PR #27, opened 2016-10-01
Move Cloud API Docs source into repo

johndmulhausen:


sanscontext:
I have a fix staged and ready to fix this PR - in the interests of not doing annoying things with Carry process, let's just merge this and apply my fix.

PR #29, opened 2016-10-03
fix bugs/html in cloud docs, remove two draft store docs

sanscontext:
Cloud: some images needed an extra carriage return so they don't follow-on in large screen resolutions.

Also (TL;DR version:fixed a bunch of HTML.)

Now, any HTML that has an argument now needs the argument enclosed in quotes. Like `colspan=2` was fine before, now must be `colspan="2"`. Similarly `target=_blank` now must be `target="_blank"`.

@johndmulhausen @joaofnfernandes and whoever else for 👀 


johndmulhausen:
Thanks a ton @sanscontext !

PR #30, opened 2016-10-04
Extend applications which suffer from poor performance

holtkamp:


londoncalling:
LGTM, there might be slight formatting/alignment problems with added items, I'll merge this then fix

londoncalling:
@holtkamp actually, looks great, thanks!

PR #33, opened 2016-10-04
Add Compose file for development

bfirsh:
What are the performance issues mentioned, out of interest?

Regeneration didn't seem to work with the command shown before because the `starefossen/github-pages` had `--force_polling` passed to the Jekyll command. This unnecessarily polled the filesystem instead of using the filesystem events in Docker for Mac/Windows. Maybe this also solves the performance issues?

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,7 +14,7 @@ You have three options:
 2.  (Slow performance on Mac/Windows, fast setup) Clone this repo, and from within the directory, run:
 
     ```bash
-    docker run -ti -v "$PWD":/usr/src/app -p "4000:4000" starefossen/github-pages
+    docker-compose up
     ```
 3.  (Edit entirely in the browser, no local clone) Fork this repo in GitHub, change your fork's repository name to `YOUR_GITHUB_USERNAME.github.io`, and make any changes.
 
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,0 +1,9 @@
+version: "2"
+services:
+  docs:
+    image: starefossen/github-pages
+    command: jekyll serve -d /_site --watch -H 0.0.0.0 -P 4000
+    ports:
+     - "4000:4000"
+    volumes:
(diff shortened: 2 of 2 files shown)


johndmulhausen:
The performance issues are just that heavy I/O operations take place through a VM layer on Mac and Windows, which aren't running the Linux containers natively. I found the `:onbuild` variant of starefossen/github-pages and made a new staging container that overrides the `CMD` so that `--force_polling` is turned off, based on your suggestions here, so thank you!

I must close this PR, but I very much appreciate the thought. If you have any thoughts on improving the staging container, edit its Dockerfile here: 

https://github.com/docker/docker.github.io/blob/master/_data/staging-container/Dockerfile

It gets uploaded to docs/docstage, here: 

https://hub.docker.com/r/docs/docstage/

bfirsh:
Why don't you want to use Compose? It's a much better solution than typing a long command or using a shell script. It's included in every Docker installation, so as far as I can see there is no downside.

johndmulhausen:
Actually that's exactly it: It isn't included in the Linux installations,
or any standalone installations that weren't done through Pinata or Toolbox.

I'm open to it if we offer both Compose and straight Docker, and link to
install instructions to help people if the command fails. Having just gone
through the rigamarole for installing Compose on an Ubuntu machine so I
could stand up the doc archives on AWS, I can tell you, it really wasn't
any more convenient than installing Jekyll natively.

https://docs.docker.com/engine/installation/linux/ubuntulinux/

johndmulhausen:
Oops wrong link:

https://docs.docker.com/compose/install/

bfirsh:
Sweet, okay. I'll add a link to the install docs.

We're trying to encourage people to use Compose as the standard way of doing development environments, and it looks a bit silly if we aren't ourselves using it. The idea is that you should be able to spin up any piece of software in development by running `docker-compose up` -- and if you see a `docker-compose.yml`, you know that'll just work.

PR #34, opened 2016-10-04
added note re: symlinks

londoncalling:
@rneugeba I updated the docs in this new comprehensive repo per https://github.com/docker/for-win/issues/109#issuecomment-251307391. Please have a look. 

(We already had this info in [OSFX topic on Mac](https://docker.github.io/docker-for-mac/osxfs/#/symlinks), but didn't add it to Windows. Let me know if the way that is written is better than what I wrote here for Windows.)

We should be serving docs.docker.com from this repo any minute now (maybe today). I'll make an announcement to the team, when that happens. In the meantime, after the PR is merged, the changes will show up on this preview site: https://docker.github.io/


londoncalling:
LGTM, merging, we can update further if needed.

PR #46, opened 2016-10-05
Exoscale: fix letter casing

marcaurele:
Following change https://github.com/docker/docker.github.io/commit/a059edd94987b0cc45a7bea536dbb98cb2380701

Signed-off-by: Marc-Aurèle Brothier <m@brothier.org>

### DIFF
--- a/machine/drivers/exoscale.md
+++ b/machine/drivers/exoscale.md
@@ -1,11 +1,11 @@
 ---
-description: exoscale driver for machine
+description: Exoscale driver for machine
 keywords:
 - machine, exoscale, driver
 menu:
   main:
     parent: smn_machine_drivers
-title: exoscale
+title: Exoscale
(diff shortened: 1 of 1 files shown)


johndmulhausen:
Thanks, @marcaurele !

CC @londoncalling FYI :)

marcaurele:
@johndmulhausen Is the change made in this PR live? In the documentation, I can still see "exoscale" in the left menu instead of "Exoscale". Is there something else I should change to get the left menu in the correct letter casing?

londoncalling:
@marcaurele that should be coming from the change in the "Title:" portion of the file in the metadata at the top. It is changed there, but not reflected in the docs, as you say. I can look into it this morning, after a meeting I have, if @johndmulhausen doesn't get to it first.

PR #47, opened 2016-10-05
Fix markdown display error

jamesmstone:
The issue occurs on the [Dockerfile reference](https://docs.docker.com/engine/reference/builder/#arg) and can be seen below:
image


jamesmstone:
I believe I fixed the issue, but my markdown viewer seemed to be displaying the original markdown correctly. So it's hard to tell

jamesmstone:
Haha there you go! The power of opensource

Issue #44, opened 2016-10-05
Feedback for: engine/reference/commandline/ps.md


sixcorners:
URL: engine/reference/commandline/ps.md

Actions are missing from the format parameters of examples on docs.docker.com.
I see `$ docker ps --format "table \t"` on the website and `$ docker ps --format "table {{.ID}}\t{{.Labels}}"` when looking the file being rendered by github. inspect.md has the same problem.

iBobik:
It is probably because `{{ }}` is also syntax for Liquid, what is run by Jekyll, base of Github Pages.

Workaround:
```
{% raw %}
$ docker ps --format "table {{.ID}}\t{{.Labels}}"
{% endraw %}
```

iBobik:
I asked in Jekyll forum for a better solution: https://talk.jekyllrb.com/t/how-to-turn-off-liquid-in-md/3078

johndmulhausen:
Thanks for asking them for their thoughts, @iBobik . Looks like the only other option is crazy syntax such as {{ "string" }}}} which works in spite of intent/implementation rather than because of it. I tried {{ content | markdownify }} and it not only didn't solve the issue, it increased the build time by over 300%.

I think we're stuck using {% raw %}, which #48 applied. While that PR did have too wide of a scope on the pages, that's always something we can adjust if Liquid ever needs to be added to those 35 files. 

If you ever dig up a novel solution, let us know!

PR #48, opened 2016-10-05
Fix formatting for Markdown files containing `{{`

benizi:
They're usually examples of Go-style text templates.  But, Liquid templating is treating them as variable substitutions.  E.g., in `engine/reference/commandline/ps.md`:

```bash
$ docker ps --filter volume=remote-volume --format "table {{.ID}}\t{{.Mounts}}"
```

is rendered as:

```bash
$ docker ps --filter volume=remote-volume --format "table \t"
```

Solution is to wrap entire files in `{% raw %} ... {% endraw %}` blocks.


benizi:
I think this fixes docker/docker.github.io#44.

Not sure whether the (much-)simpler option of changing `{{ content }}` to `{{ content | markdownify }}` in `_layouts/docs.html` would work, since there are some Markdown files that contain legitimate usage of templating directives.

johndmulhausen:
I'd prefer that we put {% raw %} and {% endraw %} around every instance of a {{ bracketed }} string, like so:

`{% raw %}{{ foobar }} {% endraw %}`

Or at least, just the affected code blocks. The scope should be narrower, is what I mean.

If you wrap raw/endraw around the entire page like this, that will prevent any Liquid templating language from processing. So if anyone touches these 35 pages in the future and tries to include a file, add some logic, print a variable, etc., they're going to see the raw code instead of the interpreted/processed code.

iBobik:
> Not sure whether the (much-)simpler option of changing {{ content }} to {{ content | markdownify }} in _layouts/docs.html would work, since there are some Markdown files that contain legitimate usage of templating directives.

This if very nice solution, because docs contributors should not need to thing about Jekyll. How about to improve it by adding condition to _layouts/docs.html to turn on Luquid optionally?

Like this:
```
{% if page.use_liquid %}
  {{ content }}
{% else %}
  {{ content | markdownify }}
{% endif %}
```

And in all content files where Liquid is needed just include this on top:
```
---
use_liquid: True
---
```

johndmulhausen:
@benizi @iBobik 

Tried the markdownify at the layout level; that doesn't work, and it; also SUBSTANTIALLY increases build time. Accepting this fix for now; we'll adjust the scope of the RAW tags in the future if we have problems w/liquid.

PR #55, opened 2016-10-05
Fix broken documentation (after ARG). 

karinepires:
The documentation is broken after ARG. This modification fix it.


johndmulhausen:
Sorry, I got to #68 first. But thanks for the PR, @karinepires !

PR #68, opened 2016-10-05
don't split verbatim text over few lines

rutsky:
Should fix improperly formatted documentation page.

This page https://docs.docker.com/engine/reference/builder/#/arg looks currently like this:


johndmulhausen:
Thanks, @rutsky !

PR #69, opened 2016-10-05
Update compose-file.md

sixcorners:
Document that the journald logging driver also allows docker-compose logs to work.

### DIFF
--- a/compose/compose-file.md
+++ b/compose/compose-file.md
@@ -468,7 +468,7 @@ The default value is json-file.
     driver: "syslog"
     driver: "none"
 
-> **Note:** Only the `json-file` driver makes the logs available directly from
+> **Note:** Only the `json-file` and `journald` drivers make the logs available directly from
 > `docker-compose up` and `docker-compose logs`. Using any other driver will not
 > print any logs.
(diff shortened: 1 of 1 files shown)


johndmulhausen:
@londoncalling can you or someone on the Compose team adjudicate the claim being made here?

londoncalling:
LGTM, @sixcorners can this be merged and published now or is it for an upcoming release?

sixcorners:
https://docs.docker.com/engine/admin/logging/overview/ says journald works with docker logs so I tested it with docker-compose logs and it seems to work fine.
I think this is for already released stuff. The server version I tested with is 1.2.1 and docker-compose is 1.8.1. The client I have installed is 1.12.2-rc1 but I don't think that matters.

londoncalling:
@sixcorners , excellent! That makes it easier for me :-) I'll merge this now and you should see it published momentarily. Thanks!

londoncalling:
@sixcorners fyi, it took a minute, but it should be there now: https://docs.docker.com/compose/compose-file/#/logging (you might have to refresh your browser)

PR #73, opened 2016-10-05
Added a link at "right here" at docker for windows instructions

lack3r:
Added a link at "right here" at docker for windows instructions


johndmulhausen:
Actually, a better approach might be to just delete this entirely, if you would be so kind:

`(or [right here](https://github.com/docker/for-win/issues/new) &#9786;)`

lack3r:
Of course :) I've made another PR with this change

londoncalling:
Closing this PR without merging, as I've already merged #75

PR #84, opened 2016-10-05
Document Compose's syntax for variable defaults

nVitius:
This feature was not supported previously, but
was added in https://github.com/docker/compose/pull/3108

### DIFF
--- a/compose/compose-file.md
+++ b/compose/compose-file.md
@@ -1144,10 +1144,16 @@ port mapping is `:5000` (which is of course an invalid port mapping, and will
 result in an error when attempting to create the container).
 
 In the case of environment variables that are not set, it is also possible to
-define default values like so: `${VARIABLE:-default}`. In this case, Compose
-will use the default value if `VARIABLE` is *unset* or *empty*. If you need
-the substitution to occur only if `VARIABLE` is *empty*, use this syntax:
-`${VARIABLE-default}`.
+define default values  using one of the following syntax variants:
+
+
+* To provide a default value that will be used if `VARIABLE` is either *unset* or *empty*:
+
+  `${VARIABLE:-default}`
+
+* To provide a default value only if `VARIABLE` is *empty*:
+
+  `${VARIABLE-default}`
 
 Both `$VARIABLE` and `${VARIABLE}` syntax are supported. Extended
 shell-style features, such as `${VARIABLE/foo/bar}`, are not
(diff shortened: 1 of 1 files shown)


aanand:
Thanks! We should document both forms of the inline default syntax:

- `${VAR-default}` (only fall back if VAR is _unset_)
- `${VAR:-default}` (fall back if VAR is _unset or empty_)

johndmulhausen:
Question: Is this for upcoming Docker Compose functionality or is this functionality already released?

nVitius:
@aanand: thanks, I missed that part.
@johndmulhausen: It looks like it's slated for 1.9.

mstanleyjones:
In that case, please rebase this on the branch `vnext-compose`.

nVitius:
Okay, I rebased over the `vnext-compose` branch and added the comment regarding the different syntax.

compose/compose-file.md
```diff
@@ -1143,8 +1143,14 @@ string. In the example above, if `EXTERNAL_PORT` is not set, the value for the
 port mapping is `:5000` (which is of course an invalid port mapping, and will
 result in an error when attempting to create the container).
 
-Both `$VARIABLE` and `${VARIABLE}` syntax are supported. Extended shell-style
-features, such as `${VARIABLE-default}` and `${VARIABLE/foo/bar}`, are not
+In the case of environment variables that are not set, it is also possible to
```
mstanleyjones: Consider "you can" instead of "it is possible to".

compose/compose-file.md
```diff
@@ -1143,8 +1143,14 @@ string. In the example above, if `EXTERNAL_PORT` is not set, the value for the
...
+define default values like so: `${VARIABLE:-default}`. In this case, Compose
```
mstanleyjones: Consider rewriting this a bit:

...define default values using one of the following syntax variants.

- To provide a default value which will be used if `VARIABLE` is unset or empty:

   ```
   ${VARIABLE:-default}
   ```

- To provide a default value if `VARIABLE` is empty:
   ```bash
   ${VARIABLE-default}
   ```

Both...

londoncalling:
@mstanleyjones can you look at the updates I made here per your comments, and if this looks good, review and approve? Then either you or I can merge this...?

mstanleyjones:
LGTM, thought I would still reconsider "it is possible to"...

PR #87, opened 2016-10-05
Adding fix for builder reference markdown

emcniece:
Currently the https://docs.docker.com/engine/reference/builder/ page is broken near the end of the ARG section. This appears to be due to the inline code highlighting parsing HTML tags incorrectly - to fix this, the code has been moved to a full code block.


johndmulhausen:
Thanks, @emcniece !

PR #91, opened 2016-10-06
update attach.go filepath and code snippet

samwagg:
Doc contained out-of-date file path and code from attach.go. Updated according to attach.go's current location and code.


londoncalling:
@thaJeztah it looks like @samwagg caught some of the updates needed to the published version of these docs that you did per https://github.com/docker/opensource/pull/94#issuecomment-252894798. Can you verify and say whether this PR looks good to you? Then I'll merge it, and address your other comments after..?

PR #114, opened 2016-10-07
Prevent Liquid from mangling `{{` in a more targeted way

benizi:
Per [feedback on #48](https://github.com/docker/docker.github.io/pull/48#issuecomment-251749322), don't wrap entire files in `{% raw %} ... {% endraw %}` blocks. Instead, just wrap code blocks or paragraphs that contain `{{`.


johndmulhausen:
@benizi Thanks so much for this. Am I crazy or did we lose some of the {% raw %} directives? I don't see the word "raw" anywhere here: https://raw.githubusercontent.com/benizi/docker.github.io/362e94b2bca42da5dc833175e13d02935c26d25b/apidocs/cloud-api-source/source/includes/node.md

benizi:
Sorry, meant to mention that: the rendering pipeline for the Cloud API docs is apparently different enough that they're unnecessary for those files. (I assume it has to do with how it parses out the various toggleable, language-specific examples.)

benizi:
^^ @johndmulhausen

sanscontext:
@benizi Correct - to regen the Cloud API docs we're actually still running `make release` from the `cloud-api-source` folder, which generates a bunch of HTML files we just include. You can take a peek at the `Makefile` in `apidocs/cloud-api-source` if you're curious.

johndmulhausen:
Thanks for the clarification, that makes sense. And thanks to @sanscontext for adding context. :)

This is great, @benizi and I really appreciate you taking the time to come back to this issue a second time to address the feedback. Merging now!

PR #117, opened 2016-10-07
Add deferred deletion to recommended dm options

mbentley:
Signed-off-by: Matt Bentley <matt.bentley@docker.com>

### DIFF
--- a/engine/userguide/storagedriver/device-mapper-driver.md
+++ b/engine/userguide/storagedriver/device-mapper-driver.md
@@ -314,10 +314,10 @@ assumes that the Docker daemon is in the `stopped` state.
 
 14. Configure the Docker daemon with specific devicemapper options.
 
-    There are two ways to do this. You can set options on the command line if you start the daemon there:
+    Now that your storage is configured, configure the Docker daemon to use it. There are two ways to do this. You can set options on the command line if you start the daemon there:
 
     ```bash
-    --storage-driver=devicemapper --storage-opt=dm.thinpooldev=/dev/mapper/docker-thinpool --storage-opt dm.use_deferred_removal=true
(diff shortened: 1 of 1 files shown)


engine/userguide/storagedriver/device-mapper-driver.md
```diff
@@ -314,10 +314,10 @@ assumes that the Docker daemon is in the `stopped` state.
 
 14. Configure the Docker daemon with specific devicemapper options.
 
-    There are two ways to do this. You can set options on the command line if you start the daemon there:
+    Now that your storage has been configured, we have to instruct the Docker daemon to use it.  There are two ways to do this. You can set options on the command line if you start the daemon there:
```
mstanleyjones: Please avoid the use of second person ("we"). I suggest "Now that your storage is configured, configure the Docker daemon to use it."

Nit: Please use only a single space between sentences.

engine/userguide/storagedriver/device-mapper-driver.md
```diff
@@ -327,11 +327,14 @@ assumes that the Docker daemon is in the `stopped` state.
...
        "storage-opts": [
          "dm.thinpooldev=/dev/mapper/docker-thinpool",
-         "dm.use_deferred_removal=true"
+         "dm.use_deferred_removal=true",
+         "dm.use_deferred_deletion=true"
        ]
     }
     ```
 
+    >**Note**: It is suggested to set both `dm.use_deferred_removal=true` and `dm.use_deferred_deletion=true` to prevent unintentional leaking of mount points.
```
mstanleyjones: Avoid "It is suggested". Instead, try something like "Always set both ....to prevent unintentional leaking of mount points."

mbentley:
@mstanleyjones Thanks for the feedback! Updated my PR and squashed my commits.

PR #119, opened 2016-10-07
add leading url to feedback link email template

sanscontext:
For @johndmulhausen  - this just adds the domain name, because I can't figure out how to rewrite the `page.path` to replace `.md` with `/` to make it an actual correct url. What I was going for is to make the "URL" part of any ticket filed a live link to the page, rather than a flat file path.

Feel free to carry or close, or point me at something I could reverse-engineer.


sanscontext:
Sweet, didn't know about that variable. Updated, PTAL.

johndmulhausen:
Looks great, @sanscontext ! Thanks a ton :)

PR #121, opened 2016-10-07
remove prefix space

BSCheshir:
strange space on ALL code block on https://docs.docker.com/compose/gettingstarted/

github clear it on preview .md


johndmulhausen:
Thanks @BSCheshir! :)

PR #124, opened 2016-10-08
Update step_four.md

RStrydom:
It is not normal to provide the file you are looking for when running `ls`


johndmulhausen:
Thanks, @RStrydom :)

PR #125, opened 2016-10-08
Wrong destination folder in container

sicwolf:
/opt/webapp of web container includes app.py, if set destination folder as /opt/webapp, app.py is not visible and below error log presents and the web container will exit, so it's better to set the destination folder as /webapp. 

_docker logs -f web
python: can't open file 'app.py': [Errno 2] No such file or directory_

"docker run -d -P --name web -v /src/webapp:/opt/webapp:ro training/webapp python app.py" this line aslo shall be update as "docker run -d -P --name web -v /src/webapp:/webapp:ro training/webapp python app.py"

### DIFF
--- a/engine/tutorials/dockervolumes.md
+++ b/engine/tutorials/dockervolumes.md
@@ -99,7 +99,7 @@ In addition to creating a volume using the `-v` flag you can also mount a
 directory from your Docker engine's host into a container.
 
 ```bash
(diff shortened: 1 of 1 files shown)


sicwolf:
Dear mstanleyjones,

All possible wrong /opt/webapp already changed in dockervolumes.md, could you pls. review the two commits related to this issue?
thank you so much for your review on my first contribution on docker project!

PR #139, opened 2016-10-10
fix formatting in container-communication.md

vishesh92:
fix formatting in container-communication.md so page is rendered perfectly at https://docs.docker.com/engine/userguide/networking/default_network/container-communication/


johndmulhausen:
Thanks @vishesh92 !

PR #141, opened 2016-10-10
Fix incorrect information about swarm port assignment

mstanleyjones:
The docs imply that services can only use ports in the 30000-32767. In fact, you can specify any port, but if you don't specify one, a port in that range is assigned.


mstanleyjones:
@thaJeztah @dhiltgen PTAL

thaJeztah:
Do we know already if commits need a sign-off in this repository btw?

johndmulhausen:
@thaJeztah They don't at the moment, we need to get that set up

PR #142, opened 2016-10-10
Small update to the scale documentation

drlukeangel:
Just a small update to make it a bit clearer with a bit more functionality.

### DIFF
--- a/engine/reference/commandline/service_scale.md
+++ b/engine/reference/commandline/service_scale.md
@@ -25,8 +25,8 @@ Options:
 
 ### Scale a service
 
-If you scale a service, you set the *desired* number of replicas. Even though
-the command returns directly, actual scaling of the service may take some time.
+Scales a service both up and down to the desired number of replicas. Even though the command returns directly, actual scaling of the service may take some time. To keep a service active but have no instances of it you can scale it to 0.
+
 
 For example, the following command scales the "frontend" service to 50 tasks.
 
--- a/engine/reference/commandline/service_scale.md
+++ b/engine/reference/commandline/service_scale.md
@@ -25,7 +25,7 @@ Options:
 
 ### Scale a service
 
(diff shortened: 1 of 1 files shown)


engine/reference/commandline/service_scale.md
```diff
@@ -25,8 +25,8 @@ Options:
 
 ### Scale a service
 
-If you scale a service, you set the *desired* number of replicas. Even though
-the command returns directly, actual scaling of the service may take some time.
+Scales a service both up and down to the desired number of replicas. Even though the command returns directly, actual scaling of the service may take some time. To keep a service active but have no instances of it you can scale it to 0.
```
mstanleyjones: Change "both up and down" to "either up or down" as you really can't do both at the same time.

Maybe reword the second part to something like "The command returns immediately, but the service may take some time to finish scaling."

Maybe "You can set a service's scale to 0 to keep the service active but running zero instances." Maybe say why you would do that, too.

mstanleyjones:
This time I will squash and merge your changes. In the future, please squash them and make sure there is only one commit per committer per PR. Thanks!

drlukeangel:
Hey @mstanleyjones will do on the squash next time it wasn't documented to do that anywhere but no biggie.  Did you know can force users to do that on github https://github.com/blog/2141-squash-your-commits might be helpful. Cheers

mstanleyjones:
Yeah, the thing is that only allowing squashed commits doesn't work nicely with the new feature of being able to collaborate in a PR. If you push a commit and i push a commit, they shouldn't be squashed together, because that obscures who did the work. But if you pushed 3 commits and I pushed 1, your 3 should be squashed if possible. Make sense? 

PR #172 is open to add issue and PR templates. Let us know if you have suggestions for what should be in the templates, beyond what is there now.

PR #148, opened 2016-10-10
Clarify what happens when a failed swarm node recovers

### LINKED ISSUE #105: Not documented: Restore containers on crashed node in Swarm
I could not find the docs in `engine/swarm/admin_guide.md` for this scenario:

I am looking for a command that allows me to restore the containers to a node that had crashed and is now back up and running.
This node does not get the previous containers be default - how can I trigger this?

mstanleyjones:
Fixes #105

### DIFF
--- a/engine/swarm/admin_guide.md
+++ b/engine/swarm/admin_guide.md
@@ -262,3 +262,14 @@ The `--force-new-cluster` flag puts the Docker Engine into swarm mode as a
 manager node of a single-node swarm. It discards swarm membership information
 that existed before the loss of the quorum but it retains data necessary to the
 Swarm such as services, tasks and the list of worker nodes.
+
+### Joining a previously failed node
+
+If a node becomes unavailable, it cannot communicate with the rest of the swarm
+and its workload is redistributed among the other nodes.
+If access to that node is restored, it will join the swarm automatically, but it
+will join with no workload because the containers it was assigned have been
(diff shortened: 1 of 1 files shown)


mstanleyjones:
@thaJeztah PTAL

engine/swarm/admin_guide.md
```diff
@@ -262,3 +262,14 @@ The `--force-new-cluster` flag puts the Docker Engine into swarm mode as a
 manager node of a single-node swarm. It discards swarm membership information
 that existed before the loss of the quorum but it retains data necessary to the
 Swarm such as services, tasks and the list of worker nodes.
+
+### Re-joining a previously failed node
+
+If a node becomes unavailable, it cannot communicate with the rest of the swarm.
+If access to that node is restored, it will re-join the swarm, but it will
+function as a completely new node. In fact, when it first re-joins, its current
```
aaronlehmann: I agree. Let's not say that re-joins the swarm, because according to our terminology, it never really left; it was just unreachable for a period of time. During that time, its workload would have been rescheduled elsewhere.

mstanleyjones:
Thanks for the feedback. I've reworded it a bit. Can y'all review it again?

aaronlehmann:
I don't see the changes - looks like they haven't been pushed yet.

mstanleyjones:
Oops, you are right! Should be pushed now.

aaronlehmann:
LGTM

PR #149, opened 2016-10-10
Provide a backup for old config when configuring devicemapper

### LINKED ISSUE #19: docs: improve devicemapper configuration instructions
transferred to here from https://github.com/docker/docker/issues/27020

https://docs.docker.com/engine/userguide/storagedriver/device-mapper-driver/#/for-a-direct-lvm-mode-configuration

This doc informs users to push all images to a remote registry and then informs to rm -f /var/lib/docker, thus removing everything including certs that may be stored under this directory tree. For good reasons, we instruct users to change from the default devicemapper loopback configuration that we recommended in early releases, of which they maybe running UCP on those nodes. Deleting the certs in /var/lib/docker/discovery_certs creates a far more complex recovery scenario than just copying the certs back after the lvm changes have been completed. This is the error received when you try to start up the docker daemon after following the doc instructions on a UCP node.

Initializing discovery with TLS
Error starting daemon: discovery initialization failed (Could not read CA certificate \ "/var/lib/docker/discovery_certs/ca.pem\" open /var/lib/docker/discovery_certs/ca.pem: no such file or directory

Can additional steps be written into the referenced doc perhaps such as copy the /var/lib/docker directory to /tmp to enable the recovery of items such as certs, etc in order to keep from placing users into extensive recovery scenarios. thnx

mstanleyjones:
Fixes #19

### DIFF
--- a/engine/userguide/storagedriver/device-mapper-driver.md
+++ b/engine/userguide/storagedriver/device-mapper-driver.md
@@ -303,13 +303,18 @@ assumes that the Docker daemon is in the `stopped` state.
     $ lvs -o+seg_monitor
     ```
 
-13. If the Docker daemon was previously started, clear your graph driver directory.
+13. If the Docker daemon was previously started, move your existing graph driver
+    directory out of the way.
 
-    Clearing your graph driver removes any images, containers, and volumes in your
-    Docker installation.
+    Moving the graph driver removes any images, containers, and volumes in your
+    Docker installation. These commands move the contents of the
+    `/var/lib/docker` directory to a new directory named `/var/lib/docker.bk`.
+    If any of the following steps fail and you need to restore, you can remove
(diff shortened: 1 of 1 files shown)


mstanleyjones:
@bgrissin PTAL

mstanleyjones:
Thanks for that, I wasn't sure where to recommend the deletion step. Let me move it around a bit.

mstanleyjones:
Actually, looking more closely, I think this is actually what I do recommend. You start Docker in step 16 (the last step of the procedure). After that, you run the `journald` command to verify that everything is fine. Only then do I tell you to delete the backup directory. Can you let me know where you think it should go instead?

mstanleyjones:
I'm going to go ahead and merge this. Thanks for the review.

PR #150, opened 2016-10-10
node1

BSCheshir:
little bit before
ID                         NAME      IMAGE  NODE   DESIRED STATE  CURRENT STATE               ERROR
63s86gf6a0ms34mvboniev7bs  my-web.1  nginx  node1  Running        Running 58 seconds ago


johndmulhausen:
So it is! Thanks, @BSCheshir  :)

PR #151, opened 2016-10-10
raw broken

BSCheshir:
https://docs.docker.com/engine/swarm/networking/#/use-swarm-mode-service-discovery has broken (may be on "{% raw %}...{% endraw %}" tag)


johndmulhausen:
Fixed in 1dc817611b6846a64ed1bb48201cff3702d33814

PR #152, opened 2016-10-10
Change 'alias:' to 'redirect_from:' so aliases work again

mstanleyjones:
@londoncalling @johndmulhausen PTAL, I think this is why our aliases don't work. :)

### DIFF
--- a/compose/compose-file.md
+++ b/compose/compose-file.md
@@ -1,5 +1,5 @@
 ---
-aliases:
+redirect_from:
 - /compose/yml
 description: Compose file reference
 keywords:
--- a/compose/link-env-deprecated.md
+++ b/compose/link-env-deprecated.md
@@ -1,5 +1,5 @@
 ---
-aliases:
+redirect_from:
 - /compose/env
 description: Compose CLI reference
 keywords:
--- a/compose/reference/overview.md
+++ b/compose/reference/overview.md
@@ -1,5 +1,5 @@
 ---
-aliases:
+redirect_from:
 - /compose/reference/docker-compose/
 description: Overview of docker-compose CLI
 keywords:
--- a/cs-engine/index.md
+++ b/cs-engine/index.md
@@ -1,5 +1,5 @@
 ---
-aliases:
+redirect_from:
 - /docker-trusted-registry/cs-engine/
 description: Learn more about the Commercially Supported Docker Engine.
 keywords:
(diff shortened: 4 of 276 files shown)


joaofnfernandes:
Just for reference, from https://github.com/jekyll/jekyll-redirect-from, it seems that the redirect paths should end with a trailing slash.
I noticed that `compose/link-env-deprecated.md` doesn't have one.

mstanleyjones:
@johndmulhausen let me know that he thought we should still be using the alias functionality rather than redirect_from, and that there might be a bug that is causing the redirect problem that Vicky reported elsewhere. I'd like him to look at it when he comes back from vacation before we decide whether we want to make this change or fix that bug. Maybe he can explain to us why we shouldn't use redirect_from. I think it actually creates the old HTML file with a 302 redirect to the new one, and maybe he thought it would make Jenkins take longer. But let's let him give more details.

compose/compose-file.md
```diff
@@ -581,10 +581,10 @@ In the example below, three services are provided (`web`, `worker`, and `db`), a
         image: mysql
         networks:
           new:
-            aliases:
+            redirect_from:
               - database
           legacy:
-            aliases:
+            redirect_from:
```
johndmulhausen: Erroneous replacement

docker-for-windows/menu.md
```diff
@@ -1,5 +1,5 @@
 ---
-aliases: []
+redirect_from: []
```
johndmulhausen: Remove

engine/getstarted/menu.md
```diff
@@ -1,5 +1,5 @@
 ---
-aliases: []
+redirect_from: []
```
johndmulhausen: Remove

johndmulhausen:
Added some comments; will be looking at build times. This might be preferred after all so we stop throwing 404s

mstanleyjones:
@johndmulhausen PTAL, I think this is ready. FYI, here is the magic command I used:

```
find . -type f -name "*.md"|while read i; do sed -i '' -e '1,10 s/aliases:/redirect_from:/' "$i"; done
```

This only replaces `alias:` with `redirect_from:` if it occurs in the first 10 lines of the file.

johndmulhausen:
Made a few nips/tucks for consistency, but this looks great, thank you!

PR #155, opened 2016-10-10
Change 'draft: true' to 'published: false' for Jekyll

mstanleyjones:
Hugo uses "draft=true|false" but Jekyll uses "published: true|false" (the opposite Boolean), so some things that should have been drafts were published. This fixes that. The files were all already public, they were just not ready to be on the website yet.


mstanleyjones:
@johndmulhausen PTAL. I verified this by building locally.

johndmulhausen:
Thanks, @mstanleyjones !

PR #159, opened 2016-10-11
Update 3_prepare_the_app.md

thaxy:
The directory name for the dockercloud-quickstart-go repository is wrong.
The change directory command wouldn't work if copy pasted.


thaxy:
Ahh there is a second cd command. Forgot to add the right directory path to the 2nd bash box.

PR #160, opened 2016-10-11
Update 3_prepare_the_app.md

thaxy:
Wrong directory names for the repository. Change directory command wouldn't work if copy pasted.


thaxy:
I am not sure if you should change the tags according to the repository with the 'dockercloud' prefix.

johndmulhausen:
Yup, makes sense. Thanks, @thaxy !

PR #166, opened 2016-10-11
Update Go API client to docker/docker/client

bfirsh:
This has moved from docker/engine-api. See https://github.com/docker/docker/pull/27232

This page needs an overhaul at some point, but this is just fixing this particular bug.


johndmulhausen:
Thanks, @bfirsh !

PR #171, opened 2016-10-11
Update the contribution guide for new docs processes

mstanleyjones:
Provide quick guidelines in the README and update the opensource contribution guide a bit. @johndmulhausen PTAL

### DIFF
--- a/opensource/project/test-and-docs.md
+++ b/opensource/project/test-and-docs.md
@@ -59,50 +59,31 @@ Docker Engine test suite.  The `Makefile` contains a target for the entire test 
 The target's name is simply `test`. The `Makefile` contains several targets for
 testing:
 
-<style type="text/css">
-.monospaced {font-family: Monaco, Consolas, "Lucida Console", monospace !important;}
-</style>
-<table>
-  <tr>
-    <th>Target</th>
-    <th>What this target does</th>
-  </tr>
-  <tr>
-    <td class="monospaced">test</td>
-    <td>Run the unit, integration and docker-py tests.</td>
-  </tr>
-  <tr>
-    <td class="monospaced">test-unit</td>
-    <td>Run just the unit tests.</td>
-  </tr>
-  <tr>
-    <td class="monospaced">test-integration-cli</td>
-    <td>Run the test for the integration command line interface.</td>
-  </tr>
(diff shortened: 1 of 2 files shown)


opensource/project/test-and-docs.md
```diff
@@ -239,99 +198,120 @@ Git for Windows installation.  **Git Bash**, just as it sounds, allows you to
...
-    ![Beta documentation](images/red_notice.png)
-
-6. Navigate to your new or changed document.
-
-7. Review both the content and the links.
-
-8. Return to your terminal and exit out of the running documentation container.
+can build the documentation from the local host.
+
+### Building the docs for docs.docker.com
```
joaofnfernandes: I think @johndmulhausen had containerized Jekyll so that you would not need to install anything other than Docker.
I think you only needed

```
docker run -ti -–rm -v "$PWD":/docs -p 4000:4000 docs/docstage
```

mstanleyjones: Cool, I didn't know about that. That doesn't work on Mac or Windows, but `docker-compose up` does. I have documented both as well as the drawbacks of using Docker vs using Jekyll directly (as they are in the README).

joaofnfernandes: If `docker-compose up` works for all platforms, then maybe we should just present two options: fork from github, or use docker-compose. What do you think?
You're saying in readme.md that the `docker-compose up` is slower, but is it by a lot?

joaofnfernandes:
LGTM 🚀

mstanleyjones:
@joaofnfernandes I just pushed an amendment that re-orders the methods for building the site in the README and adds some more explanation to the file in the contribution guide. I also removed the method that used `docker` in preference for normalizing on `docker-compose`.

opensource/project/test-and-docs.md
joaofnfernandes: I think you're right, my feedback is not a blocker so let's merge this and iterate once we have more feedback.

About people that might want to use Jekyll in other repos, I think it's the other way around. By using docker we're lowering the barrier to contribute with docs. Of course there are people that might want to use Jekyll in multiple projects, but those are not the majority.

PR #174, opened 2016-10-12
This guide used legacy backticks.

IRus:
https://github.com/koalaman/shellcheck/wiki/SC2006
https://github.com/koalaman/shellcheck/wiki/SC2046

### DIFF
--- a/compose/install.md
+++ b/compose/install.md
@@ -38,7 +38,7 @@ which the release page specifies, in your terminal.
 
      The following is an example command illustrating the format:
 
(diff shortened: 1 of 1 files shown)


johndmulhausen:
@docker/core-compose-maintainers Is this a canonical way to get Docker Compose?

johndmulhausen:
By "canonical" I mean, is this a sustainable, best-practice way to get Docker Compose that the Compose team would recommend we provide to users?

shin-:
@johndmulhausen Yes, this is the way we have recommended users install Compose for a long time.

johndmulhausen:
@shin- Thanks so much! Merging now. :)

PR #178, opened 2016-10-12
Additional information about scaling a service

mstanleyjones:
Adding additional information as a follow-on to #148. Related to #105.

### DIFF
--- a/engine/swarm/admin_guide.md
+++ b/engine/swarm/admin_guide.md
@@ -21,15 +21,15 @@ maintain the swarm.
 
 This article covers the following swarm administration tasks:
 
-* [Using a static IP for manager node advertise address](admin_guide.md#use-a-static-ip-for-manager-node-advertise-address)
-* [Adding manager nodes for fault tolerance](admin_guide.md#add-manager-nodes-for-fault-tolerance)
-* [Distributing manager nodes](admin_guide.md#distribute-manager-nodes)
-* [Running manager-only nodes](admin_guide.md#run-manager-only-nodes)
-* [Backing up the swarm state](admin_guide.md#back-up-the-swarm-state)
-* [Monitoring the swarm health](admin_guide.md#monitor-swarm-health)
-* [Troubleshooting a manager node](admin_guide.md#troubleshoot-a-manager-node)
-* [Forcefully removing a node](admin_guide.md#force-remove-a-node)
-* [Recovering from disaster](admin_guide.md#recover-from-disaster)
+* [Using a static IP for manager node advertise address](#use-a-static-ip-for-manager-node-advertise-address)
+* [Adding manager nodes for fault tolerance](#add-manager-nodes-for-fault-tolerance)
+* [Distributing manager nodes](#distribute-manager-nodes)
+* [Running manager-only nodes](#run-manager-only-nodes)
+* [Backing up the swarm state](#back-up-the-swarm-state)
+* [Monitoring the swarm health](#monitor-swarm-health)
+* [Troubleshooting a manager node](#troubleshoot-a-manager-node)
+* [Forcefully removing a node](#force-remove-a-node)
+* [Recovering from disaster](#recover-from-disaster)
 
 Refer to [How nodes work](how-swarm-mode-works/nodes.md)
 for a brief overview of Docker Swarm mode and the difference between manager and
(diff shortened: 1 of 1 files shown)


engine/swarm/admin_guide.md
```diff
@@ -263,13 +263,25 @@ manager node of a single-node swarm. It discards swarm membership information
...
-and its workload is redistributed among the other nodes.
-If access to that node is restored, it will join the swarm automatically, but it
-will join with no workload because the containers it was assigned have been
-reassigned. The node will only receive new workloads when the swarm is rebalanced.
-To force the swarm to be rebalanced, you can
-[update](../reference/commandline/service_update/) or
-[scale](../reference/commandline/service_scale/) the service.
+### Re-balancing a service after joining a new or previously failed node
+
+When you add a new node to a swarm, or a node re-joins after it has been
```
thaJeztah: Would "reconnect" instead of "re-join" be an option?

mstanleyjones: I'm not in love with that, because the verb in the CLI is 'join' rather than 'connect'. In fact, I should probably change the 'add' to 'join' for consistency. I could probably just change 're-join' to 'join'. WDYT?

thaJeztah: Hm, actually, that is the issue, because the node never "left" the swarm, it was only unreachable. When it becomes reachable again, it doesn't "join", it just erm, "dunnowhattocallit".

_joining_ a swarm generates a new cryptographic identity, which isn't the case here.

aaronlehmann: @thaJeztah is correct.`reconnect` sounds right to me. Otherwise, "...or a node returns..."?

stevvooe: `reconnect` or `re-register`, but I am not sure if we use that terminology consistently.

engine/swarm/admin_guide.md
```diff
@@ -263,13 +263,25 @@ manager node of a single-node swarm. It discards swarm membership information
...
+unavailable, the new node does not automatically get a workload if the service
+is already running at the desired scale. Notably, when a failed node recovers
+and re-joins a swarm, the workloads it was previously running have been
+reassigned to other nodes, and it does not automatically take them back.
```
stevvooe: Might be good to include reasoning behind _why_ this is the case, as I've found this behavior confusterates people.

@aluzzardi can give you more, but we don't place workload on the new server to prevent healthy services from being interrupted and to avoid dog piling newly joined nodes.

mstanleyjones:
Thanks all, this is fantastic info and I will add it to the doc here.

mstanleyjones:
OK, I tried to capture the feedback given by @aaronlehmann @stevvooe @aluzzardi . PTAL, thanks!

stevvooe:
LGTM

engine/swarm/admin_guide.md
```diff
@@ -263,13 +263,32 @@ manager node of a single-node swarm. It discards swarm membership information
...
+the idle node. This is a design decision. If the swarm periodically shifted tasks
+to different nodes for the sake of balance, the clients using those tasks would
+be disrupted. The goal is to avoid disrupting running services for the sake of
+balance across the swarm. When new tasks start, or when a node with running
+tasks becomes unavailable, those tasks are given to less busy nodes. The goal
+is eventual balance, with minimal disruption to the end user.
+
+If you are concerned about an even balance of load and don't mind disrupting
+running tasks, you can force your swarm to re-balance by temporarily scaling
+the service upward.
```
thaJeztah: Note that this PR, if accepted, would probably allow rebalancing without having to change the scale; https://github.com/docker/swarmkit/pull/1664

mstanleyjones: But I can't talk about it until / unless it is. :)

engine/swarm/admin_guide.md
```diff
@@ -263,13 +263,32 @@ manager node of a single-node swarm. It discards swarm membership information
...
+under-loaded nodes in your swarm. You may need to scale the service up by modest
+increments a few times to achieve the balance you want across all the nodes.
+
+When the load is balanced to your satisfaction, you can scale the service back
+down to the original scale. You can use `docker service ps` to assess the current
+balance of your service across nodes.
+
+See also
+[`docker service scale`](../reference/commandline/service_scale/) and
+[`docker service ps`](../reference/commandline/service_ps/).
```
mstanleyjones: Done, and also took the file portion out of the in-file links earlier up in the file.

mstanleyjones:
With the +1 from @stevvooe I'm going to merge.

engine/swarm/admin_guide.md
thaJeztah: I know; it was just a heads-up :smile:

PR #179, opened 2016-10-12
Fix broken links

QwertyZW:
Fixed a link pointing to the same page and 2 other links that are returning 404s

### DIFF
--- a/docker-hub/bitbucket.md
+++ b/docker-hub/bitbucket.md
@@ -13,7 +13,7 @@ title: Automated Builds with Bitbucket
 # Automated Builds with Bitbucket
 
 If you've previously linked Docker Hub to your Bitbucket account,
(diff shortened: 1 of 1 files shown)


sanscontext:
Hey @QwertyZW thanks for the head's up. That first link is actually linking to a subsection of the doc, and is correct as it was. (It should jump down the page.) Those other two are broken, you're correct, but were not intended to be links to the documentation - rather directly into the Hub UI. They should both go to `https://hub.docker.com/add/automated-build/bitbucket/`

sanscontext:
Hi @QwertyZW, do you want to edit your patch and re-push? :) Or do you want to close it and I'll make the correction in the next batch of Hub updates?

johndmulhausen:
@sanscontext You can actually edit his patch directly if you click the pencil in the "files" tab in this PR. :)

sanscontext:
Fixed the links. Merging.

PR #172, opened 2016-10-11
Add github templates

mstanleyjones:
One template for issues, one for pull requests,
and a CONTRIBUTING.md with some of the same
info as the README

PR #184, opened 2016-10-13
Deprecate repo:shortid syntax

tonistiigi:
https://github.com/docker/docker/pull/27207

Signed-off-by: Tonis Tiigi <tonistiigi@gmail.com>

### DIFF
--- a/engine/deprecated.md
+++ b/engine/deprecated.md
@@ -17,6 +17,13 @@ To learn more about Docker Engine's deprecation policy,
 see [Feature Deprecation Policy](index.md#feature-deprecation-policy).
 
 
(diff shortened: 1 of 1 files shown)


thaJeztah:
@mstanleyjones do you know if we'll be adding milestones / labels, to make it easier to find which release a documentation change relates to (in future)?

mstanleyjones:
@thaJeztah I think it could be a nightmare to use tags for that, but there is a pending PR #172 which would allow you to capture it in the issue or PR template. If we really needed it, we could make a versioned vnext branches. I asked @johndmulhausen about that but he preferred to try without it, for simplicity's sake.

PR #192, opened 2016-10-13
[Docs] UCP needs TCP opened for port 4789

vsaraswat:
It looks like UCP requires port TCP 4789 opened. Docs currently state just UDP, so made a quick addition to this page to add in TCP.


joaofnfernandes:
LGTM

PR #193, opened 2016-10-13
Update version to  the latest

horalstvo:
I installed docker-machine today to later find out that I just got old version. I'm new to Docker, though would it make sense to update the version?

Also don't know how to do it on github directly, but I forgot to modify the version hash. So the last change should be to 0.8.2, build e18a919.

### DIFF
--- a/machine/install-machine.md
+++ b/machine/install-machine.md
@@ -29,13 +29,13 @@ If you want only Docker Machine, you can install the Machine binaries directly b
 
     If you are running OS X or Linux:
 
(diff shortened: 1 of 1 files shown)


joaofnfernandes:
Thanks for contributing! Let's see if you can fix that version hash part yourself.

Can you [try editing this page](https://github.com/horalstvo/docker.github.io/blob/patch-1/machine/install-machine.md) with the github edit functionality? I think that should do the trick.

johndmulhausen:
Ping @horalstvo

johndmulhausen:
Looks like someone already made this update elsewhere. Thanks!

PR #194, opened 2016-10-13
Update networkingcontainers.md

DariusX:
Changed    --network=    to   --net


joaofnfernandes:
Thanks for contributing. In docker/docker#23324 the `--network` flag replaced `--net`. While `--net` is still available for backwards compatibility, we recommend using `--network`.

Issue #189, opened 2016-10-13
github markdown not processed the same as docker.github.io markdown


andrewhsu:
The way github processes markdown and shows the final rendered HTML is not the same as how the markdown is processed to show HTML on the final docker.github.io site.

For example, if you see the items in the section titled "Task 2" the ordered list is numbered correctly and the blocks of code are in their own grey background box:
https://github.com/docker/docker.github.io/blob/d53c6798c9f6b7046ae04277eb4d25ccffc1f0af/opensource/project/set-up-dev-env.md

However, the markdown is rendered on the final website with grey boxes behind each line. I think the problem starts at ordered list item number 2 in the section titled "Task 2":
https://docker.github.io/opensource/project/set-up-dev-env/

sanscontext:
Hey @andrewhsu, thanks for the report. It looks like there's a set of code-block gates that's improperly closed, and that's causing the renderer some pain. This should be a quick fix.

andrewhsu:
Is there a way to have the docker.github.io markdown processed the same way that github processes the markdown? It would make it convenient for people submitting PRs to preview the changes before it is actually rendered on the public web site.

Or...is there a way to preview the changes of a PR before it is pushed to production?

mstanleyjones:
Github.com uses a totally different stylesheet to preview Markdown and it is very simplistic. To output docs.docker.com (and docker.github.io, since they are currently one and the same), we use Jekyll, which is a little more picky about indentation in lists. You can preview the docs by using `docker compose up` in the root of the docker.github.io repo. There are a couple of other ways to do it too. See the README at the top level of this repo for more details about them.

sanscontext:
Hey @andrewhsu , I'll close this out, since it's fixed. (It took a second shot on my account - it turns out the new renderer is picky in ways I hadn't anticipated.) The page is a *lot* more readable now, and I suspect there are other pages with similar issues. Please keep reporting these if you see them!

PR #200, opened 2016-10-14
actually fix broken gates this time

sanscontext:
For #189  *again* - turns out it was Kramdown's inability to parse things in an ordered list that aren't indented exactly the same number of spaces. Also reverts non-useful `bash` highlighting.

@mstanleyjones @joaofnfernandes  (sorry for the double-tap)

### DIFF
--- a/opensource/project/set-up-dev-env.md
+++ b/opensource/project/set-up-dev-env.md
@@ -39,55 +39,55 @@ To remove unnecessary artifacts:
 
 1. Verify that you have no unnecessary containers running on your host.
 
-    ```bash
-    $ docker ps -a
-    ```
+   ```none
+   $ docker ps -a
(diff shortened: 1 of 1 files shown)


mstanleyjones:
Did you build it locally and verify that it's OK? You can also tell if something is going to work in Atom, by looking at the folding indicators and making sure everything lines up. Jekyll is much more picky than Hugo, that's for sure.

sanscontext:
I did. Turns out, Atom's standard markdown previewer gives a false positive result, and the Atom `markdown-kramdown-preview` plugin is too broken to be helpful.  I did `jekyll serve` instead.

joaofnfernandes:
Looks like it's fixed. LGTM 🚀

Issue #208, opened 2016-10-14
Enforce no local editing of files that are maintained elsewhere


mstanleyjones:
### Problem description

We want reference docs to be maintained in their code repos, along with
the code they are referencing. We want to periodically bring them up to date
in this repo using automated processes.

Two things need to happen:
- Create a manifest of directories and files that should not be the target of pull requests in this repo: Done in PR #209.
- Enforce that manifest using CI, on each pull request

The automated refreshing of the reference docs in this repo is out of scope for this issue, but
a second issue can be filed as a follow-up to this one.

/cc @thaJeztah @stevvooe @mikedougherty @johndmulhausen

mstanleyjones:
We did this by removing those files from our repo and getting them at build time.

PR #209, opened 2016-10-14
Add manifest of files maintained elsewhere

mstanleyjones:

Adds a manifest of files which should not be edited in this repo because they are
maintained elsewhere. The intent is for this file to be used by a CI process to fail
a PR which changes one of these files.

### Project version

This change will need to go into all the vnext branches.

### Related issue

#208 

https://github.com/docker/distribution/pull/1985

### Please take a look

/cc @stevvooe @thaJeztah @johndmulhausen and please CC others who will know which files or directories should be excluded from the other projects.

The plan is for this file to be used by a CI process which
will fail a PR if it includes changes to any of these files.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -2,8 +2,13 @@
 guidelines to help us review it. You can remove these comments
 as you fill out this template. -->
 
+<!--DO NOT edit files and directories listed in .NOT_EDITED_HERE.txt
+as these are maintained in upstream repos and added to the central
+docs repo periodically. Your PR will fail automated checks if it
+includes changes to these files or directories.-->
+
 <!-- **Squash your commits**: Please use an
-[interactive rebase](https://help.github.com/articles/about-git-rebase/)
+interactive rebase (https://help.github.com/articles/about-git-rebase/)
 to squash (combine) multiple commits into a single commit, to keep the
 commit history clean. The exception is when commits from multiple
 contributors exist in the same pull request. In that case, one commit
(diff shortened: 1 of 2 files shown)


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -2,8 +2,13 @@
...
 as you fill out this template. -->
 
+<!--DO NOT edit files and directories listed in _NOT_EDITED_HERE.txt
+as these are maintained in upstream repos and added to the central
+docs repo periodically. Your PR will fail automated checks if it
+includes changes to these files or directories.-->
+
 <!-- **Squash your commits**: Please use an
-[interactive rebase](https://help.github.com/articles/about-git-rebase/)
+interactive rebase (https://help.github.com/articles/about-git-rebase/)
```
thaJeztah: oh, not related, but I think we explain those steps in our contributors guide as well https://docs.docker.com/opensource/workflow/work-issue/#/pull-and-rebase-frequently

_NOT_EDITED_HERE.txt
```diff
@@ -0,0 +1,18 @@
+# Files and directories here are not edited in the docker.github.io repo.
```
thaJeztah: If this file is only used for automation, I suggest to either make it hidden (`.something`), or we can add a `hack/` directory to put everything in related to automation.

mstanleyjones:
I think it should at least be able to seen, so it c an be referenced when a PR fails and the contributor can look at it. WDYT? I added a bit more info into comments of the file, for humans.

thaJeztah:
We can link to it even if it's hidden, it's just that it's not visible when people to a checkout. No showstopper though; just something we should discuss / look at when we start to use it for automation.

stevvooe:
LGTM

🚲 🏠 on the file name:

```
external.txt
upstream.txt
```

May want to look at python or git's file matching syntax for this use case.

Either way, this perfectly addresses my original concern.

mstanleyjones:
I changed it to .NOT_EDITED_HERE.txt and left it in the root for now, as a compromise. I like the idea that if someone is browsing the repo on Github it will be right at the top. I'm going to go ahead and merge it. We can add exclusions later if we want.

PR #239, opened 2016-10-17
Add ELK stack example to UCP logging

vsaraswat:
### Please take a look

Adds an ELK stack example to the UCP logging page as a real-world example of how you could use Syslog feature. The ELK stack example has been sitting around internally for awhile, so let's provide as a customer-facing example.

### DIFF
--- a/ucp/configuration/configure-logs.md
+++ b/ucp/configuration/configure-logs.md
@@ -28,7 +28,7 @@ One popular logging stack is composed of Elasticsearch, Logstash and
 Kibana. The following example demonstrates how to set up an example
 deployment which can be used for logging.
 
-```sh
+```none
 docker volume create --name orca-elasticsearch-data
 
 docker run -d \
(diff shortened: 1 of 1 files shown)


vsaraswat:
@joaofnfernandes

dhiltgen:
It might also be nice to show some example queries to give users a taste of what they can do with their ELK stack hooked up to UCP.  For example:

* `type:"api" AND (tags:"post" OR tags:"put" OR tags:"delete")` -- Show all the modifications on the system
* `username:"admin"` -- Show all access from a given user
* `type:"auth fail"` -- Show all authentication failures on the system

ucp/configuration/configure-logs.md
```diff
@@ -22,6 +22,42 @@ You can configure UCP for sending logs to a remote logging service:
...
+    -p 5601:5601 \
+    kibana
+```
+
+Once you have these containers running, configure UCP to send logs to 
+the IP of the Logstash container. You can then browse to port 5601 on the system 
+running Kibana and browse log/event entries. You should specify the "time" 
+field for indexing.
+
+Note: When deployed in a production environment, you should secure your ELK
```
johndmulhausen: Suggest you add a leading `>` here so the note is blockquoted. :)

PR #246, opened 2016-10-18
Update DTR API docs overview

banjot:

Added a line to the DTR API doc overview page

### Project version

applies to all versions of the product available today

### Please take a look

@joaofnfernandes 

Added a specific line to call out that the interactive API documentation is a subset of the total API documentation that's provided.

### DIFF
--- a/apidocs/overview.md
+++ b/apidocs/overview.md
@@ -21,4 +21,6 @@ DTR repositories, permissions, and settings.
 To get access to interactive documentation, in your **DTR UI**, click
 on the **top-right menu** and choose **API docs**.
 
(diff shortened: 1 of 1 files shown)


vikstrous:
I think we need to be more specific. These docs are the for DTR and not for enzi. DTR proxies requests to UCP's enzi at /enzi. None of the /enzi APIs are guaranteed to be the referring to the right version of UCP you are using.

joaofnfernandes:
@banjot I'm sure there's customer feedback that led to this proposal. Can you give us more context  on that feedback?

PR #251, opened 2016-10-18
remove redundant back slash

liubin:
I think this is a redundant back slash, should be removed.


sanscontext:
Thanks, and I do believe you're correct - that looks extra.  LGTM!

PR #258, opened 2016-10-18
Fix label used in social media previews 

sanscontext:
Currently when you send a link to the docs in slack, the snippet view is created from this "twitter:title" element, which was confusingly set to "Docs team"

image
  
This changes that to say "Docker docs" instead. We can probably do something better using the pagetitle variable, but this at least fixes it for now.

### Please take a look
 
@mstanleyjones @johndmulhausen 

### DIFF
--- a/_layouts/docs.html
+++ b/_layouts/docs.html
@@ -36,7 +36,7 @@ ng\:form {
 <meta name="twitter:card" content="summary_large_image">
 <meta name="twitter:site" content="@docker_docs">
 <meta name="twitter:url" content="https://twitter.com/docker_docs">
-<meta name="twitter:title" content="Docs team">
+<meta name="twitter:title" content="Docker docs">
 <meta name="twitter:image:src" content="https://www.docker.com/sites/default/files/social/docker-twitter-share.png">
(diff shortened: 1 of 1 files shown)


_layouts/docs.html
```diff
@@ -36,7 +36,7 @@
 <meta name="twitter:card" content="summary_large_image">
 <meta name="twitter:site" content="@docker_docs">
 <meta name="twitter:url" content="https://twitter.com/docker_docs">
-<meta name="twitter:title" content="Docs team">
+<meta name="twitter:title" content="Docker docs">
```
thaJeztah: should this include the actual page title? or is that appended automatically?

mstanleyjones: Yeah I don't think the outcome will be much better unless we somehow include the page title. Maybe @johndmulhausen knows how to do it.

sanscontext: I'm not sure how/if we can invoke that variable from the `layout`, so I'm just going with the static-fix improvement for now. (As I said in the PR description. ;) )

thaJeztah: What happens if we _remove_ this meta-tag? Will it default to the page title?

johndmulhausen:
Switched to use Page's title. Thanks, @sanscontext :)

PR #265, opened 2016-10-19
fix markdown syntax issue

liubin:
Add new lines.

The markdown issue generated a strange html page at 
https://docs.docker.com/engine/reference/commandline/swarm_leave/

screen shot 2016-10-19 at 11 11 06


johndmulhausen:
Thanks, @liubin !

PR #266, opened 2016-10-19
Minor corrections to restore command arguments

anokun7:
### Please take a look

Corrected command arguments for restore. 
Listed correct link for restore documentation. 
Changed usage of /tmp to /var/tmp for storage of backups as /tmp gets cleaned out on reboot.


joaofnfernandes:
Thanks for updating this! LGTM 🚀

Issue #227, opened 2016-10-17
Incorrect command shown for running Dockerized nginx container


razasayed:
On the page https://docs.docker.com/docker-for-mac/ , the command for running a dockerized nginx container is shown as `shell docker run -d -p 80:80 --name webserver nginx` but instead it should be `docker run -d -p 80:80 --name webserver nginx` . Looks like the markdown has already been corrected in the master branch which i confirmed by doing a Preview but on the live site the typo still exists.

sanscontext:
Closing as fixed.

PR #267, opened 2016-10-19
Fix buggy markdown

therobinkim:

Changes `shell docker run -d -p 80:80 --name webserver nginx`
to `docker run -d -p 80:80 --name webserver nginx`
in bullet number 2.

Before:
screen shot 2016-10-19 at 00 17 58

After:
screen shot 2016-10-19 at 00 18 26

FYI modifying the TRIPLE-BACKTICK block to render correctly would misalign the rest of the content under bullet 2, so I opted to just change it to a SINGLE-BACKTICK snippet.

### Related issue

#227 

Signed-off-by: Robin Kim <therobinkim@gmail.com>

### DIFF
--- a/docker-for-mac/index.md
+++ b/docker-for-mac/index.md
@@ -133,9 +133,7 @@ Run these commands to test if your versions of `docker`, `docker-compose`, and `
 
 2. For something more adventurous, start a Dockerized web server.
 
-	```shell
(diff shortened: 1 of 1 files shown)


johndmulhausen:
Thanks, @therobinkim !

PR #271, opened 2016-10-19
Update django.md

DavidRabel-b1:
okay-to-publish

As of Django 1.10 the database backend django.db.backends.postgresql_psycopg2 is called django.db.backends.postgresql .

See https://docs.djangoproject.com/en/1.10/ref/settings/#engine


johndmulhausen:
Thanks, @DavidRabel-b1 ! And thanks for the link to the context/info. :)

PR #275, opened 2016-10-19
docs: add one missing blank and remove double blanks

dvoytik:
add one missing blank and remove double blanks


johndmulhausen:
Thanks, @dvoytik  :)

PR #278, opened 2016-10-19
[Registry Docs] Fix invalid YAML

CWSpear:
I have no idea what `maxthread` is, or if that's where it goes. But as it was, it was invalid YAML and confusing. This PR is more to get someone who _does_ know where `maxthread` goes to take a look and either direct a fix, or, if this _is_ correct, then to merge this.

### DIFF
--- a/registry/configuration.md
+++ b/registry/configuration.md
@@ -78,7 +78,7 @@ information about each option that appears later in this page.
     storage:
       filesystem:
         rootdirectory: /var/lib/registry
-		maxthreads: 100
+        maxthreads: 100
       azure:
         accountname: accountname
         accountkey: base64encodedaccountkey


johndmulhausen:
@docker/maintainers Can anyone speak to whether this change to the example Registry configuration is correct?

CWSpear:
Perhaps it is either there by mistake and needs to be removed, or else it needs some more documentation added. That's the only place `maxthread` appears on the page.

aaronlehmann:
This looks like a spaces-vs-tabs issue.

`maxthreads` was indented with tabs in this document, so it didn't line up with the parameters (intendeted with spaces) as it should.

The fix here looks correct to me. `maxthreads` is supposed to be intented under `filesystem`, at the same level as `rootdirectory`.

dmcgowan:
LGTM

stevvooe:
This needs to go upstream.

Where did this option come from?

johndmulhausen:
@stevvooe Thank you!

@CWSpear I'm sorry for the confusion, but this PR needs to get opened here: 
https://github.com/docker/distribution/tree/master/docs

The thing is, we import this particular doc from upstream. 

We're working on a manifest that auto-redirects "edit this page" to point you to the source repo in cases like this.

aaronlehmann:
Hm, that's confusing to me; why does the same document live in both places?

If documents such as this one are supposed to live in the upstream repository, why not check out those repositories as part of the process of building the docs?

Keeping the repositories in sync seems confusing an error-prone.

Issue #226, opened 2016-10-17
I saw the problem about text format on document page


ninearif:
### Problem description

I think there are some problem about text styling or formatting in documentation page, some indent or text style doesn't show properly.

### Problem location
- [ ] I saw the problem on the following URL: https://docs.docker.com/swarm/swarm_at_scale/deploy-app/
- [ ] This link in document is also not available now : http://vote.myenterprise.com

thaJeztah:
We should not use `myenterprise.com` as a domain name here, as it's an existing domain, and use an `example.com` or `example.org` domain name as those are intended for documentation purposes

ninearif:
I agree with you. It's make me confuse sometime.

PR #279, opened 2016-10-19
Use example.com for documentation

thaJeztah:
The myenterprise.com domain is actually
a registered domain name (and is hosting
a website), so changing to example.com.
The example.com domain is reserved for documentation
purposes (https://tools.ietf.org/html/rfc2606),
so is safe to use.

relates to https://github.com/docker/docker.github.io/issues/226

/cc @ninearif


johndmulhausen:
Thanks, @thaJeztah :)

PR #284, opened 2016-10-19
Updates compose install to 1.8.1

joaofnfernandes:
/ping @shin- to see if it makes sense to update the docs to install the latest compose.


shin-:
:+1: Thanks!

PR #297, opened 2016-10-20
Add to swarm admin docs

mstanleyjones:
- Add details about maintaining quorum
- Add details about pending tasks
### Project version

n/a
### Related issue

n/a

https://github.com/docker/docker/issues/25069#issuecomment-255218986
https://github.com/docker/docker/issues/26017#issuecomment-246405418
### Please take a look

@stevvooe @aaronlehmann 

Signed-off-by: Misty Stanley-Jones misty@docker.com

### DIFF
--- a/engine/swarm/admin_guide.md
+++ b/engine/swarm/admin_guide.md
@@ -40,10 +40,27 @@ fault-tolerant. However, additional manager nodes reduce write performance
 because more nodes must acknowledge proposals to update the swarm state.
 This means more network round-trip traffic.
 
-Raft requires a majority of managers, also called a quorum, to agree on proposed
-updates to the swarm. A quorum of managers must also agree on node additions
-and removals. Membership operations are subject to the same constraints as state
-replication.
+Raft requires a majority of managers, also called the quorum, to agree on
+proposed updates to the swarm, such as node additions or removals. Membership
+operations are subject to the same constraints as state replication.
+
+### Maintaining the quorum of managers
+
+If the swarm loses the quorum of managers, the swarm cannot perform management
+tasks. If your swarm has multiple managers, always have more than two. In order
+to maintain quorum, a majority of managers must be available. An odd number of
+managers is recommended, because the next even number does not make the quorum
+easier to keep. For instance, whether you have 3 or 4 managers, you can still
+only lose 1 manager and maintain the quorum. If you have 5 or 6 managers, you
+can still only lose two.
+
+Even if a swarm loses the quorum of managers, swarm tasks on existing worker
+nodes continue to run. However, swarm nodes cannot be added, updated, or
+removed, and new or existing tasks cannot be started, stopped, moved, or
+updated.
+
+See [Recovering from losing the quorum](recovering-from-losing-the-quorum) for
+troubleshooting steps if you do lose the quorum of managers.
 
 ## Use a static IP for manager node advertise address
 
@@ -64,8 +81,8 @@ Dynamic IP addresses are OK for worker nodes.
 
 You should maintain an odd number of managers in the swarm to support manager
 node failures. Having an odd number of managers ensures that during a network
-partition, there is a higher chance that a quorum remains available to process
-requests if the network is partitioned into two sets. Keeping a quorum is not
+partition, there is a higher chance that the quorum remains available to process
+requests if the network is partitioned into two sets. Keeping the quorum is not
 guaranteed if you encounter more than two network partitions.
 
 | Swarm Size |  Majority  |  Fault Tolerance  |
@@ -103,7 +120,7 @@ In addition to maintaining an odd number of manager nodes, pay attention to
 datacenter topology when placing managers. For optimal fault-tolerance, distribute
 manager nodes across a minimum of 3 availability-zones to support failures of an
 entire set of machines or common maintenance scenarios. If you suffer a failure
-in any of those zones, the swarm should maintain a quorum of manager nodes
+in any of those zones, the swarm should maintain the quorum of manager nodes
 available to process requests and rebalance workloads.
 
 | Swarm manager nodes |  Repartition (on 3 Availability zones) |
@@ -231,29 +248,51 @@ you demote or remove a manager
 ## Recover from disaster
 
 Swarm is resilient to failures and the swarm can recover from any number
-of temporary node failures (machine reboots or crash with restart).
+of temporary node failures (machine reboots or crash with restart) or other
+transient errors. However, a swarm cannot automatically recover if it loses a
+quorum. Tasks on existing worker nodes will continue to run, but administrative
+tasks are not possible, including scaling or updating services and joining or
+removing nodes from the swarm. The best way to recover is to bring the missing
+manager nodes back online. If that is not possible, continue reading for some
+options for recovering your swarm.
 
-In a swarm of `N` managers, there must be a quorum of manager nodes greater than
-50% of the total number of managers (or `(N/2)+1`) in order for the swarm to
-process requests and remain available. This means the swarm can tolerate up to
-`(N-1)/2` permanent failures beyond which requests involving swarm management
-cannot be processed. These types of failures include data corruption or hardware
-failures.
+In a swarm of `N` managers, a quorum (a majority) of manager nodes must always
+be available. For example, in a swarm with 5 managers, a minimum of 3 must be
+operational and in communication with each other. In other words, the swarm can
+tolerate up to `(N-1)/2` permanent failures beyond which requests involving
+swarm management cannot be processed. These types of failures include data
+corruption or hardware failures.
 
-Even if you follow the guidelines here, it is possible that you can lose a
-quorum of manager nodes. If you can't recover the quorum by conventional
-means such as restarting faulty nodes, you can recover the swarm by running
-`docker swarm init --force-new-cluster` on a manager node.
+### Recovering from losing the quorum
+
+If you lose the quorum of managers, you cannot administer the swarm. If you have
+lost the quorum and you attempt to perform any management operation on the swarm,
+an error occurs:
+
+```no-highlight
+Error response from daemon: rpc error: code = 4 desc = context deadline exceeded
+```
+
+The best way to recover from losing the quorum is to bring the failed nodes back
+online. If you can't do that, the only way to recover from this state is to use
+the `--force-new-cluster` action from a manager node. This removes all managers
+except the manager the command was run from. The quorum is achieved because
+there is now only one manager. Promote nodes to be managers until you have the
+desired number of managers.
 
 ```bash
 # From the node to recover
 docker swarm init --force-new-cluster --advertise-addr node01:2377
 ```
 
-The `--force-new-cluster` flag puts the Docker Engine into swarm mode as a
-manager node of a single-node swarm. It discards swarm membership information
-that existed before the loss of the quorum but it retains data necessary to the
-Swarm such as services, tasks and the list of worker nodes.
+When you run the `docker swarm init` command with the `--force-new-cluster`
+flag, the Docker Engine where you run the command becomes the manager node of a
+single-node swarm which is capable of managing and running services. The manager
+has all the previous information about services and tasks, worker nodes are
+still part of the swarm, and services are still running. You will need to add or
+re-add  manager nodes to achieve your previous task distribution and ensure that
+you have enough managers to maintain high availability and prevent losing the
+quorum.
 
 ### Forcing the swarm to rebalance
 
@@ -267,11 +306,16 @@ balance across the swarm. When new tasks start, or when a node with running
 tasks becomes unavailable, those tasks are given to less busy nodes. The goal
 is eventual balance, with minimal disruption to the end user.
 
-If you are concerned about an even balance of load and don't mind disrupting
-running tasks, you can force your swarm to re-balance by temporarily scaling
-the service upward.
+In Docker 1.13 and higher, you can use the `--force` or `-f` flag with the
+`docker service update` command to force the service to redistribute its tasks
+across the available worker nodes. This will cause the service tasks to restart.
+Client applications may be disrupted. If you have configured it, your service
+will use a [rolling update](swarm-tutorial.md#rolling-update).
 
-Use `docker service inspect --pretty <servicename>` to see the configured scale
+If you use an earlier version and you want to achieve an even balance of load
+across workers and don't mind disrupting running tasks, you can force your swarm
+to re-balance by temporarily scaling the service upward. Use
+`docker service inspect --pretty <servicename>` to see the configured scale
 of a service. When you use `docker service scale`, the nodes with the lowest
 number of tasks are targeted to receive the new workloads. There may be multiple
 under-loaded nodes in your swarm. You may need to scale the service up by modest
--- a/engine/swarm/how-swarm-mode-works/services.md
+++ b/engine/swarm/how-swarm-mode-works/services.md
@@ -68,6 +68,36 @@ schedules tasks to worker nodes.
 
 ![services flow](../images/service-lifecycle.png)
 
+### Pending services
+
+A service may be configured in such a way that no node currently in the
+swarm can run its tasks. In this case, the service remains in state `pending`.
+Here are a few examples of when a service might remain in state `pending`.
+
+**Note**: If your only intention is to prevent a service from
+being deployed, scale the service to 0 instead of trying to configure it in
+such a way that it will remain in `pending`.
+
+- If all nodes are paused or drained, and you create a service, it will be
+  pending until a node becomes available. In reality, the first node to become
+  available will get all of the tasks, so this is not a good thing to do in a
+  production environment.
+
+- You can reserve a specific amount of memory for a service. If no node in the
+  swarm has the required amount of memory, the service will remain in a pending
+  state until a node is available which can run its tasks. If you specify a very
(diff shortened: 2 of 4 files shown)


engine/swarm/admin_guide.md
```diff
@@ -53,6 +53,29 @@ updates to the swarm. A quorum of managers must also agree on node additions
 and removals. Membership operations are subject to the same constraints as state
 replication.
 
+### Maintaining a quorum of managers
+
+Always maintain **an odd number of manager nodes** in the swarm. Otherwise, your
+swarm may lose a quorum of managers, and will not be able to perform management
```
aaronlehmann: Having an even number of managers doesn't cause you to lose quorum, it just doesn't have any fault tolerance benefit over having one fewer manager. This is described in detail in the "Add manager nodes for fault tolerance" section below.

engine/swarm/admin_guide.md
```diff
@@ -53,6 +53,29 @@ updates to the swarm. A quorum of managers must also agree on node additions
...
+tasks. Three managers is the minimum, and five managers is recommended. In a
```
thaJeztah: Should we combine this with https://github.com/mstanleyjones/docker.github.io/blob/0db3e0bf877880a935d53b14459ee9bd09786656/engine/swarm/admin_guide.md#add-manager-nodes-for-fault-tolerance, as it's largely repeating the same message?

Note that having a single manager is "ok" (if no fault tolerance is needed at all)

aaronlehmann: One manager is the minimum. Three managers is the minimum for a setup to continue working after a manager failure. I don't know if we should go so far as to recommend 5 managers, as 3 is sufficient for a lot of use cases. But it is good to point out that 5 managers allows you to lose 2.

engine/swarm/admin_guide.md
```diff
@@ -53,6 +53,29 @@ updates to the swarm. A quorum of managers must also agree on node additions
...
+
+```no-highlight
+Error response from daemon: rpc error: code = 4 desc = context deadline exceeded
+```
+
+The only way to recover from this state is to use the `--force-new-swarm` action
+from a manager node. This removes all managers except the manager the command
+was run from. Quorum is achieved because there is now only one manager. Promote
+nodes to be managers until you have the desired number of managers. You should
+always have an odd number of managers.
```
thaJeztah: Should we point out that `--force-new-swarm` looses all swarm-related data (networks, services, etc)?

engine/swarm/how-swarm-mode-works/services.md
```diff
@@ -76,6 +76,27 @@ schedules tasks to worker nodes.
 
 ![services flow](../images/service-lifecycle.png)
 
+### Pending services
+
+You can even configure a service in such a way that no node currently in the
+swarm can run its tasks. Here are a few examples:
+
+- You can even create a service with an image that is not available in the
```
thaJeztah: You can even -> You can?

engine/swarm/admin_guide.md
aaronlehmann: I don't think we need to repeat the recommendation for an odd number of managers here. It's covered well under "Add manager nodes for fault tolerance".

engine/swarm/how-swarm-mode-works/services.md
```diff
@@ -76,6 +76,27 @@ schedules tasks to worker nodes.
...
+  registry, either because it hasn't been pushed to the registry yet or because
+  the registry is not currently available.
```
thaJeztah: Using a non-existing image is actually dangerous, as (contrary to some other constraints), swarm does not know about this when scheduling, so it will keep trying to deploy the task, and pull images. It's possible the node runs out of resources because of this, or (if no log-rotation is in place) runs out of disk space.

engine/swarm/manage-nodes.md
```diff
@@ -151,9 +151,11 @@ You can promote a worker node to the manager role. This is useful when a
...
 maintenance. Similarly, you can demote a manager node to the worker role.
 
-Regardless of your reason to promote or demote a node, you should always
-maintain an odd number of manager nodes in the swarm. For more information refer
-to the [Swarm administration guide](admin_guide.md).
+>**Note: Maintaining a quorum** Regardless of your reason to promote or demote
+a node, you should always maintain **an odd number of manager nodes** in the
+swarm. Otherwise, your swarm may lose a quorum of managers, and will not be able
+to perform management tasks. For more information refer to the
+[Swarm administration guide](admin_guide.md).
```
aaronlehmann: As above, this warning feels a little dire compared to the actual consequences.

engine/swarm/manage-nodes.md
```diff
@@ -199,6 +201,12 @@ quorum. To override the warning, pass the `--force` flag. If the last manager
 node leaves the swarm, the swarm becomes unavailable requiring you to take
 disaster recovery measures.
 
+>**Note: Maintaining a quorum** Regardless of your reason to promote or demote
+a node, you should always maintain **an odd number of manager nodes** in the
+swarm. Otherwise, your swarm may lose a quorum of managers, and will not be able
+to perform management tasks. For more information refer to the
+[Swarm administration guide](admin_guide.md).
```
aaronlehmann: Same comment about the severity of the warning.

engine/swarm/services.md
```diff
@@ -128,15 +135,22 @@ deploys a service to the node. You can apply constraints to the
 service based upon node attributes and metadata or engine metadata. For more
 information on constraints, refer to the `docker service create` [CLI  reference](../reference/commandline/service_create.md).
 
+### Reserving memory or number of CPUs for a service
+
+To reserve a given amount of memory or number of CPUs for a service, use the
+`--reserve-memory` or `--reserve-cpu` flags. If no available nodes can satisfy
+the requirement (for instance, if you request 4 CPUs and no node in the swarm
+has 4 CPUs), the service remains in a pending state until a node is available to
+run its tasks.
```
aaronlehmann: Thanks for adding this. I think someone had pointed this out to me as something that needed more documentation, recently.

engine/swarm/admin_guide.md
aaronlehmann: `--force-new-cluster` does not lose this data.

thaJeztah: oh, I was misinformed then, I thought that was mentioned at some point. Thanks!

engine/swarm/how-swarm-mode-works/services.md
```diff
@@ -76,6 +76,27 @@ schedules tasks to worker nodes.
...
+
+- You can reserve a specific number of CPUs for a service. If no node in the
+  swarm has the required number of CPUs, the service will remain in a pending
+  state until a node is available which can run its tasks.
+
+- You can impose placement constraints on the service, and the constraints may
+  not be able to be honored at a given time.
+
+This behavior is a consequence of the declarative way that Docker swarms are
```
stevvooe: Don't be a negatron. This wording makes it sound like a bad thing.

thaJeztah: "a negatron" :joy:

mstanleyjones:
Addressed comments

engine/swarm/admin_guide.md
```diff
@@ -53,6 +53,19 @@ updates to the swarm. A quorum of managers must also agree on node additions
 and removals. Membership operations are subject to the same constraints as state
 replication.
 
+### Maintaining a quorum of managers
+
+If the swarm loses a quorum, it will not be able to perform management
```
aaronlehmann: Not sure what's most correct here, but I've generally seen "loses quorum" instead of "loses a quorum". "Losing the quorum of managers" might also work.

engine/swarm/admin_guide.md
```diff
@@ -53,6 +53,19 @@ updates to the swarm. A quorum of managers must also agree on node additions
...
+tasks. If your swarm has multiple masters, always have more than two. In order
```
aaronlehmann: "managers", not masters

engine/swarm/admin_guide.md
```diff
@@ -53,6 +53,19 @@ updates to the swarm. A quorum of managers must also agree on node additions
...
+to maintain a quorum, a majority of managers must be available. An odd number
+of managers is recommended, because the next even number does not make the quorum
+easier to keep. For instance, whether you have 3 or 4 managers, you can still only
+lose 1 manager and maintain a quorum. If you have 5 or 6 managers, you an still
```
aaronlehmann: Typo: `an` should be `can`

engine/swarm/admin_guide.md
```diff
@@ -53,6 +53,19 @@ updates to the swarm. A quorum of managers must also agree on node additions
...
+only lose two.
+
+See [Recovering from losing a quorum](recovering-from-losing-a-quorum) for
+troubleshooting steps if you do lose a quorum of managers.
```
aaronlehmann: Same comment about "lose a quroum" / "losing a quorum", but again I'm not sure what's most correct.

engine/swarm/admin_guide.md
```diff
@@ -239,27 +252,40 @@ you demote or remove a manager
...
-50% of the total number of managers (or `(N/2)+1`) in order for the swarm to
-process requests and remain available. This means the swarm can tolerate up to
-`(N-1)/2` permanent failures beyond which requests involving swarm management
-cannot be processed. These types of failures include data corruption or hardware
-failures.
+### Recovering from losing a quorum
+
+If you lose a quorum of managers, you cannot administer the swarm. Attempting
+to remove a node from a swarm which does not have a quorum of managers results
+in an error like the following:
```
aaronlehmann: This is technically true but I'm not sure why we would explain what happens when you try to remove a node from a swarm that already has too few nodes to function. Is it just an example of how management features stop working?

mstanleyjones:
All the feedback has been addressed (I believe). @aaronlehmann PTAL again

engine/swarm/admin_guide.md
```diff
@@ -48,10 +48,22 @@ fault-tolerant. However, additional manager nodes reduce write performance
...
+
+### Maintaining the quorum of managers
+
+If the swarm loses the quorum of managers, the swarm cannot perform management
+tasks. If your swarm has multiple managers, always have more than two. In order
+to maintain quorum, a majority of managers must be available. An odd number
+of managers is recommended, because the next even number does not make the quorum
+easier to keep. For instance, whether you have 3 or 4 managers, you can still only
+lose 1 manager and maintain the quorum. If you have 5 or 6 managers, you can
+still only lose two.gn
```
aaronlehmann: Something wrong at the end of this paragraph.

johndmulhausen:
Deploy preview ready!

Built with commit 02ffe1f166dbbe70fe6b7f7dd7ac1e87fb494aff

https://deploy-preview-297--docker-docs.netlify.com

mstanleyjones:
@aaronlehmann I addressed your comments and also added info about forcing services to update. That would mean we need to hold back this PR until the 1.13 doc is merged in, so let me know if you want me to revert that rewording / addition and add it in a separate PR against `vnext-engine`.

engine/swarm/admin_guide.md
```diff
@@ -231,29 +243,46 @@ you demote or remove a manager
...
 # From the node to recover
 docker swarm init --force-new-cluster --advertise-addr node01:2377
 ```
 
-The `--force-new-cluster` flag puts the Docker Engine into swarm mode as a
-manager node of a single-node swarm. It discards swarm membership information
-that existed before the loss of the quorum but it retains data necessary to the
-Swarm such as services, tasks and the list of worker nodes.
+When you run the `docker swarm init` command with the `--force-new-cluster` flag
+, the Docker Engine where you run the command becomes the manager node of a
```
aaronlehmann: comma on its own line

engine/swarm/admin_guide.md
```diff
@@ -267,11 +296,16 @@ balance across the swarm. When new tasks start, or when a node with running
...
+In Docker 1.13 and higher, you can use the `--force` or `-f` flag with the
+`docker service update` command to force the service to redistribute its tasks
+across the available worker nodes. This will cause the service tasks to restart.
+Client applications may be disrupted. If you have specified an `--update-delay`
+value for your service, the service will use a rolling restart.
 
-Use `docker service inspect --pretty <servicename>` to see the configured scale
+If you use an earlier version and you want to achieve an even balance of load
+across workers and don't mind disrupting running tasks, you can force your swarm
+to re-balance by temporarily scaling the service upward. Use
```
aaronlehmann: Hm, seems surprising to include workarounds for older versions in the documentation. Do we generally do this?

mstanleyjones: I think we have to, since users of UCP and DTR won't be on 1.13 yet...

aaronlehmann: Don't we version this document? I feel like it would be very confusing to try to cover multiple product versions in the same document.

mstanleyjones:
Feedback addressed. PTAL again.

engine/swarm/admin_guide.md
mstanleyjones: This is the admin guide that lives on docker.com. It is versioned in a way (you can get to older versions), but we also need to be careful because lots of people just Google. When a feature is brand new, we need to call out when you can expect it to work. Otherwise, support will get lots of calls from people on earlier versions trying to use commands.

engine/swarm/admin_guide.md
aaronlehmann: Please fix the comma

engine/swarm/admin_guide.md
```diff
@@ -231,29 +247,46 @@ you demote or remove a manager
...
 
-The `--force-new-cluster` flag puts the Docker Engine into swarm mode as a
-manager node of a single-node swarm. It discards swarm membership information
-that existed before the loss of the quorum but it retains data necessary to the
-Swarm such as services, tasks and the list of worker nodes.
+When you run the `docker swarm init` command with the `--force-new-cluster` flag
+, the Docker Engine where you run the command becomes the manager node of a
+single-node swarm which is capable of managing and running services. The manager
+has all the previous information about services and tasks, but you need to add
+manager and worker nodes to achieve your previous task distribution and ensure
```
aaronlehmann: Workers would still be part of the swarm, but you would need to re-add managers.

engine/swarm/admin_guide.md
```diff
@@ -267,11 +296,16 @@ balance across the swarm. When new tasks start, or when a node with running
...
 is eventual balance, with minimal disruption to the end user.
 
-If you are concerned about an even balance of load and don't mind disrupting
-running tasks, you can force your swarm to re-balance by temporarily scaling
-the service upward.
+In Docker 1.13 and higher, you can use the `--force` or `-f` flag with the
+`docker service update` command to force the service to redistribute its tasks
+across the available worker nodes. This will cause the service tasks to restart.
+Client applications may be disrupted. If you have specified an `--update-delay`
+value for your service, the service will use a rolling restart.
```
aaronlehmann: See comment above. `--update-delay` does not control whether a rolling restart is used, only whether there is a delay between tasks being updated in a rolling update.

mstanleyjones:
Addressed latest feedback, including some comments I had missed before.

engine/swarm/how-swarm-mode-works/services.md
```diff
@@ -68,6 +68,31 @@ schedules tasks to worker nodes.
...
+  production environment.
+
+- You can reserve a specific amount of memory for a service. If no node in the
+  swarm has the required amount of memory, the service will remain in a pending
+  state until a node is available which can run its tasks. If you specify a very
+  large value, such as 500 GB, the task will be pending forever, unless you
+  really have a node which can satisfy it.
+
+- You can impose placement constraints on the service, and the constraints may
+  not be able to be honored at a given time.
```
thaJeztah: Reading this again; this section may be read as "instructions for creating a pending service". In some cases this is desirable (e.g. automatically start deploying when a node becomes available), however if this is just to prevent a service from being deployed, I think we should mention that scaling the service to `0` is the most logical way (or refer to the section about scaling).

engine/swarm/manage-nodes.md
```diff
@@ -6,10 +6,10 @@ title: Manage nodes in a swarm
 
 As part of the swarm management lifecycle, you may need to view or update a node as follows:
 
-* [list nodes in the swarm](manage-nodes.md#list-nodes)
```
thaJeztah: did we fix this in Jekyll? Otherwise the `manage-nodes.md` is still needed :cry:

engine/swarm/admin_guide.md
```diff
@@ -231,29 +247,47 @@ you demote or remove a manager
...
+operational and in communication with each other. In other words, the swarm can
+tolerate up to `(N-1)/2` permanent failures beyond which requests involving
+swarm management cannot be processed. These types of failures include data
+corruption or hardware failures.
+
+### Recovering from losing the quorum
+
+If you lose the quorum of managers, you cannot administer the swarm. If you have
+lost the quorum and you attempt to perform any management operation on the swarm,
+an error occurs:
```
thaJeztah: @stevvooe you think we should repeat that tasks remain to run (https://github.com/docker/docker.github.io/pull/297/files#r88358814)? Train of thought; If the lost quorum is temporary, but all tasks are still running, there may not be a direct need to try to recover.

mstanleyjones:
Addressed the new feedback.

PR #302, opened 2016-10-21
Update archlinux.md so it renders section title correctly

krerkkiat:

As reading through the installation guide, I found that the Uninstallation is render as "## Uninstallation". This seems to be the result of lack of newline before the ## Uninstallation in the md file. Thus, I just basically add a newline there.

See the [website](https://docs.docker.com/engine/installation/linux/archlinux/) for the issue.

### Project version

Not really sure. It may apply to all version that support uninstallation.

### Related issue

None.

None.

### Please take a look

None.

Add newline before the Uninstallation section since it is displayed as the raw text (## Uninstallation) on the website.

### DIFF
--- a/engine/installation/linux/archlinux.md
+++ b/engine/installation/linux/archlinux.md
@@ -87,6 +87,7 @@ IPForward=kernel
 ```
 
 This configuration allows IP forwarding from the container as expected.
+
 ## Uninstallation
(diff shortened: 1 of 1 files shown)


johndmulhausen:
Thanks, @krerkkiat !

PR #283, opened 2016-10-19
Add apt source with sudo tee

konstruktoid:
### Describe the proposed changes

@JonJagger is correct when pointing out that `sudo echo "<REPO>" > /etc/apt/sources.list.d/docker.list` will fail, and that the docs should mention `sudo tee` instead.

### Related issue

Closes #272 

Signed-off-by: Thomas Sjögren <konstruktoid@users.noreply.github.com>

PR #303, opened 2016-10-21
Fix 'Permission denied' when creating docker.list file

bmarcot:

Change to the documentation, in https://docs.docker.com/engine/installation/linux/ubuntulinux/

### Project version

Docker docs.

### Please take a look

Command was returning 'Permission Denied': "This is a known issue, when you use sudo in this fashion, it won't work right. That is because while the echo command is run as sudo, the >> for append tries to open the file target as a non-sudo user. That is where the permission issue is." See http://askubuntu.com/questions/185268/permission-denied-etc-apt-sources-list

### DIFF
--- a/engine/installation/linux/ubuntulinux.md
+++ b/engine/installation/linux/ubuntulinux.md
@@ -90,7 +90,7 @@ packages from the Docker repository:
     for the placeholder `<REPO>`.
 
     ```bash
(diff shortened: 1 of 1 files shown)


konstruktoid:
I believe this is a duplicate of #283

bmarcot:
It is, my mistake.

PR #304, opened 2016-10-21
Remove line numbers from caps link

### LINKED ISSUE #294: Feedback for: engine/security/security.md
URL: [https://docs.docker.com/engine/security/security.md](https://docs.docker.com/engine/security/security/)

In the sentence "This won’t affect regular web apps; but malicious users will find that the arsenal at their disposal has shrunk considerably! By default Docker drops all capabilities except those needed, a whitelist instead of a blacklist approach. ", there is a link from the text "those needed" that links to https://github.com/docker/docker/blob/master/oci/defaults_linux.go#L64-L79.  As the file changes, the actual selected range changes.  Not sure if we want to link to a given release or just update this to the right range on `master`.

konstruktoid:

Remove static line numbers from defaults_linux.go, since they will probably change in the future, and the source file IMHO is clear enough that `s.Process.Capabilities` is easy to find.

### Related issue

Closes #294 

### DIFF
--- a/engine/security/security.md
+++ b/engine/security/security.md
@@ -197,7 +197,7 @@ to the host.
 This won't affect regular web apps; but malicious users will find that
 the arsenal at their disposal has shrunk considerably! By default Docker
 drops all capabilities except [those
-needed](https://github.com/docker/docker/blob/master/oci/defaults_linux.go),
+needed](https://github.com/docker/docker/blob/master/oci/defaults_linux.go#L62-L77),
 a whitelist instead of a blacklist approach. You can see a full list of
 available capabilities in [Linux
 manpages](http://man7.org/linux/man-pages/man7/capabilities.7.html).
--- a/engine/security/security.md
+++ b/engine/security/security.md
@@ -197,7 +197,7 @@ to the host.
 This won't affect regular web apps; but malicious users will find that
 the arsenal at their disposal has shrunk considerably! By default Docker
 drops all capabilities except [those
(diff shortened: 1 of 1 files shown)


johndmulhausen:
@konstruktoid Only thing is, now that link doesn't have much meaning. It's a little disorienting to just get dumped onto the top of a code file. Is the issue that the wrong lines are being highlighted? Perhaps we should just adjust it so that it highlights 62-77

konstruktoid:
@johndmulhausen True, but adding highlighting requires changes to the docs every time the code file changes.

johndmulhausen:
I think that's what we have to do to give users the right information. Another option is showing the enum values inline; if those are pretty much set it may be the lowest-maintenance option

konstruktoid:
Added the correct line numbers from https://github.com/docker/docker/blob/master/oci/defaults_linux.go, we'll just have to keep track on the file.

johndmulhausen:
Thanks, @konstruktoid :)

thaJeztah:
Perhaps link to a specific commit instead? https://github.com/docker/docker/blob/8b1772c86b20d06ae6bd8ccf4b20f6df4cd86b62/oci/defaults_linux.go#L62-L77

johndmulhausen:
^ Sounds like a good idea to me :)

konstruktoid:
Should I send a new PR with @thaJeztah suggestion @johndmulhausen?

PR #309, opened 2016-10-21
Moving sudo command after echo pipe for Ubuntu install

dgvigil:
The current command fails and requires the sudo to move after the pipe. This follows the same usage as other Linux installation instructions.


johndmulhausen:
Thanks, @dgvigil !

PR #318, opened 2016-10-22
Add aliases to engine/reference/ subdirs for backwards compatible links

weslord:
Added aliases to old locations of docs in /engine/reference/commandline/ and api/ so that old incoming links (from, say, StackOverflow) will redirect properly.

Parent directory already had similar aliases, propagated them into subdirectory contents.

### Related issue
Issue #79


johndmulhausen:
Thanks, @weslord ! Sheesh, this sure looks like tedious work; we thank you for it!

PR #336, opened 2016-10-24
Update installing-cli.md to mention python-dev for pip install

wluxion:
Ran into the same problem described here:
http://stackoverflow.com/questions/36067018/install-docker-cloud-client-using-pip

It wasn't immediately obvious how to proceed -- mentioning python-dev should make it much easier.


johndmulhausen:
Hi @wluxion; I think it might be better to go ahead an include the command `apt-get install python-dev` -- can you add that?

wluxion:
Happily. Done.

johndmulhausen:
Thanks, @wluxion ! :)

PR #338, opened 2016-10-25
docker-for-mac/osxfs: clarify the ownership behavior in Beta 29

dsheets:
### Project version

Docker for Mac Beta 29

docker/pinata#5671

### DIFF
--- a/docker-for-mac/osxfs.md
+++ b/docker-for-mac/osxfs.md
@@ -86,7 +86,7 @@ values. Ownership-based permissions are only enforced at the OS X file
 system level with all accessing processes behaving as the user running
 Docker. If the user does not have permission to read extended attributes
 on an object (such as when that object's permissions are `0000`), `osxfs`
-will attempt to add an access control list (ACL) entry, which allows the user to
+will attempt to add an access control list (ACL) entry that allows the user to
 read and write extended attributes. If this attempt fails, the object will appear to be owned by the process accessing it until the extended attribute is readable again.
 
 ### File system events
--- a/docker-for-mac/osxfs.md
+++ b/docker-for-mac/osxfs.md
@@ -85,11 +85,9 @@ requests for ownership metadata will return the previously set
 values. Ownership-based permissions are only enforced at the OS X file
 system level with all accessing processes behaving as the user running
 Docker. If the user does not have permission to read extended attributes
-on an object, e.g. when that object's permissions are `0000`, `osxfs`
-will attempt to add an access control list entry allowing the user to
-read and write extended attributes. If this is not possible or extended
-attribute permissions are still denied, ownership will be reported as
-the accessing process until the extended attribute is again readable.
+on an object (such as when that object's permissions are `0000`), `osxfs`
+will attempt to add an access control list (ACL) entry, which allows the user to
+read and write extended attributes. If this attempt fails, the object will appear to be owned by the process accessing it until the extended attribute is readable again.
 
 ### File system events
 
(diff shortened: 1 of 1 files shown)


docker-for-mac/osxfs.md
```diff
@@ -85,9 +85,11 @@ requests for ownership metadata will return the previously set
 values. Ownership-based permissions are only enforced at the OS X file
```
mstanleyjones: s/OS X/macOS
s/file system/filesystem

docker-for-mac/osxfs.md
```diff
@@ -85,9 +85,11 @@ requests for ownership metadata will return the previously set
...
 system level with all accessing processes behaving as the user running
 Docker. If the user does not have permission to read extended attributes
-on an object, e.g. when that object's permissions are `0000`, ownership
-will be reported as the accessing process until the extended attribute
-is again readable.
+on an object, e.g. when that object's permissions are `0000`, `osxfs`
```
mstanleyjones: Let's try to reduce the complexity of this sentence. I'm thinking something along these lines.

"If the user doesn't have permission to read extended attributes on an object (such as when that object's permissions are `0000`), `osxfs` attempts to add an access control list (ACL) which allows the user to read and write extended attributes. If this attempt fails, the object will appear to be owned by the process accessing it until the extended attribute is readable."

Is it very likely that `osxfs` will be able to add it? I suppose it runs as root?

samoht: no it doesn't run as root.

samoht: osxfs will have the same privileges that the user running `Docker.app` which basically means the containers will be able to read/write the same sets of files as the user can.

docker-for-mac/osxfs.md
dsheets: The rest of the file uses "OS X" and "file system". Do you think they should be changed globally? I have no idea what to do about Apple's rename of their OS and the component is still called `osxfs` but maybe the docs should be universally updated... As for "file system", I interpret it as a noun phrase with "file" modifying "system" but I have seen lots of compound word spellings -- what is the convention in the rest of the Docker/Docker for Mac docs?

londoncalling:
@dsheets I incorporated @mstanleyjones 's comments, please merge when reviewed, Misty

docker-for-mac/osxfs.md
dsheets: I think this is a good simplification, thanks! I endorse Victoria's commit 415dfd7 as accurate.

As for your questions, it is not currently mentioned in the docs but this also applies to attempts to write extended attributes to persist the ownership metadata from the Linux VM. It is very likely that the reason for `chown` failure is that the UNIX permissions on the object prohibit read, write, or both which correspond to reading xattrs, writing xattrs, or both. We take advantage of a couple nice properties of the OS X FS permission model:

1. the owner of a file may change the security settings of that file regardless of the permissions
2. ACLs are included in the security settings
3. ACLs override UNIX permissions

The use case that precipitated this change was the demand for POSIX-compliance regarding the `chown`ing of read-only or mode 0 files (that are still owned by the user). By default, OS X will disallow this but now we can set a bottom priority ACL entry (or add to the first user-allow ACL entry) the ability to read and write extended attributes which then lets us treat extended attributes just like traditional UNIX ownership metadata.

The alternative, without using ACLs, is a horrible non-atomic racy dance of permission juggling as we attempt to evade the coupling of UNIX permissions and xattrs permissions.

I hope this helps elucidate the use case and issues. I'm happy to help improve these docs even more if you can think of ways they could be made clearer.

Thanks!

londoncalling:
Got it @dsheets , made that correction and now will merge this, @mstanleyjones merging now since I think I incorporated all your comments as well.

PR #339, opened 2016-10-25
Update swarm.md

dave-pollock:
I have changed the example for compose affinity filters to use the environment option rather than labels.
Scheduling filters should be configured in the environment section of compose files - not labels.


mstanleyjones:
@shin- PTAL

PR #256, opened 2016-10-18
Clarify Hyper-V and VirtualBox info for Docker for Windows

mstanleyjones:
### Describe the proposed changes

Reworded a couple spots to clarify the situation with Hyper-V and Virtualbox on Docker for Windows

### Project version

n/a

### Related issue

#134

### Related issue or PR in another project

n/a

### Please take a look

@londoncalling @engcheng

Issue #134, opened 2016-10-10
Feedback for: machine/get-started.md


engcheng:
URL: [https://docs.docker.com/machine/get-started.md](https://docs.docker.com/machine/get-started/)

Under the section, "If you are using Docker for Windows"

The first line mentions: 
> Docker for Windows uses Microsoft Hyper-V for virtualization, and Hyper-V and Oracle VirtualBox are incompatible

Is there a typo error? Because Hyper-V for virtualization and Hyper-V seems to be pointing to the same thing.

.

johndmulhausen:
@engcheng I can't repro there being more than one link. Which browser are you using?

engcheng:
@johndmulhausen thanks for feedback. I'm using Firefox 45.4.0 on Linux 64 bits.

The [url](https://docs.docker.com/machine/get-started/) with typo which I'm referring to can be traced in the following sequence:

1. Get started with Docker Machine and a local VM
2. Prerequisite Information
3. If you are using Docker for Windows
4. Line 1

I've also attached a screenshot for further clarity in [PicPaste](http://picpaste.com/docker_machine_typo-Ylj0iPnq.png)

Hope its clearer now.

johndmulhausen:
Thanks, @engcheng that does help, especially the screenshot. :)

So by "point to the same thing" you don't mean that there are two Hyper-V links pointing to the same thing, you mean that "Hyper-V" is being used twice in this sentence and it is not clear why, correct?

engcheng:
yes @johndmulhausen 

If we break the statement into 2 parts, 
First, it mentions: "Docker for Windows uses Microsoft Hyper-V for virtualization"
Then, followed by "and Hyper-V and Oracle VirtualBox are incompatible"

Substituting the product name by symbols, "Docker for Windows uses A for virtualization, and A and B are incompatible". Thus it creates an impression A is both compatible and incompatible.

Think I get it. The first statement is stating what works. The second statement is saying what don't work, right?

If that's the case, perhaps may I suggest:

- it may be clearer to state that Hyper-V needs to be the only hypervisor.
- reword using however. "Docker for Windows uses A for virtualization. **However**, A and B are incompatible. Therefore, you cannot run the two solutions simultaneously.".

Do you think its clearer? Thanks for taking time to enlighten :smile:

mstanleyjones:
[...]

PR #346, opened 2016-10-25
incorporated Misty's rewrite from pr#256 with additional info

londoncalling:
Updates to Docker Machine per PR #256 and issue #134 , @mstanleyjones , @engcheng

Needed to do a slight re-write on that other PR to clarify relationship of Docker Machine, drivers, and Docker for Mac and windows.


londoncalling:
ok LGTM, merging

@friism @thaJeztah please take a look, I'll update anything else needed.

PR #349, opened 2016-10-26
Updated uninstall.md

ntwrkguru:
### Please take a look

Corrected the uninstall command to make it copypasteable (it was missing a \ at the end of the volume argument) and also updated the expected output to include the WARN messages that will appear. Also corrected ordered list numbering (cosmetic only while editing)


johndmulhausen:
Nice, thanks a lot @ntwrkguru ! :)

PR #350, opened 2016-10-26
Use the correct service name when linking

KevinTangy:


johndmulhausen:
Thanks, @KevinTangy !

PR #363, opened 2016-10-27
compose/startup-order.md -- added clarification

michael-wise:

The page uses "wait-for-it.sh" and then shows a script. I added a filename above the script to indicate it is not the same as "wait-for-it.sh".

### Please take a look

Added a filename above the custom script to distinguish it from the wait-for-it.sh script.

### DIFF
--- a/compose/startup-order.md
+++ b/compose/startup-order.md
@@ -53,9 +53,11 @@ script:
           db:
             image: postgres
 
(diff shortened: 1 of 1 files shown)


johndmulhausen:
Thanks, @michael-wise ! And congratulations on your first pull request. :) Hope you come back to us with more!

PR #367, opened 2016-10-27
Alphabetize daemon config file options

cavemandaveman:
Just what the title says, plus change to 4 space indentation instead of tabs.


johndmulhausen:
Thanks, @cavemandaveman :)

PR #375, opened 2016-10-28
Docker Cloud: clarify use case for different types of env vars

sanscontext:
Clarify why you'd use env vars in the Cloud build settings vs when you'd declare them directly in hooks.

cc @pchico83 and @pkennedyr for review

Signed-off-by: LRubin lrubin@docker.com

### DIFF
--- a/docker-cloud/builds/automated-build.md
+++ b/docker-cloud/builds/automated-build.md
@@ -105,6 +105,24 @@ run an initial test.
     Docker Cloud on every push. Only pushes to branches that are listed as the
     source for one or more tags will trigger a build.
 
+### Environment variables for builds
+
+You can set the values for environment variables used in your build processes
+when you configure an automated build. Add your build environment variables by
+clicking the plus sign next to the **Build environment variables** section, and
+then entering a variable name and the value.
+
+When you set variable values from the Docker Cloud UI, they can be used by the
+commands you set in `hooks` files, but they are stored so that only users who
+have `admin` access to the Docker Cloud repository can see their values. This
+means you can use them to safely store access tokens or other information that
+should remain secret.
+
+> **Note**: The variables set on the build configuration screen are used during
+the build processes _only_ and should not be confused with the environment
+values used by your service (for example to create service links).
+
+
 ## Check your active builds
 
 1. To view active builds, go to the repository view and click **Timeline**.
@@ -188,7 +206,7 @@ To work around this, you can set up your automated build using the `SSH_PRIVATE`
     This step is optional, but allows you to revoke the build-only keypair without removing other access. <!-- (TODO: Link to instructions for GH & BB ) -->
 2. Copy the private half of the keypair to your clipboard.
 3. In Docker Cloud, navigate to the build page for the repository that has linked private submodules. (If necessary, follow the steps [here](automated-build.md#configure-automated-build-settings) to configure the automated build.)
-4. At the bottom of the screen, click the plus sign ( **+** ) next to **Environment Variables**.
+4. At the bottom of the screen, click the plus sign ( **+** ) next to **Build Environment variables**.
 5. Enter `SSH_PRIVATE` as the name for the new environment variable.
 6. Paste the private half of the keypair into the **Value** field.
 7. Click **Save**, or **Save and Build** to validate that the build now completes.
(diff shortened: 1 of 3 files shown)


docker-cloud/builds/automated-build.md
```diff
@@ -105,6 +105,17 @@ run an initial test.
     Docker Cloud on every push. Only pushes to branches that are listed as the
     source for one or more tags will trigger a build.
 
+### Environment variables for builds
+
+When you configure automated builds in Docker Cloud you can specify additional Environment Variables. The variables you add on the **Configure build settings** screen are used during the build process _only_.
+
+You can also declare environment variables in `hooks` files, however anything
```
pchico83: I would remove the first sentence. Instead I would say that the nice thing about build env vars is that they are accessible from your hooks without the need of storing them in your git repo.

docker-cloud/builds/automated-build.md
```diff
@@ -105,6 +105,17 @@ run an initial test.
...
+declared in these files can be read by anyone with read access to the code
+repository. If you use build environment variables to store access tokens or
+other information that should remain secret, create them on the build settings
+screen. The variables created here are only accessible to users who have `admin`
+access to the Docker Cloud repository.
```
pchico83: If it is not in other doc, I would say that env variables are only accessible from hooks, and I would add the example about how to pass these variables as build args of the dockerfile.

docker-cloud/builds/automated-build.md
sanscontext: I understand your concern, but I want to make sure I'm introducing the concept in prose in this section before I dive in to how it works. I can definitely improve this to talk about build env vars being in Hooks too.

docker-cloud/builds/automated-build.md
sanscontext: (I'll rework to add this in.)

docker-cloud/builds/automated-build.md
sanscontext: Updated - better?

sanscontext:
@pchico83 PTAL when you get a chance.

docker-cloud/builds/automated-build.md
```diff
@@ -105,6 +105,21 @@ run an initial test.
     Docker Cloud on every push. Only pushes to branches that are listed as the
     source for one or more tags will trigger a build.
 
+### Environment variables for builds
+
+You can set environment variables used in your build either by declaring them in
```
pchico83: "by declaring them in hooks files" --> we cannot say this. the message should be that the build env vars defined in the UI are accessible from your hooks. other approaches are just wrong and cannot be commented or proposed as alternatives.
If you want to dig in this idea, make a use case for env vars, tell that you need env vars in your hooks to cover this use case, say that declaring them in your git repo is unacceptable due to security concerns, and then conclude that defining them in the Docker Cloud UI is the right solution.

caervs: There are rare cases where putting confidential info in your source repo is fine. For instance, if I am one guy who has a private repo for an ssh container, I can store the private key for the container image in my source repo since no one else can read it.

We should make the benefits of each approach clear so that users can decide if build rule env vars make sense for their use cases. The biggest benefit of secrets in source repo is that the secrets are version controlled which makes rotating secrets way easier. Of course the drawback is that if you never amend your history git will preserve your secrets forever for anyone who has read access.

So, the general rule of thumb that all secrets should go into build rules is I think too simple. My opinion is that this space is a good place to make clear the subtleties of secrets but our users should already be aware that env variables declared in hooks are accessible when those hooks are run. That's just how shell scripts work.

pchico83: @caervs I don't think it is right in any case, it is an unsecure workaround. I think we shouldn't document that practice at all (which is also unrelated to Docker Cloud). If we talk about it, it should be to show the limitations of this practice and explain the solution using build env vars.

pkennedyr: My recommendation would be to keep this objective and describe how environment variables can be used to accomplish a specific use case where the user does not want to expose confidential info in their source repo (i.e. env variables can thus be declared in hooks which are accessible when those hooks are run). We're not assigning value to either approach, simply describing how a given feature can be used to accommodate a desired behavior.

pkennedyr:
LGTM

sanscontext:
@joaofnfernandes I'm stuck behind a review-issue again. Can you take a look?

PR #392, opened 2016-10-30
Update centos.md

xcellardoor:
I've added in the '--rm' argument to the 'script method' and to the 'add your user to docker group' examples to match the first instruction. Purely a consistency thing, and so new users can appreciate not having a few hello-world containers spawned. Ta.

Applies to: Docker Documentation


johndmulhausen:
Yep, I see that we start with using `--rm` then trail off. Thanks, @xcellardoor :)

PR #396, opened 2016-10-31
Update index page title and description

joaofnfernandes:
Updates /index.html so that it presents a better description when you bookmark the page, and in search engines.

/ping @docs.maintainers

### DIFF
--- a/_data/toc.yaml
+++ b/_data/toc.yaml
@@ -1,5 +1,5 @@
 toc:
-- title: Welcome to the Docs
+- title: What is Docker?
   path: /
 - sectiontitle: Docker Engine
   section:
--- a/index.md
+++ b/index.md
@@ -1,12 +1,12 @@
 ---
-description: Home page for Docker's documentation
+title: Docker documentation
+description: Learn how to build, ship, and run the next generation of distributed applications
 keywords:
 - Docker, documentation, manual, guide, reference, api
-title: Welcome to the Docs
 layout: docs
 ---
 
-# Welcome to the Docs
+# What is Docker?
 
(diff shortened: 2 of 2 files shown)


sanscontext:
Not a fan of this wording - it sort of implies that we, the docs team, don't know what Docker is. I'd prefer the "Docker Documentation" phrase you'd suggested in chat.

joaofnfernandes:
Our index page is kind of strange, in the sense that it is an index page and also shows in the navigation menu.

The current proposal looks like this

screen shot 2016-10-31 at 12 02 53
(notice the browser tab)

My idea for "what is Docker" came from the website.
screen shot 2016-10-31 at 12 01 33

_data/toc.yaml
```diff
@@ -1,5 +1,5 @@
 toc:
-- title: Welcome to the Docs
+- title: What is Docker?
```
johndmulhausen: Some possibilities: 

Docker Documentation
Documentation Home
(No entry at all) 

I think these are all better than a left-nav entry that says "What is Docker?" 

cc @sanscontext @londoncalling @mstanleyjones 

WDYT?

johndmulhausen: No entry is viable in the case of the home page; either way, the left-nav tree will be completely collapsed. The only difference is there wouldn't be a blue-colored nav entry on the left side.

mstanleyjones: I prefer "Docker Documentation" or "Docker Docs". From a quick google I found "AWS Documentation", "Oracle Help Center", "Microsoft Technical Documentation", "Guides and Sample Code - Apple Developer",

PR #402, opened 2016-10-31
Add alias to fix #388

### LINKED ISSUE #388: 404 at: /reference/logging/fluentd/
URL: /reference/logging/fluentd/

johndmulhausen:
Fixes 404

### DIFF
--- a/engine/admin/logging/fluentd.md
+++ b/engine/admin/logging/fluentd.md
@@ -1,6 +1,7 @@
 ---
 aliases:
 - /engine/reference/logging/fluentd/
+- /reference/logging/fluentd/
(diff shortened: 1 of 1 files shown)


engine/admin/logging/fluentd.md
```diff
@@ -1,6 +1,7 @@
 ---
 aliases:
 - /engine/reference/logging/fluentd/
+- /reference/logging/fluentd/
```
thaJeztah: Is it possible to add redirects for `/reference/*` -> `/engine/reference/*`, and `/reference/logging/*` -> `/admin/logging/*` ? These links exist because that's were those docs were previously (so links may exist to the other logging drivers as well (etc.)

PR #407, opened 2016-11-01
[WIP] Migrates UCP 1.1 docs with history

joaofnfernandes:
The previous migration to this tree did not persisted the history for the UCP docs. This solves that.
/ping @mstanleyjones for a quick check


mstanleyjones:
I thought we specifically didn't want to keep the history from the private repos. @johndmulhausen for opinions....

joaofnfernandes:
Having history will make it easier when we try submitting docs for the next version of UCP/DTR.
This PR preserves all the changes that have been made to UCP docs in this repo, it doesn't discard them.

johndmulhausen:
It's quite alright to preserve history if that's what you want, Joao.
Private-sourced changes getting published to the public repo are not bound
to preserving or removing history. If it serves a purpose to bring history
along, by all means do.

PR #409, opened 2016-11-01
Fix some Liquid warnings

mstanleyjones:

Wrap some Docker log template directives in `{% raw %}` because when Liquid sees `{{ }}` it gets really excited and treats them as mix-ins.

### Project version

n/a

### Related issue

n/a

n/a

### Please take a look

@johndmulhausen @joaofnfernandes @londoncalling

### DIFF
--- a/machine/reference/inspect.md
+++ b/machine/reference/inspect.md
@@ -61,8 +61,10 @@ For the most part, you can pick out any field from the JSON in a fairly
 straightforward manner.
 
 ```none
+{% raw %}
 $ docker-machine inspect --format='{{.Driver.IPAddress}}' dev
 192.168.5.99
+{% endraw %}
 ```
 
 **Formatting details:**
@@ -78,7 +80,8 @@ $ docker-machine inspect --format='{{json .Driver}}' dev-fusion
 While this is usable, it's not very human-readable. For this reason, there is
 `prettyjson`:
 
-```
+```none
+{% raw %}
 $ docker-machine inspect --format='{{prettyjson .Driver}}' dev-fusion
 {
   "Boot2DockerURL": "",
@@ -97,4 +100,5 @@ $ docker-machine inspect --format='{{prettyjson .Driver}}' dev-fusion
   "SwarmHost": "tcp://0.0.0.0:3376",
   "SwarmMaster": false
 }
+{% endraw %}
 ```
--- a/machine/reference/ls.md
+++ b/machine/reference/ls.md
@@ -103,16 +103,20 @@ The following example uses a template without headers and outputs the `Name` and
 for all running machines:
 
 ```none
+{% raw %}
 $ docker-machine ls --format "{{.Name}}: {{.DriverName}}"
 default: virtualbox
 ec2: amazonec2
+{% endraw %}
 ```
 
 To list all machine names with their driver in a table format you can use:
 
 ```none
+{% raw %}
 $ docker-machine ls --format "table {{.Name}} {{.DriverName}}"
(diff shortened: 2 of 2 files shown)


machine/reference/inspect.md
```diff
@@ -61,8 +61,10 @@ For the most part, you can pick out any field from the JSON in a fairly
 straightforward manner.
 
 ```none
+{% raw %}
 $ docker-machine inspect --format='{{.Driver.IPAddress}}' dev
```
joaofnfernandes: It looks like that even with `{% raw %}` some component (probably highligh js) is trying to interpret the `{{.Driver.IPAddress}}` part.
I'm getting a JS error on my browser, when previewing `/machine/reference/inspect/` on my browser.

I could fix this error by changing `{{.Driver.IPAddress}}` to `{{Driver.IPAddress}}`. But that's not the Docker syntax.

```
alljs.js:2 Error: [$parse:syntax] Syntax Error: Token '.' not a primary expression at column 1 of the expression [.Driver.IPAddress] starting at [.Driver.IPAddress].
http://errors.angularjs.org/1.3.14/$parse/syntax?p0=.&p1=not%20a%20primary%20expression&p2=1&p3=.Driver.IPAddress&p4=.Driver.IPAddress
    at alljs.js:1
    at Parser.throwError (alljs.js:2)
    at Parser.primary (alljs.js:2)
    at Parser.unary (alljs.js:2)
    at Parser.multiplicative (alljs.js:2)
    at Parser.additive (alljs.js:2)
    at Parser.relational (alljs.js:2)
    at Parser.equality (alljs.js:2)
    at Parser.logicalAND (alljs.js:2)
    at Parser.logicalOR (alljs.js:2)
```

johndmulhausen:
I haven't tested to the degree Joao has, but I do know for code blocks you can also use ````liquid` at the start of your code block and it will process curly braces directly.

machine/reference/inspect.md
mstanleyjones: With my PR I do not get any Liquid warnings, even after doing `jekyll clean` before `jekyll serve`. That's interesting that you see the error only in the JS console. How in the world do you know which pages to preview? How can we tell which of them have problems?

joaofnfernandes: A couple of days ago I tried fixing this, that's why I noticed the JS errors. Here's what I did:

1. jekyll clean
2. jekyll serve
3. Take not of the pages that had liquid warnings
4. Try to fix those warnings
3. jekyll serve
4. Test my changes.

I noticed that in some cases, the result was not rendering properly, there were some snippets where the `{{<field>}}` part was being removed from the html.
Since my change was essentially what you did, I wanted to inspect what html was being generated for those pages. That's when I noticed the JS error.

From what I could, when we had `{{.<field>}}` (it started with a dot), I would get JS erros. I tried fiddling with it but could not get it to work properly. In some situations the text out be removed from the generated html, in other cases it would generate JS errors.

Maybe try the syntax suggested by @johndmulhausen, where you only escape what's inside the code fences.

    ```bash
    {% raw %}
    <sample>
    {% endraw %}
    ```

mstanleyjones: @joaofnfernandes The second dot warning is not actually about the log format at all! It's about a block of invalid JSON which has a truncation mark '...' in it. Let me push some more changes into this PR to get warning-free.

johndmulhausen:
Since the JS error is related to Angular weirdness, let's get this in to stop any weirdness around Liquid/rendering.

PR #410, opened 2016-11-01
Error-on-purpose so github status check fails.

johndmulhausen:
Junk PR for GitHub support to use as repro. As you can see below, this PR never passes the built-in github/pages status check, showing the message:

`github/pages — Waiting for status to be reported`

### DIFF
--- a/index.md
+++ b/index.md
@@ -8,7 +8,7 @@ layout: docs
 
 # Welcome to the Docker Documentation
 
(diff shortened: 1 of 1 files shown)


johndmulhausen:
GitHub support got back to me that they recommend we not use this; it's for status-checking non-default gh-pages branches and they aren't going to fix it any time soon.

PR #412, opened 2016-11-01
Add topic about troubleshooting volume problems

mstanleyjones:

Lots of people have run into issues where one container bind-mounts `/var/lib/docker` and they can't stop a different container because of "device or resource busy" errors. This topic addresses this type of problem in a way that can hopefully be surfaced by searching Google or docs.docker.com. 

I chose to add this to the formal docs, but I'm open to the idea of writing up a knowledgebase article about it instead. I know "troubleshooting" topics in formal docs are contentious.

There is a vague idea that this may be fixed by some specific kernel version but I was not able to find anything to corroborate this.

### Project version

n/a

### Related issue

n/a

https://github.com/docker/docker/issues/22734
https://github.com/docker/docker/issues/17823
https://github.com/docker/docker/issues/22877

### Please take a look

@thaJeztah @cpuguy83 @jpetazzo @JustinINevill 

Signed-off-by: Misty Stanley-Jones <misty@docker.com>

### DIFF
--- a/_data/toc.yaml
+++ b/_data/toc.yaml
@@ -179,6 +179,10 @@ toc:
       title: Runtime metrics
     - path: /engine/admin/ambassador_pattern_linking/
       title: Link via an ambassador container
+  - sectiontitle: Troubleshoot Docker Engine
+    section:
+    - path: /engine/admin/troubleshooting_volume_errors.md
+      title: Troubleshoot volume problems
   - sectiontitle: Manage a swarm
     section:
     - path: /engine/swarm/
--- a/engine/admin/troubleshooting_volume_errors.md
+++ b/engine/admin/troubleshooting_volume_errors.md
@@ -1,0 +1,62 @@
+---
+description: Troubleshooting volume errors
+keywords:
+- cadvisor, troubleshooting, volumes, bind-mounts
+title: Troubleshoot volume errors
+---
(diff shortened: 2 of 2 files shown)


cpuguy83:
We can recommend using `rslave` (e.g. `-v /var/lib/docker:/var/lib/docker:ro,rslave`) to mitigate this.
This means that any changes to any mount underneath /var/lib/docker will propagate to the container.

The downside to this is it requires to root mount that `/var/lib/docker` is on to be either set to `shared` or `slave`. Most distros are now shipping with `/` set to `shared` by default, but this is relatively new.
We can give more details on setting this up if needed.
WDYT?

johndmulhausen:
I'm not sure "sectionTitle" is okay in the YAML
I'm also not sure the space between the back ticks and lexer name is okay in the snippets. :(

johndmulhausen:
Deploy preview processing.

Built with commit d223eb3a37af6761b710ea876f0644b2ced26bc7

https://app.netlify.com/sites/docker-docs/deploys/582208a7a7e2d23b2475941c

mstanleyjones:
I was playing with a different editor and had no idea it added the space between the fence and lexer. I've fixed the problems and rebased on current `master` to get in all recent fixes.

PR #415, opened 2016-11-01
Add SSH key path option to Digital Ocean doc

v4n:
Modify the docs based on this PR https://github.com/docker/machine/pull/3761


v4n:
@nathanleclaire might want to check PR along the docker/machine PR when you get a chance.

mstanleyjones:
The upstream PR only seems to be against the machine repo's `master`. If this is a new feature, the doc PR needs to be against `vnext-machine` rather than `master` Please rebase if necessary.

@londoncalling PTAL

v4n:
@mstanleyjones this is indeed a new feature that has been merged into machine's master.

Correct me if I'm wrong. But isn't this PR already pointed to be merged into `vnext-machine` rather than `master`?

> v4n wants to merge 1 commit into docker:vnext-machine from v4n:vnext-machine

PR #417, opened 2016-11-01
Update note about custom certs with system certs

dmcgowan:

Docker 1.13 will now merge custom certificates with the system certificate pool. This change updates the note about when the system certificate pool is used. Also fixed incorrect statement about example layout. The example was only showing custom certificates, and not multiple certificates.

### Project version

 Docker Engine 1.13

https://github.com/docker/docker/pull/27918

### Please take a look

@thaJeztah

### DIFF
--- a/engine/security/certificates.md
+++ b/engine/security/certificates.md
@@ -27,8 +27,10 @@ A custom certificate is configured by creating a directory under
 `localhost`). All `*.crt` files are added to this directory as CA roots.
 
 > **Note:**
-> In the absence of any root certificate authorities, Docker
-> will use the system default (i.e., host's root CA set).
+> As of docker 1.13, on Linux any root certificates authorities will be merged
+> in with the system defaults (i.e., host's root CA set). Prior to 1.13 and on
+> Windows, the system default certificates will only be used when there are no
+> custom root certificates provided.
 
 The presence of one or more `<filename>.key/cert` pairs indicates to Docker
 that there are custom certificates required for access to the desired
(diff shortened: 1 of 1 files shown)


thaJeztah:
This was closed due to some unrelated `fixes #xyz` mentions in an old commit :joy:

engine/security/certificates.md
```diff
@@ -29,8 +29,10 @@ A custom certificate is configured by creating a directory under
 `localhost`). All `*.crt` files are added to this directory as CA roots.
 
 > **Note:**
-> In the absence of any root certificate authorities, Docker
-> will use the system default (i.e., host's root CA set).
+> As of docker 1.13, on Linux any root certificates authorities will be merged
+> in with the system defaults (i.e., host's root CA set). Prior to 1.13 and on
+> Windows, the system default certificates will only be used when there are no
```
thaJeztah: asked @dmcgowan and unfortunately, the Go update required for this feature wasn't implemented for Windows

thaJeztah:
oh, hold on, this should be in the `vnext-engine` branch, let me try if I can switch it on GitHub 🌟

engine/security/certificates.md
joaofnfernandes: @thaJeztah this is only tangentially related to this change, but I'm wondering if you need to restart the Docker daemon for Docker to load these certificates?
My understanding is that if you were adding certs to the host trusted CA store, you would need to restart the daemon. If you do need to restart, maybe we need to mention it.

dmcgowan: https://github.com/golang/go/issues/16736, looks like will be fixed in 1.8

dmcgowan: @joaofnfernandes restart is not needed, these certificates are always read at the start of a push or pull operation and not kept around after

thaJeztah:
It's on the right branch now, and https://github.com/docker/docker/pull/27918 was merged, so this should probably be ready to go.

@joaofnfernandes for review/merge :smile:

joaofnfernandes:
@dmcgowan thanks for helping with docs.
LGTM, firing away!

PR #418, opened 2016-11-02
absolute links to docs.docker.com are now relative links

aduermael:

Absolute links to `docs.docker.com` are now relative links. 

I used that command: 

```bash
$ ag -l "]\(https://docs.docker.com" . | xargs sed -i "" -e 's/\](https:\/\/docs.docker.com/\](/g'
$ ag -l "]\(http://docs.docker.com" . | xargs sed -i "" -e 's/\](http:\/\/docs.docker.com/\](/g'
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -107,7 +107,7 @@ You have three options:
     ```
 
     If you haven't got Docker Compose installed,
-    [follow these installation instructions](https://docs.docker.com/compose/install/).
+    [follow these installation instructions](/compose/install/).
 
     The container runs in the background and incrementally rebuilds the site each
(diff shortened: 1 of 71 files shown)


johndmulhausen:
Thanks a ton @aduermael ! This catches all the instances where markdown links were absolute! There are still HTML links (`href="https://docs.docker.com/...`), but this is a great start. Now staging and forking will still produce links that work!

aduermael:
@johndmulhausen you're welcome! 🙂

> Now staging and forking will still produce links that work!
Yes, that's what I was experiencing in my fork! 😝

PR #421, opened 2016-11-02
Fix regex capture groups syntax

aureliojargas:
At least in Docker Hub, the correct syntax is `{\1}`, not `{/1}`.


sanscontext:
Good catch, thanks @aureliojargas!

PR #422, opened 2016-11-02
fix: Add line breaks before and after code blocks

gokaygurcan:
Missing line breaks were causing some problems on the live page, the code blocks were like these:

image

image

image


johndmulhausen:
Thank you, @gokaygurcan !

PR #430, opened 2016-11-02
Merge master into vnext-opensource

mstanleyjones:

Pick up recent fixes since the migration

### Project version

opensource next

### Related issue

n/a

n/a

### Please take a look

@johndmulhausen @thaJeztah


.mailmap
```diff
@@ -1,254 +0,0 @@
-# Generate AUTHORS: hack/generate-authors.sh
```
thaJeztah: Are we not going to do an AUTHORS file?

engine/examples/mongodb.md
```diff
@@ -69,53 +72,60 @@ Continuing, we will declare the `MAINTAINER` of the `Dockerfile`:
...
+```
 
 Lastly we set the `ENTRYPOINT` which will tell Docker to run `mongod` inside
 the containers launched from our MongoDB image. And for ports, we will use
 the `EXPOSE` instruction.
 
-    # Expose port 27017 from the container to the host
-    EXPOSE 27017
+```dockerfile
+# Expose port 27017 from the container to the host
```
thaJeztah: Hm, something we should fix "expose" doesn't actually _expose_, it only adds metadata to the dockerfile to describe that the image "exposes" a certain port (it's used for the `-P` (publish all) option)

engine/examples/mongodb.md
```diff
@@ -69,53 +72,60 @@ Continuing, we will declare the `MAINTAINER` of the `Dockerfile`:
...
+# Set usr/bin/mongod as the dockerized entry-point application
+ENTRYPOINT ["/usr/bin/mongod"]
+```
 
 Now save the file and let's build our image.
 
-> **Note:**
->
-> The full version of this `Dockerfile` can be found [here](https://github.com/docker/docker/blob/master/docs/examples/mongodb/Dockerfile).
+> **Note:**  The full version of this `Dockerfile` can be found [here](https://github.com/docker/docker/blob/master/docs/examples/mongodb/Dockerfile).
```
thaJeztah: Looks like we need to update this URL

engine/reference/commandline/build.md
```diff
@@ -94,11 +96,9 @@ Build Syntax Suffix             | Commit Used           | Build Context Used
 
 If you pass an URL to a remote tarball, the URL itself is sent to the daemon:
 
-Instead of specifying a context, you can pass a single Dockerfile in the `URL`
-or pipe the file in via `STDIN`. To pipe a Dockerfile from `STDIN`:
-
```
thaJeztah: Not sure this should be removed

engine/reference/commandline/dockerd.md
```diff
@@ -1090,71 +1092,71 @@ This is a full example of the allowed configuration options on Linux:
 
 ```json
 {
-	"authorization-plugins": [],
```
mstanleyjones: They won't be merged back at all. Instead, the contents will be re-copied from `docker/docker` at some point in the future, at which point the newly added options would show up.

engine/examples/mongodb.md
mstanleyjones: OK, but that's not really in scope for this PR.

engine/reference/commandline/build.md
mstanleyjones: You are probably right, I should do a follow-on commit that re-replaces the reference docs with what is currently in engine.

mstanleyjones: This was done in 7b41a45abe004f7d5a7d69f1eba6650f28e68e76 which may need to be reverted in `master`. I'll check into it.

mstanleyjones: https://github.com/docker/docker.github.io/pull/433

PR #453, opened 2016-11-04
Update config-storage.md

banjot:
### Please take a look

Added a line to ensure it's clear that use of filesystem is incompatible with DTR HA and multiple replica set up.


banjot:
Added a single line to point out that filesystem storage is incompatible with DTR HA.

mstanleyjones:
@joaofnfernandes PTAL

joaofnfernandes:
LGTM

PR #448, opened 2016-11-03
fix frontmatter keywords values in compose/reference/ files

gdevillele:
### Describe the proposed changes

fixes files in which the keywords value is of the wrong type.

### Project version

current stable

### Related issue

contributes to #435 

### Related issue or PR in another project

N/A

### Please take a look

N/A

Signed-off-by: Gaetan de Villele <gdevillele@gmail.com>

PR #458, opened 2016-11-04
Revert "fix frontmatter keywords values in compose/reference/ files"

mstanleyjones:
Reverts docker/docker.github.io#448

This is causing  the home page to break. Reverting.


gdevillele:
This one was not in cause in the home page issue.

PR #459, opened 2016-11-04
Revert "fix some frontmatter keywords values"

mstanleyjones:
Reverts docker/docker.github.io#437

This is breaking the home page. Reverting.


gdevillele:
This one was not in cause in the home page issue.

PR #460, opened 2016-11-04
Revert "Merge pull request #437 from gdevillele/fix_keywords_format"

mstanleyjones:
Reverting because it is breaking the home page


mstanleyjones:
cc @dgageot

PR #476, opened 2016-11-06
Clarify the fact that VOLUME does not support host-dir

alexisbellido:

### DIFF
--- a/engine/tutorials/dockervolumes.md
+++ b/engine/tutorials/dockervolumes.md
@@ -108,10 +108,6 @@ The `host-dir` can either be an absolute path or a `name` value. If you
 supply an absolute path for the `host-dir`, Docker bind-mounts to the path
 you specify. If you supply a `name`, Docker creates a named volume by that `name`.
 
-> **Note:**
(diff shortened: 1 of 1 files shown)


johndmulhausen:
Deploy preview processing.

Built with commit b01714f7f5fe2c705cda5032d5dcfa3708582722

https://app.netlify.com/sites/docker-docs/deploys/581f6323a7e2d27d7e81bd1c

johndmulhausen:
Are we sure this is necessary? This doc already says: "Note: The host directory is, by its nature, host-dependent. For this reason, you can’t mount a host directory from Dockerfile because built images should be portable. A host directory wouldn’t be available on all potential hosts."

alexisbellido:
Fair enough.  Now that I know more about how volumes and containers work, the note you mention makes more sense; but I didn't grasp the concept right away when I started with Docker. I suspect others may be in the same situation.

I even found a tutorial that showed a Dockerfile with VOLUME using host directories.

What about adding the new text at the end of the existing note? It may feel redundant but still help to get the point the across.

Issue #435, opened 2016-11-02
Fix inconsistency of frontmatter "keywords" values


gdevillele:
### Problem description

Context: some pages have a `keywords` entry in their frontmatter header. 
- The type of those entries is not consistent across the repository
- It does not make sense to have an array-typed value as there is no value in having several keyword meta tags in the pages served
- The frontmatter parser that is used currently, in some cases, fails to build the correct html meta tag.

https://docs.docker.com/docker-trusted-registry/support/ contains:
```
<meta name="keywords" content="Docker, supporthelp">
```

https://docs.docker.com/engine/swarm/how-swarm-mode-works/pki/ contains:
```
<meta name="keywords" content="dockercontainerclusterswarm modenodetlspki">
```

### Problem location

In markdown files, across the whole repository.

### Project version(s) affected

current stable

### Suggestions for a fix

Fixing the content would be a good start.

I am going to submit PRs soon.

gdevillele:
Thanks for having reviewed and merged all this. 🙂

Closing.

PR #488, opened 2016-11-08
fix type of keywords entry in frontmatter (in /docker-cloud/ dir)

gdevillele:

in /docker-cloud/ :
change keywords's type from array to string.

### Project version

current stable

### Related issue

contributes to #435 

Signed-off-by: Gaetan de Villele <gdevillele@gmail.com>

### DIFF
--- a/docker-cloud/apps/api-roles.md
+++ b/docker-cloud/apps/api-roles.md
@@ -1,9 +1,8 @@
 ---
+description: API Roles
+keywords: API, Services, roles
 redirect_from:
 - /docker-cloud/feature-reference/api-roles/
-description: API Roles
-keywords:
-- API, Services, roles
 title: Service API roles
 ---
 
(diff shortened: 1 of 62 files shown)


johndmulhausen:
Deploy preview processing.

Built with commit bcec1ddcb2366faf4067af322ba46f17f382acd5

https://app.netlify.com/sites/docker-docs/deploys/58218ab8668674753a544cfd

sanscontext:
LGTM, though I'm not sure if I should be seeing any different behavior here. (Google results?)

Also, @gdevillele, does your editor do something interesting with the final line? It looks like there's a spurious end of file change in all these diffs, which is confusing.

gdevillele:
@sanscontext Thank you.

Behavior is expected to be exactly the same.

About the last line, I have just removed trailing LF (line feed) characters.
It can be confusing in the Github UI as they are non-printable characters.

sanscontext:
Ah, interesting. I wonder if Atom is/was adding those.

gdevillele:
@sanscontext I cannot know for sure, but I suppose it is just the writer who pressed the ENTER key at the end of the file. It isn't more than that 🙂

sanscontext:
Merged! Sorry for the delay, little distracted with updates yesterday :)

gdevillele:
Thx. No problem.

PR #508, opened 2016-11-09
Implement per-page reading-time stats

mstanleyjones:

Implemented a simple reading-time metric which displays just below the page title on every page

Includes some basic styling, but I leave feedback on styling and positioning to others

Uses the metric of 180 wpm, as that is what this [Wikipedia article](http://en.wikipedia.org/wiki/Words_per_minute) thinks you can read on a screen.

Credit to https://carlosbecker.com/posts/jekyll-reading-time-without-plugins/ for the implementation

n/a

### Related issue

n/a

n/a

### Please take a look

@johndmulhausen @sanscontext @londoncalling @joaofnfernandes 

@hansfrederickbrown can you please ping the right person on the Design team to PTAL?

See the Netlify staged site when it completes.

### DIFF
--- a/css/documentation.css
+++ b/css/documentation.css
@@ -346,3 +346,17 @@ color: #F04124;
 }
 
 /* end search */
+
+/* begin reading_time */
+
+span.reading-time {
+  font-style: italic;
+  font-size: 80%;
+  display: block;
+  padding-top: 10px;
+  padding-bottom: 10px;
+  color: #999;
+}
+
+span.reading-time-label {
+}
--- a/_includes/read_time.html
+++ b/_includes/read_time.html
@@ -1,0 +1,9 @@
+<span class="reading-time" title="Estimated reading time">
+  <span class="reading-time-label">Estimated reading time: </span>
+  {% assign words = content | number_of_words %}
+  {% if words < 360 %}
+    1 minute
+  {% else %}
+    {{ words | divided_by:180 }} minutes
+  {% endif %}
+</span>
--- a/_layouts/docs.html
+++ b/_layouts/docs.html
@@ -187,6 +187,7 @@ ng\:form {
               <section class="section" id="DocumentationText">
 								{% if page.title %}<h1>{{ page.title }}</h1>{% endif %}
 								{% if page.advisory %}<blockquote><p><strong style="color:black">{{ site.data.advisories.texts[page.advisory] }}</strong></p></blockquote>{% endif %}
+              {% include read_time.html %}
               {{ content }}
 							{% if page.noratings != true %}
 							<div style="text-align: center; margin-top: 50px">


johndmulhausen:
Deploy preview processing.

Built with commit 88a1f6c680a235b41d8c787f8188b1177db743dd

https://app.netlify.com/sites/docker-docs/deploys/58237fbd71e20a5b82e0dafa

joaofnfernandes:
Why not use https://github.com/gjtorikian/jekyll-time-to-read?

mstanleyjones:
@joaofnfernandes it requires a plugin, which limits our flexibility. And I'm not sure what it gets us over my solution. What do you like better about it?

joaofnfernandes:
It leverages a component that is being used by a lot of people while freeing us from doing custom development.

mstanleyjones:
We still have to do the CSS. And Github pages (netlify) won't allow
plugins. I don't see the benefit.

On Nov 9, 2016, at 1:59 PM, Joao Fernandes <notifications@github.com> wrote:

It leverages a component that is being used by a lot of people while
freeing us from doing custom development.

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/docker/docker.github.io/pull/508#issuecomment-259538826>,
or mute the thread

joaofnfernandes:
> Github pages (netlify) won't allow plugins

From what I can see in [_config.yml](https://github.com/docker/docker.github.io/blob/6fd776fabcd88efb2d60dda72a4bc4f6ef29ef08/_config.yml) we are using 3 plugins already.

johndmulhausen:
Great discussion here. To clarify, the three plugins we are using are supported by GitHub Pages. They do allow a handful of them: 

https://pages.github.com/versions/

Also, to be clear, Misty is already using something in this PR that is provided by Jekyll and maintained by lots of contributors; `number_of_words` is a built-in feature that doesn't require a plugin to use. Search for it here: https://jekyllrb.com/docs/templates/

Given that we are essentially just dividing `number_of_words` by 180 and putting some CSS on it, I'm comfortable with the approach here, but it would be great for the UX team to have a look at the deploy preview and give us feedback.

@docker/design -- Any comments? We're talking about how best to surface estimated reading times, as shown here: https://deploy-preview-508--docker-docs.netlify.com

css/documentation.css
```diff
@@ -346,3 +346,18 @@ color: #F04124;
 }
 
 /* end search */
+
+/* begin reading_time */
+
+span.reading-time {
+  font-weight: bold;
```
johndmulhausen: @mstanleyjones  I think it looks better with bold off. Other than that, I say we pull this is and work with design on better styling in the general effort.

mstanleyjones:
I took off the bold. One thing I think that might be an obvious iteration is to make the reading time available in the Twitter card (which is shown when copying the URL into Slack and other contexts) and maybe in the Google search results somehow. I don't know how to do that, and I suspect that it would require tweaking this approach a bit to make the reading time available in the `<head>` of the page.

Assuming this looks ok when Netlify finishes, I'll merge it as is so we can iterate.

PR #531, opened 2016-11-11
Document ability to update a service's image

### LINKED ISSUE #528: Document ability to pin images by digest
### Problem description

Document the work done in https://github.com/docker/docker/pull/28173

Engine 1.13

cc/ @nishanttotla

mstanleyjones:

Document the ability to update a service's image using `service update -image`

Engine 1.13

### Related issue

Fixes #528 

https://github.com/docker/docker/pull/28173

### Please take a look

@nishanttotla

### DIFF
--- a/engine/swarm/services.md
+++ b/engine/swarm/services.md
@@ -70,6 +70,135 @@ $ docker service create --name helloworld alpine ping docker.com
 9uk4639qpg7npwf3fn2aasksr
 ```
 
+### Specify the image version the service should use
+
+When you create a service without specifying any details about the version of
+the image to use, the service uses the latest version available in Docker Hub
+or your registry. You can force the service to use a specific version of
+the image in a few different ways, depending on your desired outcome.
+
+An image version can be expressed in several different ways:
+
+- If you specify a tag, the manager resolves that tag to a digest, and directs
+  workers to uses that version.
+  ```bash
+  $ docker service create --name="myservice" ubuntu:16.04
+  ```
+
+  Some tags represent discrete releases, such as `ubuntu:16.04`. Tags like this
+  will almost always resolve to a stable digest over time. It is recommended
+  that you use this kind of tag when possible.
+
+  Other types of tags, such as `latest` or `nightly`, may resolve to a new
+  digest often, depending on how often an image's author updates the tag. It is
+  not recommended to run services using a tag which is updated frequently, to
+  prevent different service replica tasks from using different image versions.
+
+- If you don't specify a version at all, by convention the image's `latest` tag
+  is resolved to a digest. The following two commands are equivalent:
+  ```bash
+  $ docker service create --name="myservice" ubuntu
+
+  $ docker service create --name="myservice" ubuntu:latest
+  ```
+
+- If you specify a digest directly, that exact version of the image is always
+  used.
+  ```bash
+  $ docker service create --name="myservice" ubuntu:16.04@sha256:35bc48a1ca97c3971611dc4662d08d131869daa692acb281c7e9e052924e38b1
+  ```
+
+When you create a service, the swarm manager resolves the
+image's tag to the specific digest it points to at the time of service creation.
+If the manager is able to resolve the tag to a digest, worker nodes for that
+service will use that specific digest unless the service is explicitly updated.
+This feature is particularly important if you do use often-change tags such as
+`latest`, to be sure that all service tasks use the same version of the image.
+
+If the manager is not able to resolve the tag to a digest, each worker
+node is responsible for resolving the tag to a digest, and different nodes may
+use different versions of the image. If this happens, a warning like the
+following will be logged, substituting the placeholders for real information.
+
+```none
+unable to pin image <image_name> to digest: <reason>
+```
+
+To see an image's current digest, issue the command `docker inspect
+<image>:<tag>` and look for the `RepoDigests` line. The following is the current
+digest for `ubuntu:latest` at the time this content was written. The output is
+truncated for clarity.
+
+```bash
+$ docker inspect ubuntu:latest
+```
+
+```json
+"RepoDigests": [
+    "ubuntu@sha256:35bc48a1ca97c3971611dc4662d08d131869daa692acb281c7e9e052924e38b1"
+],
+```
+
+After you create a service, its image is never updated unless you explicitly run
+`docker service update` with the `--image` flag as described below. Other update
+operations such as scaling the service, adding or removing networks or volumes,
+renaming the service, or any other type of update operation do not update the
+service's image.
+
+### Update a service's image after creation
+
+Each tag represents a digest, similar to a Git hash. Some tags, such as
+`latest`, are updated often to point to a new digest. Others, such as
+`ubuntu:16.04`, represent a released software version and are not expected to
+update to point to a new digest often if at all. In Docker 1.13 and higher, when
+you create a service, it is constrained to run a specific digest of an image
+until you update the service using `service update` with the `--image` flag. If
+you use an older version of Docker Engine, you must remove and re-create the
+service to update its image.
+
+When you run `service update` with the `--image` flag, the swarm manager queries
+Docker Hub or your private Docker registry for the digest the tag currently
+points to and updates the service tasks to use that digest.
+
+Usually, the manager is able to resolve the tag to a new digest and the service
+updates, redeploying each task to use the new image. If the manager is unable to
+resolve the tag or some other problem occurs, the next two sections outline what
+to expect.
+
+#### If the manager resolves the tag
+
+If the swarm manager can resolve the image tag to a digest, it instructs the
+worker nodes to redeploy the tasks and use the image at that digest.
+
+- If a worker has cached the image at that digest, it uses it.
+
+- If not, it attempts to pull the image from Docker Hub or the private registry.
+
+  - If it succeeds, the task is deployed using the new image.
+
+  - If the worker fails to pull the image, the service fails to deploy on that
+    worker node. Docker tries again to deploy the task, possibly on a different
+    worker node.
+
+#### If the manager cannot resolve the tag
+
+If the swarm manager cannot resolve the image to a digest, all is not lost:
+
+- The manager instructs the worker nodes to redeploy the tasks using the image
+  at that tag.
+
+- If the worker has a locally cached image that resolves to that tag, it uses
+  that image.
+
+- If the worker does not have a locally cached image that resolves to the tag,
+  the worker tries to connect to Docker Hub or the private registry to pull the
+  image at that tag.
+
+  - If this succeeds, the worker uses that image.
+
+  - If this fails, the task fails to deploy and the manager tries again to deploy
+    the task, possibly on a different worker node.
+
 ## Configure the runtime environment
 
 You can configure the following options for the runtime environment in the


engine/swarm/services.md
```diff
@@ -70,6 +70,42 @@ $ docker service create --name helloworld alpine ping docker.com
...
+You can specify any tag which has been pushed to Docker Hub or your private
+registry. For instance, some of the tags available for the `ubuntu` image are
+`latest`, `devel`, `yakkety`, and `16.04`.
+
+### Update a service's image after creation
+
+Each tag represents a digest, similar to a Git hash. Some tags, such as
+`latest`, are updated occasionally to point to a new digest. Services are
+constrained to run a specific digest of an image until you update the service
+using `service update` with the `-image` flag, which is supported in Docker
```
nishanttotla: nit: `--image`

nishanttotla:
It might be useful to add some more details for the workflow here, or an example if you think suitable.

In particular:
```
docker service create --name xyz image:tag
```
will create a service with image `image:tag@sha256:34834908r...` based on the resolution of tag to digest at creation time. Even if the digest changes for this tag, any service updates will continue to use the same digest, unless the `--image` flag is added on a service update, in which case the tag will be resolved to digest again, and the service image updated if the digest has changed.

The user can also create a service providing a digest, in which case it will be used directly with no resolution attempt. Additionally, if the image with a certain digest exists locally, Swarm will not pull it (https://github.com/docker/docker/pull/28265), and just use the local copy.

WDYT @mstanleyjones?

nishanttotla:
cc @aaronlehmann

engine/swarm/services.md
```diff
@@ -70,6 +70,42 @@ $ docker service create --name helloworld alpine ping docker.com
 9uk4639qpg7npwf3fn2aasksr
 ```
 
+### Specify the image version the service should use
+
+When you create a service without specifying any details about the version of
+the image to use, the service uses the latest version available in Docker Cloud
```
aaronlehmann: I believe this should be Docker Hub. Docker Cloud is a different product.

engine/swarm/services.md
```diff
@@ -70,6 +70,42 @@ $ docker service create --name helloworld alpine ping docker.com
...
+`latest`, are updated occasionally to point to a new digest. Services are
+constrained to run a specific digest of an image until you update the service
+using `service update` with the `-image` flag, which is supported in Docker
+Engine 1.13 and higher. If you use an older version of Docker Engine, you must
+remove and re-create the service to update its image.
+
+When you run `service update` with the `-image` flag, a background function
+called `ResolveTagToDigest` queries Docker Hub or your private Docker registry
+for the digest the tag points to, and updates the service tasks to use that
+digest. If the tag does not exist or the registry is not reachable, the service
```
aaronlehmann: We should not mention the name of the function from our source code.

```
When you create a service, or run `service update` and specify a new image with
the `--image` flag, Docker queries Docker Hub or your private Docker registry for
the digest the tag points to. It stores that digest on the service so that tasks will use
the image with this digest, ensuring that each node running the service uses the
same version of the image, even if the tag is updated while the service is running.
```

engine/swarm/services.md
```diff
@@ -70,6 +70,42 @@ $ docker service create --name helloworld alpine ping docker.com
...
+continues running as before, still using the old image, and a warning is
+shown.
```
aaronlehmann: If the tag does not exist or the registry is not reachable, the service probably won't work because nodes won't be able to pull the image. The exception is if the image already exists on the nodes that will run the service tasks, perhaps because it has been successfully pulled before.

mstanleyjones: I was specifically talking about when you try to update the image and the updated version is not available. Wouldn't the service just keep running at its current image version in that case?

mstanleyjones:
Addressed all feedback except for the question above.

engine/swarm/services.md
aaronlehmann: The service still gets updated, but instead of getting an image reference with a digest in it, it gets one with a tag, like `wordpress:latest`. So it will only keep working if the node where it's running has an image called `wordpress:latest`. It may not have one, especially because it would have previously pulled something like `wordpress@sha256:d98a7343ac750ffe387e3d514f8521ba69846c216778919b01414b8617cfb3d4`.

mstanleyjones:
@aaronlehmann PTAL again, tried to reword to catch the drift of what you were saying.

engine/swarm/services.md
```diff
@@ -70,6 +70,58 @@ $ docker service create --name helloworld alpine ping docker.com
...
+### Update a service's image after creation
+
+Each tag represents a digest, similar to a Git hash. Some tags, such as
+`latest`, are updated occasionally to point to a new digest. Services are
+constrained to run a specific digest of an image until you update the service
+using `service update` with the `--image` flag, which is supported in Docker
+Engine 1.13 and higher. If you use an older version of Docker Engine, you must
+remove and re-create the service to update its image.
+
+When you run `service update` with the `--image` flag, a background function
```
aaronlehmann: Instead of "background function", let's say "Docker", "the service update operation", or something similar.

mstanleyjones:
@aaronlehmann @stevvooe @nishanttotla PTAL again and let me know if I have captured what happens if the tag can't be resolved.

engine/swarm/services.md
```diff
@@ -70,6 +70,86 @@ $ docker service create --name helloworld alpine ping docker.com
...
 ```
 
+### Specify the image version the service should use
+
+When you create a service without specifying any details about the version of
+the image to use, the service uses the latest version available in Docker Hub
+or your local registry. You can force the service to use a specific version of
+the image in a few different ways, depending on your desired outcome.
+
+When you don't specify a tag, you get the latest version of an image, because
```
aaronlehmann: nit: `latest` tag rather than latest version.

`latest` doesn't have any special properties, so it's possible for it to be out of date. It's only the latest version by convention.

engine/swarm/services.md
```diff
@@ -70,6 +70,86 @@ $ docker service create --name helloworld alpine ping docker.com
...
+Each tag represents a digest, similar to a Git hash. Some tags, such as
+`latest`, are updated occasionally to point to a new digest. Services are
+constrained to run a specific digest of an image until you update the service
+using `service update` with the `--image` flag, which is supported in Docker
+Engine 1.13 and higher. If you use an older version of Docker Engine, you must
+remove and re-create the service to update its image.
+
+When you run `service update` with the `--image` flag, Docker
+queries Docker Hub or your private Docker registry for the digest the tag points
+to. and updates the service tasks to use that digest.
```
aaronlehmann: extra period

engine/swarm/services.md
stevvooe: I'd add some very strong warnings against leaving the state of your cluster to mercy of the `latest` tag.

engine/swarm/services.md
```diff
@@ -70,6 +70,86 @@ $ docker service create --name helloworld alpine ping docker.com
 9uk4639qpg7npwf3fn2aasksr
 ```
 
+### Specify the image version the service should use
+
+When you create a service without specifying any details about the version of
+the image to use, the service uses the latest version available in Docker Hub
+or your local registry. You can force the service to use a specific version of
```
nishanttotla: nit: we can remove `local` here.

engine/swarm/services.md
```diff
@@ -70,6 +70,86 @@ $ docker service create --name helloworld alpine ping docker.com
...
+
+$ docker service create --name="myservice" ubuntu:latest
+```
+
+You can specify any tag which has been pushed to Docker Hub or your private
+registry. For instance, some of the tags available for the `ubuntu` image are
+`latest`, `devel`, `yakkety`, and `16.04`. The image's tag name is resolved to
+the specific digest it points to at the time of service creation. As an
+alternative, you can specify the digest directly, and no digest resolution
+occurs.
```
nishanttotla: Can we explicitly show what an image string looks like, after resolution to digest? Such as `ubuntu:latest@sha256:345kjhfkjaslf...`?

mstanleyjones:
Feedback addressed. PTAL.

engine/swarm/services.md
mstanleyjones: I did address this but not on this line, so Github didn't pick it up as resolved.

engine/swarm/services.md
```diff
@@ -70,6 +70,107 @@ $ docker service create --name helloworld alpine ping docker.com
...
+the image to use, the service uses the latest version available in Docker Hub
+or your registry. You can force the service to use a specific version of
+the image in a few different ways, depending on your desired outcome.
+
+When you don't specify a tag, you get digest that corresponds to the `latest`
+tag of an image, by convention. There is no guarantee that the version tagged
+with `latest` is actually the newest or most up-to-date version. In general, it
+is recommended that you choose a specific version of an image rather than
+relying on `latest`. Otherwise, service tasks that deploy at different times may
+use different versions of the image, and this may not be what you intend.
```
nishanttotla: Did this text change from before?  Sorry if I missed this earlier, but the messaging is confusing here. It's good to be clear that relying on `latest` isn't a good idea, but `service tasks that deploy at different times may use different versions of the image` is exactly what tag->digest resolution is meant to prevent. The only time this can fail is when the registry is unreachable or v1. Maybe it makes sense to specify that?

Especially since this is described below, so just a pointer to that might suffice.

engine/swarm/services.md
```diff
@@ -70,6 +70,128 @@ $ docker service create --name helloworld alpine ping docker.com
...
+- If you don't specify a version at all, by convention the image's `latest` tag
+  is resolved to a _digest_, which is a unique identifier for a version of the
+  image. The following two commands are equivalent:
+  ```bash
+  $ docker service create --name="myservice" ubuntu
+
+  $ docker service create --name="myservice" ubuntu:latest
+  ```
+
+- If you specify a different tag, Docker resolves that tag to a digest, and uses
```
nishanttotla: Let's be specific here `Docker resolves` -> `agent resolves`

engine/swarm/services.md
```diff
@@ -70,6 +70,128 @@ $ docker service create --name helloworld alpine ping docker.com
...
+```
+
+```json
+"RepoDigests": [
+    "ubuntu@sha256:35bc48a1ca97c3971611dc4662d08d131869daa692acb281c7e9e052924e38b1"
+],
+```
+
+After you create a service, its image is never updated unless you explicitly run
+`docker service update` with the `--image` flag as described below. Other update
```
nishanttotla: Again, this is only if the manager is able to resolve the tag to digest. Otherwise this is not guaranteed.

engine/swarm/services.md
```diff
@@ -70,6 +70,128 @@ $ docker service create --name helloworld alpine ping docker.com
...
+  ```
+
+- If you specify a digest directly, that exact version of the image is always
+  used.
+  ```bash
+  $ docker service create --name="myservice" ubuntu:16.04@sha256:35bc48a1ca97c3971611dc4662d08d131869daa692acb281c7e9e052924e38b1
+  ```
+
+**Note: Using `latest`** There is no guarantee that the version tagged
+with `latest` is actually the newest or most up-to-date version. In general, it
```
nishanttotla: The main reason to warn against `latest` is that over time, if image behind the `latest` tag updates, the service could end up with different versions. This text doesn't make that clear.

engine/swarm/services.md
```diff
@@ -70,6 +70,130 @@ $ docker service create --name helloworld alpine ping docker.com
...
+  will almost always resolve to a stable digest over time. It is recommended
+  that you use this kind of tag when possible.
+
+  Other types of tags, such as `latest` or `nightly`, may resolve to a new
+  digest often, depending on how often an image's author updates the tag. It is
+  not recommended to run services using a tag which is updated frequently, to
+  prevent different service replica tasks from using different image versions.
+
+- If you don't specify a version at all, by convention the image's `latest` tag
+  is resolved to aThe following two commands are equivalent:
```
nishanttotla: `aThe` -> `the`

engine/swarm/services.md
```diff
@@ -70,6 +70,130 @@ $ docker service create --name helloworld alpine ping docker.com
...
+- If you specify a digest directly, that exact version of the image is always
+  used.
+  ```bash
+  $ docker service create --name="myservice" ubuntu:16.04@sha256:35bc48a1ca97c3971611dc4662d08d131869daa692acb281c7e9e052924e38b1
+  ```
+
+When you create a service, the swarm manager resolves the
+image's tag to the specific digest it points to at the time of service creation.
+If the manager is able to resolve the tag to a digest, worker nodes for that
+service will use that specific digest unless the service is explicitly updated.
```
nishanttotla: Let's just add a line here that this is particularly important for services that use mutable tags like `latest`, so that all service tasks use the same image.

mstanleyjones:
OK, fixed the latest feedback!

nishanttotla:
LGTM

mstanleyjones:
Merged! 🎉

PR #540, opened 2016-11-11
Added GID to USER instruction

dhauenstein:

I added a GID parameter to the USER instruction. The USER instruction does not only allow setting the UID of a container, but also the GID of this user. GID may be added by adding a colon and a GID after the user name (such as: `USER www-data:www-data`. This feature is already implemented, but not documented.

### DIFF
--- a/engine/reference/builder.md
+++ b/engine/reference/builder.md
@@ -1204,8 +1204,8 @@ into the newly created volume.
 
 The `USER` instruction sets the user name or UID to use when running the image
 and for any `RUN`, `CMD` and `ENTRYPOINT` instructions that follow it in the
-`Dockerfile`. Optionaly a group name can be set by adding a `:` and a GID, which will be used as the GID 
-for `RUN`, `CMD` and `ENTRYPOINT` instructions.
+`Dockerfile`. To specify the group ID (GID), add it as a suffix to the UID, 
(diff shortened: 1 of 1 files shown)


johndmulhausen:
Deploy preview processing.

Built with commit 1cead3926fa9265bc725c3584fca9c1a0f6f0736

https://app.netlify.com/sites/docker-docs/deploys/582596a9a7e2d22c5ed91a3a

engine/reference/builder.md
```diff
@@ -1200,11 +1200,12 @@ into the newly created volume.
 
 ## USER
 
-    USER daemon
+    USER daemon[:daemon]
 
 The `USER` instruction sets the user name or UID to use when running the image
 and for any `RUN`, `CMD` and `ENTRYPOINT` instructions that follow it in the
-`Dockerfile`.
+`Dockerfile`. Optionaly a group name can be set by adding a `:` and a GID, which will be used as the GID
```
mstanleyjones: This is good info, but let's avoid passive voice:

To specify the group ID (GID), add it as a suffix to the UID, separated by a colon. For instance, `MYUSER:MYGROUP`.

This is just an idea, maybe you can do a better job of conveying it.

dhauenstein: Sounds good, I'll fix this in a minute. Cheers!

PR #547, opened 2016-11-11
remove jekyll-gist from _config.yml

aduermael:
### Proposed changes

I removed the jekyll-gist gem from  from _config.yml.
It's not being used anymore, and in my opinion we shouldn't use it in the future either. All content should come from that repository. 


johndmulhausen:
@aduermael Agreed; it was interesting to experiment with but ultimately local development/building is broken if we depend on Gist.

PR #550, opened 2016-11-11
Update compose example to work better with swarm

davidwilliamson:
Update the Wordpress example file `docker-compose.yml` to work better with swarm mode enabled, while still behaving correctly without swarm mode.

* Remove the `links` directive. This allows swarm to distribute containers across available nodes. (with the `links` directive, swarm will be forced to run both containers on the same node.) Note that the `depends_on` directive is sufficient to ensure correct container start up sequence.
* Use a docker volume instead of bind mounting a local directory.
* Give docker-compose commands for shutting down the Wordpress app, showing how to preserve or remove the persisted data in the volume.

Verified proposed change works correctly in both a swarm and a non-swarm environment.

N/A
### Related issue
None
None
### Please take a look
@shin- For correctness of the proposed changes to docker-compose.yml

### DIFF
--- a/compose/wordpress.md
+++ b/compose/wordpress.md
@@ -33,7 +33,7 @@ with Docker containers. This quick-start guide demonstrates how to use Compose t
        db:
          image: mysql:5.7
          volumes:
-           - "./.data/db:/var/lib/mysql"
+           - db_data:/var/lib/mysql
          restart: always
          environment:
            MYSQL_ROOT_PASSWORD: wordpress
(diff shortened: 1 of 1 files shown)


johndmulhausen:
Deploy preview processing.

Built with commit 186dfbb3369d19a95d7a5cd3680c98c0379480a1

https://app.netlify.com/sites/docker-docs/deploys/58264655c4d9cc7e488614bf

johndmulhausen:
Thanks @davidwilliamson this is great!

Issue #529, opened 2016-11-10
Document managing sensitive strings (secrets)


mstanleyjones:
### Problem description

Document feature described in:
- https://github.com/docker/docker/pull/27794
- https://github.com/docker/docker/pull/28306

### Project version(s) affected

Engine 1.13

### Suggestions for a fix

cc/ @ehazlett

mstanleyjones:
Closed via #568

PR #568, opened 2016-11-15
Document secrets

mstanleyjones:

Add docs for using secrets, addresses #529.

Engine 1.13

### Related issue

#529 

see #529

### Please take a look

@ehazlett @thaJeztah @mrjana @stevvooe and please add anyone else who should review.

### DIFF
--- a/_data/toc.yaml
+++ b/_data/toc.yaml
@@ -287,6 +287,8 @@ toc:
       title: Manage nodes in a swarm
     - path: /engine/swarm/services/
       title: Deploy services to a swarm
+    - path: /engine/swarm/secrets/
+      title: Manage sensitive data with Docker secrets
     - path: /engine/swarm/networking/
       title: Attach services to an overlay network
     - path: /engine/swarm/admin_guide/
--- a/engine/swarm/secrets.md
+++ b/engine/swarm/secrets.md
@@ -105,10 +105,10 @@ real-world example, continue to
 [Intermediate example: Use secrets with a Nginx service](#intermediate-example-use-secrets-with-a-nginx-service).
 
 1.  Add a secret to Docker. The `docker secret create` command reads standard
-    input.
+    input because the `-f` flag is set to `-`.
 
     ```bash
-    $ echo "This is a secret" | docker secret create my_secret_data
+    $ echo "This is a secret" | docker secret create my_secret_data -f -
     ```
 
 2.  Create a `redis` service and grant it access to the secret. By default,
@@ -337,14 +337,17 @@ generate the site key and certificate, name the files `site.key` and
 2.  Create three secrets, representing the key, the certificate, and the
     `site.conf`. You can store any file as a secret as long as it is smaller
     than 500 KB. This allows you to decouple the key, certificate, and
-    configuration from the services that will use them.
+    configuration from the services that will use them. In each of these
+    commands, the `-f` flag takes the path to the file containing the secret
+    on the host machine's filesystem. In these examples, the secret name and
+    the file name are the same.
 
     ```bash
-    $ cat site.key | docker secret create site.key
+    $ docker secret create site.key -f site.key
 
-    $ cat site.crt | docker secret create site.crt
+    $ docker secret create site.crt -f site.crt
 
-    $ cat site.conf | docker create site.conf
+    $ docker secret create site.conf -f site.conf
     ```
 
     ```bash
@@ -521,8 +524,11 @@ line.
     > can later add a new version, update the service to use it, then remove the
     > old version.
 
+    The `-f` flag is set to `-`, which indicates that the input is read from
+    STDIN.
+
     ```bash
-    $ openssl rand -base64 20 | docker secret create mysql_password
+    $ openssl rand -base64 20 | docker secret create mysql_password -f -
 
     l1vinzevzhj4goakjap5ya409
     ```
@@ -535,7 +541,7 @@ line.
     bootstrap the `mysql` service.
 
     ```bash
-    $ openssl rand -base64 20 | docker secret create mysql_root_password
+    $ openssl rand -base64 20 | docker secret create mysql_root_password -f -
     ```
 
     List the secrets managed by Docker using `docker secret ls`:
@@ -570,7 +576,9 @@ line.
     - The secrets are each mounted in a `tmpfs` filesystem at
       `/run/secrets/mysql_password` and `/run/secrets/mysql_root_password`.
       They are never exposed as environment variables, nor can they be committed
-      to an image if the `docker commit` command is run.
+      to an image if the `docker commit` command is run. The `mysql_password`
+      secret is the one used the non-privileged WordPress container will use to
+      connect to MySQL.
     - Sets the environment variables `MYSQL_PASSWORD_FILE` and
       `MYSQL_ROOT_PASSWORD_FILE` to point to the
       files `/run/secrets/mysql_password` and `/run/secrets/mysql_root_password`.
@@ -704,13 +712,13 @@ use it, then remove the old secret.
 queries or commands, as opposed to just changing a single environment variable
 or a file, since the image only sets the MySQL password if the database doesn’t
 already exist, and MySQL stores the password within a MySQL database by default.
-Rotating passwords or other secrets will often involve additional steps outside
-of Docker.
+Rotating passwords or other secrets may involve additional steps outside of
+Docker.
 
 1.  Create the new password and store it as a  secret named `mysql_password_v2`.
 
     ```bash
-    $ openssl rand -base64 20 | docker secret create mysql_password_v2
+    $ openssl rand -base64 20 | docker secret create mysql_password_v2 -f -
     ```
 
 2.  Update the MySQL service to give it access to both the old and new secrets.
@@ -733,12 +741,14 @@ of Docker.
     `/run/secrets/mysql_password`.
 
     Even though the MySQL service has access to both the old and new secrets
-    now, the MySQL root password has not yet been changed.
+    now, the MySQL password for the WordPress user has not yet been changed.
 
-3.  Now, change the MySQL password for the `wordpress` user using the `mysql`
-    CLI. This command reads the old and new password from the files in
-    `/run/secrets` but does not expose them on the command line or save them in
-    the shell history.
+    > **Note**: This example does not rottate the MySQL `root` password.
+
+3.  Now, change the MySQL password for the `wordpress` user using the
+    `mysqladmin` CLI. This command reads the old and new password from the files
+    in `/run/secrets` but does not expose them on the command line or save them
+    in the shell history.
 
     Do this quickly and move on to the next step, because WordPress will lose
     the ability to connect to MySQL.
--- a/engine/swarm/secrets.md
+++ b/engine/swarm/secrets.md
@@ -1,0 +1,837 @@
+---
+title: Manage sensitive data with Docker secrets
+description: How to securely store, retrieve, and use sensitive data with Docker services
+keywords: swarm, secrets, credentials, sensitive strings, sensitive data, security, encryption, encryption at rest
+---
+
+## About secrets
+
+In terms of Docker Swarm services, a _secret_ is a blob of data, such as a
+password, SSH private key, SSL certificate, or another piece of data that should
+not be transmitted over a network or stored unencrypted in a Dockerfile or in
+your application's source code. In Docker 1.13 and higher, you can use Docker
+_secrets_ to centrally manage this data and securely transmit it to only those
+containers that need access to it. Secrets are encrypted during transit and at
+rest in a Docker swarm. A given secret is only accessible to those services
+which have been granted explicit access to it, and only while those service
+tasks are running.
+
+You can use secrets to manage any sensitive data which a container needs at
+runtime but you don't want to store in the image or in source control, such as:
+
+- Usernames and passwords
+- TLS certificates and keys
+- SSH keys
+- Other important data such as the name of a database or internal server
+- Generic strings or binary content (up to 500 kb in size)
+
+> **Note**: Docker secrets are only available to swarm services, not to
+> standalone containers. To use this feature, consider adapting your container to
+> run as a service with a scale of 1.
+
+Another use case for using secrets is to provide a layer of abstraction between
+the container and a set of credentials. Consider a scenario where you have
+separate development, test, and production environments for your application.
+Each of these environments can have different credentials, stored in the
+development, test, and production swarms with the same secret name. Your
+containers only need to know the name of the secret in order to function in all
+three environments.
+
+## How Docker manages secrets
+
+When you add a secret to the swarm, Docker sends the secret to the swarm manager
+over a mutual TLS connection. The secret is stored in the Raft log, which is
+encrypted. The entire Raft log is replicated across the other managers, ensuring
+the same high availability guarantees for secrets as for the rest of the swarm
+management data.
+
+>**Warning**: Raft data is encrypted in Docker 1.13 and higher. If any of your
+Swarm managers run an earlier version, and one of those managers becomes the
+manager of the swarm, the secrets will be stored unencrypted in that node's Raft
+logs. Before adding any secrets, update all of your manager nodes to Docker 1.13
+to prevent secrets from being written to plain-text Raft logs.
+
+When you grant a newly-created or running service access to a secret, the
+decrypted secret is mounted into the container in an in-memory filesystem at
+`/run/secrets/<secret_name>`. You can update a service to grant it access to
+additional secrets or revoke its access to a given secret at any time.
+
+A node only has access to (encrypted) secrets if the node is a swarm manager or
+if it is running service tasks which have been granted access to the secret.
+When a container task stops running, the decrypted secrets shared to it are
+unmounted from the in-memory filesystem for that container and flushed from the
+node's memory.
+
+If a node loses connectivity to the swarm while it is running a task container
+with access to a secret, the task container still has access to its secrets, but
+cannot receive updates until the node reconnects to the swarm.
+
+You can add or inspect an individual secret at any time, or list all
+secrets. You cannot remove a secret that a running service is
+using. See [Rotate a secret](secrets.md#example-rotate-a-secret) for a way to
+remove a secret without disrupting running services.
+
+In order to update or roll back secrets more easily, consider adding a version
+number or date to the secret name. This is made easier by the ability to control
+the mount point of the secret within a given container.
+
+## Read more about `docker secret` commands
+
+Use these links to read about specific commands, or continue to the
+[example about using secrets with a service](secrets.md#example-use-secrets-with-a-service).
+
+- [`docker secret create`](../reference/commandline/secret_create.md)
+- [`docker secret inspect`](../reference/commandline/secret_inspect.md)
+- [`docker service ls`](../reference/commandline/secret_ls.md)
+- [`docker secret rm`](../reference/commandline/secret_rm.md)
+- [`--secret`](../reference/commandline/service_create.md#create-a-service-with-secrets) flag for `docker service create`
+- [`--secret-add` and `--secret-rm`](../reference/commandline/service_update.md#adding-and-removing-secrets) flags for `docker service update`
+
+## Examples
+
+This section includes three graduated examples which illustrate how to use
+Docker secrets. The images used in these examples have been updated to make it
+easier to use Docker secrets. To find out how to modify your own images in
+a similar way, see
+[Build support for Docker Secrets into your images](#build-support-for-docker-secrets-into-your-images).
+
+> **Note**: These examples use a single-Engine swarm and unscaled services for
+> simplicity.
+
+### Simple example: Get started with secrets
+
+This simple example shows how secrets work in just a few commands. For a
+real-world example, continue to
+[Intermediate example: Use secrets with a Nginx service](#intermediate-example-use-secrets-with-a-nginx-service).
+
+1.  Add a secret to Docker. The `docker secret create` command reads standard
+    input.
+
+    ```bash
+    $ echo "This is a secret" | docker secret create my_secret_data
+    ```
+
+2.  Create a `redis` service and grant it access to the secret. By default,
+    the container can access the secret at `/run/secrets/<secret_name>`, but
+    you can customize the file name on the container using the `target` option.
+
+    ```bash
+    $ docker service  create --name="redis" --secret="my_secret_data" redis:alpine
+    ```
+
+3.  Verify that the task is running without issues using `docker service ps`. If
+    everything is working, the output looks similar to this:
+
+    ```bash
+    $ docker service ps redis
+
+    ID            NAME     IMAGE         NODE              DESIRED STATE  CURRENT STATE          ERROR  PORTS
+    bkna6bpn8r1a  redis.1  redis:alpine  ip-172-31-46-109  Running        Running 8 seconds ago  
+    ```
+
+    If there were an error, and the task were failing and repeatedly restarting,
+    you would see something like this:
+
+    ```bash
+    $ docker service ps redis
+
+    NAME                      IMAGE         NODE  DESIRED STATE  CURRENT STATE          ERROR                      PORTS
+    redis.1.siftice35gla      redis:alpine  moby  Running        Running 4 seconds ago                             
+     \_ redis.1.whum5b7gu13e  redis:alpine  moby  Shutdown       Failed 20 seconds ago      "task: non-zero exit (1)"  
+     \_ redis.1.2s6yorvd9zow  redis:alpine  moby  Shutdown       Failed 56 seconds ago      "task: non-zero exit (1)"  
+     \_ redis.1.ulfzrcyaf6pg  redis:alpine  moby  Shutdown       Failed about a minute ago  "task: non-zero exit (1)"  
+     \_ redis.1.wrny5v4xyps6  redis:alpine  moby  Shutdown       Failed 2 minutes ago       "task: non-zero exit (1)"
+    ```
+
+4.  Get the ID of the `redis` service task container using `docker ps` , so that
+    you can use `docker exec` to connect to the container and read the contents
+    of the secret data file, which defaults to being readable by all and has the
+    same name as the name of the secret. The first command below illustrates
+    how to find the container ID, and the second and third commands use shell
+    completion to do this automatically.
+
+    ```bash
+    $ docker ps --filter name=redis -q
+
+    5cb1c2348a59
+
+    $ docker exec $(docker ps --filter name=redis -q) ls -l /run/secrets
+
+    total 4
+    -r--r--r--    1 root     root            17 Dec 13 22:48 my_secret_data
+
+    $ docker exec $(docker ps --filter name=redis -q) cat /run/secrets/my_secret_data
+
+    This is a secret
+    ```
+
+5.  Verify that the secret is **not** available if you commit the container.
+
+    ```bash
+    $ docker commit $(docker ps --filter name=redis -q) committed_redis
+
+    $ docker run --rm -it committed_redis cat /run/secrets/my_secret_data
+
+    cat: can't open '/run/secrets/my_secret_data': No such file or directory
+    ```
+
+6.  Try removing the secret. The removal fails because the `redis` is running
+    and has access to the secret.
+
+    ```bash
+
+    $ docker secret ls
+
+    ID                          NAME                CREATED             UPDATED
+    wwwrxza8sxy025bas86593fqs   my_secret_data      4 hours ago         4 hours ago
+
+
+    $ docker secret rm my_secret_data
+
+    Error response from daemon: rpc error: code = 3 desc = secret 'my_secret_data' is in use by the following service: redis
+    ```
+
+7.  Remove access to the secret from the running `redis` service by updating the
+    service.
+
+    ```bash
+    $ docker service update --secret-rm="my_secret_data" redis
+    ```
+
+8.  Repeat steps 3 and 4 again, verifying that the service no longer has access
+    to the secret. The container ID will be different, because the
+    `service update` command redeploys the service.
+
+    ```bash
+    $ docker exec -it $(docker ps --filter name=redis -q) cat /run/secrets/my_secret_data
+
+    cat: can't open '/run/secrets/my_secret_data': No such file or directory
+    ```
+
+7.  Stop and remove the service, and remove the secret from Docker.
+
+    ```bash
+    $ docker service rm redis
+
+    $ docker secret rm my_secret_data
+    ```
+
+### Intermediate example: Use secrets with a Nginx service
+
+This example is divided into two parts.
+[The first part](#generate-the-site-certificate) is all about generating
+the site certificate and does not directly involve Docker secrets at all, but
+it sets up [the second part](#configure-the-nginx-container), where you store
+and use the site certificate and Nginx configuration as secrets.
+
+#### Generate the site certificate
+
+Generate a root CA and TLS certificate and key for your site. For production
+sites, you may want to use a service such as `Let’s Encrypt` to generate the
+TLS certificate and key, but this example uses command-line tools. This step
+is a little complicated, but is only a set-up step so that you have
+something to store as a Docker secret. If you want to skip these sub-steps,
+you can [use Let's Encrypt](https://letsencrypt.org/getting-started/) to
+generate the site key and certificate, name the files `site.key` and
+`site.crt`, and skip to
+[Configure the Nginx container](#configure-the-nginx-container).
+
+1.  Generate a root key.
+
+    ```bash
+    $ openssl genrsa -out "root-ca.key" 4096
+    ```
+
+2.  Generate a CSR using the root key.
+
+    ```bash
+    $ openssl req \
+              -new -key "root-ca.key" \
+              -out "root-ca.csr" -sha256 \
+              -subj '/C=US/ST=CA/L=San Francisco/O=Docker/CN=Swarm Secret Example CA'
+    ```
+
+3.  Configure the root CA. Edit a new file called `root-ca.cnf` and paste
+    the following contents into it. This constrains the root CA to only be
+    able to sign leaf certificates and not intermediate CAs.
+
+    ```none
+    [root_ca]
+    basicConstraints = critical,CA:TRUE,pathlen:1
+    keyUsage = critical, nonRepudiation, cRLSign, keyCertSign
+    subjectKeyIdentifier=hash
+    ```
+
+4.  Sign the certificate.
+
+    ```bash
+    $ openssl x509 -req  -days 3650  -in "root-ca.csr" \
+                   -signkey "root-ca.key" -sha256 -out "root-ca.crt" \
+                   -extfile "root-ca.cnf" -extensions \
+                   root_ca
+    ```
+
+5.  Generate the site key.
+
+    ```bash
+    $ openssl genrsa -out "site.key" 4096
+    ```
+
+6.  Generate the site certificate and sign it with the site key.
+
+    ```bash
+    $ openssl req -new -key "site.key" -out "site.csr" -sha256 \
+              -subj '/C=US/ST=CA/L=San Francisco/O=Docker/CN=localhost'
+    ```
+
+7.  Configure the site certificate. Edit a new file  called `site.cnf` and
+    paste the following contents into it. This constrains the site
+    certificate so that it can only be used to authenticate a server and
+    can't be used to sign certificates.
+
+    ```none
+    [server]
+    authorityKeyIdentifier=keyid,issuer
+    basicConstraints = critical,CA:FALSE
+    extendedKeyUsage=serverAuth
+    keyUsage = critical, digitalSignature, keyEncipherment
+    subjectAltName = DNS:localhost, IP:127.0.0.1
+    subjectKeyIdentifier=hash
+    ```
+
+8.  Sign the site certificate.
+
+    ```bash
+    $ openssl x509 -req -days 750 -in "site.csr" -sha256 \
+        -CA "root-ca.crt" -CAkey "root-ca.key"  -CAcreateserial \
+        -out "site.crt" -extfile "site.cnf" -extensions server
+    ```
+
+9.  The `site.csr` and `site.cnf` files are not needed by the Nginx service, but
+    you will need them if you want to generate a new site certificate. Protect
+    the `root-ca.key` file.
+
+#### Configure the Nginx container
+
+1.  Produce a very basic Nginx configuration that serves static files over HTTPS.
+    The TLS certificate and key will be stored as Docker secrets so that they
+    can be rotated easily.
+
+    In the current directory, create a new file called `site.conf` with the
+    following contents:
+
+    ```none
+    server {
+        listen                443 ssl;
+        server_name           localhost;
+        ssl_certificate       /run/secrets/site.crt;
+        ssl_certificate_key   /run/secrets/site.key;
+
+        location / {
+            root   /usr/share/nginx/html;
+            index  index.html index.htm;
+        }
+    }
+    ```
+
+2.  Create three secrets, representing the key, the certificate, and the
+    `site.conf`. You can store any file as a secret as long as it is smaller
+    than 500 KB. This allows you to decouple the key, certificate, and
+    configuration from the services that will use them.
+
+    ```bash
+    $ cat site.key | docker secret create site.key
+
+    $ cat site.crt | docker secret create site.crt
+
+    $ cat site.conf | docker create site.conf
+    ```
+
+    ```bash
+    $ docker secret ls
+
+    ID                          NAME                  CREATED             UPDATED
+    2hvoi9mnnaof7olr3z5g3g7fp   site.key       58 seconds ago      58 seconds ago
+    aya1dh363719pkiuoldpter4b   site.crt       24 seconds ago      24 seconds ago
+    zoa5df26f7vpcoz42qf2csth8   site.conf      11 seconds ago      11 seconds ago
+    ```
+
+4.  Create a service that runs Nginx and has access to the three secrets. The
+    last part of the `docker service create` command creates a symbolic link
+    from the location of the `site.conf` secret to `/etc/nginx.conf.d/`, where
+    Nginx looks for extra configuration files. This step happens before Nginx
+    actually starts, so you don't need to rebuild your image if you change the
+    Nginx configuration.
+
+    > **Note**: Normally you would create a Dockerfile which copies the `site.conf`
+    > into place, build the image, and run a container using your custom image.
+    > This example does not require a custom image. It puts the `site.conf`
+    > into place and runs the container all in one step.
+
+    ```bash
+    $ docker service create \
+         --name nginx \
+         --secret site.key \
+         --secret site.crt \
+         --secret site.conf \
+         --publish 3000:443 \
+         nginx:latest \
+         sh -c "ln -s /run/secrets/site.conf /etc/nginx/conf.d/site.conf && exec nginx -g 'daemon off;'"
+    ```
+
+    This uses the short syntax for the `--secret` flag, which creates files in
+    `/run/secrets/` with the same name as the secret. Within the running
+    containers, the following three files now exist:
+
+    - `/run/secrets/site.key`
+    - `/run/secrets/site.crt`
+    - `/run/secrets/site.conf`
+
+5.  Verify that the Nginx service is running.
+
+    ```bash
+    $ docker service ls
+
+    ID            NAME   MODE        REPLICAS  IMAGE
+    zeskcec62q24  nginx  replicated  1/1       nginx:latest
+
+    $ docker service ps nginx
+
+    NAME                  IMAGE         NODE  DESIRED STATE  CURRENT STATE          ERROR  PORTS
+    nginx.1.9ls3yo9ugcls  nginx:latest  moby  Running        Running 3 minutes ago
+    ```
+
+6.  Verify that the service is operational: you can reach the Nginx
+    server, and that the correct TLS certificate is being used.
+
+    ```bash
+    $ curl --cacert root-ca.crt https://0.0.0.0:3000
+
+    <!DOCTYPE html>
+    <html>
+    <head>
+    <title>Welcome to nginx!</title>
+    <style>
+        body {
+            width: 35em;
+            margin: 0 auto;
+            font-family: Tahoma, Verdana, Arial, sans-serif;
+        }
+    </style>
+    </head>
+    <body>
+    <h1>Welcome to nginx!</h1>
+    <p>If you see this page, the nginx web server is successfully installed and
+    working. Further configuration is required.</p>
+
+    <p>For online documentation and support please refer to
+    <a href="http://nginx.org/">nginx.org</a>.<br/>
+    Commercial support is available at
+    <a href="http://nginx.com/">nginx.com</a>.</p>
+
+    <p><em>Thank you for using nginx.</em></p>
+    </body>
+    </html>
+    ```
+
+    ```bash
+    $ openssl s_client -connect 0.0.0.0:3000 -CAfile root-ca.crt
+
+    CONNECTED(00000003)
+    depth=1 /C=US/ST=CA/L=San Francisco/O=Docker/CN=Swarm Secret Example CA
+    verify return:1
+    depth=0 /C=US/ST=CA/L=San Francisco/O=Docker/CN=localhost
+    verify return:1
+    ---
+    Certificate chain
+     0 s:/C=US/ST=CA/L=San Francisco/O=Docker/CN=localhost
+       i:/C=US/ST=CA/L=San Francisco/O=Docker/CN=Swarm Secret Example CA
+    ---
+    Server certificate
+    -----BEGIN CERTIFICATE-----
+    …
+    -----END CERTIFICATE-----
+    subject=/C=US/ST=CA/L=San Francisco/O=Docker/CN=localhost
+    issuer=/C=US/ST=CA/L=San Francisco/O=Docker/CN=Swarm Secret Example CA
+    ---
+    No client certificate CA names sent
+    ---
+    SSL handshake has read 1663 bytes and written 712 bytes
+    ---
+    New, TLSv1/SSLv3, Cipher is AES256-SHA
+    Server public key is 4096 bit
+    Secure Renegotiation IS supported
+    Compression: NONE
+    Expansion: NONE
+    SSL-Session:
+        Protocol  : TLSv1
+        Cipher    : AES256-SHA
+        Session-ID: A1A8BF35549C5715648A12FD7B7E3D861539316B03440187D9DA6C2E48822853
+        Session-ID-ctx:
+        Master-Key: F39D1B12274BA16D3A906F390A61438221E381952E9E1E05D3DD784F0135FB81353DA38C6D5C021CB926E844DFC49FC4
+        Key-Arg   : None
+        Start Time: 1481685096
+        Timeout   : 300 (sec)
+        Verify return code: 0 (ok)
+    ```
+
+7.  To clean up after running this example, remove the `nginx` service and the
+    stored secrets.
+
+    ```bash
+    $ docker service rm nginx
+
+    $ docker secret rm site.crt site.key nginx.conf
+    ```
+
+### Advanced example: Use secrets with a WordPress service
+
+In this example, you create a single-node MySQL service with a custom root
+password, add the credentials as secrets, and create a single-node WordPress
+service which uses these credentials to connect to MySQL. The
+[next example](#example-rotate-a-secret) builds on this one and shows you how to
+rotate the MySQL password and update the services so that the WordPress service
+can still connect to MySQL.
+
+This example illustrates some techniques to use Docker secrets to avoid saving
+sensitive credentials within your image or passing them directly on the command
+line.
+
+> **Note**: This example uses a single-Engine swarm for simplicity, and uses a
+> single-node MySQL service because a single MySQL server instance cannot be
+> scaled by simply using a replicated service, and setting up a MySQL cluster is
+> beyond the scope of this example.
+>
+> Also, changing a MySQL root passphrase isn’t as simple as changing
+> a file on disk. You must use a query or a `mysqladmin` command to change the
+> password in MySQL.
+
+1.  Generate a random alphanumeric password for MySQL and store it as a Docker
+    secret with the name `mysql_password` using the `docker secret create`
+    command. To make the password shorter or longer, adjust the last argument of
+    the `openssl` command. This is just one way to create a relatively random
+    password. You can use another command to generate the password if you
+    choose.
+
+    > **Note**: After you create a secret, you cannot update it. You can only
+    > remove and re-create it, and you cannot remove a secret that a service is
+    > using. However, you can grant or revoke a running service's access to
+    > secrets using `docker service update`. If you need the ability to update a
+    > secret, consider adding a version component to the secret name, so that you
+    > can later add a new version, update the service to use it, then remove the
+    > old version.
+
+    ```bash
+    $ openssl rand -base64 20 | docker secret create mysql_password
+
+    l1vinzevzhj4goakjap5ya409
+    ```
+
+    The value returned is not the password, but the ID of the secret. In the
+    remainder of this tutorial, the ID output is omitted.
+
+    Generate a second secret for the MySQL `root` user. This secret won't be
+    shared with the WordPress service created later. It's only needed to
+    bootstrap the `mysql` service.
+
+    ```bash
+    $ openssl rand -base64 20 | docker secret create mysql_root_password
+    ```
+
+    List the secrets managed by Docker using `docker secret ls`:
+
+    ```bash
+    $ docker secret ls
+
+    ID                          NAME                  CREATED             UPDATED
+    l1vinzevzhj4goakjap5ya409   mysql_password        41 seconds ago      41 seconds ago
+    yvsczlx9votfw3l0nz5rlidig   mysql_root_password   12 seconds ago      12 seconds ago
+    ```
+
+    The secrets are stored in the encrypted Raft logs for the swarm.
+
+2.  Create a user-defined overlay network which will be used for communication
+    between the MySQL and WordPress services. There is no need to expose the
+    MySQL service to any external host or container.
+
+    ```bash
+    $ docker network create -d overlay mysql_private
+    ```
+
+3.  Create the MySQL service. The MySQL service will have the following
+    characteristics:
+
+    - Because the scale is set to `1`, only a single MySQL task runs.
+      Load-balancing MySQL is left as an exercise to the reader and involves
+      more than just scaling the service.
+    - Only reachable by other containers on the `mysql_private` network.
+    - Uses the volume `mydata` to store the MySQL data, so that it persists
+      across restarts to the `mysql` service.
+    - The secrets are each mounted in a `tmpfs` filesystem at
+      `/run/secrets/mysql_password` and `/run/secrets/mysql_root_password`.
+      They are never exposed as environment variables, nor can they be committed
+      to an image if the `docker commit` command is run.
+    - Sets the environment variables `MYSQL_PASSWORD_FILE` and
+      `MYSQL_ROOT_PASSWORD_FILE` to point to the
+      files `/run/secrets/mysql_password` and `/run/secrets/mysql_root_password`.
+      The `mysql` image reads the password strings from those files when
+      initializing the system database for the first time. Afterward, the
+      passwords are stored in the MySQL system database itself.
+    - Sets environment variables `MYSQL_USER` and `MYSQL_DATABASE`. A new
+      database called `wordpress` is created when the container starts, and the
+      `wordpress` user will have full permissions for this database only. This
+      user will not be able to create or drop databases or change the MySQL
+      configuration.
+
+      ```bash
+      $ docker service create \
+           --name mysql \
+           --replicas 1 \
+           --network mysql_private \
+           --mount type=volume,source=mydata,destination=/var/lib/mysql \
+           --secret source=mysql_root_password,target=mysql_root_password \
+           --secret source=mysql_password,target=mysql_password \
+           -e MYSQL_ROOT_PASSWORD_FILE="/run/secrets/mysql_root_password" \
+           -e MYSQL_PASSWORD_FILE="/run/secrets/mysql_password" \
+           -e MYSQL_USER="wordpress" \
+           -e MYSQL_DATABASE="wordpress" \
+           mysql:latest
+      ```
+
+4.  Verify that the `mysql` container is running using the `docker service ls` command.
+
+    ```bash
+    $ docker service ls
+
+    ID            NAME   MODE        REPLICAS  IMAGE
+    wvnh0siktqr3  mysql  replicated  1/1       mysql:latest
+    ```
+
+    At this point, you could actually revoke the `mysql` service's access to the
+    `mysql_password` and `mysql_root_password` secrets because the passwords
+    have been saved in the MySQL system database. Don't do that for now, because
+    we will use them later to facilitate rotating the MySQL password.
+
+5.  Now that MySQL is set up, create a WordPress service that connects to the
+    MySQL service. The WordPress service has the following characteristics:
+
+    - Because the scale is set to `1`, only a single WordPress task runs.
+      Load-balancing WordPress is left as an exercise to the reader, because of
+      limitations with storing WordPress session data on the container
+      filesystem.
+    - Exposes WordPress on port 30000 of the host machine, so that you can access
+      it from external hosts. You can expose port 80 instead if you do not have
+      a web server running on port 80 of the host machine.
+    - Connects to the `mysql_private` network so it can communicate with the
+      `mysql` container, and also publishes port 80 to port 30000 on all swarm
+      nodes.
+    - Has access to the `mysql_password` secret, but specifies a different
+      target file name within the container. The WordPress container will use
+      the mount point `/run/secrets/wp_db_password`. Also specifies that the
+      secret is not group-or-world-readable, by setting the mode to
+      `0400`.
+    - Sets the environment variable `WORDPRESS_DB_PASSWORD_FILE` to the file
+      path where the secret is mounted. The WordPress service will read the
+      MySQL password string from that file and add it to the `wp-config.php`
+      configuration file.
+    - Connects to the MySQL container using the username `wordpress` and the
+      password in `/run/secrets/wp_db_password` and creates the `wordpress`
+      database if it does not yet exist.
+    - Stores its data, such as themes and plugins, in a volume called `wpdata`
+      so these files  persist when the service restarts.
+
+    ```bash
+    $ docker service create \
+         --name wordpress \
+         --replicas 1 \
+         --network mysql_private \
+         --publish 30000:80 \
+         --mount type=volume,source=wpdata,destination=/var/www/html \
+         --secret source=mysql_password,target=wp_db_password,mode=0400 \
+         -e WORDPRESS_DB_USER="wordpress" \
+         -e WORDPRESS_DB_PASSWORD_FILE="/run/secrets/wp_db_password" \
+         -e WORDPRESS_DB_HOST="mysql:3306" \
+         -e WORDPRESS_DB_NAME="wordpress" \
+         wordpress:latest
+    ```
+
+6.  Verify the service is running using `docker service ls` and
+    `docker service ps` commands.
+
+    ```bash
+    $ docker service ls
+
+    ID            NAME       MODE        REPLICAS  IMAGE
+    wvnh0siktqr3  mysql      replicated  1/1       mysql:latest
+    nzt5xzae4n62  wordpress  replicated  1/1       wordpress:latest
+    ```
+
+    ```bash
+    $ docker service ps wordpress
+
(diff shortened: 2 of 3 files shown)


johndmulhausen:
Deploy preview failed.

Built with commit 1243ffc60f2903e76ba5164c0048693457212a0f

https://app.netlify.com/sites/docker-docs/deploys/582a634da7e2d21444fb3cf6

_data/toc.yaml
```diff
@@ -219,6 +219,8 @@ toc:
       title: Manage nodes in a swarm
     - path: /engine/swarm/services/
       title: Deploy services to a swarm
+    - path: /engine/swarm/store_sensitive_strings/
```
stevvooe: Why not `secrets`?

engine/swarm/store_sensitive_strings.md
```diff
@@ -0,0 +1,234 @@
...
+## How Docker manages secrets
+
+When you add a secret to the swarm, Docker sends the secret to each manager
+node over a network connection which uses mutual TLS. When a manager receives
+a secret, it encrypts it with an encryption key before writing it to
+the RAFT logs, just as other RAFT log data is encrypted. The encryption key
+itself has already been propagated to the swarm node using mutual TLS.
+
+>**Warning**: RAFT data is encrypted in Docker 1.13 and newer. If some of your
+Swarm nodes run an earlier version, the secrets are stored unecrypted in
```
stevvooe: > the secrets __may be__ stored unecrypted in those nodes' RAFT logs.

I'm not sure that we opaquely store secret data in older versions.

@diogomonica @cyli ?

engine/swarm/store_sensitive_strings.md
```diff
@@ -0,0 +1,234 @@
...
+those nodes' RAFT logs.
+
+When you create a service, you can give the service access to one or more
+secrets. You can update a service to grant it access to additional secrets or
+revoke its access to a given secret at any time.
+
+The Engine managing a given worker node creates a bind mount within
+`/run/secrets/` on the worker node for each secret. From the point of view of
```
stevvooe: @ehazlett `/run/secrets` runs afoul of FHS standards. Do we link to `/var/run` to `/run`?

From https://en.wikipedia.org/wiki/Filesystem_Hierarchy_Standard:

> In FHS 3.0, /var/run is replaced by /run; a system should either continue to provide a /var/run directory, or provide a symbolic link from /var/run to /run, for backwards compatibility.[10]

I think we just need to make sure that `/run` is not a fork of `/var/run`. Accessing `/var/run/secrets` or `/run/secrets` should be equivalent, recommending the later.

engine/swarm/store_sensitive_strings.md
diogomonica: If we're in a situation where there is a 1.12 manager in a 1.13 cluster, and someone adds a secret, the 1.12 manager will write the secrets unencrypted to disk.

The upgrade path should be: upgrade all the managers to 1.13, and then start using secrets.

engine/swarm/store_sensitive_strings.md
```diff
@@ -0,0 +1,234 @@
+---
+title: Manage sensitive strings (secrets) for Docker services
+description: How to securely store, retrieve, and use sensitive strings (secrets) in Docker services
```
diogomonica: the use of strings is weird. what about `sensitive data` everywhere?

engine/swarm/store_sensitive_strings.md
ehazlett: Yes /var/run is linked to /run

engine/swarm/store_sensitive_strings.md
```diff
@@ -0,0 +1,234 @@
...
+keywords: swarm, secrets, credentials, sensitive strings, security, encryption, encryption at rest
+---
+
+## About secrets
+
+In terms of Docker Swarm services, a _secret_ is a string, such as a password,
```
diogomonica: `string` -> `a blob of data`

engine/swarm/store_sensitive_strings.md
```diff
@@ -0,0 +1,234 @@
...
+credit card number, SSH private key, SSL certificate, or another piece of
```
diogomonica: Definitely not a credit card number. Lets remove that.

stevvooe:
A few nits, but, otherwise, a great doc!

engine/swarm/store_sensitive_strings.md
```diff
@@ -0,0 +1,234 @@
...
+sensitive data that should not be transmitted over a network or stored in a
+Dockerfile or in your application's source code in plain text. Starting in
```
diogomonica: starting with?

engine/swarm/store_sensitive_strings.md
```diff
@@ -0,0 +1,234 @@
...
+Docker 1.13, Docker can encrypt and manage these secrets for you. Consider the
+following examples when you might want to manage sensitive strings using Docker:
+
+- Your service relies on a password-protected database, and you want to protect
```
diogomonica: `and you want to securely distribute the credentials to your application containers`

engine/swarm/store_sensitive_strings.md
```diff
@@ -0,0 +1,234 @@
...
+  the database's credentials.
+- You need a central location to administer and manage credentials needed by
+  different services running in your swarm.
+- Your application relies upon a SSH private key, and you want to protect the
+  key.
+- You run an application, such as a web server, which needs to use a SSL
```
diogomonica: use TLS instead of SSL

engine/swarm/store_sensitive_strings.md
```diff
@@ -0,0 +1,234 @@
...
+  certificate, and you want to protect the certificate itself.
+- You use a different SSL certificate in development and production environments
```
diogomonica: TLS

diogomonica: Also, not sure exactly what you mean with this usecase.

engine/swarm/store_sensitive_strings.md
```diff
@@ -0,0 +1,234 @@
...
+  but do not want to change the image or store the SSL certificate in the image.
+- Your application uses infrastructure which is metered and needs a credential,
```
diogomonica: What about just listing the kinds of secrets instead of creating an example for each individual one?
```
Your application needs secrets like:
- API tokens
- SSH private keys
- TLS keys
...
```

engine/swarm/store_sensitive_strings.md
```diff
@@ -0,0 +1,234 @@
...
+  such as Amazon AWS, EC2, or S3, and you want to prevent unauthorized use of
+  this infrastructure.
+- You are a developer building an application which processes sensitive data
+  to which you do not need or want direct access. Your administrator may
+  populate the secret key with a mock value while you are developing the
+  application, and change it to use the real value in production.
+
+In these and many other situations, managing sensitive strings using Docker
```
diogomonica: sensitive data

engine/swarm/store_sensitive_strings.md
```diff
@@ -0,0 +1,234 @@
...
+secrets makes sense.
+
+## How Docker manages secrets
+
+When you add a secret to the swarm, Docker sends the secret to each manager
```
diogomonica: Maybe a nit, but we don't send the secret to each manager. We send the secret to a manager, and then the secret gets replicated securely across the remaining managers.

engine/swarm/store_sensitive_strings.md
```diff
@@ -0,0 +1,234 @@
...
+node over a network connection which uses mutual TLS. When a manager receives
```
diogomonica: What about: when a manager receives a secret, it stores it in RAFT, relying on the RAFT logs replication to ensure that your secrets are distributed with the remaining managers, and ensuring your secrets are highly available in the event of a manager failure.

In Docker 1.13 and newer, all the RAFT logs are encrypted using NaCl SecretBox (Salsa20/Poly1305)....

diogomonica: My point is:
- Talk about high availability of secrets due to using raft for replication
- Talk about MTLS for transport of the data
- Talk about how all raft logs are encrypted at rest using NaCl SecretBox (Salsa20/Poly1305).

engine/swarm/store_sensitive_strings.md
diogomonica: We should probably add something about how to securely upgrade from a 1.12, or a note that to use secrets securely, make sure that all of the manager nodes are running 1.13

engine/swarm/store_sensitive_strings.md
```diff
@@ -0,0 +1,234 @@
...
+
+When you create a service, you can give the service access to one or more
+secrets. You can update a service to grant it access to additional secrets or
+revoke its access to a given secret at any time.
+
+The Engine managing a given worker node creates a bind mount within
+`/run/secrets/` on the worker node for each secret. From the point of view of
+the container, the secret is available as plain text. The `/run/secrets/`
+directory is an in-memory filesystem. When the node stops or leaves the swarm,
+or when the service is stopped or removed, the secrets are no longer accessible.
```
diogomonica: add: and are completely removed from the node's memory.

engine/swarm/store_sensitive_strings.md
```diff
@@ -0,0 +1,234 @@
...
+You can add, remove, or get an individual secret at any time, or list all
+secrets. You can update a secret by removing it and creating a new secret with
+the same name. However, you cannot remove a secret that a running service is
+using. In order to update or roll back a secret, consider adding a version
+number or date to the secret name.
+
+## Read more about `docker secret` commands
+
+Use these links to read about specific commands, or continue to the
+[example about using secrets with a service](store_sensitive_strings.md@example-use-secrets-with-a-service).
```
diogomonica: I like `store sensitive data` instead of `sensitive strings`

engine/swarm/store_sensitive_strings.md
```diff
@@ -0,0 +1,234 @@
...
+use of Docker secrets. The following examples show variations on one approach.
+
+**Note**: These examples use syntax specific to shells like Bash. If you use a
+different shell, verify that shell's syntax for checking for the existence of
+a file.
+
+When evaluating the value of the environment variable,
+first check whether that value corresponds to a file path available within the
+container read the credentail from the file if so. The `mistysj/wordpress` image
+adds the following check to see if the value of `WORDPRESS_DB_PASSWORD` is a
```
diogomonica: We should probably document why we don't support environment variables in the first place. They are less secure than files.

See: https://github.com/docker/docker/pull/9176#issuecomment-99542089

engine/swarm/store_sensitive_strings.md
```diff
@@ -0,0 +1,234 @@
...
+[example about using secrets with a service](store_sensitive_strings.md@example-use-secrets-with-a-service).
+
+- [`docker secret create`](../reference/commandline/secret_create.md)
+- [`docker secret inspect`](../reference/commandline/secret_inspect.md)
+- [`docker service ls`](../reference/commandline/secret_ls.md)
+- [`docker secret rm`](../reference/commandline/secret_rm.md)
+- [`--secret`](../reference/commandline/service_create.md#create-a-service-with-secrets) flag for `docker service create`
+- [`--secret-add` and `--secret-rm`](../reference/commandline/service_update.md#adding-and-removing-secrets) flags for `docker service update`
+
+## Example: Use secrets with a service
```
diogomonica: We need an example for rotation of a secret (add a new secret, remove the old one, and ensure the new secret has the same target as the old one.

We should probably also give advanced examples on how to change the permissions of the secret file, and the use of `source=bla,target=bla,uid=1000,gid=1000,mode=0700`

engine/swarm/store_sensitive_strings.md
stevvooe: @diogomonica While RAFT messages may be written to disk opaquely, I'm not sure that a manager with 1.12 snapshot structure definition will actually do this. None of our types have opaque field preservation, as far as I can tell. Looks like this was removed in proto3, as described in https://github.com/google/protobuf/issues/272.

> The upgrade path should be: upgrade all the managers to 1.13, and then start using secrets.

I agree and this language is sufficient to protect users against leaks, even if not precisely accurate. We may want to make this stronger and warn users to make sure the _entire_ cluster is on 1.13 before performing cluster mutations.

The only data loss scenario @aaronlehmann and I could work out was if a 1.12 engine was elected leader in a mixed-mode cluster. We may be able to downweight older engines in the master election process to minimize this possibility.

The other scenarios for data loss considered are 1.12 proxying to 1.13 leader, but this is a non-issue, since the JSON API prevents opaque field transfer.

aaronlehmann: > The only data loss scenario @aaronlehmann and I could work out was if a 1.12 engine was elected leader in a mixed-version cluster.

Or if that engine is upgraded to 1.13 and later becomes the leader. It will be missing the data.

_data/toc.yaml
mstanleyjones: oh, I suppose.... I just don't think secrets means anything to anyone. :)

engine/swarm/store_sensitive_strings.md
mstanleyjones: Could you store sensitive data that isn't a string?

engine/swarm/store_sensitive_strings.md
mstanleyjones: You could have a different certificate but stored with the same secret name, and deploying to devel or production would be the same, even though the actual secret would be different.

engine/swarm/store_sensitive_strings.md
mstanleyjones: I really wanted to answer "why would I use this" as well as "What could I use this for"...

diogomonica: I get it. My feedback is that it doesn't seem to add much, and there is a shorter/sweeter way of describing potential use-cases, but I'm not known for writing great documentation :)

engine/swarm/store_sensitive_strings.md
diogomonica: Sounds good.

engine/swarm/store_sensitive_strings.md
diogomonica: String has a very specific meaning for developers and programmers. I think we want to avoid using that term. The term `blob` would technically be more adequate, but I think `sensitive data` reads a lot better.

_data/toc.yaml
diogomonica: Agreed. Let's use sensitive data.

thaJeztah:
Hm, for fun I tried adding a binary as secret, and that works :smile:

```bash
$ docker run --name hola hello-world

$ docker cp hola:/hello .
$ cat hello | docker secret create hello
r4gmhiw6fmiorqfk5reqbyryr

$ docker service create --name foo --secret source=hello,target=hello,mode=0755 nginx:alpine


$ docker exec $(docker ps -n1 -q --filter label=com.docker.swarm.service.name=foo) /run/secrets/hello

Hello from Docker!
This message shows that your installation appears to be working correctly.

To generate this message, Docker took the following steps:
 1. The Docker client contacted the Docker daemon.
 2. The Docker daemon pulled the "hello-world" image from the Docker Hub.
 3. The Docker daemon created a new container from that image which runs the
    executable that produces the output you are currently reading.
 4. The Docker daemon streamed that output to the Docker client, which sent it
    to your terminal.

To try something more ambitious, you can run an Ubuntu container with:
 $ docker run -it ubuntu bash

Share images, automate workflows, and more with a free Docker Hub account:
 https://hub.docker.com

For more examples and ideas, visit:
 https://docs.docker.com/engine/userguide/
```

Should we document what size limits there are to secrets? (I don't think we put a size limit on there, but given that secrets are in memory, using many / big data there could have consequences perhaps?)

ehazlett:
Yes there is a size limit of 500 KB (https://github.com/docker/swarmkit/blob/master/manager/controlapi/secret.go#L22) per secret.  We should document this.

_data/toc.yaml
cyli: Non-blocking: it looks like in the rest of the documentation there is a parenthetical "(secrets)" after "manage sensitive strings" - should that also be included here so folks can more easily find it in the TOC if they're specifically looking for the word "secrets"?

mstanleyjones:
@diogomonica @ehazlett @stevvooe @cyli @toli PTAL. I think the topic is starting to shape up nicely. I addressed the feedback so far, and added a section on rotating secrets. Still waiting on an update to the official `mysql` image so that I can use that in the example instead of `mysql/mysql-server`.

mstanleyjones:
cc/ @johndmulhausen @hansfrederickbrown

stevvooe:
LGTM

mstanleyjones:
Added info about binary content and 500kb limit.

engine/swarm/secrets.md
```diff
@@ -0,0 +1,278 @@
...
+Swarm managers run an earlier version, and one of those managers becomes the
+manager of the swarm, the secrets will be stored unencrypted in that node's RAFT
+logs. Before adding any secrets, update all of your manager nodes to Docker 1.13
+to prevent secrets from being written to plain-text RAFT logs.
+
+When you grant a newly-created or running service access to a secret, the
+decrypted secret is mounted into the container in an in-memory filesystem at
+`/run/secrets/<secret_name>`. You can update a service to grant it access to
+additional secrets or revoke its access to a given secret at any time. When the
+task container stops running, the secret is unavailable to that container. When
```
cyli: Non-blocking:  do you mean if you `docker start` the stopped container?

engine/swarm/secrets.md
```diff
@@ -0,0 +1,278 @@
...
+                            -e WORDPRESS_DB_HOST="mysql:3306" \
+                            wordpress:latest
+    ```
+
+6.  Access `http://localhost:30000/` from any Swarm node and begin using
+    WordPress.
+
+## Example: Rotate a secret
+
+This example builds upon the previous one. In this scenario, you create a new
```
cyli: "create a new secret with..."

engine/swarm/secrets.md
mstanleyjones: No, I was trying to cover the case where a node is running more than one task, each of which may have access to the same secret. When one of the containers stops, it doesn't have access to the secret anymore. If none of a node's task containers need access to a given secret anymore, I think the secret is flushed from the node as well, unless the node is acting as a backup master. Right? Is there a better way to put this?

mstanleyjones:
Addressed @cyli feedback, tried to reword the area about when secrets get removed from a node's memory. PTAL at that section again and give me suggestions how I can improve it.

engine/swarm/secrets.md
```diff
@@ -0,0 +1,280 @@
...
+
+In terms of Docker Swarm services, a _secret_ is a blob of data, such as a
+password, SSH private key, SSL certificate, or another piece of data that should
+not be transmitted over a network or stored in a Dockerfile or in your
+application's source code unencrypted. In Docker 1.13 and higher, you can use
+Docker _secrets_ to centrally manage this data and securely transmit it to only
+those containers that need access to it. A _secret_ is a blob of data that is
+encrypted during transit and at rest in a Docker swarm. A given secret is only
+accessible within those containers which have been granted explicit access to it,
+and only while those containers are running.
```
diogomonica: maybe we replace "accessible within those containers" to "accessible to those services"?

Other than that, beautiful.

engine/swarm/secrets.md
```diff
@@ -0,0 +1,280 @@
...
+
+You can use secrets to manage any sensitive data which a container needs at
+runtime but you don't want to store in the image or in source control, such as:
+
+- Usernames and passwords
+- Database names
```
diogomonica: database credentials

engine/swarm/secrets.md
```diff
@@ -0,0 +1,280 @@
...
+Each of these environments can have different credentials, stored in the
+development, test, and production swarms with the same secret name. Your
+containers only need to know the name of the secret in order to function in all
+three environments.
+
+## How Docker manages secrets
+
+When you add a secret to the swarm, Docker sends the secret to the swarm manager
+over a mutual TLS connection. The secret is stored in the RAFT log, which is
+encrypted. The entire RAFT log is replicated across the other managers, ensuring
```
diogomonica: We should probably link here to another document talking about `unlock` and `unlock-key`, and how the key used for encryption is stored and how it can be managed/rotated/etc.

engine/swarm/secrets.md
```diff
@@ -0,0 +1,280 @@
...
+6.  Access `http://localhost:30000/` from any Swarm node and begin using
+    WordPress.
+
+## Example: Rotate a secret
+
+This example builds upon the previous one. In this scenario, you create a new
+secret with the new MySQL password, update the `mysql` and `wordpress` services
+to use it, then remove the old secret.
+
+1.  Create the new secret, and name it `mysql_password_2016`.
```
diogomonica: let's use `mysq_password_v2` instead of 2016

engine/swarm/secrets.md
```diff
@@ -0,0 +1,280 @@
...
+5.  Remove the old secret from the swarm.
+
+    ```bash
+    $ docker secret rm mysql_password
+    ```
+
+## Build support for Docker Secrets into your images
+
+If you develop a container that can be deployed as a service and accepts a
+sensitive string, such as a credential, as an environment variable, consider
```
diogomonica: `and requires sensitive data, such as...`

engine/swarm/secrets.md
```diff
@@ -0,0 +1,280 @@
...
+
+The WordPress image has been updated so that the environment variables which
+contain important data for WordPress, such as `WORDPRESS_DB_PASSWORD`, also have
+variants which can read their values from a file (`WORDPRESS_DB_PASSWORD_FILE`).
+This strategy ensures that backward compatibility is preserved, while allowing
+your container to read the information from a Docker-managed secret instead of
+being passed directly.
+
+>**Note**: Docker secrets do not set environment variables directly. This was
+a conscious decision, since environment variables can leak between containers
```
diogomonica: I would do `since it's easy for environment variables to be unintentionally leaked`

engine/swarm/secrets.md
mstanleyjones: I really did mean the database name, separately from the credentials (username and password). For instance, you can specify that as a file with the new `wordpress` image enhancements.

engine/swarm/secrets.md
cyli: I think your new change is great - I was wondering if you were specifically trying to call out doing a commit in a running container so it'd be available later, or restarting a container, or otherwise checking the contents of the stopped container on disk, etc.

cyli:
@mstanleyjones This looks great, thank you!  I just wasn't sure if you were trying to emphasize something in particular, and this removes that confusion.  Appreciate all your hard work on this!  Other than @diogomonica's suggestion for lock and unlock, this all LGTM!

engine/swarm/secrets.md
```diff
@@ -0,0 +1,280 @@
...
+remove a secret without disrupting running services.
+
+In order to update or roll back secrets more easily, consider adding a version
+number or date to the secret name. This is made easier by the ability to control
+the mount point of the secret within a given container.
+
+## Read more about `docker secret` commands
+
+Use these links to read about specific commands, or continue to the
+[example about using secrets with a service](secrets.md.md@example-use-secrets-with-a-service).
```
ehazlett: Should this be `secrets.md@example-use-secrets-with-a-service`?

ehazlett:
one minor comment otherwise LGTM -- nice work!

engine/swarm/secrets.md
thaJeztah: probably `@` -> `#` as well?

engine/swarm/secrets.md
mstanleyjones: Done

engine/swarm/secrets.md
mstanleyjones: Reworded a bit.

engine/swarm/secrets.md
mstanleyjones: done

engine/swarm/secrets.md
mstanleyjones: Done

engine/swarm/secrets.md
```diff
@@ -0,0 +1,340 @@
...
+    $ openssl rand -base64 20 |docker secret create mysql_password
+      lwxgks2rojdbvaui5kay1suvr
+    ```
+
+    The value returned is not the password, but the ID (digest) of the secret.
+    List the secrets managed by Docker using `docker secret ls`:
+
+    ```bash
+    $ docker secret ls
+    ID                          NAME                CREATED             UPDATED             SIZE
```
cyli: Also, many apologies for springing this on you at the last minute, but https://github.com/docker/docker/pull/28727#issuecomment-262396148 removes displaying sizes and digests at all when you inspect.  Could you delete the last column in this sample output?  (the size column?)

mstanleyjones:
Addressed latest feedback. Thanks @cyli !

mstanleyjones:
Modified the instructions in the `mysql` lab now that https://github.com/docker-library/mysql/pull/237 is merged and the `mysql` image has been updated!

engine/swarm/secrets.md
mstanleyjones: Waiting for https://github.com/docker/docker.github.io/pull/694 to be merged.

engine/swarm/secrets.md
```diff
@@ -0,0 +1,338 @@
...
+
+    ```bash
+    $ docker secret ls
+    ID                          NAME                CREATED             UPDATED
+    lwxgks2rojdbvaui5kay1suvr   mysql_password      41 seconds ago      41 seconds ago
+    ```
+    The secret is stored in the encrypted RAFT logs for the swarm.
+
+4.  Create the MySQL service. The `mysql` image now supports
+    additional environemnt variables which read their values from a file within
```
cyli: "environment" mispelling :)

cyli: Also, non-blocking nitpick - this reads to me like the environment variable reads values from a file?  Whereas it's the script within the image which reads the file pointed at from the environment variable.  I'm not sure the best way to say this in a clear way, but maybe something like "additional environment variables, which lets mysql read secret values from a file..."

That may make the next sentence a bit redundant though.

cyli:
@mstanleyjones Thank you!  I have a minor spelling/phrasing nitpick, but everything looks great!  👍

mstanleyjones:
Addressed this new feedback. Thanks @cyli !

engine/swarm/secrets.md
```diff
@@ -0,0 +1,339 @@
...
+
+This simple example shows how secrets work in just a few commands. For a
+real-world example, continue to
+[Use secrets with a service](secrets.md#example-use-secrets-with-a-service).
+
+1.  Add a secret to Docker. The `docker secret create` command reads standard
+    input.
+
+    ```bash
+    $ echo "This is a secret" | docker secret create secret_data.txt
```
thaJeztah: Wonder if we should use `secret_data.txt` as a name here, this may be confusing because either users think this is the name of a local file ("where does the `secret_data.txt` file come from?"), or that it must match the filename under which it can be accessed inside the container (which is later shown that it can be set through `target`).

Perhaps something that more obviously shows it's a custom name, such as `my-little-secret` (without file-extension)

engine/swarm/secrets.md
```diff
@@ -0,0 +1,339 @@
...
+    ```
+
+2.  Create a `redis` service and grant it access to the secret.
+    ```bash
+    $ docker service  create --name="redis" --secret="secret_data.txt" redis:alpine
+    ```
+
```
thaJeztah: In this step, perhaps we should explain that (by default), the secret is accessible inside the container as a file named `/run/secrets/<name-of-your-secret>`

engine/swarm/secrets.md
```diff
@@ -0,0 +1,339 @@
...
+3.  Get the ID of the `redis` service task container using `docker ps`. The ID
+    is the first field.
+
+    ```bash
+    $ docker ps |grep redis
+
+    ea3839998cbf        redis@sha256:f1ed3708f538b537eb9c2a7dd50dc90a706f7debd7e1196c9264edeea521a86d                "docker-entrypoint..."   About a minute ago   Up About a minute   6379/tcp              redis.1.rn9sbes7lxsiaizfqzo7g4nvu
```
thaJeztah: This output will be updated before GA to match the 1.12 output again, and the digest will no longer be shown by default, just `redis:latest`

thaJeztah: (We should make this easier to do (`docker ps | grep` should not be the way, because `docker ps` does not relate to _services_))

engine/swarm/secrets.md
```diff
@@ -0,0 +1,339 @@
...
+    $ docker service rm redis
+
+    $ docker sercret rm secret_data.txt
+    ```
+
+## Example: Use secrets with a service
+
+In this example, you create a MySQL service with a custom username and password.
+Next, you add these credentials as secrets. Next, you create a WordPress
+service which uses these credentials to connect to MySQL. Finally, you update
```
thaJeztah: Perhaps avoid the repeated "Next", something like;

```
In this example, you create a MySQL service with a custom username and password,
and add these credentials as secrets. Next, you create a WordPress
service which uses these credentials to connect to MySQL. ....
```

engine/swarm/secrets.md
```diff
@@ -0,0 +1,339 @@
...
+    >**Note**: After you create a secret, you cannot update it. You can only
+    remove and re-create it, and you cannot remove a secret that a service is
+    using. However, you can grant or revoke a running service's access to
+    secrets using `service update`. If you need the ability to update a secret,
+    consider adding a version component to the secret name, so that you can later
+    add a new version, update the service to use it, then remove the old version.
+
+    ```bash
+    $ openssl rand -base64 20 |docker secret create mysql_password
+      lwxgks2rojdbvaui5kay1suvr
```
thaJeztah: Can you add something to the description like "after the `secret create` command completes, the ID of the secret is printed"

Looking at this example, I can mistake `lwxgks2rojdbvaui5kay1suvr` for being the actual secret

engine/swarm/secrets.md
```diff
@@ -0,0 +1,339 @@
...
+
+    ```bash
+    $ docker secret ls
+    ID                          NAME                CREATED             UPDATED
+    lwxgks2rojdbvaui5kay1suvr   mysql_password      41 seconds ago      41 seconds ago
+    ```
+    The secret is stored in the encrypted RAFT logs for the swarm.
+
+4.  Create the MySQL service. The `mysql` image now supports additional
+    environment variables which allows the container to read credentials and other
```
thaJeztah: I'd drop "now" and "additional" here

```Markdown
The `mysql` image supports  environment variables which allows
```

engine/swarm/secrets.md
```diff
@@ -0,0 +1,339 @@
...
+    important settings from files within the container. This example uses the new
```
thaJeztah: Remove "new" here

engine/swarm/secrets.md
```diff
@@ -0,0 +1,339 @@
...
+    `MYSQL_ROOT_PASSWORD_FILE` variable and sets it to the file path within the
+    container which contains the `mysql_password` secret.
```
thaJeztah: Should there be a link to the readme? "For an overview of all environment variables available for configuring the mysql image, refer to the image's documentation on Docker Hub https://hub.docker.com/_/mysql/

(or Docker Store; https://store.docker.com/images/3083290a-203f-4c04-b2de-cc057959d2c9?tab=description)

engine/swarm/secrets.md
```diff
@@ -0,0 +1,339 @@
...
+    $ docker service ls
+
+    ID            NAME   MODE        REPLICAS  IMAGE
+    6xw6rke8k1ul  mysql  replicated  1/1       mysql@sha256:406a4256f5092f88e0c424a0749d1258eb25f5db14eeed4d46532fbda3c63266
+    ```
+
+6.  Now that MySQL is set up, you can create a WordPress container that connects
+    to the MySQL server. The WordPress service has the following characteristics:
+
+    - Joins the service to the `wp_network` overlay network, and resolves to the
```
thaJeztah: (Same may apply to the MySQL service in point `4.`

engine/swarm/secrets.md
```diff
@@ -0,0 +1,339 @@
...
+3. Update the `wordpress` service to use the new password, keeping the target
+   path at `/run/secrets/wp_db_secret` and keeping the file permissions at `0400`.
+
+   ```bash
+   $ docker service update wordpress \
+                    --secret-add source=mysql_password_v2,target=wp_db_password,mode=0400 \
+                    --secret-rm mysql_password
+   ```
+
+4.  Verify that WordPress works by going to `http://localhost:30000/` again.
```
thaJeztah: Hm, okay, bad news; this example shows how to replace a secret, but unfortunately it doesn't work;

- both the `wordpress` and `mysql` services use an anonymous volume; each time these services are updated, a new volume is created, so any changes made are lost. After updating, I end up with a _fresh_ database, and a _fresh_ WordPress installation (i.e., uploads, and customizations are no longer there)
- the `mysql` image does not currently have an option to _update_ an existing password through environment variables. The provided secret/password is only used during first initialization, but ignored after that


To see that it doesn't work;

Create volumes in advance for both services (so that they keep modifications);

    docker volume create wp_data
    docker volume create mysql_data


Or implicitly create the volumes during `docker service create`;

    --mount source=mysql_data,target=/var/lib/mysql
    --mount source=wordpress_data,target=/var/www/html

```bash
docker service create --name="mysql" \
                        --mount source=mysql_data,target=/var/lib/mysql \
                        --network="wp_network" \
                        --replicas=1 \
                        --secret=mysql_password \
                        -e MYSQL_ROOT_PASSWORD_FILE=/run/secrets/mysql_password \
                        mysql

docker service create --name="wordpress" \
                        --mount source=wordpress_data,target=/var/www/html \
                        --network="wp_network" \
                        --replicas="1" \
                        --publish 30000:80 \
                        --secret source=mysql_password,target=wp_db_password,mode=0400 \
                        -e WORDPRESS_DB_PASSWORD_FILE="/run/secrets/wp_db_password" \
                        -e WORDPRESS_DB_HOST="mysql:3306" \
                        wordpress:latest

```bash

Now rotate the secrets;

```bash
docker service update mysql \
                 --secret-add source=mysql_password_v2,target=mysql_password \
                 --secret-rm mysql_password

docker service update wordpress \
                --secret-add source=mysql_password_v2,target=wp_db_password,mode=0400 \
                --secret-rm mysql_password
```


After this, the mysql service is running, but still uses the old password,
the WordPress service is not running, because it cannot connect to the
database

```
MySQL Connection Error: (1045) Access denied for user 'root'@'10.0.0.3' (using password: YES)
Warning: mysqli::mysqli(): (HY000/1045): Access denied for user 'root'@'10.0.0.3' (using password: YES) in - on line 19
```

engine/swarm/secrets.md
```diff
@@ -0,0 +1,339 @@
...
+    [Build support for Docker Secrets into your images](store_sensitive_strings.md#build-support-for-docker-secrets-into-your-images).
+
+    The MySQL service will have the following characteristics.
+
+    - Joins the service to the `wp_network` overlay network, and resolves to the
+      DNS name `mysql`.
+    - Because the scale is set to 1, only a single MySQL task runs.
+    - Grants the service access to the `mysql_password` secret, which will be
+      available as plain-text within the container at `/run/secrets/mysql_password`.
+    - Sets the environment variable `MYSQL_ROOT_PASSWORD_FILE` to this file path.
```
thaJeztah: I think we should follow best practice here and don't use `MYSQL_ROOT_PASSWORD` for the database connection, but specify a `MYSQL_DATABASE `, `MYSQL_USER`, and `MYSQL_PASSWORD` so that a non-database-root user is used for the WordPress connection / database

engine/swarm/secrets.md
mstanleyjones: Hmm, I need to test this again because it did work when I ran through my own example before...

thaJeztah: It's deceiving, because if you didn't actually create a site, you'll see the default "Welcome to WordPress" screen, but it's a fresh install

diogomonica: @cyli @thaJeztah ideas on how to fix this?

thaJeztah: Not directly, I think updating password needs to be done through the MySQL client / CLI (some care has to be taken as well, because usernames are not unique in MySQL, multiple users with the same name can exist 😞 http://dev.mysql.com/doc/refman/5.7/en/adding-users.html)

ping @tianon any thoughts on implementing something in the MySQL image that does not require doing it "the manual way"?

cyli: Agree it would have to be done manually.  I'm not sure it would make sense to have the image automatically change passwords - if database files already exist, it's sort of strange to have an image by default go in and modify it; just like it doesn't modify any existing schemas if the DB already exists - it just initializes the schema if the DB doesn't already exist.

Can we run a one-off service (with `--restart-condition=none`) that runs the same mysql container, with both secrets, and uses the CLI to use the existing passphrase to connect to the existing mysql container and set the passphrase to the new one?

tianon: Yeah, I definitely agree with @cyli that having the `mysql` image itself change existing database contents automatically would be strange, although it might be interesting to make a special-purpose image for that sort of thing if someone doesn't have one already.

diogomonica: I say we add a new simpler example (rotation of nginx TLS keys), and modify this mysql example to have a step where people `docker exec mysql change-password` into the running mysql on the host by using a new randomly generated password, and then add that new password as a new secret, and rotate the FE server.

This will be the way to get something like this done once exec works for swarm-mode, so we should probably just do the tutorial like this.

engine/swarm/secrets.md
```diff
@@ -0,0 +1,351 @@
...
+
+This simple example shows how secrets work in just a few commands. For a
+real-world example, continue to
+[Use secrets with a service](secrets.md#example-use-secrets-with-a-service).
+
+1.  Add a secret to Docker. The `docker secret create` command reads standard
+    input.
+
+    ```bash
+    $ echo "This is a secret" | docker secret create my_secret_data
```
cyli: I also wanted to mention the change that https://github.com/docker/docker/pull/28629 is making to this command (it'd be `docker secret create my_secret_data -f -` in this particular example that passes the data via STDIN; it can also take a file name with `docker secret create my_secret_data -f <somefile>`)

engine/swarm/secrets.md
```diff
@@ -0,0 +1,351 @@
...
+    ```bash
+    $ openssl rand -base64 20 |docker secret create mysql_password_v2
+      lwxgks2rojdbvaui5kay1suvr
+    ```
+
+2.  Update the `mysql` service to use the new password, keeping the target path
+    at `/run/secrets/mysql_password` within the container.
+
+    ```bash
+    $ docker service update mysql \
```
diogomonica: This will do a rolling restart of mysql, which isn't realistic for a database.

diogomonica: We might want to point that out, if we want to keep it.

diogomonica: I think we should add an example of certificate rotation for an nginx service. It's a lot easier/more straightforward example. Alternatively, we can show rotation of a redis access database. Running MySQL database as a service is a hard sell. We might simply show it as a normal docker run, and manual update of the passphrase using `docker exec`, and rotation of the passphrase on the wordpress using `service update`

cyli:
Also, https://github.com/docker/docker/pull/29061 updates `swarm leave` so that you have to provide `--force` in order to do so if the node is locked.  Not sure if that needs to be added to this doc (or if it should be updated in the `leave` docs)

engine/swarm/secrets.md
```diff
@@ -0,0 +1,351 @@
...
+manager of the swarm, the secrets will be stored unencrypted in that node's RAFT
+logs. Before adding any secrets, update all of your manager nodes to Docker 1.13
+to prevent secrets from being written to plain-text RAFT logs.
+
+When you grant a newly-created or running service access to a secret, the
+decrypted secret is mounted into the container in an in-memory filesystem at
+`/run/secrets/<secret_name>`. You can update a service to grant it access to
+additional secrets or revoke its access to a given secret at any time.
+
+A node only has access to (encrypted) secrets if it is a swarm manager or if it
```
cyli: Should this be "decrypted" secrets?  The manager has access to all the decrypted secrets, as well as the service.

engine/swarm/secrets.md
```diff
@@ -0,0 +1,351 @@
...
+
+## How Docker manages secrets
+
+When you add a secret to the swarm, Docker sends the secret to the swarm manager
+over a mutual TLS connection. The secret is stored in the RAFT log, which is
+encrypted. The entire RAFT log is replicated across the other managers, ensuring
+the same high availability guarantees for secrets as for the rest of the swarm
+management data.
+
+>**Warning**: RAFT data is encrypted in Docker 1.13 and higher. If any of your
```
cyli: Similarly to https://github.com/docker/docker.github.io/pull/694#discussion_r89906193, we should probably change "RAFT" to be "Raft" here and in the rest of the doc.

mstanleyjones:
Extensive reworking of the examples with a lot of help from @cyli. I'd also like @NathanMcCauley to have a look and make sure we are doing reasonable things here.

engine/swarm/secrets.md
```diff
@@ -0,0 +1,770 @@
...
+    ```
+
+3.  Verify that the task is running without issues using `docker service ps`. If
+    everything is working, the output looks similar to this:
+
+    ```bash
+    $ docker service ps --no-trunc redis
+
+    NAME                               lIMAGE         NODE              DESIRED STATE  CURRENT STATE           ERROR  PORTS
+    redis.1.cezsyv6xljjj6bzpajbc6agkb  redis:alpine   ip-172-31-46-109  Running        Running 26 seconds ago
```
thaJeztah: The output format of `docker service ps` changed back to the 1.12 format, so no `--no-trunc` is needed;

```bash
$ docker service ps web

ID            NAME   IMAGE         NODE          DESIRED STATE  CURRENT STATE          ERROR  PORTS
kht7jvzvai5z  web.1  nginx:alpine  371b354ecd17  Running        Running 3 seconds ago
```

engine/swarm/secrets.md
```diff
@@ -0,0 +1,770 @@
...
+    ```
+
+    If there were an error, and the task were failing and repeatedly restarting,
+    you would see something like this:
+
+    ```bash
+    $ docker service ps redis
```
thaJeztah: same here

engine/swarm/secrets.md
```diff
@@ -0,0 +1,770 @@
...
+
+    ```bash
+    $ docker commit b9fc6c983b36 committed_redis
+
+    sha256:dfde27e778c34d2c402d96bda227ecaa906b6bf1eb97ffd3bf00bb7485408153
+
+    $ docker run --rm -it committed_redis cat /run/secrets/my_secret_data
+
+    cat: can't open '/run/secrets/my_secret_data': No such file or directory
+    ```<!--'fix-->
```
thaJeztah: ^^ leftover comment `<!--'fix-->`, or intentionally?

engine/swarm/secrets.md
```diff
@@ -0,0 +1,770 @@
...
+    ```
+
+4.  Get the ID of the `redis` service task container using `docker ps` , so that
+    you can use `docker exec` to connect to the container and read the contents
+    of the secret data file, which defaults to being readable by all and has the
+    same name as the name of the secret.  The ID is the first field of the
+    output of `docker ps`.
+
+    ```bash
+    $ docker ps --filter name=redis
```
thaJeztah: Alternatively, use

```
docker inspect <task-id> --format '{{ .Status.ContainerStatus.ContainerID }}'
```

but that may be a bit awkward as well

engine/swarm/secrets.md
```diff
@@ -0,0 +1,770 @@
...
+
+    cat: can't open '/run/secrets/my_secret_data': No such file or directory
+    ```<!--'fix-->
+
+7.  Stop and remove the service, and remove the secret from Docker.
+
+    ```bash
+    $ docker service rm redis
+
+    $ docker secret rm my_secret_data.txt
```
thaJeztah: remove `.txt` here

engine/swarm/secrets.md
```diff
@@ -0,0 +1,770 @@
...
+    ```
+
+### Intermediate example: Use secrets with a Nginx service
+
+**Note**: This example uses a single-Engine swarm for simplicity, and only
```
thaJeztah: Should this be formatted as "note"?

```Markdown
> **Note**: ......
```

engine/swarm/secrets.md
```diff
@@ -0,0 +1,770 @@
...
+    This example does it all in one step.
+
+    ```bash
+    $ docker service create --name nginx \
+                            --secret site.key \
+                            --secret site.crt \
+                            --secret nginx.conf \
+                            --publish 3000:443 \
+                            nginx:latest \
+    sh -c "ln -s /run/secrets/nginx.conf /etc/nginx/conf.d/ nginx.conf && nginx -g 'daemon off;'"
```
thaJeztah: this should use `exec` so that nginx is PID 1 inside the container after the symlink was created;

```
sh -c "ln -s /run/secrets/nginx.conf /etc/nginx/conf.d/ nginx.conf && exec nginx -g 'daemon off;'"
```

engine/swarm/secrets.md
```diff
@@ -0,0 +1,770 @@
...
+    $ docker secret create -f nginx.conf nginx.conf
+
+    zoa5df26f7vpcoz42qf2csth8
+
+    $ docker secret ls
+
+    ID                          NAME                  CREATED             UPDATED
+    2hvoi9mnnaof7olr3z5g3g7fp   site.key       58 seconds ago      58 seconds ago
+    aya1dh363719pkiuoldpter4b   site.crt       24 seconds ago      24 seconds ago
+    zoa5df26f7vpcoz42qf2csth8   nginx.conf     11 seconds ago      11 seconds ago
```
thaJeztah: to make it easier to follow (especially with the symlink), you could consider naming the secret (e.g.) `site.conf`

engine/swarm/secrets.md
```diff
@@ -0,0 +1,770 @@
...
+    {: id="step3"}
+
+4.  Create a service that runs `nginx` and has access to the three secrets. the
+    last part of the `docker service create` command creates a symbolic link
+    from the location of the `nginx.conf` secret to `/etc/nginx.conf.d/`, where
+    Nginx looks for extra configuration files, before actually starting Nginx.
+    This ensures that you don't need to rebuild your image if you change the
+    Nginx configuration.
+
+    **Note**: Normally you would create a Dockerfile which copies the `nginx.conf`
```
thaJeztah: Should probably be formatted as `note` here as well (`> **Note**: ...`)

engine/swarm/secrets.md
mstanleyjones: It's intentional to fix the unbalanced `'` in `can't`, otherwise the rest of the file is screwed up in Atom! I'll move it closer to the source of the problem.

engine/swarm/secrets.md
mstanleyjones: That seems needlessly complicated, and it seems useful to teach the reader about the --filter thing..

mstanleyjones:
Addressed feedback so far. Will do another pass for formatting after Netlify finishes.

engine/swarm/secrets.md
```diff
@@ -0,0 +1,771 @@
...
+
+4.  Get the IP address of your host machine's main interface, likely `eth0` on
+    Linux hosts. You'll need it in the next step.
+
+5.  Now that MySQL is set up, create a WordPress service that connects to the
+    MySQL service. The WordPress service has the following characteristics:
+
+    - Because the scale is set to `1`, only a single WordPress task runs.
+      Load-balancing WordPress is left as an exercise to the reader, because of
+      limitations with storing WordPress session data on the container
```
thaJeztah: Should look into this again, maybe I was mistaken about this and it stores them in the database; let me know if you want me to find out

tianon: I think sessions are in the database, but uploads and other theme/content related bits are definitely stored in the local volume on disk, so sharing that storage somehow is going to definitely be required for load-balancing WordPress. :+1:

engine/swarm/secrets.md
```diff
@@ -0,0 +1,771 @@
...
+    - Sets the environment variable `MYSQL_ROOT_PASSWORD_FILE` to point to the
+      `mysql` password in the bind-mounted `tempdir` volume. The `mysql` image
+      reads the password string from that file when initializing the database
+      for the first time. Afterward, it is stored in the MySQL database itself.
+
+      ```bash
+      $ docker run --name mysql -d \
+                   -v $(pwd)/tempdir:/initpw \
+                   -e MYSQL_ROOT_PASSWORD_FILE=/initpw/mysql_password.txt \
+                   -p 3306:3306 \
```
cyli: I originally ran into errors attaching the container to the network, but looking it up some more it looks like I just needed to have an `--attachable` flag when creating the network?

cyli:
@thaJeztah I don't feel strongly about the container vs a service.  Regarding the gist you provided, you are right that `docker exec <mysql-container> bash -c 'mysqladmin --user=wordpress --password="$(< /run/secrets/mysql_password)" password "$(< /run/secrets/mysql_password_v2)"'` (wow your shell-fu is WAY better than mine!) doesn't seem to put anything bad anywhere.

Maybe we would also want to include a caveat about why it's ok in this case though, but that passing in secrets on the command line are generally a bad idea.  I just don't want users seeing this and thinking it's always appropriate.  For instance, in this case:

- we are not running an interactive shell, so it never goes into bash history, but even if the command were executed in an interactive shell, it would appear as `mysqladmin --user=root --password="$(< /run/secrets/mysql_password_new)" password "$(< /run/secrets/mysql_password)"` in the bash history
- `mysqladmin` masks command line params in `ps` and `/proc` - for instance, the above command is too short to capture in ps or proc, but if you run `docker exec <mysql-container> bash -c 'mysqladmin --user=wordpress --password="$(< /run/secrets/mysql_password)" password` and leave out the last password interpolation, I see the following:

    ```
    $ ps aux
    USER       PID %CPU %MEM    VSZ   RSS TTY      STAT START   TIME COMMAND
    mysql        1  0.1  9.7 1328056 198804 ?      Ssl  23:24   0:00 mysqld
    root     24626  0.0  0.1  25340  3264 ?        Ss   23:39   0:00 mysqladmin --user=root --password=x x password

    $ cat /proc/24626/cmdline
    mysqladmin--user=root--password=xxpassword
    ```
 
    Not all applications do this, and so if they passed in command line args it'd be visible from viewing processes or in /proc.

Also, in the revised example, should we mention that the service update to mysql causes the container to restart and thus would cause a disruption in the wordpress service?

cyli:
@thaJeztah Also, 👍  on the separate `wordpress` and `mysql` users!

engine/swarm/secrets.md
thaJeztah: If we do `docker run`, yes, but for a service not of course

thaJeztah:
> wow your shell-fu is WAY better than mine!

"Learn by example" - I looked into some of @tianon's scripts to see how it's done 👑 

> Also, in the revised example, should we mention that the service update to mysql causes the container to restart and thus would cause a disruption in the wordpress service?

Probably good to mention.

My goals were, first of all, to avoid having password in plain text on-disk. I know a lot of work has gone into the secrets feature to avoid just _that_, and ending up with writing to plain text files in the documentation didn't feel "right".

At first, I wanted to avoid the `docker exec` as a whole, and spin up a `mysql` container to use as client (which is the "neat" way to do it). Unfortunately, we don't have secrets yet on `docker run`, so that didn't work.

As a follow up to this PR, we should do some thinking about how to actually _use_ the secrets in various situations. For example, the WordPress image still sets the secrets through environment variables, so a stray `phpinfo()` page (yup expect those to be in a lot of setups) will happily show the password;

screen shot 2016-12-22 at 01 20 50

Preventing that needs changes to the configuration files of images, so we may want some examples for that.

cyli:
> My goals were, first of all, to avoid having password in plain text on-disk. I know a lot of work has gone into the secrets feature to avoid just that, and ending up with writing to plain text files in the documentation didn't feel "right".

Yep, don't disagree, although in theory it's on your local machine, so less terrible.  It can also be stored in a password manager such as the lastpass CLI, which would also let you pipe it into secrets, but wasn't sure if that'd be too much for an example.

> At first, I wanted to avoid the docker exec as a whole, and spin up a mysql container to use as client (which is the "neat" way to do it). Unfortunately, we don't have secrets yet on docker run, so that didn't work.

Can we use a service with `--restart-condition=none`?

```
docker service create \
  --name=mysql_password_rotation \
  --network=wpnet \
  --secret=mysql_password \
  --secret=mysql_password_v2 \
  --restart-condition=none \
  mysql \
  bash -c 'mysqladmin --host=mysql --user=wordpress --password="$(< /run/secrets/mysql_password)" password "$(< /run/secrets/mysql_password_v2)"'
```

I'm not sure if we're starting to get into weird service esoteric stuff, now, though :)

> As a follow up to this PR, we should do some thinking about how to actually use the secrets in various situations. For example, the WordPress image still sets the secrets through environment variables, so a stray phpinfo() page (yup expect those to be in a lot of setups) will happily show the password

Ah good catch.  :|  I didn't realize it'd be part of the process's environment - I'd only checked mysql's, which didn't seem to be.  cc @tianon

thaJeztah:
> I'm not sure if we're starting to get into weird service esoteric stuff, now, though :)

definitely taking it too far. we can revisit once (if) secrets are available for `docker run`

> Ah good catch. :| I didn't realize it'd be part of the process's environment - I'd only checked mysql's, which didn't seem to be.

Was already discussing with him; we could probably `unset` the env-vars directly after they have been consumed. Help me remind, I can try a pull request (or bake Tianon some cookies for christmas)

mstanleyjones:
Pushed some updates. This puts back the `mysql` service and also uses @thaJeztah magical way to update the password in MySQL. The only thing it doesn't do yet is use a non-root user for the database. Need to figure out how to do that.

engine/swarm/secrets.md
```diff
@@ -0,0 +1,812 @@
...
+
+      ```bash
+      $ docker service create \
+                       --name mysql \
+                       --replicas 1 \
+                       --network mysql_private \
+                       --mount type=volume,source=mydata,destination=/var/lib/mysql \
+                       --secret source=mysql_password,target=mysql_password \
+                       -e MYSQL_ROOT_PASSWORD_FILE="/run/secrets/mysql_password" \
+                      mysql:latest
```
thaJeztah: You need _two_ secrets here; one for the `root` password and one for the `wordpress` user;

```
docker service create \
  --name mysql \
  --replicas 1 \
  --network=mysql_private \
  --mount type=volume,source=mydata,destination=/var/lib/mysql \
  --secret=mysql_root_password \
  --secret=mysql_password \
  -e MYSQL_ROOT_PASSWORD_FILE="/run/secrets/mysql_root_password" \
  -e MYSQL_PASSWORD_FILE="/run/secrets/mysql_password" \
  -e MYSQL_DATABASE=wordpress \
  -e MYSQL_USER=wordpress \
  mysql
```

here;

- `MYSQL_ROOT_PASSWORD_FILE` sets the path to the secret for the `root` password
- `MYSQL_PASSWORD_FILE ` sets the path to the secret for the regular user's (`wordpress`) password
- `MYSQL_DATABASE` sets the name of the database to create the first time the service is started
- `MYSQL_USER` sets the name of the _user_ to create the first time the service is started

engine/swarm/secrets.md
```diff
@@ -0,0 +1,812 @@
...
+
+    ```bash
+    $ docker service create \
+                     --name wordpress \
+                     --replicas 1 \
+                     --network mysql_private \
+                     --publish 30000:80 \
+                     --secret source=mysql_password,target=wp_db_password,mode=0400 \
+                     -e WORDPRESS_DB_PASSWORD_FILE="/run/secrets/wp_db_password" \
+                     -e WORDPRESS_DB_HOST="mysql:3306" \
```
thaJeztah: You need to add `WORDPRESS_DB_USER` here, to set the name to use for connecting to the MySQL database;

```
-e WORDPRESS_DB_USER=wordpress \
```

engine/swarm/secrets.md
```diff
@@ -0,0 +1,812 @@
...
+      MySQL password string from that file and add it to the `wp-config.php`
+      configuration file.
+
+    ```bash
+    $ docker service create \
+                     --name wordpress \
+                     --replicas 1 \
+                     --network mysql_private \
+                     --publish 30000:80 \
+                     --secret source=mysql_password,target=wp_db_password,mode=0400 \
```
thaJeztah: Add a volume here, so that the WordPress data is preserved when updating the service;

```
--mount type=volume,source=wpdata,destination=/var/www/html \
```

engine/swarm/secrets.md
thaJeztah: Make sure to use the right secret for the right user.

engine/swarm/secrets.md
```diff
@@ -0,0 +1,812 @@
...
+
+7.  If you want to try the running all of these examples again or just want to
+    clean up after running through them, use these commands to remove the
+    WordPress service, the MySQL container, the Docker secrets, and the
+    `tempdir/` directory.
+
+    ```bash
+    $ docker service rm wordpress
+
+    $ docker rm -f mysql
```
thaJeztah: this will be `docker service rm mysql`, or combine with the previous one (`docker service rm wordpress mysql`)

also, the named volumes can be cleaned up afterwards

mstanleyjones:
OK, the latest update uses a separate `wordpress` database, user, and secret. I still need to set the the MySQL root password, but it's not used after MySQL starts up.

This is getting very close to being ready to go, IMHO. I guess the only outstanding concern is the WordPress environment variable, right @thaJeztah ?

engine/swarm/secrets.md
mstanleyjones: What WordPress data is that? I thought it stored everything in the MySQL database....

mstanleyjones:
BTW I don't need to use `WORDPRESS_DB_USER` because it defaults to `wordpress`.

thaJeztah:
> I guess the only outstanding concern is the WordPress environment variable

That's something to be addressed in the official WordPress image; I haven't had time to look into that yet, but it's orthogonal to this PR

> BTW I don't need to use WORDPRESS_DB_USER because it defaults to wordpress.

Oh, yes, it probably defaults to that; doesn't hurt to be explicit, but I'm also fine to skip setting it

engine/swarm/secrets.md
thaJeztah: The WordPress installation (theme, uploads, configuration, etc)

engine/swarm/secrets.md
```diff
@@ -0,0 +1,857 @@
...
+    > can later add a new version, update the service to use it, then remove the
+    > old version.
+
+    ```bash
+    $ openssl rand -base64 20 | docker secret create mysql_password
+
+    l1vinzevzhj4goakjap5ya409
+    ```
+
+    The value returned is not the password, but the ID (digest) of the secret.
```
thaJeztah: small nit; remove `(digest)` here; the ID of secrets have no relation to the content of the secrets (otherwise they may potentially "leak" the content).

engine/swarm/secrets.md
```diff
@@ -0,0 +1,857 @@
...
+    ```bash
+    $ docker network create -d overlay mysql_private
+    ```
+
+3.  Create the MySQL service. The MySQL service will have the following
+    characteristics:
+
+    - Because the scale is set to `1`, only a single MySQL task runs.
+      Load-balancing MySQL is left as an exercise to the reader and involves
+      than just scaling the service.
```
thaJeztah: nit: missing "more" before "than"

engine/swarm/secrets.md
```diff
@@ -0,0 +1,857 @@
...
+      They are never exposed as environment variables, nor can they be committed
+      to an image if the `docker commit` command is run.
+    - Sets the environment variables `MYSQL_PASSWORD_FILE` and
+      `MYSQL_ROOT_PASSWORD_FILE` to point to the
+      files `/run/secrets/mysql_password` and `/run/secrets/mysql_root_password`.
+      The `mysql` image reads the password strings from those files when
+      initializing the system database for the first time. Afterward, the
+      passwords are stored in the MySQL system database itself.
+    - Sets environment variables `MYSQL_USER` and `MYSQL_DATABASE`. A new
+      database called `wordpress` is created when the image starts, and the
```
thaJeztah: nit: "when the image starts" -> "when the container starts"

engine/swarm/secrets.md
```diff
@@ -0,0 +1,857 @@
...
+
+4.  Verify that the `mysql` container is running using the `docker service ls` command.
+
+    ```bash
+    $ docker service ls
+    ID            NAME   MODE        REPLICAS  IMAGE
+    wvnh0siktqr3  mysql  replicated  1/1       mysql:latest
+    ```
+
+    At this point, you could actually revoke the `mysql` service's access to the
```
thaJeztah: 👍

engine/swarm/secrets.md
```diff
@@ -0,0 +1,857 @@
...
+use it, then remove the old secret.
+
+**Note**: Changing the password on a MySQL database involves running extra
+queries or commands, as opposed to just changing a single environment variable
+or a file, since the image only sets the MySQL password if the database doesn’t
+already exist, and MySQL stores the password within a MySQL database by default.
+Rotating passwords or other secrets will often involve additional steps outside
+of Docker.
+
+1.  Create the new password and store it in file in the `tempdir` directory
```
thaJeztah: remove "and store it in file in the ..." part, as it's no longer needed

engine/swarm/secrets.md
```diff
@@ -0,0 +1,857 @@
...
+    z05lwief4u8wk6jc89xx2l4es
+    ```
+
+2.  Update the MySQL service to give it access to both the old and new secrets.
+    Remember that you cannot update or rename a secret, but you can revoke a
+    secret and grant access to it using a new target filename.
+
+    ```bash
+    $ docker service update \
+                     --secret-rm mysql_password mysql
```
thaJeztah: This is something we need to fix; doing `--secret-rm` and `--secret-add` should be possible in a single update

engine/swarm/secrets.md
```diff
@@ -0,0 +1,857 @@
...
+    `/run/secrets` but does not expose them on the command line or save them in
+    the shell history.
+
+    Do this quickly and move on to the next step, because WordPress will lose
+    the ability to connect to MySQL.
+
+    First, find the ID of the `mysql` container task.
+
+    ```bash
+    $ docker ps | grep mysql |awk {'print $1'}
```
thaJeztah: It's better to use the built-in filter functionality here (and `-q` to only show the `ID`;

```
docker ps -q --filter name=mysql
```

engine/swarm/secrets.md
```diff
@@ -0,0 +1,857 @@
...
+    ```
+
+5.  Verify that WordPress works by browsing to http://localhost:30000/ on any
+    swarm node again. You'll need to use the WordPress username and password
+    from when you ran through the WordPress wizard in the previous task.
+
+    Verify that the blog post you wrote still exists, and if you changed any
+    configuration values, verify that they are still changed.
+
+6.  Revoke accsss to the old secret from the MySQL service and
```
thaJeztah: typo `accsss`

mstanleyjones:
Fixed latest problems found by @thaJeztah and also got a review from my friend @bskaggs as well. He spotted things I could no longer see, I've been looking at this too long!

engine/swarm/secrets.md
```diff
@@ -0,0 +1,857 @@
...
+
+7.  If you want to try the running all of these examples again or just want to
+    clean up after running through them, use these commands to remove the
+    WordPress service, the MySQL container, the `mydata` volume, and the Docker
+    secrets.
+
+    ```bash
+    $ docker service rm wordpress mysql
+
+    $ docker volume rm mydata
```
thaJeztah: need to add `wpdata` here as well now

mstanleyjones:
@thaJeztah nit fixed, and a few more picky formatting tweaks. I think this is ready to merge but I'd like input from @cyli @diogomonica @NathanMcCauley

engine/swarm/secrets.md
```diff
@@ -0,0 +1,837 @@
...
+    ```bash
+    $ docker service update --secret-rm="my_secret_data" redis
+    ```
+
+8.  Repeat steps 3 and 4 again, verifying that the service no longer has access
+    to the secret. The container ID will be different, because the
+    `service update` command redeploys the service.
+
+    ```bash
+    $ docker exec -it <container_id> cat /run/secrets/my_secret_data
```
cyli: Non-blocking:  since we had the example above of `docker exec $(docker ps --filter name=redis -q) cat /run/secrets/my_secret_data`, can we use that same command here instead of `<container_id>`?

engine/swarm/secrets.md
```diff
@@ -0,0 +1,837 @@
...
+
+    ```bash
+    $ docker service rm redis
+
+    $ docker secret rm my_secret_data
+    ```
+
+### Intermediate example: Use secrets with a Nginx service
+
+> **Note**: This example uses a single-Engine swarm for simplicity, and only
```
cyli: Non-blocking:  Since all of these examples (redis, nginx, and wordpress) presume a single-Engine swarm, should that part of this note be moved to the top of the examples section?  (Similarly for the wordpress example)

engine/swarm/secrets.md
```diff
@@ -0,0 +1,837 @@
...
+4.  Create a service that runs Nginx and has access to the three secrets. The
+    last part of the `docker service create` command creates a symbolic link
+    from the location of the `site.conf` secret to `/etc/nginx.conf.d/`, where
+    Nginx looks for extra configuration files. This step happens before Nginx
+    actually starts, so you don't need to rebuild your image if you change the
+    Nginx configuration.
+
+    > **Note**: Normally you would create a Dockerfile which copies the `site.conf`
+    > into place, build the image, and run a container using your custom image.
+    > This example does it all in one step.
```
cyli: Maybe something like "This example does not require a custom image; it copies the `site.conf` into place and runs the container all in one step"?

engine/swarm/secrets.md
```diff
@@ -0,0 +1,837 @@
...
+
+    ```bash
+    $ docker secret ls
+
+    ID                          NAME                  CREATED             UPDATED
+    l1vinzevzhj4goakjap5ya409   mysql_password        41 seconds ago      41 seconds ago
+    yvsczlx9votfw3l0nz5rlidig   mysql_root_password   12 seconds ago      12 seconds ago
+    ```
+
+    The secrets are stored in the encrypted WAL logs for the swarm.
```
cyli: Non-blocking:  Should we just say "Raft logs" here, since above we mention that secrets are stored in the raft logs?  It's true that they are written to the WAL (and snapshot, later on), but that might be too much raft detail for this document, particularly since we never say what WAL stands for.

engine/swarm/secrets.md
```diff
@@ -0,0 +1,837 @@
...
+> **Note**: This example uses a single-Engine swarm for simplicity, and uses a
+> single-node MySQL service because a single MySQL server instance cannot be
+> scaled by simply using a replicated service, and setting up a MySQL cluster is
+> beyond the scope of this example.
+>
+> Also, changing a MySQL root passphrase isn’t as simple as changing
+> a file on disk. You must use a query or a `mysqladmin` command to change the
+> password in MySQL.
+
+1.  Generate a random alphanumeric password for MySQL and store it as a Docker
```
cyli: Should we mention that this first password is for the wordpress user for mysql?  Or at least a non-root user?

engine/swarm/secrets.md
```diff
@@ -0,0 +1,837 @@
...
+
+This example builds upon the previous one. In this scenario, you create a new
+secret with a new MySQL password, update the `mysql` and `wordpress` services to
+use it, then remove the old secret.
+
+**Note**: Changing the password on a MySQL database involves running extra
+queries or commands, as opposed to just changing a single environment variable
+or a file, since the image only sets the MySQL password if the database doesn’t
+already exist, and MySQL stores the password within a MySQL database by default.
+Rotating passwords or other secrets will often involve additional steps outside
```
cyli: Nonblocking nitpick: should this be "may"?  If it's something simple like an API key the service just access, just changing the service spec should be fine.  Alternately, this is likely always the case with databases, so maybe we can call out DBs specifically here?

engine/swarm/secrets.md
```diff
@@ -0,0 +1,837 @@
...
+         mysql
+    ```
+
+    Updating a service causes it to restart, and when the MySQL service restarts
+    the second time, it has access to the old secret under
+    `/run/secrets/old_mysql_password` and the new secret under
+    `/run/secrets/mysql_password`.
+
+    Even though the MySQL service has access to both the old and new secrets
+    now, the MySQL root password has not yet been changed.
```
cyli: We don't change the root password at all in this example, right?  Could we clarify this example to say that we don't need to adjust the service's access to the MySQL root password, since it's not being modified?

engine/swarm/secrets.md
```diff
@@ -0,0 +1,837 @@
...
+
+3.  Now, change the MySQL password for the `wordpress` user using the `mysql`
```
cyli: "`mysqladmin` CLI", as opposed to "`mysql` CLI"

mstanleyjones:
Addressed @cyli feedback, and also pushed a second commit which updates the syntax of the `-f` flag to `docker create`, as per https://github.com/docker/docker/pull/29889.

cyli:
LGTM!

PR #585, opened 2016-11-15
file for visual check of various formatting

sanscontext:
Hey @docker/docs-maintainers, this file will be a canary/smoketest that we can use as a quick visual check when we start culling excess JS and CSS. 

I've added a bunch of code format blocks at the end; if you know of one that I don't currently have, lmk or add it yourself by pushing to the patch.

### DIFF
--- a/test.md
+++ b/test.md
@@ -1,0 +1,357 @@
+---
+description: Smoketest page
+title: Testing page
+hide_from_sitemap: true
+---
+
+# Heading 1
+
+Plain block of text.
+
+Lorem ipsum dolor sit amet, consectetur adipisicing elit, sed do eiusmod tempor
+incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis
+nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat.
+Duis aute irure dolor in reprehenderit in voluptate velit esse cillum dolore eu
+fugiat nulla pariatur. Excepteur sint occaecat cupidatat non proident, sunt in
+culpa qui officia deserunt mollit anim id est laborum.
+
+## Heading 2
+
+Text with various styles, basic markdown formatting. You should **not** see a single line comment below this line.
+
+<!-- This is a comment. You should not see it rendered in the page. -->
+
+Once you create or link to a repository in Docker Cloud, you can set up [automated testing](/docker-cloud/builds/automated-testing.md) and [automated builds](/docker-cloud/builds/automated-build.md).
+
+![a small cute image](/images/footer_moby_icon.png)
+
+> **Note**: This is a note. Lorem ipsum dolor sit amet, consectetur adipisicing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam.
+
+![a pretty wide image](/images/banner_image_24512.png)
+
+<center>
+This line is centered with HTML.
+</center>
+
+This line is centered with curly-brace injection.
+{: style="text-align:center" }
+
+Some Lorem ipsum text with formatting and styling.
+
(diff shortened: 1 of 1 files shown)


sanscontext:
Also of note: I'm not sure where we want this to live? Is root okay?  I've omitted it from all the menuing, so that should be enough I'd hope, but maybe we should make a utility folder (for the `allpagelinks` and other test files).

johndmulhausen:
Deploy preview ready!

Built with commit b24357cb87b53f36c1b05cfd27a9cd042987719c

https://deploy-preview-585--docker-docs.netlify.com

test.md
```diff
@@ -0,0 +1,318 @@
+---
+description: Smoketest page
```
joaofnfernandes: Does it make sense to add this as a draft, instead of publishing it live?
https://jekyllrb.com/docs/drafts/

sanscontext: Not sure. Will that make it inaccessible on the live site?

johndmulhausen: Drafts are for "posts" -- as in, blog posts. We only use pages, which don't have draft functionality.

johndmulhausen: I think if we leave this out of the nav and exclude it from the sitemap by adding `hide_from_sitemap: true` to the front-matter, that's probably enough.

test.md
```diff
@@ -0,0 +1,318 @@
...
+
+Once you create or link to a repository in Docker Cloud, you can set up [automated testing](/docker-cloud/builds/automated-testing.md) and [automated builds](/docker-cloud/builds/automated-build.md).
+
+![a small cute image](/images/footer_moby_icon.png)
+
+> **Note**: This is a note. Lorem ipsum dolor sit amet, consectetur adipisicing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam.
+
+![a pretty wide image](/images/banner_image_24512.png)
+
+<center>
```
johndmulhausen: We should avoid HTML here:

```
This line is formatted with HTML.
{: style="text-align: center" }
```

test.md
```diff
@@ -0,0 +1,318 @@
...
+Some Lorem ipsum text with formatting and styling.
+
+**Lorem ipsum dolor sit amet, consectetur adipisicing elit, sed do eiusmod tempor incididunt** ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo _consequat. Duis aute irure dolor in reprehenderit in voluptate velit esse cillum dolore_ eu fugiat nulla pariatur. Excepteur sint occaecat cupidatat non proident, sunt _**in culpa quiofficia deserunt mollit anim id est laborum.**_
+
+
+### Heading 3
+
+A selection of lists, ordered and unordered, with indented sub elements.
+
+> **Note**: This is some note text.
```
johndmulhausen: How about a blockquote with a snippet in it?

test.md
sanscontext: Have we gone through the repos to remove all instances of this? My goal in including it was to have a test case for something we currently have that might break.

test.md
sanscontext: A code snippet? I can't get a code *block* to work sadly.

test.md
johndmulhausen: We should probably be testing the correct usages instead of all of our various off-the-rails behavior. This can be both a demo and a unit test if we figure out how to properly implement common scenarios.

test.md
johndmulhausen: Try: 

```
> **Note**: Things and stuff
> 
> ```
> docker run -ti -p 4000:4000 docsocker.github.io:latest
> ```
```

test.md
sanscontext: For now, let's test both. Once we've made a cleanup pass, we can easily remove the line from the test file.

johndmulhausen:
Alright let's get this in, it's been quiet for awhile and I can't think of anything more to add. We can always iterate. Thanks for this, @sanscontext :)

PR #602, opened 2016-11-16
Pre-render left nav so it's immediately visible

johndmulhausen:
Currently the left nav takes a second to show up and JavaScript threadblocks for a moment while the left-nav is rendered. This PR makes it so every page gets the left-nav as part of its output. 

Effects:

- Pro: The left-nav is immediately visible when the page loads, though the JavaScript that opens the left-nav to your current URL still takes a second to fire.
- Con: Build time doubles from 45 seconds to 90 seconds.


johndmulhausen:
Deploy preview ready!

Built with commit d549e4c504099f9b858014fbc61cbb382a2b1389

https://deploy-preview-602--docker-docs.netlify.com

johndmulhausen:
Clicked through the entire TOC and this looks good -- I'm waiving it in despite the build time hit because we need to be able to have rendered HTML to support `<noscript>` for Section 508 compliance and it's objectively better to have objects appearing on the screen faster.

PR #609, opened 2016-11-16
Make left nav persist in mobile, fix blue menu pop-in

johndmulhausen:
Makes it so the left-nav stays pinned to the top while viewing the docs in mobile. While not perfect, it at least provides navigability and is in line with what we did in previous versions. (view https://docs.docker.com/v1.11 in mobile, for example).

Edit: Also fixed blue mobile menu popping in by setting it as invisible until after the TOC renders (and it has completed its transition to being offscreen).

cc @docker/design @sanscontext @londoncalling @joaofnfernandes @mstanleyjones 

Note: The diff is so large because I had to de-minify allcss.css and find the specific rule that was causing this.


johndmulhausen:
Deploy preview ready!

Built with commit 8eaa59ff8bce35c1404b4e82de339d5313fdfde6

https://deploy-preview-609--docker-docs.netlify.com

johndmulhausen:
Merging this in, these bugfixes are pretty objectively a good thing; obviously we should evolve the mobile experience with better design but at a minimum we shouldn't be hiding the menus as they were intended to work.

PR #610, opened 2016-11-16
Add DNS SubjectAltName to certificate recipe

### LINKED ISSUE #607: engine/security/https.md leads to generate invalid SSL Certs.
[engine/security/https.md](https://docs.docker.com/engine/security/https/)

Certificates generated by the script in this page generates certificates using the deprecated CommonName and will not be valid.

When using docker-py with latest requests version (2.12.1), the generated certificates [will not be trusted by urllib3 1.19.1][2]. This page should lead to generate certificate with a DNSname Entry in the SubjectAltName along with the IP address to respect the RFC Deprecation.

The [RFC6125][1] supports that client should not check the common name when SubjectAltName IP is provided.

    As noted, a client MUST NOT seek a match for a reference identifier
    of CN-ID if the presented identifiers include a DNS-ID, SRV-ID,
    URI-ID, or any application-specific identifier types supported by the
    client.

[1]: https://tools.ietf.org/html/rfc6125#section-6.4.4
[2]: https://github.com/shazow/urllib3/issues/1043

fbouliane:

If a SubjectAltName is present, the certificates CommonName should be
discarded by the clients. This ensure that people following the guide
generates valid certificates.

### Related issue

Fixes #607

### DIFF
--- a/engine/security/https.md
+++ b/engine/security/https.md
@@ -78,7 +78,7 @@ Since TLS connections can be made via IP address as well as DNS name, they need
 to be specified when creating the certificate. For example, to allow connections
 using `10.10.10.20` and `127.0.0.1`:
 
-    $ echo subjectAltName = IP:10.10.10.20,IP:127.0.0.1 > extfile.cnf
+    $ echo subjectAltName = DNS:$HOST,IP:10.10.10.20,IP:127.0.0.1 > extfile.cnf
 
     $ openssl x509 -req -days 365 -sha256 -in server.csr -CA ca.pem -CAkey ca-key.pem \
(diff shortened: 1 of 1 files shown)


johndmulhausen:
Deploy preview ready!

Built with commit 71368b49c396e264a6866bb7a037746d8d9ca899

https://deploy-preview-610--docker-docs.netlify.com

cyli:
LGTM!  Thank you for this fix!

We could expand the extensions to include basic constraints and extended key usage, but this is more of a comment on additional options it may be good to include in all the certs (root, server, and client) and does not directly address #607, so this should probably be discussed in a separate issue.

mstanleyjones:
Thanks @fbouliane !

PR #611, opened 2016-11-17
modify docs feedback to route support questions properly and encourag…

londoncalling:
### Proposed changes

Strawman update to docs feedback footer. Goal is to re-route issues and questions that are better sent to Support, Forums, and Knowledge Hub and keep this feedback link for docs issues.

### Please take a look

@johndmulhausen @sanscontext @mstanleyjones @joaofnfernandes 

### DIFF
--- a/_layouts/docs.html
+++ b/_layouts/docs.html
@@ -200,9 +200,9 @@ ng\:form {
 							{% if page.noratings != true %}
 							<div style="text-align: center; margin-top: 50px">
 								<img src="/images/chat.png" alt="chat icon" style="margin-right: 10px">
-								<b>Feedback?</b> Questions? Suggestions?<br/>
-							  <a href="https://github.com/docker/docker.github.io/edit/master/{{ page.path }}" class="nomunge">Edit this page</a>,
-								<a href="https://github.com/docker/docker.github.io/issues/new?title=Feedback for: {{ page.path }}&assignee={% if page.assignee %}{{ page.assignee }}{% else %}{{ page.defaultassignee }}{% endif %}&body=File: [{{ page.path }}](https://docs.docker.com{{ page.url }})" class="nomunge">file a ticket</a>, or rate this page:
+								<b>Feedback?</b> Suggestions? Can't find something in the docs?<br/>
+							  <a href="https://github.com/docker/docker.github.io/edit/master/{{ page.path }}" class="nomunge">Edit this page</a> <span style="color:#D8E0E0">&#9679;</span>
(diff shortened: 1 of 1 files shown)


johndmulhausen:
Deploy preview ready!

Built with commit 9d354d74fcedf12d447134e9abf6014ab3f8ff82

https://deploy-preview-611--docker-docs.netlify.com

sanscontext:
WDYT about "`Edit this page`, or `request doc changes`, or `contact Support`?"
I would also suggest linking directly to support.docker.com instead of the www site page.

londoncalling:
@sanscontext @mstanleyjones @johndmulhausen please take a look. 

@sanscontext I didn't want to link directly to https://support.docker.com/ for a couple of reasons - a) only commercial customers with paid subscriptions can log in to that site, so for everyone else it will be a little bit of a dead-end and might increase frustration, b) linking to the the higher level page, https://www.docker.com/docker-support-services,  gives users an idea of other ways to get help or log issues (forums, knowledge base)

Also, Misty and I discussed that there are other ways for users to log bugs and get help, depending on the products they are using. For example, if you have Docker for Mac, you can log issues at https://github.com/docker/for-mac/issues, and if  you have Docker for Windows, you can log issues at https://github.com/docker/for-win/issues. Assuming there might be more of these product-specific channels for logging tickets, I'd like to chat with the website team about adding one or more links to https://www.docker.com/docker-support-services so that if we send users there from our docs, they'll get the full palette of choices (commercial Docker support, community forums, knowledge hub, and product-specific github issues boards).

Looping in @FSHA, @hansfrederickbrown  to follow up on this.

johndmulhausen:
+1 to linking to the www support page for the reasons @londoncalling points out. Let's also not forget that there is a FAQ there that clarifies feedback channels even further for certain support levels.

Other than the fact that "Support" should be lowercase, this LGTM

sanscontext:
Hm, my understanding was that even non-paying users can at least file support tickets via that portal. Maybe that was the old system only? @bryceryan-docker can you confirm?

londoncalling:
@sanscontext From the sound of the description, of "Support" on that page, I think they need to have a paid subscription. But even if they can log support requests, you do still have to log in, and I wouldn't want to send them to something that isn't a little more open than that at first landing. Also, it's useful to see the other options.  Merging this for now, and I'm going to follow up with @FSHA on that Support web page. Also, will see what you hear from @bryceryan-docker as a follow up.

sanscontext:
FWIW, they also have to log in to Github to use either of the other links.

PR #606, opened 2016-11-16
Automatically generate Engine API documentation

bfirsh:
I have replaced the Engine API docs with a version that is automatically generated from a Swagger definition, using [ReDoc](https://github.com/Rebilly/ReDoc). It looks a bit like this:

screenshot 2016-11-16 18 11 40

I have also reorganised the menu for the Engine API to make it simpler and clearer:

screenshot 2016-11-16 21 29 18

Before, it was under "reference", even though some of it wasn't reference material, and it also seemed quite hidden that deep in the menu tree.

I've done a few other things, too:

- Tidied up some URLs (removed superfluous words, `s/_/-/`, etc)
- Moved the accounts API documentation the Docker Hub section. It makes more sense here than the Engine reference, I think – the Hub _serves_ the API, the Engine _consumes_ it.
- Rename Remote API to Engine API. See https://github.com/docker/docker/issues/28319 for details.

### Known issues

- ReDoc is a client-side app that takes a bit of time to load. [The author is working on a fix](https://github.com/Rebilly/ReDoc/issues/62) and I think the downside of this outweighs the benefit of _much_ better documentation.

### Related issues

- This vendors swagger.yaml from this Engine change: https://github.com/docker/docker/pull/28506. It would probably be worth waiting until that is merged so we can be sure we're vendoring the correct thing.
- This is a subset of #210. Just the API docs, and none of the additional SDK docs. As discussed with @johndmulhausen, this seemed like a much less controversial subset that we could get in to start with.
- Background on Swagger definition: https://github.com/docker/docker/issues/27919
- Closes https://github.com/docker/docker/issues/22931

/cc @dnephin @johndmulhausen

PR #634, opened 2016-11-18
Meta Docker ID topic

sanscontext:
This is an adaptation of the current Hub doc on signing up and logging in. We don't (yet) have a central page for login, so I'm using the Cloud page. Also provides links to the Accounts API page where it currently lives (@bfirsh I know you're moving it).

I'm not sure if there's a better way to make the Docker ID a section ( @johndmulhausen ?) , or if we really want it up at the top. 

PTAL @docker/docs-maintainers 


docker-id/index.md
```diff
@@ -0,0 +1,45 @@
+---
+description:  Docker ID
+keywords:
+- accounts, docker ID, billing, paid plans, support, Cloud, Hub, Store, Forums, knowledge base, beta access
+title: Docker ID Accounts
+---
+
+Your free Docker ID grants you access to Docker services such as the Docker
+Store, Docker Cloud, Docker Hub repositories, and some beta programs. Your Docker ID becomes repository namespace used by hosted services such as Docker Hub and Docker Cloud. All you need is an email address.
```
mstanleyjones: nit: long lines

docker-id/index.md
```diff
@@ -0,0 +1,45 @@
...
+
+Your Docker ID becomes your user namespace for hosted Docker services, and becomes your username on the Docker Forums. A Docker ID also
+
+1. Go to the [Docker Cloud sign up page](https://cloud.docker.com).
+
+2. Enter a username, a valid email address, and a password.
+
+3. Click **Sign Up**.
+
+  Docker sends a verification email to the address you provided.
```
mstanleyjones: nit: misalignment may cause the list to go funny.

docker-id/index.md
```diff
@@ -0,0 +1,45 @@
...
+
+## Register for a Docker ID
+
+Your Docker ID becomes your user namespace for hosted Docker services, and becomes your username on the Docker Forums. A Docker ID also
+
+1. Go to the [Docker Cloud sign up page](https://cloud.docker.com).
+
+2. Enter a username, a valid email address, and a password.
+
+3. Click **Sign Up**.
```
mstanleyjones: why not collapse steps 2 and 3?

docker-id/index.md
```diff
@@ -0,0 +1,45 @@
...
+
+  Docker sends a verification email to the address you provided.
+
+  > **Note**: You must click this link to activate your account. You cannot log in with your Docker ID until you verify your email address.
```
mstanleyjones: Seems funny to say you must click the link before telling them to click the link in the next step?

docker-id/index.md
```diff
@@ -0,0 +1,45 @@
...
+
+4. Click the link in the email to verify your email address.
+
+## Login
+
+Once you register for a Docker ID and verify your email address, you can log in
```
mstanleyjones: consider s/you can log in/log in

johndmulhausen:
Deploy preview ready!

Built with commit aec1c4848f31148c031979ee0f9f419767fb8c09

https://deploy-preview-634--docker-docs.netlify.com

docker-id/index.md
```diff
@@ -0,0 +1,45 @@
...
+to Docker services using either the web console with your Docker ID:
+
+![Login using the web interface](images/login-cloud.png)
+
+Or by using the the `docker login` command. (You can read more about `docker login` [here](../engine/reference/commandline/login/) )
```
mstanleyjones: Incomplete sentence. Maybe something like "To log in using the the Docker CLI,..."

docker-id/index.md
```diff
@@ -0,0 +1,45 @@
...
+
+> **Note:** When you use the `docker login` command, your authentication
+> credentials are stored in the `.dockercfg` file in your home directory.
```
mstanleyjones: Are they stored as plain text?

docker-id/index.md
sanscontext: Copypasta. Fixed.

docker-id/index.md
sanscontext: You can, because there's plenty of stuff you can access without having logged in.

docker-id/index.md
sanscontext: Artifact of previous writing. Fixed.

johndmulhausen:
Couple of things:

1. Keywords should be a string: 

```
keywords: foo, bar, etc
```

The Rosebud people will send bot-generated PRs later if you don't fix it here. 

2. I know this supersedes login information in a few places; can you delete the information in those places and add `redirect_from:` metadata to this new topic so we don't have cruft? 

3. Can you make the title less about a thing with a name ("Docker IDs") and more about what its use case is? If not - wouldn't the better term be "Docker accounts" rather than "Docker ID accounts"?

docker-id/index.md
sanscontext: Hashed

sanscontext:
@johndmulhausen for 2: At this point the Docker ID information is intermingled with information about what you get with your Docker ID for each account. That'll be a longer term project. It's easy to remove it for Hub, but less so for Cloud.

3) `Docker ID` is the brand/product name that we want to surface. I added "Accounts" because without it, we have a section called "Docker ID" and the subtopic is also "Docker ID".

docker-id/index.md
```diff
@@ -0,0 +1,50 @@
...
+
+You can also log in using the the `docker login` command. (You can read more about `docker login` [here](../engine/reference/commandline/login/).)
+
+> **Note:** When you use the `docker login` command, your credentials are stored
+in your home directory in `.docker/config.json`. The password is hashed in this
+file.
+
+## The Accounts API
+
+Once you create and verify your Docker ID, you can also update it using the [Docker Accounts API](../engine/reference/api/docker_io_accounts_api/).
```
johndmulhausen: Change to `/docker-hub/accounts-api/` per PR #606 from @bfirsh -- it's likely that we're merging that before this.

johndmulhausen:
Thanks, Laura -- do have a look at these places for where to remove dupe information:

https://docs.docker.com/docker-cloud/dockerid/
https://docs.docker.com/docker-hub/accounts/
https://docs.docker.com/engine/getstarted/step_five/

thaJeztah:
@johndmulhausen regarding "keywords" (and to an extend "description"); Do we actually still need those? AFAIK, Google doesn't use them, and ranking / relevancy is based on Title and actual content of the page.

If I'm right here, I'd really prefer to get rid of them, as it's easy for them to get outdated, so would save a lot of maintenance.

e.g.;
- https://www.searchenginejournal.com/important-tags-2016-seo/156440/
- http://www.hobo-web.co.uk/definitive-guide-to-using-important-meta-tags/

PR #640, opened 2016-11-18
Rewrites to Ubuntu installation

### LINKED ISSUE #203: docs concerning dns-config for Ubuntu are outdated
this concerns the docs at https://docs.docker.com/engine/installation/linux/ubuntulinux/#/configure-a-dns-server-for-use-by-docker

i followed the instructions to configure the DNS to be used, i also knew from prior experiences that it had worked.

however, today i installed Docker on an Ubuntu 16.04 for the first time and i took me a while do figure out that setting the `DOCKER_OPTS` in `/etc/default/docker` does not work, but i had to create a `/etc/docker/daemon.json` and store the config in it.

mstanleyjones:

Rewrite section on DNS config for Ubuntu to use the `daemon.json` config file instead of needing
custom instructions for `upstart` and `systemd`. Clarified that you only need to mess with DNS
settings if you are using `dnsmasq` or see the warning.

Lots of clean-up and gentle rewriting to the file as well.

### Related issue

Fixes #203

### Please take a look

@thaJeztah @funkyfuture @konstruktoid

### DIFF
--- a/engine/installation/linux/ubuntulinux.md
+++ b/engine/installation/linux/ubuntulinux.md
@@ -14,58 +14,61 @@ Docker is supported on these Ubuntu operating systems:
 - Ubuntu Trusty 14.04 (LTS)
 - Ubuntu Precise 12.04 (LTS)
 
-This page instructs you to install using Docker-managed release packages and
-installation mechanisms. Using these packages ensures you get the latest official
+This page instructs you to install Docker on Ubuntu, using packages provided by
+Docker. Using these packages ensures you get the latest official
 release of Docker. If you are required to install using Ubuntu-managed packages,
-consult the Ubuntu documentation.
+consult the Ubuntu documentation. Some files and commands may be different if
+you use Ubuntu-managed packages.
 
 >**Note**: Ubuntu Utopic 14.10 and 15.04 exist in Docker's `APT` repository but
 are no longer officially supported.
 
 ## Prerequisites
 
-Docker requires a 64-bit installation regardless of your Ubuntu version.
-Additionally, your kernel must be 3.10 at minimum. The latest 3.10 minor version
-or a newer maintained version are also acceptable.
+Docker has two important installation requirements:
 
-Kernels older than 3.10 lack some of the features required to run Docker
-containers. These older versions have known bugs which cause data loss
-and frequently panic under certain conditions.
+- Docker only works on a 64-bit Linux installation.
+- Docker requires version 3.10 or higher of the Linux kernel. Kernels older than
+  3.10 lack some of the features required to run Docker containers and contain
+  known bugs which cause data loss and frequently panic under certain conditions.
 
-To check your current kernel version, open a terminal and use `uname -r` to
-display your kernel version:
+  To check your current kernel version, open a terminal and use `uname -r` to
+  display your kernel version:
 
-    $ uname -r
-    3.11.0-15-generic
-
->**Note**: If you previously installed Docker using `APT`, make sure you update
-your `APT` sources to the new Docker repository.
+  ```bash
+  $ uname -r
+  3.11.0-15-generic
+  ```
 
 ### Update your apt sources
 
-Docker's `APT` repository contains Docker 1.7.1 and higher. To set `APT` to use
-packages from the Docker repository:
+To set `APT` to use packages from the Docker repository:
 
 1.  Log into your machine as a user with `sudo` or `root` privileges.
 
 2.  Open a terminal window.
 
-3.  Update package information, ensure that APT works with the `https` method, and that CA certificates are installed.
+3.  Update package information, ensure that APT works with the `https` method,
+    and that CA certificates are installed.
 
     ```bash
     $ sudo apt-get update
     $ sudo apt-get install apt-transport-https ca-certificates
     ```
-4.  Add the new `GPG` key.
+4.  Add the new `GPG` key. This commands downloads the key with the ID
+    `58118E89F3A912897C070ADBF76221572C52609D` from the keyserver
+    `hkp://ha.pool.sks-keyservers.net:80` and adds it to the `adv` keychain.
+    For more info, see the output of `man apt-key`.
 
     ```bash
-    $ sudo apt-key adv --keyserver hkp://ha.pool.sks-keyservers.net:80 --recv-keys 58118E89F3A912897C070ADBF76221572C52609D
+    $ sudo apt-key adv \
+                   --keyserver hkp://ha.pool.sks-keyservers.net:80 \
+                   --recv-keys 58118E89F3A912897C070ADBF76221572C52609D
     ```
 
-5.  Find the entry for your Ubuntu operating system.
-
-    The entry determines where APT will search for packages. The possible entries
-    are:
+5.  Find the entry in the table below which corresponds to your Ubuntu version.
+    This determines where APT will search for Docker packages. When possible,
+    run a long-term support (LTS) edition of Ubuntu.
 
     | Ubuntu version      | Repository                                                  |
     | ------------------- | ----------------------------------------------------------- |
@@ -116,36 +119,33 @@ packages from the Docker repository:
               500 https://apt.dockerproject.org/repo/ ubuntu-trusty/main amd64 Packages
            1.12.0-0~trusty 0
               500 https://apt.dockerproject.org/repo/ ubuntu-trusty/main amd64 Packages
-           1.11.2-0~trusty 0
-              500 https://apt.dockerproject.org/repo/ ubuntu-trusty/main amd64 Packages
-           1.11.1-0~trusty 0
-              500 https://apt.dockerproject.org/repo/ ubuntu-trusty/main amd64 Packages
     ```
 From now on when you run `apt-get upgrade`, `APT` pulls from the new repository.
 
 ### Prerequisites by Ubuntu Version
 
-- Ubuntu Xenial 16.04 (LTS)
-- Ubuntu Wily 15.10
-- Ubuntu Trusty 14.04 (LTS)
+#### Ubuntu Xenial 16.04 (LTS), Wily 15.10, Trusty 14.04 (LTS)
 
-For Ubuntu Trusty, Wily, and Xenial, it's recommended to install the
-`linux-image-extra-*` kernel packages. The `linux-image-extra-*` packages
-allows you use the `aufs` storage driver.
+For Ubuntu Trusty, Wily, and Xenial, install the `linux-image-extra-*` kernel
+packages, which allows you use the `aufs` storage driver.
 
 To install the `linux-image-extra-*` packages:
 
-1. Open a terminal on your Ubuntu host.
+1.  Open a terminal on your Ubuntu host.
 
-2. Update your package manager.
+2.  Update your package manager.
 
-        $ sudo apt-get update
+    ```bash
+    $ sudo apt-get update
+    ```
 
-3. Install the recommended packages.
+3.  Install the recommended packages.
 
-        $ sudo apt-get install linux-image-extra-$(uname -r) linux-image-extra-virtual
+    ```bash
+    $ sudo apt-get install linux-image-extra-$(uname -r) linux-image-extra-virtual
+    ```
 
-4. Go ahead and install Docker.
+4.  Go ahead and [install Docker](ubuntulinux.md#install).
 
 #### Ubuntu Precise 12.04 (LTS)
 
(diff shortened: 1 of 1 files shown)


johndmulhausen:
Deploy preview ready!

Built with commit 461c7a984e9cb7e52e837016a7aa5590ba0ee904

https://deploy-preview-640--docker-docs.netlify.com

funkyfuture:
lgtm

mstanleyjones:
@funkyfuture were you able to follow the rewritten procedure to get your Daemon configured correctly?

engine/installation/linux/ubuntulinux.md
```diff
@@ -116,365 +117,430 @@ packages from the Docker repository:
...
-containers. Or, you can disable `dnsmasq` in NetworkManager. Though, disabling
-`dnsmasq` might make DNS resolution slower on some networks.
+**You only need to use one of these methods.**
 
-The instructions below describe how to configure the Docker daemon
-running on Ubuntu 14.10 or below. Ubuntu 15.04 and above use `systemd`
-as the boot and service manager. Refer to [control and configure Docker
-with systemd](../../admin/systemd.md#custom-docker-daemon-options) to
-configure a daemon controlled by `systemd`.
+#### Specify DNS servers for Docker
```
thaJeztah: Same for this; this is not specify to Ubuntu. We should move this to a general configuration section

engine/installation/linux/ubuntulinux.md
```diff
@@ -116,365 +117,430 @@ packages from the Docker repository:
...
 
-* [Create a docker group](ubuntulinux.md#create-a-docker-group)
+* [Create a `docker` group](ubuntulinux.md#create-a-docker-group)
 * [Adjust memory and swap accounting](ubuntulinux.md#adjust-memory-and-swap-accounting)
 * [Enable UFW forwarding](ubuntulinux.md#enable-ufw-forwarding)
 * [Configure a DNS server for use by Docker](ubuntulinux.md#configure-a-dns-server-for-use-by-docker)
 * [Configure Docker to start on boot](ubuntulinux.md#configure-docker-to-start-on-boot)
 
-### Create a Docker group
+### Create a `docker` group
```
thaJeztah: I'd try to avoid using back-ticks in headings (we removed some in the past), as the formatting can look quite ugly.

thaJeztah: This section is also quite "generic". Giving "non root" access is also mentioned here; https://docs.docker.com/engine/installation/binaries/#/giving-non-root-access

Perhaps good to move it to a general "configuration" section as well

engine/installation/linux/ubuntulinux.md
```diff
@@ -116,365 +117,430 @@ packages from the Docker repository:
...
-              500 https://apt.dockerproject.org/repo/ ubuntu-trusty/main amd64 Packages
     ```
 From now on when you run `apt-get upgrade`, `APT` pulls from the new repository.
 
 ### Prerequisites by Ubuntu Version
 
-- Ubuntu Xenial 16.04 (LTS)
-- Ubuntu Wily 15.10
-- Ubuntu Trusty 14.04 (LTS)
+#### Ubuntu Xenial 16.04 (LTS), Wily 15.10, Trusty 14.04 (LTS)
```
thaJeztah: Note that this will conflict with the `vnext-engine` branch; Support for Ubuntu Wily 15.10 has been removed for Docker 1.13, because it's EOL.

Perhaps we should mention that it's recommended to run an LTS version, or at least make users aware that non-LTS versions only have a 9-month life-span, after which _no_ updates, including security updates are provided by Ubuntu.

engine/installation/linux/ubuntulinux.md
```diff
@@ -14,32 +14,31 @@ Docker is supported on these Ubuntu operating systems:
...
-    3.11.0-15-generic
-
->**Note**: If you previously installed Docker using `APT`, make sure you update
-your `APT` sources to the new Docker repository.
+  ```bash
+  $ uname -r
+  3.11.0-15-generic
+  ```
 
 ### Update your apt sources
```
thaJeztah: I think by now it's safe to remove the first sentence ("Docker's `APT` repository contains Docker 1.7.1 and higher."). We added it because we switched to a new apt repository (so older versions have to be obtained from the old apt repo)

engine/installation/linux/ubuntulinux.md
```diff
@@ -59,13 +59,14 @@ packages from the Docker repository:
 4.  Add the new `GPG` key.
 
     ```bash
-    $ sudo apt-key adv --keyserver hkp://ha.pool.sks-keyservers.net:80 --recv-keys 58118E89F3A912897C070ADBF76221572C52609D
+    $ sudo apt-key adv \
+                  --keyserver hkp://ha.pool.sks-keyservers.net:80 \
+                  --recv-keys \
+                  58118E89F3A912897C070ADBF76221572C52609D
```
thaJeztah: can you move the key _after_ `--recv-keys` here? Make it clearer that it's a parameter for that flag, and not "output" of the command

engine/installation/linux/ubuntulinux.md
```diff
@@ -116,365 +117,430 @@ packages from the Docker repository:
...
+    image.
 
-        $ sudo docker run hello-world
+    ```bash
+    $ sudo docker run hello-world
+    ```
 
     This command downloads a test image and runs it in a container. When the
-    container runs, it prints an informational message. Then, it exits.
+    container runs, it prints an informational message and exits.
```
thaJeztah: One question that keeps popping up is "how to install a specific version". Do you think we should add a short section after this explaining that?

So far we considered it "general knowledge" when using `apt`, but perhaps it's good to show.

For _production_, it's also recommended to pin to a specific version, to prevent inadvertently upgrading docker to a new version (which should never go without thought).

funkyfuture:
@mstanleyjones yes, that reflects how i solved it.

engine/installation/linux/ubuntulinux.md
```diff
@@ -116,365 +117,430 @@ packages from the Docker repository:
...
 
-To prevent these messages, enable memory and swap accounting on your
-system.  Enabling memory and swap accounting does induce both a memory
-overhead and a performance degradation even when Docker is not in
-use. The memory overhead is about 1% of the total available
-memory. The performance degradation is roughly 10%.
+If you don't care about these capabilities, you can ignore the warning. You can
+enable these capabilities in your kernel by following these instructions. Memory
+and swap accounting incur about a 1% memory overhead and a 10% overall
+performance degradation.
```
julianfoad: In changing this paragraph you lose some information that I think was valuable:

- "even when Docker is not in use"
- "of the total available memory"

mstanleyjones:
Addressed feedback except about moving other stuff to a more generic section. I'll reconsider that some more tomorrow.

engine/installation/linux/ubuntulinux.md
```diff
@@ -116,365 +119,494 @@ packages from the Docker repository:
...
+    $ sudo apt-get install docker-engine=1.12.0-0~xenial
+    ```
+
+    If you already have a newer version installed, you will be prompted to
+    downgrade Docker. Otherwise, the specific version will be installed.
+
+3.  Follow steps 4 and 5 of
+    [Install the latest version](ubuntulinux.md#install-the-latest-version).
+
+## Install a testing version
```
thaJeztah: Should we call this "pre-release"?

It may be important to call out that "final" versions are not released through this repository, so to go back to "stable" builds, it's needed to switch back to the "main" repository.

also /cc @andrewhsu @vieux for input (should this be part of the documentation?)

engine/installation/linux/ubuntulinux.md
thaJeztah: See my earlier comment on back-ticks in headings https://github.com/docker/docker.github.io/pull/640/files#r88769224 (wasn't sure if you missed it, or decided to keep it as-is :smile:)

Perhaps better; describe what we're trying to achieve (e.g. "run docker commands without sudo", "run docker commands as a non-privileged user". Need to be careful picking the right words to not have users assume the docker _daemon_ can be run as a non-privileged user)

engine/installation/linux/ubuntulinux.md
```diff
@@ -116,365 +119,494 @@ packages from the Docker repository:
...
+    ```bash
+    $ docker run hello-world
+    ```
 
-5. Verify your work by running `docker` without `sudo`.
+	  If this fails, you will see an error:
 
-        $ docker run hello-world
+    ```none
+		Cannot connect to the Docker daemon. Is 'docker daemon' running on this host?
```
thaJeztah: FWIW, we're improving this error message in 1.13;

if unable to reach the daemon;

    Cannot connect to the Docker daemon at <hostname/address>. Is the docker daemon running?

If the user has insufficient permissions;

    Got permission denied while trying to connect to the Docker daemon socket at <socket path>

engine/installation/linux/ubuntulinux.md
```diff
@@ -116,365 +119,494 @@ packages from the Docker repository:
...
+    ```
 
-	If this fails with a message similar to this:
+	  Check that the `DOCKER_HOST` environment variable is not set for your shell.
+	  If it is, unset it.
 
-		Cannot connect to the Docker daemon. Is 'docker daemon' running on this host?
+    ```bash
+    $ unset DOCKER_HOST
```
thaJeztah: Now that we show how to unset, may be worth showing how to find if it's set, something like:

```bash
$ env | grep DOCKER_HOST
```

engine/installation/linux/ubuntulinux.md
```diff
@@ -116,365 +119,494 @@ packages from the Docker repository:
...
+forwarding traffic. You must set UFW's forwarding policy appropriately.
 
-Also, UFW's default set of rules denies all incoming traffic. If you want to
-reach your containers from another host allow incoming connections on the Docker
-port. The Docker port defaults to `2376` if TLS is enabled or `2375` when it is
-not. If TLS is not enabled, communication is unencrypted. By default, Docker
-runs without TLS enabled.
+In addition, UFW blocks all incoming traffic by default. If you want to access
+your containers from another host, you need to configure UFW to allow incoming
+connections on the Docker port, which defaults to `2376` if TLS encrypted
```
thaJeztah: Looks like some things are mashed together here (already in the original text);

- ports 2376 or 2375 are not needed to access your containers, and only used to remotely access the Docker Remote API. By default, remote access is disabled (the daemon only listens on a socket), so unless reconfigured, these ports don't have to be opened.
- yes, port 2375 (non-TLS) can be used for the remote API, but that's the _biggest_ "no no", as not having it protected by TLS means giving anyone root access to your machine (without logging in, or any security whatsoever) - possibly a warning should be repeated here as well (wdyt?)

thaJeztah: So, step 3 is indeed to make containers accessible, but enabling 2375/2376 is only if the user configured the daemon to allow remote API connections.

engine/installation/linux/ubuntulinux.md
```diff
@@ -116,365 +119,494 @@ packages from the Docker repository:
...
 
-To specify a DNS server for use by Docker:
+The instructions below work whether your Ubuntu installation uses `upstart` or
+`systemd`.
 
-1. Log into Ubuntu as a user with `sudo` privileges.
+1.  Log into Ubuntu as a user with `sudo` privileges.
 
-2. Open the `/etc/default/docker` file for editing.
+2.  Check to see if the Docker daemon has been started using a `--config-file`
```
thaJeztah: Wondering if we should begin with the "possible" exceptions,
or assume the default configuration is used. Maybe we should
"swap" this; something like

```Markdown
The default location of the configuration file is `/etc/docker/daemon.json`.
You can change the location of the configuration file using the `--config-file`
daemon flag. The documentation below assumes the configuration file is
located at `/etc/docker/daemon.json`.
```

engine/installation/linux/ubuntulinux.md
mstanleyjones: I do mention that now. I'll deal with the conflict when it comes up.

engine/installation/linux/ubuntulinux.md
mstanleyjones: Done.

engine/installation/linux/ubuntulinux.md
mstanleyjones: Understood, I'll make a note to change this in `vnext-engine` after I merge `master` into it after this PR is done.

engine/installation/linux/ubuntulinux.md
mstanleyjones: Good idea.

mstanleyjones:
Addressed new feedback from @thaJeztah

engine/installation/linux/ubuntulinux.md
```diff
@@ -116,365 +119,501 @@ packages from the Docker repository:
...
 
 ## Optional configurations
 
-This section contains optional procedures for configuring your Ubuntu to work
-better with Docker.
+This section contains optional procedures for configuring Ubuntu to work better
+with Docker.
 
-* [Create a docker group](ubuntulinux.md#create-a-docker-group)
+* [Create a `docker` group](ubuntulinux.md#create-a-docker-group)
```
thaJeztah: nit: missed renaming this one to "Manage Docker as a non-root user"

engine/installation/linux/ubuntulinux.md
```diff
@@ -116,365 +119,501 @@ packages from the Docker repository:
...
+1.  Log into Ubuntu as a user with `sudo` privileges.
 
-2. Create the `docker` group.
+2.  Create the `docker` group.
+    ```bash
+    $ sudo groupadd docker
+    ```
 
-        $ sudo groupadd docker
+3.  Add your user to `docker` group.
```
thaJeztah: nit: missing "the" (Add your user to _the_ docker group)

engine/installation/linux/ubuntulinux.md
thaJeztah: Cool, it was just a heads up :smile:

engine/installation/linux/ubuntulinux.md
```diff
@@ -116,365 +119,501 @@ packages from the Docker repository:
...
 
-        $ sudo ufw status
+    If `ufw` is not enabled, the remaining steps will not be helpful.
 
-3. Open the `/etc/default/ufw` file for editing.
+3.  If you need to enable access to the Docker Remote API from external hosts
+    and understand the security implications (see the section before this
+    procedure) edit the UFW configuration file, which is usually
+    `/etc/default/ufw` or `/etc/sysconfig/ufw`. Set the `DEFAULT_FORWARD_POLICY`
+    policy to `ACCEPT`.
```
thaJeztah: Can you do a quick check with the network team if this only applies to the API? I think the warning may need to be on step 4., not here

engine/installation/linux/ubuntulinux.md
```diff
@@ -116,365 +119,501 @@ packages from the Docker repository:
...
-4. Set the `DEFAULT_FORWARD_POLICY` policy to:
+    Save and close the file.
 
-        DEFAULT_FORWARD_POLICY="ACCEPT"
+4.  Configure UFW to allow incoming connections on the Docker port, which is
+    2375 if you do not use TLS, and 2376 if you do.
 
-5. Save and close the file.
+    ```bash
+    $ sudo ufw allow 2375/tcp
```
thaJeztah: nit: change to 2376 (to prevent showing 2375 wherever we can)

engine/installation/linux/ubuntulinux.md
```diff
@@ -116,365 +119,501 @@ packages from the Docker repository:
...
 
-7. Allow incoming connections on the Docker port.
+Ubuntu systems which use `networkmanager` use a `dnsmasq` instance that runs on
+a loopback address such as `127.0.0.1` or `127.0.1.1` and adds this entry to
+`/etc/resolv.conf`. The `dnsmasq` service provides a local DNS cache to speed up
+DNS look-ups and also provides DHCP services. However, Docker needs access to a
+fully-functional DNS server. If it detects that no DNS server referenced in
+`/etc/resolv.conf` is a fully functional DNS server, the following warning
+occurs and Docker uses the public DNS servers provided by Google at `8.8.8.8`
+and `8.8.4.4` for DNS resolution.
```
thaJeztah: fwiw; the main reason here is not "fully functional DNS", but the IP-address; the container has its own network namespace, and `127.x.x.x` inside the container is "localhost" _inside_ the container, not the host that the container runs on. Because of that, the container cannot connect to the DNS server, and will fail.

(The only exception is if the container is running with `--network=host`, because then it runs in the host's network namespace / network stack)

engine/installation/linux/ubuntulinux.md
```diff
@@ -116,365 +119,501 @@ packages from the Docker repository:
...
+not automatically removed. To delete all images, containers, and volumes run the
+following command:
 
-    $ rm -rf /var/lib/docker
+```bash
+$ rm -rf /var/lib/docker
+```
 
-You must delete the user created configuration files manually.
+You must delete any customized configuration files manually.
```
thaJeztah: nit: s/files/file/

engine/installation/linux/ubuntulinux.md
mstanleyjones: Great info!!

engine/installation/linux/ubuntulinux.md
mstanleyjones: I don't think so, there may be more than one customized configuration file. "Any files" is fine grammar, just like "any file". :)

thaJeztah: Oh, it felt "strange" (compared it to, e.g. `any person` which is always singular) :blush:

thaJeztah:
nits addressed, merging :)

PR #641, opened 2016-11-18
Update deprecation notes about v1 registry

dmcgowan:
Adds section about the hub deprecating the v1 protocol. Adds note about intent to disable by default and remove support.

Change the wording `Though disabled by default,` to `Though enabled by default,` since the current default is disabled equal to false, or enabled. I found this confusing, but there might still be a better way to word it to make it clearer.

I choose 1.14 as disabled by default since we are already in freeze for 1.13 and making such a change seems unnecessary. We did add the flag in 1.9 but did not mention when the default will change, however I don't believe that means we need to wait 3 more releases to change the default if we are already waiting 3 to remove.

ping @thaJeztah @nwt @dmp42

### DIFF
--- a/engine/deprecated.md
+++ b/engine/deprecated.md
@@ -184,7 +184,13 @@ $ docker -help
 
 ### Interacting with V1 registries
 
-Version 1.9 adds a flag (`--disable-legacy-registry=false`) which prevents the docker daemon from `pull`, `push`, and `login` operations against v1 registries.  Though disabled by default, this signals the intent to deprecate the v1 protocol.
+**Disabled By Default In Release: v1.14**
+
+**Target For Removal In Release: v1.17**
+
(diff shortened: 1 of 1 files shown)


johndmulhausen:
Deploy preview failed.

Built with commit 2a58dc730b7beba4b2ba47a1d2d6c84b705439ef

https://app.netlify.com/sites/docker-docs/deploys/582f4822c4d9cc31880b4492

dmcgowan:
> `You don't have access to this build log.`

Might be helpful to have access to that :smile:

johndmulhausen:
@dmcgowan You appear to be using an old fork. If you just rebase (grab the latest from master on this repo) the deploy preview will work.

johndmulhausen:
Note: We'd love to give the community build log access, but Netlify is still working on that.

dmcgowan:
Closing, looks like it moved back to docker/docker

PR #653, opened 2016-11-19
Add `docker ps -a`

rogererens:
### Please take a look

When one wants to remove an image no longer needed (as in the subsequent example w.r.t. nginx), but still used by a stopped container, one would not find that container using only `docker ps`. Introduction of `docker ps -a` seems warranted to me here.


johndmulhausen:
Deploy preview ready!

Built with commit fd6e83829d37d8b7de93027bd06d1af5e9af124a

https://deploy-preview-653--docker-docs.netlify.com

johndmulhausen:
Thanks, @rogererens :)

PR #684, opened 2016-11-22
Updates to Compose getting started guide

aanand:
- Some wording and formatting tweaks
- Use Python 3.4, because it's 2016
- Use Alpine-based images for faster downloads and less disk space usage
- Use the list form of `CMD` in the Dockerfile, so it's not wrapped in sh
- Remove the `depends_on` from `docker-compose.yml`, as it's not really needed - the web app connects to Redis when it's visited, not on startup
- Add a section on updating the app and seeing the changes, because that's the point of mounting the code in!
- Add a note about `docker-compose down`

### DIFF
--- a/compose/gettingstarted.md
+++ b/compose/gettingstarted.md
@@ -4,26 +4,25 @@ keywords: documentation, docs, docker, compose, orchestration, containers
 title: Get started with Docker Compose
 ---
 
-On this page you build a simple Python web application running on Docker Compose. The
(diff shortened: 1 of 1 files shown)


johndmulhausen:
Deploy preview ready!

Built with commit b08c014fd85a9c4d495adaa947cd8539b5614fb0

https://deploy-preview-684--docker-docs.netlify.com

londoncalling:
LGTM!

Issue #689, opened 2016-11-22
Document locking and unlocking of swarm managers


mstanleyjones:
### Problem description

The raft logs are always encrypted on disk, and the encryption key is written to disk sometimes encrypted (it's encrypted if autolock managers is enabled)

But if your managers are not auto-locked, the key that encrypts the raft data is written basically in plain text in the TLS key's headers.  If it's auto-locked, the key that encrypts the raft data gets rotated, and is stored encrypted in the TLS key's headers.

if managers are auto-locked, they can't be restarted automatically after a reboot, for instance - someone has to go in manually to unlock each individual manager).  But if managers aren't auto-locked, security-wise, the raft data may as well not be encrypted, since the encryption key is stored right there.

we encrypt, and write the key to disk, because it's easier to just rotate the encryption key rather than have to forcibly delete all the unencrypted raft data

### Project version(s) affected

Engine 1.13

### Related PRs

- #568 
- https://github.com/docker/swarmkit/pull/1598
- https://github.com/docker/docker/pull/27967

cc/ @diogomonica @cyli

mstanleyjones:
Issue to track docs for lock / unlock of swarm managers: #689 (PR coming later today)

diogomonica:
- As of 1.13 all raft data is stored encrypted at rest in all the managers by default
- By default we generate a key for you and store it also on disk, so that managers can restart without operator intervention
- However, if a disk gets stolen, or you accidentally backup both the data and the key, you will be effectively leaking all of the cluster data, which starting with 1.13 might include secrets
- Because of this, we allow you to take ownership of this key, and you are able to enable `autolock`, which effectively means that the key used for encryption of your data never gets persisted to disk.
- Taking ownerhip of the key also means that a manager can no longer restart without human intervention, since providing the key is now of the responsibility of the operator (or external application).
- At any point in time you may rotate the key that is used to unlock the cluster, or give the responsibility of managing the key back to the managers, so that they can go back to restarting without external intervention.

mstanleyjones:
Closed via #694

PR #694, opened 2016-11-22
Document locking and unlocking swarm managers

mstanleyjones:

Document locking a swarm.

Engine 1.13

### Related issue

#689 

https://github.com/docker/docker/pull/27967/

### Please take a look

@cyli @aaronlehmann @diogomonica @rahim888


johndmulhausen:
Deploy preview failed.

Built with commit 96672c0fe87de44bdaac9587eea41df4c52fcc90

https://app.netlify.com/sites/docker-docs/deploys/5834aace6686742e1efa016b

engine/swarm/swarm_manager_locking.md
```diff
@@ -0,0 +1,153 @@
...
+
+When Docker restarts, the encryption key needed to encrypt and decrypt RAFT logs
+is loaded into each manager node's running memory from disk. Docker 1.13
+introduces the ability to protect the encryption key at rest by _autolocking_
+your swarm.
+
+When Docker restarts or a new manager attempts to join the swarm, or whenever
+the encryption key needs to be loaded from disk to a node's memory, you must
+[unlock the swarm](swarm_manager_locking.md#unlock-a-swarm) first, using a key
+provided by you or generated by Docker when the swarm was locked. You can rotate
```
cyli: Users cannot provide your own key - it's similar to join tokens.  It is generated by docker when you enable auto-locking, or when you rotate the unlock key.

engine/swarm/swarm_manager_locking.md
```diff
@@ -0,0 +1,153 @@
...
+
+In Docker 1.13 and higher, the RAFT logs used by swarm managers are encrypted on
+disk by default. This at-rest encryption protects your service's configuration
+and data from attackers who gain access to the encrypted RAFT logs. One of the
+reasons this feature was introduced was in support of the new [Docker
+secrets](secrets.md) feature.
+
+When Docker restarts, the encryption key needed to encrypt and decrypt RAFT logs
+is loaded into each manager node's running memory from disk. Docker 1.13
+introduces the ability to protect the encryption key at rest by _autolocking_
```
cyli: It also protects the TLS key (the same unlock key is used to encrypt the TLS key) that the manager node uses to connect to the rest of the swarm, because having a mTLS connection to the rest of the swarm means that you can get all the swarm data (including secrets).

engine/swarm/swarm_manager_locking.md
```diff
@@ -0,0 +1,153 @@
...
+your swarm.
+
+When Docker restarts or a new manager attempts to join the swarm, or whenever
```
cyli: When a new manager joins, the unlock key does not need to be provided, because that new manager does not have any data on disk that needs to be loaded.  The new manager will obtain the the unlock key from the raft leader via propagation of the rest of the raft data.

engine/swarm/swarm_manager_locking.md
```diff
@@ -0,0 +1,153 @@
...
+command and provide the following key:
+
+    SWMKEY-1-+MrE8NgAyKj5r3NcR4FiQMdgu+7W72urH0EZeSmP/0Y
+
+Please remember to store this key in a password manager, since without it you
+will not be able to restart the manager.
+```
+
+To disable autolock, set `--autolock` to `false`. The encryption key used to
+read and write RAFT logs will be stored unencrypted on disk, so this
```
cyli: Maybe "The mutual TLS key and the encryption key used to ..."?

cyli:
Thanks for the great work and quick turnaround on this @mstanleyjones!  A couple of nitpicks, but this looks fantastic.

mstanleyjones:
Addressed feedback so far.

engine/swarm/swarm_manager_locking.md
```diff
@@ -0,0 +1,153 @@
...
+In Docker 1.13 and higher, the RAFT logs used by swarm managers are encrypted on
+disk by default. This at-rest encryption protects your service's configuration
+and data from attackers who gain access to the encrypted RAFT logs. One of the
+reasons this feature was introduced was in support of the new [Docker
+secrets](secrets.md) feature.
+
+When Docker restarts, both the TLS key used to encrypt communication among
+swarm nodes, and the key used to encrypt and decrypt RAFT logs on disk, are
+loaded into each manager node's running memory from disk. Docker 1.13 introduces
+the ability to protect the encryption key at rest by _autolocking_ your swarm.
```
diogomonica: `by allowing you to take ownership of this key and require manual unlocking of your managers. We call this feature autolocking`.

engine/swarm/swarm_manager_locking.md
```diff
@@ -0,0 +1,153 @@
...
+
+When Docker restarts or whenever the encryption key needs to be loaded from disk
+to a node's memory, you must
+[unlock the swarm](swarm_manager_locking.md#unlock-a-swarm) first, using a key
+generated by Docker when the swarm was locked. When a new node joins the swarm,
```
diogomonica: I don't think this `When a new node joins the swarm...` part.

engine/swarm/swarm_manager_locking.md
```diff
@@ -0,0 +1,153 @@
...
+
+    SWMKEY-1-+MrE8NgAyKj5r3NcR4FiQMdgu+7W72urH0EZeSmP/0Y
+
+Please remember to store this key in a password manager, since without it you
+will not be able to restart the manager.
+```
+
+To disable autolock, set `--autolock` to `false`. The mutual TLS key and the
+encryption key used to read and write RAFT logs will be stored unencrypted on
+disk, so this configuration is not recommended.
```
diogomonica: `so, depending on the level of security you're striving for, this configuration might not be recommended.`

engine/swarm/swarm_manager_locking.md
```diff
@@ -0,0 +1,153 @@
...
+
+To unlock a locked swarm, use `docker swarm unlock`.
+
+```bash
+$ docker swarm unlock
+
+Please enter unlock key:
+```
+
+Enter the encryption key that was emitted when you locked the swarm, and the
```
diogomonica: emitted -> generated.

engine/swarm/swarm_manager_locking.md
cyli: I think making it clear here that the unlock key does not need to be provided when adding a manager (only the join token is needed, because the unlock key is propagated via the raft) could be useful - the user might think it needs to be added otherwise, if they're not sure when loading from disk is necessary.

Nitpick though, "the is" is missing a word :)  Maybe "the key is propagated..."?

engine/swarm/swarm_manager_locking.md
cyli: @diogomonica  ownership of both the encryption and TLS key you mean?  Or of the unlock key?

engine/swarm/swarm_manager_locking.md
mstanleyjones: Well, it was generated but it was also emitted to standard output... let me see what I can do.

mstanleyjones:
PTAL again. Thanks!

engine/swarm/swarm_manager_locking.md
```diff
@@ -0,0 +1,162 @@
...
+
+In Docker 1.13 and higher, the RAFT logs used by swarm managers are encrypted on
+disk by default. This at-rest encryption protects your service's configuration
+and data from attackers who gain access to the encrypted RAFT logs. One of the
+reasons this feature was introduced was in support of the new [Docker
+secrets](secrets.md) feature.
+
+When Docker restarts, both the TLS key used to encrypt communication among swarm
+nodes, and the key used to encrypt and decrypt RAFT logs on disk, are loaded
+into each manager node's running memory from disk. Docker 1.13 introduces the
```
diogomonica: drop "from disk"?

engine/swarm/swarm_manager_locking.md
```diff
@@ -0,0 +1,162 @@
...
+ability to protect the mutual TLS encryption key and the key used to encrypt and
+decrypt RAFT logs at rest (on disk), by allowing you to take ownership of this
```
diogomonica: drop "on disk"

engine/swarm/swarm_manager_locking.md
```diff
@@ -0,0 +1,162 @@
...
+key and to require manual unlocking of your managers. This feature is called
```
diogomonica: This feature is called `autolock`.

engine/swarm/swarm_manager_locking.md
```diff
@@ -0,0 +1,162 @@
...
+_autolocking_ your swarm.
+
+When Docker restarts or whenever the encryption key needs to be loaded from disk
```
diogomonica: "When a Docker restarts and needs to access the encrypted RAFT state on disk, you must..."

There aren't other motives to unlock a manager other than starting it (or restarting it).

engine/swarm/swarm_manager_locking.md
```diff
@@ -0,0 +1,162 @@
...
+to a node's memory, you must
+[unlock the swarm](swarm_manager_locking.md#unlock-a-swarm) first, using a key
+generated by Docker when the swarm was locked. You can rotate this key at any time.
+
+>**Note**: You don't need to unlock the swarm when a new node joins the swarm,
+because the key is propagated to it over mutual TLS.. You only need to provide
```
diogomonica: two ".."

engine/swarm/swarm_manager_locking.md
```diff
@@ -0,0 +1,162 @@
...
+the join token. The unlock key will be propagated as part of the RAFT log, which
+is encrypted at rest and transmitted over a mutual TLS connection.
```
diogomonica: Do we need the second part? Maybe just this is enough:

```
**Note**: You don't need to unlock the swarm when a new node joins the swarm,
 +because the key is propagated to it over mutual TLS.
```

engine/swarm/swarm_manager_locking.md
```diff
@@ -0,0 +1,162 @@
...
+
+## Initialize a swarm with autolocking enabled
+
+When you initialize a new swarm, you you can use the `--autolock` flag to
+automatically lock swarm manager nodes.
```
diogomonica: `automatically enable autolocking of swarm manager nodes`. It doesn't actually "lock" anything.

engine/swarm/swarm_manager_locking.md
```diff
@@ -0,0 +1,162 @@
...
+
+To add a manager to this swarm, run 'docker swarm join-token manager' and follow the instructions.
+
+To unlock a swarm manager after it restarts, run the `docker swarm unlock`
+command and provide the following key:
+
+    SWMKEY-1-WuYH/IX284+lRcXuoVf38viIDK3HJEKY13MIHX+tTt8
+```
+
+Record the key in a safe place.
```
diogomonica: Record -> Store?

diogomonica: Also add: We recommend a password manager?

engine/swarm/swarm_manager_locking.md
```diff
@@ -0,0 +1,162 @@
...
+
+When Docker restarts, a new manager node attempts to join the swarm, or Docker
```
diogomonica: please remove: `a new manager node attempts to join the swarm`

engine/swarm/swarm_manager_locking.md
```diff
@@ -0,0 +1,162 @@
...
+
+Please remember to store this key in a password manager, since without it you
+will not be able to restart the manager.
+```
+
+To disable autolock, set `--autolock` to `false`. The mutual TLS key and the
+encryption key used to read and write RAFT logs will be stored unencrypted on
+disk. There is a trade-off between the risk of storing the encryption key
+unencrypted at rest and the convenience of being able to restart a swarm without
+needing to unlock the swarm.
```
diogomonica: `without needing to unlock each manager.`

There is no swarm-wide unlock. This is individual per manager (if all nodes go down, and you have a 3 manager swarm, you need to unlock each manager individually).

engine/swarm/swarm_manager_locking.md
```diff
@@ -0,0 +1,162 @@
...
+```
+
+Enter the encryption key that was generated and shown in the command output when
+you locked the swarm or rotated the key, and the swarm unlocks.
+
+## View the current unlock key for a running swarm
+
+Consider a situation where your swarm is
+running as expected, then manager node becomes unavailable. You troubleshoot the
+problem and bring the physical node back online, but you need to rejoin the
```
diogomonica: You're not rejoining the node to the swarm. I think we should only use "join" when it's the first time a node is actually joining the swarm. Maybe rephrase:

`You troubleshoot the problem and bring the physical node back online, but you need to provide the key necessary to read the encrypted credentials and RAFT logs on disk.`

engine/swarm/swarm_manager_locking.md
```diff
@@ -0,0 +1,162 @@
...
+To unlock a swarm manager after it restarts, run the `docker swarm unlock`
+command and provide the following key:
+
+    SWMKEY-1-8jDgbUNlJtUe5P/lcr9IXGVxqZpZUXPzd+qzcGp4ZYA
+
+Please remember to store this key in a password manager, since without it you
+will not be able to restart the manager.
+```
+
+If you ever need to view the unlock key in this way,
```
diogomonica: Doesn't make a lot of sense, since rotating a swarm also displays the new key. I think we should remove it, or at least replace it with a warning around: please don't share this key anywhere, and store it securely, etc`

engine/swarm/swarm_manager_locking.md
cyli: Maybe just "enable autolocking of swarm manager nodes", since "autolocking" implies automatic?

mstanleyjones:
Addressed @diogomonica's feedback. Please let me know if you see more areas for improvement.

engine/swarm/swarm_manager_locking.md
```diff
@@ -0,0 +1,157 @@
+---
+description: Automatically lock Swarm managers to protect encryption keys
+keywords:
+- swarm, manager, lock, unlock, autolock, encryption
+title: Lock your swarm to protect its encryption key
+---
+
+In Docker 1.13 and higher, the RAFT logs used by swarm managers are encrypted on
```
aaronlehmann: In other docs we seem to use the capitalization `Raft`. This is also what the paper calls it.

engine/swarm/swarm_manager_locking.md
```diff
@@ -0,0 +1,157 @@
...
+
+Error response from daemon: Swarm is encrypted and needs to be unlocked before it can be used. Please use "docker swarm unlock" to unlock it.
+```
+
+## Enable or disable autolock on an existing swarm
+
+To enable autolock on an existing swarm, set the `autolock` flag to `true`.
+
+```bash
+$ docker swarm update --autolock="true"
```
aaronlehmann: The quotes around `true` are not necessary.

engine/swarm/swarm_manager_locking.md
```diff
@@ -0,0 +1,157 @@
...
+```
+
+To disable autolock, set `--autolock` to `false`. The mutual TLS key and the
+encryption key used to read and write RAFT logs will be stored unencrypted on
+disk. There is a trade-off between the risk of storing the encryption key
+unencrypted at rest and the convenience of being able to restart a swarm without
+needing to unlock each manager.
+
+```bash
+$ docker swarm update --autolock="false"
```
aaronlehmann: The quotes around `false` are not necessary.

engine/swarm/swarm_manager_locking.md
```diff
@@ -0,0 +1,157 @@
+---
+description: Automatically lock Swarm managers to protect encryption keys
+keywords:
+- swarm, manager, lock, unlock, autolock, encryption
+title: Lock your swarm to protect its encryption key
+---
+
+In Docker 1.13 and higher, the RAFT logs used by swarm managers are encrypted on
+disk by default. This at-rest encryption protects your service's configuration
```
aaronlehmann: We always encrypt the logs, but may or may not store the key alongside them. `--autolock=true` sets it up not to store the key.

engine/swarm/swarm_manager_locking.md
aaronlehmann: It's technically valid, but it's very atypical to have double quotes there. You could also write it as `docker "swarm" "update" "--autolock"=true`, but people don't do that either. So it looks like there are quotation marks in an arbitrary place where they do nothing, and I think it would look strange to people who understand the shell's syntax rules.

mstanleyjones: OK, removed the quotes.

mstanleyjones:
OK, I think this is ready to merge, just need some +1s.

diogomonica:
LGTM

engine/swarm/swarm_manager_locking.md
```diff
@@ -0,0 +1,157 @@
...
+and data from attackers who gain access to the encrypted Raft logs. One of the
+reasons this feature was introduced was in support of the new [Docker
+secrets](secrets.md) feature.
+
+When Docker restarts, both the TLS key used to encrypt communication among swarm
+nodes, and the key used to encrypt and decrypt Raft logs on disk, are loaded
+into each manager node's memory. Docker 1.13 introduces the ability to protect
+the mutual TLS encryption key and the key used to encrypt and decrypt Raft logs
+at rest, by allowing you to take ownership of this key and to require manual
```
cyli: Minor nitpick:  The first part of the sentence mentions 2 keys, and the second part says "take ownership of this key" - should it be "these keys"?  Or is this referring to the key-encrypting-key (then unlock key) (which hasn't been mentioned before).

engine/swarm/swarm_manager_locking.md
```diff
@@ -0,0 +1,157 @@
...
+unlocking of your managers. This feature is called _autolock_.
+
+When Docker restarts, you must
+[unlock the swarm](swarm_manager_locking.md#unlock-a-swarm) first, using a key
```
cyli: Non-blocking:  I don't know whether it'd be too much detail to mention that this is a key encrypting key - there are two other keys already mentioned in this section, and it might be useful to provide more disambiguation.  Not sure whether it'd just be more confusing, though.

cyli:
One minor nitpick, but this LGTM.  Thanks @mstanleyjones!

mstanleyjones:
Addressed @cyli feedback. If this is good to go, I will merge it after the rebase of `vnext-engine` branch is complete.

PR #701, opened 2016-11-23
rails.md: change the database.yml default database

ryanore:
In this example,  database.yml uses the default value of `postgres` for the `database` key.  
The database postgres is normally not used for user data, but for administrative purposes.  This means people following this example will not be able to drop the database, and in general it's just not a great idea.

### DIFF
--- a/compose/rails.md
+++ b/compose/rails.md
@@ -122,7 +122,7 @@ Replace the contents of `config/database.yml` with the following:
     development: &default
       adapter: postgresql
       encoding: unicode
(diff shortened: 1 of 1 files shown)


johndmulhausen:
Deploy preview ready!

Built with commit 2cee94c0af8493130cfd6be24d00de6a6c1043df

https://deploy-preview-701--docker-docs.netlify.com

johndmulhausen:
Thanks, @ryanore this is great. :)

PR #712, opened 2016-11-24
Update docs around entrypoint option

aanand:
Replaces https://github.com/docker/compose/pull/3764

- Clarify that setting `entrypoint` both overrides the entrypoint and resets the command on the image (taking @SvenDowideit's comments into account)
- Fix the examples in the doc on startup order, which incorrectly assumed that the default command on the image would not be reset when setting `entrypoint`

### DIFF
--- a/compose/compose-file.md
+++ b/compose/compose-file.md
@@ -271,6 +271,11 @@ The entrypoint can also be a list, in a manner similar to
         - memory_limit=-1
         - vendor/bin/phpunit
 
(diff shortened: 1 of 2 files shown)


johndmulhausen:
Deploy preview ready!

Built with commit bd370d86eabb2f9875ea54ebafbde657c4a937a1

https://deploy-preview-712--docker-docs.netlify.com

johndmulhausen:
Thanks, @aanand :)

PR #722, opened 2016-11-27
Extend Logging documentation

Kukunin:

Current documentation about logging in Docker seems confusing. It even causes wrong PR being merged into Ansible's `docker_container` module: https://github.com/ansible/ansible-modules-core/pull/5482

[Documentation](https://docs.docker.com/engine/admin/logging/overview/) says:

> Use the --log-driver=VALUE with the docker run command to configure the container’s logging driver. If the --log-driver option is not set, docker uses the default (json-file) logging driver.

However, it's not completely true. If my daemon is running with different `log_driver` option, all containers without explicit `log_driver` option will fall back to daemon's log_driver, which is not always `json-file`.

With this PR I make documentation more explicit, and fix the confusion

https://github.com/docker/docker/issues/28860
https://github.com/ansible/ansible-modules-core/pull/5482

### DIFF
--- a/engine/admin/logging/overview.md
+++ b/engine/admin/logging/overview.md
@@ -6,11 +6,17 @@ redirect_from:
 title: Configure logging drivers
 ---
 
+Docker comes with support of multiple logging mechanisms. Use the
+`--log-driver=VALUE` when starting the `dockerd` daemon to set default logging
+driver. You can use the `--log-opt NAME=VALUE` flag to specify extra options for
+selected driver.
+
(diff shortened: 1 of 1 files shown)


Kukunin:
btw, I'm not native speaker, tried as hard as possible to avoid mistakes, but feel free to correct me

johndmulhausen:
Deploy preview ready!

Built with commit 72a0e94f467c28a0334db60df11cb8b12ef331d8

https://deploy-preview-722--docker-docs.netlify.com

PR #737, opened 2016-11-29
Initial dump of CLI in yaml

FrenchBen:

Initial dump of CLI commands in YAML format. Keep in mind that multiple iterations may be needed in order to get the final desired output

1.13 branch of docker/docker

https://github.com/docker/docker/pull/26830

### Please take a look

@johndmulhausen @mstanleyjones 

Couple of notes:
`pname` parent command name
`plink` parent file link
`cname` child command name
`clink` child file link

Most of the parent command had child commands - I didn't loop through all of them, but can do so if we think that the dependency should be bi-directional.

### DIFF
--- a/_data/engine-cli/docker.yaml
+++ b/_data/engine-cli/docker.yaml
@@ -1,0 +1,3 @@
+command: docker
+cname: docker wait
+clink: docker_wait.yaml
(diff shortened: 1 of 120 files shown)


johndmulhausen:
I'm waving these in so we have something to work with; feel free to send updates anytime, @FrenchBen !

PR #742, opened 2016-11-29
Update Compose UI application .yml example

vsaraswat:
### Please take a look

Current Compose UI example is v1 and out of date. Changing code to v2 example from Compose Quickstart (also in docs) so that this works. We do still need to change the screenshots, but at least this will bring working code.


johndmulhausen:
Deploy preview ready!

Built with commit a566d9aeeae12f86f6042b130f92d506a4a91567

https://deploy-preview-742--docker-docs.netlify.com

PR #765, opened 2016-12-02
Removing leave off the `-P` flag and also

gurubelli:
### Please take a look

I feel this sentence 'leave off the `-P` flag and also ' not necessary as that is not used that flag in the previous. User gets confused about this flag.
I am not sure any significance of that flag. If it is necessary to mention , we should better document it.

Thanks,
Srinivasarao Gurubelli


johndmulhausen:
Deploy preview ready!

Built with commit 234fe335292de52eb549e79ba309c2ac05bbef2b

https://deploy-preview-765--docker-docs.netlify.com

johndmulhausen:
Yep, looks good; we never even tell people to use `-P` in the earlier examples, heh. Thanks, @gurubelli :)

PR #823, opened 2016-12-07
CI - don't allow absolute links to docs.docker.com

aduermael:

Added a test to prevent absolute links to `docs.docker.com`. 
Also made a few absolute links relative for this test to pass.

### DIFF
--- a/docker-cloud/builds/advanced.md
+++ b/docker-cloud/builds/advanced.md
@@ -86,7 +86,7 @@ docker build --build-arg CUSTOM=$VAR -t $IMAGE_NAME
 used by the builder, so you must include a similar build command in the hook or
 the automated build will fail.
 
-To learn more about Docker build-time variables, see the [docker build documentation]( https://docs.docker.com/engine/reference/commandline/build/#/set-build-time-variables---build-arg).
+To learn more about Docker build-time variables, see the [docker build documentation](/engine/reference/commandline/build/#/set-build-time-variables---build-arg).
 
 #### Two-phase build
 
--- a/docker-cloud/builds/automated-build.md
+++ b/docker-cloud/builds/automated-build.md
@@ -85,7 +85,7 @@ the code repository service where the image's source code is stored.
 
 10. For each branch or tag, enable or disable the **Build Caching** toggle.
 
-    [Build caching](https://docs.docker.com/engine/userguide/eng-image/dockerfile_best-practices/#/build-cache) can save time if you are building a large image frequently or have many dependencies. You might want to leave build caching disabled to make sure all of your dependencies are resolved at build time, or if you have a large layer that is quicker to build locally.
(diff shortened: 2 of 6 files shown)


johndmulhausen:
Deploy preview ready!

Built with commit e68bbc60163c97964b249bfbb4318cbb39d68fc6

https://deploy-preview-823--docker-docs.netlify.com

aduermael:
@johndmulhausen I'll now add more tests in the same scope to check for dead links or infinite redirection loops... But this one can be merged already.

sanscontext:
Will we have the ability to disable this for specific layout features? I'm specifically thinking of links in banners which might be applied at build time, and which since they could be applied anywhere, would need a static url.

aduermael:
@sanscontext sure, it's super flexible. so far it only checks published markdown files. we can add exceptions if needed.

sanscontext:
SGTM! Thanks.

aduermael:
@johndmulhausen can you merge it please? I'll submit new tests today, but I would prefer that one to be merged first. Thanks!

aduermael:
I rebased and had to fix a new absolute link in `engine/installation/binaries.md`...

johndmulhausen:
@aduermael I agree with the goal of this test, however it appears you're parsing links with RegEx here. I don't want to set a precedent of relying on clever/arcane RegEx Markdown processing instead of freely available Go libraries such as Blackfriday.

With Blackfriday you can use Bluemonday to sanitize input:

```
import (
    "github.com/microcosm-cc/bluemonday"
    "github.com/russross/blackfriday"
)

// ...
unsafe := blackfriday.MarkdownCommon(input)
html := bluemonday.UGCPolicy().SanitizeBytes(unsafe)
```

Then, if that completes (obviously the CI check should fail if it doesn't), run `html` through [the html Go library](https://godoc.org/golang.org/x/net/html) and do the link checking there with a data structure that represents the document object model. 

Parsing raw Markdown with RegEx is a recipe for unintended consequences.

aduermael:
@johndmulhausen It's way easier to parse markdown than html... I know `blackfriday` and we may have to use it for other tests, but I don't understand why in this case I would transform our markdown into html before parsing it. Sorry I just don't get it.

> Parsing raw Markdown with RegEx is a recipe for unintended consequences.

Please give me a concrete example.

johndmulhausen:
I understand it's easier to implement it this way, but I'm afraid it is also a low-quality approach that is fragile. Just to cite one concrete example: What if we want to document how our Markdown works for contributors? We've talked about such a project.

The probem is, your code has no concept of what a link is, it just hopes that matching `[]()` will always do what you want. When you started validating YAML with CI tests you knew it was important to include a YAML parser so that you would have strongly-typed data and ready-made functions that had a notion of a YAML structure. Now you want to venture into processing page content but with no concept of a document? I don't think we should go there.

PR #832, opened 2016-12-09
Ucp dtr release notes

joaofnfernandes:


johndmulhausen:
Deploy preview ready!

Built with commit cb09cd06887f4753d5e571e913617fb8821c0150

https://deploy-preview-832--docker-docs.netlify.com

datacenter/dtr/2.1/guides/release-notes.md
```diff
@@ -13,6 +13,23 @@ known issues for each DTR version.
...
 to upgrade your installation to the latest release.
 
+## DTR 2.1.2
+
+(8 Dec 2016)
+
+**Features**
+
+* The web UI now alerts when no backups have been made in a week
+* Google Cloud Storage driver now throttles data if there's heavy load
```
pdevine: This is more of a bug to me than a feature.

datacenter/dtr/2.1/guides/release-notes.md
```diff
@@ -13,6 +13,23 @@ known issues for each DTR version.
...
+
+
+**Bug fixes**
+
+* Restore operation now prints logs
+* Upgraded alpine images used by the DTR services to fix security vulnerabilities
+* Upgrading from 2.0 to 2.1 no longer makes tags disappear
```
pdevine: This only happens in certain circumstances which you can see in the issue.

datacenter/dtr/2.1/guides/release-notes.md
```diff
@@ -13,6 +13,23 @@ known issues for each DTR version.
...
+**Features**
+
+* The web UI now alerts when no backups have been made in a week
+* Google Cloud Storage driver now throttles data if there's heavy load
+
+
+**Bug fixes**
+
+* Restore operation now prints logs
+* Upgraded alpine images used by the DTR services to fix security vulnerabilities
```
pdevine: Only the expat2 library is changed.  The problem is there are false positives right now with two other packages which can be confusing to users who think we have vulnerabilities if you don't call out which CVE was fixed.

PR #844, opened 2016-12-10
Automate display of urls for DDC offline install

joaofnfernandes:


johndmulhausen:
Deploy preview ready!

Built with commit 9cdcc0dedef7cb6a3f405dc65413838e61bcfa2a

https://deploy-preview-844--docker-docs.netlify.com

joaofnfernandes:
@mbentley what do you think [about this?](https://deploy-preview-844--docker-docs.netlify.com/datacenter/ucp/2.0/guides/installation/install-offline/)

mbentley:
Nice, that definitely helps with the history. Seems like it would be ideal
if the tar balls were decoupled as UCP and DTR instead of both but I know
we are just talking about the docs and having the links to each version is
perfect.

joaofnfernandes:
Thanks for the feedback. You're right, it's kind of confusing that we provide a single tar file for both UCP and DTR. I still need to improve the docs, but a caveat that is not mentioned anywhere is that the images in the tar file are not tagged as latest, so you'll have to mention the version when running the install command.

mbentley:
No problem.  Yeah, I think that having people specify a version tag does make sense to prevent accidentally installing a version that isn't anticipated.  A bit annoying if you're used to always just running `latest` but I supposed it is better better to be explicit.

PR #845, opened 2016-12-10
formatting updates on d4mac

londoncalling:
### Proposed changes

Added a troubleshooting topic for container access problem as described in the referenced issue.

### Related issue

https://github.com/docker/for-win/issues/197

### Please take a look

@friism @rneugeba @dgageot @aysark I'm sure this is not a good description of the issue, or solution, but it's a strawman ... can you all take a look and help craft a useful troubleshooting description along with a simple example that is a little simpler and more generic than that given in the issue?  I started to try and do this but really couldn't figure it out. 

It would be nice to base something off of the `nginx` example we start with ([#5 under Explore the Application in the Getting Started](https://docs.docker.com/docker-for-windows/#/step-4-explore-the-application-and-run-examples)) . Can you walk me through a good example like this? Then I can add it here and cross link from FAQ's if needed.

Also, would the same troubleshooting tip apply on Docker for Mac? 

Thanks

p.s. I fixed a formatting issue on Docker for Mac, unrelated. The file to look at here is the Docker for Windows file.

### DIFF
--- a/docker-for-mac/troubleshoot.md
+++ b/docker-for-mac/troubleshoot.md
@@ -278,15 +278,14 @@ environments, see [Docker for Mac vs. Docker Toolbox](docker-toolbox.md).
 
 <p></p>
 
-* If you see errors like `Bind for 0.0.0.0:8080 failed: port is already allocated` or
-  `listen tcp:0.0.0.0:8080: bind: address is already in use`:
+* If you see errors like `Bind for 0.0.0.0:8080 failed: port is already allocated` or `listen tcp:0.0.0.0:8080: bind: address is already in use`:
 
-	  These errors are often caused by some other software on the Mac using those ports.
-		Run `lsof -i tcp:8080` to discover the name and pid of the other process and
-		decide whether to shut the other process down, or to use a different port in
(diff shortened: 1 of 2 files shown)


johndmulhausen:
Deploy preview ready!

Built with commit 506f81063c8bd3e7f472b8f511082324d47ba885

https://deploy-preview-845--docker-docs.netlify.com

londoncalling:
@justincormack @friism I took out the troubleshooting tip with which I was trying address [docker/for-win#197](https://github.com/docker/for-win/issues/197). Per your Slack comments, it doesn't sound like there is anything appropriate I can add to the FAQ's or Troubleshooting on this. 

Justin:
@victoria I don't think this is right. People have expectations that are not eg  even guaranteed in swarm mode. They should not make the assumption that docker traffic comes from localhost

Michael:
@victoria while really painful, I'm not sure we should tell people to use rsync and similar to work around fs filesystem performance problems: https://github.com/docker/docker.github.io/pull/845 - what do others think?

If you disagree, please re-open. 

I'm going to submit this PR just for the formatting updates on d4Mac.

PR #849, opened 2016-12-12
CI - no absolute links to local pages

aduermael:
### Describe the proposed changes

New tests to make sure there are no absolute links to `docs.docker.com` in local pages.
- <s>makes sure all local pages referenced by links do exist (and not though redirects)</s>
- <s>makes sure all redirects point to existing pages</s>

**edit:** I'll take care of relative links checking in a different PR.

**work in progress...**

PR #852, opened 2016-12-12
replaced href by src in img tag (v1.7)

aduermael:
### Proposed changes

Fixed a few img tags, replacing `href` by `src` in **v1.7**.
I need this for my new tests to pass: #849 

It's the reason why we can't see the Docker logo on some pages, like this one: https://docs.docker.com/v1.7/registry/introduction/


johndmulhausen:
Thanks, @aduermael :)

PR #871, opened 2016-12-13
Dockerfile - don’t copy files from /usr/src/app/allvbuild to /_site

aduermael:
### Proposed changes

`jekyll serve -s /usr/src/app/allvbuild -d /_site --no-watch -H 0.0.0.0 -P 4000` builds what's in  `/usr/src/app/allvbuild` and puts result in `/_site`. 

But everything has already been built `/usr/src/app/allvbuild`, so we could just serve what's in there instead of copying ~800MB when running the container:

`jekyll serve -s /usr/src/app/allvbuild --no-watch -H 0.0.0.0 -P 4000`


johndmulhausen:
Thanks, @aduermael !

PR #872, opened 2016-12-13
Dockerfile - remove absolute links to docs.docker.com

aduermael:
### Proposed changes

Most absolute links to `docs.docker.com` were made relative when building the Docker image. Most of them but these ones:
- links to `http://docs.docker.com` (`http` without the `s`)
- absolute links in `` tags
- absolute links in current version (only `/v1.x` folders were considered)

@johndmulhausen I don't really like to be permissive with markdown writing. I would personally parse these markdowns in the tests to fix the root cause not the symptoms. But that's your call, at least we make sure there aren't any absolute links to `docs.docker.com` in the image. 🙂

cc @gdevillele

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -9,8 +9,15 @@ RUN git clone https://www.github.com/docker/docker.github.io temp; \
 		&& mkdir -p allvbuild/${VER} \
 		&& jekyll build -s temp -d allvbuild/${VER} \
 		&& find allvbuild/${VER} -type f -name '*.html' -print0 | xargs -0 sed -i 's#href="/#href="/'"$VER"'/#g' \
+		&& find allvbuild/${VER} -type f -name '*.html' -print0 | xargs -0 sed -i 's#href="https://docs.docker.com/#href="/'"$VER"'/#g' \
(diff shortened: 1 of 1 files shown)


johndmulhausen:
Deploy preview ready!

Built with commit 8cba3e792e089fbfc270d27103f09b1e4f876bec

https://deploy-preview-872--docker-docs.netlify.com

johndmulhausen:
@aduermael I agree with you that we should probably fix these in the source rather than in the Dockerfile. This will add significant processing time; if we just fix it in the source and look for docs.docker.com links with CI logic, I think that's the better approach. We have long build times already -- we could actually lose the search for docs.docker.com instances instead of adding more searches if we fixed these in the source. What do you think?

aduermael:
It's not adding a lot of build time, and it's a quick way to fix them all. 
We can go and parse the markdown to fix the problem at its source, but that would be only for current and future versions, as htmls are already generated for older ones. 
We can fix the htmls easily as they're not going to evolve much...

johndmulhausen:
Yeah, let's fix the HTML in the source, kinda like you did in that PR that processed the `v1.7` branch recently; this is just piling on top of something we don't need to be constantly re-doing.

PR #883, opened 2016-12-14
fixed absolute links

aduermael:
### Proposed changes

Fixed absolute links in current version's content for new tests in #849  to pass.

### DIFF
--- a/docker-cloud/builds/push-images.md
+++ b/docker-cloud/builds/push-images.md
@@ -15,7 +15,7 @@ Images pushed to Docker Hub automatically appear for you in the **Launch new
 service** wizard.
 
 > **Note**: You must use Docker Engine 1.6 or later to push to Docker Hub.
-Follow the <a href="http://docs.docker.com/installation/" target="_blank">official installation instructions</a> depending on your system.
+Follow the <a href="/installation/" target="_blank">official installation instructions</a> depending on your system.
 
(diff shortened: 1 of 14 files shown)


sanscontext:
Does this actually work without breaking these links in the Docker Cloud docs? I wasn't aware that HTML would allow relative links on `target="_blank"` anchors.  

(For reference, I'm making a specific design decision in the Cloud docs to open a new tab so that the reader isn't distracted from a tutorial flow. This modifier isn't supported by markdown, which is why it's in HTML.)

aduermael:
@sanscontext 
>  I wasn't aware that HTML would allow relative links on target="_blank" anchors.

I assume it works... Can you point me to a page that could bring that kind of issue?

aduermael:
@sanscontext 
I just tried this link in `/docker-cloud/builds/push-images/`:
screen shot 2016-12-14 at 12 20 15 pm

It opens a new tab, even with a relative link.

sanscontext:
Sorry, I'd been waiting for the Netlify preview to show up - forgot it's now hidden three clicks deep. :P  All my specific links appear to be working!

aduermael:
ping @johndmulhausen

johndmulhausen:
Thanks, @aduermael :)

aduermael:
thanks @johndmulhausen!

PR #886, opened 2016-12-14
Fixed absolute links in external content

aduermael:
### Proposed changes

The files we’re pulling from docker/docker may include links to `docs.docker.com`. And we can’t forbid that because relative links wouldn’t make sense in the context of `docker/docker` repository in some situations. So let’s just fix these links right after they get imported.

This is necessary for tests in #849 to pass.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -29,6 +29,8 @@ RUN svn co https://github.com/docker/docker/branches/$ENGINE_BRANCH/docs/referen
  && svn co https://github.com/docker/distribution/branches/$DISTRIBUTION_BRANCH/docs/spec allv/registry/spec \
  && wget -O allv/registry/configuration.md https://raw.githubusercontent.com/docker/distribution/$DISTRIBUTION_BRANCH/docs/configuration.md \
  && jekyll build -s allv -d allvbuild \
+ && find allvbuild/engine/reference -type f -name '*.html' -print0 | xargs -0 sed -i 's#href="https://docs.docker.com/#href="/#g' \
+ && find allvbuild/engine/extend -type f -name '*.html' -print0 | xargs -0 sed -i 's#href="https://docs.docker.com/#href="/#g' \
  && rm -rf allv
 
 # Serve the site, which is now all static HTML


aduermael:
ping @johndmulhausen

Dockerfile
```diff
@@ -29,6 +29,8 @@ RUN svn co https://github.com/docker/docker/branches/$ENGINE_BRANCH/docs/referen
  && svn co https://github.com/docker/distribution/branches/$DISTRIBUTION_BRANCH/docs/spec allv/registry/spec \
  && wget -O allv/registry/configuration.md https://raw.githubusercontent.com/docker/distribution/$DISTRIBUTION_BRANCH/docs/configuration.md \
  && jekyll build -s allv -d allvbuild \
+ && find allvbuild/engine/reference -type f -name '*.html' -print0 | xargs -0 sed -i 's#href="https://docs.docker.com/#href="/#g' \
+ && find allvbuild/engine/extend -type f -name '*.html' -print0 | xargs -0 sed -i 's#href="https://docs.docker.com/#href="/#g' \
```
thaJeztah: Adding the obligatory link to the "don't use regex to parse HTML" post http://stackoverflow.com/a/1732454/1811501 here 😇

thaJeztah: But yeah, I guess this works; when thinking about this, I wondered if we could simply strip `https://docs.docker.com/` in case there's links somehow that used single quotes, but I assume that's not a problem here

aduermael: hehe, yes, I don't like doing this too much...

aduermael:
@johndmulhausen can you merge that one please if you don't have comments?

These links are the only reason why this new test fails: #849 

```
18:55:52 --- FAIL: TestURLs (2.11s)
18:55:52 	html_test.go:53: found absolute link: <a href="https://docs.docker.com/engine/admin/logging/overview/"> - /engine/reference/commandline/logs/index.html
18:55:52 	html_test.go:53: found absolute link: <a href="https://docs.docker.com/engine/tutorials/dockervolumes/"> - /engine/reference/commandline/run/index.html
18:55:52 	html_test.go:53: found absolute link: <a href="https://docs.docker.com/engine/swarm/ingress/"> - /engine/reference/commandline/service_create/index.html
```

I would like to move forward and keep fixing content in docker/docker.github.io. If we don't need these 2 lines in a near future, we can simply remove them. 

Thanks!

aduermael:
thanks @thaJeztah !

Dockerfile
johndmulhausen: Nobody likes doing this - it's just the worst. :/

johndmulhausen:
Thanks, @aduermael :)

PR #889, opened 2016-12-15
update django.md

oonid:
### Related issues (optinoal)

update database ENGINE to use django.db.backends.postgresql_psycopg2
instead of django.db.backends.postgresql (error no module).


johndmulhausen:
Pretty sure we don't want to use that - both exist.  https://github.com/django/django/tree/master/django/db/backends

PR #892, opened 2016-12-15
Remove links to non-existent anchors

mstanleyjones:
### Proposed changes

Remove links to non-existent anchors
Remove "64-bit" info from supported versions because we already cover that in the prerequisites


johndmulhausen:
Thanks, @mstanleyjones :)

PR #902, opened 2016-12-15
updated doc for OSS docker engine version

amitsehgal:
### Related issues (optinoal)

FATA[0000] Your engine version 1.12.3 OSS is too old.  UCP requires at least version 1.12.1 CS or 1.13.0 OSS.

Also, can you publish 1.13.0 for https://yum.dockerproject.org/repo/main/centos/7/Packages/


joaofnfernandes:
@amitsehgal what version of UCP were you trying to install when you saw this message?

johndmulhausen:
I think this is probably inaccurate; we don't require both CS and open-source Docker for UCP as this PR implies. It's easy to have your environment improperly configured, my guess is that's what happened here. No response from the author for over a month. Closing.

PR #913, opened 2016-12-16
WIP: First attempt at one Dockerfile per version with nginx as reverse proxy

mstanleyjones:
### Proposed changes

- Create one Docker image per archive branch (and master)
- Use nginx to reverse-proxy requests to the archive URLs to the correct container
- Use Docker Compose to orchestrate the service

This is a work in progress!

To build all the images, run `_data/docsarchive/build_all_archives.sh`
To stand up the service, change to `_data/docsarchive/proxy/` and run `docker-compose up --build`


_data/docsarchive/Dockerfile
```diff
@@ -0,0 +1,14 @@
+FROM starefossen/github-pages
+# The branch name to check out, also used in some find/replace
+# This is just the default, specify at build-time using --build-arg VER=<foo>
+ARG VER="v1.11"
```
thaJeztah: We need to "bake" the build-arg into the image so that it's available at runtime

_data/docsarchive/Dockerfile
```diff
@@ -0,0 +1,14 @@
...
+
+# Get the branch contents without history
+RUN echo "Started $VER" \
+    && git clone --depth 1 --branch $VER https://github.com/docker/docker.github.io docs_src \
+    && jekyll build -s docs_src -d docs \
+    && rm -rf docs_src \
+    && find docs -type f -name '*.html' -print0 | xargs -0 sed -i 's#href="https://docs.docker.com/#href="/'"$VER"'/#g'
```
thaJeztah: looks like relative links should also be replaced, and `src=` (images)

thaJeztah:
Here's a patch that seems to go in the right direction;


```patch
diff --git a/_data/docsarchive/Dockerfile b/_data/docsarchive/Dockerfile
index 9c5a9e0..3ba39d3 100644
--- a/_data/docsarchive/Dockerfile
+++ b/_data/docsarchive/Dockerfile
@@ -3,12 +3,19 @@ FROM starefossen/github-pages
 # This is just the default, specify at build-time using --build-arg VER=<foo>
 ARG VER="v1.11"
 
+# Bake the version into the image
+# it's needed at runtime (CMD)
+ENV VER=$VER
+
 # Get the branch contents without history
 RUN echo "Started $VER" \
     && git clone --depth 1 --branch $VER https://github.com/docker/docker.github.io docs_src \
     && jekyll build -s docs_src -d docs \
     && rm -rf docs_src \
-    && find docs -type f -name '*.html' -print0 | xargs -0 sed -i 's#href="https://docs.docker.com/#href="/'"$VER"'/#g'
+    && find docs -type f -name '*.html' -print0 | xargs -0 sed -i 's#href="https://docs.docker.com/#href="/'"$VER"'/#g' \
+    && find docs -type f -name '*.html' -print0 | xargs -0 sed -i 's#src="https://docs.docker.com/#src="/'"$VER"'/#g' \
+    && find docs -type f -name '*.html' -print0 | xargs -0 sed -i 's#href="/#href="/'"$VER"'/#g' \
+    && find docs -type f -name '*.html' -print0 | xargs -0 sed -i 's#src="/#src="/'"$VER"'/#g'
 
 
-CMD jekyll serve -b $VER -s docs --no-watch -H 0.0.0.0 -P 4000
+CMD jekyll serve -b /$VER -s docs --no-watch -H 0.0.0.0 -P 4000
diff --git a/_data/docsarchive/proxy/nginx.vh.default.conf b/_data/docsarchive/proxy/nginx.vh.default.conf
index 3f00ca9..601036d 100644
--- a/_data/docsarchive/proxy/nginx.vh.default.conf
+++ b/_data/docsarchive/proxy/nginx.vh.default.conf
@@ -6,15 +6,15 @@ server {
   charset utf-8;
 
   location / {
-    proxy_pass http://docs-current:4000/;
+    proxy_pass http://docs-current:4000/master/;
   }
 
   location ~ ^/(current|v1.12)(/?)(.*)$ {
-    proxy_pass http://docs-current:4000/$3$is_args$args;
+    proxy_pass http://docs-current:4000/master$2$3$is_args$args;
   }
 
   location ~ ^/v1\.([0-9]+)(/?)(.*)$ {
-    proxy_pass http://docs-archive-1-$1:4000/$3$is_args$args;
+    proxy_pass http://docs-archive-1-$1:4000/v1.$1$2$3$is_args$args;
   }
 
   #error_page  404              /404.html;
```

_data/docsarchive/Dockerfile
johndmulhausen: Let's fix references to docs.docker.com in the source once and for all and get rid of this line.

_data/docsarchive/build_all_archives.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
+ARCHIVE_BRANCHES=("v1.4" "v1.5" "v1.6" "v1.7" "v1.8" "v1.9" "v1.10" "v1.11" "master")
```
thaJeztah: eventually, this won't be needed if each branch has a Dockerfile in it, then we can setup auto builds for each branch

_data/docsarchive/Dockerfile
thaJeztah: Don't think they are super problematic, I can see them being useful while we still have some parts of the docs in a separate repo. We need to rewrite here anyway

johndmulhausen:
The testing procedure for a full stack redesign like this must involve:

- setting up a parallel stack on Docker Cloud that uses this
- setting up a three-node cluster with deploy-on-each strategy that runs the stack (so that each node is a clone of each other, each running only one stack instance)
- setting up a load balancer in the Stackfile for Docker Cloud that forwards to the proxy addresses on the individual nodes (you can reuse the one in `docs/docker.github.io:ssl`)

Once you verify that the loadbalancer is properly forwarding traffic to all three nodes, is getting redeploys, and is performing sequential redeploys, we can think about adopting this provided you have a plan for scale -- currently we have a single server instance running on each micro-sized VM on AWS, each with 1GB of RAM, that's it. It's very low-resource use. You're about to have 9 containers running `jekyll serve` -- what happens when we get up to 30? Are we resigning ourselves to a perpetual arms race? Near as I can tell the footprint

mstanleyjones:
@thaJeztah can you go ahead and commit your patch here?

johndmulhausen:
This was a valiant effort but I think we had better close this one in favor of PR #941 don't you think, @mstanleyjones ?

PR #915, opened 2016-12-16
improved combined Cloud landing/info pages

sanscontext:
Combines the html grid landing page with the Overview page, redirects from `overview` into the landing page so that info is surfaced on the higher-traffic one. Also cleans up some verbiage, updates some links, condenses the image into the space, and just generally makes the information denser on-page.

@joaofnfernandes and @johndmulhausen ptal if you feel like it!


johndmulhausen:
LGTM :)

PR #716, opened 2016-11-25
Update Compose file reference for version 3

aanand:
Currently a rough draft. This document will only be true once Engine 1.13 and Compose 1.10 are released.

TODO:

- [x] `healthcheck`
- [x] `stop_grace_period`

PR #921, opened 2016-12-16
Update Compose file reference for version 3

dnephin:
Re-doing #716 against vnext-engine

cc @londoncalling

### DIFF
--- a/compose/compose-file.md
+++ b/compose/compose-file.md
@@ -30,10 +30,9 @@ full details.
 
 ## Service configuration reference
 
-> **Note:** There are several versions of the Compose file format – version 1
-> (the legacy format, which does not support volumes or networks) and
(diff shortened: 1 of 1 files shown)


londoncalling:
LGTM @dnephin @aanand , merging

thaJeztah:
This is going to be confusing if vnext-engine is making changes to the _compose_ documentation (given that there's also a vnext-compose branch https://github.com/docker/docker.github.io/tree/vnext-compose)

Which branch is going to be leading?

londoncalling:
@thaJeztah you are right, this should have been merged into `vnext-compose` , we will fix it

londoncalling:
Reverting this PR and cherry-picking the original changes into `vnext-compose` instead.

PR #929, opened 2016-12-19
Corrected command to create db

daudt:


johndmulhausen:
Sorry, we just switched from `rake` to `rails` if you look at the history, based on this recommended practice: http://guides.rubyonrails.org/command_line.html#db

PR #932, opened 2016-12-19
Add a note to `env_file` option

dankeder:


mstanleyjones:
I added a follow-on commit to add some clarification. Thanks for your contribution.

PR #948, opened 2016-12-20
fix the grammer in aws.md

SRaddict:
### Proposed changes
There are some  errors in aws.md.
- change "in same region" to "in the same region"
- change "to Choose a replica..." to "to choose a replica..."
- change "Full documentations is available" to "Full documentations are available"


johndmulhausen:
Thanks, @SRaddict :)

PR #949, opened 2016-12-20
fix punctuation typos

luojie233:
Signed-off-by: Jie Luo <luo612@zju.edu.cn>

### Proposed changes

  I think the doc should in the same style, and the curly quote is not formally in standard English documents.


johndmulhausen:
Thanks, @luojie233 :)

PR #953, opened 2016-12-20
fix the error of handing a and an

SRaddict:
### Proposed changes
change "an highly..." to "a highly..."


johndmulhausen:
Thanks, @SRaddict :)

PR #954, opened 2016-12-20
Update create-swarm.md

sanscontext:
Removes some spaces that were apparently b0rking the code block there.


johndmulhausen:
Thanks, @sanscontext :)

PR #941, opened 2016-12-20
[WIP] Dockerfile per version with nginx as reverse proxy (take2)

thaJeztah:
Pushing as a separate branch for now, as I couldn't get my Git remote setup correctly :blush:

This is still WIP, but adds some modifications to @mstanleyjones's PR https://github.com/docker/docker.github.io/pull/913

- Adding a minimal(ish) base image for the docs to publish
- Updated the Dockerfile and build-script for archive builds
- Updated the nginx-configuration to route /v1.xx/ urls to the right container

To test;

- go to the `_data/docsarchive/` directory
- run `./build_all_archives.sh`
- wait for the builds to complete
- go to the `_data/docsarchive/proxy/` directory
- run `docker-compose up --build -d`
- run `docker-compose logs -f` to watch the logs
- open http://localhost:4000/ in your browser
- open http://localhost:4000/v1.12/ in your browser to view the current docs as well (test if navigation works correctly)
- open http://localhost:4000/v1.4/ for a trip down memory lane

In a terminal;

run `docker images docs`; images should be smaller than previously.Note that 143MB of the size is taken up by the base image (which is shared by all images).

```
REPOSITORY          TAG                 IMAGE ID            CREATED             SIZE
docs                latest              b1cf8892f1ef        About an hour ago   316 MB
docs                v1.12               1544ee4f8a92        About an hour ago   319 MB
docs                v1.11               112c42e65884        About an hour ago   286 MB
docs                v1.10               88b6585ff788        About an hour ago   270 MB
docs                v1.9                bd84e9b071a6        About an hour ago   211 MB
docs                v1.8                cfc0b2e4ea65        About an hour ago   205 MB
docs                v1.7                adde0f58abb6        About an hour ago   202 MB
docs                v1.6                93b6d13fb81c        About an hour ago   172 MB
docs                v1.5                61d4cafff996        About an hour ago   161 MB
docs                v1.4                dee7a0fef8db        About an hour ago   156 MB
docs                base                ef794735ce91        3 days ago          143 MB
```

In a terminal, run

    docker ps --filter label=com.docker.compose.project=proxy --format '{{.Names}}' | xargs docker stats

To view memory and cpu-use of the containers;

```
CONTAINER                    CPU %               MEM USAGE / LIMIT    MEM %               NET I/O             BLOCK I/O           PIDS
proxy_docs-archive-1-6_1     0.00%               2.004 MiB / 25 MiB   8.02%               3.18 kB / 648 B     205 kB / 0 B        2
[...]

PR #971, opened 2016-12-22
Modify Dockerfile to use shallow clones

mstanleyjones:
### Proposed changes

This changes the Dockerfile to use shallow clones to populate and layer the sources from the archives and master. This is an incremental improvement while we experiment with other more broad improvements (#941).

This takes about **2 minutes longer** to build the Dockerfile than the other version, but results in a Dockerfile that is over **300MB smaller** in size. It seems like this would make it faster to deploy.

To test, build the Dockerfile in this branch and build it again in master. Here are my results:

**master**:

```bash
$ time docker build --no-cache -f Dockerfile -t mistything .

real	7m57.189s
user	0m4.683s
sys	0m2.102s

$ docker images |grep mistything

mistything                 latest                                  f2495e274448        About a minute ago   1.9 GB
```

**this PR**:

```bash
$ time docker build --no-cache -f Dockerfile -t mistything .

Successfully built 2a8cf4f91450

real	9m21.299s
user	0m4.576s
sys	0m1.935s

$ docker images | grep mistything

mistything                 latest                                  2a8cf4f91450        12 minutes ago      1.588 GB

```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,36 +2,32 @@ FROM starefossen/github-pages
 
 ENV VERSIONS="v1.4 v1.5 v1.6 v1.7 v1.8 v1.9 v1.10 v1.11"
 
-# Create archive; check out each version, create HTML, tweak links
-RUN git clone https://www.github.com/docker/docker.github.io temp; \
- for VER in $VERSIONS; do \
-		git --git-dir=./temp/.git --work-tree=./temp checkout ${VER} \
-		&& mkdir -p allvbuild/${VER} \
-		&& jekyll build -s temp -d allvbuild/${VER} \
-		&& find allvbuild/${VER} -type f -name '*.html' -print0 | xargs -0 sed -i 's#href="/#href="/'"$VER"'/#g' \
-		&& find allvbuild/${VER} -type f -name '*.html' -print0 | xargs -0 sed -i 's#src="/#src="/'"$VER"'/#g' \
-		&& find allvbuild/${VER} -type f -name '*.html' -print0 | xargs -0 sed -i 's#href="https://docs.docker.com/#href="/'"$VER"'/#g'; \
-	done; \
-	rm -rf temp
-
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -2,36 +2,32 @@ FROM starefossen/github-pages
...
- && find allvbuild/engine/reference -type f -name '*.html' -print0 | xargs -0 sed -i 's#href="https://docs.docker.com/#href="/#g' \
- && find allvbuild/engine/extend -type f -name '*.html' -print0 | xargs -0 sed -i 's#href="https://docs.docker.com/#href="/#g' \
- && rm -rf allv
+RUN git clone --depth 1 --branch master https://www.github.com/docker/docker.github.io allv \
+    && svn co https://github.com/docker/docker/branches/$ENGINE_BRANCH/docs/reference allv/engine/reference \
+    && svn co https://github.com/docker/docker/branches/$ENGINE_BRANCH/docs/extend allv/engine/extend \
+    && wget -O allv/engine/deprecated.md https://raw.githubusercontent.com/docker/docker/$ENGINE_BRANCH/docs/deprecated.md \
+    && svn co https://github.com/docker/distribution/branches/$DISTRIBUTION_BRANCH/docs/spec allv/registry/spec \
+    && wget -O allv/registry/configuration.md https://raw.githubusercontent.com/docker/distribution/$DISTRIBUTION_BRANCH/docs/configuration.md; \
+    for VER in $VERSIONS; do \
```
johndmulhausen: Do an `rm -rf temp` between lines 19 and 20 here so that you don't have detritus between clones

mstanleyjones:
I do it at line 25 instead...

But it turns out that I screwed up by removing the COPY line and I don't gain much on time or size if I put it back, so this PR is probably a non-starter...

johndmulhausen:
I think we realized this is not that beneficial after all; too bad :(

PR #978, opened 2016-12-25
Step 6 of Prereq was not easy for this newb to implement!

Oscar-21:
Hey, this is my first time proposing changes of any sort. So go easy on me if my
proposal is wack, lol.

My issue is with the angle brackets around REPO on line 92.

`$ echo "<REPO>" | sudo tee /etc/apt/sources.list.d/docker.list`

I think the following would be better because it doesn't imply that the user include the angle brackets
in their prompt.

`$ echo "REPO" | sudo tee /etc/apt/sources.list.d/docker.list`

When implementing this line it created a problem for me that was not
fixed until I went into the directory that contained docker.list
and deleted the angle brackets. I think this is a bit more user 
friendly.  Again sorry if I am 100% crazy on this!

### DIFF
--- a/engine/installation/linux/ubuntulinux.md
+++ b/engine/installation/linux/ubuntulinux.md
@@ -89,7 +89,7 @@ To set `APT` to use packages from the Docker repository:
     for the placeholder `<REPO>`.
 
     ```bash
-    $ echo "<REPO>" | sudo tee /etc/apt/sources.list.d/docker.list
+    $ echo "REPO" | sudo tee /etc/apt/sources.list.d/docker.list
     ```
 
 7.  Update the `APT` package index.


mstanleyjones:
Hi there. I'm sorry you had trouble with this step. The thing is the angle brackets are a UNIXy convention that means the thing within them should be replaced by the real value. If we left it the way you suggest instead, most people would echo the literal word "REPO" into that file, which is also not what we want. I actually did explain it just above the command itself. Can you help me come up with a way that sounds more clear but doesn't confuse people who are used to the convention?

josephmosby:
What about <REPO_NAME_HERE>? Won't trip up people who are used to the convention, but also won't trip up people who *aren't* used to the convention.

sanscontext:
@josephmosby that sounds good. Or maybe a variant `<YOUR_REPO_NAME>`?

Oscar-21:
@mstanleyjones You did actually explicitly state that the user needs to replace the placeholder `REPO`  I figured it was a convention. I like the suggestion by @sanscontext.
 
Why, though, is it a useful convention to add characters to a placeholder that will not be used in the actual command? These are almost always stumbling blocks for me, and probably many others.

Imagine a newbie asks two questions:

> 1.How  do I open a text file in **bash** using **nano**?

>2. How do I open a text file in **bash** using **vim**?

The answer that they receive to the first question is:
`nano file.txt`

And the second is:
`vim <file.txt.>`

I think the first method has greater clarity, because, in the second example, the person providing the example would have to explicitly state that they replace the entire placeholder `<file.txt>` with the name of the file, while this is not necessary in the first.

mstanleyjones:
I still think newbies will type literally `nano file.txt` and get a file called `file.txt` instead of the intended result. That's why the `< >` formulation is used to enclose what you should replace. This is taken from the syntax on man pages and other docs from time immemorial. Here are some examples of who uses this:

- Microsoft: https://technet.microsoft.com/en-us/library/dd632565.aspx

- IBM (sometimes): http://www.ibm.com/support/knowledgecenter/SSDKKW_6.2.0/com.ibm.wpg.entadv.doc/config/integration/typographic_conventions.html (but IBM contradicts itself, see below)

- ODIN: http://www.odin.com/r/docs/poa/poa-localization-guide-2_9/index.htm?fileName=13524.htm

- Blackboard: http://library.blackboard.com/ref/8fb6da4a-0569-4d9e-89f3-c32ec8370cd9/content/moderators_and_participants_guides/using_this_guide/conventions_in_this_guide.htm

- Tizen: https://source.tizen.org/documentation/developer-guide/getting-started-guide/typographic-conventions

Some others just use italics within the codeblock to show that something is user-relaceable, but this is VERY subtle indeed:

- Oracle: http://docs.oracle.com/cd/E19340-01/820-6424/6nhokb1gq/index.html

- Salesforce: https://developer.salesforce.com/docs/atlas.en-us.pages.meta/pages/apex_qs_conventions.htm

- IBM (sometimes): http://www.ibm.com/support/knowledgecenter/SSPK3V_7.0.0/com.ibm.swg.im.soliddb.common.doc/doc/typographic.html

- AppDynamics: https://docs.appdynamics.com/display/PRO42/ADQL+Typographical+Conventions+and+Symbols

- O'Reilly: http://oreillymedia.github.io/production-resources/styleguide/

Some use bold:

- Amazon: http://docs.aws.amazon.com/general/latest/gr/docconventions.html

I even found examples of bold italic:

- NOAA: http://ferret.pmel.noaa.gov/LAS/documentation/typographical-conventions

I am sorry for your confusion and I'm glad you figured it out. Typographical conventions ultimately come down to a choice, and it's clear from the above that lots of different documentation sets end up using lots of different ways to indicate the same thing. 

cc/ @johndmulhausen

Oscar-21:
@mstanleyjones  Thanks for those links! It'll definitely help me to learn to read documentation  more effectively.

johndmulhausen:
@Oscar-21 Thanks for the PR; I have to concur with @mstanleyjones on this one and close this, but I hope you send us edits in the future for anything that looks out of whack

PR #979, opened 2016-12-25
Update index.md

sn0opr:

### DIFF
--- a/docker-hub/index.md
+++ b/docker-hub/index.md
@@ -19,7 +19,7 @@ Log in to Docker Hub and Docker Cloud using [your free Docker ID](accounts.md).
 
 Docker Hub provides the following major features:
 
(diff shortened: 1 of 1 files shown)


sanscontext:
Hey @sn0opr thanks for the feedback! Technically this is correct as written, because the line combines the verbs "push and pull" to apply to `community, official, and private image libraries` . If you were writing this as an ordered list, it'd be:

- find
- manage
- push and pull

Because those last two can be logically grouped together, where the previous two are logically separate. 

I do agree that this is a pretty awkward and unwieldy sentence though. Would you try splitting it into two sentences here?

sanscontext:
Hey @sn0opr this needs rebase and has been sitting around awhile. I'm going to close this out and make a different change to address your feedback.

PR #1006, opened 2016-12-30
Updated to remove 'json' from parsed output

rbramwell:
The use of the ```json MD formatting erroneously displays the word 'json' before the expected json. This results in the highlighted command as 'json { "dns": ["8.8.8.8", "8.8.4.4"]}' instead of '{ "dns": ["8.8.8.8", "8.8.4.4"]}'. Using ```bash resolves this issue.

### DIFF
--- a/engine/installation/linux/ubuntulinux.md
+++ b/engine/installation/linux/ubuntulinux.md
@@ -509,7 +509,7 @@ at `/etc/docker/daemon.json`.
 
 2.  Add a `dns` key with one or more IP addresses as values. If the file has
     existing contents, you only need to add or edit the `dns` line.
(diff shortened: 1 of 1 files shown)


rbramwell:
See step 3. Specifying DNS servers for Docker @ https://docs.docker.com/engine/installation/linux/ubuntulinux/#/configure-a-dns-server-for-use-by-docker for the output. This does not display incorrectly in Github's markdown but does on the website.

johndmulhausen:
I think I see what's wrong; we need a space between the code example and the text. JSON is the correct lexer, however, so I must shut this PR down, I'm sorry.

PR #1007, opened 2016-12-30
Update overview.md

NickvdMeij:
### Proposed changes

Update overview.md to reference possible issues regarding windows to unix path convert issues

### Related issues (optional)

Reference for docker github repo: [link](https://github.com/docker/docker/issues/29753)

### DIFF
--- a/compose/overview.md
+++ b/compose/overview.md
@@ -95,6 +95,8 @@ runs, if it finds any containers from previous runs, it copies the volumes from
 the old container to the new container. This process ensures that any data
 you've created in volumes isn't lost.
 
+If you use `docker-compose` on a Windows machine, See [Environment variables](reference/envvars.md) and adjust the nessecary environment variables for you specific needs
+
 
 ### Only recreate containers that have changed
 


joaofnfernandes:
Thanks for contributing 🐳
I understand that you're trying to include information so that Windows users don't get caught by surprise with the differences in environment variables, but I don't think we should do that in the Overview article. That article should explain what is Compose and why you should care. It should not go into nitty-gritty details, that can confuse new users.

Maybe what we need is a separate article explaining all the caveats of running Compose on Windows. @londoncalling what do you think?

londoncalling:
Thanks @joaofnfernandes .

@NickvdMeij I'll either merge this or find a better way of incorporating it. I looked at the [link](https://github.com/docker/docker/issues/29753) you provided, and see it definitely needs to be addressed. I'm working on Compose docs today, so I'll look at this as a part of that work. Thanks! More soon.

londoncalling:
@NickvdMeij @thaJeztah @joaofnfernandes given the background on this per the [discussion thread in this Compose issue](https://github.com/docker/docker/issues/29753), I'm merging this. Will make a couple of minor corrections and look to see if there is elsewhere we can add it as well. Sorry you had to spend 10 hours on this @NickvdMeij . Thanks for contributing to save someone else that struggle.

NickvdMeij:
@londoncalling No problem, I love docker and I dont mind giving feedback and enhancing the documentation ;-) that's the power of open source

PR #1008, opened 2016-12-30
Update envvars.md

NickvdMeij:
### Proposed changes

Update envvars.md to reflect the supported values of COMPOSE_CONVERT_WINDOWS_PATHS

### Related issues

Addition documentation for people using windows. See [link](https://github.com/docker/docker/issues/29753)

### DIFF
--- a/compose/reference/envvars.md
+++ b/compose/reference/envvars.md
@@ -82,7 +82,8 @@ Supported values are: `TLSv1`, `TLSv1_1`, `TLSv1_2`.
 ## COMPOSE\_CONVERT\_WINDOWS\_PATHS
 
 Enable path conversion from Windows-style to Unix-style in volume definitions.
-Users of Docker Machine and Docker Toolbox on Windows should always set this.
+Users of Docker Machine and Docker Toolbox on Windows should always set this. Defaults to `0`
+Supported values: `0`, `1`
 
 
 ## Related Information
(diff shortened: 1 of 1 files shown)


shin-:
Actually as far as the current release is concerned, the only recognized values are whether the variable is set or unset. I intend to fix this in 1.10 but until then this should not be merged.

johndmulhausen:
@shin- Could this be merged if it were opened against vnext-compose?

shin-:
@johndmulhausen Yes, it would be fine (although we will also support `true` and `false` notably)

NickvdMeij:
@shin- i would say `true` and `false` OR `1` and `0` should be supported. That way its always clear what the value is (if its set). I can imagine a situation where you might want to toggle this off for some projects, and toggle it back on for other projects. Doing this now would mean you need to remove the whole environment variable and create it again when you need it, instead of just changing the value

mstanleyjones:
Please rebase your work on `vnext-compose` and resubmit.

shin-:
@NickvdMeij 'false' (i-case), '0', unset and empty values will all be recognized as falsy, with anything else being recognized as true.

johndmulhausen:
Edited to make more accurate according to @shin- 's comment. Now that `vnext-compose` is merged, I'm merging. Thanks, @NickvdMeij for the PR and @thaJeztah and @mstanleyjones for the reviews :)

PR #1016, opened 2017-01-03
Correct wrong default value

mastertinner:
### Proposed changes

I think the default value for `v4auth` is actually `true`, not `false`.


mstanleyjones:
@dmcgowan can you comment on this? I looked through the code but couldn't figure it out. I think the file is https://github.com/docker/distribution/blob/96f63cbe7f9d06f213289df97b6bd3102cf65903/registry/storage/driver/s3-aws/s3.go.

dmcgowan:
This change looks correct. LGTM

PR #1021, opened 2017-01-04
remove highlight.js

thaJeztah:
Code is already highlighted through "rouge", so
enabling highlight.js only resulted in code
being parsed/highlighted _twice_.

This patch removes highlight.js and updates
the "github" stylesheet to be compatible
with rouge.

/cc @joaofnfernandes @mstanleyjones @johndmulhausen


joaofnfernandes:
Datacenter looks good. @sanscontext where's that page with all possible markdown permutations you've created? Is that page OK?

sanscontext:
@joaofnfernandes https://deploy-preview-1021--psychologist-duck-16578.netlify.com/test/#/some-code-block-samples

Looks pretty good, but looks like YAML changed how it's broken, and it might not have `dockerfile`?

joaofnfernandes:
It seems that Rouge has support for Docker files in their latest version (2.0.7), but
we're importing Rouge through github-pages, which in turn imports Rouge 1.11.1.
https://github.com/jneen/rouge/blob/master/lib/rouge/lexers/docker.rb

johndmulhausen:
While I'd love to do this, Jekyll does not support the Rouge version where Dockerfile support was added (which was 2.0.4). As you see here, it's dependency is set to `~>1.7`: https://rubygems.org/gems/jekyll

```
$ bundle update
Fetching gem metadata from https://rubygems.org/...........
Fetching version metadata from https://rubygems.org/..
Fetching dependency metadata from https://rubygems.org/.
Resolving dependencies...
Bundler could not find compatible versions for gem "rouge":
  In Gemfile:
    rouge (~> 2.0)

    jekyll (~> 3.3) was resolved to 3.3.0, which depends on
      rouge (~> 1.7)
```

It's not a GitHub Pages thing; no idea when Jekyll will get their act together.

johndmulhausen:
Trying this Gemfile:

```
source "https://rubygems.org"

gem "rouge","2.0.7"
gem "jekyll-seo-tag"
gem "kramdown"
gem "jekyll-redirect-from"
gem "jekyll","3.3.1"
```

You get the incompatibility error when running `bundle update`

Trying this: 

```
source "https://rubygems.org"

gem "rouge","2.0.7"
gem "jekyll-seo-tag"
gem "kramdown"
gem "jekyll-redirect-from"
gem "jekyll"
```

Succeeds but bundler reverts way back to Jekyll 2.5.3, probably because that was before it had dependency checking for Rouge: https://rubygems.org/gems/jekyll/versions/2.5.3

sanscontext:
In the mean time we can all watch this PR: https://github.com/jekyll/jekyll/issues/5556

johndmulhausen:
Closing in favor of #2757. We can use `conf` instead of native Dockerfile support for now.

PR #1024, opened 2017-01-04
Add notices for ignored and unsupported options when deploying a stack

Thanaton:
### Proposed changes

Add notices to the Compose file options that are ignored or unsupported when deploying a stack in swarm mode.

I was quite surprised when I tried to deploy a stack in swarm mode with Docker 1.13-rc4 and some options were unsupported. As these options won't be added before the Docker 1.13 release (see related issue), we can at least document them to prevent surprises for others.

Docker 1.13 (using the new Compose version 3 format for deploying stacks in swarm mode)

### Related issues

See https://github.com/docker/docker/issues/29685

### DIFF
--- a/compose/compose-file.md
+++ b/compose/compose-file.md
@@ -504,6 +504,11 @@ returns an error if it encounters one.
 For more on `extends`, see the
 [the extends documentation](extends.md#extending-services).
 
+> **Note:** This option is not yet supported when
+> [deploying a stack in swarm mode](/engine/reference/commandline/stack_deploy.md)
+> with a (version 3) Compose file. Use `docker-compose config` to generate a
+> configuration with all `extends` options resolved, and deploy from that.
+
 ### external_links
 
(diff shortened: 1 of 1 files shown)


mstanleyjones:
@londoncalling @johndmulhausen PTAL, relates to compose.

londoncalling:
@ManoMarks @nathanleclaire @shin- can you all review these changes re: Compose v3 and let me know if they are all correct?

londoncalling:
@ManoMarks I just checked, and although there is a discussion of `docker stack deploy`, no there is no flag on `build` that I can see. I'll merge this and add that. Thanks!

londoncalling:
LGTM, merging, then updating for the qualification on `build` per @ManoMarks  comment above.

londoncalling:
@MagnusS , the flag on the `build` was in there! Thanks!

PR #1044, opened 2017-01-06
docker upgrade command fix

protometa:
`apt-get upgrade` doesn't take an argument. To upgrade docker specifically would require `apt-get install docker-engine`.


johndmulhausen:
Thanks @protometa :)

PR #1065, opened 2017-01-09
Add info about image resolution with content trust

### LINKED ISSUE #1004: Document content trust for swarm services
### Problem description

Document feature added in https://github.com/docker/docker/pull/29156/

`--with-registry-auth`

Engine 1.13

mstanleyjones:
### Proposed changes

Added info about how image resolution works when content trust is enabled

Engine 1.13

Fixes #1004

### DIFF
--- a/engine/swarm/services.md
+++ b/engine/swarm/services.md
@@ -78,14 +78,17 @@ flag. For more information, see
 ### Specify the image version the service should use
 
 When you create a service without specifying any details about the version of
-the image to use, the service uses the latest version available in Docker Hub
-or your registry. You can force the service to use a specific version of
-the image in a few different ways, depending on your desired outcome.
+the image to use, the service uses the version tagged with the `latest` tag.
+You can force the service to use a specific version of the image in a few
+different ways, depending on your desired outcome.
 
 An image version can be expressed in several different ways:
 
-- If you specify a tag, the manager resolves that tag to a digest, and directs
-  workers to uses that version.
+- If you specify a tag, the manager (or the Docker client, if you use
+  [content trust](#image_resolution_with_trust)) resolves that tag to a digest.
+  When the request to create a container task is received on a worker node, the
+  worker node only sees the digest, not the tag.
+
   ```bash
   $ docker service create --name="myservice" ubuntu:16.04
   ```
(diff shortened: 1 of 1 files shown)


cyli:
This LGTM - cc @aaronlehmann who added this fix.

engine/swarm/services.md
```diff
@@ -80,8 +80,10 @@ flag. For more information, see
 
 When you create a service without specifying any details about the version of
 the image to use, the service uses the latest version available in Docker Hub
-or your registry. You can force the service to use a specific version of
-the image in a few different ways, depending on your desired outcome.
+or your registry, or using Notary if content trust is enabled. See the
+[note about content trust](#image_resolution_with_trust).You can force the
```
aaronlehmann: I think the sentence about creating a service is becoming long, and mixes different concepts.

The original sentence was saying that if you omit a tag, it will be treated as `<image>:latest`. Instead of saying "the latest version", I'd change this to say "the `latest` tag", since `:latest` may not actually be the most recent version.

We also want to say that if content trust is enabled, Notary is used to authenticate the tags. This happens whether you specify a custom tag, or use the default of `latest`.

There's a missing period after the link to the note.

mstanleyjones: Addressed the feedback by removing the links to the note, and just having the note at the bottom. I think it's more clear that way. Let me know what you think.

engine/swarm/services.md
```diff
@@ -79,14 +79,16 @@ flag. For more information, see
...
+the image to use, the service uses the version tagged with the `latest` tag.
+You can force the service to use a specific version of the image in a few
+different ways, depending on your desired outcome.
 
 An image version can be expressed in several different ways:
 
-- If you specify a tag, the manager resolves that tag to a digest, and directs
-  workers to uses that version.
+- If you specify a tag, the manager resolves that tag to a digest. If you are
+  using [content trust](#image_resolution_with_trust), Notary verifies that the
```
cyli: Maybe we should say "the docker client" instead of "Notary"?  Or "the docker client uses notary to verify..."?  It both verifies that it is signed and also resolves the tag to a digest.  The manager does not resolve the tag in this case (if content trust is enabled).

Also, wondering if we should link to https://docs.docker.com/engine/security/trust/content_trust/ when saying "signed" so users can look up more detail about how that works?

cyli: Update:  Ah right, sorry, that note goes with https://github.com/docker/docker.github.io/pull/1065/files#diff-77d91a1ccd7779ca36fd0f5a1cfd5353R127.  

Hm... maybe that sentence can say something like:  "When you create a service, if you don't specify a digest directly, the image's tag gets resolved to the specific digest it points to **at the time of service creation**..." ?  Thus avoiding the manager vs client resolution in that sentence.

mstanleyjones:
@cyli can you give this another look-over? I think I addressed your feedback.

PR #1079, opened 2017-01-10
Fix root_ca, json structure in Notary client config docs

riyazdf:
The JSON key should be `"root_ca"` instead of `"root-ca"`, and a comma was missing.  Ports over https://github.com/docker/notary/pull/1080

Signed-off-by: Riyaz Faizullabhoy <riyaz.faizullabhoy@docker.com>


johndmulhausen:
Thanks, @riyazdf :) LGTM

PR #1080, opened 2017-01-11
Add cs-engine 1.12.6-cs6 release notes

cpuguy83:
ping @joaofnfernandes


joaofnfernandes:
@aduermael how can I retry the Jenkins job? It seems that the problem was an HTTP 502 when doing an svn checkout during the build. I've seen Github having problems with svn checkouts in the past, and usually retrying the build would fix it.

PR #1082, opened 2017-01-11
Update get-started-overlay.md

glensc:
attempt to fix renderer glitches in docs

https://docs.docker.com/engine/userguide/networking/get-started-overlay/


johndmulhausen:
@glensc Had to push a removal of docs.docker.com so your PR would pass the Jenkins test, but we're good now. Thanks!

glensc:
@johndmulhausen wow. how you were able push to PR wasn't created by you?

https://github.com/docker/docker.github.io/pull/1082/commits

johndmulhausen:
@glensc I think anyone with push access to the source repo can also push to forks. GitHub security model is pretty weird that way. :man_shrugging:

PR #1091, opened 2017-01-11
Document the ability to publish a port locally on a swarm node

### LINKED ISSUE #643: Document ability to expose Swarm service ports without going through the routing mesh
### Problem description

Today you have:

node-1:80 -> routing mesh -> proxy -> overlay network -> nginx/php-fpm 

By exposing the port locally, you would have:

node-1:80 -> proxy -> overlay network -> nginx/php-fpm 

Which would result in one less hop.

cc/ @aluzzardi

mstanleyjones:
### Proposed changes

Document https://github.com/docker/swarmkit/commit/dfa1fc0bb99b95339e2988fe4c1e306179a73d77

Engine 1.13

### Related issues (optional)

Fixes #643

### DIFF
--- a/engine/swarm/services.md
+++ b/engine/swarm/services.md
@@ -53,8 +53,7 @@ anixjtol6wdf  my_web  1/1       nginx
 ```
 
 To make the web server accessible from outside the swarm, you need to
-[publish the port](#publish-ports-externally-to-the-swarm) where the swarm
-listens for web requests.
+[publish the port](#publish-ports) where the swarm listens for web requests.
 
 You can include a command to run inside containers after the image:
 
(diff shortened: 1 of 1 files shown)


engine/swarm/services.md
```diff
@@ -264,24 +264,50 @@ Swarm mode lets you network services in a couple of ways:
...
 
-For example, imagine you want to deploy a 3-replica nginx service to a 10-node
-swarm as follows:
+- [You can rely on the routing mesh](#publish-a services-ports-using-the-routing-mesh).
+  When you publish a service port, the swarm makes the service accessible at the
+  target port on every node, regardless of whether there is a task for the
+  service running on that node or not. This is less complex but creates some
+  additional networking overhead.
+
+- [You can publish a service task's port directly on the swarm node](#publish-a-services-ports-directly-on-the-swarm-node)
```
dongluochen: `node` -> `nodes`

engine/swarm/services.md
```diff
@@ -264,24 +264,50 @@ Swarm mode lets you network services in a couple of ways:
...
+  where that service is running. This feature is available in Docker 1.13 and
+  higher. This creates less networking overhead by passing the routing mesh, but
+  you are responsible for routing requests to your service's nodes, and you can
+  only run one task for a given service on a given swarm node.
```
stevvooe: This restriction is only in place if you specify the published port. If you don't specify it, an ephemeral port will be allocated on the node.

engine/swarm/services.md
```diff
@@ -264,24 +264,50 @@ Swarm mode lets you network services in a couple of ways:
...
+                        --replicas 3 \
+                        --publish 8080:80 \
+                        nginx
 ```
 
-The scheduler will deploy nginx tasks to a maximum of 3 nodes. However, the
-swarm makes nginx port 80 from the task container accessible at port 8080 on any
-node in the swarm. You can direct `curl` at port 8080 of any node in the swarm
-to access the web server:
+Three tasks will run on up to three nodes. Multiple tasks can run on the same
```
dongluochen: > Multiple tasks can run on the same node.

While this is possible, Swarmkit scheduler tries to spread the tasks to different nodes as long as it can. This sentence may be a little misleading.

engine/swarm/services.md
```diff
@@ -290,29 +316,47 @@ $ curl localhost:8080
...
+...truncated...
 </html>
 ```
 
+Subsequent connections may be routed to the same swarm node or a different one.
+
+#### Publish a service's ports directly on the swarm node
+
+Using the routing mesh may not be the right choice for your application if you
+need to minimize network overhead or your application topology requires that
```
stevvooe: Maybe break this down into bullet points.

engine/swarm/services.md
mstanleyjones: The intent was to make it clear that you don't know (and shouldn't need to care) where the 3 tasks are run across the 10 nodes. It could be that 8 of the nodes are down for whatever reason. The routing mesh will do the right thing for you. I'll try to make it more clear.

dongluochen:
LGTM

mstanleyjones:
I think this is good to go, merging it.

Issue #1068, opened 2017-01-09
Broken links


aduermael:
### Problem description

I have a pending PR ( #1052 )  to check all links to local resources. But now I need all broken links to be fixed in order to merge it. There are a lot, and in many cases, I personally don't know how to fix them. 

Can we please divide it up and get this done as soon as possible?

Thank you!

```
html_test.go:55: /allpagelinks/index.html
	broken: <a href="404.md">
html_test.go:55: /apidocs/overview/index.html
	broken: <a href="apidocs/overview.md" class="button darkblue-btn nomunge" style="color:#FFFFFF; width:100%; margin: 0px;">
```

**EDIT**

#1052 has been merged. But then changes for `1.13` have been merged, without running the tests. So we now have new broken links. I listed them in [that spreadsheet](https://docs.google.com/spreadsheets/d/1y_svpzGHNkZmuQJjzrTzq7TC0a72S-iuj7NDxo3dQXc/edit?usp=sharing)

aduermael:
cc @johndmulhausen

sanscontext:
For the Docker Cloud API docs ones, if you navigate to the page it works: 

https://docs.docker.com/apidocs/docker-cloud/#service goes to https://docs.docker.com/docker-cloud/apps/service-scaling/ which is a valid page

I *think* that what’s going on is the link checker is looking at files that are generated by the build, but not actually linked anywhere. I believe the `make release` build process there builds a “single page” layout, and then a series of one-pagers per topic, but AFAICT we don't actually *use* the single-pagers anywhere. They're broken because it generates them into a subfolder, but from the same source files as those that generate the working ones (and which don't trigger the link checker).

aduermael:
Thanks @joaofnfernandes for your commit! 
I updated the list (~15% of it has been fixed)

aduermael:
@sanscontext i don't understand your comment. 
Can you point me to specific lines that shouldn't appear?

aduermael:
@sanscontext I fixed links in `/apidocs/docker-cloud/includes`: #1087 (waiting for CI to pass)

sanscontext:
@aduermael  Those files are programmatically generated. Did you fix them and see if running `make release` in that dir clobbers your changes?

sanscontext:
@aduermael Please take a look at https://github.com/docker/docker.github.io/blob/master/apidocs/cloud-api-source/README.md for more background.

aduermael:
[...]

PR #1093, opened 2017-01-11
exclude /apidocs/cloud-api-source when building

aduermael:
### Proposed changes

Excluding `/apidocs/cloud-api-source` when building the website.
Contributes to #1068.


sanscontext:
SGTM!

PR #1100, opened 2017-01-12
docker-for-mac: add a FAQ about reducing the qcow2 size

djs55:
### Proposed changes

Add a FAQ about reducing the size of the qcow2 on Docker for Mac.

In Docker for Mac 1.12 the only way to free space on the host is to delete the qcow2 which means all containers and images have to be rebuilt.

In Docker for Mac 1.13 there is preliminary support for shrinking the qcow2 file non-destructively using "TRIM" (as also used on SSDs). Unfortunately this isn't (yet) fully automatic -- it runs in the background and requires the app to be occasionally restarted.

In future versions I'm hoping to make this happen automatically in the background.

### Related issues

See [docker/for-mac#371]

### DIFF
--- a/docker-for-mac/faqs.md
+++ b/docker-for-mac/faqs.md
@@ -136,18 +136,15 @@ For a complete explanation of how to do this, see the blog post [Adding Self-sig
 to Docker & Docker for
 Mac](http://container-solutions.com/adding-self-signed-registry-certs-docker-mac/).
 
-### What are system requirements for Docker for Mac?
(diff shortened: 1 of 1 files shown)


londoncalling:
@djs55 looks good, I will get this in today, along with updates for the Beta's release notes. I might  go ahead and add a link to the issue within the FAQ if that's okay.

londoncalling:
LGTM, merging!

PR #1102, opened 2017-01-12
Correction of the Install-Module command

Clorr:
The command line was Import-Module which seems to be wrong. It seemed to work for me with Install-Module. So I make the correction. Please note, I'm only a newbie, so check this twice please ;-)


johndmulhausen:
It should probably be `Install-Module` which is what we use elsewhere as well. Thanks, @Clorr :)

PR #1108, opened 2017-01-12
Add new redirect for DTR install

joaofnfernandes:


joaofnfernandes:
Jenkins is borked, but nothing to do with this change.

PR #1111, opened 2017-01-12
Document the whitelist filter of the swarm scheduler

alexmavr:
Signed-off-by: Alex Mavrogiannis <alex.mavrogiannis@docker.com>

### Proposed changes

This PR documents the new whitelist scheduler filter for Docker Swarm.

Implemented at https://github.com/docker/swarm/pull/2551

This change applies to Docker Swarm 1.2.6 which is currently unreleased.

### DIFF
--- a/swarm/scheduler/filter.md
+++ b/swarm/scheduler/filter.md
@@ -21,6 +21,7 @@ Each filter has a name that identifies it. The node filters are:
 * `constraint`
 * `health`
 * `containerslots`
+* `whitelist`
 
 The container configuration filters are:
 
@@ -42,10 +43,10 @@ $ swarm manage --filter=health --filter=dependency
 
 ## Node filters
 
-When creating a container or building an image, you use a `constraint` or
-`health` filter to select a subset of nodes to consider for scheduling.
-If a node in Swarm cluster has a label with key `containerslots`
-and a number-value, Swarm will not launch more containers than the given number.
+When creating a container or building an image, you can use a `constraint`,
+`whitelist`, or `health` filter to select a subset of nodes to consider for
+scheduling. If a node in a swarm has a label with key `containerslots`
+and an integer, Swarm will only launch that integer number of containers.
 
 ### Use a constraint filter
 
@@ -72,7 +73,7 @@ practical applications:
 
 * Schedule based on specific host properties, for example,`storage=ssd` schedules
   containers on specific hardware.
-* Force containers to run in a given location, for example region=us-east`.
+* Force containers to run on nodes with a given metadata value.
 * Create logical cluster partitions by splitting a cluster into
   sub-clusters with different properties, for example `environment=production`.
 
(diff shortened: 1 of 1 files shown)


alexmavr:
\cc @mstanleyjones

swarm/scheduler/filter.md
```diff
@@ -42,9 +43,9 @@ $ swarm manage --filter=health --filter=dependency
 
 ## Node filters
 
-When creating a container or building an image, you use a `constraint` or
-`health` filter to select a subset of nodes to consider for scheduling.
-If a node in Swarm cluster has a label with key `containerslots`
+When creating a container or building an image, you use a `constraint`,
+`whitelist` or `health` filter to select a subset of nodes to consider for
+scheduling.  If a node in Swarm cluster has a label with key `containerslots`
```
mstanleyjones: Only a single space between sentences, please.

"If a node in a swarm"... missing 'a', no need to capitalize, no need to say both "swarm" and "cluster".

swarm/scheduler/filter.md
```diff
@@ -42,9 +43,9 @@ $ swarm manage --filter=health --filter=dependency
...
 and a number-value, Swarm will not launch more containers than the given number.
```
mstanleyjones: Maybe "Swarm will only launch that number-value of containers."

Is `number-value` the literal parameter? Should we say `integer` instead of `number-value`?

swarm/scheduler/filter.md
```diff
@@ -72,7 +73,7 @@ practical applications:
 
 * Schedule based on specific host properties, for example,`storage=ssd` schedules
   containers on specific hardware.
-* Force containers to run in a given location, for example region=us-east`.
+* Force containers to run in a given location, for example `region=us-east`.
```
mstanleyjones: Maybe "Force containers to run on nodes with a given metadata value"?

swarm/scheduler/filter.md
```diff
@@ -166,6 +167,18 @@ ouruser/sinatra     v2                  cd70495a1514        35 seconds ago
 ubuntu              14.04               a5a467fddcb8        11 days ago         187.9 MB
 ```
 
+### Use the whitelist filter
+
+The node `whitelist` filter directs the scheduler to only consider a specific
```
mstanleyjones: Maybe "configures the scheduler to only schedule tasks to run on a given list of nodes."

swarm/scheduler/filter.md
```diff
@@ -166,6 +167,18 @@ ouruser/sinatra     v2                  cd70495a1514        35 seconds ago
...
+list of nodes for running containers on. The nodes are specified as the value to
```
mstanleyjones: Maybe "The whitelist is comprised of a list of node keys separated by the `|` character."

swarm/scheduler/filter.md
```diff
@@ -166,6 +167,18 @@ ouruser/sinatra     v2                  cd70495a1514        35 seconds ago
...
+the `node` key and separated using the `|` character. For example, the following
+command instructs the scheduler to deploy a mysql container on one of three
```
mstanleyjones: deploy a `mysql` continer

swarm/scheduler/filter.md
alexmavr: I would opt against using the term "tasks", as it is a swarm-mode construct which does not apply in Docker Swarm

alexmavr:
Addressed comments in the same commit. The whitelist feature will only be present in Docker Swarm and not in swarm-mode.

@mstanleyjones PTAL

swarm/scheduler/filter.md
```diff
@@ -72,7 +73,7 @@ practical applications:
 
 * Schedule based on specific host properties, for example,`storage=ssd` schedules
   containers on specific hardware.
-* Force containers to run in a given location, for example region=us-east`.
+* Force containers to run on nodes in a given location, for example `region=us-east`.
```
mstanleyjones: If I understand it right, then `us-east` is only a label, and does not in and of itself give you a location, unless you have labeled your nodes that way. That's what I was trying to convey when I asked you to change it to 'label'. I don't want people to think that it's magic. Does that make sense or am I missing something?

alexmavr:
Addressed newer comment in the same commit

PR #1112, opened 2017-01-12
Jenkinsfile: avoid container collisions (unique tags)

aduermael:
### Proposed changes

<s>Added one command in the Jenkinsfile to remove all Docker containers that might be present before running the tests. (because we've had issues where the container name was already taken)
`sh "docker rm -fv $(docker ps -a -q)"`</s>

Now tagging images and containers with JOB_BASE_NAME and BUILD_NUMBER to avoid collisions.

cc @joaofnfernandes

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -12,11 +12,11 @@ wrappedNode(label: 'linux && x86_64') {
     are not generated when building with Jekyll. */
   sh "awk '/jekyll-redirect-from/{n=1}; n {n--; next}; 1' < _config.yml > _config.yml.tmp"
   sh "mv _config.yml.tmp _config.yml"
-  
-  sh "docker build -t docs `pwd`"
-  sh "docker build -t tests `pwd`/tests"
-  sh "docker run -v /usr/src/app/allvbuild --name docs docs /bin/true"
-  sh "docker run --rm --volumes-from docs -v `pwd`:/docs tests"
-  sh "docker rm -fv docs"
-  sh "docker rmi docs tests"
+
+  sh "docker build -t docs:${JOB_BASE_NAME}-${BUILD_NUMBER} `pwd`"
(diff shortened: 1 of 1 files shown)


joaofnfernandes:
I don't have much experience with Jenkins, but this might cause parallel jobs running on the same machine to interfere with one another.
Example

1. Build 1 starts serving the docs container and starts running the tests
2. While this is happening, build 2 is scheduled to the same machine. It removes the docs container and proceeds to create a new one and run the tests
3. Build 1 fails because there is no container serving the docs. Depending on how things get scheduled it might also happen that the tests on build 1 are running against docs from build 2.

Not sure if this is a real problem or not. If it is, we can consider tagging the docs container as `docs-<commit-hash>`.

```
container_name=docs-$(git rev-parse HEAD)
```

aduermael:
@joaofnfernandes oh... I don't have much experience with Jenkins either. I thought each build was made in its own clean environment. I work too much with Docker... 😋
You're right I'll do that differently, using a unique tag for the docs container.

sanscontext:
@aduermael +1 giving each one a PR/build-specific container name would probably both fix the current issue we're having and also make cleanup easier if container delete fails afterwards somehow. :)

aduermael:
@sanscontext ok, I'll see what I can do!

joaofnfernandes:
Looks good to me!

sanscontext:
LGTM, but like @joaofnfernandes I don't have a lot of experience with Jenkins. Should we call @mikedougherty for a look?

aduermael:
@sanscontext I ended up using environment variables that are available in the build context, nothing fancy. Just waiting for tests to pass, then it can be merged.

aduermael:
🙄
```
Could not update commit status. Message: GitHub API rate limit exceeded

Finished: SUCCESS
```

johndmulhausen:
Is all of Docker using one API key? Might be at the point where we need to break it up. :/

aduermael:
Oh! All checks have passed! 
Can someone merge it please?

PR #1136, opened 2017-01-16
remove leading zeros

stkevintan:
### Proposed changes
short the css properties by removing the useless leading zeros.


joaofnfernandes:
Thanks for trying to make our docs render faster! We know that is is an issue for people with slower internet access.
But your change won't significantly change the performance of the site, and should be done during our build process by a CSS minifier, to ensure it's consistently applied across the board.

For that reason I'm closing this PR.

PR #1137, opened 2017-01-16
add in a reboot

bladster:

### DIFF
--- a/engine/installation/linux/centos.md
+++ b/engine/installation/linux/centos.md
@@ -68,11 +68,12 @@ which also installs via the `yum` package manager.
     $ sudo yum install docker-engine
     ```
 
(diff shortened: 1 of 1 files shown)


joaofnfernandes:
@thaJeztah can you confirm that you need to reboot when installing Docker on CentOS?

thaJeztah:
I don't think a reboot should be needed; @bladster what issue did you have without doing a reboot?

johndmulhausen:
This is not necessary for any distro, and there formatting problems with this PR. At most you may need to reload your equivalent of `~/.bash_profile` so that new aliases etc get picked up. Suggest we close.

mstanleyjones:
Closed due to the concerns of SMEs.

PR #1141, opened 2017-01-17
Correct Python application port for local browsing

patoobe:


joaofnfernandes:
We we're telling users to map port 80 to 5000, and then send them to browse `localhost:49155` 😂
Thanks for fixing our docs 🐳

PR #1052, opened 2017-01-06
CI - added tests for relative links [DO NOT MERGE]

aduermael:
### Proposed changes

New tests to make sure we don't have broken relative links.

PR #1143, opened 2017-01-17
Updates allpagelinks.md to use full paths

joaofnfernandes:
@aduermael I think this will fix the last broken link reported by #1068.
Can you run your tests on top of this PR?

### DIFF
--- a/allpagelinks.md
+++ b/allpagelinks.md
@@ -3,7 +3,7 @@ title: All Page Links
 hide_from_sitemap: true
 ---
 
-{% assign sorted_pages = site.pages | sort:"path" %}
-{% for thispage in sorted_pages %}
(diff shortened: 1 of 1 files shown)


johndmulhausen:
@joaofnfernandes It's a harmless fix; I think if we merge this that his check will work. Let's do it!

aduermael:
@joaofnfernandes I rebased #1052, but I don't understand how that is going to fix the last broken link.

joaofnfernandes:
It seems that for most urls in `allpages.md` it doesn't make a difference. But for the 404 it does.

```
before /404/
after /404.html
```

That should address the broken link (and hopefully not introduce new ones)

aduermael:
@joaofnfernandes great! :)

PR #1144, opened 2017-01-17
Fix documentation regarding protocol 50 (ESP)

Intellicode:
The documentation should clarify that its protocol 50 and not port 50 that should be allowed between nodes.

### Proposed changes

Fixed documentation to clarify that instead of opening port 50, protocol 50 should be allowed.


johndmulhausen:
Thanks, @Intellicode :)

PR #1150, opened 2017-01-18
Restart the docker daemon after group changes

stephenconnolly1:
Otherwise the verification step won't work.

### Proposed changes

Added a new step to restart the docker daemon. You have to do this otherwise the verification step won't work.


mstanleyjones:
Sorry, this was merged by mistake. After seeing the feedback in #1137 I'm going to revert this.

Issue #1133, opened 2017-01-15
Feedback for: docker-cloud/apps/deploy-to-cloud-btn.md - Docker Button


niccokunzmann:
[docker-cloud/apps/deploy-to-cloud-btn.md](https://docs.docker.com/docker-cloud/apps/deploy-to-cloud-btn/)

When I go to the link https://cloud.docker.com/stack/deploy/?repo=https://github.com/loklak/loklak_server
I get a 404 page.
Also I get a 404, when I access the example page: https://cloud.docker.com/stack/deploy/?repo=https://github.com/docker/dockercloud-quickstart-python

Related Issue: loklak/loklak_server#1001

I would like to get a cloud deployment as mentioned in the docs.

sanscontext:
Hi @niccokunzmann, you usually see this when you are not logged in to Docker Cloud when you click the button. (See the "Note" on the first page.)  Please confirm that you're logged in and see if it works after that? Since this isn't an issue with the documentation, I'm going to close this ticket. If you determine that there's a bug, please email support@docker.com for assistance.

niccokunzmann:
I would rather see that I need to log in. Where do I need to go to implement this?

PR #1151, opened 2017-01-18
Solve #1133

niccokunzmann:
### Proposed changes

This would solve the issue #1133 for me.

Reason: When I read through the page, I only read the bold text and if it is meaningless, I also skip the non-bold text next to it.

Implementation:
- The note is about being logged in and seeing a 404 if not. So I made this bold.
- That it is about the deploy button is clear - I removed the bold text from it.
- That is is a note is clear be structure. Being a note should not rival with the content of the note - so i removed the bold text.

### Related issues (optional)

This would solve the issue #1133 for me.

### DIFF
--- a/docker-cloud/apps/deploy-to-cloud-btn.md
+++ b/docker-cloud/apps/deploy-to-cloud-btn.md
@@ -21,7 +21,7 @@ The button redirects the user to the **Launch new Stack** wizard, with the stack
 
 The user can still modify the stack definition before deployment.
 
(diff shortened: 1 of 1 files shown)


sanscontext:
Hi @niccokunzmann  thanks for your feedback. I appreciate that you think this would solve the problem, however it goes against our formatting conventions and it doesn't actually improve the readability. I'm going to close this PR and do some other work in this file to move this note to a more readable location.

Issue #1026, opened 2017-01-04
Missing instructions for installing CS Engine offline


anweiss:
### Problem description

There are currently no instructions for downloading the CS Engine packages (and their dependencies) and installing them offline. These should be included.

### Problem location

The installation page at https://docs.docker.com/cs-engine/install/ should include download locations for the .rpm, .deb, and SUSE packages, along with their required package dependencies.

### Project version(s) affected

All versions of CS Engine

### Suggestions for a fix

Either edit the existing install page (https://github.com/docker/docker.github.io/blob/master/cs-engine/install.md) or create a new one under `cs-engine` with Offline install instructions (similar to https://github.com/docker/docker.github.io/blob/master/datacenter/ucp/2.0/guides/installation/install-offline.md).

joaofnfernandes:
@mavenugo, @cpuguy83 can you help with this?

banjot:
@joaofnfernandes I think it would be similar to this page: https://docs.docker.com/engine/installation/binaries/

mstanleyjones:
I don't think that's necessarily the right approach, because this needs to be documented for each supported distribution, both for Engine and CS. I made #1029 for that.

PR #1159, opened 2017-01-18
Rewrite and reorganize Linux install instructions

### LINKED ISSUE #703: Feedback for: engine/installation/linux/fedora.md Fedora 25 lacks support
[engine/installation/linux/fedora.md](https://docs.docker.com/engine/installation/linux/fedora/)

The section on creating a dnf|yum repo for docker on fedora 25 does not work, because there is not a 25/ folder at this url currently.

https://yum.dockerproject.org/repo/main/fedora/

Fix by adding the folder and the appropriate contents or by mentioning that support for Fedora 25 isn't available yet.

mstanleyjones:
### Proposed changes

- Rewrote installation instructions for Ubuntu, Debian, RHEL, CentOS, Fedora, Oracle

- Added instructions for installing from packages without using the repositories

- Moved lots of non-dist-specific post-installation instructions to a new file

- Added `docker.repo` files which can be imported by RHEL, CentOS, Fedora, Oracle users (the URLs to these will not work until this PR is merged)

- Added steps for importing and verifying the GPG signing key on all the dists where it is relevant

- Reorganized content in the left-hand nav to emphasize the most popular and well-tested distros

These instructions have already undergone exhaustive review and testing.

Engine 1.13

### Related issues and PRs

Fixes #703
Goes some way to eventually fixing #1026


mstanleyjones:
Here we goooo...

Djelibeybi:
I have a serious objection to this rewrite because it will result in increased support calls. All of the distro-specific support information has been removed. This needs to be restored so that Oracle (and RHEL and SUSE) customers don't think they can install upstream binaries while continuing to receive distro support for Docker. 

I do appreciate the simplification, though.

Djelibeybi:
Also, there are actual inaccuracies in the Oracle text, as it incorrectly refers to Red Hat.

mstanleyjones:
I'll fix the Oracle text relating to Red Hat. Sorry I merged this without seeing your comment. As far as the other one, I'm really not sure what to do about this. Docker has no insight into distro-specific support arrangements with customers. We do call out that not everyone can use third-party binaries right at the top of each of these topics.

Djelibeybi:
Except all the distro-specific content was in the original text. You're removing a whole bunch of content without really articulating the benefit.

engine/installation/linux/oracle.md
```diff
@@ -1,204 +1,236 @@
 ---
-description: Installation instructions for Docker on Oracle Linux.
-keywords: Docker, Docker documentation, requirements, linux, rhel, centos, oracle,  ol
+description: Instructions for installing Docker on Oracle Linux
+keywords: Docker, Docker documentation, requirements, installation, oracle, oel, rpm, install, uninstall, upgrade, update
```
Djelibeybi: Please remove the `oel` keyword and add `ol`.

engine/installation/linux/oracle.md
```diff
@@ -1,204 +1,236 @@
...
+
+- Most users
+  [set up the official Docker repositories](#install-using-the-repository) and
+  install from them, for ease of installation and upgrade tasks. This is the
+  recommended approach.
+
+- Some users download the RPM package and install it manually and manage
+  upgrades completely manually.
+
+- Some users cannot use third-party repositories, and must rely on
```
Djelibeybi: This needs to be updated to refer to the Oracle repositories. Also, it is not clear enough that this is the preferred option for customers who wish to retain Oracle support for the `docker-engine` package. It needs to be clearer that this is the only supported install option by Oracle.

engine/installation/linux/oracle.md
```diff
@@ -1,204 +1,236 @@
...
-5. Start the Docker daemon.
+## Install Docker
 
-     On Oracle Linux 6:
+You can install Docker in different ways, depending on your needs:
+
+- Most users
+  [set up the official Docker repositories](#install-using-the-repository) and
+  install from them, for ease of installation and upgrade tasks. This is the
+  recommended approach.
```
Djelibeybi: This is only the recommended approach by Docker. It is not the recommended approach by the vendor who may be providing support. It needs to be clearer that this is only recommended for customers who do not have an existing support subscription with Oracle.

engine/installation/linux/oracle.md
```diff
@@ -1,204 +1,236 @@
...
+
+- Some users download the RPM package and install it manually and manage
+  upgrades completely manually.
+
+- Some users cannot use third-party repositories, and must rely on
+  the version of Docker in the Red Hat repositories. This version of Docker may
+  be out of date. Those users should consult the Red Hat documentation and not
```
Djelibeybi: A link to the Oracle documentation that outlines the installation and upgrade procedures should be included. It was included in the previous version of the Docker documentation.

engine/installation/linux/oracle.md
```diff
@@ -1,204 +1,236 @@
...
+    ```
 
-1. Log into Oracle Linux as a user with `sudo` privileges.
+    > **Warning**: If you have both stable and unstable repositories enabled,
+    > installing or updating Docker without specifying a version in the
+    > `yum install` or `yum upgrade` command will always install the highest
+    > available version, which will almost certainly be an unstable one.
 
-2. Create the `docker` group.
+3.  On production systems, you should install a specific version of Docker
```
Djelibeybi: You do not explain how to determine which specific version of Docker should be installed. How is the end-user supposed to determine which version of Docker to install here? Also, why shouldn't the user use the latest version of Docker? That's not explained either. Not using the latest version could leave the user vulnerable to security flaws.

engine/installation/linux/oracle.md
```diff
@@ -1,204 +1,236 @@
...
+    append the version string to the package name and separate them by a hyphen
+    (`-`):
 
-    This ensures your user is running with the correct permissions.
+    ```bash
+    $ sudo yum -y install docker-engine-<VERSION_STRING>
+    ```
 
-5. Verify your work by running `docker` without `sudo`.
+    The Docker daemon starts automatically.
```
Djelibeybi: This may have changed in Docker 1.13, but up to 1.12, the daemon is not started automatically in the standard RPM. You need to use `systemctl` to both enable and start the Docker daemon (on OL7) or the `service` and `chkconfig` commands to start the daemon on OL6.

engine/installation/linux/oracle.md
```diff
@@ -1,204 +1,236 @@
...
-should ensure the filesystem is mounted properly prior to restarting the Docker
-service.
+2.  Images, containers, volumes, or customized configuration files on your host
+    are not automatically removed. To delete all images, containers, and
+    volumes:
 
-On Oracle Linux 7, you can use a `systemd.mount` definition and modify the
-Docker `systemd.service` to depend on the btrfs mount defined in systemd.
+    ```bash
+    $ sudo rm -rf /var/lib/docker
```
Djelibeybi: This won't work when the `btrfs` graph driver has been used, as the `rm -rf` command cannot remove the subvolumes that Docker creates.

engine/installation/binaries.md
```diff
@@ -4,243 +4,199 @@ keywords: binaries, installation, docker, documentation, linux
...
+    ```
 
-    https://get.docker.com/builds/Darwin/x86_64/docker-<version>.tgz
+    This command downloads a test image and runs it in a container. When the
+    container runs, it prints an informational message and exits.
 
-For example:
 
-    https://get.docker.com/builds/Darwin/x86_64/docker-1.11.0.tgz
+## Install server and binaries on Windows
```
friism: Here are the instructions in a nutshell:

```
nvoke-WebRequest https://get.docker.com/builds/Windows/x86_64/docker-1.13.0.zip -UseBasicParsing -OutFile docker.zip
Expand-Archive docker.zip -DestinationPath $Env:ProgramFiles
Remove-Item -Force docker.zip

dockerd --register-service

Start-Service docker
```

engine/installation/binaries.md
```diff
@@ -4,243 +4,199 @@ keywords: binaries, installation, docker, documentation, linux
...
 
-You can extract the downloaded archive either by double-clicking the downloaded
-`.tgz` or on the command line, using `tar -xvzf docker-1.11.0.tgz`. The client
-binary can be executed from any location on your filesystem.
+You can install Docker from binaries on Windows Server 2016 or Windows 10.
 
+- To install both client and server binaries, download the 64-bit binary. The
+  archive includes `x86.64` in the file name.
```
friism: it's x86_64 and it's in the path, not the filename (that goes for the client below too)

engine/installation/binaries.md
```diff
@@ -4,243 +4,199 @@ keywords: binaries, installation, docker, documentation, linux
...
 
-    https://get.docker.com/builds/Windows/i386/docker-latest.zip
+2.  Extract the archive using OS utilities. A `docker` directory is created,
+    and contains a `docker.exe` binary and a `dockerd.exe` binary if you
+    downloaded the 64-bit version.
 
-    https://get.docker.com/builds/Windows/x86_64/docker-latest.zip
 
-To download a specific version for Windows, use the following URL pattern:
+3.  **Optional**: Move the binaries to a directory on your executable path. If
```
friism: this doesn't make sense. Instead we should tell people to change their path to include where docker is

engine/installation/binaries.md
```diff
@@ -4,243 +4,199 @@ keywords: binaries, installation, docker, documentation, linux
...
+    ```none
+    C:\> dockerd
+    ```
 
-For example:
+5.  Verify that Docker is installed correctly by running the `hello-world`
+    image.
 
-    https://get.docker.com/builds/Windows/i386/docker-1.11.0.zip
+    > **Note**: To connect the `docker` client to a remote host, either set the
```
friism: I don't think this is necessary - we've just taken reader through local install, they're unlikely to need this

engine/installation/binaries.md
```diff
@@ -4,243 +4,199 @@ keywords: binaries, installation, docker, documentation, linux
...
 
-### Get the macOS binary
+    ```bash
+    $ sudo cp docker/docker /usr/local/bin/
+    ```
 
-The macOS binary is only a client. You cannot use it to run the `docker`
-daemon. To download the latest version for macOS, use the following URLs:
+4.  To connect the `docker` client to a remote host, either set the
+    `DOCKER_HOST` environment variable to the hostname or IP address of the
```
friism: I don't think this is necessary - we've just taken reader through local install, they're unlikely to need this

engine/installation/binaries.md
```diff
@@ -4,243 +4,199 @@ keywords: binaries, installation, docker, documentation, linux
...
-the Unix socket read/writable by the *docker* group when the daemon
-starts. The `docker` daemon must always run as the root user, but if you
-run the `docker` client as a user in the *docker* group then you don't
-need to add `sudo` to all the client commands.
-
-> **Warning**:
-> The *docker* group (or the group specified with `-G`) is root-equivalent;
-> see [*Docker Daemon Attack Surface*](../security/security.md#docker-daemon-attack-surface) details.
+    ```none
+    C:\> docker run hello-world
```
friism:
```
C:\code\tmp> docker run hello-world
Unable to find image 'hello-world:latest' locally
latest: Pulling from library/hello-world
C:\Program Files\Docker\Docker\Resources\bin\docker.exe: image operating system "linux" cannot be used on this platform.
See 'C:\Program Files\Docker\Docker\Resources\bin\docker.exe run --help'.
```

The correct thing is `hello-world:nanoserver`

engine/installation/linux/SUSE.md
```diff
@@ -1,112 +1,301 @@
...
+  [set up Docker's repositories](#install-using-the-repository) and install
+  from them, for ease of installation and upgrade tasks. This is the
+  recommended approach.
+
+- Some users download the RPM package and install it manually and manage
+  upgrades completely manually.
+
+- Some users cannot use third-party repositories, and must rely on the version
+  of Docker in the OpenSUSE or SLES repositories. This version of Docker may be
+  out of date. Those users should consult the CentOS documentation and not
```
friism: "centos"

engine/installation/linux/SUSE.md
```diff
@@ -1,112 +1,301 @@
...
+    ```
+
+2.  Install the latest version of Docker, or go to the next step to install a
+    specific version.
+
+    ```bash
+    $ sudo zypper install docker-engine
+    ```
+
+    > **Warning**: If you have both stable and unstable repositories enabled,
```
friism: This warning is duplicated, no?

engine/installation/linux/SUSE.md
```diff
@@ -1,112 +1,301 @@
...
+    source packages, omit the `-t package` flag from the command.
+
+    ```bash
+    $ zypper search -s --match-exact -t package docker-engine
+
+      Loading repository data...
+      Reading installed packages...
 
-## openSUSE
+      S | Name          | Type    | Version                               | Arch   | Repository
```
friism: @mstanleyjones as previously suggested, can we abbreviate these listings and just show a couple of rows with elipsis? I don't think there's value in listing out everything that happened to be in there when this was written.

This applies to other distros too.

engine/installation/linux/centos.md
```diff
@@ -1,232 +1,234 @@
...
+2.  Install the latest version of Docker, or go to the next step to install a
+    specific version.
 
     ```bash
-    $ curl -fsSL https://get.docker.com/ | sh
+    $ sudo yum -y install docker-engine
     ```
 
-    This script adds the `docker.repo` repository and installs Docker.
+    > **Warning**: If you have both stable and unstable repositories enabled,
```
friism: @mstanleyjones why are we duplicating this warning?

Djelibeybi:
I apologise if my previous tone was too abrupt or rude. I realise I am probably missing some larger context or motivation behind such large changes to the documentation. 

If you could, would you mind explaining the rationale behind these changes so we can better review them for the target audience?

engine/installation/linux/SUSE.md
mstanleyjones: Yes once while setting up the repos and once when actually installing. It's pretty important.

mstanleyjones:
Thanks for all the feedback. Please move any more feedback to #1172.

engine/installation/linux/centos.md
mstanleyjones: Once during enabling the repos, once when actually making the decision about what version to install.

engine/installation/linux/oracle.md
mstanleyjones: This is just saying not to always update to the latest just because it's there, but to be mindful when upgrading. This is not about which repo the bits come from. Sorry for the confusion.

engine/installation/linux/oracle.md
mstanleyjones: You're right, I verified this behavior and updated the contents in #1172.

engine/installation/linux/oracle.md
mstanleyjones: Thanks, I've added info about this in #1172.

engine/installation/linux/oracle.md
mstanleyjones: Done in #1172.

engine/installation/linux/SUSE.md
mstanleyjones: Fixed in #1172.

engine/installation/binaries.md
mstanleyjones: Thanks! Added to #1172.

engine/installation/binaries.md
mstanleyjones: Removed in #1172.

engine/installation/binaries.md
mstanleyjones: Removed in #1172.

engine/installation/binaries.md
mstanleyjones: Done.

engine/installation/binaries.md
mstanleyjones: Done.

engine/installation/linux/SUSE.md
mstanleyjones: Done.

friism:
@Djelibeybi it's all good - thanks for the review and your constructive comments, it's really helpful. I also want to apologize if this seems rushed - we're crunching for a release.

Ultimately, we want these docs to be short and to the point, and we want to help people install the official Docker packages that are built, supported and maintained by Docker.

Further, while I'm sure that the `docker-engine` packages that Oracle maintains are great (and I commend Oracle for having up-to-date releases), from our standpoint they suffer from the same shortcomings that you identify with Docker's `docker-engine` packages: Oracle's packages are not supported by Docker. And similar to what we're trying to achieve with our docs, the Oracle docs don't mention the risks inherent in using Oracle's `docker` package, nor do they point users to Docker's packages for a Docker-supported solution.

Again, thanks for your help reviewing the docs and please keep the feedback coming.

Djelibeybi:
@friism I think a generic note is probably sufficient then, across the board. Something like, "These instructions install the official Docker build which may not be supported by your distribution vendor. Support for this build is available from Docker, please see..." and a link to the Docker support subscription page or product page or whatever.

Essentially just to make it a little clearer to anyone who installs this particular build that support is available, but from Docker themselves.

PR #1161, opened 2017-01-18
replaces edit from #979

sanscontext:
replaces edit from #979, clarifies what you can do in Hub.

cc @joaofnfernandes for 👀 


joaofnfernandes:
Ship it 🚀

PR #1162, opened 2017-01-18
adds more info about valid docker ID

### LINKED ISSUE #1104: Feedback for: docker-id/index.md
[docker-id/index.md](https://docs.docker.com/docker-id/)

There should be a section to include the Docker ID naming convention to notify users about what a valid Docker ID should be:

- a minimum of 4 characters 
- lowercase alphanumeric characters only (no spaces and dashes/hyphens `-`)

sanscontext:
Closes #1104 

Adds details about what makes a valid Docker ID, since this was missing from the validation on the field. (Also filed internal tickets to address this.)

### DIFF
--- a/docker-id/index.md
+++ b/docker-id/index.md
@@ -19,7 +19,13 @@ Your Docker ID becomes your user namespace for hosted Docker services, and becom
 
 1. Go to the [Docker Cloud sign up page](https://cloud.docker.com).
 
-2. Enter a username, a valid email address, and a password.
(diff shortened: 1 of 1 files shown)


joaofnfernandes:
Feel free to merge this, but I think that details like these don't add much value to the docs:

* I'll not read the docs to learn how to create a Docker account. If my password needs to have at least 6 characters I'll get immediate feedback when creating the account
* These details will change as our security policies evolve, so not having this info in the docs will help the docs to be up to date

PR #1172, opened 2017-01-19
Addressing feedback from new installation instructions

mstanleyjones:
### Proposed changes

Addressed technical feedback from #1159.

Engine 1.13

### DIFF
--- a/engine/installation/linux/oracle.md
+++ b/engine/installation/linux/oracle.md
@@ -67,12 +67,23 @@ Docker from the repository.
     $ sudo yum install -y yum-utils
     ```
 
-2.  Use the following command to set up the **stable** repository:
+2.  Use one of the following commands to set up the **stable** repository,
+    depending on your version of Oracle Linux:
+
+    **Oracle Linux 7**:
 
     ```bash
     $ sudo yum-config-manager \
         --add-repo \
-        https://docs.docker.com/engine/installation/linux/repo_files/oracle/docker.repo
+        https://docs.docker.com/engine/installation/linux/repo_files/oracle/docker-ol7.repo
+    ```
+
+    **Oracle Linux 6**:
+
+    ```bash
+    $ sudo yum-config-manager \
+        --add-repo \
+        https://docs.docker.com/engine/installation/linux/repo_files/oracle/docker-ol6.repo
     ```
 
 3.  **Optional**: Enable the **testing** repository. This repository is included
--- a/engine/installation/linux/oracle.md
+++ b/engine/installation/linux/oracle.md
@@ -1,6 +1,6 @@
 ---
 description: Instructions for installing Docker on Oracle Linux
-keywords: Docker, Docker documentation, requirements, installation, oracle, oel, rpm, install, uninstall, upgrade, update
+keywords: Docker, Docker documentation, requirements, installation, oracle, ol, rpm, install, uninstall, upgrade, update
 redirect_from:
 - /engine/installation/oracle/
 title: Get Docker for Oracle Linux
@@ -14,9 +14,13 @@ To get started with Docker on Oracle Linux, make sure you
 
 ### OS requirements
 
-To install Docker, you need the 64-bit version of Oracle Linux 6 or 7, running
-the Unbreakable Enterprise Kernel Release 4 (4.1.12) or higher. This kernel
-supports the Docker `btrfs` storage engine.
+To install Docker, you need the 64-bit version of Oracle Linux 6 or 7.
+
+To use `btrfs`, you need to install the Unbreakable Enterprise Kernel (UEK)
+version 4.1.12 or higher. running the Unbreakable Enterprise Kernel Release 4
+(4.1.12) or higher. For Oracle Linux 6, you need to enable extra repositories
+to install UEK4. See
+[Obtaining and installing the UEK packages](https://docs.oracle.com/cd/E37670_01/E37355/html/ol_obtain_uek.html){: target="_blank" class="_" }.
 
 ### Remove unofficial Docker packages
 
@@ -56,7 +60,14 @@ Docker from the repository.
 
 #### Set up the repository
 
-1.  Use the following command to set up the **stable** repository:
+1.  Install the `yum-utils` plugin, which provides the `yum-config-manager`
+    plugin.
+
+    ```bash
+    $ sudo yum install -y yum-utils
+    ```
+
+2.  Use the following command to set up the **stable** repository:
 
     ```bash
     $ sudo yum-config-manager \
@@ -64,7 +75,7 @@ Docker from the repository.
         https://docs.docker.com/engine/installation/linux/repo_files/oracle/docker.repo
     ```
 
-2.  **Optional**: Enable the **testing** repository. This repository is included
+3.  **Optional**: Enable the **testing** repository. This repository is included
     in the `docker.repo` file above but is disabled by default. You can enable
     it alongside the stable repository. Do not use unstable repositories on
     on production systems or for non-testing workloads.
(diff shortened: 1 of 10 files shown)


engine/installation/linux/oracle.md
```diff
@@ -14,9 +14,13 @@ To get started with Docker on Oracle Linux, make sure you
 
 ### OS requirements
 
-To install Docker, you need the 64-bit version of Oracle Linux 6 or 7, running
-the Unbreakable Enterprise Kernel Release 4 (4.1.12) or higher. This kernel
-supports the Docker `btrfs` storage engine.
+To install Docker, you need the 64-bit version of Oracle Linux 6 or 7.
+
+To use `btrfs`, you need to install the Unbreakable Enterprise Kernel (UEK)
```
Djelibeybi: This is not just required for `btrfs` support. Oracle will only support `docker-engine` when running the UEK4. Technically, `btrfs` is supported on both the UEK and RHCK kernel variants.

Though, I guess this is somewhat of a moot point, because the rest of the documentation outlines how to install a completely unsupported version of Docker anyway.

engine/installation/linux/oracle.md
```diff
@@ -144,9 +167,24 @@ Docker from the repository.
...
-    The Docker daemon starts automatically.
+    The Docker daemon does not start automatically.
+
+5.  Start the Docker daemon. Use `systemctl` on Oracle Linux 7 or `service` on
+    Oracle Linux 6.
+
+    **Oracle Linux 7**:
+
+    ```bash
+    $ sudo systemctl start docker
```
Djelibeybi: You also need to run `systemctl enable docker` in order for Docker to start automatically on boot.

engine/installation/linux/oracle.md
```diff
@@ -144,9 +167,24 @@ Docker from the repository.
...
+    ```
+
+    **Oracle Linux 6**:
 
-4.  Verify that `docker` is installed correctly by running the `hello-world`
+    ```bash
+    $ sudo service docker start
```
Djelibeybi: You also need to run `chkconfig docker on` to enable Docker to start on boot.

engine/installation/linux/oracle.md
```diff
@@ -106,29 +137,21 @@ Docker from the repository.
     > `yum install` or `yum upgrade` command will always install the highest
     > available version, which will almost certainly be an unstable one.
 
-3.  On production systems, you should install a specific version of Docker
+4.  On production systems, you should install a specific version of Docker
```
Djelibeybi: There is no explanation of why you should install a specific version of Docker, nor how to decide which version to install. In the absence of this decision-making guidance, the best advice should be to install the latest stable version, which is what would happen automatically if the `testing` repository is not enabled.

engine/installation/linux/oracle.md
```diff
@@ -187,9 +225,24 @@ a new file each time you want to upgrade Docker.
     $ sudo yum install /path/to/package.rpm
     ```
 
-    The Docker daemon starts automatically.
+    The Docker daemon does not start automatically.
```
Djelibeybi: Step 6 appears to be a duplication of information from step 4. Perhaps it should be split out into its own section on "Enabling and Starting the Docker daemon" which covers both install options.

engine/installation/linux/oracle.md
```diff
@@ -227,6 +279,11 @@ instead of `yum -y install`, and pointing to the new file.
     $ sudo rm -rf /var/lib/docker
     ```
 
+    > **Note**: This won't work when the `btrfs` graph driver has been used,
+    > because the `rm -rf` command cannot remove the subvolumes that Docker
+    > creates. See the output of `man btrfs-subvolume` for information on
+    > removing `btrfs` subvolumes.
```
Djelibeybi: Add something like, "Subvolumes need to be removed before the rest of the `/var/lib/docker` content can be erased."

engine/installation/linux/oracle.md
```diff
@@ -14,9 +14,13 @@ To get started with Docker on Oracle Linux, make sure you
 
 ### OS requirements
 
-To install Docker, you need the 64-bit version of Oracle Linux 6 or 7, running
-the Unbreakable Enterprise Kernel Release 4 (4.1.12) or higher. This kernel
-supports the Docker `btrfs` storage engine.
+To install Docker, you need the 64-bit version of Oracle Linux 6 or 7.
```
Djelibeybi: It would still be great to have a link to the official Oracle Linux Docker User Guide to help guide users to the installation process for the fully supported Docker release from Oracle.

engine/installation/linux/oracle.md
```diff
@@ -94,7 +116,27 @@ Docker from the repository.
     ```bash
     $ sudo yum -y check-update
     ```
-2.  Install the latest version of Docker, or go to the next step to install a
+
+2.  Verify and import Docker's public key, which is used to sign packages in
```
Djelibeybi: This entire step is not required because you specify the GPG key in the YUM repository configuration and you've enabled GPG checking as part of the yum install process. Therefore, the install of the GPG will automatically be triggered by YUM when the package is installed. 

Thus, this step should be removed and an additional step should be added later to verify the GPG key when the `yum` command prompts the user to accept it.

engine/installation/linux/oracle.md
```diff
@@ -94,7 +116,27 @@ Docker from the repository.
     ```bash
     $ sudo yum -y check-update
```
Djelibeybi: This command does not update the `yum` package index. In order to do that, you should run `yum makecache fast` instead.

engine/installation/linux/oracle.md
```diff
@@ -210,7 +274,6 @@ To upgrade Docker, download the newer package file and repeat the
 [installation procedure](#install-from-a-package), using `yum -y upgrade`
```
Djelibeybi: Why are you advocating the use of the `-y` parameter? This would prevent the user from confirming the install. It is usually only recommended when triggering `yum` from an automation or management system.

engine/installation/linux/oracle.md
```diff
@@ -144,9 +178,24 @@ Docker from the repository.
     $ sudo yum -y install docker-engine-<VERSION_STRING>
```
Djelibeybi: I would remove the `-y` option from this command-line example. It would suppress the GPG key validation check.

engine/installation/linux/oracle.md
```diff
@@ -227,6 +290,11 @@ instead of `yum -y install`, and pointing to the new file.
     $ sudo rm -rf /var/lib/docker
     ```
 
+    > **Note**: This won't work when the `btrfs` graph driver has been used,
```
Djelibeybi: "This won't work" implies that it's somehow broken instead of addressing the fact that Docker creates `btrfs` subvolumes that would need to be removed first. 

Rather, it should be something like "If you're using the `btrfs` graph driver, the Docker-created subvolumes need to be removed prior to removing the rest of the Docker content. You can review the [Oracle Linux documentation](http://docs.oracle.com/en/operating-systems/linux.html) to learn how to list and remove subvolumes from a `btrfs` volume.

engine/installation/linux/repo_files/oracle/docker-ol6.repo
```diff
@@ -1,27 +1,27 @@
 [docker-main]
 name=Docker Repository
-baseurl=https://yum.dockerproject.org/repo/main/oraclelinux/$releasever/
```
Djelibeybi: If you renamed your directory (or created a symlink) on your YUM server to `7Server` and `6Server`, you could keep the `$releasever` variable and simplify back to a single `docker.repo` file for both Oracle Linux 6 and 7.

mstanleyjones:
I really apologize for merging while your review was in flight. I was asked to go ahead and merge. I appreciate your feedback and it hasn't been lost. I will continue in the morning.

Djelibeybi:
Could I ask that you ask whomever asked you to merge to not do that tomorrow? :) 

I can't be the only person interested in ensuring the quality and effectiveness of the documentation and submitting and merging immediately implies that you're not really interested in external review.

mstanleyjones:
Thanks again for your feedback and I'm sorry about the crossed wires. I'm going to address the feedback very soon, but we are in the midst of a release right now, so I appreciate your patience.

To address the one point about starting Docker automatically at boot, those steps have been moved to the post-install configuration topic, as they are common across most distributions of Linux.

To address the point about duplication between the "Install using the repository" and "Install using a package" sub-procedures, the thinking was that people will be doing one or the other, not both, and it's convenient for them to have all the steps they need within their sub-procedure. I'm happy to re-evaluate the structure.

PR #1175, opened 2017-01-19
WIP: voting app example updated Compose v3 features

londoncalling:
Draft 2 of voting app example walkthrough, work-in-progress.

* Revised list of Compose v.3 features, and added better links into the Compose file reference with regard to upgrading and versioning

Still TBD:

* add links to @ManoMarks voting app lab topics and source code when it is published
* update code examples per chat with @johndmulhausen 
* add app diagram or other images on first page
* incorporate any review comments

@johndmulhausen , I considering moving the Compose v3 features topic above the `docker-stack.yml` code block as we discussed, but it doesn't make as much sense when it precedes the full `.yml` file. Let's discuss tomorrow a.m.

### DIFF
--- a/engine/getstarted-voting-app/index.md
+++ b/engine/getstarted-voting-app/index.md
@@ -58,6 +58,14 @@ Each service will run in its own container. Using swarm mode,
 we can also scale the application to deploy replicas
 of containerized services distributed across multiple nodes.
 
+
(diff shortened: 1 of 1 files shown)


johndmulhausen:
Let's merge and iterate on this.

PR #1181, opened 2017-01-19
redeploying should not use existing volume

sardjv:


sanscontext:
Hey @sardjv, thanks for the feedback. Could you edit your PR to use the actual syntax used by the command? (https://docs.docker.com/apidocs/docker-cloud/#redeploy-a-service FWIW)

sanscontext:
Merging, since this is actually confirmed correct. Thanks @sardjv !

PR #1185, opened 2017-01-19
Update systemd.md

paddyboyle:
Removed the quotes from the --graph option on line 130.  Using quotes works for version 1.12 of Docker but with 1.13, Docker will create a directory named " under / and then create a directory of mnt/docker-data" under that directory.


johndmulhausen:
Thanks, @paddyboyle :)

PR #1222, opened 2017-01-20
Renamed SUSE.md to suse.md

mstanleyjones:
### Proposed changes

It's confusing the link checker.


sanscontext:
Hah, some of the links I fixed in the linkchecker leadup were to change the case - apparently the linkchecker is case-sensitive. Just FYI you'll probably see that break again.

PR #1224, opened 2017-01-20
Corrections to ubuntu install docs, based on attempting to install on Wily

nafg:
### Proposed changes

* Added `-` to apt-key command so it should use stdin
* Changed apt repo layout in add-apt-repository command

The apt repo layout seems wrong, it only worked after the proposed change.

Also apt-key seems to need `-` to look at stdin; otherwise I got `gpg: can't open `': No such file or directory`

### DIFF
--- a/engine/installation/linux/debian.md
+++ b/engine/installation/linux/debian.md
@@ -94,7 +94,7 @@ Docker from the repository.
 2.  Add Docker's official GPG key:
 
     ```bash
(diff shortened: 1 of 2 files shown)


mstanleyjones:
I tested the instructions on Ubuntu Xenial and it worked with and without the `-`, so I'll allow the  change. However, Wily is not supported for Docker 1.13.

I rebased your patch and added the change to the Debian instructions as well, to be consistent. I'll merge after the tests complete. Thanks for your contribution.

PR #1238, opened 2017-01-20
Adding a dash at the end of `apt-key add` command

artheus:
The dash at the end of the `apt-key add` command is important as it will otherwise expect to read from a file, and not `stdin`

### DIFF
--- a/engine/installation/linux/ubuntu.md
+++ b/engine/installation/linux/ubuntu.md
@@ -71,7 +71,7 @@ Docker from the repository.
 2.  Add Docker's official GPG key:
 
     ```bash
-    $ curl -s http://yum.dockerproject.org/gpg | sudo apt-key add -
+    $ curl -fsSL https://yum.dockerproject.org/gpg | sudo apt-key add -
     ```
 
     > **Note**: The URL is correct, even for Linux distributions that use `APT`.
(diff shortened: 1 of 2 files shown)


engine/installation/linux/ubuntu.md
```diff
@@ -71,7 +71,7 @@ Docker from the repository.
 2.  Add Docker's official GPG key:
 
     ```bash
-    $ curl -s http://yum.dockerproject.org/gpg | sudo apt-key add
+    $ curl -s http://yum.dockerproject.org/gpg | sudo apt-key add -
```
thaJeztah: Should we use this perhaps;

```bash
$ apt-key adv --fetch-keys http://yum.dockerproject.org/gpg
```

thaJeztah: I notice it's using `http://` not `https://` as well, so at least that needs to be updated, although `apt-key` doesn't support that without `gnupg-curl` being installed, so if we go for that, we should add it to step 1.


If we _do_ stick with `curl`, we probably should change to;

```bash
$curl -fsSL https://yum.dockerproject.org/gpg | sudo apt-key add -
```

mstanleyjones: If we need to use https, then the second option is better. Can you also change it for the Debian instructions, @artheus ? Mind the space between the `$` and the `curl` though.

mstanleyjones:
I went ahead and made those changes and pushed them into the PR. Thanks @artheus

PR #1241, opened 2017-01-20
Update ubuntu.md

davros1973:
### Proposed changes

Please forgive me for my ignorance ... I'm not very familiar with Linux yet.  I kept getting a 403 error (malformed URL really ... "dists" repeated) and failure to find the release file etc..  I'm not sure if what I've done is a fix, or an ugly work-around ... but it works for me without any issue I can see doing it this way.  (Ubuntu - xenial 16.04 Docker version 1.13.0 build 49bf474).  Removed pool/ from repository url, and prefixed "ubuntu-" for lsb_release codename.

### DIFF
--- a/engine/installation/linux/ubuntu.md
+++ b/engine/installation/linux/ubuntu.md
@@ -93,8 +93,8 @@ Docker from the repository.
 
     ```bash
     $ sudo add-apt-repository \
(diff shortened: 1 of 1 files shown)


mstanleyjones:
@thaJeztah I'd like to test that first. This change is good for now. Thanks @davros1973

PR #1251, opened 2017-01-20
Fixed bash command and repo url

lrkwz:


mstanleyjones:
This was also reported and fixed in #1224. Thank you!

PR #1255, opened 2017-01-20
Use fixed $releasever for CentOS / RHEL

thaJeztah:
The repo-files used in the installation instructions currently use `$releasever`
to configure the right YUM repository to use, however _Server_ releases output
the version in a different format (i.e., `7Server` instead of just `7`).

Given that we don't have separate repositories for "Server" and "Desktop" editions,
and we only support version 7, setting the repo-files to use a fixed version.

This issue was originally reported in
https://github.com/docker/docker/issues/23376

And fixed through 2b36087597b67b23c8309fe9330a60f17952496b, but this change
got lost during the rewrite of the installation docs.

ping @mstanleyjones PTAL

### DIFF
--- a/engine/installation/linux/repo_files/centos/docker.repo
+++ b/engine/installation/linux/repo_files/centos/docker.repo
@@ -1,27 +1,27 @@
 [docker-main]
 name=Docker Repository
-baseurl=https://yum.dockerproject.org/repo/main/centos/$releasever/
+baseurl=https://yum.dockerproject.org/repo/main/centos/7/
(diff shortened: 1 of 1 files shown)


andrewhsu:
The only 2 packages in the centos6 yum repo are docker 1.7.0 and 1.7.1. Plus centos6 has kernel 2.6.32 which i don't think we want to support for running docker.

I'd recommend just hardcoding the `docker.repo` definition to point to the centos7 yum directory. Do not use `$releasever`.

PR #1282, opened 2017-01-20
Fixed bash commands format.

mcheptea:
### Proposed changes

This is a simple tutorial page rendering fix. The webpage https://docs.docker.com/engine/swarm/swarm-tutorial/create-swarm/ doesn't render correctly the code blocks. I removed the `bash` tags to improve the rendering.

### Related issues (optional)

None.


sanscontext:
Hey @mcheptea thanks for the feedback. Unfortunately, this is a render issue with the indentation of the block, not the format cue.

johndmulhausen:
Yep, I'll fix it in my own commit. Thanks for bringing it to our attention, @mcheptea :)

PR #1289, opened 2017-01-20
Fix installation instructions for Ubuntu

Intellicode:
The previous url was incorrect and returns error while doing sudo apt-get update

### Proposed changes

Fix repository path for Ubuntu installation instructions. The original instructions cause problems while performing ```sudo apt-get update```

```
W: The repository 'https://apt.dockerproject.org/repo/pool xenial Release' does not have a Release file.
N: Data from such a repository can't be authenticated and is therefore potentially dangerous to use.
N: See apt-secure(8) manpage for repository creation and user configuration details.
E: Failed to fetch https://apt.dockerproject.org/repo/pool/dists/xenial/main/binary-amd64/Packages  403  Forbidden
E: Some index files failed to download. They have been ignored, or old ones used instead.
```

### DIFF
--- a/engine/installation/linux/ubuntu.md
+++ b/engine/installation/linux/ubuntu.md
@@ -93,8 +93,8 @@ Docker from the repository.
 
     ```bash
     $ sudo add-apt-repository \
-           "deb https://apt.dockerproject.org/repo/pool/ \
-           $(lsb_release -cs) \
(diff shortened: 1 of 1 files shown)


Intellicode:
Nvm, seems someone else already fixed it!

PR #1292, opened 2017-01-20
Close 'code' html elements

snesneros:
There were missing slashes in the closing tags, causing the page to display incorrectly.


johndmulhausen:
Thanks, @snesneros :)

PR #1301, opened 2017-01-20
CI - broken link checking improvements

aduermael:
### Proposed changes

Improving link checking...

- [x] check all links in `_data/toc.yaml` prior to checking html files
- [x] consider html base tag
- [x] consider self closing html tags
- [x] don't fail on redirects (but show them in the logs)
- [x] logs: print broken links and redirects in CSV format
- [x] overall timeout for Jenkins tests

PR #1302, opened 2017-01-20
CI - turn off broken link checking

aduermael:
### Proposed changes

While I'm working on #1301, it would be better to turn off broken link checking.


aduermael:
@johndmulhausen What do you think? 
Also would really love these tests to be made required. I do prefer to turn off problematic ones to take the time make them better. We'll turn them back on next week.

mstanleyjones:
Merging at @aduermael request.

PR #1306, opened 2017-01-21
Update `docker-stack` command to `docker stack`

axsuul:
There is no such thing as `docker-stack`


johndmulhausen:
Thanks, @axsuul -- right you are :)

PR #1308, opened 2017-01-21
Update broken link in documentation (step_three.md)

timazet:
### Proposed changes

Incorrect format of link displaying was fixed
Link was changed to correct one


johndmulhausen:
Actually, we're encouraging people to go try the search in the walkthrough; it might be interesting to link people right to whalesay and change the focus of this page a bit but as is this kind of breaks the instructions.

PR #1315, opened 2017-01-22
Fixed codeformating

VeeeneX:
### Proposed changes

Edited wrong formating


johndmulhausen:
This doesn't appear to render incorrectly for me. Sorry, @VeeeneX -- I do see weirdness later in the file (with ````bash` code fences showing up) but not this.

PR #1320, opened 2017-01-23
browse to docker hub link now in markup

codypearce:
### Proposed changes

Change browse to docker hub link on the getting started step three page to markup language link rather then an anchor tag because it was showing as an anchor element on the site rather then a normal link. 

Can be seen at https://docs.docker.com/engine/getstarted/step_three/


johndmulhausen:
Duplicate of #1347

PR #1324, opened 2017-01-23
Fix usage of wait-for-postgres.sh

xaqq:
The docker-compose example file passed `db:5432` as the `-h` option for psql. However, `-h` is to specify the hostname, and doesn't
accept a port number.


johndmulhausen:
Thanks, but I see this usage in wait-for-it.sh's docs: 

`wait-for-it.sh host:port [-s] [-t timeout] [-- command args]`

PR #1331, opened 2017-01-23
Cloud: clarify registration runs on all manager nodes

sanscontext:
Clarify that the Cloud registration container runs on all managers nodes in a swarm.

(cc @friism does this help?)

### DIFF
--- a/docker-cloud/cloud-swarm/index.md
+++ b/docker-cloud/cloud-swarm/index.md
@@ -78,7 +78,11 @@ To register an existing swarm in Docker Cloud:
 
     To register a swarm with an organization, prefix the new name with the organization name, for example `myorganization/myteamswarm`.
 
-The manager node pulls the `dockercloud/registration` container and runs it on the swarm. The swarm then appears in the **Swarms** screen in Docker Cloud.
+The manager node pulls the `dockercloud/registration` container which creates a
+global service called `dockercloud-server-proxy`. This service runs on _all_ of
+the swarm's manager nodes.
+
+The swarm then appears in the **Swarms** screen in Docker Cloud.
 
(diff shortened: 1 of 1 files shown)


friism:
@sanscontext we should check with @fermayo. I suspect that the `registration` container actually creates a service powered by a different image (but I don't know that for a fact). So what happens is:

 1. User `docker run` the registration image
 2. The registration container now uses the Docker API to create a new service
 3. That service runs on all managers, keeps running, etc.
 4. The `docker run` container stops and exits

tifayuki:
@friism 
Yes, you are correct.
the registration container creates a `global` service(`dockercloud/server-proxy`) with `node.role == manager` and exits.

sanscontext:
Thanks @tifayuki - I re-edited it to clarify, can you look one more time? 🙏

docker-cloud/cloud-swarm/index.md
```diff
@@ -78,7 +78,11 @@ To register an existing swarm in Docker Cloud:
 
     To register a swarm with an organization, prefix the new name with the organization name, for example `myorganization/myteamswarm`.
 
-The manager node pulls the `dockercloud/registration` container and runs it on the swarm. The swarm then appears in the **Swarms** screen in Docker Cloud.
+The manager node pulls the `dockercloud/registration` container which creates a
+global service called `dockercloud/serverproxy`. This service runs on _all_ of
```
tifayuki: the service name is `dockercloud-server-proxy`. The image it uses is `dockercloud/server-proxy``

sanscontext: Doh. Sorry about that. Fixed.

PR #1340, opened 2017-01-23
Reapply pr 694

mstanleyjones:
@cyli noticed that somehow pull request #694 disappeared when we released Engine 1.13. I have no explanation at all, but this patch re-applies it.


cyli:
LGTM on green - thanks for fixing this quickly @mstanleyjones!

mstanleyjones:
Failure is unrelated. merging.

PR #1341, opened 2017-01-23
Don't fetch the registry api template

mstanleyjones:
This file causes build errors (see https://jenkins.dockerproject.org/job/docker/job/docker-github-io.b3m3d0/job/PR-1340/2/console) and it only used by the Registry project, not by us. We consume the files that they generate using the template.

cc @stevvooe


stevvooe:
LGTM

johndmulhausen:
Merging; I don't think Netlify or Jenkins are gonna help us test this one out. :)

PR #1351, opened 2017-01-24
fix link to logging context

thaJeztah:
this source file no longer exists on "master",
so replacing the link with a permalink on the
1.13 tag.

fixes https://github.com/docker/docker/issues/30403


johndmulhausen:
Change the URL to:

https://github.com/docker/docker/blob/{{ page.enginebranch }}/daemon/logger/context.go

That way, when we change _config.yml in the future this link will stay current.

thaJeztah:
@johndmulhausen won't work because the file is renamed in 1.14, so it will still break then

thaJeztah:
all green

PR #1352, opened 2017-01-24
Fix Markdown

sathieu:
The formating of https://docs.docker.com/engine/userguide/storagedriver/device-mapper-driver/#configure-docker-with-devicemapper is incorrect


johndmulhausen:
Thanks @sathieu :)

PR #1347, opened 2017-01-24
Fix broken link

edmundojr:
### Proposed changes

Fix link format by updating its format to markdown.

URL: https://docs.docker.com/engine/getstarted/step_three/
Screenshot:
screen shot 2017-01-24 at 09 41 25

PR #1354, opened 2017-01-24
Updated browse to the Docker Hub link formatting

ajinkyapisal:
### Proposed changes

https://docs.docker.com/engine/getstarted/step_three/ Link on this page was not rendered correctly. Updated it to use markdown link style instead of anchor tag


johndmulhausen:
Duplicate of #1347

PR #1359, opened 2017-01-24
iam-permissions missing

usultis:
Docker for AWS Stack (1.13.0-1) fails without ec2:AuthorizeSecurityGroupEgress and ec2:AuthorizeSecurityGroupIngress permissions


johndmulhausen:
Thanks, @usultis :)

PR #1368, opened 2017-01-25
Fix language within devicemapper thinpool setup

RonWilliams:
### Proposed changes

Previous language indicated that a user would be "creating a thin pool". This language is confusing as it indicates that the logical volume created is of the thin pool type; instead, the volume being created is a linear volume group which is then converted to a thin pool in step 6. This is an important nuance when implementing setup through configuration management tools such as Salt or Ansible. When initially provisioning the two logical volumes as thin pool, the error "Pool metadata LV docker/thinpoolmeta is of an unsupported type". Updating the language reflects the actual command behaviors. 

### Related issues (optional)

N/A

### DIFF
--- a/engine/userguide/storagedriver/device-mapper-driver.md
+++ b/engine/userguide/storagedriver/device-mapper-driver.md
@@ -237,7 +237,7 @@ assumes that the Docker daemon is in the `stopped` state.
     $ vgcreate docker /dev/xvdf
     ```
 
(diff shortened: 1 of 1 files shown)


RonWilliams:
Here's a link to the page within netlify for review:

https://deploy-preview-1368--docker-docs.netlify.com/engine/userguide/storagedriver/device-mapper-driver/

PR #1369, opened 2017-01-25
getting started button format update, voting app services topic intro

londoncalling:
### Proposed Changes

- updated CSS to implement white buttons for links to docs (vs. download installer links), based on Josh's new `secondary-btn` CSS with some tweaks to get it to work with current `documentaton.css`

- updated `anatomy of voting app` topic title and intro

@johndmulhausen @joaofnfernandes @mstanleyjones @Raphaelbouchard 

### DIFF
--- a/_data/toc.yaml
+++ b/_data/toc.yaml
@@ -140,7 +140,7 @@ toc:
   - sectiontitle: Define and deploy your app
     section:
     - path: /engine/getstarted-voting-app/
-      title: Tour the voting app sample
+      title: Sample app overview
     - path: /engine/getstarted-voting-app/node-setup/
       title: Set up Dockerized machines
     - path: /engine/getstarted-voting-app/create-swarm/
--- a/css/documentation.css
+++ b/css/documentation.css
@@ -137,13 +137,34 @@
 /* reset away from #008AB5 */
 
 #DocumentationText a.button {
-	color: #fff;
 }
 
 #DocumentationText .darkblue-btn {
     color: #ffffff;
     margin-top: 15px;
 }
+
+#DocumentationText .secondary-btn,
+   button .secondary-btn {
+     border: 2px solid #1488c6;
+     color: #1488c6;
+     background: white;
+     margin-top: 15px;
+     font-size: 14px;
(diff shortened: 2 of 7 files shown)


johndmulhausen:
What do we think, guys? 
https://deploy-preview-1369--docker-docs.netlify.com/

johndmulhausen:
I'm cool with it but I defer to @Raphaelbouchard

Raphaelbouchard:
I agree, I like the distinction between CTAs that take you to more docs, downloads or tutorials

mstanleyjones:
👍

londoncalling:
@johndmulhausen @Raphaelbouchard **re: button format**Thanks, just to clarify re: button formats, I used Josh's button, but currently in order to get it to work, you have to override some things in `documentation.css` (e.g., white text is defined as a default color on an `#DocumentationText a.button` which is a little weird. So I had to get rid of that. It doesn't affect anything, because we are using because we already explicitly define the text color for the buttons we are using, e.g., 

```
#DocumentationText .darkblue-btn {
    color: #ffffff;
    margin-top: 15px;
}
```

Hopefully, Josh's new CSS will not have these problems. I'll check in with him today.

@johndmulhausen @ManoMarks @mstanleyjones @thaJeztah  Based on a good conversation I had with @joaofnfernandes re: tutorials yesterday, I added a little more context and explanation about artifacts like like `Dockerfiles` and  `Compose files`, and concepts like `services`, `containers`, and `images` so that users have a reference point for what they learn in beginner [Getting Started with Docker](https://docs.docker.com/engine/getstarted/) to [Define and deploy your app](https://docs.docker.com/engine/getstarted-voting-app/).  See what you think ...

This is just a start and making these tutorials easier to follow. I'd like to use this approach more thoroughly as we re-do the tutorials and choose new sample apps for them; perhaps a palette of several tutorials that build on each other, or if one big one that uses the same sample, then bite-sized topics within it, each that build on the previous.

londoncalling:
LGTM, waiting for Jenkins migration to be done, then I'll merge this

PR #1373, opened 2017-01-25
Change cloud.docker.com/login to id.docker.com/login in site template

mstanleyjones:
The old URL redirects to the new one, so let's use the new one. This currently comes in the W3C validator as "broken".


mstanleyjones:
LGTM, merging.

PR #1377, opened 2017-01-25
Fixed a typo

Thoma5:
Dcoker instead of Docker

### Proposed changes

Replaced Dcoker with Docker, also removed an unnecessary space.


sanscontext:
Thanks @Thoma5 !

PR #1393, opened 2017-01-26
Fix info message to be more clear

KickingTheTV:
### Proposed changes
Current description mentions making the swarm "unavailable" in Docker Cloud, but can be confusing because we have a swarm state "unavailable"


sanscontext:
Thanks @KickingTheTV !

PR #1394, opened 2017-01-26
Missing privileged flag

clemenko:
Added privileged flag.

### DIFF
--- a/datacenter/ucp/2.0/guides/installation/upgrade.md
+++ b/datacenter/ucp/2.0/guides/installation/upgrade.md
@@ -78,7 +78,7 @@ that you've installed using that node's root CA material.
 
     ```bash
     $ docker run --rm -it \
-      --name ucp \
+      --name ucp --privileged\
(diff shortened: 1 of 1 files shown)


datacenter/ucp/2.0/guides/installation/upgrade.md
```diff
@@ -78,7 +78,7 @@ that you've installed using that node's root CA material.
 
     ```bash
     $ docker run --rm -it \
-      --name ucp \
+      --name ucp --privileged\
```
joaofnfernandes: @alexmavr can you confirm that the UCP installer needs to run in privileged mode? 🤔

alexmavr:
No, that is incorrect. The UCP bootstrapper has never needed privileged. Let's close this to avoid confusing further reviewers. @clemenko why did you think UCP needs privileged?

clemenko:
It was due to an selinux issue.

PR #1396, opened 2017-01-26
make links to https://docs.docker.com relative

aduermael:
make links to https://docs.docker.com relative (all of them, not just the ones in /engine/extend & /engine/reference)


aduermael:
fixes link for #1068 
@mstanleyjones, @johndmulhausen

mstanleyjones:
This is OK for now but we should revisit this using a HTMLM parser later for safety. I'm worried about the edge cases where the URL gets caught wrapping across lines.

PR #1408, opened 2017-01-27
EG about sharing dir btw host and container

Henri-Lo:
Very basic example is lacking here about how to share a directory and where it appears so added the example

`docker run -it -v ~/Desktop:/Desktop r-base bash`

and demonstrating it with bash.

### Proposed changes

The current docs are very inaccessible, simple demos missing about how to share the host directories with the containers.

### DIFF
--- a/docker-for-mac/osxfs.md
+++ b/docker-for-mac/osxfs.md
@@ -37,24 +37,31 @@ Docker user who started the containers.
 
 ### Namespaces
 
-Much of the OS X file system that is accessible to the user is also available to
-containers using the `-v` bind mount syntax. For example, the command 
+Much of the macOS file system that is accessible to the user is also available to
+containers using the `-v` bind mount syntax. The following command runs a container
+from an image called `r-base` and shares the macOS user's `~/Desktop/` directory as
+`/Desktop` in the container.
 
-`docker run -it -v ~/Desktop:/Desktop r-base bash`
+```bash
+$ docker run -it -v ~/Desktop:/Desktop r-base bash
+```
 
-runs the Docker image called `r-base`, shares the Desktop of OSX in `/Desktop` and 
-starts bash. You can see the folder called Desktop inside the container such that
+The user's `~/Desktop/` directory is now visible in the container as a directory
+under `/`.
 
 `root@2h30fa0c600e:/# ls
 bin  boot  Desktop  dev  etc  home  lib  lib32	lib64  libx32  media  mnt  opt	proc  root  run  sbin  srv  sys  tmp  usr  var`
 
-and, by default, you can share files in `/Users`, `/Volumes`, `/private`, and `/tmp` 
-directly. To add or remove directory trees that are exported to Docker, use the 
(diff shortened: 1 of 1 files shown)


docker-for-mac/osxfs.md
```diff
@@ -38,10 +38,19 @@ Docker user who started the containers.
...
 Much of the OS X file system that is accessible to the user is also available to
-containers using the `-v` bind mount syntax. By default, you can share files in
-`/Users`, `/Volumes`, `/private`, and `/tmp` directly. To add or remove
-directory trees that are exported to Docker, use the **File sharing** tab in
-Docker preferences <img src="../images/whale-x.png"> -> **Preferences** ->
+containers using the `-v` bind mount syntax. For example, the command 
+
+`docker run -it -v ~/Desktop:/Desktop r-base bash`
+
+runs the Docker image called `r-base`, shares the Desktop of OSX in `/Desktop` and
```
mstanleyjones: The file path here is wrong. `/Desktop` is not the same as `~/Desktop`.

docker-for-mac/osxfs.md
```diff
@@ -38,10 +38,19 @@ Docker user who started the containers.
...
-containers using the `-v` bind mount syntax. By default, you can share files in
-`/Users`, `/Volumes`, `/private`, and `/tmp` directly. To add or remove
-directory trees that are exported to Docker, use the **File sharing** tab in
-Docker preferences <img src="../images/whale-x.png"> -> **Preferences** ->
+containers using the `-v` bind mount syntax. For example, the command 
+
+`docker run -it -v ~/Desktop:/Desktop r-base bash`
+
+runs the Docker image called `r-base`, shares the Desktop of OSX in `/Desktop` and 
+starts bash. You can see the folder called Desktop inside the container such that
```
mstanleyjones: starts `bash`.

s/such that/using the `ls`  command:

Also consider using `ls -c1` so that the listing will not be so wide. :) Then you can also truncate the listing maybe after `home` and use `...` to indicate that you truncated it.

docker-for-mac/osxfs.md
```diff
@@ -38,10 +38,19 @@ Docker user who started the containers.
...
+
+`root@2h30fa0c600e:/# ls
+bin  boot  Desktop  dev  etc  home  lib  lib32	lib64  libx32  media  mnt  opt	proc  root  run  sbin  srv  sys  tmp  usr  var`
+
+and, by default, you can share files in `/Users`, `/Volumes`, `/private`, and `/tmp`
```
mstanleyjones: Maybe rewrite this a little, to something like "You can configure Docker for Mac to always share one or more of the directories <list> with containers....

mstanleyjones:
Ping?

londoncalling:
@justincormack @rneugeba can you guys review this? I'll look at it also, but please have a look.

mstanleyjones:
LGTM, going to go ahead and merge this.

Issue #1405, opened 2017-01-26
Prefer `docker <object> <verb>` over `docker <verb>`


mstanleyjones:
### Problem description

To keep things consistent, prefer the format `docker <object> <verb>` where possible in examples. The old way still works.

cc/ @joaofnfernandes

joaofnfernandes:
This will probably be a lot of work. Maybe we should address this in three stages:
1. Fix get started tutorials, and top X doc pages with most page views
2. There are some reference pages like the one for `docker run` that have content maintained by hand. Start migrating that content to the new version.
3. Fix the rest of the docs

mstanleyjones:
I've got a local branch where I've been taking a different approach, doing semi-automated search-and-replace within the `engine/` subdirectory. I am concentrating on `docker container` and `docker image` as those are the two where the short form has been the most common. When I start the PR, it might make the most sense for each of us to push into the PR for our areas of responsibility. This may not get done for 13.1.

joaofnfernandes:
Yup, compose 3.1 is higher priority than this for the 1.13.1 release. This is maintenance work and if it doesn't fit on the 1.13.1 release timeline, that's ok.

PR #1410, opened 2017-01-27
Prefer 'docker <object> <verb>' syntax instead of 'docker <verb>'

mstanleyjones:
**To address other areas than `engine/`, please push separate commits into this PR.**

### Proposed changes

This commit changes instances within the engine/ subdirectory
and covers the following commands:

docker container:
- docker attach
- docker cp
- docker create
- docker commit
- docker diff
- docker exec
- docker export
- docker kill
- docker logs
- docker ls
- docker pause
- docker port
- docker rename
- docker restart
- docker rm
- docker run
- docker start
- docker stats
- docker stop
- docker top
- docker unpause
- docker wait

docker image:
- docker build
- docker history
- docker import
- docker load
- docker pull
- docker push
- docker save
- docker tag

The following commands are **not** covered because they can apply to
different Docker objects:

docker inspect
docker prune
docker rm
docker update

### Related issues (optional)

#1405


thaJeztah:
I'm in doubt if we should change `docker run` to `docker container run` (`docker run` will _not_ be deprecated).

True, `docker container run` was added for a _consistent_ naming, but `docker run` is so baked in the history of docker that I would consider `docker run` the "canonical", and `docker container run` the alternative.

nathanleclaire:
I'm strongly -1 on `docker container run` as well FWIW... Though I'm not sure the docs maintainers can do anything about it @thaJeztah ?  Where's the place to discuss it upstream?

johndmulhausen:
Please suspend work on this pending further discussion.

thaJeztah:
We can probably discuss here

/cc @dnephin @shykes

joaofnfernandes:
Reposting my opinion about this here.

In Docker 1.13 we made awesome progress on our CLI client:
* Almost all commands now follow the pattern `docker <object> <verb>`
* We worked hard for providing backwards compatibility

There is still work to be done, like for operations involving registries like `docker login, push, pull, search`, but this change ensures that commands are not ambiguous.

We live and breath Docker, so for us `docker run` is obvious. But for someone that's starting to learn our tech, it's totally not obvious if you're running a container, a service, and image, a stack, or something else. I've seen several new users confused about this when running informal user research.

Knowing about this, and also knowing that we need to continue working to decrease the barrier for devs and IT ops to get started with Docker means that our docs should follow suit with the changes made to the Docker CLI:

1. New users should be exposed to the new syntax when starting to learn Docker
2. We need to slowly start transitioning existing users to the new syntax. Our community creates way more content than us. This means that if we don't nudge our community to also use the new syntax, new users will for sure be exposed to the old one. This will cause confusion, like what I've seen happening with the `--network` and `--net` flags.

I also want to address some feedback here.

> docker run will not be deprecated

It won't, but it doesn't mean that our docs need to promote it.

> docker run is so baked in the history of docker

This is not a good argument. A better way to guide our decisions is to think how many users are currently using Docker, and how fast the adoption curve is. If we have 2x year over year adoption, that means that the number of new users learning Docker in 2017 is higher than all the current Docker users combined.

For all those users, "Docker history" is whatever UX we create for them today, not whatever we created in previous versions.

> I'm strongly -1 on docker container run as well

You can find part of that discussion in docker/docker#28527, and for sure it's an heated debate.

dnephin:
I think I agree that new for users it helps to be more explicit about the commands by using the full canonical names. As they get more familiar with the commands maybe they would benefit from [shorter aliases](https://github.com/docker/docker/pull/27468). I think the object-verb form should help users understand the different docker concepts.

Maybe in the command line reference section we can mention that some commands have top-level aliases (like `run`).

nathanleclaire:
Why use the CLI to try and strong-arm user understanding of the core concepts?  

Whether it's called `docker container run` or just `docker run`, that command still contains a TON of magic that's unrelated to _just_ working with containers (processes).

You'll still have to somehow explain this to people -- "Well, you `docker container run`, but actually what that `container run` command does is that in the background it authenticates to a registry if you're logged in, gets some stuff from that registry, creates a container filesystem and namespaces,  creates and attaches volumes, creates and attaches networks, starts the process, attaches to it, and then `wait`s"?

Taking this idea to its logical conclusion, you end up with something like this, which is probably a lot more clear, but a lot less usable:

```console
$ docker image pull ubuntu
$ docker container create ubuntu
asdfasdf
$ docker network connect bridge ubuntu
$ docker container start asdfasdf
$ docker container attach asdfasdf
# / 
```

While it'd be great to have those available to teach "Docker under the hood" / "the plumbing" more directly, I feel strongly the defaults should get the hell out of people's way and allow them to do their daily work faster.  (By the way, "just use an alias" / "people should just be scripting this" is a huge blowoff IMO, if people have to script or alias your CLI to make it usable, it's broken...)

johndmulhausen:
+1 to Nathan's comments. `docker run` is practically part of our brand. "Just run `docker run hello-world` and magic happens."

The docs team suddenly subjecting our users to a nascent taxonomy fetish -- when we have well-worn, widely adopted, never-to-be-deprecated, shorter, useful, and brand-iconic commands like `docker run` out there -- would be a mistake. 

In a docs team discussion we hashed it out a bit, and there are times when the more disambiguated commands are useful. For example, `docker rmi` and `docker rm` are ambiguous compared to `docker image delete` and `docker container delete`. 

So, we should roll out these changes in a sensible way that is the most helpful in all cases. 

In the case of `docker run` I'd say at most we can mention that it's an alias for `docker container run` when introducing it in high-level places like getting started and leave it at that. The CLI refdocs will cover the aliasing as well, and our job is done. 

For other commands that are as iconic: `docker pull` and `docker push` for example, I'd consider it fair game to do the same. 

For commands that do exhibit an unclear scope when the namespace is left out or have other disambiguation problems, we should do a sort of inverse -- introduce the command `docker image delete` and mention that `docker rmi` is an alias.

mstanleyjones:
Closing for now, in light of feedback.

PR #1416, opened 2017-01-27
Rename .github/CONTRIBUTING.md to CONTRIBUTING.md

mueller-ma:
So it matches the link in .github/PULL_REQUEST_TEMPLATE.md and is easier to find. It now has the same path like in docker/docker.


johndmulhausen:
So, I had to research where this file should live and this seems okay; we don't really surface contributing.md well without it being in root; it might behoove us to link to it from Readme.md as well. At any rate, thanks @mueller-ma :)

PR #1419, opened 2017-01-28
Engine updates 1.13.1

mstanleyjones:
### Proposed changes

Engine changes for 1.13.1

Engine 1.13.1


mstanleyjones:
I don't get why those first 5 commits are there. I rebased this and didn't get any issues....

engine/installation/linux/linux-postinstall.md
```diff
@@ -97,6 +97,69 @@ For information about the different storage engines, see
...
 your host's Linux distribution and available kernel drivers.
 
+## Enable experimental features
+
+Some features of Docker are introduced as experimental features for one or more
+releases. Previously, you needed to download and install an
+**experimental build** to use these features. In Docker 1.13 and higher,
+experimental features are included in all stable Docker builds and packages,
+but disabled by default. To enable experimental features, start the Docker
+daemon (`dockerd`) with the `--experimental` flag. You can do this temporarily
```
johndmulhausen: Links to the temp and perm instructions?

engine/installation/linux/linux-postinstall.md
```diff
@@ -97,6 +97,69 @@ For information about the different storage engines, see
...
+or permanently.
+
+> **Note**: Docker for Windows and Docker for Mac ship with experimental
+> features enabled by default.
```
thaJeztah: I think that's only for the beta releases, stable releases have them disabled? But you can enable them in the UI

@justincormack ?

engine/installation/linux/linux-postinstall.md
```diff
@@ -97,6 +97,69 @@ For information about the different storage engines, see
...
+
+1.  Edit the Docker daemon configuration file, which defaults to
+    `/etc/docker/daemon.json`.
+
+2.  Adhering to JSON syntax, add the key `experimental` with the value `true`.
+    The following example shows a very simple `daemon.json`, but yours may be
+    different. Be sure that all but the last key-value pairs have a comma at
+    the end.
+
+    ```none
```
thaJeztah: can you add "json" language hint here?

engine/installation/linux/linux-postinstall.md
```diff
@@ -97,6 +97,69 @@ For information about the different storage engines, see
...
+    ```bash
+    $ sudo /usr/bin/dockerd --experimental
+    ```
+
+The Docker daemon starts in the foreground. To stop it, use `Ctrl+C`.
+
+### Permanently enable experimental features
+
+> **Warning**: Experimental features are not appropriate for production
+> environments or workloads.
```
thaJeztah: Wonder if we should rephrase this, and explain why; experimental features should be as reliable as non-experimental, but design may change, or the feature as a whole may be decided to not leave experimental.

engine/installation/linux/linux-postinstall.md
londoncalling: @mstanleyjones @thaJeztah The Docker for Mac and Docker for Windows Beta channel releases ship with experimental features enabled by default, and you have the option to switch this off and use the last stable release of Engine. The Stable releases now also have the option to use experimental features as an option on the Docker Daemon, but experimental features are off by default on the Stable channel. 

This is all documented but I will to update those docs just a bit to underscore the fact that you have the same options on stable and on beta (https://docs.docker.com/docker-for-mac/#/experimental-mode, https://docs.docker.com/docker-for-windows/#/experimental-mode)

engine/installation/linux/linux-postinstall.md
londoncalling: @thaJeztah @mstanleyjones here is how I describe experimental features in both d4mac and d4win downloads: https://docs.docker.com/docker-for-mac/#/download-docker-for-mac, and I also link directly to the [Docker Experimental Features README,](https://github.com/docker/docker/tree/master/experimental) which I think is a good idea. You might do the same here.

justincormack:
We are currently shipping experimental enabled by default on both channels
since it became a daemon flag.

londoncalling:
@justincormack thanks for the clarification, I'll fix those descriptions in d4mac, 4win right now

engine/installation/linux/linux-postinstall.md
mstanleyjones: 👍

engine/installation/linux/linux-postinstall.md
mstanleyjones: 👍 I'll add links to my note.

engine/installation/linux/linux-postinstall.md
mstanleyjones: Great idea!

engine/installation/linux/linux-postinstall.md
mstanleyjones: I don't like to add it because there is a bug in Atom that makes it break horribly, and also it doesn't seem to act differently from `none` in our current CSS.

mstanleyjones: But I'll try it and see.

mstanleyjones:
Relates to https://github.com/docker/docker/issues/31009

PR #1420, opened 2017-01-28
Tabs-based organization

johndmulhausen:
We've talked about it; let's beat up on this together and make it great. :)

cc @hansfrederickbrown @mstanleyjones @londoncalling @sanscontext @Raphaelbouchard @jsouth @chikaando @joaofnfernandes


sanscontext:
The tabs are more visible, but I'd still like to put some sort of visual anchor on them. Not sure what we have access to.

londoncalling:
@johndmulhausen , I think the tabs look good. Here are things I'm concerned about:

* I'm not sure we closed the issue re: organization of Reference that we started discussing in the meeting.

* I don't see the rest of the guides for Docker for Mac and Docker for Windows, and those other pages are frequently browsed, fairly critical and updated frequently (e.g., Troubleshooting, Release Notes, FAQs, Networking, etc.). If we are going to break up those books, users need to know how to get to the rest of it. I can do that work, but let's plan it out.

londoncalling:
@johndmulhausen I'd suggest adding one other way of getting back to the docs "home page" (i.e., the top page of the Guides). I see now that you can get there by clicking on the "Guides" tab, but I think it would be worth having an additional way, as people are used to having the option of clicking on "Welcome to the Docs" at the top of the left nav.

Perhaps something in the top left that says "Docs Home". See example mockup attached. Let me know what you think, @johndmulhausen @mstanleyjones 

docs-home-icon

johndmulhausen:
@londoncalling I think that's probably too much; why don't we add a shoutout to the "Guides" section in @mstanleyjones Honeypot tour of the nav?

mstanleyjones:
While you're in here can you fix the padding on the search auto-complete div?

johndmulhausen:
@mstanleyjones Done

allpagelinks.md
```diff
@@ -1,5 +1,5 @@
 ---
-title: All Page Links
+layout: null
```
joaofnfernandes: This frontmatter sets `layout` to null twice. Delete line 4

PR #1429, opened 2017-01-30
zfs: fixed assumptions about thin provisioning

Mic92:
unlike lvm thin provisioning is fast and does not slow down the filesystem


mstanleyjones:
@thaJeztah can you please provide technical review for this, or delegate it?

PR #1436, opened 2017-01-30
Cloud: add note on machine users

sanscontext:
CLOUD-3921
🤖
@pkennedyr @marcusmartins @pchico83 maybe?


pkennedyr:
LGTM 🤖

Let's wait for confirmation from Pablo and Marcus before publishing. Thanks!

PR #1437, opened 2017-01-30
Add instructions for manual start of Docker daemon

onelovelyname:
Docker daemon did not automatically start for me; running `sudo docker run hello` resulted in the following error message: 
```bash
docker: Cannot connect to the Docker daemon at unix:///var/run/docker.sock. Is the docker daemon running?.
See 'docker run --help'.
```

After starting the docker daemon manually using `dockerd`, I saw the expected information.

### DIFF
--- a/engine/installation/linux/centos.md
+++ b/engine/installation/linux/centos.md
@@ -145,9 +145,13 @@ Docker from the repository.
     $ sudo yum -y install docker-engine-<VERSION_STRING>
     ```
 
-    The Docker daemon starts automatically.
+4.  Start Docker.
 
-4.  Verify that `docker` is installed correctly by running the `hello-world`
+    ```bash
+    $ sudo systemctl docker start
+    ```
+
+5.  Verify that `docker` is installed correctly by running the `hello-world`
     image.
 
     ```bash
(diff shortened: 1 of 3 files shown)


nathanleclaire:
It seems more likely it should be started via system service manager, no?  So in CentOS I'd expect that it's systemd and probably something like `sudo systemctl start docker`?

mstanleyjones:
I'm more curious to know why it didn't start automatically for you. What version of Docker were you installing and what version of CentOS? I'd like to try to replicate this.

Also, if this really is an issue with CentOS not starting the daemon automatically, we need to just tell people to start it, rather than give them a cautionary thing to do if it doesn't start. I'd prefer to avoid it altogether though, so if you can get me that info, it would be great!

nathanleclaire:
@mstanleyjones FWIW, IIRC some package managers won't start installed daemons / services automatically, so it's not unexpected to have to start them after the fact.

nathanleclaire:
I agree we should verify it's out-of-the-box behavior before covering in docs though 👍

mstanleyjones:
OK, verified that Docker does **not** start automatically on CentOS. Thank you bringing this up. I'm going to do the fix in a slightly different way than you have it, but I'll commit into your PR, if that's OK.

PR #1069, opened 2017-01-09
Add examples to CLI generation templates

mstanleyjones:
### Proposed changes

Add the examples for the CLI commands into the CLI-generation templates until they can be included as part of the YAML. Workaround as discussed with @johndmulhausen until the examples can be shipped to us in the YAML. cc/ @FrenchBen @dnephin 

### Unreleased project version (optional)

Engine 1.13

PR #1441, opened 2017-01-31
Include example from latest docker docs

FrenchBen:
### Proposed changes

Fixes the need to have hard-coded examples introduced in https://github.com/docker/docker.github.io/pull/1069 and adds them to the CLI as seen within the `docker/docker` repo

cc @johndmulhausen @mstanleyjones


mstanleyjones:
Interesting! Maybe @johndmulhausen can play with this. I think we might have to munge it a bit to make the TOC levels match.

johndmulhausen:
Let's get it in and I can munge later. Excellent work @FrenchBen :)

PR #1443, opened 2017-01-31
make docs - isn't this outdated ?

krasi-georgiev:
make doesn't seem to have docs target anymore so should I change the docs ?


johndmulhausen:
Yup, thanks @krasi-georgiev :)

PR #1450, opened 2017-01-31
Modify the in-page TOC

mstanleyjones:
### Proposed changes

This is one possible remedy for the current in-page TOC not being useful to people. It is only one possible solution.

- The in-page TOC is generated by Kramdown and Liquid, not by Javascript at page load time.

- By default, every pages gets an in-page TOC:
   - H2 headings only. The current TOC shows H2 and H3 and it gets very long and hard to read.
   - Between the reading time and the start of the content

- To opt out of the in-page TOC, add `notoc: true` to the frontmatter.
   This no longer deletes the search box. You can see it in action on the front page.

- This removes manually-maintained TOCs on a few pages that I knew had them

- This also changes the content that is in the right-hand nav. It's static, and it's the search, Edit this page link, Request docs changes link, and Get support link.

It is possible that some pages will have both the automatic and manually TOCs because I don't know all the pages where people have manually created them. In these cases, those authors would need to either remove the manually-generated TOC or set `notoc: true` for the page.

Another interesting thing this shows is the lack of parallelism among H2 headings within a page. Having the TOC so visible might help us to reorganize pages to show our true intent.

Check it out on Netlify.


londoncalling:
@mstanleyjones I don't think this is the best solution. I would suggest instead:

- remove ✎ Edit this page, ✓ Request docs changes, ? Get support from the right-hand nav. Having them all at the bottom of every page is enough.

- Replace the original "Edit this page" (which was sitting above the on-page TOC list) with a title of "On this page" or "Topics on this page" .. something like that

- Keep the H1s and H2s, they are very useful for pages with lots of topics e.g., d4mac, d4win release notes. Even if we shorten those pages, it's useful for relatively long pages.

- Remove the on-page lists on pages where they don't make sense (as you've done here)

- Ultimately, find a way of having the on-page TOC float/follow the scrolling if possible. If we only put the topic list at the top, we will never have this option. Also, if the topic list is even relatively long, having it only at the top is really less useful.

johndmulhausen:
I like it though I *think* we might want to hold off -- we're shipping some significant changes in 2 weeks.

sanscontext:
FWIW, I do _not_ like it. I would prefer to keep the original side-nav and just make it scroll with you/sticky.

mstanleyjones:
This is just a POC to see what it's like to put the TOC right in the reader's line of sight, since we pretty much proved it's currently in a blind spot for them. Along the way I fixed the problem where the search disappeared if you disabled the TOC on the page. I'm not 100% wedded to this idea. Regardless, I figured out how to inject the TOC using Kramdown and Liquid, and not relying on Javascript. I'll do another POC where we keep it in the right-hand nav and make it sticky.

I personally do not like having the H3 and H4 headings in the TOC, and it seems dumb to have the H1 heading in there since it's already the title of the page. The H3 and H4 ones are super cluttery IMO. But I think we can riff off this. I'd be happy to talk through with anyone what I was thinking, it might make more sense in person.

mstanleyjones:
Closed in favor of #1474

PR #1453, opened 2017-02-01
Update baseimages.md

j-wang:
Docs used a `:` instead of a `.` for `ubuntu:16.04`.

### Proposed changes

Change `ubuntu:16:04` to `ubuntu:16.04` in the example code, because the former is not a valid repository/tag spec and doesn't actually run. If attempted, the error message returned is:

`docker: Error parsing reference: "ubuntu:16:04" is not a valid repository/tag: invalid reference format.`


johndmulhausen:
Yep, thank you @j-wang :)

PR #1458, opened 2017-02-01
updates to voting app for better readability

londoncalling:
### Changes

* simplified several topics

* added in better implicit comparison of Dockerfiles vs. Compose files but tried to do so with less words rather than more

* reduced number of links, and made the existing links go to better targets

### Reviewers

@ManoMarks @mstanleyjones @johndmulhausen 


londoncalling:
LGTM, I'm going to go ahead and merge this to get it published, as I think the changes are rather uncontroversial and make it better. But please send review comments and we can iterate on it as needed.

PR #1460, opened 2017-02-01
Add documentation around pipefail to address issue 1459

### LINKED ISSUE #1459: Document best practice to avoid pipefail masking errors during image building
### Problem description

When writing a Dockerfile it is common to see lines such as:

```
RUN curl https://some.thing/file.tgz | tar -xf -
```

However what is not obvious is that if the first part of a pipe command fails, but the second still returns a successful return code, Docker will consider the command successful, cache the result, and continue the build.

For example:

```
FROM alpine:3.2
RUN wget https://wrong | wc
```

Succeeds (although the error is printed by `wget`, `wc` happily continues and returns an exit code of `0`)

```
Sending build context to Docker daemon 2.048 kB
Step 1/2 : FROM alpine:3.2
 ---> d95450e87d1e
Step 2/2 : RUN wget https://wrong | wc
 ---> Running in 4527cb186fb4
wget: bad address 'wrong'
        0         0         0
 ---> 617373b273b0
Removing intermediate container 4527cb186fb4
Successfully built 617373b273b0
```

Given how `RUN` is documented to operate, this behavior is technically correct and appropriate, however it doesn't feel obvious (it caught us out in our CI system) and I did find it surprising.

If a user does in fact want their build to fail if part of a pipe fails, this can be done in most cases by prepending the command with `set -o pipefail`, e.g.:

```
RUN set -o pipefail && wget https://wrong | wc
```

Results in a build failure as desired:

```
Sending build context to Docker daemon 2.048 kB
Step 1/2 : FROM alpine:3.2
 ---> d95450e87d1e
Step 2/2 : RUN set -o pipefail && wget https://wrong | wc
 ---> Running in e647fe34893a
wget: bad address 'wrong'
        0         0         0
The command '/bin/sh -c set -o pipefail && wget https://wrong | wc' returned a non-zero code: 1
```

### Problem location

- I couldn't find the information I wanted. I expected to find it near the following URL:
<https://docs.docker.com/engine/userguide/eng-image/dockerfile_best-practices/#/run>
and/or:
<https://docs.docker.com/engine/reference/builder/#run>

All

### Suggestions for a fix

Add documentation to the best practices to suggest that users running commands that pipe using consider using `set -o pipefail` in their `Dockerfile`.

I'd be happy to contribute a Pull Request along the lines of the above if you think this suggestion has merit.

aeijdenberg:
Signed-off-by: Adam Eijdenberg <adam.eijdenberg@gmail.com>

### Proposed changes

Add documentation around `pipefail` to address issue 1459.

### Related issues (optional)

Fixes #1459.

### DIFF
--- a/engine/userguide/eng-image/dockerfile_best-practices.md
+++ b/engine/userguide/eng-image/dockerfile_best-practices.md
@@ -283,6 +283,19 @@ keep the image size down. Since the `RUN` statement starts with
 > **Note**: The official Debian and Ubuntu images [automatically run `apt-get clean`](https://github.com/docker/docker/blob/03e2923e42446dbb830c654d0eec323a0b4ef02a/contrib/mkimage/debootstrap#L82-L105),
 > so explicit invocation is not required.
 
+#### Using pipes
+
+Where your `RUN` command depends on pipes, for example:
+
+
+	RUN wget -O - https://some.site | wc -l > /number
+
+Remember that Docker executes this with `/bin/sh -c` which considers the command successful if the last operation in the pipe is successful. In our example above this build step will succeed and result in a new image produced so long as `wc -l` succeeds, even if the `wget` fails.
+
+To instead fail on an error at any stage in the pipe, prepend `set -o pipefail &&` to ensure that an unexpected error prevents the build from inadvertently succeeding, for example:
+
+	RUN set -o pipefail && wget -O - https://some.site | wc -l > /number
+
 ### CMD
 
 [Dockerfile reference for the CMD instruction](../../reference/builder.md#cmd)
--- a/engine/userguide/eng-image/dockerfile_best-practices.md
+++ b/engine/userguide/eng-image/dockerfile_best-practices.md
@@ -285,16 +285,24 @@ keep the image size down. Since the `RUN` statement starts with
 
 #### Using pipes
 
-Where your `RUN` command depends on pipes, for example:
+Some `RUN` commands depend on the ability to pipe the output of one command into another, using the pipe character (`|`), as in the following example:
 
+```Dockerfile
+RUN wget -O - https://some.site | wc -l > /number
+```
 
-	RUN wget -O - https://some.site | wc -l > /number
+Docker executes these commands using the `/bin/sh -c` interpreter, which only evaluates the exit code of the last operation in the pipe to determine success. In the example above this build step succeeds and produces a new image so long as the `wc -l` command succeeds, even if the `wget` command fails.
 
-Remember that Docker executes this with `/bin/sh -c` which considers the command successful if the last operation in the pipe is successful. In our example above this build step will succeed and result in a new image produced so long as `wc -l` succeeds, even if the `wget` fails.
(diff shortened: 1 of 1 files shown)


engine/userguide/eng-image/dockerfile_best-practices.md
```diff
@@ -283,6 +283,19 @@ keep the image size down. Since the `RUN` statement starts with
 > **Note**: The official Debian and Ubuntu images [automatically run `apt-get clean`](https://github.com/docker/docker/blob/03e2923e42446dbb830c654d0eec323a0b4ef02a/contrib/mkimage/debootstrap#L82-L105),
 > so explicit invocation is not required.
 
+#### Using pipes
+
+Where your `RUN` command depends on pipes, for example:
```
mstanleyjones: Not all of our users are UNIX experts. Let's give them some context, and also not start with a sentence fragment:

Some `RUN` commands depend on the ability to pipe the output of one command into another, using the pipe character (`|`), as in the following example:

engine/userguide/eng-image/dockerfile_best-practices.md
```diff
@@ -283,6 +283,19 @@ keep the image size down. Since the `RUN` statement starts with
...
+
+
+	RUN wget -O - https://some.site | wc -l > /number
```
mstanleyjones: Please use the triple-tick fencing syntax, with `Dockerfile` hint, instead of the 4-space indentation.

engine/userguide/eng-image/dockerfile_best-practices.md
```diff
@@ -283,6 +283,19 @@ keep the image size down. Since the `RUN` statement starts with
...
+
+Remember that Docker executes this with `/bin/sh -c` which considers the command successful if the last operation in the pipe is successful. In our example above this build step will succeed and result in a new image produced so long as `wc -l` succeeds, even if the `wget` fails.
```
mstanleyjones: s/Remember that Docker executes this...successful./Docker executes these commands using the `/bin/sh -c` interpreter, which only evaluates the exit code of the last operation in the pipe to determine success.

s/In our example above/In the example above

s/will succeed and result/succeeds and produces a new image

s/the `wc -l` succeeds/the `wc -l` command succeeds (and the same for `wget`)

engine/userguide/eng-image/dockerfile_best-practices.md
```diff
@@ -283,6 +283,19 @@ keep the image size down. Since the `RUN` statement starts with
...
+
+To instead fail on an error at any stage in the pipe, prepend `set -o pipefail &&` to ensure that an unexpected error prevents the build from inadvertently succeeding, for example:
```
mstanleyjones: If you want the command to fail due to an error at any stage in the pipe,

aeijdenberg:
Hi @mstanleyjones , thanks for the suggested edits. I've changed as suggested and also added a note about shells that don't support `-o pipefail`.

If/when you're happy with the overall PR, I'm happy to rebase/squash the commits. Figured best keep them separate while we're still discussing changes.

PR #1462, opened 2017-02-01
Update Deploy to Azure button

tsubasaxZZZ:
This page has to have Azure button but now AWS button.
AWS page too.


mstanleyjones:
This was fixed in #1466, but thanks for your report!

PR #1468, opened 2017-02-01
Added a new remove unofficial docker step

tradfor:
I found an error when trying to install the official docker-engine, as there a second package that conflicts with the new package.  Added instructions on how to remove it.

### DIFF
--- a/engine/installation/linux/rhel.md
+++ b/engine/installation/linux/rhel.md
@@ -27,6 +27,13 @@ version of Docker, remove it using the following command:
 $ sudo yum -y remove docker
 ```
 
+You may also have to remove the package `docker-selinux` which will conflict with
+the new official `docker-engine` package.  Remove it with the following command:
+
+```bash
+$ sudo yum -y remove docker-selinux
+```
+
 The contents of `/var/lib/docker` are not removed, so any images, containers,
 or volumes you created using the older version of Docker are preserved.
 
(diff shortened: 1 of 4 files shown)


engine/installation/linux/rhel.md
```diff
@@ -27,6 +27,13 @@ version of Docker, remove it using the following command:
 $ sudo yum -y remove docker
 ```
 
+You may also have to remove the package `docker-selinux` which will conflict with
```
mstanleyjones: I did a quick Google and it looks like this applies to CentOS, Fedora, and Oracle as well (but on Oracle, the package name is `docker-engine-selinux`), so if you could add it to the CentOS and Oracle docs as well, that would be great.

mstanleyjones: s/will conflict/conflicts

engine/installation/linux/rhel.md
```diff
@@ -27,6 +27,13 @@ version of Docker, remove it using the following command:
...
+the new official `docker-engine` package.  Remove it with the following command:
```
mstanleyjones: s/new official/official

engine/installation/linux/rhel.md
tradfor: Done.

tradfor: Changed.

engine/installation/linux/rhel.md
tradfor: Changed.

mstanleyjones:
These look great, if you add Fedora, it will be good to go. Don't forget, that Fedora uses `dnf` instead of `yum`.

PR #1474, opened 2017-02-02
POC 2 for in-page TOC

mstanleyjones:
### Proposed changes

This approach is fully liquid. You can configure with:

- `no_toc: true`
- `toc_min: 2`
- `toc_max: 6`

The site defaults are 2 and 3 for the min and max.

This retains the Edit / Request / Support links on the side, as well.


londoncalling:
I like it.

mstanleyjones:
The last commit takes the `js/alljs.js` file from 1m to 28k by removing the inline paste of Angular.js and JQuery from the file. Angular doesn't appear to have been used at all. I added a CDN for JQuery.

This fixes the double-rendering of the TOC fonts and speeds the page-load time significantly (again).

mstanleyjones:
Had to give up on that last commit for now. At least the JS is getting an overhaul!

mstanleyjones:
This is ready to merge! Thanks @allejo for all of the help in the implementation!

PR #1187, opened 2017-01-19
Release Engine 1.13 documentation

mstanleyjones:
### Proposed changes

Rebased `vnext-engine` on `master`. After merging into master, `vnext-engine` needs to be reset to `master`.

### Unreleased project version (optional)

Engine 1.13

PR #369, opened 2016-10-27
Add NATS client as option for publishing container logs

wallyqs:
### Describe the proposed changes

Adds documentation of feature for enabling the Docker Engine to be able to use a [NATS](https://github.com/nats-io/gnatsd) transport for emitting the log entries under a customizable subject via `-log-driver nats` logging driver.
### Project version

Based on the `vnext-engine` branch
### Related issue or PR in another project

Related to proposed feature at: https://github.com/docker/docker/pull/27841

Signed-off-by: Waldemar Quevedo waldemar.quevedo@gmail.com

PR #1482, opened 2017-02-02
Adds details about new non-blocking logging mode

cpuguy83:
This functionality was added in docker/docker#28762.
It allows a user to specify what to do in cases when a log message can't
be sent, block the container's stdio or drop messages.

Also removes an erroneous entry about a `nats` log driver, which does
not exist.


cpuguy83:
ping @thaJeztah

mstanleyjones:
LGTM pending tech review. Where in the world did the `nats` driver content come from??

cpuguy83:
Not sure but it looks like it got cherry-picked in somehow in #1187

mstanleyjones:
The NATS stuff was added in #369. Should I just revert that change? It is my fault because I merged it without realizing that the feature was rejected.

mstanleyjones:
I can't revert #369 cleanly so I'll just merge this one, which effectively does the same thing.

PR #1483, opened 2017-02-02
Fix left-hand nav font re-render on page load

mstanleyjones:
### Proposed changes
I think this fixes the font re-render on opened nav menus on the left-hand nav. Have a page open like http://localhost:4000/engine/installation/ to test it. In this case, I chose to remove the earlier overridden definition rather than our later definition, since presumably we chose it mindfully.

PTAL @johndmulhausen


johndmulhausen:
Yessss, thank you @mstanleyjones !

PR #1487, opened 2017-02-03
Various Jekyll optimizations

mstanleyjones:
- Add `tree: true|false` front-matter key (default to true) to disable the left-hand nav tree on a per-page basis
- Disabling the tree also disables reading time
- Disabled tree on search, sorry, thank-you-subscribing-docker-weekly
- Use null layout for allpagelinks.md and limit it to pages that have a title (which leaves out redirects, which are not real pages)
- Add site-generation date and time to footer, in a tiny unobtrusive font

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -32,6 +32,7 @@ defaults:
       enginebranch: 1.13.x
       toc_min: 2
       toc_max: 3
+      tree: true
   - scope:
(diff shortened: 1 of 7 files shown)


johndmulhausen:
Looks great; merging so it's not blocked by the Netlify chicanery from today.

PR #1493, opened 2017-02-03
Fix left-hand nav font re-render on page load

mstanleyjones:
### Proposed changes

This is #1483 but without trying to yank out `css/app2.css` which broken the header and footer.

Let's wait to be sure that the Netlify test works before merging this. :)


mstanleyjones:
This looks good now. The only font change that is still there is when it switches from Helvetica to the web font after loading the web font, and that can't be helped.

PR #1497, opened 2017-02-04
move info about local users to note at the top

tfoxnc:
To prevent users from configuring LDAP and *then* realizing their local users are deleted, it should also be stated at the beginning that a recovery admin should be created and that all the local users are deleted.

### DIFF
--- a/datacenter/ucp/2.0/guides/configuration/integrate-with-ldap.md
+++ b/datacenter/ucp/2.0/guides/configuration/integrate-with-ldap.md
@@ -8,8 +8,10 @@ title: Integrate with LDAP
 Docker UCP integrates with LDAP services, so that you can manage users from a
 single place.
 
-**Note:** When you switch from the built-in authentication to using LDAP
-authentication, all manually created users become inactive. Use the **Recovery admin username** and **Recovery admin password** fields to create a local user account that can be used to access UCP if the LDAP server becomes unreachable.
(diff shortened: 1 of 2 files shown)


datacenter/ucp/2.0/guides/configuration/integrate-with-ldap.md
```diff
@@ -8,6 +8,9 @@ title: Integrate with LDAP
 Docker UCP integrates with LDAP services, so that you can manage users from a
 single place.
 
+**Note:** When you switch from the built-in authentication to using LDAP
```
mstanleyjones: Just a little bit of rewording:

s/from the built-in/from using built-in

s/manually created/local

Also, try to wrap your lines around 80 characters so that diffs are easier to review later on. Thanks!

joaofnfernandes:
@tfoxnc I think this one slipped through the cracks. I've reworded your proposal a bit and applied it to the docs of UCP 2.0 and 2.1.
When the checks pass, I'll merge it. Sorry for the delay.

datacenter/ucp/2.0/guides/configuration/integrate-with-ldap.md
joaofnfernandes: I've addressed most of the feedback here. I've not changed from 'manually created' to 'local' because we use that term consistently in more places.

PR #1498, opened 2017-02-04
Update fedora.md to correct systemd invocation

narfman0:
```systemctl docker start``` is not correct for systemd; it must be ```systemctl start docker```. This was probably a leftover from init script - an easy oversight when switching over.


mstanleyjones:
This is going to be fixed in #1523 for all affected Linux distributions. Thanks for reporting it!

PR #1500, opened 2017-02-04
fix starting docker command

shoji-k:
### Proposed changes

fix starting docker command


mstanleyjones:
This is going to be fixed in #1523 for all of the affected Linux distributions. Thanks for reporting it. By the way, I do not think the `.service` suffix is necessary.

PR #1503, opened 2017-02-04
typo in command

tillmanj:


mstanleyjones:
This is going to be fixed in #1523 for all of the affected Linux distributions. Thanks for reporting it!

PR #1505, opened 2017-02-04
Fix: Start Docker Section

pyontko-nazar:
'sudo systemctl docker start' doesn't work.
It should be 'sudo systemctl start docker'.
Details at http://www.unix.com/man-page/centos/1/systemctl


mstanleyjones:
This is going to be fixed in #1523 for all of the affected Linux distributions. Thanks for reporting it!

PR #1507, opened 2017-02-05
fix systemctl command error.

eacdy:
fix systemctl command error.


mstanleyjones:
This is going to be fixed in #1523 for all three affected distributions. Thanks for reporting it!

PR #1508, opened 2017-02-05
Corrected a typo and error in docker start command

sivakumar2205:
- Corrected a typo in comment.
- Corrected a command to start docker.

### DIFF
--- a/engine/installation/linux/fedora.md
+++ b/engine/installation/linux/fedora.md
@@ -57,7 +57,7 @@ Docker from the repository.
 
 #### Set up the repository
 
-1.  Install the `dnf-plugins.core` package which provides the commands to manage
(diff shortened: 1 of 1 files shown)


engine/installation/linux/fedora.md
```diff
@@ -152,7 +152,7 @@ Docker from the repository.
 4.  Start Docker.
 
     ```bash
-    $ sudo systemctl docker start
```
mstanleyjones: This part is going to be addressed in #1523, for all three affected distributions. If you remove that change from this PR, I'll merge this one. I checked and the problem in line 60 only affects Fedora. Thanks!

sivakumar2205: @mstanleyjones Updated the PR. Thanks.

PR #1515, opened 2017-02-05
Move to GHP 112, nginx-based server, docs-base

johndmulhausen:
1) The Gemfile was pinned to GH Pages 105 but the production image was running "latest" -- now they're both pinned to 112 so they stay in sync. 112 was the most recent, specifically-tagged version of starefossen/github-pages on Docker Hub.

2) Moves to nginx, a production-ready static server, over `jekyll serve` by grabbing it from the Jessie aptitude repo and installing it in userland. Seems to work fine, no `sudo` required. 

3) Creates a docs/docs-base image that contains github-pages, nginx, and the built archives -- and changes the production Dockerfile to use `FROM docs/docs-base` which massively improves build times.

As a result of this: 

- Random updates to gh-pages don't make our site unbuildable; both the local Gemfile and the production environment are pinned to 112
- We got some neat new plugins/capabilities from GH Pages release 112
- Our build times are massively reduced by the use of a base image that has the archives pre-built
- We are using a production-ready server, nginx, instead of `jekyll serve` which is not meant to handle what we do. The use of nginx opens up a lot of possibilities for better site performance, such as true 301 redirects, etc.


Dockerfile
```diff
@@ -46,7 +46,8 @@ RUN svn co https://github.com/docker/docker/branches/$ENGINE_BRANCH/docs/extend
  && jekyll build -s allv -d allvbuild \
  && rm -rf allvbuild/apidocs/layouts \
  && find allvbuild -type f -name '*.html' -print0 | xargs -0 sed -i 's#href="https://docs.docker.com/#href="/#g' \
- && rm -rf allv
+ && rm -rf allv \
+ && apt-get update \
+ && apt-get -y install nginx
```
thaJeztah: I suggest to;

- move installation of nginx to a separate `RUN`, as it's not related to the generation of the archives. The installation step can be performed _before_ generation of the archives
- to stick close to the official nginx image, which is also based on `debian:jessie`, we can copy the same approach as that image; https://github.com/nginxinc/docker-nginx/blob/e950fa7dfcee74933b1248a7fe345bdbc176fffb/mainline/jessie/Dockerfile#L5-L23
- some of those modules are not needed, so we can omit those, so it'll be something like;

```Dockerfile
ENV NGINX_VERSION 1.11.9-1~jessie

RUN apt-key adv --keyserver hkp://pgp.mit.edu:80 --recv-keys 573BFD6B3D8FBC641079A6ABABF5BD827BD9BF62 \
	&& echo "deb http://nginx.org/packages/mainline/debian/ jessie nginx" >> /etc/apt/sources.list \
	&& apt-get update \
	&& apt-get install --no-install-recommends --no-install-suggests -y \
						ca-certificates \
						nginx=${NGINX_VERSION} \
	&& rm -rf /var/lib/apt/lists/*

# forward request and error logs to docker log collector
RUN ln -sf /dev/stdout /var/log/nginx/access.log \
	&& ln -sf /dev/stderr /var/log/nginx/error.log
```

Dockerfile
```diff
@@ -46,7 +46,8 @@ RUN svn co https://github.com/docker/docker/branches/$ENGINE_BRANCH/docs/extend
...
 
-# Serve the site, which is now all static HTML
-CMD jekyll serve -s /usr/src/app/allvbuild --no-watch -H 0.0.0.0 -P 4000
+CMD echo "Server running at http://0.0.0.0:4000" && nginx -c /usr/src/app/allvbuild/nginx.conf
```
thaJeztah: Probably cleaner to copy the nginx.conf to it's default location;

```Dockerfile
COPY nginx.conf /etc/nginx/
```

Also nginx is started in a shell here, therefore not being PID1 (this line will run `/bin/sh -c 'echo .. && nginx -c ...`).

If we are ok to omit the `"Server running at .."`, change to;

```Dockerfile
CMD ["nginx", "-g", "daemon off;"]
```

Otherwise, I _think_ this will work (adding `exec`);

```Dockerfile
CMD echo "Server running at  http://0.0.0.0:4000" && exec nginx -c /usr/src/app/allvbuild/nginx.conf
```

You can verify it's working correctly by running the container, and checking the `nginx` is PID1

Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM starefossen/github-pages
+FROM starefossen/github-pages:112
```
mstanleyjones: What do we get by updating?

johndmulhausen: Besides just stability and ensuring we ship a site that actually builds, these:

- New plugin: jekyll-include-cache
- Bump jekyll-titles-from-headings to v0.1.3
- Bump jekyll-readme-index to v0.0.3
- The following enabled-by-default plugins (+ a default theme):

Jekyll Optional Front Matter - Don't require optional front matter to process a page as a page
Jekyll README Index - If a repo has a readme, but not an index, use the readme as the index
Jekyll Default Layout - Set a default per-page layout if no layout is specified and a logical one exists for the site
Jekyll Titles from Headings - If there is no YAML title, use the first heading as the page's title
- Jekyll Relative Links

johndmulhausen:
@thaJeztah OK, I think I have addressed your feedback; the `exec` thing worked, the log output is showing in the terminal now because of that change, which is great, and we're using the official nginx instructions from their Dockerfile (thanks for paring those down).

Dockerfile
```diff
@@ -2,6 +2,24 @@ FROM starefossen/github-pages:112
 
 ENV VERSIONS="v1.4 v1.5 v1.6 v1.7 v1.8 v1.9 v1.10 v1.11 v1.12"
```
thaJeztah: I'd put every docs/branch related ENV-vars after the installation of NGINX; we could then even decide to create a base-image containing Jekyll and NGINX

johndmulhausen:
@thaJeztah Done

mstanleyjones:
I'm testing this locally now.

johndmulhausen:
@thaJeztah: @mstanleyjones couldn't use the more involved install instructions for nginx, but `apt-get update && apt-get -y install nginx` worked fine. It seemed to work fine last night but perhaps something has changed on the server end. Regardless, in the more involved instructions we're using a specific GPG key to enable that install and doing custom repo installation by echoing into config files, etc -- maybe it's a good idea just to use the Debian distro's `apt-get` package, whatever that is? We're not doing anything too fancy.

mstanleyjones:
I added a commit on top of this that does a few things:

- Renames 'allvbuild' to 'target'
- Renames 'allv' to 'md_source' (when building master) and 'archive_source' (when dealing with the archives). These directories get deleted at the end of each layer so it doesn't really matter what they're called, but it's a little less confusing
- Makes sure that all intermediate directories get nuked
- Adds `_site` to the Dockerignore because it's very likely to be present when building the Dockerfile locally as a one-off
- Changes the order of the build so that we do `master` first. It made more sense to me when thinking about the temporary directories and making sure that what we pass down to the next layer is only in a single directory.

You can use `docker history` and see that this improves the sizes of the layers:

**Before**: 

```
a3d80c092520        19 seconds ago      /bin/sh -c svn co https://github.com/docke...   579 MB              
3638f49523fb        3 minutes ago       /bin/sh -c #(nop) COPY dir:fff0d7e4749adba...   87.9 MB             
ba89de24c06b        3 minutes ago       /bin/sh -c git clone https://www.github.co...   736 MB 
```

**After**:

```
a88705620282        About a minute ago   /bin/sh -c git clone https://www.github.co...   736 MB              
3ba456f16f6f        7 minutes ago        /bin/sh -c mv allvbuild allv  && svn co ht...   205 MB              
d7cbe5f24528        11 minutes ago       /bin/sh -c #(nop) COPY dir:fff0d7e4749adba...   87.9 MB  
```

It shaves off about 300MB from the step where we bring in all the archives. Also the copy went down from 720 MB in master (but that is really variable, because you may or may not have `_site` or any uncommitted files in your context).

mstanleyjones:
PS if you don't like my commit, just do a hard-reset in the `nginx` branch to the SHA before it. I didn't mean to force my changes into the PR.

mstanleyjones:
@nathanleclaire I'd love it if you could take a look and see if we are doing anything else silly. I appreciated your feedback yesterday.

mstanleyjones:
:1st_place_medal: for that last commit!

PR #1518, opened 2017-02-06
Fix systemctl commands

jrdnull:
### Proposed changes

Updated systemctl commands as they do not work


mstanleyjones:
I think this is a duplicate of #1523. Thanks for reporting it!

PR #1521, opened 2017-02-06
Update centos.md

farrukhnajmi:
Fixing typo in syntax for systemctl command to start docker service.


mstanleyjones:
I think this is a duplicate of #1523. Thanks for reporting it!

PR #1522, opened 2017-02-06
Update rhel.md

helgein:
Corrected sytemctl calls and added hint about how to configure a proxy.


helgein:
I do not thing it is wise that i modify documentation regarding distributions i am not hands on personally.
As for the proxy config, please feel free to replace it with a link to the appropriate paragraph.

mstanleyjones:
The proxy server settings are already covered in the [postinstallation](https://docs.docker.com/engine/installation/linux/linux-postinstall/#systemd) topic.

PR #1525, opened 2017-02-06
Fixed bug in posh-docker setup instructions

hjoelr:
There were two problems that this fixes.

1. Before this change, posh-docker was being installed twice. Once for the system (via NuGet) and secondly in the user's PowerShell context. This was confusing. I understand the reason was probably to create the user's $PROFILE file if it didn't already exist. I've replaced that functionality with clearer code.

2. The instructions had incorrectly said to add "Install-Module posh-docker" to the $PROFILE. Instead, this needed to be "Import-Module posh-docker".

### DIFF
--- a/docker-for-windows/index.md
+++ b/docker-for-windows/index.md
@@ -400,15 +400,18 @@ PowerShell Module as follows.
     command to a `$PROFILE` by typing these commands at the PowerShell prompt.
 
     ```none
-    Install-Module -Scope CurrentUser posh-docker -Force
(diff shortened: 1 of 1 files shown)


mstanleyjones:
@londoncalling @friism PTAL

londoncalling:
@hjoelr Thanks, Joel. I checked my `$PROFILE` and you are right, it's got `Import-Module posh-docker` in it. So, it sounds like these corrections you made are all great. Merging.

PR #1529, opened 2017-02-06
Adds a Hopscotch tour of the new navigation introduced in #1420

mstanleyjones:
This patch depends on #1420. It adds a 4-step on-screen tour of the navigation enhancements introduced in #1474 and #1420. The tour is implemented using http://linkedin.github.io/hopscotch/.

The tour is triggered by clicking on a button, which displays in the upper right of the new tab bar. Because the tour is expected to be run from the welcome page, which doesn't have an in-page TOC, the 4th step of the tour, which shows you the in-page TOC, goes to `/learn/index.html`. When you end the tour, or if you stop it before it's finished, you are returned back to the page where you started.

I think it would be great to re-do the tour when we have the big site re-skin, as well.

Don't merge this until #1420 is merged! I wanted to make the PR now to let Netlify build it, so people could try it out.


js/my_first_tour.js
```diff
@@ -0,0 +1,75 @@
...
+      content: "Use the left navigation for a structured view of content within a top-level category.",
+      target: "left-nav",
+      placement: "right"
+    },
+    {
+      title: "Feedback Links",
+      content: "Use the feedback links to edit the page, provide feedback, or find out how to get support.",
+      target: "feedback-links",
+      placement: "left",
+      multipage: "true",
```
mstanleyjones: For these three lines, I really only need to do it this way if the page they start the tour on doesn't have the div `#side-toc`. But I'm not sure how to do that without duplicating the whole `var tour` with two variants. @jsouth or @johndmulhausen do you have any suggestions?

js/my_first_tour.js
```diff
@@ -0,0 +1,75 @@
+// Define the tour!
+var tour = {
+  id: "hello-hopscotch",
+  steps: [
+    {
+      title: "Navigation improvements!",
```
mstanleyjones: I am not attached to the text in the titles, contents, buttons, etc. I am sure others can come up with better text.

londoncalling:
@mstanleyjones I think this is great! It is such a nice, concise tour, that it might not be a bad idea to extend it a bit. 

How about adding a tab or even 4 more to briefly describe the content on each of the tabs? It will even better queue people that there are several tabs there, and give them an idea of what they will find on each one.

mstanleyjones:
@londoncalling PTAL at the latest version when it builds. I added a step for each of the horizontal tabs and I also tweaked the alignment of the arrows on some of the bubbles.

js/my_first_tour.js
```diff
@@ -0,0 +1,120 @@
...
+      title: "Glossary",
+      content: "Use the <b>Glossary</b> tab to learn terminology specific to Docker.",
+      target: "top-nav",
+      placement: "bottom",
+      width: "570px",
+      arrowOffset: "280px"
+    },
+    {
+      title: "Reference",
+      content: "Use the <b>Reference</b> tab to go to Docker reference information, such as API, CLI, and commands topics.",
```
mstanleyjones: I think "CLI" and "command" refer to the same thing, right? Other than that, these changes look good, @londoncalling.

mstanleyjones:
Rebased on current iteration of `tabs` branch and squashed commits. Let me try my hand at theming.

londoncalling:
@mstanleyjones I had a thought about this last night. I think the text/font should be bigger.

londoncalling:
@mstanleyjones looking good! I would just say some of the content could be shortened up, i.e., no need to repeat titles of the tabs etc. in the content twice. I can make the changes here, but why does it say "some checks not successful"? Should I go ahead and make changes and not worry about that?

johndmulhausen:
Can we change the font face of the Tour button to be the more plain-jane, serious font used on the buttons for the Getting Started links on the front page? There's just something off about the way that button looks now with that font. I know you're working on the centering/positioning too, so -- while you're in there?

londoncalling:
@mstanleyjones okay, let me know what you think of the new wording :-) the tour looks great, you are brilliant

mstanleyjones:
Closed in favor of #1420

PR #1535, opened 2017-02-07
update ubuntu installation instructions

thaJeztah:
the `software-properties-common` package is already
installed in step one, so there's not need to re-install
it in step 3.

Also moved the instructions for adding the "testing" repository
after the instructions for adding the "main" repository. The
warning "Do not use these unstable repositories on production
systems or for non-testing workloads." war prominently visible,
and therefore gave the impression it was refering to the instructions below;

screen shot 2017-02-06 at 7 20 28 pm

### DIFF
--- a/engine/installation/linux/ubuntu.md
+++ b/engine/installation/linux/ubuntu.md
@@ -87,19 +87,18 @@ Docker from the repository.
       uid                  Docker Release Tool (releasedocker) <docker@docker.com>
     ```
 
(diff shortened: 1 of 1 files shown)


mstanleyjones:
If I am right, I inadvertently fixed these in #1560 . If so, can you close this one? Sorry about that!

thaJeztah:
Sure! I noticed, let me close this one (i just stumbled upon these while assisting a user)

Issue #1516, opened 2017-02-06
Feedback for: engine/installation/linux/ubuntu.md


sravan-s:
[engine/installation/linux/ubuntu.md](https://docs.docker.com/engine/installation/linux/ubuntu/)

I'm using Linux Mint Rafaela. Since mint is a distribution based on Ubuntu, I believe people would usually use this page

Now this part
```
sudo add-apt-repository \
       "deb https://apt.dockerproject.org/repo/ \
       ubuntu-$(lsb_release -cs) \
       main"  
```

Creates an additional-repositories.list file saying:
`deb https://apt.dockerproject.org/repo/        ubuntu-rafaela        main`
Which in turn throws 403 forbidden error while running **apt-get update**
And I wasn't able to finish my installation

But when I used 

```
sudo add-apt-repository \
       "deb https://apt.dockerproject.org/repo/ \
        ubuntu-precise \
       main"  
```
Everything worked fine.
I guess it would be good if we add a pointer saying **ubuntu-$(lsb_release -cs)** means give the stable version of your parent distribution

What are your opinions?

sravan-s:
@mstanleyjones ,

Is it fine, if I try to contribute to this?

sravan-s:
@mstanleyjones,
Is it fine to close this issue now?

mstanleyjones:
Yep! In the future, if you add a line like "Fixes #<issue_number>", Github will do this automatically when your PR is closed.

PR #1537, opened 2017-02-07
Update repository setup step for Linux Mint users

sravan-s:
### Proposed changes
Add a note to remind people using distributions derived from Ubuntu(Linux Mint)
to change the ubuntu-$(lsb_release -cs) part to their parent Ubuntu release

The output of $(lsb_release -cs) won't necessarily be a stable Ubuntu release, it would
sometimes be something like 'rafaelia' and there could be problems in the installation

### Related issues (optional)
https://github.com/docker/docker.github.io/issues/1516

### DIFF
--- a/engine/installation/linux/ubuntu.md
+++ b/engine/installation/linux/ubuntu.md
@@ -101,6 +101,11 @@ Docker from the repository.
 
     To disable the `testing` repository, you can edit `/etc/apt/sources.list`
     and remove the word `testing` from the appropriate line in the file.
+    
+    > **Note**: Sometimes, in a distribution like Linux Mint, you might have
+    > to change **ubuntu-$(lsb_release -cs)** to your parent Ubuntu distribution.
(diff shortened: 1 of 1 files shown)


engine/installation/linux/ubuntu.md
```diff
@@ -101,6 +101,11 @@ Docker from the repository.
 
     To disable the `testing` repository, you can edit `/etc/apt/sources.list`
     and remove the word `testing` from the appropriate line in the file.
+    
+    > **Note**: Sometimes, in a distribution like Linux Mint, you might have
+    > to change **ubuntu-$(lsb_release -cs)** to your parent Ubuntu distribution.
```
mstanleyjones: Please use backticks instead of asterisks, so this will be monotype instead of bold. Thanks!

sravan-s: @mstanleyjones I changed it. IS this good? :)

mstanleyjones:
Thank you! 👍

PR #1544, opened 2017-02-07
Fix markdown rendering of example configuration

pedep:
### Proposed changes

Fix markdown formatting of example config in compose/compose-file.md in the `update_config` and `restart_policy` section.
Added comments between list and code block. This renders the lines as a code block, instead of a paragraph within a list item. The problem only happens when a code block is used directly after a list.

stackexchange on the topic:
http://meta.stackexchange.com/questions/34292/code-blocks-after-a-list-but-not-within-a-list-in-markdown-is-it-possible

### DIFF
--- a/compose/compose-file.md
+++ b/compose/compose-file.md
@@ -256,11 +256,11 @@ updates.
 - `monitor`: Duration after each task update to monitor for failure `(ns|us|ms|s|m|h)` (default 0s).
 - `max_failure_ratio`: Failure rate to tolerate during an update.
 
-<!-- fix code block after list -->
-
+```
     update_config:
       parallelism: 2
       delay: 10s
+```
 
 #### resources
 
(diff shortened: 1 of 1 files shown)


compose/compose-file.md
```diff
@@ -256,9 +256,11 @@ updates.
 - `monitor`: Duration after each task update to monitor for failure `(ns|us|ms|s|m|h)` (default 0s).
 - `max_failure_ratio`: Failure rate to tolerate during an update.
 
+```
```
pedep: @johndmulhausen if you wish to use this syntax, you have to remove the whitespace in front

johndmulhausen: I thought of that; arguably, the whitespace belongs, since this section of YAML does happen at a certain indentation level. What do you think?

pedep: Now that you mention it, it seems to fit quite nicely - Since the general text is indented due to the list, keeping this whitespace keeps it more uniform in general. 

But overall, either way should be fine. Actually having formatting, be it indented or not, makes the examples much more readable.

johndmulhausen:
@pedep Thanks for the PR and the chat! Come back anytime :)

pedep:
Cheers mate :)

PR #1549, opened 2017-02-07
Expand menu items after 'click', not 'touchend'

### LINKED ISSUE #1128: Pure scrolling annoyingly expands menu items
When scrolling through the docs on my iPhone, menu items are expanded. The shall not be expanded by touchdown/up when scrolling, only when clicking. It is extremely annoying and makes the docs almost unreadable in terms of user experience.

### Problem location

- Other: <DETAILS>

Enselic:
The problem with using 'touchend' is that if you scroll with your finger on a touch screen, like an an iPhone or on a PC laptop with a touch screen, the menu items will be expanded when you let go, even if you wanted to just scroll.

Fixes #1128

### DIFF
--- a/js/menu.js
+++ b/js/menu.js
@@ -44,7 +44,7 @@ function matches(inputTxt,searchTxt)
 function hookupTOCEvents()
 {
   // do after tree render
-  $('.expand-menu').on('mouseup touchend', function(elem) {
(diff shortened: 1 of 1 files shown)


johndmulhausen:
Thanks, @Enselic !

PR #1553, opened 2017-02-08
bump engine API to 1.26

vieux:
see https://github.com/docker/docker/pull/30806 for details


johndmulhausen:
Thanks, @vieux -- this LGTM. When the API ships, ping us and we'll make sure this gets pushed. :)

PR #1566, opened 2017-02-09
Add info about backing up swarms

### LINKED ISSUE #1072: Feedback for: engine/swarm/admin_guide.md
[engine/swarm/admin_guide.md](https://docs.docker.com/engine/swarm/admin_guide/)

On this section: https://docs.docker.com/engine/swarm/admin_guide/#back-up-the-swarm-state

It looks like we can backup a docker swarm cluster easily:
`cp -r /var/lib/docker/swarm/raft /tmp/backup && tar ... && scp ....`

#### Here are my questions and my assumptions:

1) Does the backup process care about online vs offline backups? I assume we can do a hot backup
2) Can I simply cp -r or rsync that folder without worrying about the integrity of the backup? I assume yes.
3) Do I need to disable the swarm cluster to ensure an accurate backup? I assume it's no.
4) Can I ensure that the swarm state is consistent at backup time? No idea
5) Does the backup is reliable no matter which manager it's made from ? Or should we do it from the leader node ?

Thanks!

mstanleyjones:
### Proposed changes

Add more explicit info about backup and recovery as outlined in #1072 

Some clean-up and reorganization along the way

Fixes #1072

### DIFF
--- a/engine/swarm/admin_guide.md
+++ b/engine/swarm/admin_guide.md
@@ -11,24 +11,11 @@ for managing the swarm and storing the swarm state. It is important to
 understand some key features of manager nodes in order to properly deploy and
 maintain the swarm.
 
-This article covers the following swarm administration tasks:
-
-* [Using a static IP for manager node advertise address](#use-a-static-ip-for-manager-node-advertise-address)
-* [Adding manager nodes for fault tolerance](#add-manager-nodes-for-fault-tolerance)
-* [Distributing manager nodes](#distribute-manager-nodes)
-* [Running manager-only nodes](#run-manager-only-nodes)
-* [Backing up the swarm state](#back-up-the-swarm-state)
-* [Monitoring the swarm health](#monitor-swarm-health)
-* [Troubleshooting a manager node](#troubleshoot-a-manager-node)
-* [Forcefully removing a node](#force-remove-a-node)
-* [Recovering from disaster](#recover-from-disaster)
-* [Forcing the swarm to rebalance](#forcing-the-swarm-to-rebalance)
-
 Refer to [How nodes work](/engine/swarm/how-swarm-mode-works/nodes.md)
 for a brief overview of Docker Swarm mode and the difference between manager and
 worker nodes.
 
-## Operating manager nodes in a swarm
+## Operate manager nodes in a swarm
 
 Swarm manager nodes use the [Raft Consensus Algorithm](/engine/swarm/raft.md) to manage the
 swarm state. You only need to understand some general concepts of Raft in
@@ -45,7 +32,7 @@ Raft requires a majority of managers, also called the quorum, to agree on
 proposed updates to the swarm, such as node additions or removals. Membership
 operations are subject to the same constraints as state replication.
 
-### Maintaining the quorum of managers
+### Maintain the quorum of managers
 
 If the swarm loses the quorum of managers, the swarm cannot perform management
 tasks. If your swarm has multiple managers, always have more than two. In order
@@ -63,7 +50,7 @@ updated.
 See [Recovering from losing the quorum](#recovering-from-losing-the-quorum) for
 troubleshooting steps if you do lose the quorum of managers.
 
-## Use a static IP for manager node advertise address
+## Configure the manager to advertise on a static IP address
 
 When initiating a swarm, you have to specify the `--advertise-addr` flag to
 advertise your address to other manager nodes in the swarm. For more
@@ -107,7 +94,7 @@ While it is possible to scale a swarm down to a single manager node, it is
 impossible to demote the last manager node. This ensures you maintain access to
 the swarm and that the swarm can still process requests. Scaling down to a
 single manager is an unsafe operation and is not recommended. If
-the last node leaves the swarm unexpetedly during the demote operation, the
+the last node leaves the swarm unexpectedly during the demote operation, the
 swarm will become unavailable until you reboot the node or restart with
 `--force-new-cluster`.
 
@@ -115,7 +102,7 @@ You manage swarm membership with the `docker swarm` and `docker node`
 subsystems. Refer to [Add nodes to a swarm](/engine/swarm/join-nodes.md) for more information
 on how to add worker nodes and promote a worker node to be a manager.
 
-## Distribute manager nodes
+### Distribute manager nodes
 
 In addition to maintaining an odd number of manager nodes, pay attention to
 datacenter topology when placing managers. For optimal fault-tolerance, distribute
@@ -131,7 +118,7 @@ available to process requests and rebalance workloads.
 | 7                   |                  3-2-2                 |
 | 9                   |                  3-3-3                 |
 
-## Run manager-only nodes
+### Run manager-only nodes
 
 By default manager nodes also act as a worker nodes. This means the scheduler
 can assign tasks to a manager node. For small and non-critical swarms
@@ -154,18 +141,15 @@ When you drain a node, the scheduler reassigns any tasks running on the node to
 other available worker nodes in the swarm. It also prevents the scheduler from
 assigning tasks to the node.
 
-## Back up the swarm state
+## Add worker nodes for load balancing
 
-Docker manager nodes store the swarm state and manager logs in the following
-directory:
-
-```bash
-/var/lib/docker/swarm/raft
-```
-
-Back up the `raft` data directory often so that you can use it in case of
-[disaster recovery](#recover-from-disaster). Then you can take the `raft`
-directory of one of the manager nodes to restore to a new swarm.
+[Add nodes to the swarm](/engine/swarm/join-nodes.md) to balance your swarm's
+load. Replicated service tasks will be distributed across the swarm as evenly as
+possible over time, as long as the worker nodes are matched to the requirements
+of the services. When limiting a service to run on only specific types of nodes,
+such as nodes with a specific number of CPUs or amount of memory, remember that
+worker nodes that do not meet these requirements will not be able to run these
+tasks.
 
 ## Monitor swarm health
 
@@ -232,12 +216,14 @@ To cleanly re-join a manager node to a cluster:
 For more information on joining a manager node to a swarm, refer to
 [Join nodes to a swarm](/engine/swarm/join-nodes.md).
 
-## Force remove a node
+## Forcibly remove a node
 
-In most cases, you should shut down a node before removing it from a swarm with the `docker node rm` command. If a node becomes unreachable, unresponsive, or compromised you can forcefully remove the node without shutting it down by passing the `--force` flag. For instance, if `node9` becomes compromised:
+In most cases, you should shut down a node before removing it from a swarm with
+the `docker node rm` command. If a node becomes unreachable, unresponsive, or
+compromised you can forcefully remove the node without shutting it down by
+passing the `--force` flag. For instance, if `node9` becomes compromised:
 
-<!-- bash hint breaks block quote -->
-```
+```none
 $ docker node rm node9
 
 Error response from daemon: rpc error: code = 9 desc = node node9 is not down and can't be removed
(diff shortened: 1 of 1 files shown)


engine/swarm/admin_guide.md
```diff
@@ -11,24 +11,11 @@ for managing the swarm and storing the swarm state. It is important to
 understand some key features of manager nodes in order to properly deploy and
 maintain the swarm.
 
-This article covers the following swarm administration tasks:
```
mstanleyjones: Removed this manually-maintained TOC because the right-hand TOC is sufficient

engine/swarm/admin_guide.md
```diff
@@ -11,24 +11,11 @@ for managing the swarm and storing the swarm state. It is important to
...
-* [Forcefully removing a node](#force-remove-a-node)
-* [Recovering from disaster](#recover-from-disaster)
-* [Forcing the swarm to rebalance](#forcing-the-swarm-to-rebalance)
-
 Refer to [How nodes work](/engine/swarm/how-swarm-mode-works/nodes.md)
 for a brief overview of Docker Swarm mode and the difference between manager and
 worker nodes.
 
-## Operating manager nodes in a swarm
+## Operate manager nodes in a swarm
```
mstanleyjones: Prefer non-gerunds for headings

engine/swarm/admin_guide.md
```diff
@@ -45,7 +32,7 @@ Raft requires a majority of managers, also called the quorum, to agree on
 proposed updates to the swarm, such as node additions or removals. Membership
 operations are subject to the same constraints as state replication.
 
-### Maintaining the quorum of managers
+### Maintain the quorum of managers
```
mstanleyjones: Put this beneath the fault-tolerance part

engine/swarm/admin_guide.md
```diff
@@ -154,18 +141,15 @@ When you drain a node, the scheduler reassigns any tasks running on the node to
...
 
-## Back up the swarm state
-
-Docker manager nodes store the swarm state and manager logs in the following
-directory:
-
-```bash
-/var/lib/docker/swarm/raft
-```
+## Add worker nodes for load balancing
```
mstanleyjones: I felt like we needed to talk about why you'd add worker nodes somewhere

engine/swarm/admin_guide.md
```diff
@@ -251,8 +237,51 @@ Before you forcefully remove a manager node, you must first demote it to the
...
+To do a _cold backup_, completely shut down the swarm before backing it up.
+Generally, _hot backups_ are sufficient. However, if any files are being changed
+while the copy is occurring, you may lose Some data. Be sure to test backup
+and recovery  thoroughly before settling on a backup strategy.
+
+#### Which manager to back up from
+
+In Raft, the leader can change at any time. In theory the followers may fall
+behind the leader, so a backup from a follower may miss some of the very latest
+data. However, this should not be an issue. Depending on your data integrity
```
mstanleyjones: @aaronlehmann it feels weaselly to say "this should not be an issue" but it was direct quote from you. Can you elaborate?

engine/swarm/admin_guide.md
```diff
@@ -251,8 +237,51 @@ Before you forcefully remove a manager node, you must first demote it to the
...
+[Restore from a backup](#restore-from-a-backup).
+
+### Things to know before backing up
+
+#### Hot or cold backups
+
+To do a _cold backup_, completely shut down the swarm before backing it up.
+Generally, _hot backups_ are sufficient. However, if any files are being changed
+while the copy is occurring, you may lose Some data. Be sure to test backup
+and recovery  thoroughly before settling on a backup strategy.
```
aaronlehmann: I think I've changed my mind about this. Let's recommend that people do cold backups. It's not very reassuring to say that hot backups will work, except when they don't work.

engine/swarm/admin_guide.md
aaronlehmann: If we recommend cold backups, then this is moot, because a manager that has been temporarily shut down can't be the leader.

engine/swarm/admin_guide.md
```diff
@@ -251,8 +237,51 @@ Before you forcefully remove a manager node, you must first demote it to the
...
+### Restore from a backup
+
+If you have backed up the Raft directory as described in
+[Backing up the swarm](#backing-up-the-swarm), you can restore the data to a new
+swarm.
+
+1.  Initialize a new single-node swarm.
+2.  Shut down Docker on that swarm.
+3.  Restore the `/var/lib/docker/swarm/raft` directory with the contents of the
+    backup.
```
aaronlehmann: This has become a little more involved with Docker 1.13 because the raft files are encrypted now. The operator will also need the encryption keys for this to work. @cyli what do you think is the best way to restore from a backup? Should the admin back up and restore the `certificates` directory as well?

engine/swarm/admin_guide.md
```diff
@@ -251,8 +237,51 @@ Before you forcefully remove a manager node, you must first demote it to the
...
+4.  Verify the contents of the backup.
```
aaronlehmann: Not sure I understand this step

engine/swarm/admin_guide.md
cyli: @aaronlehmann Yes they'd have to, because otherwise it can't be decrypted.  This means that it'd be the same CA and cert+key, though.   Not sure if the expectations are that the old swarm and the new swarm are both still running (I'd assume not, but never know), and if so it'd not be great to have both using the same certs, although this would be fixed when we had root ca rotation.

cyli: Clarification on the comment above:  I think we should still recommend that they back up and restore the certificates directory, because otherwise they won't be able to restore but make sure to say that the new swarm cluster will use the same CA as the old one, and the new node will have the same TLS key and cert as the backed-up node.  I don't think this should be a problem, unless someone is trying to run both the old and new swarms side by side, which is not really the use case.

engine/swarm/admin_guide.md
mstanleyjones: What I had in mind is that they should figure out their own strategy to verify the contents of the backup, probably based on the application they are running (like an integration test of sorts). In ops land, verifying that restores work is an important step.

mstanleyjones:
Addressed feedback so far. @cyli @aaronlehmann PTAL and let me know if more changes are needed, or if we need to go over anything f2f.

engine/swarm/admin_guide.md
```diff
@@ -251,8 +237,68 @@ Before you forcefully remove a manager node, you must first demote it to the
 worker role. Make sure that you always have an odd number of manager nodes if
 you demote or remove a manager
 
+## Back up the swarm
+
+Docker manager nodes store the swarm state and manager logs in the
+`/var/lib/docker/swarm/raft` directory. In 1.13 and higher, the Raft data is
+encrypted, and the encryption keys are stored in
+`/var/lib/docker/swarm/certificates`. Without these keys, you will not be able
+to restore the swarm.
```
aaronlehmann: We should instead ask users to back up and restore the entirety of `/var/lib/docker/swarm`. I realized there are a few other files here which are significant.

engine/swarm/admin_guide.md
```diff
@@ -251,8 +237,68 @@ Before you forcefully remove a manager node, you must first demote it to the
...
+3.  Remove the contents of `/var/lib/docker/swarm/raft` directory on the new
+    swarm, and on 1.13 and higher, the `/var/lib/docker/swarm/certificates`
+    directory.
+
+4.  Restore the `/var/lib/docker/swarm/raft` directory with the contents of the
+    backup. In Engine 1.13 and higher, also restore the
+    `/var/lib/docker/swarm/certificates` directory.
+
+    > **Note**: The new swarm will use the same encryption keys as the old one.
+    > It is not possible to change the encryption keys on a swarm at this time.
```
aaronlehmann: "The new node will use the same encryption key for on-disk storage as the old one. It is not possible to change the on-disk storage encryption keys at this time."

engine/swarm/admin_guide.md
cyli: @aaronlehmann Would that also be the case in 1.12?  (not sure if the `state.json` was in 1.12)

engine/swarm/admin_guide.md
```diff
@@ -251,8 +237,68 @@ Before you forcefully remove a manager node, you must first demote it to the
...
+1.  Stop Docker on the manager before backing up the data, so that no data is
+    being changed during the backup. It is possible to take a backup while the
+    manager is running (a "hot" backup), but this is not recommended and your
+    results will be less predictable when restoring. While the manager is down,
+    other nodes will continue generating swarm data that will not be part of
+    this backup.
+
+    > **Note**: Be sure to maintain the quorum of swarm managers. For instance,
+    > if you normally run with three managers, you need to add a manager before
+    > you can stop one, because two managers are not sufficient to maintain the
```
cyli: Two managers are a quorum given a cluster of 3 managers, so you shouldn't need to add another before bringing this one down.  Maybe something like "...if you normally run with 3 managers, and one of them is unreachable, stopping one of the remaining ones will break quorum and disable the cluster until either the unreachable one or the stopped one comes back up"?

engine/swarm/admin_guide.md
```diff
@@ -251,8 +237,68 @@ Before you forcefully remove a manager node, you must first demote it to the
...
+    directory.
+
+4.  Restore the `/var/lib/docker/swarm/raft` directory with the contents of the
+    backup. In Engine 1.13 and higher, also restore the
+    `/var/lib/docker/swarm/certificates` directory.
+
+    > **Note**: The new swarm will use the same encryption keys as the old one.
+    > It is not possible to change the encryption keys on a swarm at this time.
+
+5.  Restart the new swarm. If applicable, verify that the state of the swarm is
```
cyli: Should we mention that the unlock key is also the same as the old swarm, and that the unlock key will be needed to restore if the previous cluster is locked?

engine/swarm/admin_guide.md
```diff
@@ -251,8 +237,68 @@ Before you forcefully remove a manager node, you must first demote it to the
...
+
 ## Recover from disaster
 
+### Restore from a backup
+
+After backing up the swarm as described in
+[Backing up the swarm](#backing-up-the-swarm), use the following procedure to
+restore the data to a new swarm.
+
+1.  Initialize a new single-node swarm.
```
cyli: If we're restoring the whole swarm directory, I think you may not need to initialize a single-node swarm.  Just stop docker, restore the swarm directory, and start again.  At least, if I back up the `/var/lib/docker/swarm` dir, then do `docker swarm leave --force`, I just have to copy the swarm directory back and restart docker in order for that node to think it's in the swarm again.

engine/swarm/admin_guide.md
cyli: Also, are we specifically trying to create a new swarm?  If so, should we recommend `docker swarm init --force-new-cluster` so this node doesn't try to reach out to connect to old peers, etc.?

mstanleyjones:
Addressed the new (awesome) feedback. I think this is looking good. I talked about the need to rotate the unlock key after the restoration too.

engine/swarm/admin_guide.md
```diff
@@ -251,8 +237,85 @@ Before you forcefully remove a manager node, you must first demote it to the
...
+    > storage as the old one. It is not possible to change the on-disk storage
+    > encryption keys at this time.
+    >
+    > In the case of a swarm with auto-lock enabled, the unlock key is also the
+    > same as on the the old swarm, and the unlock key will be needed to
+    > restore.
+
+5.  Start Docker on the new node. Unlock the swarm if necessary. Re-initialize
+    the swarm using the following command, so that old nodes do not attempt to
+    reconnect to the swarm.
```
aaronlehmann: It's more so that this node doesn't attempt to connect to old nodes. If you had other nodes still running, presumably you wouldn't need to use the backup.

pascalandy:
Note to myself. Read this when merged.

engine/swarm/admin_guide.md
mstanleyjones: OK, I think I explained this more clearly.

engine/swarm/admin_guide.md
```diff
@@ -251,8 +237,86 @@ Before you forcefully remove a manager node, you must first demote it to the
...
+
+2.  Stop Docker on the manager before backing up the data, so that no data is
+    being changed during the backup. It is possible to take a backup while the
+    manager is running (a "hot" backup), but this is not recommended and your
+    results will be less predictable when restoring. While the manager is down,
+    other nodes will continue generating swarm data that will not be part of
+    this backup.
+
+    > **Note**: Be sure to maintain the quorum of swarm managers. For instance,
+    > if you normally run with three managers, you may need to add a manager
```
cyli: I think you might have to add 2 - if you start with 3, and add a node to become a 4-node cluster, you can still only lose 1 in order to maintain quorum.  Maybe we should just mention that your cluster is vulnerable to another node loss at this point?  Alternately, maybe we should recommend that if folks want to regularly take down nodes for backups, it's safer if they run with a 5-node cluster?

mstanleyjones: I reworded the note. PTAL.

engine/swarm/admin_guide.md
```diff
@@ -251,8 +237,87 @@ Before you forcefully remove a manager node, you must first demote it to the
...
+    manager is running (a "hot" backup), but this is not recommended and your
+    results will be less predictable when restoring. While the manager is down,
+    other nodes will continue generating swarm data that will not be part of
+    this backup.
+
+    > **Note**: Be sure to maintain the quorum of swarm managers. During the
+    > time that a manager is shut down, your swarm is more vulnerable to
+    > losing the quorum if further nodes are lost.The number of managers you
+    > run is a trade-off. If you regularly take down managers to do backups,
+    > consider running a 5-manager swarm, so that you can lose two managers
```
cyli: nitpick about wording:  I'm not sure whether this is counting the downed backing-up manager when it says that "you can lose two managers while the backup is running".    It implies to me that in addition to the manager that is down (and being backed up), you can lose an additional two.  

But you only lose one more manager while the backup is running, because if your backup is running, you're down to 4/5, and then you can only use 1 more and still maintain quorum.

(also in L263, should there be a space after the period?  `lost.The number`)

aaronlehmann:
LGTM

cyli:
LGTM!  Thank you for adding this and for the super quick response times!

PR #1567, opened 2017-02-09
Updated Oracle Linux instructions

Djelibeybi:
* Change the note about the UEK4, because it's not only required for `btrfs` support
* Added a note about the fact that Oracle also ships a supported package with the same name as upstream
* Added a note about the fact that we've already deprecated docker-engine-selinux
* Added a link to the Oracle Linux Docker User's Guide
* Fixed the GPG key section as it will not be shown during the `makecache fast` run but rather will automatically be accepted when the yum install command is run. 
* Added a note about disabling the `-y` parameter to prevent the automatic acceptance of the upstream GPG key
* Provided the specific GPG key output for better validation
* Provide context about the fact that `docker-engine` exists in the Oracle Linux addons repo as well and may appear in the results of the `yum ilst` command.
* Provide more context for the `btrfs` subvolume removal

### DIFF
--- a/engine/installation/linux/oracle.md
+++ b/engine/installation/linux/oracle.md
@@ -21,6 +21,9 @@ version 4.1.12 or higher. For Oracle Linux 6, you need to enable extra repositor
 to install UEK4. See
 [Obtaining and installing the UEK packages](https://docs.oracle.com/cd/E37670_01/E37355/html/ol_obtain_uek.html){: target="_blank" class="_" }.
 
+The [OverlayFS2 storage driver](https://docs.docker.com/engine/userguide/storagedriver/overlayfs-driver/) is only supported
+when running the UEK4.
+
 ### Remove unofficial Docker packages
 
 Oracle's repositories may contain an older version of Docker, with the package name
--- a/engine/installation/linux/oracle.md
+++ b/engine/installation/linux/oracle.md
@@ -16,14 +16,14 @@ To get started with Docker on Oracle Linux, make sure you
 
 To install Docker, you need the 64-bit version of Oracle Linux 6 or 7.
 
-To use `btrfs`, you need to install the Unbreakable Enterprise Kernel (UEK)
+Oracle recommends using the Unbreakable Enterprise Kernel (UEK)
 version 4.1.12 or higher. For Oracle Linux 6, you need to enable extra repositories
 to install UEK4. See
 [Obtaining and installing the UEK packages](https://docs.oracle.com/cd/E37670_01/E37355/html/ol_obtain_uek.html){: target="_blank" class="_" }.
 
 ### Remove unofficial Docker packages
 
-Oracle's repositories contain an older version of Docker, with the package name
+Oracle's repositories may contain an older version of Docker, with the package name
 `docker` instead of `docker-engine`. If you installed this version of Docker,
 remove it using the following command:
 
@@ -31,12 +31,18 @@ remove it using the following command:
 $ sudo yum -y remove docker
 ```
 
+Oracle's repositories also contain the Oracle-supported version of Docker which uses
+the same `docker-engine` name as the official Docker package. If you have installed this
+version of Docker, it will automatically be upgraded by the official version.
+
 You may also have to remove the package `docker-engine-selinux` which conflicts with
 the official `docker-engine` package.  Remove it with the following command:
 
 ```bash
 $ sudo yum -y remove docker-engine-selinux
 ```
+This package has already been deprecated by Oracle so will only exist on systems that have
+an older version of the Oracle-supported package installed.
 
 The contents of `/var/lib/docker` are not removed, so any images, containers,
 or volumes you created using the older version of Docker are preserved.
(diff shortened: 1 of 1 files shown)


Djelibeybi:
Is this PR going to be reviewed or merged?

engine/installation/linux/oracle.md
```diff
@@ -16,27 +16,36 @@ To get started with Docker on Oracle Linux, make sure you
 
 To install Docker, you need the 64-bit version of Oracle Linux 6 or 7.
 
-To use `btrfs`, you need to install the Unbreakable Enterprise Kernel (UEK)
+Oracle recommends using the Unbreakable Enterprise Kernel (UEK)
```
mstanleyjones: Please remove "Oracle recommends". Maybe you can quantify why the UEK kernel is recommended instead.

engine/installation/linux/oracle.md
```diff
@@ -16,27 +16,36 @@ To get started with Docker on Oracle Linux, make sure you
...
 to install UEK4. See
 [Obtaining and installing the UEK packages](https://docs.oracle.com/cd/E37670_01/E37355/html/ol_obtain_uek.html){: target="_blank" class="_" }.
 
+The [OverlayFS2 storage driver](https://docs.docker.com/engine/userguide/storagedriver/overlayfs-driver/) is only supported
+when running the UEK4.
+
 ### Remove unofficial Docker packages
 
-Oracle's repositories contain an older version of Docker, with the package name
+Oracle's repositories may contain an older version of Docker, with the package name
```
mstanleyjones: I would say if the package name is called `docker`, it is definitely older, right?

engine/installation/linux/oracle.md
```diff
@@ -16,27 +16,36 @@ To get started with Docker on Oracle Linux, make sure you
...
 `docker` instead of `docker-engine`. If you installed this version of Docker,
 remove it using the following command:
 
 ```bash
 $ sudo yum -y remove docker
 ```
 
+Oracle's repositories also contain the Oracle-supported version of Docker which uses
```
mstanleyjones: This seems superfluous, since the right thing will happen anyway: The latest version will always be chosen.

engine/installation/linux/oracle.md
```diff
@@ -55,7 +64,7 @@ You can install Docker in different ways, depending on your needs:
 
 - Some users cannot use third-party repositories, and must rely on
   the version of Docker in the Oracle repositories. This version of Docker may
-  be out of date. Those users should consult the Oracle documentation and not
+  be out of date. Those users should consult the [Oracle Linux Docker User's Guide](http://docs.oracle.com/cd/E52668_01/E75728/html/index.html) and not
```
mstanleyjones: Please remove this link. It will confuse users of Docker's packages since the Oracle documentation does not mention the existence of Docker's packages.

engine/installation/linux/oracle.md
```diff
@@ -122,18 +131,30 @@ Docker from the repository.
...
+    If this is the first time you are installing a package from
+    the Docker repositories, the GPG key and fingerprint will be shown and
+    automatically accepted.
+    
+    If you do not want to automatically accept the GPG key, remove the `-y` 
+    parameter so that yum prompts you to accept the GPG manually.
+    
+    Ensure the GPG details match the following:
+    
+    ```
```
mstanleyjones: Please add a `none` hint here.  Otherwise, this could be interpreted as something off the wall, like YAML.

engine/installation/linux/oracle.md
```diff
@@ -203,9 +229,10 @@ users to run Docker commands and for other optional configuration steps.
 
 #### Upgrade Docker
 
-To upgrade Docker, first run `sudo yum makecache fast`, then follow the
-[installation instructions](#install-docker), choosing the new version you want
-to install.
+To upgrade Docker, run `sudo yum update docker-engine` to upgrade to the latest available
+version. If you would prefer to upgrade to a specific version, first run
```
mstanleyjones: I think this can be shortened. Consider `s/If you would prefer....first run/To upgrade to a specific version, first run`

engine/installation/linux/oracle.md
Djelibeybi: Yes, but the package doesn't exist any more in our repositories, so the "may" also indicates that it may not exist at all.

engine/installation/linux/oracle.md
Djelibeybi: This can't be shortened. The two sentences are unrelated. You either upgrade to the latest version via `sudo yum update docker-engine` or you install a specific version.

engine/installation/linux/oracle.md
Djelibeybi: But users need to be warned that the Oracle supported version will automatically be replaced by a non-supported version.

Djelibeybi:
Closing this PR because the source branch was lost. I will open a new PR which includes updates based on these comments.

engine/installation/linux/oracle.md
mstanleyjones: Then why not "previously contained"?

engine/installation/linux/oracle.md
mstanleyjones: Yes, I was imagining a construction like:

To upgrade Docker to the latest version, run `sudo yum update docker-engine`. To upgrade to a specific version, first run....

PR #1572, opened 2017-02-09
Tweak 1.13-cs engine install/release notes for cs1

cpuguy83:


cs-engine/1.13/release-notes.md
```diff
@@ -16,7 +16,7 @@ cannot be adopted as quickly for consistency and compatibility reasons.
 
 ## CS Engine 1.13.0-cs1
```
mavenugo: This should be 1.13.1-cs1 ?

PR #1579, opened 2017-02-09
Update contact method for getting offline DB file

sarahpark:
Temporary fix until we get to docker/dhe-deploy#4075

### Proposed changes

We need a temporary way to get DB .tar files to users (who have offline mode for their DB updates). B/c Store team won't get to the ideal solution until after their 17 Feb sprint starts, we'll manually email users links. Support Team shouldn't handle this (since they'll be the middle man, haven't been onboarded on this situation yet, and will add more logistic chaos).

(I'm aware the email is "nautilus" which is an internal name - it's a google group we already have, so easy to use this one for now)

### Related issues (optional)
relates to docker/dhe-deploy#4075, https://docker.atlassian.net/browse/MER-1444


johndmulhausen:
Thanks, @sarahpark :)

PR #1629, opened 2017-02-13
Update upgrade.md to include socket

bgrissin:


bgrissin:
upgrade does not work on initial node without specifying the socket

PR #1633, opened 2017-02-13
Add links to video overviews

sarahpark:
### Proposed changes

Docker Inc. released (awesome) videos for the 1.13 / DDC release last week (see latest videos): https://www.youtube.com/user/dockerrun

Since "content is king", these videos seem like great ways to support the current content in docs (for users who prefer to watch videos first as an overview, and then read the rest of the docs for details).

---

B/c you can't embed Youtube videos in markdown, I have thumbnails of the videos that will open a new tab on click.

I have just 2 videos as a test :) If we find* they're generating more views to the videos + helping users through the docs, then will submit a PR to add the other videos (or I can just add them all anyways - up to you, Docs team).

*Not sure how to track this, so I can just manually revisit this and the docs' analytics (cross referenced with the videos' views) in a few weeks.


joaofnfernandes:
Thanks for putting in the effort for doing this 🚀

PR #1636, opened 2017-02-13
Remove unsupported IPAM config entries in v3

shin-:
Signed-off-by: Joffrey F <joffrey@docker.com>

### Proposed changes

Remove unsupported IPAM config entries in v3 because they are unsupported and will cause Compose to raise an error when used.

### Related issues (optional)

docker/compose#4474 , docker/compose#4366

### DIFF
--- a/compose/compose-file/index.md
+++ b/compose/compose-file/index.md
@@ -803,7 +803,7 @@ In the example below, three services are provided (`web`, `worker`, and `db`), a
 
 Specify a static IP address for containers for this service when joining the network.
 
-The corresponding network configuration in the [top-level networks section](compose-file.md#network-configuration-reference) must have an `ipam` block with subnet and gateway configurations covering each static address. If IPv6 addressing is desired, the [`enable_ipv6`](compose-file.md#enableipv6) option must be set.
+The corresponding network configuration in the [top-level networks section](compose-file.md#network-configuration-reference) must have an `ipam` block with subnet configurations covering each static address. If IPv6 addressing is desired, the [`enable_ipv6`](compose-file.md#enableipv6) option must be set.
 
 An example:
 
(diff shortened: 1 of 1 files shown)


blop:
So how are we supposed to define the gateway and ip_range using the macvlan driver for example ?
It was working just fine in v2. Is macvlan not supported anymore ?

mstanleyjones:
cc/ @shin-

shin-:
At the moment, those options are not available for Swarm services, hence their absence from the v3 format. Until then, if you depend on those options, please continue to use v2.

kelmansouri:
not sure if I quite understood here.

do you mean that it is not possible to specify a custom subnet in swarm using docker-compose v3?

intech:
Why this solution was blocked by swarm. Setting up ip_range is required without swarm.
This change has limited the ability to automatically configure routing for multi-host networks where swarm is not used. :angry: 
So far, I had to use version 2. I believe that you should reconsider this question regarding the ip_range parameter. :thinking: 
I will give you an example of network configuration:
```yml
# datacenter1 / server1
subnet: 10.1.1.0/8
ip_range: 10.1.1.1/24 # allow for container
```
```yml
# datacenter2 / server123
subnet: 10.2.123.0/8
ip_range: 10.2.123.1/24 # allow for container
```
and I cannot specify a subnet as 10.1.1.1/24 otherwise the containers on the multi-hosts will not see each other without additional settings. :exclamation:

coLATin:
"gateway" and "ip_range" shoule be useful lots of users

cbittencourt:
The removal of this feature in v3 is very unfortunate. Are there any workarounds for this when also depending on v3.x for other features such as https://docs.docker.com/compose/compose-file/#target?

haneef95:
Nice one @shin- 

Thanks, that worked!

intech:
Awesome SLA 4 years :tada: :smile:

PR #1646, opened 2017-02-14
Update aws.md

nicolaka:
updated example to reflect new HRM syntax

### DIFF
--- a/datacenter/install/aws.md
+++ b/datacenter/install/aws.md
@@ -342,7 +342,7 @@ provides multiple advantages to easily deploy and access your application.
     docker service create -p 8080 \
       --network ucp-hrm \
       --name demo-hrm-app \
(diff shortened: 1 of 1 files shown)


datacenter/install/aws.md
```diff
@@ -342,7 +342,7 @@ provides multiple advantages to easily deploy and access your application.
     docker service create -p 8080 \
       --network ucp-hrm \
       --name demo-hrm-app \
-      --label com.docker.ucp.mesh.http=8080=http://foo.example.com \
+      --label com.docker.ucp.mesh.http.8080=external_route=http://foo.example.com,internal_port=8080 \
```
joaofnfernandes: Based on https://docs.docker.com/datacenter/ucp/2.1/guides/admin/upgrade/incompatibilities-and-breaking-changes/

I think you can make `...http.8080=external_route` be just `...http=external_route`

PR #1662, opened 2017-02-14
Document secrets in compose file reference

mstanleyjones:
### Proposed changes

Documented secret info in compose file reference

Included info about currently-unresolved issue https://github.com/docker/docker/issues/30991 which makes short syntax unusable.

Engine 1.13.2

### DIFF
--- a/compose/compose-file/index.md
+++ b/compose/compose-file/index.md
@@ -850,6 +850,97 @@ port (a random host port will be chosen).
      - "127.0.0.1:5000-5010:5000-5010"
      - "6060:6060/udp"
 
+### secrets
+
+Grant access to secrets on a per-service basis using the per-service `secrets`
+configuration. Two different syntax variants are supported.
+
+> **Note**: The secret must already exist or be
+> [defined in the top-level `secrets` configuration](#secrets-configuration-reference)
+> of this stack file, or stack deployment will fail.
+
+#### Short syntax
+
+The short syntax variant only specifies the secret name. This grants the
+container access to the secret and mounts it at `/run/secrets/<secret_name>`
+within the container. The source name and destination mountpoint are both set
+to the secret name.
+
+> **Warning**: Due to a bug in Docker 1.13.1, using the short syntax currently
+> mounts the secret with permissions `000`, which means secrets defined using
+> the short syntax are unreadable within the container if the command does not
+> run as the `root` user. The workaround is to use the long syntax instead if
+> you use Docker 1.13.1 and the secret must be read by a non-`root` user.
+
+The following example uses the short syntax to grant the `redis` service
+access to the `my_secret` and `my_other_secret` secrets. The value of
+`my_secret` is set to the contents of the file `./my_secret.txt`, and
+`my_other_secret` is defined as an external resource, which means that it has
+already been defined in Docker, either by running the `docker secret create`
+command or by another stack deployment. If the external secret does not exist,
+the stack deployment fails with a `secret not found` error.
+
+```none
+version: "3.1"
+services:
+  redis:
+    image: redis:latest
+    deploy:
+      replicas: 1
+    secrets:
+      - my_secret
+      - my_other_secret
+secrets:
+  my_secret:
+    file: ./my_secret.txt
+  my_other_secret:
+    external: true
+```
+
+#### Long syntax
+
+The long syntax provides more granularity in how the secret is created within
+the service's task containers.
+
+- `source`: The name of the secret as it exists in Docker.
+- `target`: The name of the file that will be mounted in `/run/secrets/` in the
+  service's task containers. Defaults to `source` if not specified.
+- `uid` and `gid`: The numeric UID or GID which will own the file within
+  `/run/secrets/` in the service's task containers. Both default to `0` if not
+  specified.
+- `mode`: The permissions for the file that will be mounted in `/run/secrets/`
+  in the service's task containers, in octal notation. For instance, `0777`
+  represents world-readable. The default in Docker 1.13.1 is `0000`, but will
+  be `0444` in the future.
+
+The following example sets name of the `my_secret` to `redis_secret` within the
+container, sets the mode to `0440` (group-readable) and sets the user and group
+to `103`. The `redis` service does not have access to the `my_other_secret`
+secret.
+
+```none
+version: "3.1"
+services:
+  redis:
+    image: redis:latest
+    deploy:
+      replicas: 1
+    secrets:
+      - source: my_secret
+        target: redis_secret
+        uid: '103'
+        gid: '103'
+        mode: 0440
+secrets:
+  my_secret:
+    file: ./my_secret.txt
+  my_other_secret:
+    external: true
+```
+
+You can grant a service access to multiple secrets and you can mix long and
+short syntax. Defining a secret does not imply granting a service access to it.
+
 ### security_opt
 
 Override the default labeling scheme for each container.
@@ -1212,6 +1303,33 @@ refer to it within the Compose file:
       outside:
         external:
           name: actual-name-of-network
+
+## secrets configuration reference
+
+The top-level `secrets` declaration defines or references
+[secrets](/engine/swarm/secrets.md) which can be granted to the services in this
+stack. The source of the secret is either `file` or `external`.
+
+- `file`: The secret is created with the contents of the file at the specified
+  path.
+- `external`: If set to true, specifies that this secret has already been
+  created. Docker will not attempt to create it, and if it does not exist, a
+  `secret not found` error occurs.
+
+In this example, `my_first_secret` will be created (as
+`<stack_name>_my_first_secret)`when the stack is deployed,
+and `my_second_secret` already exists in Docker.
+
+```none
+secrets:
+  my_first_secret:
+    file: ./secret_data
+  my_second_secret
+    external: true
+```
+
+You still need to [grant access to the secrets](#secrets) to each service in the
+stack.
 
 ## Variable substitution
 


compose/compose-file/index.md
```diff
@@ -882,6 +882,73 @@ port (a random host port will be chosen).
...
+within the container. The source name and destination mountpoint are both set
+to the secret name.
+
+> **Warning**: Due to a bug, using the short syntax currently mounts the secret
+> with permissions `000`, which means the secret is currently unreadable if you
+> define it using the short syntax. The workaround is to use the long syntax
+> instead.
+
+The following example uses the short syntax to grant the `redis` service
+access to the `my_secrets` secret, which has been defined as an external
```
shin-: Nitpick: For consistency with `volumes` and `networks`, the examples should probably use local secrets over external ones (which require additional knowledge of the Docker Engine/CLI to use)

compose/compose-file/index.md
```diff
@@ -882,6 +882,73 @@ port (a random host port will be chosen).
...
+configuration. Two different syntax variants are supported.
+
+#### Short syntax
+
+The short syntax variant only specifies the secret name. This grants the
+container access to the secret and mounts it at `/run/secrets/<secret_name>`
+within the container. The source name and destination mountpoint are both set
+to the secret name.
+
+> **Warning**: Due to a bug, using the short syntax currently mounts the secret
```
dnephin: I don't really like the idea of calling this out in the reference. It should be soon. Are we going to remember to come back here and remove this warning?

I guess including a version number would be an improvement.

mstanleyjones: I think we have to include it, otherwise people on 1.13.1 will have a hell of a hard time figuring out what is going on (like I did, and @nathanleclaire did). I'm OK with putting the version on it. We can't assume everyone will instantly update to 1.13.2, or that people who aren't on 1.13 yet will go straight to 1.13.2. In fact, I don't think we **can** remove the note as long as 1.13.1 is a supported configuration. @thaJeztah ?

compose/compose-file/index.md
mstanleyjones: Maybe I should use one of each?

mstanleyjones:
Addressed feedback.

compose/compose-file/index.md
```diff
@@ -882,6 +882,81 @@ port (a random host port will be chosen).
      - "127.0.0.1:5000-5010:5000-5010"
      - "6060:6060/udp"
 
+### secrets
+
+After you
+[define or reference secrets in the top-level `secrets` configuration](#secrets-configuration-reference),
+grant access to secrets on a per-service basis using the per-service `secrets`
+configuration. Two different syntax variants are supported.
```
dnephin: Maybe this is a minor nit pick, but it kind of bothers me that this reads like narrative documentation, when it's supposed to be reference docs.

So it might say something like:

"Secrets allows you to define or reference..."

compose/compose-file/index.md
```diff
@@ -882,6 +882,81 @@ port (a random host port will be chosen).
...
+#### Short syntax
+
+The short syntax variant only specifies the secret name. This grants the
+container access to the secret and mounts it at `/run/secrets/<secret_name>`
+within the container. The source name and destination mountpoint are both set
+to the secret name.
+
+> **Warning**: Due to a bug in Docker 1.13.1, using the short syntax currently
+> mounts the secret with permissions `000`, which means the secret is unreadable
+> if you define it using the short syntax. The workaround is to use the long
```
dnephin: It's only uneradable if you use a non-root user. Since the root user is default, many people will never see this bug. Can we update to say "unreadable by non-root users"?

compose/compose-file/index.md
```diff
@@ -882,6 +882,81 @@ port (a random host port will be chosen).
...
+> syntax instead, if you use Docker 1.13.1.
+
+The following example uses the short syntax to grant the `redis` service
+access to the `my_secret` and `my_other_secret` secrets. The value of
+`my_secret` is set to the contents of the file `./my_secret.txt`, and
+`my_other_secret` is defined as an external resource, which means that it has
+already been defined in Docker, outside of this stack file.
```
dnephin: maybe "... is defined as an external resource, which means that it must be created with `docker secret create`" ?

Saying "has already been defined" isn't necessarily true. Maybe someone is using an external secret they haven't defined (which would give them an error).

compose/compose-file/index.md
```diff
@@ -882,6 +882,81 @@ port (a random host port will be chosen).
...
+  my_other_secret:
+    external: true
+```
+
+#### Long syntax
+
+When you use the long syntax to grant access to a secret, you can specify a
+different destination name for the secret within the container, different
+numeric user ID or group ID, or different permissions for the secret. The
+following example sets name of the `my_secret` to `redis_secret` within the
```
dnephin: Similar nit pick to above about sounding like narative docs.  

How about: "The long syntax grants access..."

Could we also remove "different" ? It's not really different, just non-default.

compose/compose-file/index.md
mstanleyjones: I'm not on-board with this distinction, because the compose file reference is literally the only source for a lot of this stuff. This and the generic Secrets topic are all that people get, and this is a different way of thinking about configuring secrets than doing it all by hand.

Also, "secrets" can't allow you to do anything. "secrets" are not an actor. You are the actor, using a tool. So I don't think we should artifically use passive voice here. Can you unpick why this bothers you?

compose/compose-file/index.md
mstanleyjones: I can, but it's only readable by root because root can read things that are set as un-readable. The secret's mode is set to 000, which is technically readable by nobody. The reason it works at all is kind of a fluke, and probably highlights some faulty assumptions in the testing of the feature. I think that by documenting that it still works for root users, that encourages people to run within containers as root, which I don't think is something we actively want to encourage.

compose/compose-file/index.md
mstanleyjones: It's not accurate to say that it must be created by `docker secret create` since another stack file could have created it, right? It's interesting to point out that you'll get an error if you declare an external secret that doesn't yet exist. In fact, what will it do? I'll test it to see.

compose/compose-file/index.md
mstanleyjones: The long syntax itself does not grant access. It's a way for you to grant access. As above, I'm curious what bothers you about using second person and active voice here. Presumably the person is reading the reference in order to get things done. This isn't academia.

compose/compose-file/index.md
dnephin: >  the compose file reference is literally the only source for a lot of this stuff.

That's not really an argument for confusing the type of documentation, it's just an argument for including some more docs about it somewhere else (assuming we need it, which I'm not sure we do).

>  "secrets" can't allow you to do anything. "secrets" are not an actor.

Ok, we can be more verbose. "The `secrets` field allows you..."

> Can you unpick why this bothers you?

I think it's because I prefer to read reference docs and it feels more comfortable to me to read things written in the passive voice.

Wouldn't it be good to remain consistent with the rest of the document?

mstanleyjones: Maybe for the very very short term. Technical writers tend to avoid the passive voice. Maybe @johndmulhausen can weigh in on that in general. To be honest, I did not read the entire document before contributing this PR, because this PR is very important and late, and this is not normally my content to maintain.

compose/compose-file/index.md
dnephin: > The reason it works at all is kind of a fluke

Not really, it's by design that root can read and edit these files.

By omitting this detail we mislead people into thinking the feature is entirely broken on 1.13.1, which it is not. It's only broken if you are a non-root user.

nathanleclaire: FYI: With `000` it is not readable by _anyone_, including `root`. That's the bug in 1.13.1, nobody can read the file.

But (unsolicited opinion) we should be encouraging and enabling folks to run containers as non-root wherever possible...

nathanleclaire:
👍

compose/compose-file/index.md
dnephin: It's not a major issue for me. I'm fine keeping it the way it is for now. It'll be a bit inconsistent with the tone of the rest of the doc, but I agree it's better we get the docs live.

compose/compose-file/index.md
dnephin: Being accurate about the scope of a bug is hardly encouraging anyone to run containers as root.

The bug is that non-root users can not read the file. Claiming otherwise is misleading and more likely to prevent people from adopting secrets. If they need them as a non-root user, we already have the workaround documented (they can set the mode with the long form), which is much easier than running as root.

compose/compose-file/index.md
dnephin: @mstanleyjones It will give you an error:

```
secret not found: external-secret-name
```

> If I understand the declarative model of services correctly, it shouldn't actually be an error to create a service which relies on a secret that doesn't yet exist

It is an error because the client needs to resolve that secret to a secret id. If the secret doesn't exist it can't resolve it, so it can't even create the service.

>  since another stack file could have created it, right? 

Technically yes. What's important is that the user understands they need to create the secret. I don't think the previous wording made that clear.

For external volumes (and networks) we say this:

> If set to `true`, specifies that this volume has been created outside of Compose. `docker-compose up` will not attempt to create it, and will raise an error if it doesn't exist.

Can we use that here as well?

compose/compose-file/index.md
dnephin: Ok, I still prefer the tone of the current document, but I'm fine with this inconsistency for now. It's not a major issue, so let's ignore it.

mstanleyjones:
@dnephin I think I made a good compromise. Also addressed your other feedback. PTAL. Sorry for the delay on this, we were hack-a-thoning Thursday and Friday.

mstanleyjones:
I'm going to merge for now. I think all the facts are accurate. We can fix things later to do with the consistency of the file.

dnephin:
Thanks! LGTM

PR #1666, opened 2017-02-15
clarified logging into a swarm manager to list nodes

londoncalling:
Added a reminder about how to log into a swarm node in a later step in this task.


londoncalling:
LGTM

PR #1668, opened 2017-02-15
Added a bit more detail to restart.md

erd0s:
This killed me for a while, I expected that since environment variables were set after the build process that they would be reloaded on restart.


johndmulhausen:
This is great, thanks @erd0s :)

PR #1672, opened 2017-02-15

scjane:

### DIFF
--- a/engine/userguide/eng-image/dockerfile_best-practices.md
+++ b/engine/userguide/eng-image/dockerfile_best-practices.md
@@ -275,10 +275,10 @@ used an older version, specifying the new one causes a cache bust of `apt-get
 update` and ensure the installation of the new version. Listing packages on
 each line can also prevent mistakes in package duplication.
 
-In addition, cleaning up the apt cache and removing `/var/lib/apt/lists`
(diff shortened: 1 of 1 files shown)


engine/userguide/eng-image/dockerfile_best-practices.md
```diff
@@ -275,8 +275,8 @@ used an older version, specifying the new one causes a cache bust of `apt-get
 update` and ensure the installation of the new version. Listing packages on
 each line can also prevent mistakes in package duplication.
 
-In addition, cleaning up the apt cache and removing `/var/lib/apt/lists` helps
-keep the image size down. Since the `RUN` statement starts with
+In addition, cleaning up the apt cache and removing `/var/lib/apt/lists`
+reduce the image size. Since the `RUN` statement starts with
```
johndmulhausen: s/reduce/reduces

mstanleyjones: Actually the previous wording was grammatically correct. Cleaning up and removing are presented as two separate steps, so both steps together need to be treated as plural. However, in reality, it's only one step (removing) which itself cleans. I cleaned up the wording to make it more clear. I'll squash my commit.

PR #1676, opened 2017-02-15
Update Docker for AWS release notes page

kencochrane:
### Proposed changes

- Updated Docker for AWS release notes page

Signed-off-by: Ken Cochrane <KenCochrane@gmail.com>

### DIFF
--- a/docker-for-aws/release-notes.md
+++ b/docker-for-aws/release-notes.md
@@ -6,29 +6,77 @@ title: Docker for AWS Release Notes
 
 {% include d4a_buttons.md %}
 
-## 1.13.0-1
-Release date: 1/18/2017
+## Stable Channel
+
+### 1.13.1-2 (stable)
+Release date: 02/08/2017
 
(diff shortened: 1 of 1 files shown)


docker-for-aws/release-notes.md
```diff
@@ -6,41 +6,89 @@ title: Docker for AWS Release Notes
...
-Release date: 1/18/2017
+## Stable Channel
+
+### 1.13.1-2 (stable)
+Release date: 02/08/2017
 
 {{aws_button_latest}}
 
-### New
+**New**
```
mstanleyjones: I don't mind changing these from H-level headings to just bold text, but it means that you can't permalink straight to the new features for each version if you wanted to. Up to you!

kencochrane: That is fine, we would want to link directly to a release version anyway. It also cleans up the side menu since we don't have New after every release version. Our release notes aren't that big, so it won't be that much of a hassle to our users.

PR #1705, opened 2017-02-15
remove duplicate array in redirect_from directive

parkr:
This PR fixes a build failure I saw locally with the latest version of github-pages.

```text
  Liquid Exception: undefined method `end_with?' for ["/cs-engine/release-notes/prior-release-notes/"]:Array in 404.md
  bundler: failed to load command: jekyll (/Users/parkr/.rbenv/versions/2.3.1/lib/ruby/gems/2.3.0/bin/jekyll)
  NoMethodError: undefined method `end_with?' for ["/cs-engine/release-notes/prior-release-notes/"]:Array
```

PR #1707, opened 2017-02-16
Gemfile: add note about upgrade process of github-pages gem

parkr:
The Gemfile cannot be upgraded to a newer version of the github-pages
gem without having a corresponding tag in the starefossen/github-pages
image. For future contributors, this is good to know so they don't
accidentally hold up a PR of theirs like I did in #1705. :wink:


johndmulhausen:
Thanks, @parkr !

PR #1711, opened 2017-02-16
Add info about fetching credentials from IAM role

mstanleyjones:
Fixes https://github.com/docker/distribution/issues/741

@stevvooe PTAL. I'm a little out of my depth. There was already advice about setting these to an empty string for one behavior, and now we have another behavior if they are completely omitted. Is that right?

### DIFF
--- a/registry/storage-drivers/s3.md
+++ b/registry/storage-drivers/s3.md
@@ -20,10 +20,10 @@ Amazon S3 or S3 compatible services for object storage.
       <code>accesskey</code>
     </td>
     <td>
-      yes
+      no
     </td>
     <td>
-      Your AWS Access Key.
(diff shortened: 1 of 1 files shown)


registry/storage-drivers/s3.md
```diff
@@ -31,7 +32,8 @@ Amazon S3 or S3 compatible services for object storage.
       <code>secretkey</code>
     </td>
     <td>
-      yes
+      yes, unless you use [IAM roles](http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/iam-roles-for-amazon-ec2.html).<br />
```
stevvooe: It might just be better to say no, then have this IAM bit in the description field.

mstanleyjones:
Addressed feedback. I'll merge this when Netlify finishes if there isn't any more feedback.

PR #1726, opened 2017-02-16
CC the assignee (since direct issue assignment doesn't work)

johndmulhausen:
When filing bugs, we can CC the owner of the page, in lieu of assigning the bug to them directly. That way there's at least some visibility on the author's radar. 

Also, updating the assignees to reflect the departure of sanscontext.

To test out the behavior, go to any page in the Netlify preview and click "Request docs changes" on the right or bottom of the page.

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -48,7 +48,7 @@ defaults:
   - scope:
       path: "docker-cloud"
     values:
-      assignee: "sanscontext"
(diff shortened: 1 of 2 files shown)


johndmulhausen:
Note: I'm merging this because we have a lot of feedback coming in on the docs today that I want to get routed properly; we can refine the approach later.

PR #1731, opened 2017-02-16
Update zfs-driver.md

scjane:


johndmulhausen:
Fair enough, thanks @scjane :)

PR #1763, opened 2017-02-16
fix(compose:extends): not supported from version 3

### LINKED ISSUE #1607: Feedback for: compose/compose-file/index.md
[compose/compose-file/index.md](https://docs.docker.com/compose/compose-file/)

We should remove `extends` from this list and call out the change on the [compose versioning](https://docs.docker.com/compose/compose-file/compose-versioning/) page.

See: docker/compose#4470, docker/compose#4315

Electron-libre:
docker compose : `extends` is not supported since v3.

This PR update the compose reference, migration instructions , and add note to the extends documentation.

Fixes #1607

### DIFF
--- a/compose/compose-file/compose-versioning.md
+++ b/compose/compose-file/compose-versioning.md
@@ -227,7 +227,7 @@ several options have been removed:
     `deploy`. Note that `deploy` configuration only takes effect when using
     `docker stack deploy`, and is ignored by `docker-compose`.
 
-- `extends`: This option has been removed. You wont be able to use `extends` along version "3"
+-   `extends`: This option has been removed for `version: "3"` Compose files.
 
 ### Version 1 to 2.x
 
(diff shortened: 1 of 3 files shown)


londoncalling:
@shin- @dnephin can you all verify this is accurate? (and maybe we can chat about I can verify next time myself, I scanned the Compose repo but couldn't figure it out immediately)

compose/compose-file/compose-versioning.md
```diff
@@ -227,6 +227,8 @@ several options have been removed:
     `deploy`. Note that `deploy` configuration only takes effect when using
     `docker stack deploy`, and is ignored by `docker-compose`.
 
+-   `extends`: This option has been removed for `version: "3"` Compose files.
```
shin-: ```s/`version: "3"`/version 3.x/```

londoncalling:
LGTM

compose/compose-file/compose-versioning.md
shin-: @londoncalling I think you missed this instance - afaik we don't use that notation anywhere in the docs, so it should be modified to version 3.x (especially because the comment applies to 3.1 as well)

londoncalling: I'll fix it via the `edit page` link on the published doc, thanks @shin-

PR #1772, opened 2017-02-16
Changed the label information

clemenko:
Updated the label information to include the objects that it can be applied to.

### DIFF
--- a/datacenter/ucp/2.0/guides/user-management/permission-levels.md
+++ b/datacenter/ucp/2.0/guides/user-management/permission-levels.md
@@ -43,7 +43,7 @@ users can see the containers they deploy in the cluster.
 
 ## Team permission levels
 
-Teams and Labels give the administrator fine-grained control over permissions. Each team can have multiple labels. Each label has a key of `com.docker.ucp.access.label`. The label is then applied to the containers, services, networks, secrets and volumes. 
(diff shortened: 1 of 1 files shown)


datacenter/ucp/2.0/guides/user-management/permission-levels.md
```diff
@@ -43,8 +43,7 @@ users can see the containers they deploy in the cluster.
 
 ## Team permission levels
 
-Teams allow you to define fine-grain permissions to services, containers, and
-networks that have the label `com.docker.ucp.access.label` applied to them.
+Teams and Labels give the administrator fine-grained control over permissions. Each team can have multiple labels. Each label has a key of `com.docker.ucp.access.label`. The label is then applied to the containers, services, networks, secrets and volumes.
```
joaofnfernandes: Can we also make it explicit what resources don't have RBAC? Is it just images and nodes @alexmavr?

joaofnfernandes: Make "Labels" lower case. Only product names should be capitalized :)

alexmavr: RBAC does not currently apply for images, nodes, plugins and ports

clemenko:
Updated branch.

Issue #1703, opened 2017-02-15
How to configure a specific load balancer for DTR


joaofnfernandes:
We have an [overview page ](https://docs.docker.com/datacenter/dtr/2.2/guides/admin/configure/use-a-load-balancer/), but we need docs with example configurations for specific load balancers.

- [ ] haproxy
- [ ] elb
- [ ] nginx

Before starting, create a new issue where you specify the load balancer config you're going to tackle, and reference this issue on the new one.

pdevine:
This should include how the /health check endpoint works.  We need to be explicit about how to get per replica health, as well as cluster health.

nicolaka:
dibs

joaofnfernandes:
@nicolaka if you're tackling just a single load balancer, create a separate issue to track that. But given your chops I bet you'll try to have config files for all three :D

nicolaka:
anyone has an idea why in haproxy https healthchecks are failing with the following config?

```
backend dtr_upstream_servers_443
        mode tcp
        option httpchk /health
        server node01:12391 172.31.39.165:12391 check check-ssl verify none weight 100
        server node02:12391 172.31.21.168:12391 check check-ssl verify none weight 100
        server node03:12391 172.31.14.53:12391 check check-ssl verify none weight 100
```

@pdevine @vikstrous ?

whoshuu:
By default `option httpchk` uses `OPTIONS` as the HTTP method. `OPTIONS` is not a valid method on that endpoint. Try this instead:

```
option httpchk get /health
```

nicolaka:
same result @whoshuu with :

```
backend dtr_upstream_servers_443
        mode tcp
        option httpchk get /health
        server node01 172.31.39.165:12391 weight 100 check
        server node02 172.31.21.168:12391 weight 100 check
        server node03 172.31.14.53:12391 weight 100 check
```

if i check directly with curl it also doesn't work unless I add `https://` to the URL.

vikstrous:
This is the template we use:

```
frontend localnodes
	bind *:%s
	mode tcp
	default_backend nodes
	timeout client 1h

backend nodes
	mode tcp
	balance roundrobin
	{{range .}}
	server clone{{.Index}} {{.IP}}:{{.Port}} check
	{{end}}
	timeout connect 20s
	timeout server 1h
```

nicolaka:
Still no luck here. I think this should be the right setting to USE https to do healthchecks but without verifying server cert.

```
[...]

PR #1774, opened 2017-02-16
Load balancer health check explanations

### LINKED ISSUE #1702: Explain how to use the /health endpoint for load-balancing DTR
Check if there's documentation that explains the health-checking endpoint of DTR.
- What does this endpoint really validates? Just the node, database, or the status of all internal components?

### LINKED ISSUE #1771: Explain how to use the /api/v0/meta/cluster_status endpoint for load-balancing DTR
Would be helpful for checking the health of the entire cluster.

whoshuu:
Fixes #1702
Fixes #1771 
References #1703

### DIFF
--- a/datacenter/dtr/2.2/guides/admin/configure/use-a-load-balancer.md
+++ b/datacenter/dtr/2.2/guides/admin/configure/use-a-load-balancer.md
@@ -14,14 +14,14 @@ This allows users to access DTR using a centralized domain name. If a replica
 goes down, the load balancer can detect that and stop forwarding requests to
 it, so that the failure goes unnoticed by users.
 
-## Load-balancing DTR
+## Load balancing DTR
 
 DTR does not provide a load balancing service. You can use an on-premises
 or cloud-based load balancer to balance requests across multiple DTR replicas.
 
 Make sure you configure your load balancer to:
 
-* Load-balance TCP traffic on ports 80 and 443
+* Load balance TCP traffic on ports 80 and 443
 * Not terminate HTTPS connections
 * Use the `/health` endpoint on each DTR replica, to check if
 the replica is healthy and if it should remain on the load balancing pool or
--- a/datacenter/dtr/2.2/guides/admin/configure/use-a-load-balancer.md
+++ b/datacenter/dtr/2.2/guides/admin/configure/use-a-load-balancer.md
@@ -24,18 +24,27 @@ Make sure you configure your load balancer to:
 
 * Load balance TCP traffic on ports 80 and 443
 * Not terminate HTTPS connections
-* Use the `/health` endpoint (note the lack of an `/api/v0/` in the path) on each
-DTR replica, to check if the replica is healthy and if it should remain on the
-load balancing pool or not
+* Use the unauthenticated `/health` endpoint (note the lack of an `/api/v0/` in
+the path) on each DTR replica, to check if the replica is healthy and if it
+should remain in the load balancing pool or not
 
 ## Health check endpoints
 
-The `/health` endpoint returns a JSON object for the replica being queried with
-`"Healthy"` as one of the keys. Any response other than a 200 HTTP status code
-and `"Healthy":true` means the replica is unsuitable for taking requests. If
-the API server is still up, the returned JSON object will have an `"Error"` key
-with more details. More specifically, these issues can be in any of these
-services:
+The `/health` endpoint returns a JSON object for the replica being queried of
+the form:
+
+```json
+{
+  "Error": "error message",
+  "Health": true
+}
+```
+
+A response of `"Healthy": true` means the replica is suitable for taking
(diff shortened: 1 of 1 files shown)


datacenter/dtr/2.2/guides/admin/configure/use-a-load-balancer.md
```diff
@@ -14,18 +15,45 @@ This allows users to access DTR using a centralized domain name. If a replica
...
-* Use the `/health` endpoint on each DTR replica, to check if
-the replica is healthy and if it should remain on the load balancing pool or
-not
+* Use the `/health` endpoint (note the lack of an `/api/v0/` in the path) on each
+DTR replica, to check if the replica is healthy and if it should remain on the
+load balancing pool or not
+
+## Health check endpoints
+
+The `/health` endpoint returns a JSON object for the replica being queried with
```
joaofnfernandes: This is really helpful, but show me instead of describing it :D
A sample of the json structure would be more helpful, and then you can use text to clarify some fields that might be misleading.

datacenter/dtr/2.2/guides/admin/configure/use-a-load-balancer.md
```diff
@@ -14,18 +15,45 @@ This allows users to access DTR using a centralized domain name. If a replica
...
+* Storage container (registry)
+* Authorization (garant)
+* Metadata persistence (rethinkdb)
+* Content trust (notary)
+
+Note that this endpoint is for checking the health of a *single* replica. To get
+the health of every replica in a cluster, querying each individual replica is
+the preferred way to do it in real time.
+
+The `/api/v0/meta/cluster_status` endpoint returns a JSON object for the entire
```
joaofnfernandes: I think you also need to be authenticated to be able to query this endpoint. If that's the case, can regular DTR users do it or does it need admin credentials?

datacenter/dtr/2.2/guides/admin/configure/use-a-load-balancer.md
```diff
@@ -14,18 +15,45 @@ This allows users to access DTR using a centralized domain name. If a replica
...
+cluster *as observed* by the replica being queried. Health status for the
+replicas is available in the `"replica_health"` key. These statuses are taken
+from a cache which is updated by each replica individually.
+
+In addition, this endpoint returns a dump of the rethink system tables
+which can be rather large (~45 KB) for a status endpoint.
```
joaofnfernandes: This is useful but has two assumptions:
* that users know DTR internals like rethink
* they know mathematical notation

Lets make this accessible to anyone, so maybe
"The response of this endpoint contains information about the internal DTR storage, making the response json be around 45 KB."

datacenter/dtr/2.2/guides/admin/configure/use-a-load-balancer.md
whoshuu: Is this something that would be more appropriate as a link to official swagger docs? (We actually don't have any for the `/health` endpoint. I don't see any where else in the docs that shows actual JSON responses.

joaofnfernandes: From what I can [see here](https://docs.docker.com/datacenter/dtr/2.2/reference/api/) the `/health` endpoint is not part of the swagger docs 😓
But for the `cluster_status` let's link, [here's the source](https://github.com/docker/docker.github.io/blob/master/datacenter/dtr/2.2/reference/api/index.html). I'm not sure we can do relative links in this case, since the swagger api is essentially an html app we just dump on that directory.

datacenter/dtr/2.2/guides/admin/configure/use-a-load-balancer.md
whoshuu: Admin only :octocat:

datacenter/dtr/2.2/guides/admin/configure/use-a-load-balancer.md
whoshuu: Yeah I've tried linking it, and it breaks down at the `index` part of the path, which is unfortunate. Maybe there's a way to get an already expanded version of the docs that's static html, that would make permalinking it much easier.

datacenter/dtr/2.2/guides/admin/configure/use-a-load-balancer.md
joaofnfernandes: Then let's mention that in the doc so that it is clear that `/health` is more appropriate for load balancers and generic things that don't need to care about getting authentication tokens.

datacenter/dtr/2.2/guides/admin/configure/use-a-load-balancer.md
joaofnfernandes: For the time being we can link to https://docs.docker.com/datacenter/dtr/2.2/reference/api/

PR #1779, opened 2017-02-16
fix link to Cloud docs on Manuals tab

londoncalling:
@pkennedyr @johndmulhausen 

Fixed link to Docker Cloud docs on Manuals tab

(we should also look at the UCP docs which are re-directing right now to a version specific URL)


londoncalling:
LGTM, merging

PR #1781, opened 2017-02-16
Remove Javascript for New Relic

mstanleyjones:
Nobody seems to be using whatever data is gathered by this behemoth. When unminified, it is over 200 lines of code, processed on every page load. 👎


jstepka:
Def not used by the data team. +1

johndmulhausen:
Thanks, @mstanleyjones :)

PR #1786, opened 2017-02-17
Update uninstall.md with details about the Swarm mode CA with UCP

### LINKED ISSUE #1653: swarm-mode wont have a valid CA after uninstalling UCP
@dhiltgen commented 

If the user is uninstalling UCP, we should fix-up the swarm configuration so that it is no longer using our external CA, which will no longer be present after we complete the uninstall.

We should be able to take the cluster root CA material and move it from the named volume back to the host in the location where swarm looks for it, then remove the remote CA configuration.

Hopefully the swarm mode manager will "do the right thing" at that point after we update the config removing the remote CA, but worst case scenario, we may have to instruct the user to restart their docker daemons to complete the uninstall.

---

@dhiltgen commented 

Latest word from the core team is this isn't really going to be possible in 1.12 GA cleanly.  I've filed https://github.com/docker/swarmkit/issues/1157 but they'd like to take the time to make sure this is done right, and we just don't have enough time left in 1.12 to do that.  I'm going to move this issue to a later seattle sprint so we can re-evaluate before release if there's anything we can do to mitigate, or perhaps the core team will have had more time to work on a fix for a patch release.

---

@dhiltgen commented 

Given that we can't actually fix this for seattle, I'm switching this to a doc issue.  In short, we'll have to doc a known issue that after uninstalling UCP, your swarm cluster will not be able to issue certificates so you will have to re-install UCP before your certs expire.

wsong:
### Proposed changes

Added a paragraph about uninstalling UCP and its effects on the Swarm mode CA.

### Related issues (optional)

Fixes #1653.

### DIFF
--- a/datacenter/ucp/2.1/guides/admin/install/uninstall.md
+++ b/datacenter/ucp/2.1/guides/admin/install/uninstall.md
@@ -33,6 +33,16 @@ for any necessary configuration values.
 [Check the reference documentation](../../../reference/cli/index.md) to learn the options
 available in the `uninstall-ucp` command.
 
+## Swarm mode CA
+
+After uninstalling UCP, the nodes in your cluster will still be in swarm mode, but you cannot
+join new nodes until you reinstall UCP, because swarm mode was relying on UCP to provide the
+CA certificates that allow nodes in the cluster to identify each other. Additionally, since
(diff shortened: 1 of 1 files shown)


joaofnfernandes:
Thanks for the contribution 🚀

PR #1787, opened 2017-02-17
document dtr backup/restore better

### LINKED ISSUE #1722: How to backup and restore DTR
DTR 2.2 has backup and restore [docs here](https://docs.docker.com/datacenter/dtr/2.2/guides/admin/backups-and-disaster-recovery/), but a bunch of users are having problems when following these instructions.

- [x] Explain when backup and restores should be used. Check #1720
- [x] Review that the current docs work on DTR 2.2
- [x] Make it clear what the DTR backup saves and what doesn't. Check #1704 
- [x] How do I test if a backup succeeded
- [ ] How do I backup/restore images when using an external storage
- [x] When restoring you must your the same docker/dtr tag that created the backup. Check #1685

vikstrous:
fix #1722

### DIFF
--- a/datacenter/dtr/2.2/guides/admin/backups-and-disaster-recovery.md
+++ b/datacenter/dtr/2.2/guides/admin/backups-and-disaster-recovery.md
@@ -4,13 +4,10 @@ keywords: docker, registry, high-availability, backup, recovery
 title: DTR backups and recovery
 ---
 
-DTR relies on consensus to operate and if it loses consensus permanently
-and the replica that failed can't be brought back to life, the only
-option is to restore from backups. This is why it's very important to perform
-periodic backups. Note that backups are tied to specific DTR versions and work
-best with those DTR versions. You can backup/restore across patch versions
-at your own risk, but not across minor versions as those require more complex
-migrations.
+DTR replicas rely on having a majority available at any given time for writes to
+succeed. Therefore if a majority of replicas are permanently lost, the only way
+to restore DTR to a working state is to recover from backups. This is why it's
+very important to perform periodic backups.
 
 ## DTR data persistence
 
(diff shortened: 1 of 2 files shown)


datacenter/dtr/2.2/guides/admin/backups-and-disaster-recovery.md
```diff
@@ -4,10 +4,13 @@ keywords: docker, registry, high-availability, backup, recovery
 title: DTR backups and recovery
 ---
 
-When you use Docker Trusted Registry on a production setting, you should first
-configure it for high availability.
-
-The next step is creating a backup policy and disaster recovery plan.
+DTR relies on consensus to operate and if it loses consensus permanently
```
joaofnfernandes: When you say "DTR relies on consensus to operate" it's very clear in your head what you mean, but people with less experience in distributed systems might not really understand what you mean. So you need to provide a bit more context. What about

The way multiple DTR instances replicate data is by using log consensus. When you data is changed in one of the instances, it its replicated into the other instances, but is only persisted if a majority or replicas acknowledges the change.
This means that if the majority of the replicas are unhealthy, DTR data cannot be replicated, and you'll have to restore the DTR data from an existing backup.

The rest of the sentence is not really important for the first part of the explanation, we can move it down below.

datacenter/dtr/2.2/guides/admin/backups-and-disaster-recovery.md
```diff
@@ -37,17 +43,30 @@ command creates a backup of DTR:
 
 These files are added to a tar archive, and the result is streamed to stdout.
 
+Things DTR's backup command doesn't back up:
```
joaofnfernandes: Write as if you were explaining this to me face to face. In this case I bet you'd say

DTR doesn't backup:

datacenter/dtr/2.2/guides/admin/backups-and-disaster-recovery.md
```diff
@@ -37,17 +43,30 @@ command creates a backup of DTR:
...
+
+* The image scanning database (if using image scanning)
+* Image contents
+* Users, orgs, teams
+
+There is no way to back up the image scanning database. You can re-download it
```
joaofnfernandes: Since you start the sentence with "there is no way you can backup", you're giving people a mini hearth attack.
Maybe

Once you restore DTR from a backup, you can download the latest vulnerability database.

datacenter/dtr/2.2/guides/admin/backups-and-disaster-recovery.md
```diff
@@ -71,34 +90,67 @@ Where:
...
+docker client command with the -e flag which in turn relays the password to the
+DTR bootstrapper.
+
+## Testing backups
+
 To validate that the backup was correctly performed, you can print the contents
 of the tar file created:
 
 ```none
 $ tar -tf /tmp/backup.tar
```
joaofnfernandes: What should I visually look for? I'll see a bunch of files but should I look for something specific?

datacenter/dtr/2.2/guides/admin/backups-and-disaster-recovery.md
```diff
@@ -71,34 +94,67 @@ Where:
...
+manage users, orgs or teams so if you try to
+restore DTR on a cluster other than the one it was backed up on, DTR
+repositories will be associated with users that don't exist and it will appear
+as if the restore operation didn't work.
 
-This command does not restore Docker images. You should implement a separate
-restore procedure for the Docker images stored in your registry, taking in
-consideration whether your DTR installation is configured to store images on
-the filesystem or using a cloud provider.
+Note that to restore DTR, you must first remove any left over containers from
```
joaofnfernandes: Can I use the `restore` or `destroy` commands to achieve this? If so guide me on how to do it, it or link to the uninstall page.

datacenter/dtr/2.2/guides/admin/backups-and-disaster-recovery.md
```diff
@@ -71,34 +94,67 @@ Where:
...
+the previous installation.
 
-You can check the
-[reference documentation](../../reference/cli/backup.md), for the
-backup command to learn about all the available flags.
+The restore command performs a fresh installation of DTR, and reconfigures it with
+the configuration created during a backup. The command starts by installing DTR,
+restores the configurations from the
+backup and then restores the repository metadata. All install time flags are the same
+for restore and apply the same way they would to a new installation.
```
joaofnfernandes: I don't really understand what you mean by "All install time flags are the same for restore and apply the same way they would to a new installation." Are you trying to explain that all configurations will be restored?

vikstrous: I'm saying that if you set any install flags, they will override any backed up configs. Some things such as port numbers are install time configs, not something that will be restored.

joaofnfernandes: ohhhh. So say it like that.

If you use any flags with the `restore` command, you'll override the settings from the backup.

datacenter/dtr/2.2/guides/admin/backups-and-disaster-recovery.md
```diff
@@ -71,34 +91,122 @@ Where:
...
 To validate that the backup was correctly performed, you can print the contents
 of the tar file created:
 
 ```none
 $ tar -tf /tmp/backup.tar
 ```
 
+The structure of the archive should look something like this:
+
+```none
```
joaofnfernandes: Can you can this sample to be just 4-6 lines. Right now we don't have vertical scroll 😭
So it will take a lot of space in the page and make it look strange.

datacenter/dtr/2.2/guides/admin/backups-and-disaster-recovery.md
```diff
@@ -71,34 +91,122 @@ Where:
...
-command.
-This command performs a fresh installation of DTR, and reconfigures it with
-the configuration created during a backup.
+You can restore a DTR node from a backup using the `restore` command.
 
-The command starts by installing DTR, restores the configurations stored on
-etcd, and then restores the repository metadata stored on RethinkDB. You
-can use the `--config-only` option, to only restore the configurations stored
-on etcd.
+Note that backups are tied to specific DTR versions and work
```
joaofnfernandes: "it works best" is technically true but I think this is one of those situation where we need to take a strong stance and say "it only works with the same version".
I bet we don't have tests to exercise this across versions so we don't provide any guarantees that it will work.

PR #1790, opened 2017-02-17
move redirects from older install page to here

londoncalling:
Moving redirects from v 1.12 CS engine install page to this newest version. 

Found these legacy redirects on Manuals tab of docs site.


cs-engine/1.13/index.md
```diff
@@ -2,6 +2,11 @@
 title: Install CS Docker Engine 1.13
 description: Learn how to install the commercially supported version of Docker Engine.
 keywords: docker, engine, install
+redirect_from:
```
joaofnfernandes: I think you forgot to commit the `cs-engine/1.12` file. This way you'll have the same redirects on both articles.

PR #1795, opened 2017-02-17
Fix command docker ps ... --name

pascalandy:
Plain easy :)


johndmulhausen:
Yep, checks out. Thanks, @pascalandy !

pascalandy:
Always welcome!

PR #1801, opened 2017-02-17
fixed problems related to tabs restructure

londoncalling:
- Fixed nav menus to surface Compose file reference

- Changed several links on Manuals home page to link to overviews instead of index pages

- Modified index pages to set `notoc: true`


londoncalling:
@johndmulhausen FYI, LGTM, merging

PR #1820, opened 2017-02-17
Typo

Deseao:

### DIFF
--- a/engine/reference/commandline/volume_create.md
+++ b/engine/reference/commandline/volume_create.md
@@ -20,7 +20,7 @@ $ docker volume create hello
 hello
 $ docker run -d -v hello:/world busybox ls /world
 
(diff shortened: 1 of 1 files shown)


Deseao:
Ah I see. Can the Edit this page link at the bottom be changed to point to the correct file?

joaofnfernandes:
Yup, right now it's not clear at all. Do you care to create an issue on this repo so that we can improve that?

mstanleyjones:
So, there are a few things going on here. For 1.13.0, we could not parse all the info we needed from the upstream files, so we actually did have the static content in our own repo (and still do). So your PR was actually correct. However, there #1829 will remove that duplicated content. The good news, is the issue raised in your PR actually is fixed upstream so when #1829 goes in, the content will be fixed.

@joaofnfernandes I'm sorry you were not fully aware of the situation.

PR #1821, opened 2017-02-17
Screenshots when switching to Windows Container

ekitagawa:
### Proposed changes

Uploading screenshots to switch from Linux to Windows container.  Now that we have Windows version of container, vast majority of Docker for Windows will shift to Windows only folks, so we have to make it very simple to switch.


londoncalling:
@ekitagawa thanks for this, you are right that we should have had that screen snap in the location where we talk about the ability to switch container types (Windows, Linux). We do have the screen snap, but only at the beginning of the `Settings` topic.

I'm going to close this PR though because we just updated the Windows menus for the next Beta release. I tagged you on PR #1930 to give you credit for the update. On that PR, I've included the update with the new menu. It's building now, but you can have a look at that topic on the Netlify preview here: https://deploy-preview-1930--docker-docs.netlify.com/docker-for-windows/#/switch-between-windows-and-linux-containers

This should be published within the hour tonight on the docs site. Thanks very much for the feedback!

PR #1828, opened 2017-02-17
Added docs for adding SANs to UCP

### LINKED ISSUE #935: Clarify UCP TLS certs and SANs
When using the default configurations, UCP uses a self-signed TLS certificate for its server component. If you're using a load balancer to load-balance requests across several UCP manager nodes, and the TLS certificates don't have a valid Subject Alternative Name, client tools are not going to trust the TLS certificate.

Explain how to add Subject Alt Names to the TLS certs used by the UCP controller service

### LINKED ISSUE #1693: [Docs] Document CLI SAN update process
@alexmavr commented

Updating SANs through the CLI has several gotchas so we should try to explain that process in detail.

Assigning to @joaofnfernandes because of our triage docs, but I can spend the time to write down the actual article

---

@joaofnfernandes commented

It would be awesome™ if you could draft something either on a PR or even just a dump here on the issue itself.

wsong:
### Proposed changes
Added a new page about adding SANs to an existing UCP cluster through the web UI or
the CLI.

### Related issues (optional)
Fixes #935.
Fixes #1693.


datacenter/ucp/2.1/guides/admin/configure/add-sans-to-cluster.md
```diff
@@ -0,0 +1,45 @@
+---
+title: Add SANs to cluster certificates
+description: Learn how to add new SANs to cluster nodes, allowing you to connect to UCP with a different hostname
+keywords: Docker, cluster, nodes, labels, certificates, SANs
+---
+
+When you connect to UCP, you need to make sure that the hostname that you use to connect is recognized
```
joaofnfernandes: This is looking great, but can we structure this intro to make it a bit more obvious there there are two choices. I would structure the intro like this

1. UCP has HTTPS out of the box. This means that you need to make sure the TLS certificates have information about the hostname you're using to access UCP. (You already cover this)
2. If you're using your own TLS certs, make sure they respect this, and learn more at `/datacenter/ucp/2.1/guides/admin/configure/use-your-own-tls-certificates.md`
3. If you want to use the self-signed certificate that UCP has out of the box, ensure the certificate respects this. Here's how you do it. (You already cover this)

If you don't do a good job of explaining these two "modes" most users will strugle with it and continue asking support for help.
Now that I think about it, I can configure UCP to use my own TLS certs, but then go to the node and add more SANS. Can you check if this still happens and file an issue with UCP so that we can make this more obvious?

datacenter/ucp/2.1/guides/admin/configure/add-sans-to-cluster.md
```diff
@@ -0,0 +1,45 @@
...
+```
+
+This will get the current set of SANs for the given manager node. Append your desired SAN to this
+list (e.g. `default-cs,127.0.0.1,172.17.0.1,example.com`) and then run:
+
+```none
+docker node update --label-add com.docker.ucp.SANs=<SANs-list> <node-id>
+```
+
+`<SANs-list>` is the list of SANs with your new SAN appended at the end. As in the web UI,
```
joaofnfernandes: Do you know if we we can do this from the CLI? I'm not seeing anything on the `docker/ucp` image. If the only way to do this is using our internal API, then let's not talk about it 😅

wsong: @alexmavr can weigh in more, but I'm pretty sure this is the "right" way to do this on the CLI today.

alexmavr: That's the supported CLI version, no internal APIs are involved in these instructions whatsoever

datacenter/ucp/2.1/guides/admin/configure/add-sans-to-cluster.md
wsong: Updated the intro.

If you use your own TLS certs, you actually can add SANs as much as you want; nothing will break. We will happily add the new SANs to those certs.

PR #1829, opened 2017-02-17
Remove duplicated content from CLI ref stub files

mstanleyjones:
Thanks to @FrenchBen we now pull long descriptions and examples for the Engine Reference CLI from the reference files in `docker/docker`. This PR removes the redundant info that we brute-force added to the stub files for the 1.13.0 release. This also adds a README to explain the process for generating the CLI reference YAML. @FrenchBen PTAL at the readme and make sure I explained the process correctly.


johndmulhausen:
LGTM :) Thanks Misty, I'm sure this was tedious work. Feel free to wait for FrenchBen before merging.

engine/reference/commandline/README
```diff
@@ -0,0 +1,26 @@
...
+which parses YAML files generated from the
+[`docker/docker`](https://github.com/docker/docker) repository. The YAML files
+are parsed into output files like
+[https://docs.docker.com/engine/reference/commandline/build/](https://docs.docker.com/engine/reference/commandline/build/).
+
+## How the output is generated
+
+The output files are composed from two sources:
+
+- The **Description**, **Usage**, and **Description** sections comes directly from
```
FrenchBen: `**Description**` is in that list twice

engine/reference/commandline/README
```diff
@@ -0,0 +1,26 @@
...
+the CLI source code in that repository.
+
+- The **Extended Description** and **Examples** sections are pulled into the
+  YAML from the files in [https://github.com/docker/docker/tree/master/docs/reference/commandline](https://github.com/docker/docker/tree/master/docs/reference/commandline)
+  Specifically, the Markdown inside the `## Description` and `## Examples`
+  headings are parsed. Please submit corrections to the text in that repository.
+
+# Updating the YAML files
+
+To generate the YAML files, run `make yamlfiles` in the `docker/docker` branch
```
FrenchBen: My PR was shut down, @thaJeztah has a different process in mind he'd like to implement

thaJeztah: @tiborvass was looking into it; the TL;DR was that the actual YAML-generator would never be in sync with the code-base, so result in generating the wrong options/flags

mstanleyjones: Adjusted the instructions for the `docker/docker` side to say to talk to @frenchben or @thajestah 👼

mstanleyjones:
Merging for now, just so we can get rid of the duplicate content.

PR #1834, opened 2017-02-17
no need for `grep -v` you can use `grep [d]ocker`

dominis:

### DIFF
--- a/engine/admin/index.md
+++ b/engine/admin/index.md
@@ -159,7 +159,7 @@ you installed Docker, you may already have this file.
 7. Verify that the `docker` daemon is running as specified with the `ps` command.
 
     ```bash
-    $ ps aux | grep docker | grep -v grep
+    $ ps aux | grep [d]ocker
     ```
 
 ### Logs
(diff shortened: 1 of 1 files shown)


engine/admin/index.md
```diff
@@ -159,7 +159,7 @@ you installed Docker, you may already have this file.
 7. Verify that the `docker` daemon is running as specified with the `ps` command.
 
     ```bash
-    $ ps aux | grep docker | grep -v grep
+    $ ps aux | grep [d]ocker
```
joaofnfernandes: Can you explain why you're going `grep [d]ocker` instead of `grep docker`? Isn't it the exact same thing?

And the `grep -v grep` is just removing noise from the output since without it you'll have one line like
```
docker   17370  0.0  0.1   9760  1040 pts/0    S+   22:58   0:00 grep docker
```

mstanleyjones: It's a little trick that's hard to explain, but the `grep` result from `ps` will contain the bracket so it won't match the regular expression. Try it and see.

dominis: @joaofnfernandes http://askubuntu.com/a/153430

joaofnfernandes: The trick makes sense, but I think that documentation (and code samples) should be explicit even if it is a bit more verbose. This minimizes the prior knowledge you need to have before starting to learn Docker.

So I think we should keep the snippet as it is today, but I'll defer to @mstanleyjones on this.

mstanleyjones:
I agree. I think we can't expect everyone who uses Docker to be a shell ninja. :)

PR #1835, opened 2017-02-17
UCP upgrade offline for 2.1

### LINKED ISSUE #1576: How to upgrade DDC offline
@joaofnfernandes commented

The process is essentially the same as the regular upgrade, you just need to fetch the latest images and preload them into the nodes before running the upgrade command.

(From dhiltgen)
Further, we should talk about pre-loading the images for people that use offline bundles - they need to pre-load the image everywhere before starting the upgrade. I haven't tested, but from memory, I'm pretty sure if the admin does the docker load < ucp_images.2.1.0.tar.gz (or whatever the files is called... typing from memory) while connected to UCP itself, then the images will be loaded on all nodes, which is a lot less tedious than loading one at a time on every underlying node.
Note: we also have "load image" in the GUI now, so they can just do this from the comfort of their browser

### LINKED ISSUE #1628: Feedback for: datacenter/ucp/2.1/guides/admin/install/install-offline.md
[datacenter/ucp/2.1/guides/admin/install/install-offline.md](https://docs.docker.com/datacenter/ucp/2.1/guides/admin/install/install-offline/)

Remove DTR from this table, it doesn't belong.

ry4nz:
### Proposed changes
Add Upgrade offline section
Adjust the upgrade section so it is no longer a subsection of install
Display the bundle link and it only shows UCP images

Signed-off-by: Ryan Zhang <ryan.zhang@docker.com>

### Related issues (optional)

Fixes #1576 
Fixes #1628

### DIFF
--- a/datacenter/ucp/2.1/guides/admin/upgrade/upgrade-offline.md
+++ b/datacenter/ucp/2.1/guides/admin/upgrade/upgrade-offline.md
@@ -1,17 +1,17 @@
 ---
-description: Learn how to install Docker Universal Control Plane. on a machine with
+description: Learn how to upgrade Docker Universal Control Plane. on a machine with
   no internet access.
-keywords: docker, ucp, install, offline
-title: Install UCP offline
+keywords: docker, ucp, upgrade, offline
+title: Upgrade UCP offline
 ---
 
-The procedure to install Docker Universal Control Plane on a host is the same,
+The procedure to upgrade Docker Universal Control Plane on a host is the same,
 whether that host has access to the internet or not.
 
-The only difference when installing on an offline host,
+The only difference when upgrading on an offline host,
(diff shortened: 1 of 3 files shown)


datacenter/ucp/2.1/guides/admin/upgrade/upgrade-offline.md
```diff
@@ -0,0 +1,60 @@
+---
```
joaofnfernandes: I think this will require a bit more work, since this is what I'm seeing right now

screen shot 2017-02-17 at 15 36 19

Since this change looks good, do you mind submitting a PR containing just the upgrade offline article with the necessary change to the `_data/toc.yaml` file, but keep reusing the `ddc_url_list.html component`?

Then we can work on another PR to separate the UCP and DTR offline tar files.

Issue #1816, opened 2017-02-17
Feedback for: compose/reference/scale.md


funkydorian:
[compose/reference/scale.md](https://docs.docker.com/compose/reference/scale/), CC: @londoncalling 

The alternative to that in docker-compose.yml which is "replicas" can be indicated also in that page.

PR #1837, opened 2017-02-17
added x-ref from scale reference to replicas in v3 Compose

londoncalling:
### What changed

- Added x-ref to `replicas` from `scale` in Compose reference per user feedback

### Related issues

#1816 

### Reviewers

@funkydorian, @shin- 

### DIFF
--- a/_layouts/docs.html
+++ b/_layouts/docs.html
@@ -301,7 +301,6 @@ ng\:form {
 								{% endif %}
 								<div id="side-toc"><div id="side-toc-title">On this page:</div><div id="side-toc-contents">{% include toc_pure_liquid.html html=my_content sanitize=true class="inline_toc" id="my_toc" toc_min=my_min toc_max=my_max page_name=my_name %}</div></div>
 								{% endunless %}
-								<div>{{content}}</div>
                 </section>
             </div>
           </div>
--- a/compose/reference/scale.md
+++ b/compose/reference/scale.md
@@ -13,3 +13,9 @@ Sets the number of containers to run for a service.
 Numbers are specified as arguments in the form `service=num`. For example:
 
     docker-compose scale web=2 worker=3
+
+>**Tip:** Alternatively, in
+[Compose file version 3.x](/compose/compose-file/index.md), you can specify
+[`replicas`](/compose/compose-file/index.md#replicas)
+under [`deploy`](/compose/compose-file/index.md#deploy) as part of the
+service configuration for [Swarm mode](/engine/swarm/).


shin-:
cc @dnephin 

I guess it's similar in effect but not exactly equivalent.

compose/reference/scale.md
```diff
@@ -13,3 +13,10 @@ Sets the number of containers to run for a service.
 Numbers are specified as arguments in the form `service=num`. For example:
 
     docker-compose scale web=2 worker=3
+
+>**Tip:** Alternatively, in
+[Compose file version 3.x](/compose/compose-file/index.md)
+with Docker Engine 1.13 and above, you can specify
```
shin-: "Engine 1.13" is redundant since we're already talking about "Compose file version 3.x"

compose/reference/scale.md
```diff
@@ -13,3 +13,10 @@ Sets the number of containers to run for a service.
...
+[`replicas`](/compose/compose-file/index.md#replicas) under
+[`deploy`](/compose/compose-file/index.md#deploy) as part of the service
```
shin-: I'll leave it up to you, but is "under deploy" here really necessary? It should become obvious as soon as readers click through the `replicas` link.

compose/reference/scale.md
```diff
@@ -13,3 +13,10 @@ Sets the number of containers to run for a service.
...
+configuration for a [swarm](/engine/swarm/).
```
shin-: "for Swarm mode"

londoncalling:
@shin- , thanks, updated per your comments. 

Per your suggestion, I tried leaving out the additional reference to "under `deploy`" but the way the examples are shown in the Compose file, seems more thorough with it in. It might be a little overkill, but I don't think it hurts anything, and possibly adds clarity.

I'll wait and see if @dnephin weighs in now.

londoncalling:
@shin- @dnephin @mstanleyjones leveraged (and fixed a small issue) in #1842 which enabled us to surface H1 topic titles in the right side "on-page" navigation menus if needed. I think this is good for these Compose file format references as an extra reminder of which version you are looking at.

Merging this as soon as it builds and passes checks to get these fixes in (especially fix re: #1842  because it has global impact on docs).

@dnephin feel free to suggest any changes to the tip on `scale.md` if you don't like it.

_layouts/docs.html
```diff
@@ -301,7 +301,6 @@
 								{% endif %}
 								<div id="side-toc"><div id="side-toc-title">On this page:</div><div id="side-toc-contents">{% include toc_pure_liquid.html html=my_content sanitize=true class="inline_toc" id="my_toc" toc_min=my_min toc_max=my_max page_name=my_name %}</div></div>
 								{% endunless %}
-								<div>{{content}}</div>
```
dnephin: This seems unrelated. I guess it was just a cleanup?

compose/reference/scale.md
```diff
@@ -13,3 +13,9 @@ Sets the number of containers to run for a service.
 Numbers are specified as arguments in the form `service=num`. For example:
 
     docker-compose scale web=2 worker=3
+
+>**Tip:** Alternatively, in
+[Compose file version 3.x](/compose/compose-file/index.md), you can specify
+[`replicas`](/compose/compose-file/index.md#replicas)
+under [`deploy`](/compose/compose-file/index.md#deploy) as part of the
+service configuration for [Swarm mode](/engine/swarm/).
```
dnephin: Do you think it's worth including the caveat that `replicas` only works for `docker stack deploy` and not `docker-compose up`?

funkydorian: i think this should be indicated as today one of developer over here got mixed up because of that

PR #1839, opened 2017-02-17
improve dtr upgrade docs

vikstrous:


datacenter/dtr/2.2/guides/admin/upgrade/index.md
```diff
@@ -4,34 +4,61 @@ keywords: docker, dtr, upgrade, install
...
-is ensuring you're running DTR 2.1. If that's not the case, start by upgrading
-your installation to version 2.1, and then upgrade to 2.2.
+DTR uses [semantic versioning](http://semver.org/) and we aim to achieve specific
+guarantees while upgrading between versions. We never support downgrading. We
+support upgrades according to the following rules:
 
-There is no downtime when upgrading a highly-available DTR cluster. If your
-DTR deployment has a single replica, schedule the upgrade to take place outside
-business peak hours to ensure the impact on your business is close to none.
+* When upgrading from one patch version to another you can skip patch versions
```
joaofnfernandes: To make it a bit more organized maybe we can put this in a table like

|From| To| Description| Supported| 
|:----|:---|:------------|----------|
|x.x.0| x.x.1| From one patch to the one after| yes|
|x.x.0| x.x.5| Skipping patches| yes|
|x.0.x| x.1.x| From one minor version to the next| yes|
|x.0.x| x.5.x| Skip minor versions| No. You can't skip minor versions|

datacenter/dtr/2.2/guides/admin/upgrade/index.md
```diff
@@ -4,34 +4,61 @@ keywords: docker, dtr, upgrade, install
 title: Upgrade DTR
 ---
 
-The first step in upgrading to a new minor version or patch release of DTR 2.2,
-is ensuring you're running DTR 2.1. If that's not the case, start by upgrading
-your installation to version 2.1, and then upgrade to 2.2.
+DTR uses [semantic versioning](http://semver.org/) and we aim to achieve specific
```
joaofnfernandes: The left nav now has a section just about upgrading, so it makes sense for us to spin this information about upgrade policies into it's own article.

vikstrous: let's merge this first and then decide separately if we should split the pages?

joaofnfernandes: Also works.

datacenter/dtr/2.2/guides/admin/upgrade/index.md
```diff
@@ -4,34 +4,65 @@ keywords: docker, dtr, upgrade, install
...
 
-## Step 1. Upgrade DTR to 2.1
+|From| To| Description| Supported|
+|:----|:---|:------------|----------|
+| 2.2.0 | 2.2.1 | patch upgrade | yes |
+| 2.2.0 | 2.2.2 | skip patch version | yes |
+| 2.2.2 | 2.2.1 | patch downgrade | no |
+| 2.1.0 | 2.2.0 | minor upgrade | yes |
+| 2.1.1 | 2.2.0 | minor upgrade | yes |
+| 2.1.2 | 2.2.2 | minor upgrade | yes |
```
alexmavr: @vikstrous what about 2.1.3 and 2.1.4? why are these missing from this table?

datacenter/dtr/2.2/guides/admin/upgrade/index.md
```diff
@@ -43,6 +74,16 @@ By default the upgrade command runs in interactive mode and prompts you for
...
 
+The upgrade command will start replacing every container in your DTR cluster,
+one replica at a time. It will also perform certain data migrations. If anything
+fails or the upgrade is interrupted for any reason, you can re-run the upgrade
+command and it will resume from where it left off.
+
+## Patch upgrade
+
+A patch upgrade changes only the DTR containers and it should be a relatively
+safe operation. The command is the same as for a minor upgrade.
```
alexmavr: what is relatively safe? what's the difference with "absolutely" safe? is there any downtime involved at all and, if so, how long does it last?

datacenter/dtr/2.2/guides/admin/upgrade/index.md
vikstrous: These are examples

datacenter/dtr/2.2/guides/admin/upgrade/index.md
vikstrous: My bad. That was really vague. I guess I was trying to say is that it's safer than a minor upgrade because it doesn't do any data migration.

joaofnfernandes: I see a lot of sentences like that in docs related to install, upgrade, and security. Generally I agree with @alexmavr and avoid sentences like that altogether.
When we say "this is safer", "there are risks" or things like that:

* Expert users will know it (probably better than us), so they won't learn anything
* Novice users will be scared because now we're telling them that something can go wrong, but won't know what to do, since we're not teaching them the subtle differences and what to do in each case.

So in most cases both users are better off if we don't even include that sentence.

PR #1842, opened 2017-02-18
Allow h1 in in-page toc

mstanleyjones:
#1474 added the ability to control the minimum and maximum level of HTML headings that would display in the in-page TOC for a given page. This PR fixes a bug that presented `toc_min: 1` from being honored. It also adds documentation about valid front-matter elements to the `README`, for the edification of all.

To test: Check out this PR, edit a page (maybe `testing.md`), set `toc_min: 1`, and verify that the page title is in the in-page TOC and when you click it, you get to the `#` anchor.

### DIFF
--- a/_layouts/docs.html
+++ b/_layouts/docs.html
@@ -289,8 +289,19 @@ ng\:form {
 								{% assign my_min = page.toc_min | default: site.toc_min | default: 2 %}
 								{% assign my_max = page.toc_max | default: site.toc_max | default: 3 %}
 								{% assign my_name = page.url | default: "unnamed" %}
-								<div id="side-toc"><div id="side-toc-title">On this page:</div><div id="side-toc-contents">{% include toc_pure_liquid.html html=content sanitize=true class="inline_toc" id="my_toc" toc_min=my_min toc_max=my_max page_name=my_name %}</div></div>
+								{% if my_min < 2 %}
+									{% comment %}
+										the 'content' variable doesn't include the <h1> by default
+										so if we request the h1 heading to be in the in-page TOC,
+										append it to the content before we pass it to the TOC generator
+									{% endcomment %}
+									{% assign my_content = "<h1>" | append: page.title | append: "</h1>" | append: content %}
+								{% else %}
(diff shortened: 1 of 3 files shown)


mstanleyjones:
@londoncalling It's merged! :1st_place_medal:

_layouts/docs.html
```diff
@@ -289,8 +289,19 @@
...
 								{% assign my_max = page.toc_max | default: site.toc_max | default: 3 %}
 								{% assign my_name = page.url | default: "unnamed" %}
-								<div id="side-toc"><div id="side-toc-title">On this page:</div><div id="side-toc-contents">{% include toc_pure_liquid.html html=content sanitize=true class="inline_toc" id="my_toc" toc_min=my_min toc_max=my_max page_name=my_name %}</div></div>
+								{% if my_min < 2 %}
+									{% comment %}
+										the 'content' variable doesn't include the <h1> by default
+										so if we request the h1 heading to be in the in-page TOC,
+										append it to the content before we pass it to the TOC generator
+									{% endcomment %}
+									{% assign my_content = "<h1>" | append: page.title | append: "</h1>" | append: content %}
```
allejo: Ahhh I _think_ if an ID attr is added to this `<h1>` (i.e. make it `<h1 id="page-title">`) then the change to the `_includes/toc_pure_liquid.html` wouldn't be necessary. Just as a quick note 👍

mstanleyjones: You might think that, but actually when you give it an ID then the link it generates encodes the ID. So if you do `<h1 id="page-title-in-toc">...</h1>`, then the link to the `<h1>` in the TOC becomes: `http://localhost:4000/compose/compose-file/#/page-title-in-toc`

It doesn't really make it do the wrong thing, since that ID doesn't really exist on the page, but it's kind of not really what you want either. I'm OK with the fix we have (and the follow-on PR where @londoncalling removed my inadvertent extra `{{content}}` div that I was using for debugging!

allejo: Ohhh derp! I see what you mean, good call

PR #1846, opened 2017-02-18
removed on-page topic menu from Compose CLI ref pages

londoncalling:
Removed on-page TOC menus from Compose CLI reference pages with one or fewer subtopics on the page, since it's not needed.


londoncalling:
Well, inexplicably now these CLI reference commands seem to be displaying properly on the current docs in master _without_ the `notoc: true` in the header.

 But I'm going to merge this anyway. I don't think it can do any harm, and Netlify looks good.

Issue #1733, opened 2017-02-16
Improve DTR caching docs


joaofnfernandes:
We already have docs for DTR caching, but users are flooding support and the field team with questions, so we should create more content around this.

- [x] What is a DTR cache and why you should care
- [x] How cache works behind the scenes (network flows)
- [x] Example for IT ops to deploy a simple cache
- [x] Example for IT ops to deploy cache with TLS
- [x] Example for IT ops to deploy complex cache (chaining)
- [ ] How can IT ops configure multiple users to use a cache (scripting)
- [ ] How can IT ops monitor a cache and check if it's running or not
- [ ] How can IT troubleshoot if a cache goes down or has problems
- [x] How can developers choose a cache
- [ ] How can developers troubleshoot if a cache is down and what are the consequences

Bonus points: How to set up geo-based caching

---
This is the documentation we have right now
- [Deploy a cache](https://docs.docker.com/datacenter/dtr/2.2/guides/admin/configure/deploy-a-cache/)
- [Use a cache](https://docs.docker.com/datacenter/dtr/2.2/guides/user/access-dtr/use-a-cache/)

hinshun:
#dibs

joaofnfernandes:
Most of the issues here where addressed, so this is no longer a P0

johndmulhausen:
Closing per triage discussion

PR #1847, opened 2017-02-18
Update documentation for Docker Content Cache

hinshun:
Related to: #1733

- Revised vocabulary to be consistent using 'Docker Content Cache'
- Added 'What is Docker Content Cache' section
- Added examples for deploying with TLS and cache chaining
- Moved configuration table before examples section


datacenter/dtr/2.2/guides/admin/configure/deploy-a-cache.md
```diff
@@ -1,57 +1,168 @@
...
+    no
+  </td>
+  <td>
+    A list of externally-reachable addresses for upstream registries for cache chaining. If more than one host is specified, pulls from upstream content caches will be done in round-robin order.
+  </td>
+</tr>
+</table>
 
-## Deploy a DTR cache
+## Deploying a simple Docker Content Cache
```
joaofnfernandes: Our titles use the simple form instead of gerund.
Deploying -> Deploy

datacenter/dtr/2.2/guides/admin/configure/deploy-a-cache.md
```diff
@@ -1,57 +1,168 @@
...
 
-You can deploy a DTR cache on any host that has Docker installed. The only
-requirements are that:
+You can deploy a Docker Content Cache on any host that has Docker installed.
+The only requirements are that:
 
 * Users need to have access to both DTR and the cache
 * The cache needs access to DTR
```
joaofnfernandes: Do all the caches need access to DTR, or they just need access to their upstreams?

datacenter/dtr/2.2/guides/admin/configure/deploy-a-cache.md
```diff
@@ -96,7 +207,7 @@ docker run --detach --restart always \
   docker/dtr-content-cache:<version> /config.yml
 ```
 
-You can also run the command in interactive mode instead of detatched by
+You can also run the command in interactive mode instead of detached by
 replacing `--detached` with `--interactive`. This allows you to
```
joaofnfernandes: --detached -> --detach

datacenter/dtr/2.2/guides/admin/configure/deploy-a-cache.md
```diff
@@ -140,96 +251,135 @@ In the host where you've deployed the `region-us` cache, run:
...
-  </td>
-  <td>
-    A list of externally-reachable addresses for upstream registries for cache chaining. If more than one host is specified, pulls from upstream content caches will be done in round-robin order.
-  </td>
-</tr>
-</table>
+### Let's Encrypt
+
+Content Cache also supports using Let's Encrypt to automatically obtain a
+browser-trusted certificate.  For more information on Let's Encrypt, see
```
joaofnfernandes: I'm not sure browser-trusted certificate is really a thing. Maybe

You can also use Let's Encrypt to generate certificates for your caches.

datacenter/dtr/2.2/guides/admin/configure/deploy-a-cache.md
```diff
@@ -140,96 +251,135 @@ In the host where you've deployed the `region-us` cache, run:
...
+[https://letsencrypt.org/how-it-works/](https://letsencrypt.org/how-it-works/)
+and the relevant section of the [registry configuration](/registry/configuration.md#letsencrypt).
+
+### Alternatives
```
joaofnfernandes: I think this section adds too much ambiguity. When we say `While rarely advisable, you may want to use self-signed certificates`, this is very clear in your head what you mean, but consider that

1. Users that know the consequences of using self-signed certs don't really need that disclaimer
2. Users who don't are even worse. You've scared them but didn't really taught them what are the consequences of using self-signed certs.

Just remove the section.

datacenter/dtr/2.2/guides/admin/configure/deploy-a-cache.md
```diff
@@ -1,57 +1,168 @@
 ---
-title: Cache Docker images
-description: Learn how to configure DTR to have caching and making pulls faster.
+title: Configure your Docker Content Cache
```
hinshun: I wanted to include the sentence `service that is deployed separately from DTR` because of the confusion reported from field engineers that didn't know that it's a separate service. The importance is not that its a different image, but rather the deployment should be on a different machine because if they are together then the cache is useless.

@pdevine If we do word it as a feature like `Cache Docker images`, then we should change our frontend user settings to reflect that. Because currently it's called `Content Cache` in DTR. What should we do?

datacenter/dtr/2.2/guides/admin/configure/deploy-a-cache.md
```diff
@@ -1,57 +1,168 @@
...
-pull Docker images from a cache that is geographically closer to them or
-that allows them faster downloads.
+## What is Docker Content Cache
 
-## How DTR caches works
+Docker Content Cache is a service that is deployed separately from DTR, and
+can be configured to securely cache images from DTR. Users can configure
+their DTR user accounts to specify which cache to pull from. Then, when a
+user pulls from DTR, they will be redirected to pull from their chosen
+Content Cache. By deploying caches geographically closer to remote offices
```
hinshun: Just to clarify geolocation is not supported by the product.
I can make the suggested change.

datacenter/dtr/2.2/guides/admin/configure/deploy-a-cache.md
```diff
@@ -1,57 +1,168 @@
...
+and low connectivity areas, users can benefit from faster pulls.
 
-You start by deploying one or more DTR caches.
+## Authentication and Freshness
+
+Pulling from Content Cache still enforces authentication, ensuring that
+images users don't have access to on DTR remain protected on the cache.
```
hinshun: Yeah that's better, thanks.

datacenter/dtr/2.2/guides/admin/configure/deploy-a-cache.md
```diff
@@ -1,57 +1,168 @@
...
+However, there are caveats associated with this. You should avoid creating
+a topology that has chained caches of more than two levels as the total
+round trip time between the caches may become higher than the time saved
+from caching the image.
 
-In this example, users can go to their DTR profile page, and configure their
-user account to use the US cache. Then, when using the
+## How it works
+
+After you've deployed the caches, user can configure which cache to
```
hinshun: :+1:

datacenter/dtr/2.2/guides/admin/configure/deploy-a-cache.md
hinshun: :+1:

datacenter/dtr/2.2/guides/admin/configure/deploy-a-cache.md
hinshun: Only their upstreams, which may or may not be DTR, so we should change this.

datacenter/dtr/2.2/guides/admin/configure/deploy-a-cache.md
hinshun: :+1:

datacenter/dtr/2.2/guides/admin/configure/deploy-a-cache.md
hinshun: I copied this text mainly from: https://docs.docker.com/registry/deploying/#/lets-encrypt
If this is a problem we should change it in both locations.

datacenter/dtr/2.2/guides/admin/configure/deploy-a-cache.md
hinshun: :+1: Though this is also in: https://docs.docker.com/registry/deploying/#alternatives-1

datacenter/dtr/2.2/guides/admin/configure/deploy-a-cache.md
joaofnfernandes: @pdevine we we're discussing this the other day. Here's the current behavior ^^

datacenter/dtr/2.2/guides/admin/configure/deploy-a-cache.md
joaofnfernandes: I know. The distribution docs are in need of some tough love 😅

datacenter/dtr/2.2/guides/admin/configure/deploy-a-cache.md
joaofnfernandes: Yup. We need to improve the distribution docs to remove ambiguities like that. Don't use those docs as a reference of the quality we should be aiming for.

PR #1852, opened 2017-02-18
fix in logging/overview.md

rafaeleyng:
Changes `<NAME>-<VALUE>` to `<NAME>=<VALUE>`


johndmulhausen:
Ah, thanks @rafaeleyng :)

PR #1866, opened 2017-02-20
Update trust_delegation.md

scjane:
Fix subject-pronoun disaggrement.


johndmulhausen:
Thanks, @scjane :)

engine/security/trust/trust_delegation.md
```diff
@@ -141,11 +141,11 @@ IDs to collaborators yourself should you need to remove a collaborator.
 
 ## Removing a delegation key from an existing repository
 
-To revoke a collaborator's permission to sign tags for your image repository, you must
-know the IDs of their keys, because you need to remove their keys from the
-`targets/releases` delegation.
+To revoke a collaborator's ability to sign tags for your image repository, you
```
mstanleyjones: I kept this singular because plural adjectives can be difficult for some readers to parse. Thanks for the edits!

scjane: No problem :) 

In fact, I wanted to use singular at first and changed "their" to "his" or "her" which is acceptable as described in the style guide, but it still sounded a little weird to me. So, I switched to plural😌

PR #1867, opened 2017-02-20
Update deploying_notary.md

scjane:
Add an ending period for step 1 as it is a complete setence and step 2 and step 3 both end with a period.


johndmulhausen:
Fair enough, thanks @scjane :)

PR #1874, opened 2017-02-20
Update systemd.md

r1k0:
restart docker before showing the Environment variable.
https://github.com/docker/docker/issues/22568#issuecomment-281092636

### DIFF
--- a/engine/admin/systemd.md
+++ b/engine/admin/systemd.md
@@ -182,16 +182,17 @@ you will need to add this configuration in the Docker systemd service file.
     $ sudo systemctl daemon-reload
     ```
 
-5.  Verify that the configuration has been loaded:
+5.  Restart Docker:
+
+    ```bash
+    $ sudo systemctl restart docker
+    ```
+
+6.  Verify that the configuration has been loaded:
 
     ```bash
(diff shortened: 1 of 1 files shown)


engine/admin/systemd.md
```diff
@@ -182,16 +182,17 @@ you will need to add this configuration in the Docker systemd service file.
     $ sudo systemctl daemon-reload
     ```
 
-5.  Verify that the configuration has been loaded:
+5.  Restart Docker:
```
mstanleyjones: Are you sure this is needed? I don't think so, because the `daemon-reload` above will send a HUP to the daemon, which will make it re-read its configuration. Did you test to make sure it's needed?

r1k0: Did you read the context issue?
Did you actually test the doc?

r1k0:
feel free to remove. no time to debate.

ricardogsilva:
I just stumbled upon this and I can confirm that (at least on ubuntu server 16.04) restarting docker is needed in order to have the ssh-proxy environment variable showing. So IMHO the changes originally proposed do make sense and should be merged.

PR #1876, opened 2017-02-21
Fixed volume cli output

FrenchBen:
### Proposed changes

Added raw comments around the CLI output, which was getting truncated.


mstanleyjones:
I think #1892 fixed the ones of these that actually needed fixing, and the others are superfluous. You only need the `{% raw %}` if you have something surrounded by double curly braces like `{{ thing }}`, because Jekyll / Liquid will try to interpret it as Liquid templating language otherwise.

PR #1879, opened 2017-02-21
Update seccomp.md

scjane:


johndmulhausen:
Always better to do this. Thanks, @scjane :)

PR #1888, opened 2017-02-21
Via @thaJeztah we no longer support Wheezy

fxdgear:
https://github.com/docker/docker/issues/15692#issuecomment-174098882

### DIFF
--- a/engine/installation/linux/debian.md
+++ b/engine/installation/linux/debian.md
@@ -20,7 +20,6 @@ Raspbian versions:
 
 - Stretch (testing)
 - Jessie 8.0 (LTS) / Raspbian Jessie
-- Wheezy 7.7 (LTS)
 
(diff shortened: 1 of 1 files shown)


thaJeztah:
Hm, actually, I may have been mistaken there (sorry it's been a while), I thought it was discussed at some point, but not sure what the current status is. Not sure if we should _recommend_ it though, with "stretch" probably becoming the current version soon

fxdgear:
> Not sure if we should recommend it though

So maybe we should remove it from the docs? And then if people want to use it they can.

mstanleyjones:
@fxdgear as I pointed out, there are additional steps for Wheezy that we can't pull until we no longer support it.

mstanleyjones:
Closing for now. Sorry about the confusion.

PR #1890, opened 2017-02-21
Fix broken DTR storage link

alexmavr:
Signed-off-by: Alex Mavrogiannis <alex.mavrogiannis@docker.com>
I'm not sure if the missing page was fully moved to the index, but this link will at least work now

I'm actually not sure if this target link is correct, or whether I should redirect to `./index.md`

### Proposed changes
Fixed a broken link, duh

### DIFF
--- a/datacenter/dtr/2.2/guides/admin/configure/external-storage/nfs.md
+++ b/datacenter/dtr/2.2/guides/admin/configure/external-storage/nfs.md
@@ -70,4 +70,4 @@ add it back again.
 
 ## Where to go next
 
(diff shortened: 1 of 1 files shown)


joaofnfernandes:
Pointing to the index should to the trick, and the syntax is `index.md`. Since this is a simple change I'm submitting #1903 and closing this one. For more involved changes I'll push to your branch :)

joaofnfernandes:
What's interesting is that your PR was passing the url checker. So that means that Jekyll also accepts the `../` format or the url checker is buggy.

PR #1892, opened 2017-02-21
Fix Liquid errors introduced recently

mstanleyjones:
Some template code got in without being wrapped in `{% raw %}` tags.


mstanleyjones:
Fixes #1876

nathanleclaire:
👍

PR #1905, opened 2017-02-22
Convert Glossary to be procedurally generated from YAML

johndmulhausen:
As the title specifies; allows us to do things like grab the definition for a term anywhere we are in the site.

Also, we are cross-referencing between the glossary and the keywords in topic metadata.

Demo: 
https://deploy-preview-1905--docker-docs.netlify.com/glossary/


johndmulhausen:
@mstanleyjones The short answer is: we won't sync with the `docker/docker` repo's Glossary anymore. Essentially we have to worry about more platform-wide things than the engine does.

mstanleyjones:
LGTM except for the buttons for related topics...

Issue #1875, opened 2017-02-20
Feedback for: engine/getstarted-voting-app/deploy-app.md


SegersIan:
[engine/getstarted-voting-app/deploy-app.md](https://docs.docker.com/engine/getstarted-voting-app/deploy-app/), CC: @mstanleyjones

In the very last step where they instruct you to run the command: `docker stack services <APP-NAME>` the output mismatched for me:

| ID       |     NAME    |         MODE   |     REPLICAS  IMAGE |
| --- |---| --- |--- |
| ...  | vote_db     |     replicated  0/1    |   postgres:9.4 |
| ...  | vote_visualizer | replicated  0/1  |      dockersamples/visualizer:stable |
| ...  | vote_worker   |   replicated  0/1     |  dockersamples/examplevotingapp_worker:latest |
| ...  | vote_redis     |  replicated  0/2     |  redis:alpine |
| ...  | vote_result    |  replicated  0/**1**      | dockersamples/examplevotingapp_result:before |
| ...  | vote_vote       | replicated  1/2      | dockersamples/examplevotingapp_vote:before |

There is only **1** replica for `vote_result`. For some reason, the majority of my services didn't launch but I guess that is because I made a mistake myself while following the tutorial.

I guess the issue is that this [commit](https://github.com/docker/example-voting-app/commit/45eb3aaf9d8373294936ddb4aff0d3b5ea023553) makes the documentation instructions inconsistent. Not sure how this change might escalate further in the tutorial.

mstanleyjones:
@SegersIan can you please verify that the stack really didn't deploy? It can take some time. Services are "eventually consistent" so not everything happens right away. You can do `docker stack ls` and `docker service ls` to see.

SegersIan:
The stack did deploy in the end correctly except for the `vote_result`. It did stick with just one replicate instead of two. So I guess it's just a matter of updating the *Example output* in the documentation. ( I know, we're talking really about a nitpick minor details here )

londoncalling:
@SegersIan thanks for the feedback. I'm cc'ing @ManoMarks as I cannot replicate the problem you are seeing. (Also ccing @marcosnils and @AdHominem who commented on the referenced [commit](https://github.com/docker/example-voting-app/commit/45eb3aaf9d8373294936ddb4aff0d3b5ea023553).) 

[...]

PR #1906, opened 2017-02-22
adding new images for change in voting app .yml

londoncalling:
Update voting app example walkthrough per changes to `.yml` (`result` service has only one replica)

@ManoMarks 


londoncalling:
fixes for issue #1875

londoncalling:
LGTM, merging

PR #1912, opened 2017-02-22
Update content_trust.md

scjane:
Is "can" more appropriate here? I think although "must" can imply suggestions that someone should do something, it's better to use the most apparent meaning of a word in technical docs. Or, did I miss something?

### DIFF
--- a/engine/security/trust/content_trust.md
+++ b/engine/security/trust/content_trust.md
@@ -19,7 +19,7 @@ ability to use digital signatures for data sent to and received from remote
 Docker registries. These signatures allow client-side verification of the
 integrity and publisher of specific image tags.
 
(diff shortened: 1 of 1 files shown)


engine/security/trust/content_trust.md
```diff
@@ -19,7 +19,7 @@ ability to use digital signatures for data sent to and received from remote
 Docker registries. These signatures allow client-side verification of the
 integrity and publisher of specific image tags.
 
-Currently, content trust is disabled by default. You must enable it by setting
+Currently, content trust is disabled by default. You can enable it by setting
```
mstanleyjones: Why not "To enable it, set..."?

PR #1914, opened 2017-02-22
Fixing a typo

arun-gupta:
Instructions says to list name but the format lists container id


arun-gupta:
@mstanleyjones updated

PR #1921, opened 2017-02-22
Fix error about permissions in secrets doc

mstanleyjones:
### Proposed changes

Fix error about permissions in secrets doc
Also link to a resource for figuring out the right mode, since not all our readers are UNIX gurus.

### DIFF
--- a/compose/compose-file/index.md
+++ b/compose/compose-file/index.md
@@ -911,9 +911,10 @@ the service's task containers.
 - `mode`: The permissions for the file that will be mounted in `/run/secrets/`
   in the service's task containers, in octal notation. For instance, `0444`
   represents world-readable. The default in Docker 1.13.1 is `0000`, but will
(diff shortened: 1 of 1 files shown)


compose/compose-file/index.md
```diff
@@ -909,9 +909,14 @@ the service's task containers.
   `/run/secrets/` in the service's task containers. Both default to `0` if not
   specified.
 - `mode`: The permissions for the file that will be mounted in `/run/secrets/`
-  in the service's task containers, in octal notation. For instance, `0777`
+  in the service's task containers, in octal notation. For instance, `0444`
   represents world-readable. The default in Docker 1.13.1 is `0000`, but will
-  be `0444` in the future.
+  be `0444` in the future. Secrets cannot be writable because they are mounted
+  in a temporary filesystem, so if you set the writable bit, it is ignored. The
```
diogomonica: in a read-only temporary filesystem

Just because it is temporary doesn't mean it has to be RO

PR #1922, opened 2017-02-22
Updating the Oracle Linux installation documentation

Djelibeybi:
### Proposed changes

Following from #1567, this PR includes updates incorporated after receiving feedback from the Docker team.

### Related issues (optional)

#1567

### DIFF
--- a/engine/installation/linux/oracle.md
+++ b/engine/installation/linux/oracle.md
@@ -14,11 +14,10 @@ To get started with Docker on Oracle Linux, make sure you
 
 ### OS requirements
 
-To install Docker, you need the 64-bit version of Oracle Linux 6 or 7.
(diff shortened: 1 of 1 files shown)


engine/installation/linux/oracle.md
```diff
@@ -45,18 +53,18 @@ or volumes you created using the older version of Docker are preserved.
 
 You can install Docker in different ways, depending on your needs:
 
-- Most users
+- Users who do not require Docker support from Oracle
```
friism: I think this is missing a "can"

mstanleyjones: I'll fix that in follow-up. Thanks @friism

Djelibeybi: I left it in the active voice, i.e. "these users do this" as opposed to "these users could do this". But you could add "can" back into this sentence as well.

PR #1929, opened 2017-02-23
[ucp] Improve/modularize the UCP HRM docs

### LINKED ISSUE #1732: Improve UCP http routing mesh docs
We currently have documentation for the HTTP routing mesh but the documentation is mixing a lot of concepts so some users are getting confused by it.

- [ ] What is this feature, and why should you care?
- [ ] I'm an IT ops and I want to understand how this works behind the scenes
- [ ] How IT ops can enable this feature
- [ ] How a developer can deploy a simple app that is routed using HTTP. Emphasize the UI workflow but also explain the CLI flow
- [ ] How a developer can route HTTPs requests
- [ ] How a developer can implement sticky sessions
- [ ] Reference for the label syntax

---

These are the current docs we have:
- [Use hostnames to access your service](https://docs.docker.com/datacenter/ucp/2.1/guides/user/services/use-hostnames-to-access-your-service/)
- [Use domain names to access services](https://docs.docker.com/datacenter/ucp/2.1/guides/admin/configure/use-domain-names-to-access-services/)
- [Reference architecture](http://success.docker.com/KBase/Docker_Reference_Architecture%3A_Universal_Control_Plane_2.0_Service_Discovery_and_Load_Balancing)

avuserow:
### Proposed changes

This separates out the use cases into three audiences:
administrators who configure UCP and HRM, users who use the web UI, and
users who use the Docker CLI or API. The CLI reference also provides a
guide to the label syntax and the full set of features.

cc @joaofnfernandes. I'm sure there's way more improvements but want you to take a look. We should also investigate an update to the reference architecture.

### Related issues (optional)

Resolves #1732.


joaofnfernandes:
Let's coordinate a review to the reference architecture once we merge this. It will make that review easier because we know this is the source of truth and others can reuse it.

datacenter/ucp/2.1/guides/user/services/hrm-labels.md
```diff
@@ -0,0 +1,118 @@
+---
```
joaofnfernandes: When you're adding new articles, you also need to add [an entry here](https://github.com/docker/docker.github.io/blob/master/_data/toc.yaml#L1268-L1269), to make the article show in the left navigation bar of docs.docker.com

PR #1935, opened 2017-02-23
Update rolling-update.md

scjane:


scjane:
@mstanleyjones Sorry for always using the auto-generated name "Upadting xxx.md". Do you mean it will be more clear if I modify this title according to the change I make or I still use the default title but describe my change in a comment? Will it be a little bit redundant if a change is apparent enough, such as adding or deleting a period? That is to say, what kind of changes should be particularly titled as you think? Looking forward to your suggestions.

PR #1939, opened 2017-02-23
Update manage-nodes.md

scjane:


scjane:
@mstanleyjones Thank you for the changes you made and your kindly commit about them, where I learned a useful tip. I'll try doing some improvements on format and words in my future PRs while modifying tiny mistakes.

PR #1947, opened 2017-02-23
added link target for legacy system requirements location 

londoncalling:
### What changed

- added link target (id) for legacy location of system requirements for Docker for Mac and Windows so that hard-coded links in template responses to certain classes of bug report reference won't break

- updated the note in Docker for Mac and Windows home pages to x-ref specifically to system requirements in the install page, and say that the topic moved there.

- directly fixed the links on the install pages for both Mac and Windows to link to the on-page topic system requirements (that was a mistake in previous link updates I made in the docs, not the same problem as in the templates)

**Explanation:** The links breaking in the app templates is due to the fact that we restructured all the docs to split out install topics (`install.md` from the rest of the getting started (`index.md`). The above updates should catch the template links at the note. At some point, if you want to modify the link from `index.md#what-to-know-before-install` to `install.md#what-to-know-before-install`, then those links would go directly to the topic.

### Related issues

https://github.com/docker/pinata/issues/6624

### Reviewers

@ijc25 @friism @jeanlaurent @ebriney 

### DIFF
--- a/docker-for-mac/index.md
+++ b/docker-for-mac/index.md
@@ -17,8 +17,11 @@ Welcome to Docker for Mac!
 Docker is a full development platform for creating containerized apps, and
 Docker for Mac is the best way to get started with Docker on a Mac.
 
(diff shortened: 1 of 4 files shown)


londoncalling:
LGTM, merging to address this issue in published docs ASAP

PR #1950, opened 2017-02-23
tweaked note re: d4win on sample workflow

londoncalling:


londoncalling:
LGTM

PR #1956, opened 2017-02-24
Expand the docs for configuring UCP and DTR for content trust

### LINKED ISSUE #1730: Improve content trust docs
We have several documentation pages that are mixing everything together, and users are having a hard time understanding content trust in DDC. We should have different pages explaining

- [x] What is content trust, and why should you care. Diagrams would help a lot
- [x] How IT ops can configure UCP to only run trusted images
- [x] How IT ops can configure which teams/users are trusted by UCP
- [x] How IT ops/repo owner can initialize trust for a DTR repository
- [x] How IT ops/repo owner can delegate image signing on a DTR repository
- [x] How DTR users should set their CLI to sign images that UCP can trust. Diagrams would help a lot

- [ ] Common pitfalls in the current docs. Check #1686

---

Here's links to the docs we have available today:
- [Only allow running signed images](https://docs.docker.com/datacenter/ucp/2.1/guides/admin/configure/only-allow-running-signed-images/)
- [Use trusted images for continuous integration](https://docs.docker.com/datacenter/ucp/2.1/guides/admin/configure/use-trusted-images-for-ci/)
- [Notary docs](https://docs.docker.com/notary/getting_started/)
- [Matt Bentley's content trust guide](https://docker.atlassian.net/wiki/pages/viewpage.action?pageId=113673243)

mstanleyjones:
### Proposed changes

Fixes #1730 

- Adds conceptual info
- Splits tasks up into admin and end user tasks
- Adds info about configuring the client for macOS and Windows
- Adds screenshots


datacenter/ucp/2.1/guides/admin/configure/content-trust/client_configuration.md
```diff
@@ -0,0 +1,255 @@
...
+
+    ```none
+    x509: certificate signed by unknown authority
+    ```
+
+    This procedure is different if you are on a Linux or macOS client:
+
+    - **Linux**:
+
+      1.  Download the certificate add it to a subdirectory of the
```
joaofnfernandes: Yup, we should not be storing CAs in the `/etc/docker/certs.d`, instead they should configure the OS for this. This has the added benefit that any other client tools will trust the CA.
E.g. If IT decides to rollout DDC with their own PKI, they can configure employee laptops to trust the CA once, and Docker, compose, curl, Chrome, Firefox and others will trust that CA.

We already have docs on [how to achieve this](https://docs.docker.com/datacenter/dtr/2.2/guides/user/access-dtr/)

datacenter/ucp/2.1/guides/admin/configure/content-trust/admin_tasks.md
```diff
@@ -0,0 +1,387 @@
...
+
+    Click the **REQUIRE SIGNATURE FROM ALL OF THESE TEAMS** field and choose
+    one or more teams. DDC will consider an image to be trustworthy only if it
+    is signed by a member of _every team_ you select. The requirement can be
+    fulfilled by a single user who is a member of all the teams, or by a signer
+    in each team.
+
+    > **Note**: If you don't specify any team (by leaving the field blank), an
+    > image will be trusted as long as it is signed by any UCP user whose keys
+    > are [configured in the Notary client](#set-up-the-docker-notary-client).
```
cyli: Non-blocking: I think this may make more sense as just "any UCP user", because it sounds like in order to be trusted, they have to be a UCP user *and* also something else.  I get what you are saying, though - in order for them to sign at all, their keys have to first be configured to sign in notary, so effectively that is true.  

It mostly seems confusing because the contrast with the above (where configuring keys is not mentioned at all) - like if you don't specify teams, you have to configure the keys.  But if you specify teams, you do not.

datacenter/ucp/2.1/guides/admin/configure/content-trust/admin_tasks.md
```diff
@@ -0,0 +1,387 @@
...
+
+3.  Click **Update** to apply the changes.
+
+UCP is now configured to only allow use of signed images, but you don't have the
+ability to sign images yet. Next,
+[set up the Docker Notary CLI client](#set-up-the-docker-notary-client).
+
+## Set up the Docker Notary CLI client
+
+After [configuring UCP](#configure-ucp), you need to specify which Docker images
```
cyli: Should this be "you need to specify who sign specific Docker images"?  "specify which Docker images can be trusted" sounds sort of like whitelisting a list of images, when it's setting up the trust data for specific images.

datacenter/ucp/2.1/guides/admin/configure/content-trust/admin_tasks.md
```diff
@@ -0,0 +1,387 @@
...
+can be trusted, using the Docker Notary server that is built into Docker Trusted
+Registry (DTR). The following procedure configures the Notary server to store
```
cyli: Maybe "The following procedure sets signed metadata on the Notary server about the Docker images you trust"?   

"Configuring the Notary server to store..." to me suggests something like configuring the database back end, or something.

datacenter/ucp/2.1/guides/admin/configure/content-trust/admin_tasks.md
```diff
@@ -0,0 +1,387 @@
...
+2.  Configure the Notary client to communicate with the DTR server and store
+    its metadata in the correct location. You can either use a
+    [Notary configuration file](/notary/reference/client-config.md) or manually
+    specify the following flags when you run the `notary` command.
+
+    |Flag               |Purpose                                               |
+    |-------------------|------------------------------------------------------|
+    | `-s <dtr_url>`    | The hostname or IP address of the DTR server         |
+    | `-d <trust_directory>` | The path to the local directory where trust metadata will be stored |
+    | `--tlscacert <dtr_ca.pem>` | The path to the trust certificate for DTR. Only required if your DTR registry is not using certificates signed by a globally trusted certificate authority, such as self-signed certificates. Download the trust certificate from `https://<dtr_url>/ca` either from your browser or using `curl` or `wget`.|
```
cyli: Non-blocking:  perhaps "CA certificate" as opposed to "trust certificate", just to distinguish it from the rest of the trust data when discussing notary?

datacenter/ucp/2.1/guides/admin/configure/content-trust/admin_tasks.md
```diff
@@ -0,0 +1,387 @@
...
+      authority for DTR. To get the certificate, download `https://<dtr_url>/ca`
+      from your browser or using `curl` or `wget`. This certificate is
+      different from the UCP trust certificate in the UCP client bundle.
+
+    - `fatal: <dtr_url> does not have trust data for ddc-staging-dtr.qa.aws.dckr.io/engineering/redis`:
+      The repository has not yet been initialized and you need to run
+      `notary init`. Continue to step 2.
+
+    - `No targets present in this repository.`: The repository has been
+      initialized, but contains no signed images. You do not need to do step 2.
```
cyli: Maybe in this and the next item, rather than saying that step 2 can be skipped (because step 3 can also be skipped) the user can be linked to the "Delegate image signing" step instead?

datacenter/ucp/2.1/guides/admin/configure/content-trust/admin_tasks.md
```diff
@@ -0,0 +1,387 @@
...
+Continue to [delegate image signing](#delegate-image-signing) to give the
+appropriate users the ability to sign images.
+
+You can also
+[learn more about the keys used by Notary](/engine/security/trust/trust_key_mng.md).
+
+
+## Delegate image signing
+
+The administrator who manages Docker Trusted Registry is often not part of the
```
cyli: Even if they are a member of a group allowed to sign images, they probably still have to add their UCP bundle key to a delegation - the targets key or root key is not recognized by UCP as a signing user.

datacenter/ucp/2.1/guides/admin/configure/content-trust/admin_tasks.md
```diff
@@ -0,0 +1,387 @@
...
+
+    The response may be one of the following:
+
+    - `fatal: client is offline`: Either the repository server can't be reached,
+      or DTR is using certificates which are not signed by a globally trusted
+      certificate authority, such as self-signed certificates. Run `notary list`
+      again, adding the `--tlscacert` flag, with the path to the certificate
+      authority for DTR. To get the certificate, download `https://<dtr_url>/ca`
+      from your browser or using `curl` or `wget`. This certificate is
+      different from the UCP trust certificate in the UCP client bundle.
```
cyli: Non-blocking: maybe just "certificates" or "CA certificate" in the UCP bundle?

datacenter/ucp/2.1/guides/admin/configure/content-trust/admin_tasks.md
```diff
@@ -0,0 +1,387 @@
...
+    The following command adds `user1` to the `targets/engineering` delegation:
+
+    ```bash
+    $ notary delegation add -p <GUN> targets/engineering --all-paths user1.pem
+    ```
+
+    You will be prompted for your DTR credentials followed by the passphrase
+    for the `targets` key.
+
+5.  Securely remove the `.pem` files of the users you added delegations to. If
```
cyli: These `.pem` files are certs containing public keys, and can't alone be used to sign - they're certificates that specify the username and the fingerprint of the actual signing key.

cyli: The users should definitely not send admin their `key.pem` file, just the `cert.pem`, which is not sensitive.  If they did send the admin the `key.pem` file by accident, yes those should be securely removed.  :)

datacenter/ucp/2.1/guides/admin/configure/content-trust/client_configuration.md
```diff
@@ -0,0 +1,255 @@
...
+[delegate signing to them](#delegate-image-signing).
+
+Import the private key associated with the user certificate. You must specify
+the trust directory `~/.docker/private`. If the `~/.docker/private` directory
+does not yet exist, Notary will create it.
+
+- **Linux or macOS**:
+
+  ```bash
+  $ notary -d ~/.docker/trust key import /path/to/key.pem
```
cyli: It should be `~/.docker/trust` above.

datacenter/ucp/2.1/guides/admin/configure/content-trust/client_configuration.md
```diff
@@ -0,0 +1,255 @@
...
+      prompt.
+
+    - To set the environmment variable for just a single command, add it before
+      the command:
+
+      ```bash
+      $ DOCKER_CONTENT_TRUST=1 docker pull...
+      ```
+
+    - To set the environment variable permanently, edit the `~/.profile` file
```
cyli: TIL apparently so:  https://en.wikipedia.org/wiki/Unix_shell#Configuration_files

vikstrous: Heh. Someone needs to add fish to that table.

PR #1961, opened 2017-02-24
Update scale-service.md

scjane:
Deleted "you can" because it shoud be "you must" in fact, so use imperative instead.


johndmulhausen:
Always good to lose some uses of passive voice. Thanks, @scjane :)

PR #1963, opened 2017-02-24
Update btrfs-driver.md

scjane:
It seems that some word is missing. Added a "for".


johndmulhausen:
Thanks for doing a few things in one PR, @scjane -- much appreciated :)

PR #1965, opened 2017-02-24
Update index.md

scjane:
Is a "delete" missing here? Or, shall we change this to "connect containers to and disconnect from it"?


johndmulhausen:
Add/delete works better than connect/disconnect, I think. Thanks, @scjane :)

PR #1966, opened 2017-02-24
Update trust_sandbox.md

scjane:
Added a colon (:). But I'm not sure whether a period (.) is more appropriate here.


johndmulhausen:
Thanks, @scjane :)

scjane:
@johndmulhausen Actually, I'm not sure whether a period (.) is more appropriate here.

PR #1970, opened 2017-02-24
split out redirect note to system requirements for both mac and windows

londoncalling:
### Proposed changes

Updated notes that act as manual redirects to system requirements topic, which moved to new `install.md` pages in both Docker for Mac and Docker for Windows.

Split out references to install in general vs. system requirements to make sure it catches the attention of users who already have the apps installed, but are looking for more details on hardware/software requirements.

### Related

#1947 

### Reviewers

@ijc25 


londoncalling:
@ijc25 I'm going to go ahead and merge this because I think it's better, and closer to what you want. We can always tweak it again if needed. Thanks!

PR #1976, opened 2017-02-25
Fix sentence.

mike706574:
### Proposed changes

Fix a broken sentence - should be self-explanatory, but I'm not sure how well my change expresses the original intended meaning.


johndmulhausen:
Thanks, @mike706574 :) LGTM

PR #1982, opened 2017-02-26
feedback on PowerShell env setup, clarified env vs. ssh

londoncalling:
### Changes

- explicitly added PowerShell `env` command, and also linked to the command reference

- clarified that using `docker-machine env` vs. `docker-machine ssh` are two ways of getting access to the machines, and you can use one or the other method (or both) to work through the tutorial steps

### Reviewers

@JimGalasyn  I'll probably merge this before you see it, but wanted to tag you on it. Feel free to submit more comments if it still isn't accurate or needs improvement. Thanks for the feedback.


londoncalling:
LGTM

PR #1994, opened 2017-02-27
Update meetups.md

scjane:
Modify the format of the last item in the list and add missing period (.).


johndmulhausen:
Thanks, @scjane :)

PR #2007, opened 2017-02-28
update deprecation policy

vieux:
ping @friism @thaJeztah

### DIFF
--- a/engine/index.md
+++ b/engine/index.md
@@ -98,8 +98,8 @@ on the separate [Release Notes page](/release-notes)
 As changes are made to Docker there may be times when existing features
 will need to be removed or replaced with newer features. Before an existing
 feature is removed it will be labeled as "deprecated" within the documentation
-and will remain in Docker for, usually, at least 3 releases. After that time
-it may be removed.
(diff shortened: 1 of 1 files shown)


engine/index.md
```diff
@@ -98,8 +98,8 @@ on the separate [Release Notes page](/release-notes)
 As changes are made to Docker there may be times when existing features
 will need to be removed or replaced with newer features. Before an existing
 feature is removed it will be labeled as "deprecated" within the documentation
-and will remain in Docker for, usually, at least 3 releases. After that time
-it may be removed.
+and will remain in Docker for, usually, at least 3 stable releases (roughly 9 months).
```
mstanleyjones: Can we remove `, usually,`? What are the circumstances in which it would be shorter than 3 months?

vieux: @mstanleyjones never shorter, but could be longer.

mstanleyjones: Then "at least" covers it and we don't need "usually".

vieux: I just pushed an update

PR #2008, opened 2017-02-28
"old.example.com.com" --> "old.example.com"

clemenko:
Fixing : "old.example.com.com" --> "old.example.com"


johndmulhausen:
Fair enough, thanks @clemenko :)

PR #2016, opened 2017-02-28
move system auth description

xolotl:
The requirement to auth to one's local system was in step 1, install, when in practice, the auth comes during step 2, first launch.


londoncalling:
LGTM

PR #2019, opened 2017-02-28
Adds documentation about the `--ignore-override` option

dennisgove:
For PR https://github.com/docker/compose/pull/4543

When option `--ignore-override` is provided the `docker-compose.override.yml`
file, if it exists, will not be considered for any command. This option
has no effect when the `-f` option is used as that explicitly lists the
set of files to be considered.

Signed-off-by: Dennis Gove <dpgove@gmail.com>

### DIFF
--- a/compose/extends.md
+++ b/compose/extends.md
@@ -26,13 +26,17 @@ services.
 
 If a service is defined in both files Compose merges the configurations using
 the rules described in [Adding and overriding
-configuration](extends.md#adding-and-overriding-configuration).
+configuration](extends.md#adding-and-overriding-configuration). You can turn off
+the automatic use of `docker-compose.override.yml` with command-line option
+`--ignore-override`.
 
 To use multiple override files, or an override file with a different name, you
 can use the `-f` option to specify the list of files. Compose merges files in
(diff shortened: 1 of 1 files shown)


dennisgove:
Thanks! I just want to be totally upfront here, the PR this documentation change is for has not yet been accepted - https://github.com/docker/compose/pull/4543.

mstanleyjones:
Ugh. OK, keep me posted. We can revert the commit if necessary.

dennisgove:
So far there haven't been any additional comments from accepters in https://github.com/docker/compose/pull/4543

mstanleyjones:
@shin- any update on this?

shin-:
As I pointed out in https://github.com/docker/compose/pull/4543#issuecomment-283456247, this feature already exists. The argument for it ("There's a way to do this already, but it's too verbose so let's add another flag that does exactly the same thing for an equivalent amount of verbosity") seems tenuous at best. I'd recommend reverting this as it's unlikely this will make it into the project.

dennisgove:
Thank you @shin-. 

@mstanleyjones, please revert this commit. I'm very sorry for the inconvenience.

PR #2022, opened 2017-02-28
DTR 2.2.2

joaofnfernandes:


joaofnfernandes:
The tests are failing because S3 is down, so you can't pull images from Hub in the mean time.

PR #2030, opened 2017-03-01
Update work_issue.md

scjane:
Change "re-start" to "restart". Though not included in "Prefered usages" in the documentation guide, I think "restart" is better and used more frequently. Besides, some other docs here, such as "Keep containers alive during daemon downtime" of "Admin Guide", also use "restart".


johndmulhausen:
Thanks, @scjane :)

PR #2032, opened 2017-03-01
There is no OpenSuSE and there never was

ralflang:
though we had SuSE and S.u.S.E.


johndmulhausen:
Not to be pedantic but, upon Googling: isn't it openSUSE?

PR #2033, opened 2017-03-01
List other keyservers to try for cs-engine install

cpuguy83:
Sometimes ha.pool.sks-keyservers.net goes down, so let's provide some
other keyservers to try in such cases.


cpuguy83:
Well, I've tried formatting the notes a couple of ways but none are working as expected.
Please let me know how you'd like to have them formatted.

PR #2044, opened 2017-03-02
Update work_issue.md

scjane:
I think this change is not a must but an optimization. Please review it. Thanks.


johndmulhausen:
Thanks, @scjane :)

PR #2058, opened 2017-03-02
Fix Docker CE repo location

mstanleyjones:
@cpuguy83 PTAL


mstanleyjones:
Not actually a problem.

PR #2061, opened 2017-03-02
Fix alignment with buttons and headings

mstanleyjones:
@jsouth why do I need to monkey with the headings like this? We might need to revert.

@mavenugo PTAL at Netlify when it finishes to confirm that this is OK for now.


jsouth:
@mstanleyjones  - Ah i see. The markup is missing the base "button" class prior to the specific dark button.

This is fine for now, but add will add this to cleanup list  :)

PR #2066, opened 2017-03-02
Update step_four.md

canzanos:
### Proposed changes

When running on Windows if you do notepad Dockerfile (at least using Windows 10) the file will be saved as Dockerfile.txt.  By using notepad Dockerfile. when saving the file it will not be saved with the .txt extension.  Later on in Step 2: Build an image from your Dockerfile the command
$ docker build -t docker-whale .

will fail if Dockerfile has the .txt extension

### DIFF
--- a/engine/getstarted/step_four.md
+++ b/engine/getstarted/step_four.md
@@ -61,7 +61,7 @@ creating an image which runs on Linux.
     **Windows**:
 
     ```none
-    C:\> notepad Dockerfile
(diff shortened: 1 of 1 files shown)


johndmulhausen:
Thanks, @canzanos -- we love it when our docs get more Microsoft-friendly, and we aren't well staffed to test these use cases. If you find anything else, keep the PRs coming! :)

PR #2070, opened 2017-03-02
Put temp styles in the right place

mstanleyjones:
@johndmulhausen PTAL


johndmulhausen:
Thanks, @mstanleyjones :)

PR #2076, opened 2017-03-02
Lots of formatting fixes to do with buttons

mstanleyjones:
- Change all the dark blue buttons to outline buttons to match the current theme

- The outline buttons need some padding at the bottom or they will crush into the next header

- A single button doesn't need to be inside a table so I removed tables where unnecessary

- A few tables were relying on `<br />` within a `<td>`. I just added another `<tr>` and put the second line's content there.

- Updated the examples of buttons in `test.md`.

This doesn't address a few things for follow-up work:

- We really need to figure out how to use cards for horizontal content so we can avoid having two-column 50% tables.
- I didn't change all of the HTML links into Markdown format, though I did change some.
- Probably in mobile, this kind of button should change from `inline-block` to `inline` and stack vertically.

@FrenchBen @nathanleclaire PTAL at D4A
@londoncalling PTAL at D4M, D4W, Toolbox
@johndmulhausen @jsouth PTAL overall

### DIFF
--- a/docker-for-aws/index.md
+++ b/docker-for-aws/index.md
@@ -12,7 +12,7 @@ redirect_from:
 ## Docker Enterprise Edition (EE) for AWS
 This deployment is fully baked and tested, and comes with the latest Enterprise Edition version of Docker. <br/>This release is maintained and receives <strong>security and critical bugfixes for one year</strong>.
 
-<a class="button darkblue-btn" href="https://store.docker.com/editions/enterprise/docker-ee-aws?tab=description" target="_blank">Deploy Docker Enterprise Edition (EE) for AWS</a>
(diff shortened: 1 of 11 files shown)


docker-for-aws/index.md
```diff
@@ -40,13 +40,10 @@ using CloudFormation. For more about stable and edge channels, see the
   </td>
   <td width="50%">
   {{aws_blue_edge}}
+  {{aws_blue_vpc_edge}}
   </td>
   </tr>
   <tr valign="top">
```
FrenchBen: this `<tr>` and following `</td></tr>`

nathanleclaire:
Don't know that much about the CSS in docs but LGTM

docker-for-aws/index.md
mstanleyjones: The `<tr>` has two `<td>` in it just like it's supposed to. I think it's right. Both buttons are in one `<td>`.

johndmulhausen:
LGTM as well :)

mstanleyjones:
Merging for now. 👀

PR #2093, opened 2017-03-03
3 typo changes

debianivanov:
```bash
    sudo nano /etc/docker/daemon.json
    ```
Here a $ sign is missing and the command in the web site appears as "bash sudo nano /etc/docker/daemon.json", which is not correct. It shoud look like: "$ sudo nano /etc/docker/daemon.json".

"which is 2375 if you do not use TLS, and 2376 if you do."
On all other lines the ports are in backquotes, so I have added backquotes here also.

"In this case, steps 3 and 4."
Missing verb like "repeat", I have added it.

### DIFF
--- a/engine/installation/linux/linux-postinstall.md
+++ b/engine/installation/linux/linux-postinstall.md
@@ -202,7 +202,7 @@ at `/etc/docker/daemon.json`.
     configuration.
 
     ```bash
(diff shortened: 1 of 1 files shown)


johndmulhausen:
Thanks, @debianivanov :)

PR #2095, opened 2017-03-03
Update step_six.md

travisjs:
I need to add the -f option to remove image, otherwise: Error response from daemon: conflict: unable to delete e867bc17a0bb (must be forced) - image is referenced in multiple repositories


johndmulhausen:
@travisjs Looks like it should be `image rm` and the option needs to go before the hash, per https://docs.docker.com/engine/reference/commandline/image_rm/

So I patched that in real quick. 

Thanks for the PR!

PR #2099, opened 2017-03-03
Update docker-machine create command

uncompiled:
If `docker-machine --create` is invoked, the user will receive an error: "Incorrect Usage."  This PR changes the example command to `docker-machine create`.


johndmulhausen:
Thanks @uncompiled :)

PR #2100, opened 2017-03-04
clean up code and images

londoncalling:
#### What changed

- Cleaned up images, made terminal window background pale color (considered making these plain code blocks in markdown, but there is no way to highlight a snip-it that way, and I also don't think people like looking at all that code when it isn't necessary. For this case, I think images work better until we get more control over highlighting or styles within code blocks.

- Cleaned up code blocks by specifying "none" for markdown fences. This got rid of a lot of whacky colors in the code samples.

- Simplified stack file example and instructions.

- Removed platform specific commands and suggestions (e.g., `curl`, `vi`, `cat` that won't work on Windows). I either need to find equivalents, or just not have these in there. For now, I've removed them.

### FYI 

@johndmulhausen @mstanleyjones 


londoncalling:
LGTM

PR #2106, opened 2017-03-04
mention check-config.sh script

savayg:
### Proposed changes

Mention the check-config.sh script as I randomly ran through it, it could have saved me a lot of time.

I was trying to figure out why my ports weren't publishing in my swarm although everything seemed correct. It was due to missing CONFIG_IP_VS in the kernel of my host provider.

### DIFF
--- a/engine/installation/linux/linux-postinstall.md
+++ b/engine/installation/linux/linux-postinstall.md
@@ -93,6 +93,21 @@ your host's Linux distribution and available kernel drivers.
 
 ## Troubleshooting
 
+### Kernel compatibility
+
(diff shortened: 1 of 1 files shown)


thaJeztah:
This script applies to any Linux version, it checks kernel configuration and available options. It should probably be added to a troubleshooting section.

Also, I'd prevent directly piping the script to `sh`, as it's not considered best practice

mstanleyjones:
Thanks for your work. I put this info into the troubleshooting section for Linux instead and expanded upon it. I pushed directly into the PR.

PR #2110, opened 2017-03-04
FIX Duplicate "are"

debianivanov:
"on Mac and Windows are are as follows"
Removed the duplicated "are".


johndmulhausen:
Thanks, @debianivanov :)

PR #2111, opened 2017-03-04
Update consistency of <manager-IP>:<port>

debianivanov:
On the "Sample app overview" page all blocks of IP address and port are shown as <manager-IP>:5000 or <manager-IP>:5001.
On this page "Try it out and vote" I see some inconsistency of these forms and fixed them, because I think the style format must be in the form "<manager-IP>:5000", e.g., colon after character ">".


johndmulhausen:
Thanks, @debianivanov :)

PR #2117, opened 2017-03-05
Update consistency of <manager-IP>:<port>

debianivanov:
This is the same issue as the page "Try it out and vote". I have updated the consistency of format to be the same through all pages. (and I hope I do not overdo these small things, but I just like the ease of proposing changes to the documentation and I hope that it is getting better this way)


johndmulhausen:
Much thanks, @debianivanov !

PR #2119, opened 2017-03-06
Update next_steps.md

scjane:
Is "ID" the right one? It appears that "id", "ID", or "Id" is used here and there.


johndmulhausen:
ID is right, thanks @scjane :)

scjane:
@johndmulhausen OK. I get it now. Thanks :)

PR #2121, opened 2017-03-06
Update hyper-v.md

endlos:
### Proposed changes

I specified that it is not necessary to create a new external switch if there is one already created. Otherwise the reader would follow the steps unnecessarily.


johndmulhausen:
- Use a blockquote instead of italics
- Precede this with **Note**:
- Append "(Optional)" to the heading so people know they might be able to skip the whole seciton

endlos:
@johndmulhausen done :-)

johndmulhausen:
Beautiful! Thanks a ton @endlos :)

PR #2162, opened 2017-03-07
Update test.md

scjane:
Is "as" missing?


johndmulhausen:
Indeed, thanks, @scjane :)

PR #2164, opened 2017-03-07
docker swarm does not respect depends_on

vHanda:
### Proposed changes

Swarmkit does not seem to have the concept of starting services in a particular order.

Tested with docker 1.13.1. Also, `docker-compose bundle` warns us about depends_on being unsupported.

It should be obvious from the documentation.

### DIFF
--- a/compose/compose-file/index.md
+++ b/compose/compose-file/index.md
@@ -348,6 +348,10 @@ Simple example:
 > for a service to be ready, see [Controlling startup order](/compose/startup-order.md)
 > for more on this problem and strategies for solving it.
 
+> **Note:** This option is ignored when
+> [deploying a stack in swarm mode](/engine/reference/commandline/stack_deploy.md)
+> with a (version 3) Compose file.
+
 A healthcheck indicates you want a dependency to wait
 for another container to be "healthy" (i.e. its healthcheck advertises a
 successful state) before starting.


mstanleyjones:
@aaronlehmann @nishanttotla PTAL

aaronlehmann:
ping @dnephin

dnephin:
The title of this issue is a bit misleading. I believe it should say "Document that `docker stack deploy` does not support `depends_on`"

I guess a note [here](https://docs.docker.com/compose/compose-file/#dependson) wouldn't hurt.

There's some background in https://github.com/docker/docker/issues/31333. The idea is that on swarm mode services are restarted until they are stable, so the order they are started shouldn't matter. There's also no way to say "has started" because `service create` is async.

arvenil:
@dnephin 

Definitely docs should state ```Note: These options are ignored when deploying a stack in swarm mode with a (version 3) Compose file.```
Also `docker stack deploy` doesn't say anything that those options are ignored in contrast to raising warning for `restart: always`.

While it makes sense there is no such concept as deployment order with swarm, availability of those options in doc suggested I could use it to improve starting issues (e.g. some services die because other are not yet available so I get bunch containers that are restarted at initialization couple of times before they finally connect to each other).

Anyway, would be nice to update docs and maybe rise warning for ignored options as I lost plenty of time on this and on https://github.com/docker/docker/issues/30404

dnephin:
I've included this fix in #2278 because they are going to conflict.

PR #2168, opened 2017-03-07
Use shallow clone and shallow checkout

mstanleyjones:
This does an initial shallow clone of the repo (just the tip of `master`) and then does a shallow fetch of each branch before checking it out, so we just get the tip of each of the archive branches.

This doesn't reduce the size of the image but reduces the time to do the `git clone` and `git fetch`es. It currently saves about 2 minutes locally and will save more as the repo grows and the number of branches in the archives increase.

Before:

```
real    10m7.755s
user    0m1.169s
sys    0m2.311s
```

After:

```
real    8m11.720s
user    0m1.115s
sys    0m2.141s
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -30,9 +30,12 @@ COPY nginx.conf /etc/nginx/nginx.conf
 
 ENV VERSIONS="v1.4 v1.5 v1.6 v1.7 v1.8 v1.9 v1.10 v1.11 v1.12 v1.13"
 
-RUN git clone https://www.github.com/docker/docker.github.io archive_source; \
+## Use shallow clone and shallow check-outs to only get the tip of each branch
+
+RUN git clone --depth 1 https://www.github.com/docker/docker.github.io archive_source; \
  for VER in $VERSIONS; do \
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -30,9 +30,12 @@ COPY nginx.conf /etc/nginx/nginx.conf
 
 ENV VERSIONS="v1.4 v1.5 v1.6 v1.7 v1.8 v1.9 v1.10 v1.11 v1.12 v1.13"
 
-RUN git clone https://www.github.com/docker/docker.github.io archive_source; \
+## Use shallow clone and shallow check-outs to only get the tip of each branch
+
+RUN git clone --depth 1 https://www.github.com/docker/docker.github.io archive_source; \
  for VER in $VERSIONS; do \
-		git --git-dir=./archive_source/.git --work-tree=./archive_source checkout ${VER} \
+    git --git-dir=./archive_source/.git --work-tree=./archive_source fetch origin ${VER}:${VER} --depth 1 \
```
thaJeztah: looks like you swapped tabs for spaces here; indentation is off 😇

PR #2176, opened 2017-03-07
Trigger build of docs-base after successful build of archive branch

mstanleyjones:
Bring v1.12 up to date with v1.4..v1.11. Needed to make a couple tweaks since v1.12 actually builds from source.


mstanleyjones:
Failure is nothing to do with this PR. @aduermael

PR #2181, opened 2017-03-07
Update step_six.md

iHrushikesh:


johndmulhausen:
We are actually encouraging the more explicit `docker image rm` these days. Thanks for the PR, though. :)

PR #2196, opened 2017-03-08
Fix docker ps

ry4nz:
### Proposed changes

Current documentation at https://docs.docker.com/engine/reference/commandline/ps/ looks incorrect.  This patch fixes it.
screen shot 2017-03-08 at 10 31 00 am


johndmulhausen:
Thank you, @ry4nz ! Much better, obviously :)

PR #2201, opened 2017-03-08
updates to intro, moved services descriptions  under config file topic

londoncalling:
- topic reorganization on intro page so that descriptions of services and keys makes more sense in context
- simplify, clarify some concepts


londoncalling:
LGTM

PR #2203, opened 2017-03-08
Missing ssh tunnel option

lorenzo-pizzari:
### Proposed changes
I've tried to connect with ssh tunnel and i found that the port option is missing.
Added port option to doc.
If no port option is provided the tunnel don't work properly

### DIFF
--- a/docker-for-azure/deploy.md
+++ b/docker-for-azure/deploy.md
@@ -43,7 +43,7 @@ Once you are logged into the container you can run Docker commands on the swarm:
 
 You can also tunnel the Docker socket over SSH to remotely run commands on the cluster (requires [OpenSSH 6.7](https://lwn.net/Articles/609321/) or later):
 
-    $ ssh -NL localhost:2374:/var/run/docker.sock docker@<ssh-host> &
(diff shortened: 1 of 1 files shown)


nathanleclaire:
👍 

BTW, just noticed this -- the `&` at the end of the command is not needed if the `-f` option for SSH ("fork" into the background) is used.

Also: Obligatory mention to say "don't expose Docker API unencrypted over the network", even on `localhost`.

nathanleclaire:
Well, in Azure we use pubkey, so path to SSH key is not _strictly_ needed like it might be with AWS case...

lorenzo-pizzari:
> BTW, just noticed this -- the & at the end of the command is not needed if the -f option for SSH ("fork" into the background) is used.
You're right! I've just pushed in this way.
> Well, in Azure we use pubkey, so path to SSH key is not strictly needed like it might be with AWS case...
Yes, but in the previous command there is the SSH key path, also the Azure SSH key can be different from the "system default" (~/.ssh/id_rsa)

nathanleclaire:
> Yes, but in the previous command there is the SSH key path, also the Azure SSH key can be different from the "system default" (~/.ssh/id_rsa)

That's true !

PR #2238, opened 2017-03-09
V1.4 archive admonition (JS solution)

mstanleyjones:
Adds an admonition to the top of every page stating that this is an archive, and giving a drop-down menu with a choice of other docs versions you can use instead.

This relies on an internet connection and fails gracefully if there isn't one.

It also relies on a file that we host in the latest version of the docs. That file is
built automatically as part of the `master` build. If that file doesn't exist,
a Javascript error will be logged but the archived page will load normally.

This is for 1.4. I suspect that we may need to tweak it slightly for other archive versions
as the version of Bootstrap changes.

The build that puts the JSON file up in its permanent location is still in flight, so it may be a few minutes before this works.

Since we don't have Netlify, you can check out this branch and do `jekyll serve` locally.


mstanleyjones:
- Updated styling
- Replaced old Google Analytics code with Segment
- Removed old combo box

PR #2240, opened 2017-03-09
Remove logdriver details for NATS

cpuguy83:
The NATS driver was never integrated into Docker and should not be in
the docs.


johndmulhausen:
Thanks, @cpuguy83!

PR #2247, opened 2017-03-10
Update image-scan.md

scjane:
Is "and" necessary here, or can be omitted?


johndmulhausen:
I think the "and" helps it read. The period is appreciated, too. Thanks, @scjane :)

PR #2253, opened 2017-03-10
Update tail docs to match built-in docs

ryansanford:
Default value for tail helps me remember it's not a switch, but requires a value. This change makes it match the built-in docs.


johndmulhausen:
Much appreciated, @ryansanford !

PR #2260, opened 2017-03-10
Update install.md to fix image render

MWransky:
## Bug
[Docker for Mac](https://docs.docker.com/docker-for-mac/install/#install-and-run-docker-for-mac) `install.md` was not displaying an image properly. Instead it showed the improperly formatted Markdown.

screen shot 2017-03-10 at 2 26 01 pm

## Fix
Simple edit to complete the Markdown to render the image properly.


MWransky:
Closing because issue already resolved.

PR #2264, opened 2017-03-10
Go code: ImagePull: repository name must be canonical

netgusto:
### Proposed changes

Corrected documentation because example Go code results in panic.

If name given to imagePull is not canonical (see here for a description of a "canonical" name : https://github.com/docker/distribution/blob/245ca46/reference/normalize.go#L19-L23 ), the Go example panics : `repository name must be canonical`

### DIFF
--- a/engine/api/index.md
+++ b/engine/api/index.md
@@ -56,7 +56,7 @@ As an example, the `docker run` command can be easily implemented in various pro
   		panic(err)
   	}
 
-  	_, err = cli.ImagePull(ctx, "alpine", types.ImagePullOptions{})
+  	_, err = cli.ImagePull(ctx, "docker.io/library/alpine", types.ImagePullOptions{})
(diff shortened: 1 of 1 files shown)


joaofnfernandes:
Were you able to run the code sample successfully after your change?

I'm trying to reproduce this with the original sample and I'm getting a `panic: Error: No such image: alpine` when creating the container, which is strange because go fails to pull alpine and doesn't panic, but does so when trying to create the container.

mstanleyjones:
Ping, what's the status on this?

joaofnfernandes:
I think we can rebase and merge it. This change makes things less ambiguous and potentially fixes a problem, so we should take it in.

mstanleyjones:
I don't have enough background to resolve the conflict, so I will leave that with y'all.

Issue #1129, opened 2017-01-14
docker.github.io/machine/reference/scp.md is very sparse


zfrenchee:
### Problem description

This page could use a few more examples. My specific issue is here: 

http://stackoverflow.com/questions/41651212/how-to-move-a-local-volume-onto-a-remote-docker-machine

And the docs don't answer my question, which I feel like they should, since it's a narrow question about `docker-machine scp` and a common use case I suspect. 

### Problem location

- I couldn't find the information I wanted. I expected to find it near the following URL: https://docs.docker.com/machine/reference/scp/

### Suggestions for a fix

I think a couple more examples could go a long way to improving this page. I'm particularly interested in my use case but I imagine there are other common ones as well.

londoncalling:
@fresnizky thanks for your feedback on this.

I just added some detail to `docker-machine scp` reference around your example. I know it's an issue specific to your use case, but I think the understanding the idea of copying to remote systems for deployment is useful in a more general sense. I contextualized your use case as a sort of second example, beyond the use of `scp` and more about file paths on remote systems. We are somewhat de-emphasizing `docker-machine` going forward but if you have ideas for more pointers that can help those docs, please do send, and we'll have a look at them.

zfrenchee:
@londoncalling no worries I solved these issues long long ago. =) Thanks for patching up the docs though.

PR #2265, opened 2017-03-10
fleshed out Machine scp command description, fixed Compose file notes

londoncalling:
### What's changed

- improved notes and some links in Compose files

- added examples to Machine `docker-machine scp` CLI reference topic per user feedback

### Related issues

https://github.com/docker/docker.github.io/issues/1129

### Reviewers

@zfrenchee @mstanleyjones @nathanleclaire 

### DIFF
--- a/machine/reference/scp.md
+++ b/machine/reference/scp.md
@@ -10,6 +10,8 @@ machine to your local host using `scp`.
 The notation is `machinename:/path/to/files` for the arguments; in the host
 machine's case, you don't have to specify the name, just the path.
 
+## Example
+
 Consider the following example:
 
 ```none
@@ -27,5 +29,43 @@ A file created remotely!
 Just like how `scp` has a `-r` flag for copying files recursively,
 `docker-machine` has a `-r` flag for this feature.
 
-In the case of transferring files from machine to machine, they go through the
-local host's filesystem first (using `scp`'s `-3` flag).
+In the case of transferring files from machine to machine,
+they go through the local host's filesystem first (using `scp`'s `-3` flag).
+
+## Specifying file paths for remote deployments
+
+When you copy files to a remote server with `docker-machine scp` for app
+deployment, make sure `docker-compose` and the Docker daemon know how to find
+them. You can specify absolute paths, e.g. `/home/myuser/workspace` in a
+[Compose file](/compose/compose-file/index.md), which will be mounted into the
+container at `/workspace`, from the absolute path on the remote host where the
+Docker daemon is running. Local client paths (e.g., on your laptop) will not
+work for daemons running on a remote machine, so avoid using relative paths.
+
+For example, imagine you want to transfer your local directory
+`/Users/londoncalling/webapp` to a remote machine and bind mount it into a
+container on the remote host. (We'll suppose the remote user is `ubuntu`.) You
+could do something like this:
+
+```none
+$ docker-machine scp -r /Users/londoncalling/webapp MACHINE-NAME:/home/ubuntu/webapp
+```
+
+Then write a docker-compose file that bind mounts it in:
+
+```none
+version: "3.1"
+services:
+  webapp:
+    image: alpine
+    command: cat /app/root.php
+    volumes:
+    - "/home/ubuntu/webapp:/app"
+```
+
+And we can try it out like so:
+
+```none
+$ eval $(docker-machine env MACHINE-NAME)
+$ docker-compose run webapp
+```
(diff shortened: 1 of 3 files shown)


machine/reference/scp.md
```diff
@@ -27,5 +29,48 @@ A file created remotely!
...
 
-In the case of transferring files from machine to machine, they go through the
-local host's filesystem first (using `scp`'s `-3` flag).
\ No newline at end of file
+In the case of transferring files from machine to machine,
+they go through the local host's filesystem first (using `scp`'s `-3` flag).
+
+## Understanding file paths for remote deployments
+
+If you copy files to a remote server with `docker-machnie scp`
```
nathanleclaire: typo here: `docker-machine scp`

londoncalling: :-) thanks @nathanleclaire fixing now

machine/reference/scp.md
```diff
@@ -27,5 +29,48 @@ A file created remotely!
...
+  links:
+    - postgres:postgres
+  volumes:
+    - /usr/src/app/static
+    - ./data:/usr/src/app/data
+  env_file: .env
+  command: /usr/local/bin/gunicorn --workers 4 --timeout 120 --bind :8000 app:app
+```
+
+* The first path given for [volumes](/compose/compose-file/index.md#volumes-volumedriver), `/usr/src/app/data`, is the path _inside_ the container.
```
nathanleclaire: Wrap?

machine/reference/scp.md
```diff
@@ -27,5 +29,48 @@ A file created remotely!
...
+
+* The second path, `./data:/usr/src/app/data`, is the path where
```
nathanleclaire: If you're writing this up for remote deployments, I'd shy away from the "relative path method" (i.e., `.:/foo` indicating that `.`, the current directory, should be bind-mounted into `foo`) completely. Arguably it was somewhat of a mistake to include in Compose in the first place as it's caused no end of confusion and does not accurately map to what gets sent to the Docker API. It's client side magic.

My point is that in a writeup on remote paths you should call it out explicitly that this won't work and use absolute paths. There's a big opportunity to clear up a long-term point of confusion here. I'd address it something like, "You can specify an absolute path, e.g. `/home/myuser/workspace`, which will be mounted into the container at `/workspace`, from the absolute path on the _remote host_ where the Docker daemon is running. Local client paths (e.g., your Mac laptop) will not work for daemons running on a remote VM."

machine/reference/scp.md
londoncalling: Because of the link and the bullet, it doesn't really wrap properly no matter what I do. This way, the link formats properly. I'm not sure how to fix this.

londoncalling: Never mind, I just fixed it.

machine/reference/scp.md
londoncalling: Cool, great!

londoncalling: The first example with `.:/foo` was in the original topic, I'll try to update/change that whole example per your comments.

machine/reference/scp.md
mstanleyjones: If you line up the wrapping with the T in the `The`, it will work.

machine/reference/scp.md
```diff
@@ -27,5 +29,50 @@ A file created remotely!
...
+    - ./data:/usr/src/app/data
+  env_file: .env
+  command: /usr/local/bin/gunicorn --workers 4 --timeout 120 --bind :8000 app:app
+```
+
+* The first path given for
+[volumes](/compose/compose-file/index.md#volumes-volumedriver),
+`/usr/src/app/data`, is the path _inside_ the container.
+
+* The second path, `./data:/usr/src/app/data`, is the path where
```
nathanleclaire: I'm a little confused what the example would like to illuminate -- you could certainly `docker-machine scp` a directory over and bind mount it in to a container, but if that were the case I'd:

1. Leave out a reference to the "in-container" volume (`/usr/src/app/data`) since it's not really relevant to the issue at hand
2. Illustrate with a simple example including `docker-machine scp`

e.g.,

For instance, imagine if you wanted to transfer your local directory `/Users/londoncalling/webapp` to a remote VM and bind mount it into a container on the remote host (we'll suppose the remote user is `ubuntu`). You could do something like this:

```
$ docker-machine scp -r /Users/londoncalling/webapp machinename:/home/ubuntu/webapp
```

Then write a `docker-compose` file which bind mounts it in:

```
version: "3.1"
services:
  webapp:
    image: alpine
    command: cat /app/root.php
    volumes:
    - "/home/ubuntu/webapp:/app"
```

And we can try it out like so:

```
$ eval $(docker-machine env machinename)
$ docker-compose run webapp
```

PR #2268, opened 2017-03-10
Add new Azure regions in FAQ

ddebroy:
### Proposed changes

Azure introduced support for two new regions in Korea. We also started supporting Gov regions. Add those new regions in our FAQ notes of supported region strings.

@mstanleyjones @FrenchBen


mstanleyjones:
Is it feasible for us to keep documenting supported regions? Do we not support all of Azure's regions? Can we just link to an Azure document for this?

I'm going to go ahead and merge this, but feel free to comment.  cc/ @johndmulhausen

PR #2275, opened 2017-03-13
AWS CloudFormation service role policy size error

jakubknejzlik:
When trying to create AWS Policy for docker swarm, the response in console I get is:

`An error occurred: Cannot exceed quota for PolicySize: 5120`

### Proposed changes

Shrink EC2 actions to `ec2:*` to reduce policy size.


mstanleyjones:
@frenchben ptal

FrenchBen:
@jakubknejzlik Thank you for the suggestion. 

You're more than welcome to apply the `ec2:*` to your policy, but we want to stay as explicit as we can on what permissions are required and thus can't merge this change. 

If you would like to instead split the policy into chunks that can be created within AWS, it would be better IMHO

mstanleyjones:
Please address @FrenchBen's comments in a new PR if you still wish to propose this change.

PR #2276, opened 2017-03-13
Update orgs.md

scjane:
I think these are two separate sentences. Shall we simply use a semicolon (;), or end the first one with a period (.) and then write the second one?


johndmulhausen:
Thanks, @scjane :)

PR #2278, opened 2017-03-13
Fix depends_on in Compose file v3 reference

dnephin:
includes #2164
Fixes https://github.com/docker/docker/issues/30404#issuecomment-285972049

### DIFF
--- a/compose/compose-file/index.md
+++ b/compose/compose-file/index.md
@@ -348,6 +348,8 @@ Simple example:
 > for a service to be ready, see [Controlling startup order](/compose/startup-order.md)
 > for more on this problem and strategies for solving it.
 
+> **NOTE:** version 3 no longer supports the `condition` form of `depends_on`
+
 > **Note:** This option is ignored when
 > [deploying a stack in swarm mode](/engine/reference/commandline/stack_deploy.md)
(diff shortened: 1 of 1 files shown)


compose/compose-file/index.md
```diff
@@ -348,33 +348,12 @@ Simple example:
 > for a service to be ready, see [Controlling startup order](/compose/startup-order.md)
 > for more on this problem and strategies for solving it.
 
-A healthcheck indicates you want a dependency to wait
-for another container to be "healthy" (i.e. its healthcheck advertises a
-successful state) before starting.
+> **NOTE:** version 3 no longer supports the `condition` form of `depends_on`
```
johndmulhausen: @dnephin I think this note can be moved to https://docs.docker.com/compose/compose-file/compose-versioning/#version-3

dnephin: I disagree. That section is for fields that have been removed entirely, because there is no where else to document them. In this case the field still exists, but it only supports the "short form".  I think users would expect to see this documented in the field section.

mstanleyjones:
I just committed a change that combined the 3(!) admonitions into a single one.

PR #2282, opened 2017-03-13
updated command variables for consistency

londoncalling:
Update to command variables to create the virtual machines


londoncalling:
LGTM

PR #2287, opened 2017-03-13
Fix the template not to include Related Commands section for dockerd

mstanleyjones:
The logic that prevented the `docker` command from having a Related Commands table failed to prevent the same thing for the `dockerd` command. Just had to adjust the logic. Also adjusted the logic of a couple other areas to use an `unless` loop instead of `if` loop because it reads more clearly to me.

### DIFF
--- a/_includes/cli.md
+++ b/_includes/cli.md
@@ -14,7 +14,7 @@
 
 ## Usage
 
-```shell
+```none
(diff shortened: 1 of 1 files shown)


mstanleyjones:
LGTM in Netlify!

PR #2290, opened 2017-03-13
Use the jekyll-include-cache plugin to shave off some page building time

mstanleyjones:
This is a very naive first implementation which only caches truly static content like the footer and archives. If we could make use of this for building the left-hand nav tree, I think it would be pretty cool.

cc/ @allejo


johndmulhausen:
Naive or no, this is actually as good as we can do without significantly changing the plumbing on treebuilder.html and tree.html. Absolutely everything else has per-page specificity that doesn't gain us anything.

Testing locally, it finished the build in 143 seconds. Ran the current master build and it finished in roughly the same time, so I don't think this buys us much but it sets the stage for us doing more with caching where we can.

PR #2291, opened 2017-03-14
Fix CSS loading order, add bootstrap JS

mstanleyjones:
The bootstrap JS was not included, and CSS was getting loaded out of order, causing the page color to be wrong.


johndmulhausen:
LGTM; thanks, @mstanleyjones :)

PR #2299, opened 2017-03-14
Brief introduction of "swarm"

AyushyaChitransh:
### Proposed changes

It was unclear what swarm is. I had no clue. Simple Google results were unable to give an exact idea to what swarm mode is. Hence I wasn't able to decide where to go next.
Word Swarm has been used here for the first time, and one line introduction, to what swarm is, would be very beneficial in helping a newbie, to decide where to begin with.

### Related issues (optional)

Closes ##2225


johndmulhausen:
Simple enough, thanks @AyushyaChitransh :)

PR #2301, opened 2017-03-14
Update node-setup.md

fredyagomez:
### Proposed changes

You guys have the instruction for Windows 10, but most people run other version like me (Windows 7). The Hyper-V to create virtual machine only works in Windows 10. I am proposing a really straightforward approach for Windows 7, which actually is 100% similar to the Mac approach,: download Virtualbox from Oracle and start creating virtual machines from the command line.

### Related issues (optional)

No time for this. Sorry.

### DIFF
--- a/engine/getstarted-voting-app/node-setup.md
+++ b/engine/getstarted-voting-app/node-setup.md
@@ -31,7 +31,9 @@ using the Toolbox installer because it can
 here](https://www.virtualbox.org/wiki/Downloads), and follow install
 instructions. You do not need to start VirtualBox. The `docker-machine create`
 command will call it via the driver.
+
 <p />
+
 * **Hyper-V driver on Docker for Windows** - On Docker for Windows, you
 will use `docker-machine` with the [`Hyper-V`](/machine/drivers/hyper-v/) driver
 to create machines. You will need to follow the instructions in the [Hyper-V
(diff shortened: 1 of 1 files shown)


johndmulhausen:
@friism

friism:
What @fredyagomez is suggesting is Docker Toolbox: https://docs.docker.com/toolbox/toolbox_install_windows/

@fredyagomez what docs where you looking at? Maybe we should make the toolbox links more prominent?

fredyagomez:
@friism So you are right, this link tells the user how to create virtual machines in Windows 7/8. Since the main goal of the node-setup page is to create virtual machines, there should be a link or comment to this page that redirects those users using Windows 7/8 on the process. In my case, I found the info for Mac, then Windows 10, since none of these two applied to me, then I figured out the way in Windows 7/8 which is exactly the link you pasted.

mstanleyjones:
@londoncalling ptal

londoncalling:
@fredyagomez you are absolutely right, I need to add links to Toolbox for Windows. I'd like to go ahead and re-write this from scratch, using some of your content here as a guideline. I'm going to open a new PR and tag you and @friism on it, then close this one.

londoncalling:
@fredyagomez you are absolutely right, I need to add links to Toolbox for Windows. I've created a PR #2369, as it was going to be a bit too complicated to modify your PR. 

I used some of your content as a guide, but added links to the appropriate Toolbox topics on the node setup page and the first page of the sample app. (cc @friism )

londoncalling:
Closing this PR, see my previous comment ^^ and PR #2369

PR #2309, opened 2017-03-14
Update step_six.md

lostwong:
### Proposed changes

Minor fix to indicate that the Image ID is in the third column and not the second column.


johndmulhausen:
Nice catch - thanks, @lostwong :)

PR #2318, opened 2017-03-14
Added ability to designate a branch based off master as an archive

mstanleyjones:
- Set `is_archive: true` in `_config.yml`.
- Be sure the correct version is set at the top of `js/archive.js`. You will also need to pull down
  the `archives.json` from https://docs.docker.com/js/archives.json and put it in your `js/` directory,
  and change the path to the JSON file in `archive.js`. This is just for testing and shouldn't
  be committed.

When you set this, three things happen:
- the archive.js is loaded in both the global header and the normal header
- the archive header shows on every page
- the docs archive link is not shown in the left-hand nav

**This is not ready to go yet.** The archive header is currently hidden behind the
fixed-position header in normal pages. To see what I mean, set `is_archive: true` in `_config.yml`
and build the site locally. Go to a page like http://localhost:4000/engine/getstarted/ and look
in the inspector for the div before the `<header>` tag. @johndmulhausen PTAL

Please don't commit with `is_archive` set to `true`.

### DIFF
--- a/engine/swarm/services.md
+++ b/engine/swarm/services.md
@@ -507,7 +507,7 @@ after that is not counted.
 ## Roll back to the previous version of a service
 
 In case the updated version of a service doesn't function as expected, it's
-possible to roll back to the previous version of the service using
+possible to manually roll back to the previous version of the service using
 `docker service update`'s `--rollback` flag. This will revert the service
 to the configuration that was in place before the most recent
 `docker service update` command.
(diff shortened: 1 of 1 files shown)


johndmulhausen:
Looks solid in the PR; I'll have to peek at what this looks like when you test it in an `is_archive` branch, though. Peek over your shoulder tomorrow or something? :)

mstanleyjones:
Sure! It's a CSS problem to do with the fixed top header. We need to get the archive div above it, to match the other pages. I'm not sure the right way to do it.

PR #2337, opened 2017-03-15
Update deployment-strategies.md

scjane:
Is "when" unnecessary?


johndmulhausen:
It isn't, neither is that parenthesis. Thanks, @scjane :)

PR #2342, opened 2017-03-15
bridge option syntax

lbt:
It seems that the syntax is either:
-b <bridge>
or
--bridge=<bridge>


mstanleyjones:
The short options don't take an = sign, just a space. Thanks!

PR #2343, opened 2017-03-15
Update !#/bin/bash to !#/bin/sh startup-order.md

kellyjandrews:
### Proposed changes

Changed `!#/bin/bash` to `!#/bin/sh` because the script above was throwing `standard_init_linux.go:178: exec user process caused "no such file or directory"`.  Maybe I was doing something incorrectly, but switching to `sh` allowed it to run.

### DIFF
--- a/compose/startup-order.md
+++ b/compose/startup-order.md
@@ -53,7 +53,7 @@ script:
     check. For example, you might want to wait until Postgres is definitely
     ready to accept commands:
 
-        #!/bin/bash
+        #!/bin/sh
         # wait-for-postgres.sh
 
         set -e


mstanleyjones:
You can do some troubleshooting on this by running the `postgres` image interactively:

```
$ docker run --rm -it postgres sh
```

The `--rm` switch says to remove the container when it exits. The `-it` switches run the container interactively and attached to your terminal. The `sh` is what to run inside the container, instead of the default endpoint which would be the database.

You're dropped into the container at a root prompt. Then you can run some commands:

```
# which sh
/bin/sh

# which bash
/bin/bash

# ls -l /bin/sh
lrwxrwxrwx 1 root root 4 Nov  8  2014 /bin/sh -> dash

# ls -ld /bin/bash
-rwxr-xr-x 1 root root 1029624 Nov  5 21:22 /bin/bash
```

You can stop and remove the container by just typing `exit` (because of the `--rm` switch).

So, in this case, `sh` is actually a symbolic link to `dash`, which is a lightweight shell that doesn't have all the features of Bash. However, Bash does exist and is where it is supposed to be.

You might want to verify that your `postgres` image is up to date and try again. I don't like to change these docs if Bash really is there. Can you double-check by running the  commands above and let us know?

kellyjandrews:
Ah - ok - that makes sense. I was running the script as an `ENTRYPOINT` to my node app running the `node:7.7.2-alpine` image.  

```shell
# ls -l /bin/sh
lrwxrwxrwx    1 root     root            12 Mar  3 11:20 /bin/sh -> /bin/busybox
# ls -ld /bin/bash
ls: /bin/bash: No such file or directory
```
`bash` doesn't exist on Busy Box at all - which does explain the error I was seeing. The above troubleshooting, however, is super useful, and I've learned something today, so all good.

I ended by changing to `node:7.7.2-slim` where `bash` is preinstalled, and I get the results you have above.

```shell 
# ls -l /bin/sh
lrwxrwxrwx 1 root root 4 Nov  8  2014 /bin/sh -> dash

# ls -ld /bin/bash
-rwxr-xr-x 1 root root 1029624 Nov  5 21:22 /bin/bash
```

I don't want to change documentation if not required - but it might be good to note somewhere that the alpine version won't have `bash` installed, unless that is just something I should have know already and was completely naive here. 

Thank you so much for the really helpful response.  I will go ahead and close this, and let you all reopen if warranted.

mstanleyjones:
So, I think it is general knowledge among people who use Alpine that Alpine doesn't have `bash` by default. The shell for Alpine is `ash`. I'm not sure that the Docker docs are the right place to talk about Alpine details, though. I'm sorry this tripped you up.

PR #2344, opened 2017-03-15
Docs base

mtanlee:


mstanleyjones:
What is the purpose of this pull request?

johndmulhausen:
@mtanlee I think you meant to propose an update to `docs-base`. This PR would replace the contents of `master` with `docs-base` which is not something we want to do. Feel free to try again. :)

PR #2349, opened 2017-03-15
Run bundle update before running jekyll build

mstanleyjones:
Also added a few more comments


johndmulhausen:
Dupe of #2348 ?

Reminder of a couple things we talked about:

1) I believe you'll need to actually run `gem install bundler` first for this to work
2) This is being run on the archives, which I don't believe have `Gemfile`s to use
3) If we do upload `Gemfile`s to the archive branches, and want to use bundler to install separate Ruby environments for each build, you'll want to run `bundle exec jekyll build` instead of just `jekyll build` when building the archives.

Issue #1997, opened 2017-02-27
Feedback for: engine/admin/systemd.md


MattacksWTW:
[engine/admin/systemd.md](https://docs.docker.com/engine/admin/systemd/), CC: @mstanleyjones

"# or on older distributions, you may need to use"
What does 'older' mean?

The help is useless without telling me what it means, or a means to find out what it means

mstanleyjones:
Fixed by #2353

PR #2353, opened 2017-03-15
Update systemd.md

mstanleyjones:
Addressing feedback in #1997


thaJeztah:
Oh, sorry I didn't notice your ping. I don't think we should remove all instructions for configuring through systemd, because there are cases where people either _prefer_ to use systemd, or systemd-specific options have to be overridden. Although we should direct users to the `daemon.json` configuration file, we _do_ need a section explaining the use of systemd drop-in files (for example to override the default options set in the systemd unit file https://github.com/docker/docker/blob/17.04.x/contrib/init/systemd/docker.service.rpm#L14-L30)

PR #2354, opened 2017-03-15
Update URL for Windows binaries, rewrite install.sh topic

mstanleyjones:
Fixes https://github.com/docker/docker/issues/31514#issuecomment-284856636

### DIFF
--- a/engine/getstarted/linux_install_help.md
+++ b/engine/getstarted/linux_install_help.md
@@ -3,36 +3,72 @@ description: Getting started with Docker
 keywords: beginner, getting started, Docker, install
 redirect_from:
 - /mac/started/
-title: 'Example: Install Docker on Ubuntu Linux'
+title: Install Docker using a shell script
 ---
 
-This installation procedure for users who are unfamiliar with package
-managers, and just want to try out the Getting Started tutorial while running Docker on Linux. If you are comfortable with package managers, prefer not to use
-`curl`, or have problems installing and want to troubleshoot, please use our
-`apt` and `yum` <a href="/engine/installation/"
-target="_blank">repositories instead for your installation</a>.
+This installation procedure for users who don't want to use a package manager to
+install Docker. The script works on a "best effort" basis to determine your
+operating system and environment, and attempts to provide reasonable defaults.
+The script may allow you to install Docker in environments that are not actually
+supported configurations.
 
-1. Log into your Ubuntu installation as a user with `sudo` privileges.
+If you can use a package manager, you should use the
+[recommended installation procedure for your operating system](/engine/installation/)
+instead. Using a package manager ensures that you get upgrades when they are
+available, and allows you to install a specific version of Docker, rather than
+the very latest version.
 
-2. Verify that you have `curl` installed.
+## Prerequisites
 
-        $ which curl
+- You need `sudo` access on Linux, or administrator access on Windows.
+- You need `curl` or `wget` installed. These instructions use `curl`, but you
+  can adapt them to use `wget`.
 
-    If `curl` isn't installed, install it after updating your manager:
+## Install Docker using the `install.sh` script
 
-        $ sudo apt-get update
-        $ sudo apt-get install curl
+> **Warning**: Always examine shell scripts you download from the internet before
+> running them.
 
-3. Get the latest Docker package.
+1.  Open [https://get.docker.com](https://get.docker.com/) in your web browser
+    so that you can examine the script before running it. This is important
+    because the script will run with elevated privileges.
 
-        $ curl -fsSL https://get.docker.com/ | sh
+2.  Run the script, using `curl` to download it and piping it through `sh`:
 
-    The system prompts you for your `sudo` password. Then, it downloads and
-    installs Docker and its dependencies.
+    ```bash
+    $ curl -fsSL https://get.docker.com/ | sh
+    ```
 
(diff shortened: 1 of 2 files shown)


engine/getstarted/linux_install_help.md
```diff
@@ -3,36 +3,49 @@ description: Getting started with Docker
...
+1.  Open [https://get.docker.com](https://get.docker.com/) in your web browser
+    so that you can examine the script before running it. This is important
+    because the script will run with elevated privileges.
 
-3. Get the latest Docker package.
+2.  Run the script, using `curl` to download it and piping it through `sh`:
 
-        $ curl -fsSL https://get.docker.com/ | sh
+    ```bash
+    $ curl -faSL https://get.docker.com/ | sh
```
thaJeztah: should this be `-fsSL`? I don't think `-a` applies here?

engine/getstarted/linux_install_help.md
```diff
@@ -3,36 +3,49 @@ description: Getting started with Docker
...
-    >command fails for the Docker repo during installation. To work around this,
-    >add the key directly using the following:
+    > **Note**: If your company is behind a filtering proxy, the
+    > `apt-key` may fail during Docker installation. To work around this,
+    > use the following command to manually add the Docker key:
     >
     >       $ curl -fsSL https://get.docker.com/gpg | sudo apt-key add -
+
+  3.  The script attempts to start Docker and to configure the system to start
+      Docker automatically.
```
thaJeztah: Only on systems using "apt" (`.deb`) packages, on RPM based platforms, docker is not started, or configured to start automatically (per convention for RPM based installs)

engine/getstarted/linux_install_help.md
```diff
@@ -3,36 +3,49 @@ description: Getting started with Docker
...
+> running them.
 
-        $ sudo apt-get update
-        $ sudo apt-get install curl
+1.  Open [https://get.docker.com](https://get.docker.com/) in your web browser
+    so that you can examine the script before running it. This is important
+    because the script will run with elevated privileges.
 
-3. Get the latest Docker package.
+2.  Run the script, using `curl` to download it and piping it through `sh`:
```
Silex: `2.  Run` has an extra space (2 instead of 1)

Silex: `2.  Run` has an extra space (2 instead of 1) between the dot and `Run`.

mstanleyjones: Actually, the others are missing a space. Because tab stops are at 2 spaces, numbered lists need two spaces so that anything indented within them lands on a tab stop. Jekyll is picky about this! Thanks for catching this.

Silex: Ah, didn't know that. Any number of spaces seems to work for the title, only the body needs a multiple of 2 spaces, but you're right the raw text looks prettier that way as it's aligned.

mstanleyjones: Yep, and if you mis-align something like a codeblock inside a numbered list, it can be hard to debug.

mstanleyjones:
Addressed the feedback, @thaJeztah

engine/getstarted/linux_install_help.md
```diff
@@ -3,36 +3,65 @@ description: Getting started with Docker
...
 
-        $ curl -fsSL https://get.docker.com/ | sh
+    ```bash
+    $ curl -fsSL https://get.docker.com/ | sh
+    ```
 
-    The system prompts you for your `sudo` password. Then, it downloads and
-    installs Docker and its dependencies.
+    You are prompted for your `sudo` password. The script determines your
+    operating system, downloads Docker and its dependencies, and installs Docker.
```
thaJeztah: nit; "downloads, and installs, the Docker package and its dependencies. " perhaps (duplicate "Docker" in the existing sentence)?

engine/getstarted/linux_install_help.md
```diff
@@ -3,36 +3,65 @@ description: Getting started with Docker
...
 
-    >**Note**: If your company is behind a filtering proxy, you may find that the
-    >`apt-key`
-    >command fails for the Docker repo during installation. To work around this,
-    >add the key directly using the following:
+    > **Note**: Usuntu or Debian users whose host is behind a filtering proxy
```
thaJeztah: oops, missed this one; typo s/Usunty/Ubuntu/

engine/getstarted/linux_install_help.md
```diff
@@ -3,36 +3,65 @@ description: Getting started with Docker
...
-managers, and just want to try out the Getting Started tutorial while running Docker on Linux. If you are comfortable with package managers, prefer not to use
-`curl`, or have problems installing and want to troubleshoot, please use our
-`apt` and `yum` <a href="/engine/installation/"
-target="_blank">repositories instead for your installation</a>.
+This installation procedure for users who don't want to use a package manager to
+install Docker. If you can use a package manager, you should use the
+[recommended installation procedure for your operating system](/engine/installation/)
+instead. Using a package manager ensures that you get upgrades when they are
+available, and allows you to install a specific version of Docker, rather than
+the very latest version.
```
thaJeztah: Probably for a follow-up, but we should point out that the script is working on a "best effort" base, both to determine the operating system, and to configure docker with a default configuration "that works", but may not be tuned for production.

mstanleyjones:
Addressed remaining feedback.

PR #2358, opened 2017-03-16
fixed typo from adressses to addresses

adamoverton:


johndmulhausen:
Thanks, @adamoverton :)

PR #2371, opened 2017-03-17
Update ssh-into-a-node.md

scjane:


johndmulhausen:
Good catches, thanks @scjane :)

PR #2378, opened 2017-03-17
Add a timeout to the Jenkinsfile

mstanleyjones:
Adds a 60-minute timeout, after which the Jenkins job will fail. This should prevent jobs hanging around for months in weird states.


mstanleyjones:
Verified this works. We may need to adjust the timeout in the future, but right now our CI job on master takes about 10 minutes, so 60 seems more than adequate.

PR #2383, opened 2017-03-18
Add Glyphicon for pull request links

mstanleyjones:
Use Glyphicon for pull request links


mstanleyjones:
@johndmulhausen I don't think we actually have this glyph. When I build this locally, the span doesn't have any image in it and I can't find the glyph mentioned in any of our CSS by grepping it.

mstanleyjones:
We don't have the glyph and not paying $60 to get it just for this.

PR #2385, opened 2017-03-18
Improved word usage

vishnu-narayanan:


johndmulhausen:
Thanks, @vishnu-narayanan :)

PR #2386, opened 2017-03-18
Add a procedure to sort

kotonikushidori:
### Proposed changes
The search results  described in the manual, many containers were displayed, 
and we could not reach "docker/whalesay".


johndmulhausen:
@kotonikushidori Sorry, but this section of text has been removed. Thanks for the PR though, don't let it keep you from sending us more changes in the future. :)

PR #2391, opened 2017-03-19
update step_three.md

ql-eilde:
Update docker hub link by docker store link (docker/whalesay is no more in the results of docker hub search). Would be interesting to update the illustration images with docker store illustrations.


johndmulhausen:
Thanks, nipped and tucked it a bit. :)

PR #2399, opened 2017-03-20
Update docker-upgrade.md

scjane:
Is my change appropriate? Or shall we list them as two steps as follows:
Before you upgrade your nodes:

1.Go to the detail page, xxx

2.If a new version is available, xxx


johndmulhausen:
This looks right; not sure why it was so weird to begin with.

johndmulhausen:
Thanks, @scjane :)

PR #2401, opened 2017-03-20
Fix external links from compose file docs

mkurek:
Compose file v1 and v2 docs had relative links to some non-existing pages (especially about extending services). It was probably introduced when `compose-file.md` was detached into directory with single file per compose file version.

### Proposed changes

Fix for external links from compose file docs.


johndmulhausen:
Thanks, @mkurek 
CC @londoncalling FYI

PR #2405, opened 2017-03-20
updated url and download script

btro:

### DIFF
--- a/machine/completion.md
+++ b/machine/completion.md
@@ -14,12 +14,9 @@ for the bash and zsh shell.
 Make sure bash completion is installed. If you use a current Linux in a non-minimal installation, bash completion should be available.
 On a Mac, install with `brew install bash-completion`
 
-Place the completion scripts in `/etc/bash_completion.d/` (`` `brew --prefix`/etc/bash_completion.d/`` on a Mac), using e.g.
+Place the completion script in `/etc/bash_completion.d/` (`` remove `brew --prefix`/etc/bash_completion.d/`` on a Linux), using e.g.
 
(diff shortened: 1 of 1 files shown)


machine/completion.md
```diff
@@ -14,12 +14,9 @@ for the bash and zsh shell.
 Make sure bash completion is installed. If you use a current Linux in a non-minimal installation, bash completion should be available.
 On a Mac, install with `brew install bash-completion`
 
-Place the completion scripts in `/etc/bash_completion.d/` (`` `brew --prefix`/etc/bash_completion.d/`` on a Mac), using e.g.
+Place the completion script in `/etc/bash_completion.d/` (`` remove `brew --prefix`/etc/bash_completion.d/`` on a Linux), using e.g.
```
mstanleyjones: I don't understand this change. What is the "remove" word for? Also "on a Linux" is not clear or grammatically correct. Can you please work with @londoncalling on the problem you are trying to address with this PR?

btro: Oh, yes the "on a Linux" part is incorrect. I just replaced "Mac" with "Linux" in this line, thats why the error is in there. The remove part states, that you have to remove the "brew --prefix`/etc/bash_completion.d/" from the following command if you use Linux.

londoncalling:
@btro I've pushed an update, and I think this captures the spirit of the changes with a little more clarity. @mstanleyjones can you review and approve these changes and merge this? I think you have to give your approval before we can merge the change?

londoncalling:
LGTM, merging

PR #2410, opened 2017-03-20
Improve doc tests

mstanleyjones:
Reorganized the test page
Added more info about typography, mixing HTML and Markdown, using Bootstrap grid classes, setting HTML styles on both HTML and Markdown elements, and some other stuff

I think this organization makes a little more sense from a "how do I" standpoint.

I noticed some CSS that is missing while working on this:

- For all the Bootstrap `panel` classes, we set the border color to transparent.
- If you look at the 2-column HTML table that has buttons inside it, we are not clearing the buttons, even when I put them inside `<p>` tags.

cc/ @JimGalasyn who I can't choose as a reviewer yet

### DIFF
--- a/test.md
+++ b/test.md
@@ -144,7 +144,8 @@ Some tables in markdown and html.
 
 | Permission level | Access        |
 | ---------------- | ------------- |
-| **Bold** or _italic_ within a table cell | |
+| **Bold** or _italic_ within a table cell. Next cell is empty on purpose. | |
+|                  | Previous cell is empty. A `--flag` in mono text. |
 | Read             | Pull |
 | Read/Write       | Pull, push |
 | Admin            | All of the above, plus update description, create and delete |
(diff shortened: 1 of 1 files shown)


test.md
```diff
@@ -180,33 +171,153 @@ only need to do it on the first one. If you have a `<th>`, set it there.
...
+break the Markdown block up. This example does it with Markdown. You can't have any blank lines or it will
+break the Markdown block up. This example does it with Markdown. You can't have any blank lines or it will
+break the Markdown block up. This example does it with Markdown. You can't have any blank lines or it will
+break the Markdown block up. This example does it with Markdown. You can't have any blank lines or it will
+break the Markdown block up. This example does it with Markdown. You can't have any blank lines or it will
+break the Markdown block up. This example does it with Markdown. You can't have any blank lines or it will
+break the Markdown block up.
+{: style="column-count: 3 "}
+
+## Admonitions
```
mstanleyjones: OK, but it's fairly standard "tech writer" speak for those boxes themselves.

test.md
```diff
@@ -180,33 +171,153 @@ only need to do it on the first one. If you have a `<th>`, set it there.
...
+You can use block-level HTML tags too. This paragraph is centered.
+</center>
+
+Keep reading for more examples, such as creating tabbed content within the
+page or displaying content as "cards".
+
+## Jekyll / Liquid tricks
+
+This paragraph is centered and colored green by setting CSS directly on the element.
+{: style="text-align:center; color: green" }
```
mstanleyjones: I don't think we want to encourage contributors to add CSS styles. I agree with you in principal, but ...

PR #2426, opened 2017-03-21
Fix Windows container reference

sixeyed:
### Proposed changes

Clarified the sentence on running Linux containers on Windows. You can't do that, but you can run Linux containers with *Docker for Windows*, which spins up a Moby VM for Linux containers.


johndmulhausen:
Fair enough, thanks @sixeyed :)

PR #2429, opened 2017-03-21
Fixing path to make.sh

ManoMarks:
### Proposed changes

Fixed path to make.sh on Work with a development container page


ManoMarks:
Sorry mis-interpreted what I was looking at in the docs

PR #2437, opened 2017-03-22
Update faqs.md

Paxxi:
### Proposed changes

setting DOCKER_HOST to `http://` fails with error `Invalid bind address format: http://localhost:2375`


johndmulhausen:
Thanks, @Paxxi :)

PR #2454, opened 2017-03-23
Fix docker-machine GCE options listing

zigarn:
### Proposed changes

Fix docker-machine GCE options listing.

There was too many spaces and it was renedered as a code block.


johndmulhausen:
Thanks, @zigarn !

zigarn:
You're welcome @johndmulhausen

PR #2457, opened 2017-03-23
Update content_trust.md

scottstientjes:
### Proposed changes

Removed "Using default tag: latest" from sample command output.  The testing tag was specified in the command, so it would not being using the default tag.


joaofnfernandes:
Thanks for the contribution 🐳

PR #2322, opened 2017-03-14
clarified Compose file topic structure, and file examples

londoncalling:
### What's changed

- In Compose file v.3, linked to topics that have examples of Compose files (Docker Labs, Voting app getting started sample)

- In Voting app getting started sample, moved link to full Compose file closer to the example snip-it of a service on the first page

### Related issues

#1250 

### Reviewers

@sbrattla @shin- @dnephin @ManoMarks 

Signed-off-by: Victoria Bialas <victoria.bialas@docker.com>

Issue #1250, opened 2017-01-20
Indicate if a configuration option is a build option or runtime option


sbrattla:
Please refer to https://docs.docker.com/compose/compose-file/ for this issue.

I find it hard to see from the documentation if a configuration option is a build option or runtime option. By this, I mean that it's hard to see if an option is respected by docker-compose when building the container or when running the container.

For instance, the `dockerfile` option is a build option. There is no way to distinguish this option from the `expose` option which is a runtime option. There might be options which are taken into consideration both during build and run; but then that should be indicated as well. Right now, there is a **build** heading / section on that page but the **build** heading  is on the same level as all other headings (`h3`). I would at least expect the build sub headings to be `h4` and to see a corresponding runtime (`h3`) heading where sub headings would be `h4` as well.

A better way, in my opinion, would be to add a build or runtime label (or both) just next to the configuration option name. 

The reason I came across this is that we use the same `docker-compse.yml` for both local development (`docker-compose build`) and swarm deployments. The "runtime" configuration options in `docker-compose.yml` must be provided to `service create` as arguments - and it would be nice to use the configuration as a definitive guide to which options we would have to provide during `service create`.

![image](https://cloud.githubusercontent.com/assets/1845372/22149056/f1e7e1fc-df11-11e6-803c-5083ffe7df6b.png)

dnephin:
The fields that apply to build are under the `build` section (https://docs.docker.com/compose/compose-file/#build). They are H4 headings.

The rest are runtime config. This was fixed in c2 of the Compose file.

londoncalling:
@sbrattla Please see Daniel's response ^^. I hope this clarifies build v. runtime for you. Please ping us if you have more questions or feedback.

sbrattla:
@londoncalling I have to admit that I still think there is something odd with how things are presented. Considering that **build** (the h3 header) represents build parameters, why is then **expose** also a h3 header? **build** is an "abstract" term representing build parameters, while **expose** is a part of the Dockerfile syntax. 

My (naive and simplistic?) approach would be that there are two "top" groups : (1) build and (2) runtime. I would then add "subsections" to these groups such as "Volumes", "Networking" etc. Then I would add the actual Dockerfile syntax terms to these subgroups again. This would make it much easier understand the context of a specific syntax term.

If a term for some reason would be both a runtime and build parameter, then maybe something as simple as a @see notation could be introduced.

dnephin:
[...]

PR #2477, opened 2017-03-25
added compose file example in accordion

londoncalling:
### What's changed

- added the voting app sample Compose file (`docker-stack.yml`) to the Compose file version 3 reference

- the goal is to help clarify the structure and format of Compose files, per the referenced PR's and issues below

- fyi, `markdownify` didn't work here (it worked for the code block with the exception that the backticks showed), so I used the HTML tags recommended on the Bootstrap.js site

### Related issues and PR's

Issue #1250, PR #2322 

### Reviewers, fyi

@sbrattla @shin- @dnephin @ManoMarks @jsouth @johndmulhausen 

### DIFF
--- a/compose/compose-file/index.md
+++ b/compose/compose-file/index.md
@@ -22,14 +22,120 @@ The topics on this reference page are organized alphabetically by top-level key
 to reflect the structure of the Compose file itself. Top-level keys that define
 a section in the configuration file such as `build`, `deploy`, `depends_on`,
 `networks`, and so on, are listed with the options that support them as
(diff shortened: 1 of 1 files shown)


londoncalling:
LGTM

dnephin:
Is there no way to include this sample from the file? Duplicating the same feels like it will be extra work to maintain.

londoncalling:
@dnephin Hi, I could link them out to file on GitHub, but I just felt the file would be more useful for this purpose if it were inline. I can change this if you prefer.

PR #2520, opened 2017-03-28
Cloudstor docs

ddebroy:
### Proposed changes

The documentation of the `cloudstor` plugin for our cloud editions is incorrect. Users cannot get swarm services using cloudstor to come up using what is specified in the documentation. This has led to multiple questions from uses in forums/slack channels and unnecessary debugging sessions. The documentation got out of sync due to a change in the installation mechanism of `cloudstor` to use an alias. The changes fixes the docs so the example commandlines in the docs uses the alias.

### Related issues (optional)

https://github.com/docker/editions/commit/18c5eead86645eb25fd59648356db011500ed1cf

@FrenchBen @mstanleyjones @mavenugo


mstanleyjones:
LGTM, assuming you have tested this procedure.

PR #2537, opened 2017-03-29
Suggestion to add the offline bundle step

rahim888:
Suggestion to add the offline bundle step since user needs to run against the cluster


joaofnfernandes:
In the past we had something like this in the docs, we eventually switched to asking users to ssh into a ucp node, but in the end we removed all that because you can run the DTR installer from any node, even outside the UCP cluster.

The only requirement is that the node where you're running the command can reach the UCP nodes and trusts the TLS certificate it presents.

Due to that, I'm closing this PR, but feel free to open it again if you see that a lot of users are getting confused by this.

PR #2544, opened 2017-03-29
Note that China and US Gov Cloud are not supported.

kastork:
I may not be in full conformance to the contribution guidelines because the page `CONTRIBUTING.md` contains a broken link to the actual guidelines.
(currently the broken link is https://github.com/docker/docker.github.io/blob/master/opensource/project/who-written-for

### Proposed changes

The US Gov cloud region is not supported in this template.  This
seems like useful info to have before trying.  This PR adds a note
in the prerequisites area and in the FAQ.

### DIFF
--- a/docker-for-aws/faqs.md
+++ b/docker-for-aws/faqs.md
@@ -104,7 +104,7 @@ If you are using the `10.0.0.0/16` CIDR in your VPC. When you create a docker ne
 
 ## Which AWS regions will this work with?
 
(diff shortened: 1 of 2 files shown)


joaofnfernandes:
@kencochrane can you confirm that Docker for AWS doesn't currently support AWS the US Gov region?

kencochrane:
As of Today we do not support AWS Gov Cloud, but we are working on adding support. This change is fine, since it is currently correct. 

We will need to remember to update once we have rolled out support for AWS Gov Cloud.

PR #2545, opened 2017-03-29
Update zfs-driver.md

dhiltgen:
The daemon will not auto-detect - the next step in the instructions explain how to configure it properly.


johndmulhausen:
Thanks, @dhiltgen :)

PR #2546, opened 2017-03-29
Improve bootstrap tabs

mstanleyjones:
- Use data-target instead of href
- Move tab-syncing code from DDC topic that was custom-loading it into site-wide JS so everyone can use it
- Add example for syncing tab groups on the same page

The reason for using `data-target` instead of `href` is that it avoids creating non-working fake links that the link checker will fail due to our weird `<base href>` situation.

Make sure to check out the multi-tab example in `/test/` because I think we will use this a lot!!

### DIFF
--- a/test.md
+++ b/test.md
@@ -232,22 +232,22 @@ You can nest captures within each other to represent more complex logic with Liq
 Here are some tabs:
 
 <ul class="nav nav-tabs">
-  <li class="active"><a data-toggle="tab" href="#tab1">TAB 1 HEADER</a></li>
-  <li><a data-toggle="tab" href="#tab2">TAB 2 HEADER</a></li>
+  <li class="active"><a data-toggle="tab" data-target="#tab1">TAB 1 HEADER</a></li>
+  <li><a data-toggle="tab" data-target="#tab2">TAB 2 HEADER</a></li>
 </ul>
 <div class="tab-content">
   <div id="tab1" class="tab-pane fade in active">TAB 1 CONTENT</div>
   <div id="tab2" class="tab-pane fade">TAB 2 CONTENT</div>
 </div>
 
-You need to adjust the IDs and HREFs to match your use case.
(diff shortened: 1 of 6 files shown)


test.md
```diff
@@ -270,6 +270,38 @@ print it and run it through the `markdownify` Liquid filter. Here's a demo:
 </div>
 </div>
 
+#### Synchronizing multiple tab groups
```
joaofnfernandes: Thanks for documenting this! I should have done it when I first created it 😅
Looking back at the script, I think we can can improve it so that you don't need to use the `data-group` attribute.

Do you think declaring tabs like I'm doing below is easier than what we have today?

```
<ul class="nav nav-tabs">
  <li class="active"><a data-toggle="tab" data-target="#go">Go</a></li>
  <li><a data-toggle="tab" data-target="#python">Python</a></li>
</ul>
...
<ul class="nav nav-tabs">
  <li class="active"><a data-toggle="tab" data-target="#go-1">Go</a></li>
  <li><a data-toggle="tab" data-target="#python-1">Python</a></li>
</ul>
...
```

Instead of using the `data-group` attribute, you would use `#go` in the first name, and then `#go-1`, `#go-2`...
This is a bit less explicit, so what do you think?

mstanleyjones: No, I think using `data-group` is better because that's what it is intended for. Later on we will have some more universal data groups that users will be able to set cookies to prefer (like always show me python if it's available).

joaofnfernandes: We discussed this a bit yesterday. The `data-group` thing is my invention, and not really something that bootstrap has out of the box. If you think this is easier to use let's leave it as is and I can tweak it some other time.

mstanleyjones: I think it's dangerous to do string sub-parsing like the other option you outlined. Since Bootstrap doesn't have this built in, we would have had to make something up anyway. :)

PR #2556, opened 2017-03-30
fix missing '-y' parameter

alex-gru:
Fixed a cmd in the intro on `Creating a Dockerfile for MongoDB`.

### Proposed changes
The missing parameter leads to abort when installing `software-properties-common`, since `apt-get install` expects a confirmation.


johndmulhausen:
Thanks, @alex-gru :)

PR #2557, opened 2017-03-30
Update deploy-app.md

barksten:
Removes an extra '.' in the path when copying the docker-stack.yml with scp


johndmulhausen:
Thanks, @barksten :)

PR #2565, opened 2017-03-30
404 improvement follow-up

mstanleyjones:
Fix ordering of 404 page
Add a note to `test.md` about running in-page JS

I tested this by building the Dockerfile locally and it worked for me, but then it worked for me on Netlify before. Let's hope it works this time!

### DIFF
--- a/404.md
+++ b/404.md
@@ -11,7 +11,7 @@ tree: false
 We tried to forward you to where we think you might be going, but we couldn't
 find a good match. You've got some choices:
 
-<div id="sorryMsg" />
+<div id="sorryMsg"></div>
 
 <!-- Script needs to be at the bottom, after any content -->
 <script language="JavaScript">
@@ -34,7 +34,7 @@ if (forwardingURL.indexOf(".md") > -1)
 {% for item in site.data.docsarchive.docker-compose %}
   if (forwardingURL == "/{{ item[0] }}/")
   {
-    console.log("Found via Docker Compose file for Archive")
+    console.log("Found via Docker Compose file for Archive");
     gonnaFwd = true;
     archive = true;
     // make it so redirects cascade; first, use the base URL, then append path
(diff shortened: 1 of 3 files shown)


404.md
```diff
@@ -83,28 +91,13 @@ if (gonnaFwd) {
...
-    output.push("<li><a href='/search/?q=" + phrase + "'</a>Search for <tt>&quot;" + phrase + "&quot;</tt></a></li>");
-    output.push("<li><a href='/search/'>Try a different search</a></li>");
-    output.push("<li><a href='/'>Go to the index</a></li></ul><br />");
-    document.getElementById("sorryMsg").innerHTML = output.join("");
-  }
-  window.onload = populateTicket;
+  console.log("Showing 404 choices");
+  var searchVal = phrase;
+  var output = new Array();
+  output.push("<ul><li><a href='https://github.com/docker/docker.github.io/issues/new?title=404 at: " + path + "&body=URL: " + path + "' class='nomunge'>File a ticket</a></li>");
```
thaJeztah: wondering if we need escaping in these concatenations (`encodeURIComponent()` or similar); will give it a try when it's built, to see if there's no HTML/script injection here (and other parts)

404.md
```diff
@@ -6,6 +6,14 @@ notoc: true
 tree: false
 ---
 
+<br/>
+
+We tried to forward you to where we think you might be going, but we couldn't
+find a good match. You've got some choices:
+
+<div id="sorryMsg" />
```
thaJeztah: Better to change this to

```html
<div id="sorryMsg"></div>
```

http://stackoverflow.com/questions/3558119/are-non-void-self-closing-tags-valid-in-html5

404.md
mstanleyjones: This isn't new code, just migrated from elsewhere, and it WFM ... I did think maybe it needed `&nbsp` treatment, but it seems OK.

404.md
mstanleyjones: hmm.... ok....

mstanleyjones:
I addressed the feedback!

test.md
```diff
@@ -347,6 +347,15 @@ break the Markdown block up. This example does it with Markdown. You can't have
 break the Markdown block up.
 {: style="column-count: 3 "}
 
+## Running in-page Javascript
```
johndmulhausen: This probably belongs in README.MD, no?

mstanleyjones: I'm prepared to be convinced about that. I was not sure where people would go to look for this kind of info, and I think if it were me, I'd look in the `test.md` because I think of it as more of a "how-do-i".

PR #2576, opened 2017-03-31
Update gettingstarted.md

scjane:


johndmulhausen:
Good catch, @scjane thanks

PR #2579, opened 2017-03-31
Update envvars.md

scjane:


johndmulhausen:
Thanks @scjane -- lots of nice fixes here :)

PR #2581, opened 2017-03-31
Clean get Docker

joaofnfernandes:
* Instead of having a massive wall of text about the differences about CE and EE, we can just show it, and create a consistent experience between docs and https://www.docker.com/pricing.
* ~~Removed mentions to stable and edge, because we're already talking about them once you click on the platform where you want to install Docker~~
* Updated mentions to stable and edge to say "channel" instead of "build" to be consistent with the rest of the docs
* Made the tables consistent. CE is always the first column, EE the second
* Removed some mentions to Docker Cloud that were pointing to anchors that no longer existed


engine/installation/index.md
```diff
@@ -8,51 +8,30 @@ redirect_from:
...
-- **Docker Community Edition (Docker CE)** is ideal for developers and small
-  teams looking to get started with Docker and experimenting with
-  container-based apps. Docker CE is available on many platforms, from desktop
-  to cloud to server. Docker CE is available for macOS and Windows and provides
-  a native experience to help you focus on learning Docker. You can build and
-  share containers and automate the development pipeline all from a single
-  environment.
+{% include docker_ce_ee.md %}
 
-  Docker CE gives you the option to run **stable** or **edge** builds.
```
mstanleyjones: You totally removed any mention of Stable and Edge from this page. I don't think that is the right way to go.

joaofnfernandes: > Removed mentions to stable and edge, because we're already talking about them once you click on the platform where you want to install Docker.

I can add it back, but right now we are presenting a ton of options upfront. You need to understand the difference between CE and EE, and if it exists for your platform.
Since edge is not for all users, and once you navigate to the platform we present that choice there, I think we can remove a bit of the complexity here.

mstanleyjones: I'd like @johndmulhausen and @friism to weigh in.

londoncalling:
@joaofnfernandes @mstanleyjones I think this is a good place to describe `stable` and `edge`, where you are providing an overview of installers / supported OS's and CE, EE. @friism is probably a the best judge, but it makes sense to me.

joaofnfernandes:
It's 2 against 1, so I've added back the explanation about stable and edge releases.

mstanleyjones:
Resolved conflicts. Let's see how it works after Netlify finishes.

joaofnfernandes:
Thanks for resolving the conflicts. It seems like we lost some changes in the process, so I've reintroduced them. If you agree with this we should merge since this has been here for a while.

londoncalling:
@joaofnfernandes it isn't building for some reason

joaofnfernandes:
Kicked off Jenkins again. The build could not do a git clone for some reason.

londoncalling:
Cool, thanks

joaofnfernandes:
Let's get this in and we can continue to iterate on subsequent PRs.

londoncalling:
Nice.

PR #2599, opened 2017-04-02
added "docker" to systemd is-active command

shanecastle:
The command was missing "docker" on the end for the systemd example.


johndmulhausen:
So it was. Thanks @shanecastle :)

PR #2600, opened 2017-04-02
Fix coy-paste error

dproteus13:
### Proposed changes

Looks like the description of Stable vs Edge was copied from the page for the Mac version.  One instance of Mac was left, so I fixed that to Windows.


johndmulhausen:
Thanks, @dproteus13 :)

dproteus13:
Not a problem :-D

PR #2620, opened 2017-04-04
Adjust heading level

### LINKED ISSUE #2617: HTML "leaking" in compose file reference
[compose/compose-file/index.md](https://docs.docker.com/compose/compose-file/), CC @londoncalling 

screen shot 2017-04-04 at 1 48 49 pm

mstanleyjones:
Fixes #2617 

Since `toc_max` is 4, the `<h4>` in the table is causing HTML to leak into the right-hand nav. This fixes it by changing that to a level-5 heading.

cc/ @londoncalling


mstanleyjones:
LGTM!!

PR #2629, opened 2017-04-05
Update step_four.md

PlatonicWannabe:
Windows users --like me-- may miss the period, and docker will not find `Dockerfile.txt.`


johndmulhausen:
Thanks, @PlatonicWannabe :)

PR #2631, opened 2017-04-05
There is a colon omitted in this section.

wrfly:
Just a colon omitted.


johndmulhausen:
Thank you @wrfly good catch :)

PR #2662, opened 2017-04-07
Implement browsing by metadata tag w/Glossary integration

johndmulhausen:
As it says. Browse and page and click the "tags" on the bottom, or visit the Glossary tab.


johndmulhausen:
Eh, let's merge and iterate. It's important that we get some visibility into metadata done, because I want to file some bugs to improve it during hackathon.

PR #2666, opened 2017-04-07
Update upgrade-offline.md

scjane:


johndmulhausen:
Thanks, @scjane :)

johndmulhausen:
CC @joaofnfernandes I'm seeing a lot of instances of "login into" when it should simple be "log in to"

PR #2673, opened 2017-04-07
Update restrict-services-to-worker-nodes.md

scjane:


johndmulhausen:
Nice, that's a good change. Thanks, @scjane :)

PR #2676, opened 2017-04-07
Fix enable/disable flags of repos in Fedora

Yajo:
### Proposed changes

Run `man dnf.plugin.config_manager` and you'll see that current docs were wrong.


johndmulhausen:
Ah, indeed. Thanks, @Yajo :)

PR #2685, opened 2017-04-07
Remove redundant feedback links in footer

mstanleyjones:
Now that we have the feedback links in the right-hand TOC, no need to have them in the page footer too.


londoncalling:
LGTM

PR #2698, opened 2017-04-10
Update recover-a-user-password.md

scjane:
1.Is "choose the user you want to change the password" acceptable? 
2.If not, which of the following modifications is better?
   "choose the user whose password you want to change"
   "choose the user you want to change the password for"
   ...


johndmulhausen:
Hi @scjane, "choose the user whose password you want to change" is better.

Also, let's just have it say "log in to" instead of "log directly in to"

Issue #2563, opened 2017-03-30
Admonition styling should be different for info, caution, warning


mstanleyjones:
[test.md](https://docs.docker.com/test/), CC @johndmulhausen 

I'm not sure how to do this, maybe some JS that looks for the words inside the first `<strong>` tag and adds a class to the blockquote based on it? Right now there is no programmatic way to know that they are not just plain notes.

joaofnfernandes:
No JS 🙏

We can add classes to the block quotes like

```
> This is going to make your life miserable
{: .warning}
```

And then create css like

```
blockquote.warning {
  border-left-color: #ce4844;
}
```

Easy peasy. I would also prefer to stick to just 2 types:
* Info: the default so that you don't need to even apply the class
* Warning: this is really bad

This will force us to be careful when using notes so that we don't end up creating a crappy experience like this (not wanting to offend the authors)

<img width="921" alt="screen shot 2017-03-31 at 15 09 46" src="https://cloud.githubusercontent.com/assets/1525937/24571134/2c6e692e-1624-11e7-8565-3ac0e3913a2a.png">

johndmulhausen:
Everything Joao said, basically; I'll have a looksee.

mstanleyjones:
I don't like us having to specify a class as well as the admonition type text itself. Maybe we should have the admonition title automatically added based on the class. I think we should default to Note since that is what we have always done and we have tons and tons of notes.

joaofnfernandes:
If we standardize on specific colors (info/note is always blue, warning is always red) we'll have some consistency so we won't need to start the block quote with `**note**` or `**warning**`. We can do something like the bootstrap docs (see screenshot in my first comment), where they start the note with a title.

Here's a PR to get the ball rolling #2710. [Check the staged demo](https://deploy-preview-2710--docker-docs.netlify.com/test/#admonitions-notes).

johndmulhausen:
Closing as obsolete; I think @londoncalling's latest pass has us in a good place and we should start a new discussion if we want to evolve from here.

PR #2710, opened 2017-04-10
Adds style for warnings and notes

joaofnfernandes:
This PR adds a way for us to have warnings and notes.

screen shot 2017-04-11 at 14 41 23

It also maintains backwards compatibility with the older note style so that we can start migrating slowly (the first note in the screenshot).

### DIFF
--- a/_scss/_notes.scss
+++ b/_scss/_notes.scss
@@ -1,0 +1,35 @@
+/*
+* Infos, notes, and warnings
+*/
+
+$info-color: #1488C6;
+$warning-color: #aa6708;
+$danger-color: #ce4844;
+
+blockquote {
+  border-left-color: $info-color;
+}
+
+blockquote > p:first-child {
+  margin-top: 0;
+}
+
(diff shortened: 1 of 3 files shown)


johndmulhausen:
This is great! Can you do a gold/caution variant for "Be Careful" messages?

joaofnfernandes:
I'd prefer to have just two colors, for the reasons pointed in #2563, but we can create 3 colors and. @jsouth what hex colors should we use for these banners? We need colors for:

* info
* caution
* warning

Right now I'm using the colors of the bootstrap docs but I'd prefer if we had our own colors.

joaofnfernandes:
Now we have 3 colors. You can use regular blockquotes for the default blue, or add the class `warning` for yellow/brown, or `danger` for red.

jsouth:
@joaofnfernandes - I'll collect these intent colors today and post here :)

joaofnfernandes:
Merging. We can iterate the colors of a new PR.

Issue #2691, opened 2017-04-09
Searchability is problematic


gdennie:
[engine/reference/builder.md](https://docs.docker.com/engine/reference/builder/), CC @mstanleyjones 

The site's resident search is limited. Recently tried searching for the term, entrypoint, and was disappointed at the weirdness. Perhaps your site needs a search-able glossary to hold all its concepts in one place.

mstanleyjones:
The very first search result I get from https://docs.docker.com/search/?q=entrypoint is https://www.google.com/url?q=https://docs.docker.com/engine/reference/builder/, which seems to be the right one. The second and third results, which are both on forums.docker.com, also seem to be high quality results. If you want to narrow it down to only docs.docker.com content, you can click the "Docs" tab in the search results.  The top four search results are IMO very high quality.

If you think we should have a definition for Entrypoint in the glossary, I can definitely see an argument for that.

gdennie:
The search results brings up the article and not the definition or synopsis.

________________________________

johndmulhausen:
@gdennie You'll be happy to know we're working on it: https://github.com/docker/docker.github.io/pull/1387/files#diff-8fd00abf6369bddcce9b428356326a1aR13

PR #2712, opened 2017-04-10
Add term to search results page

### LINKED ISSUE #2708: Feature request: show glossary term on search page
### Problem description

If someone searches for a term we have a definition for, it would be cool to show the definition on the search page, maybe right above the search results themselves (since we  can't put it into the search results themselves).

We already have YAML for the glossary, so I think we could do this by saving a `glossary.json` when we build the site (similar to how I generate the archive.json now) and having JS to search through it for the exact term in the `q` parameter of the search.

Positioning the definition might be a little bit tricky. Maybe we can reuse the code from the tag pages for that.

I think this will end up being easy, so I will go ahead and start a PR on it and maybe we can work together.

mstanleyjones:
### Proposed changes

Fixes #2708
Relates to #2691 (but does not fix it, since it is asking for a definition for Entrypoint to be added)

If a given search term (or phrase) exists verbatim in the glossary, show its definition above the search results.

To demo, go to the Netlify preview and search for a term like `docker` or `data volume`, and see that the definition is displayed. Now search for `foo` and see that nothing is displayed.

There is currently some logging info being displayed in the console log for debugging. I'll remove it when this is approved to merge.

I'm open to suggestions on the JS code. It's a little specialized to the search page, in the same way that the glossary definition display is specific to the glossary page. If we can reuse the logic in a better way, I am all for it.

### DIFF
--- a/js/menu.js
+++ b/js/menu.js
@@ -239,10 +239,21 @@ function renderTopicsByTagTable(tagToLookup,divID)
 }
 function renderTagsPage()
 {
-  if(window.location.pathname.indexOf("/glossary/")>-1)
+  if(window.location.pathname.indexOf("/glossary/")>-1 || window.location.pathname.indexOf("/search/")>-1)
   {
-    var tagToLookup = decodeURI(queryString().term);
-    $("#keyword").html(tagToLookup);
+    var tagToLookup;
+    if (window.location.pathname.indexOf("/glossary/")>-1)
+    {
+      // Get ?term=<value>
+      tagToLookup = decodeURI(queryString().term);
+      $("#keyword").html(tagToLookup);
+    }
+    else
+    {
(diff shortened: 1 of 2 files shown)


mstanleyjones:
Bug in the current implementation: If you are on the search page and you do another search, it does not update the search page URL accordingly, so the new term won't display a definition.

mstanleyjones:
@johndmulhausen and I worked on this and we got it working better. Now you just use the search field in the header, even if you are on the search page. If there is a definition for your search term, it is shown. Magic!

mstanleyjones:
Now the search field should retain focus if you are still on the search page and typed in a new search term.

johndmulhausen:
LGTM; great work, @mstanleyjones :)

PR #2725, opened 2017-04-11
Updating table to remove extra closing table row

beenanner:
### Proposed changes

Fix for table markup showing on: https://docs.docker.com/registry/storage-drivers/gcs/


johndmulhausen:
Thanks, @beenanner great catch

PR #2726, opened 2017-04-11
Fix table formatting and endpoint template examples

beenanner:
### Proposed changes

Fix for closing table showing on https://docs.docker.com/registry/storage-drivers/oss/ and endpoint example missing template values due to html being parsed.


johndmulhausen:
Thanks, @beenanner!

PR #2731, opened 2017-04-11
There's no cow, there's a whale

nedbat:
### Proposed changes

The instructions mention a cow, but there's no cow, there's a whale.


johndmulhausen:
Fair enough, not sure how that happened. Thanks, @nedbat :)

Issue #2720, opened 2017-04-10
Version selector points to broken links if you're seeing the edge docs


joaofnfernandes:
![example](https://cloud.githubusercontent.com/assets/1525937/24886879/bb5082b8-1e0c-11e7-909c-ba8bcc9781ae.gif)

johndmulhausen:
Closing as obsolete

PR #2736, opened 2017-04-12
Go back to small /edge

mstanleyjones:
### Proposed changes

Rather than having two whole versions of the site, one `edge` and one `stable`, just have the one version, but have a `/edge/` subdirectory with its own index, CLI reference stubs, and API reference. In the Edge CLI reference stubs, have an admonition that they are edge and give a link back to the non-Edge one.

A few new files are fetched in the Dockerfile and put into the right places in `/edge/`.

So there is an entry for `/edge/` in the left-hand TOC, and there are links to the edge CLI and API refs near the stable ones in the Reference tab.

This PR also pulls out the no-longer-used `glossary.md` from the Dockerfile.

I retained the `Try the Edge docs` link in the tab bar, for visibility.

Netlify will get you some of the way there, but to really test, you need to build the Dockerfile and run locally.


mstanleyjones:
Fixes #2720

mstanleyjones:
OK, I popped off the top commit. Merge at will.

johndmulhausen:
Thanks, @mstanleyjones :)

PR #2743, opened 2017-04-12
Remove "Got Docker?" from index.md

demahum:
### Proposed changes

The section "Got Docker?" with the link is not needed since, whatsoever, next step in the tutorial is the link that this section is pointing to. If there is no this section, user will follow the tutorial and come across the part for downloading and installing Docker. If this section is present (as it is now), user who clicks the link will skip everything till the end of the page and move to the next one.


johndmulhausen:
This topic no longer exists. Thanks, @demahum :)

demahum:
Of course @johndmulhausen. Thanks for the answer. :)

PR #2756, opened 2017-04-12
Fix problem with topic title

mstanleyjones:
This page is taking a long time to load and I can only guess it's because of the colon in the title.


mstanleyjones:
Red herring.

PR #2757, opened 2017-04-13
Default highlight.js off, dockerfile > conf highlighting, pygments

johndmulhausen:
## Changes

- highlight.js is disabled unless `highlightjs: true` is enabled in front-matter
- import of all pygments CSS files, which work with Rouge lexer
- change of `dockerfile` to `conf`, which works fine, syntax-highlighting-wise
- Removal of excess inactive scrollbars, e.g. https://docs.docker.com/engine/examples/running_redis_service/
- Using slightly modified perldoc.css file from pygments, nice mild color scheme (https://github.com/richleland/pygments-css)

## Why? 

- Fixes terrible JS lag, especially on pages such as this: https://docs.docker.com/compose/aspnet-mssql-compose/ which is killing browser tabs due to syntax collision between C# and highlight.js's regex
- Better highlighting in general


_scss/_content.scss
```diff
@@ -129,10 +129,9 @@ pre {
...
     line-height: 1.42857143;
     color: #333;
-    word-break: normal!important; 
-    word-wrap: break-word; 
+    word-break: normal!important;
+    word-wrap: break-word;
     background-color: #f5f5f5;
     border: 1px solid #ccc;
     border-radius: 4px;
-    overflow-x: scroll!important;
```
mstanleyjones: What will happen now if we have long code lines? We don't want them to wrap because it can cause trouble in pasting.

_scss/_content.scss
```diff
@@ -129,10 +129,9 @@ pre {
     font-size: 13px;
     line-height: 1.42857143;
     color: #333;
-    word-break: normal!important; 
-    word-wrap: break-word; 
+    word-break: normal!important;
```
mstanleyjones: You should configure atom not to change white space in lines you didn't otherwise touch.

css/pygments/autumn.css
```diff
@@ -0,0 +1,66 @@
+.highlight .hll { background-color: #ffffcc }
```
mstanleyjones: We only need to import the theme we actually use, right?

_scss/_content.scss
johndmulhausen: They get scrollbars; this is not the rule that actually implements them, this is creating double scrollbars (one active, one inactive)

johndmulhausen:
Fix for SQL compose example (page loads with no lag): https://deploy-preview-2757--docker-docs.netlify.com/compose/aspnet-mssql-compose/

Code blocks from test page:
https://deploy-preview-2757--docker-docs.netlify.com/test/#code-blocks

johndmulhausen:
LGTM, merging

PR #2761, opened 2017-04-13
Change mount syntax on windows from //c/ to c:\

joozek78:
### Proposed changes

Changed the syntax of `-v` option for windows version. It doesn't work if you pass `//c/path`, but works if you pass `c:\path`


joaofnfernandes:
@JimGalasyn 
I know we just merged this, but in the context of this tutorial, you're running the `training/webapp` image, which is a Linux-based image. So I think this snippet should be 

```
c:\<path>:/<container path>
```

Because inside the container `c:\` won't exist.

PR #2762, opened 2017-04-13
Update CS Engine redirects

mstanleyjones:
### Proposed changes

If you try to go to `/cs-engine/release-notes/` you get redirected to 1.12 instead of 1.13. This fixes that and also adds a link from the 1.13 release notes to the 1.12 ones for customers who need to find those.

Thanks @pvnovarese for the report.


mstanleyjones:
LGTM, merging. It's important for people to be able to find the release notes.

PR #2777, opened 2017-04-14
Make it apply to everyone, not just Mac users.

abdullah2:
The tutorial is universal. How do we know that user is on the Mac?


johndmulhausen:
Thanks, @abdullah2 :)

PR #2791, opened 2017-04-14
Use interested instead of invested

lvlds:
### Proposed changes

I think that **interested** is better than **invested**, in this case.


johndmulhausen:
Fair enough, thanks, @lvlds :)

PR #2795, opened 2017-04-14
Improve accessibility of notes

### LINKED ISSUE #2784: Improve note accessibility
### Problem description

After #2710, notes are less accessible. We need to add the note type to the note title, so that people with color blindness and other color-differentiating issues can still discern the note severity and type.

One suggestion is to modify the CSS and use the `:before` selector to add the content of either `Note: `, `Important: `, or `Warning: ` before the note title.

Please don't use Javascript to do this, as we already have too much Javascript in the site.

Relates to #2781 

### Problem location

- Other: <DETAILS>

joaofnfernandes:
Extends #2710 to
fix #2784

### DIFF
--- a/_scss/_notes.scss
+++ b/_scss/_notes.scss
@@ -12,24 +12,45 @@ blockquote {
 
 blockquote > p:first-child {
   margin-top: 0;
+  font-weight: 700;
+  color: $info-color;
 }
 
-blockquote > p > strong {
-  color: $info-color;
+blockquote > p:first-child::before {
+  content: 'Note: ';
 }
 
 blockquote.warning {
   border-left-color: $warning-color;
 }
 
-blockquote.warning > p > strong {
+blockquote.warning > p:first-child {
   color: $warning-color;
+}
+
+blockquote.warning > p:first-child::before {
+  content: 'Important: ';
 }
 
 blockquote.danger {
   border-left-color: $danger-color;
 }
 
-blockquote.danger > p > strong {
+blockquote.danger > p:first-child {
   color: $danger-color;
+}
+
+blockquote.danger > p:first-child::before {
+  content: 'Warning: '
+}
+
+/* Maintain backwards compatibility with old
+*  note style
+*/
+blockquote > p:only-child {
+  font-weight: inherit;
+  color: inherit;
+}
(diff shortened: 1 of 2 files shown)


joaofnfernandes:
I don't understand the feedback. Can you detail or add a screenshot?

mstanleyjones:
When I checked in /test in  netlify I did not see the "Important:" on the important note. Let me double check.

mstanleyjones:
https://deploy-preview-2795--docker-docs.netlify.com/test/#images

_scss/_notes.scss
```diff
@@ -12,24 +12,45 @@ blockquote {
...
   border-left-color: $warning-color;
 }
 
-blockquote.warning > p > strong {
+blockquote.warning > p:first-child {
   color: $warning-color;
 }
 
+blockquote.warning > p:first-child::before {
+  content: 'Be careful: ';
```
mstanleyjones: Oh I see. Is this the prefix we want to use?

joaofnfernandes:
It the PR as is right now, you have 3 messages if you use two paragraphs in the blockquote style

screen shot 2017-04-18 at 14 13 30

_scss/_notes.scss
joaofnfernandes: I'm using the `warning` and `danger` classes, since other front-end frameworks like bootstrap use classes like that with the same semantics.

mstanleyjones:
Yes, I just don't like the `Be careful`. I do not think it aligns with admonition conventions that readers will have learned to understand, which tend to use words like `Important` or `Caution` or `Info` to denote severity level. I don't think we need to buck well-established trends.

joaofnfernandes:
@mstanleyjones I know that there are a few docs out there that converged in a common terminology  but I don't think they are well established trends. They are mostly conventions developed by older docsets. If you look at the new kids in the block, they have very few notes/warnings, and tend to only have a single style for them.

I'm shooting for a compromise between fun and accessibility. We're the fun side of enterprise 🐳

joaofnfernandes:
@mstanleyjones the feedback was addressed. If you're ok merge this and label #2784 as counted.

PR #2797, opened 2017-04-15
Grammar and formatting check

lvlds:
### Proposed changes

- If the first word starts with a vowel sound you use an. If the first word has a consonant sound use a.
- Removing backticks because probably the text needs to be bold and not marked as code, to make **/usr/bin** bold as seen in item 8 in Task 3.


johndmulhausen:
Thanks, @lvlds :)

Issue #2627, opened 2017-04-04
Cleanup Items Round Dos


jsouth:
### Round Dos
original task list for launch: https://github.com/docker/docker.github.io/issues/2052

- [x] Mobile type fixes for small devices (currently "samples" link busts at the seams and floats to secondary line on iphone 4 and other very small viewports)
- [x] Window Rescale leakage between 400-800px
- [x] No longer any difference between notes, warnings, info admonitions (all blockquotes are blue)(Carried over from original fix list)
- [x] Keyboard highlighting in search autocomplete: Functional,  but needs refined: Padding issues
- [x] Create custom class for inline images (display prop)
- [x] Safari overflow issues (full sweep of safari)

joaofnfernandes:
Some pages are overflowing the viewport. Check #2730

PR #2804, opened 2017-04-16
#2627 - Right Sidebar

jsouth:
There are more issues in #2627 in progress. However this resolves the vital sidebar issue on window rescale.

- Fix right sidebar overflow on rescale
- Resolve version select dropdown overflow issue.

http://recordit.co/SrnIpxajcg

### DIFF
--- a/_scss/_navigation.scss
+++ b/_scss/_navigation.scss
@@ -3,6 +3,10 @@
 * Primary navigation - top header **********************************************************
 *
 */
+.nav-container {
+    overflow: hidden;
+    height: 50px;
+}
 
 
 .primary {
@@ -58,7 +62,9 @@
 
 .nav-secondary {
     background: $bg-secondary;
+    box-shadow: 0 1px 13px rgba(0, 0, 0, 0.24);
     height: 55px;
+    overflow: hidden;
 }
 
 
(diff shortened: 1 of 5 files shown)


johndmulhausen:
Seems much better in Netlify. Thanks, @jsouth :)

_scss/_navigation.scss
```diff
@@ -58,7 +62,9 @@
 
 .nav-secondary {
     background: $bg-secondary;
+    box-shadow: 0 1px 13px rgba(0, 0, 0, 0.24);
     height: 55px;
+    overflow: hidden;
```
Alex-CS: Adding `overflow: hidden;` here and in `.nav-container` makes the version select menu unusable (because almost all of it is hidden). 
hidden-version-dropdown
@jsouth I've only ever tried to use that menu after this commit, so I'm not familiar with the "version select dropdown overflow issue" you mentioned in the PR, but what I'm seeing seems like a new issue caused by this.

If this isn't the right place to bring this to your attention, please direct me to the appropriate place.

jsouth: @Alex-CS - Thanks for locating this! Will get this resolved asap!

PR #2805, opened 2017-04-16
Update overview.md

padoo:
### Proposed changes

It should say 'predecessors', not 'successors'. Otherwise the sentence and the subsequent example don't make sense.


johndmulhausen:
Thanks, @padoo :)

PR #2814, opened 2017-04-17
Make install instructions for Docker Compose clearer

### LINKED ISSUE #2435: Compose install instructions for Linux are unclear
[compose/install.md](https://docs.docker.com/compose/install/)

it should be mentioned, that docker-compose can be installed with `sudo apt install docker-compose` on ubuntu. 

After reading this article (and several other) I assumed you always had to build it from source, because for some reason it isn't in the official repositories.

shubheksha:
### Related issues (optional)
Fixes #2435

### DIFF
--- a/compose/install.md
+++ b/compose/install.md
@@ -18,14 +18,15 @@ To install Compose, do the following:
 
     * [Other systems](/engine/installation/index.md){: target="_blank" class="_"}
 
-2.  The Docker Toolbox installation includes both Engine and Compose, so Mac and
-    Windows users are done installing. Others should continue to the next step.
+2.  Docker for Mac, Docker for Windows, and Docker Toolbox installs Docker Compose as part of those suites
+    along with Docker Engine and Docker Machine, so Mac and Windows users are done installing.
+    Others should continue to the next step.
 
-3.  Go to the
+3.  The binaries for Compose are located at the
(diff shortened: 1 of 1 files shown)


bhavin192:
I think installing `docker-compose` using `apt-get` will install some old version of compose instead of latest one.

joaofnfernandes:
@bhavin192 is right. Currently `apt-get install` installs compose 1.5, which is pretty hold. I'm placing this PR on hold until @shin- weights on this.

mstanleyjones:
The correct way to install Docker Compose on Linux is by following the instructions here: https://github.com/docker/compose/releases

compose/install.md
```diff
@@ -59,6 +60,14 @@ To install Compose, do the following:
 
 ## Alternative install options
 
+### Install using apt-get on Ubuntu
+
+Compose can be installed on Ubuntu using the following command:
+
+```bash
+$ sudo apt-get install docker-compose
+```
```
dnephin: I don't think we should suggest this method of install. The package is not provided by Docker, and it is not well maintained (version is version old).

shubheksha:
Removed the section on install instructions for Ubuntu.

johndmulhausen:
Thanks to @joaofnfernandes, @mstanleyjones, @dnephin, and @bhavin192 for the reviews. Thanks to @shubheksha for the PR!

PR #2817, opened 2017-04-17
Add the --fail to `curl` commands when installing compose

### LINKED ISSUE #1448: Add --fail option to curl examples
[compose/install.md](https://docs.docker.com/compose/install/)

The Curl example should include `--fail` to cause a hard failure if the remote server doesn't return an HTTP 200.

shubheksha:
### Related issues (optional)
Fixes #1448

### DIFF
--- a/compose/install.md
+++ b/compose/install.md
@@ -34,7 +34,7 @@ To install Compose, do the following:
     The following is an example command illustrating the format:
 
     ```bash
-    $ curl -L "https://github.com/docker/compose/releases/download/1.11.2/docker-compose-$(uname -s)-$(uname -m)" -o /usr/local/bin/docker-compose --fail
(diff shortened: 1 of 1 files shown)


compose/install.md
```diff
@@ -34,7 +34,7 @@ To install Compose, do the following:
     The following is an example command illustrating the format:
 
     ```bash
-    $ curl -L "https://github.com/docker/compose/releases/download/1.11.2/docker-compose-$(uname -s)-$(uname -m)" -o /usr/local/bin/docker-compose
```
joaofnfernandes: Leaving the `--fail` flag at the end makes it a bit difficult to read the command. Can you move it to the beginning, to be like `curl -L --fail ...`

PR #2821, opened 2017-04-17
Add note about --all option being the default behavior for `docker-compose rm`

### LINKED ISSUE #2281: Remove -a flag from reference doc
[compose/reference/rm.md](https://docs.docker.com/compose/reference/rm/)

Documentation refers to an optional flag `-a`.
> Options:
    -a, --all     Also remove one-off containers created by
                  docker-compose run

But when running latest (docker-compose version 1.11.2, build dfed245) `docker-compose rm` with the mentioned flag a warning is displayed:

```
WARNING: --all flag is obsolete. This is now the default behavior of `docker-compose rm`
```

Documentation should be updated to reflect the new behaviour.

shubheksha:
### Related issues (optional)
Fixes #2281

### DIFF
--- a/compose/reference/rm.md
+++ b/compose/reference/rm.md
@@ -31,3 +31,6 @@ Going to remove djangoquickstart_web_run_1
 Are you sure? [yN] y
 Removing djangoquickstart_web_run_1 ... done
 ```
+>**Note**: 
+> The `--all` or the `-a` option is obselete. It's now the default behavior of
+`docker-compose rm`.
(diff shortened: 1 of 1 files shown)


shubheksha:
Should the option be removed altogether from the doc?

mstanleyjones:
Yes, I think a better solution is to remove it from the doc. Maybe document instead how to show only a sub-set.

joaofnfernandes:
@shin- I don't see this mentioned in the Compose release notes on github. Can you add it to the relevant release?

shin-:
@joaofnfernandes That was back in 1.8.0:  https://github.com/docker/compose/blob/master/CHANGELOG.md#180-2016-06-14

PR #292, opened 2016-10-20
Update debian.md

emalloy:
### Update debian.md

Add dirmngr , required for gnupg, setting up publickey entries in apt

Otherwise, the `apt-key` command fails:

```
# apt-key adv --keyserver hkp://p80.pool.sks-keyservers.net:80 --recv-keys 58118E89F3A912897C070ADBF76221572C52609D
Executing: /tmp/tmp.3nIvpf0TjG/gpg.1.sh --keyserver
hkp://p80.pool.sks-keyservers.net:80
--recv-keys
58118E89F3A912897C070ADBF76221572C52609D
gpg: failed to start the dirmngr '/usr/bin/dirmngr': No such file or directory
gpg: connecting dirmngr at '/tmp/tmp.3nIvpf0TjG/S.dirmngr' failed: No such file or directory
gpg: keyserver receive failed: No dirmngr
```

PR #2823, opened 2017-04-17
[Debian] add gnupg2 requirement for stretch

bamarni:
Unlike jessie, debian stretch doesn't include `gnupg` by default.

This was already tackled with #292 but forgot during a refactoring (50e75ff936200c50462c8135c0ae6d1baddc2e3e).


johndmulhausen:
Thanks, @bamarni :)

PR #2825, opened 2017-04-17
Fix password mismatch root vs wordpress user

### LINKED ISSUE #1667: Fix for wordpress password issue
[compose/wordpress.md](https://docs.docker.com/compose/wordpress/)

Quickstart fails if I set MYSQL_ROOT_PASSWORD to be different from MYSQL_PASSWORD (as one ought to for security purposes)

This is resolved by defining WORDPRESS_DB_USER in the wordpress environment to match MYSQL_USER in the db environment.

johnharris85:
### Proposed changes

Fixes #1667


johnharris85:
The conflict is with another PR fixing the same issue merged previously. They solve the same thing so I'll close this PR :)

PR #2828, opened 2017-04-17
Corrects running_ssh_service examples sshd_config to be image agnostic

### LINKED ISSUE #1055: Corrections to running_ssh_service.md
[engine/examples/running_ssh_service.md](https://docs.docker.com/engine/examples/running_ssh_service/)

The following RUN line should be changes:
    from
RUN sed -i 's/PermitRootLogin prohibit-password/PermitRootLogin yes/' /etc/ssh/sshd_config
    to
RUN sed -i 's/PermitRootLogin [a-zA-Z-]*/PermitRootLogin yes/' /etc/ssh/sshd_config

thus making it agnostic of image setting...

jadametz:
### Proposed changes

Fixes #1055

### DIFF
--- a/engine/examples/running_ssh_service.Dockerfile
+++ b/engine/examples/running_ssh_service.Dockerfile
@@ -3,7 +3,7 @@ FROM ubuntu:16.04
 RUN apt-get update && apt-get install -y openssh-server
 RUN mkdir /var/run/sshd
 RUN echo 'root:screencast' | chpasswd
-RUN sed -i 's/PermitRootLogin prohibit-password/PermitRootLogin yes/' /etc/ssh/sshd_config
+RUN sed -i 's/PermitRootLogin [a-zA-Z-]*/PermitRootLogin yes/' /etc/ssh/sshd_config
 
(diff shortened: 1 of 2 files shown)


thaJeztah:
Hm, I'm not entirely sure we should do this; this is just an example that uses Ubuntu to illustrate the basic concept. It's not a general example to setup sshd

joaofnfernandes:
@thaJeztah the current regex captures
```
PermitRootLogin prohibit-password
```

while this PR makes it a bit more flexible to capture

```
PermitRootLogin prohibit-password
PermitRootLogin no
```
I think this makes the sample future proof and more portable to other distros. What's your main concern in merging this PR? Making that Dockerfile more obscure to new users?

joaofnfernandes:
@mstanleyjones this regex will run in a Dockerfile right after installing sshd, so the config file you're editing is always going to be the default. But given your and @thaJeztah 's feedback I think we should leave the snippet as is and not merge this change.

mstanleyjones:
Not merging for now, but thanks for your work @jadametz

Issue #1850, opened 2017-02-18
Add commands for CentOS installation


Californian:
[engine/installation/linux/centos.md](https://docs.docker.com/engine/installation/linux/centos/), CC: @mstanleyjones
I would suggest adding the following commands to the installation instructions either for engine or compose, at least for CentOS 7.
```
sudo usermod -aG docker <username>
newgrp docker
```
Without the above commands, `docker-compose` fails because of a socket permission issue.

prashant171992:
Raised a PR here : https://github.com/docker/docker.github.io/pull/2830

prashant171992:
@JimGalasyn can u please review this PR ?

mstanleyjones:
This is covered in the post-install instructions which are linked from each of the Linux installation topics. This is common across Linux distributions, and we don't want to duplicate it. Also, it is not strictly required for installation, as you can still run all the Docker commands using `sudo`.

You'll still get the credit, @prashant171992

PR #2830, opened 2017-04-17
Add commands for CentOS installation

prashant171992:
Changes related to https://github.com/docker/docker.github.io/issues/1850

### DIFF
--- a/engine/installation/linux/centos.md
+++ b/engine/installation/linux/centos.md
@@ -216,9 +216,19 @@ Repository set-up instructions are different for [Docker CE](#docker-ce) and
     This command downloads a test image and runs it in a container. When the
     container runs, it prints an informational message and exits.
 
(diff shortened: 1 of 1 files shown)


londoncalling:
@prashant171992 On step 6 in the Netlify preview (https://deploy-preview-2830--docker-docs.netlify.com/engine/installation/linux/centos/#install-using-the-repository) the `bash` markdown element to denote the code block type is showing. 

I think you might be able to fix this simply by adding a return after the preceding step or sentence. For numbered steps you will have to indent the code block to align with the start of the preceding sentence.

mstanleyjones:
As mentioned in #1850, this is not the approach we are going to go with, as adding the `docker` group is an optional post-installation step common to all Linux distributions. See https://docs.docker.com/engine/installation/linux/linux-postinstall/#manage-docker-as-a-non-root-user

PR #2831, opened 2017-04-17
Add different types of volumes to the glossary

### LINKED ISSUE #2732: Add different types of volumes to the glossary
Some source material: https://success.docker.com/KBase/Different_Types_of_Volumes

Don't add these as separate entries, just expand on what's in the existing entry for volumes

shubheksha:
### Related issues (optional)
Fixes #2732

### DIFF
--- a/_data/glossary.yaml
+++ b/_data/glossary.yaml
@@ -290,3 +290,13 @@ volume: |
   delete volumes when you remove a container, nor will it "garbage collect"
   volumes that are no longer referenced by a container.
   *Also known as: data volume*
+
+  There are three types of volumes: *host, anonymous, and named*:
+
+  - A **host volume** lives on the Docker host's filesystem and can be accessed from within the container.  
+
+  - An **anonymous volume** is useful for when you would rather have Docker handle where the files are stored.
+    It can be difficult, however, to refer to the same volume over time when it is an anonymous volumes.
+
+  - A **named volume** is similar to an anonymous volume. Docker manages where on disk the volume is created,
+    but you give it a volume name.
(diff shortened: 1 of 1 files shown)


shubheksha:
@mstanleyjones something went wrong with that PR, created a new one

_data/glossary.yaml
```diff
@@ -289,4 +289,14 @@ volume: |
...
   volumes that are no longer referenced by a container.
-  *Also known as: data volume*
+  It is also known as *data volume.*
+
+  There are three types of volumes: *host, anonymous, and named*:
+
+  - A **host volume** lives on the Docker host's filesystem and can be accessed from within the container.
+
+  - An **anonymous volume** is useful for when you would rather have Docker handle where the files are stored.
+    It can be difficult, however, to refer to the same volume over time when it is an anonymous volumes.
```
mstanleyjones: You should not use anonymous volumes unless you know for a fact that you don't want to reuse the volume. It's an implementation detail that anonymous volumes have randomly-generated names.

_data/glossary.yaml
```diff
@@ -289,4 +289,14 @@ volume: |
...
+
+  - A **named volume** is similar to an anonymous volume. Docker manages where on disk the volume is created,
```
mstanleyjones: I would actually defined a named volume first, and define an anonymous volume in terms of the named volume.

Actually, I think maybe you can reuse some content from https://success.docker.com/KBase/Different_Types_of_Volumes.

shubheksha: Hey, this is linked in the original issue as well and that's what I used to add the info. I tried to make some changes based on what you suggested and what I understood, PTAL.

shubheksha:
@mstanleyjones, I reworded it and changed the order based on what I understood. PTAL.

mstanleyjones:
You need to remove the unrelated merge commit which is adding commits from another of your PRs here.

shubheksha:
Fixed it.

Issue #2607, opened 2017-04-03
error or other example?


jjmonsalveg:
[engine/getstarted/step_four.md](https://docs.docker.com/engine/getstarted/step_four/), CC @mstanleyjones 
Hello!, in Step 2 and 4 seem to be the result of the same run but the 3 seems to be another run.

Step 2: Successfully built c2c3152907b5
Step 3: Successfully built success 7d9495d03763

Step 4 (docker images):
Docker-whale latest c2c3152907b5 4 minutes ago 275.1 MB

It is not clear whether step 3 is a sequence of 2 or it is a separate example

jpooler:
#dibs for the hackathon

johndmulhausen:
This topic no longer exists. @jpooler still gets the points for #2832.

PR #2832, opened 2017-04-17
Change image IDs to indicate the same run.

bhavin192:
As mentioned in #2607 the image IDs should be same in order to avoid confusion in step 2,3,4, cause the whole text is explaining the one run of `docker build`


johndmulhausen:
This topic no longer exists. But don't worry, you'll get your points. Thanks @bhavin192 :)

Issue #2130, opened 2017-03-06
404 in package download engine links table


mbentley:
[cs-engine/1.13/index.md](https://docs.docker.com/cs-engine/1.13/)

Under the table of RPMs, the `docker-selinux` link is incorrect.  It has an `11` instead of `1` in the URL.

`RHEL 7.x and CentOS 7`	`docker-engine, docker-engine-debuginfo, docker-engine-selinux`

https://packages.docker.com/1.13/yum/repo/main/centos/7/Packages/docker-engine-selinux-1.13.1.cs2-11.el7.centos.noarch.rpm
should be
https://packages.docker.com/1.13/yum/repo/main/centos/7/Packages/docker-engine-selinux-1.13.1.cs2-1.el7.centos.noarch.rpm

scjane:
#dibs

johndmulhausen:
Sorry @scjane, looks like you claimed this but never submitted a PR. @bhavin192 jumped on this one.

scjane:
@johndmulhausen OK. It is because when I was about to submit my PR, I saw @bhavin192 had already submitted one. So, I think it's breaking rules if I submit mine again. Anyway, this issuse is resolved.

bhavin192:
@scjane I was not aware of dibs, Sorry about that.

scjane:
@bhavin192 Never mind. Forget it🙂

londoncalling:
Points to @bhavin192 on this one.

PR #2835, opened 2017-04-17
Fix the typo in package link 

bhavin192:
Fixes the #2130


johndmulhausen:
@bhavin192 Thanks! Be mindful of when people call dibs on bugs though. But, @scjane claimed bug #2130 and never submitted a PR, so I guess you get this one. :)

bhavin192:
@johndmulhausen I would take care of that from next time. I was not aware of dibs while I was sending PRs.

PR #2838, opened 2017-04-17
Removed 'Request docs change' link from 404 page.

### LINKED ISSUE #2783: On the 404 page, the "Request Docs Changes" link should include the referrer, not 404.md, in the report
If you click Request Docs Changes on the 404 page, it reports that the error is in 404.md. This may be the case but it is much more likely that someone wants to report a 404 and has not realized that there is a link to do that in the 404 page itself. You can try it by going to https://docs.docker.com/foo and clicking "Request docs changes" and comparing its behavior to if you clicked "File a ticket" in the same page.

The fix is either to disable the Report docs  changes link on the 404 page (probably bad UX) or to make it  behave the same as the File an issue link on the same page. Either way it needs a Liquid if/then kind of thing in `_layouts/docs.html`.

prashant171992:
Removed 'Request docs change' link from 404 page. 
Issue : https://github.com/docker/docker.github.io/issues/2783

Fixes #2783

### DIFF
--- a/_layouts/docs.html
+++ b/_layouts/docs.html
@@ -9,6 +9,8 @@
 else %}{% assign edit_url = "" %}{% endif %} {% break %} {% endif %} {% endfor %}
 
 <!-- End of logic for 'edit this button' -->
+
+
 <!DOCTYPE html>
 <html lang="en">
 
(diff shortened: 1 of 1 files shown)


bhavin192:
Removing the link from the _docs.html_ removes it from all pages (it's layout html not 404). This means no option to Request docs changes on any page.

prashant171992:
@bhavin192 made the check for 404 page in layout html , please review.

bhavin192:
What if the link to _request docs changes_ creates new issue with the link on which user gets 404?

prashant171992:
@bhavin192 didnt get u , I have removed the 'Request docs change' link from 404 page.  So it will not appear on a 404 page. There was an option to  disable the 'Report docs change' link on the 404 page in the requirement. So i did it . Please lemme know if I missed something.

bhavin192:
I just want to say that keep that _request docs changes_ 
If user gets 404 on /foo, create new issue for /foo (Same as file an issue)

> The fix is either to disable the Report docs changes link on the 404 page (probably bad UX) or to make it behave the same as the File an issue link on the same page.

prashant171992:
@mstanleyjones made the change.

prashant171992:
@mstanleyjones fixed , please check.

mstanleyjones:
You've worked hard on this and I am going to give you the points, but I'm going to push into this PR to show you how to do it in Liquid, and not use JS. Give me a few minutes.

mstanleyjones:
Actually, you are right, and I did it in JS even in the 404.md page! 😨 So your fix is great!

prashant171992:
@mstanleyjones I was struggling to find the page URL (/foo) using liquid, so used JS finally.

Issue #2634, opened 2017-04-05
Formatting messed up with button on Docker for AWS release page


ManoMarks:
[docker-for-aws/release-notes.md](https://docs.docker.com/docker-for-aws/release-notes/), CC @johndmulhausen 
![screen shot 2017-04-05 at 9 17 29 am](https://cloud.githubusercontent.com/assets/113855/24715578/c074656e-19e0-11e7-8a26-40ec41de98d1.png)

scjane:
#dibs

johndmulhausen:
@scjane Careful about claiming bugs and not submitting PRs! @stevenhanna6 got in there. Only claim bugs when you are actively working on them -- as close to one-at-a-time as you can. It jams up the works, otherwise. 

@stevenhanna6 gets the points here.

PR #2846, opened 2017-04-18
Formatting messed up with button on release pages

stevenhanna6:
### Proposed changes

/docker-for-aws/release-notes.md
/docker-for-azure/release-notes.md

Both have wrapping issues with text after buttons.

/_includes/d4a_buttons.md is used to insert the buttons on those pages.

Because the anchor tag uses the button class it applies a float:left and causes wrapping issues after. 

A simple div tag with class clearfix can fix this. The buttons are also used inside a couple tables, but there is no need for float: left in those cases and the div causes no issues.

### Related issues (optional)
Fixes: #2634


johndmulhausen:
Much better, @stevenhanna6 -- thank you!

PR #2850, opened 2017-04-18
Change gerund form of verbs (-ing) to non-gerund form in join-nodes.md

### LINKED ISSUE #2082: Change gerund form of verbs (-ing) to non-gerund form in topic
[engine/swarm/join-nodes.md](https://docs.docker.com/engine/swarm/join-nodes/)
At the beginning of this doc:

> To take full advantage of swarm mode you can add nodes to the swarm:
> 
> - Adding worker nodes increases capacity. xxx
> - Manager nodes increase fault-tolerance. xxx
> 

Will it be better if the two items in the list using parallelism, such as:

To take full advantage of swarm mode you can add nodes to the swarm:

 - Adding worker nodes increases capacity. xxx
 - Adding manager nodes increases fault-tolerance. xxx

Or:

To take full advantage of swarm mode you can add nodes to the swarm:

 - Worker nodes increase capacity. xxx
 - Manager nodes increase fault-tolerance. xxx

Only my ideas. Perhaps it's unnecessary to change as sentence flexibility helps make a doc, even a technical doc, have interesting qualities. But sometimes it's difficult for me to find a balance. Look forward to your suggestions. Thanks.

scjane:
Fix #2082

### DIFF
--- a/engine/swarm/join-nodes.md
+++ b/engine/swarm/join-nodes.md
@@ -5,13 +5,13 @@ title: Join nodes to a swarm
 ---
 
 When you first create a swarm, you place a single Docker Engine (Engine) into
-swarm mode. To take full advantage of swarm mode you can add nodes to the swarm:
+swarm mode. To take full advantage of swarm mode, you can:
 
-* Adding worker nodes increases capacity. When you deploy a service to a swarm,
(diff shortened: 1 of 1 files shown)


johndmulhausen:
Thanks, @scjane :)

Issue #2801, opened 2017-04-16
Document what --memory-swap=0 means


jamshid:
[engine/admin/resource_constraints.md](https://docs.docker.com/engine/admin/resource_constraints/), CC @mstanleyjones 

The latest docs https://docs.docker.com/engine/admin/resource_constraints/#/memory-swap-details do not describe the effect of using `--memory-swap=0`. The default is `-1` which means unlimited, and using a positive integer (i.e. >1) is defined. 

Does `0` mean no swap will be allowed, or do I have to set `--memory` and `--memory-swap` to the same value to accomplish that?

Filing here as recommended in https://github.com/docker/docker/issues/30716.

joeuo:
I'll take this one up!

johndmulhausen:
Points to @joeuo

PR #2851, opened 2017-04-18

pareshdkc:
### Proposed changes

--memory-swap=""   :	Total memory limit (memory + swap, format: <number>[<unit>])
So here if we do not want to allow any swap, we have to keep swap value in above to 0 which means --memory-swap value becomes --memory value.


mstanleyjones:
Does this relate to a specific GitHub issue?

johndmulhausen:
@mstanleyjones Could be addressing #2801

pareshdkc:
yes..!! it is for issue #2851

mstanleyjones:
Closing for now. Please reopen if you can provide more context.

PR #2852, opened 2017-04-18
Typo on raft.md page

### LINKED ISSUE #2205: Typo on raft.md page
[engine/swarm/raft.md](https://docs.docker.com/engine/swarm/raft/)

For the sentence:

> The existing tasks will keep running but the scheduler will not be able to rebalance tasks to cope with failures if when the manager set is not healthy.

Should it be "if or when"?

scjane:
Fix #2205


johndmulhausen:
Thanks, @scjane :)

PR #2854, opened 2017-04-18
describe the effect of using --memory-swap=0

joeuo:
Signed-off-by: Zhiguo Deng <bjzgdeng@linux.vnet.ibm.com>

### Related issues (optional)
#2801

### DIFF
--- a/engine/admin/resource_constraints.md
+++ b/engine/admin/resource_constraints.md
@@ -39,7 +39,10 @@ for [Memory Resource Controller](https://www.kernel.org/doc/Documentation/cgroup
 ### `--memory-swap` details
 
 `--memory-swap` is a modifier flag that only has meaning if `--memory` is also
-set. Swap is disk-based storage, so is usually less performant than memory.
+set. Using swap allows the container to write excess memory requirements to disk
+when the container has exhausted all the RAM that is available to it. There is a
+performance penalty for applications that swap memory to disk often.
+
 Its setting can have complicated effects:
 
 - If `--memory-swap` is set to a positive integer, then both `--memory` and
--- a/engine/admin/resource_constraints.md
+++ b/engine/admin/resource_constraints.md
@@ -51,17 +51,17 @@ Its setting can have complicated effects:
   non-swap memory. So if `--memory="300m"` and `--memory-swap="1g"`, the
   container can use 300m of memory and 700m (`1g - 300m`) swap.
 
-- If `--memory-swap` is unset, and `--memory` is set, the container can use as much swap
-  as the `--memory` setting, if the host container has swap memory configured.
-  For instance, if `--memory="300m"` and `--memory-swap` is not set, the
-  container can use 300m of memory and 300m of swap.
-
 - If `--memory-swap` is set to `0`, the setting is ignored, and the value is
   treated as unset.
 
 - If `--memory-swap` is set to the same value as `--memory`, and `--memory` is
   set to a positive integer, the effect is the same as setting `--memory-swap`
-  to `0`.
+  to `0` (the value is treated as unset).
+
+- If `--memory-swap` is unset, and `--memory` is set, the container can use
+  twice as much swap as the `--memory` setting, if the host container has swap
+  memory configured. For instance, if `--memory="300m"` and `--memory-swap` is
+  not set, the container can use 300m of memory and 600m of swap.
 
 - If `--memory-swap` is explicitly set to `-1`, the container is allowed to use
   unlimited swap, up to the amount available on the host system.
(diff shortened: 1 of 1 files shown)


mstanleyjones:
I'm about to push in an edit over this.

engine/admin/resource_constraints.md
```diff
@@ -38,16 +38,22 @@ for [Memory Resource Controller](https://www.kernel.org/doc/Documentation/cgroup
 
 ### `--memory-swap` details
 
-- If unset, and `--memory` is set, the container can use twice as much swap
+- A limit value equal to memory plus swap. Must be used with `-m (--memory)` flag.
+      The limit should always be larger than `-m (--memory)` value. By default, it is
+      set to double the value of --memory.
```
mstanleyjones: How can this be true and at the same time the default can be `-1` (as stated below), which means unlimited swap memory?

mstanleyjones:
@justincormack has looked into this a bit, I believe. He says it is gross and confusing. :(

engine/admin/resource_constraints.md
justincormack: Yes, one of those statements must be wrong I agree. The kernel docs do not mention the double memory thing https://www.kernel.org/doc/Documentation/cgroup-v1/memory.txt but it may be that Docker does that I need to check the code (its DockerCon and I won't have time today!)

I remember looking at this before and finding it confusing, I think there have been a few issues about it previously, and we should resolve it.

joeuo: @justincormack @mstanleyjones The --unset and set to default (-1) are two different settings. @justincormack I agree we should resolve it and this is based on the man page and my use cases. Maybe check the codes when you have time.

mstanleyjones:
I don't think we are quite there yet with this, but part of the limitation is on the technical review side.

mstanleyjones:
I tried my hand at editing this. @justincormack PTAL

engine/admin/resource_constraints.md
```diff
@@ -38,16 +38,30 @@ for [Memory Resource Controller](https://www.kernel.org/doc/Documentation/cgroup
...
+set. Swap is disk-based storage, so is usually less performant than memory.
+Its setting can have complicated effects:
+
+- If `--memory-swap` is set to a positive integer, then both `--memory` and
+  `--memory-swap` must be set. `--memory-swap` represents the total amount of
+  memory and swap that can be used, and `--memory` controls the amount used by
+  non-swap memory. So if `--memory="300m"` and `--memory-swap="1g"`, the
+  container can use 300m of memory and 700m (`1g - 300m`) swap.
+
+- If `--memory-swap` is unset, and `--memory` is set, the container can use as much swap
```
crosbymichael: will have to check but I thought it was 2 * (memory)

engine/admin/resource_constraints.md
```diff
@@ -38,16 +38,30 @@ for [Memory Resource Controller](https://www.kernel.org/doc/Documentation/cgroup
...
-      For instance, if `--memory="300m"` and `--memory-swap` is not set, the
-      container can use 300m of memory and 600m of swap.
-- If set to a positive integer,  and if both `--memory` and `--memory-swap`
-      are set, `--memory-swap` represents the total amount of memory and swap
-      that can be used, and `--memory` controls the amount used by non-swap
-      memory. So if `--memory="300m"` and `--memory-swap="1g"`, the container
-      can use 300m of memory and 700m (1g - 300m) swap.
-- If set to `-1` (the default), the container is allowed to use unlimited swap memory.
+`--memory-swap` is a modifier flag that only has meaning if `--memory` is also
+set. Swap is disk-based storage, so is usually less performant than memory.
```
crosbymichael: I would refer to swap as "Swap allows the container do write excess memory requirements to disk when it can no longer fit into RAM.  There is a performance penalty for applications that swap memory to disk often."

I'm sure you will word it better but its a little more explanation.

engine/admin/resource_constraints.md
mstanleyjones: I don't see anywhere in the code where it is doubling memory...

crosbymichael: https://github.com/moby/moby/blob/master/daemon/daemon_unix.go#L266

mstanleyjones:
I reworded the part describing what swap is.

engine/admin/resource_constraints.md
mstanleyjones: Thanks, I reworked things to clarify this. This is gnarly! Basically there are three different ways it can be treated as "unset".

johndmulhausen:
Thanks, @joeuo :)

engine/admin/resource_constraints.md
tao12345666333: I send a PR for fix this. https://github.com/docker/docker.github.io/pull/9340 

`MemorySwap` is `Memory` + `Swap`
```
	if hostConfig.Memory > 0 && hostConfig.MemorySwap == 0 {
		// By default, MemorySwap is set to twice the size of Memory.
		hostConfig.MemorySwap = hostConfig.Memory * 2
	}

```

Issue #1528, opened 2017-02-06
Add note that Compose version 3 doesn't support additional IPAM configs


qmcree:
### Problem description

Need to mention that Docker Compose version 3 does not support any additional IPAM configurations except for subnet. See docker/compose#4366.

### Problem location

- I saw a problem on the following URL: https://docs.docker.com/compose/compose-file/#ipam

### Suggestions for a fix

Add note that the configurations are only acceptable for version 2 at the moment.

Rubab-Syed:
I'll take this one.

johndmulhausen:
Points to @Rubab-Syed -- thanks!

PR #2858, opened 2017-04-18
Add note that compose v3 doesnt support additional ipam configs

Rubab-Syed:
### Proposed changes

Adds note as mentioned in #1528


johndmulhausen:
Thanks, @Rubab-Syed :)

PR #2860, opened 2017-04-18
Fix sidebar scrolling

### LINKED ISSUE #2232: Sidebar menu not visible when many items are expanded
Here is a short video that shows the problem

https://screencast.com/t/dtWyUkzDdy

### LINKED ISSUE #2403: Can't scroll down left side sometimes
[index.md](https://docs.docker.com/)
sometimes I can not scroll down the left

allejo:
### Proposed changes

Move away from a JS solution to keep the sidebars in place and use a more robust CSS solution to take advantage of [`position: sticky`](http://caniuse.com/#feat=css-sticky). A polyfill will be necessary to support IE+Edge, which are the only modern browsers lacking support. I've previously used [`wilddeer/stickyfill`](https://github.com/wilddeer/stickyfill) and has worked for me, however I don't know if there's a different one you'd like to explore (modernizr has a polyfill as well).

### Related issues (optional)

Fixes #2403
Fixes #2232


mstanleyjones:
Ping when Netlify finishes and I'll have a look.

allejo:
Sweet, I'll clean/fix this up later today

allejo:
@mstanleyjones does the latest netlify preview fix the issue you were having?

mstanleyjones:
Hmm, not quite yet. We miss the sticky header at the top. Go to https://docs.docker.com and scroll down a bit, and notice that it sticks to the top.

mstanleyjones:
I kicked Netlify.

allejo:
All of my commits have been squashed and cleaned up to have meaningful messages. Netlify has the latest preview of this PR. I've also updated the [first post](https://github.com/docker/docker.github.io/pull/2860#issue-222494521). Feedback and issues welcome 👍

mstanleyjones:
Not quite yet. There is a problem with the affixed header disappearing on mobile. Almost there!!

mstanleyjones:
@johndmulhausen WDYT? I think this is quite an improvement. Are these changes in the right places so that @jsouth won't wipe them out?

allejo:
Since the affixed header behaves the same way as the production site, i reported the mobile navbar issue it in #2938.

stevenhanna6:
All pages
Chrome, Firefox, Edge: vertical scrollbars always visible in left and right sidebars (before they were hidden until needed)

Homepage
Edge: left and right sidebars do not move down when scrolling

Other pages
Chrome, Firefox, Edge: right side bar horizontal scroll bar always visible (was hidden before)
Edge: left and right sidebars do not move down when scrolling

My solution doesn't have these problems but your solution is quite different so I can't offer any help. I guess the edge browser bugs are the main ones to be fixed, the others and more visual differences on how the site behaved before.

allejo:
Fixed scrolling in the overflow in 7d01ffc. IE + Edge doesn't support `position: sticky` yet so I'm waiting on a yay or nay on using a polyfill and which one if it's a yay.

stevenhanna6:
Just checked your scrolling overflow fix, it fixed the scrollbar issues. Great!

Can't wait to see how you fix the Edge bug.

I think your overall your solution is much cleaner then the checkNavSizes() and affix usage. Well done.

johndmulhausen:
Loving the left nav resize logic -- the scrollbars and affix logic is looking good. The problem right now is that we're losing the hamburger menu at the top-right for a large swath of resolutions -- and without that, you can't navigate the site. 

Watch how narrow it has to get before the hamburger menu comes back in the top right, as I gradually narrow the viewport:

img

stevenhanna6:
@allejo your pollyfill trick worked on Edge, looks good to me. Cleared up all the issues I saw

@johndmulhausen looks like that bug is in the original site and wasn't introduced by allejo, there is also a bug I just noticed on the original site. Click the hamburger, close it, then reisize your browser back to get rid of the hamburger...the left nav panel is now blank.

allejo:
@stevenhanna6 woo! glad that works 👍 

@johndmulhausen as Steven pointed out, I'm not to blame for the issue (whew!) but fixed it in this PR none the less

stevenhanna6:
@allejo Just want to confirm same thing you see, just the hompage has that one last hamburger bug.

Everything else works perfectly, if you can't figure out that last bug I would suggest breaking it out into a separate issue so the rest of your work gets looked over and merged. To me it a huge improvement over the original code.

allejo:
I've unified the nav bar on the homepage and the rest of the website with a single `include` in b307c2c; this should fix the missing hamburger on the homepage.

However, the nav bar as a whole really needs some TLC, which would be better suited for its own PR/project.

jsouth:
So awesome to see all of these improvements taking place!

Thank you rockstars for helping us tighten this up.🌟🌟🌟

johndmulhausen:
@allejo This is just excellent work. Thanks so much for being responsive and for helping us with this, it's a big one.

PR #2872, opened 2017-04-18
Clarify docs hack room times

londoncalling:
Clarified days and times for docs hack room


johndmulhausen:
Thanks, @londoncalling :)

PR #2874, opened 2017-04-18
Fix markdown for Docker Machine overview

cs-shadow:
Currently, the formatting for https://docs.docker.com/machine/overview
is broken due to a hanging inline code block.

Fix it by moving the inline code block to a new line.

Signed-off-by: Chandan Singh <csingh43@bloomberg.net>


johndmulhausen:
This is great, thanks @cs-shadow :)

PR #2878, opened 2017-04-18
(#2637) Updated documentation for UCP to clarify roles of nodes in a...

### LINKED ISSUE #2637: UCP architecture - a bit confusing UCP Worker/Manager/Nodes
[datacenter/ucp/2.1/guides/architecture.md](https://docs.docker.com/datacenter/ucp/2.1/guides/architecture/), CC @joaofnfernandes

I feel that there is some confusing terminology in the ucp system architecture documentation
In the first diagram you say that its a UCP is a container based application running on top of docker engine which makes sense.
https://docs.docker.com/datacenter/ucp/2.1/guides/images/architecture-1.svg

The next diagram is very confusing.
What i know a swarm is a set of docker engines running in swarm mode. Some or managers some are workers. Then they say a bit vauge that the "node" is a manager / worker.

UCP Agent has two modes of operation as i understand:
   - Worker Mode
      - Serves docker commands to other ucp services only authorized users.
   - Manager Mode
      - Serves all UCP components including the UCP Web UI and data stores.
    
UCP Worker orange box:
   Is it a representation of the UCP agent working in "UCP Worker" mode ?
UCP Manager green box:
   Is it a representation of the UCP agent working in "UCP Manager" mode ? 
 
I'm tried before to look for the actual "UCP Worker or Manager" component in the container table.... 

// Marten

wcgcoder:
… Docker swarm. Also updated architecture diagram to reflect new Enterprise engine branding.

### Proposed changes

Updated documentation for UCP version 2.1 for Docker Docs Hackathon. https://docs.docker.com/hackathon/

### Related issues (optional)

Fixes #2637

### DIFF
--- a/datacenter/ucp/2.1/guides/architecture.md
+++ b/datacenter/ucp/2.1/guides/architecture.md
@@ -4,8 +4,7 @@ keywords: docker, ucp, architecture
 title: UCP architecture
 ---
 
-Universal Control Plane is a containerized application that runs on the
-Commercially Supported (CS) Docker Engine, and extends its functionality to
+Universal Control Plane is a containerized application that runs on [Docker Enterprise Edition](/enterprise/index.md) and extends its functionality to
 make it easier to deploy, configure, and monitor your applications at scale.
 
 It also secures Docker with role-based access control so that only authorized
@@ -25,9 +24,7 @@ and Docker Compose.
 Docker UCP leverages the clustering and orchestration functionality provided
 by Docker.
 
-<a name="figure-architecture-2">
 ![](images/architecture-2.svg)
-</a>
 
 A swarm is a collection of nodes that are in the same Docker swarm. [Nodes](/engine/swarm/key-concepts.md) in a Docker swarm operate in one of two modes: Manager or Worker. If nodes are not already running in a swarm when installing UCP, nodes will be configured to run in swarm mode.
 
--- a/datacenter/ucp/2.1/guides/architecture.md
+++ b/datacenter/ucp/2.1/guides/architecture.md
@@ -36,7 +36,7 @@ and stops UCP services, based on whether that node is a
 If the node is a:
 
 * **Manager**: the `ucp-agent` service automatically starts serving all UCP
-components including the UCP web UI and data stores used by UCP. The `ucp-agent` accomplishes this by [deploying several containers](#ucp-components-in-manager-nodes) on the node. By promoting a node to manager, UCP automatically becomes highly available and fault tolerant. <br/><br/>
+components including the UCP web UI and data stores used by UCP. The `ucp-agent` accomplishes this by [deploying several containers](#ucp-components-in-manager-nodes) on the node. By promoting a node to manager, UCP automatically becomes highly available and fault tolerant.
 * **Worker**: on worker nodes the `ucp-agent` service starts serving a proxy
 service that ensures only authorized users and other UCP services can run Docker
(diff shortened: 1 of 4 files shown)


datacenter/ucp/2.1/guides/architecture.md
```diff
@@ -25,8 +25,11 @@ and Docker Compose.
 Docker UCP leverages the clustering and orchestration functionality provided
 by Docker.
 
+<a name="figure-architecture-2">
```
joaofnfernandes: You're changing from markdown to html, but what does this html version do that the markdown version doesn't already?
This takes more space and `name` attribute is deprecated in html5, so if this does exactly the same thing as the markdown version, you should revert this.

datacenter/ucp/2.1/guides/images/architecture-2.svg
```diff
@@ -22,8 +22,8 @@
                         </g>
                         <g id="engine" transform="translate(1.000000, 79.000000)">
                             <rect id="Rectangle-138" fill="#1488C6" x="0" y="0" width="95" height="22" rx="2"></rect>
-                            <text id="CS-Docker-Engine" font-family="OpenSans, Open Sans" font-size="10" font-weight="normal" fill="#FFFFFF">
-                                <tspan x="7.26025391" y="15">CS Docker Engine</tspan>
+                            <text id="Docker-Enterprise-Engine" font-family="OpenSans, Open Sans" font-size="8" font-weight="normal" fill="#FFFFFF">
+                                <tspan x="5.26025391" y="15">Docker Enterprise Edition</tspan>
```
joaofnfernandes: Can you change this from "Docker Enterprise Edition" to "Docker EE"? Right now the text is too long and overflowing.

datacenter/ucp/2.1/guides/architecture.md
wcgcoder: IIRC, I did this for something I couldn't find in markdown, but don't remember what that was. Regardless, I removed my text linking to it. Will remove. Will also update the `architecture-1.svg` you mentioned.

datacenter/ucp/2.1/guides/images/architecture-2.svg
wcgcoder: Thanks for responding to the PR, btw.
I'm not seeing an overflow on my system. Maybe it's rendering differently for you. Regardless, I agree and will change to Docker EE.

screen shot 2017-04-20 at 9 44 45 am

datacenter/ucp/2.1/guides/architecture.md
```diff
@@ -36,11 +36,10 @@ and stops UCP services, based on whether that node is a
 If the node is a:
 
 * **Manager**: the `ucp-agent` service automatically starts serving all UCP
-components including the UCP web UI and data stores used by UCP. By promoting
-a node to manager, UCP automatically becomes highly available and fault tolerant.
+components including the UCP web UI and data stores used by UCP. The `ucp-agent` accomplishes this by [deploying several containers](#ucp-components-in-manager-nodes) on the node. By promoting a node to manager, UCP automatically becomes highly available and fault tolerant. <br/><br/>
```
joaofnfernandes: Can you also remove these `<br/>`? If we need more space between list items we can do it with CSS. Doing this will make it harder to restyle the docs because everything has a random number of line breaks.

wcgcoder: Good call. The default listing space was just bothering me 😂

PR #2885, opened 2017-04-19
Issue #1465 Specify that all nodes need to be the same exact Docker v…

### LINKED ISSUE #1465: Specify that all nodes need to be the same exact Docker version for production UCP clusters
[datacenter/ucp/2.0/guides/installation/plan-production-install.md](https://docs.docker.com/datacenter/ucp/2.0/guides/installation/plan-production-install/)

Can we specify on this page that for production UCP clusters, unless a temporary upgrade is occurring, all engines need to be the same x.y.z version number? We have seen several situations where this has caused issues and it should be documented.

constanzekratel:
Fixes #1465

Just made a minor correction as part of fixing this bug. Added "with" to comply and also added "exact" to emphasize the importance of running the same version. Other than that this bug already appears to have been fixed.

### DIFF
--- a/datacenter/ucp/2.0/guides/installation/plan-production-install.md
+++ b/datacenter/ucp/2.0/guides/installation/plan-production-install.md
@@ -14,8 +14,8 @@ deploying Docker Universal Control Plane for production.
 Before installing UCP you should make sure that all nodes (physical or virtual
 machines) that you'll manage with UCP:
 
-* [Comply the system requirements](system-requirements.md)
-* Are running the same version of CS Docker Engine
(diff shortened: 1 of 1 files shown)


joaofnfernandes:
Thanks for the contribution! It looks like you've fixed this for version 2.0 of UCP. Can you make the same change to the `2.1` version?

constanzekratel:
@joaofnfernandes yes sure I'll be happy to do that

joaofnfernandes:
3...2...1... We're going live with this 🚀

PR #2887, opened 2017-04-19
Possible left/right navbar sidebar menu scrollbar not visible fix

stevenhanna6:
### Proposed changes
The homepage and many other pages have scrolling issues with the left and right scrollbars if content is expanded within them. There is a function called checkNavSizes() in docs.js that tried to fix this issue I believe but was incomplete. Because the homepage is different from the other page, a check is needed and the sidebar height to set it accordingly. The other pages just needed to set the correct tags with the correct height when the footer comes into view as was previously attempted.

layout.scss "sidebar-wrapper" needed the transition style removed or setting the height would not be instant on the right side menu. Also "sidebar" the "height: 100% !important" had to be changed to just "height: 100%" to allow the height to be set.

### Related issues (optional)
Fixes: #2232

### DIFF
--- a/_scss/_layout.scss
+++ b/_scss/_layout.scss
@@ -234,6 +234,6 @@ section.section {
     overflow-y: auto;
     padding: 20px 10px 100px 0;
     position: fixed;
-    height: 100% !important;
+    height: 100%;
     width: 270px;
 }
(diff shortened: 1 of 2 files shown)


johndmulhausen:
Sorry, @stevenhanna6 but we went with #2860 for this bug. We'll shoot you 25 points for the attempted PR. :)

stevenhanna6:
Yah I was impressed with allejo's fix, much cleaner and better solution by a mile, thanks for the 25pts.

PR #2889, opened 2017-04-19
Docker for AWS: Added missing -i <path-to-ssh-key> parameter to the tunneling paragraph

jimsander:
Added missing `-i <path-to-ssh-key>` parameter to the tunneling paragraph

### DIFF
--- a/docker-for-aws/deploy.md
+++ b/docker-for-aws/deploy.md
@@ -42,7 +42,7 @@ Once you are logged into the container you can run Docker commands on the swarm:
 
 You can also tunnel the Docker socket over SSH to remotely run commands on the cluster (requires [OpenSSH 6.7](https://lwn.net/Articles/609321/) or later):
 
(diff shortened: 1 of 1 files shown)


johndmulhausen:
@FrenchBen Does this look right to you?

FrenchBen:
@jimsander I'd assumed that people used their _default_ `id_rsa` key for this; 
I see the inconsistence between the previous line and the regular ssh one. 
`ssh -i <path-to-ssh-key>`

IMHO, the first entry should have the `-i <path-to-ssh-key>` part removed; users with multiple keys will be familiar with adding their keys to the cli or a config file.

jimsander:
@FrenchBen both points have merit, but the consistency is what caught my eye.    

As a sidenote, not everybody uses a single/default identity file, esp when folks are dealing with multiple aws iam accounts

FrenchBen:
@jimsander fair enough, the consistency makes sense to me.

@johndmulhausen LGTM

PR #2892, opened 2017-04-19
Added timeout parameters so that request to application completes qui…

shekhargulati:
…ckly

This change will make sure request gets completed quickly. Without these it takes more than a minute for request to finish. This is especially important for beginners because they might not understand why request is taking so much time. Adding these parameters request completes quickly.

### DIFF
--- a/get-started/part2.md
+++ b/get-started/part2.md
@@ -109,7 +109,7 @@ import os
 import socket
 
 # Connect to Redis
-redis = Redis(host="redis", db=0)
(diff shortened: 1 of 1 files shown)


johndmulhausen:
I have no problem with this, although I don't get that behavior at all, @shekhargulati.

shekhargulati:
@johndmulhausen yes, I ran all the commands and all the commands finish successfully. The curl command takes close to a minute.
The only difference I noticed is that I am using Docker for Mac.

Issue #2098, opened 2017-03-03
 Clarify whether a unique repository can be used both for Debian and Raspbian


jcberthon:
[engine/installation/linux/debian.md](https://docs.docker.com/engine/installation/linux/debian/)

Previous version of this document had clearer instruction on how to setup the Deb repository for Raspbian. In current release, Raspbian is still mentioned in the supported OS, but there is no longer a specific version for it.

Current document state:

```
$ sudo add-apt-repository \
   "deb [arch=amd64] https://download.docker.com/linux/debian \
   $(lsb_release -cs) \
   stable"
```

It seems that for Raspbian we could possibly use:

```
$ sudo add-apt-repository \
   "deb [arch=armhf] https://download.docker.com/linux/debian \
   $(lsb_release -cs) \
   stable"
```

But that does not work. The `add-apt-repository` command does not work on Raspbian. What I did is the following:

```
$ echo "deb [arch=armhf] https://download.docker.com/linux/debian \
   $(lsb_release -cs) stable" | sudo tee /etc/apt/sources.list.d/docker.list
```

That seems to work and I am able to install `docker-ce` but when doing `sudo apt-get update` I have the following warning:

```
W: Conflicting distribution: https://download.docker.com jessie InRelease (expected jessie but got )
```

Anyway, previous Docker release (e.g. 1.13.1) supported Debian and Raspbian for armhf (there was 2 distinct deb packages). The documentation is unclear now if a unique repository can be used for both Debian and Raspbian.

sandor-juhasz:
I have ran into the same issue today. Another workaround until the page is fixed is to follow the quick install command from the Raspberry PI homepage's Docker article at https://www.raspberrypi.org/blog/docker-comes-to-raspberry-pi/ It will install the latest CE version.

NonCreature0714:
I'm working on this 💯  @mstanleyjones @londoncalling

londoncalling:
Fixed by PR #2893

jcberthon:
I cannot reopen this issue, but I'm going to propose a PR for this issue with slightly different changes which in my opinion are not perfect but getting closer what it should be.

jcberthon:
@londoncalling @sandor-juhasz @NonCreature0714 

Perhaps you could all have a look at my 2 PRs and let me know what you think?

One is about consistency so that the Debian specific parts stays pure Debian (and not a mixture of Debian/Raspbian).

[...]

PR #2893, opened 2017-04-19
corrected Raspian steps

NonCreature0714:
@londoncalling @mstanleyjones

### Proposed changes

Addressing bug #2098, this documentation specifies more clearly what steps are necessary for installing docker-ce on Raspian.

### DIFF
--- a/engine/installation/linux/debian.md
+++ b/engine/installation/linux/debian.md
@@ -146,11 +146,15 @@ from the repository.
     **armhf**:
 
     ```bash
-    $ sudo add-apt-repository \
-       "deb [arch=armhf] {{ download-url-base }} \
-       $(lsb_release -cs) \
-       stable"
+    $ echo "deb [arch=armhf] https://apt.dockerproject.org/repo \
+        raspbian-jessie main" | \
+        sudo tee /etc/apt/sources.list.d/docker.list
     ```
+
+    > **Warning**: It is HIGHLY recommended you use
+    > ```bash $ curl -sSL https://get.docker.com | sh ``` in order to install Docker on
+    > Raspberry Pi's!! This will also install bash auto-completion and
+    > appropriately mirror the download depending on your country.
 
 4.  **Wheezy only**: The version of `add-apt-repository` on Wheezy adds a `deb-src`
     repository that does not exist. You need to comment out this repository or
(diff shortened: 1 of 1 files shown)


engine/installation/linux/debian.md
```diff
@@ -146,12 +146,16 @@ from the repository.
...
+    $ echo "deb [arch=armhf] https://apt.dockerproject.org/repo \
+        raspbian-jessie main" | \
+        sudo tee /etc/apt/sources.list.d/docker.list
     ```
 
+    > **Warning**: It is HIGHLY recommended you use
+    > ```bash $ curl -sSL https://get.docker.com | sh ``` in order to install Docker on
+    > Raspberry Pi's!! This will also install bash auto-completion and
+    > appropriately mirror the download depending on your country.
+
```
londoncalling: @NonCreature0714 just a minor comment to replace the double exclamations with just a period :-) We can leave the all-caps HIGHLY, but I think that and the warning formatting should be enough. (Giving the pros for doing this is great, but if you want to add any particular downside of _not_ using this command, that might be more compelling than the exclamation marks.)

londoncalling:
LGTM - great! I'll merge this and just make a few formatting changes.

 @jcberthon, @sandor-juhasz can you review this for technical content since you all are on the original issue?

sandor-juhasz:
The updated doc looks good, it describes both the convenience-script-based installation and the apt-based manual installation clearly. Thanks for the fix!

jcberthon:
The proposed changes are using the old URL which was used prior to the introduction of Docker CE/EE. It seems that Docker provides a path for those old URL to upgrade to Docker CE, but as far as I can tell this is not the recommended upgrade path on Debian and Ubuntu where the documentation switch from `https://apt.dockerproject.org/repo` to `https://download.docker.com/linux`. This is also hinted by the fact that one would need to install `docker` and not `docker-ce`.

I think the proper installation way on Raspbian is to do as for Debian but not using `add-apt-repository` as I reported:

```
echo "deb [arch=armhf] https://download.docker.com/linux/debian \
```

Something was fixed on Docker side, and now I do not see the warning I earlier reported in the issue. Docker run fine with this installation and it even installs the Bash completion files. I will propose a PR for this change. Perhaps someone from Docker could see is instead of using the URL `https://download.docker.com/linux/debian` there is one dedicated to Raspbian or not and correct my PR.

If what I propose above is correct, then mention of the recommended installation via downloading and running a script should be removed, much like it does not exist for Debian nor Ubuntu.

Other than that the PR looks good to me. And my PR will only be a suggestion, I do not mean that this PR was wrong :-)

PR #2898, opened 2017-04-19
command  error

CodingAdai:
"docker-machine scp myvm1 docker-compose.yml myvm1:~ " is wrong，should "docker-machine scp docker-compose.yml myvm1:~"


johndmulhausen:
Yes, thank you, @CodingAdai :)

PR #2899, opened 2017-04-19
Fix the file name in the command

CodingAdai:
According to the context, this file name is incorrect. The correct file name should be: docker-compose.yml


johndmulhausen:
Fair enough, thank you @CodingAdai :)

PR #2904, opened 2017-04-19
Fix links for Docker for AWS/Azure

jalmansor:
Both links pointed to docker-cloud/, updated to the correct locations.


johndmulhausen:
Eep! Thanks, @jalmansor :)

PR #2905, opened 2017-04-19
fixes command execution ambiguity in compose/install.md

### LINKED ISSUE #1678: Command execution ambiguity in compose/install.md
In [compose/install.md](https://docs.docker.com/compose/install/#/install-as-a-container) it is unobvious whether you should run the command as root or as a user. The $ in the beginning of the lines suggests that one runs it as user, though actually you should be running this as root (and also prepending sudo to the first of these lines would not work as with many others).

surajnarwade:
### Proposed changes

Fixes command execution ambiguity by adding `sudo` when required and corrected some commands which were wrong

### Related issues (optional)

Resolves #1678

### DIFF
--- a/compose/install.md
+++ b/compose/install.md
@@ -60,7 +60,7 @@ To install Compose, do the following:
 ### Install using pip
 
 Compose can be installed from [pypi](https://pypi.python.org/pypi/docker-compose)
-using `pip`. If you install using `pip`, it is highly recommended that you use a
+using `pip`. If you install using `pip`, we recommend that you use a
 [virtualenv](https://virtualenv.pypa.io/en/latest/) because many operating systems
 have python system packages that conflict with docker-compose dependencies. See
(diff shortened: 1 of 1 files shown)


londoncalling:
@surajnarwade Your changes look good.

I know this wasn't explicitly part of the issue described, but in testing the procedure I noticed the `curl` download example doesn't seem to work anymore. Can you replace the `curl` example shown under step 3 with this up-to-date example from the [release](https://github.com/docker/compose/releases) downloads page? Then should be good to go.

```
curl -L https://github.com/docker/compose/releases/download/1.12.0/docker-compose-`uname -s`-`uname -m` > /usr/local/bin/docker-compose
chmod +x /usr/local/bin/docker-compose
```

surajnarwade:
@londoncalling sure

londoncalling:
LGTM

PR #2909, opened 2017-04-19
Self-signed certificates setup on Windows (Part 1)

tippexs:
Copied the "Windows" section from https://docs.docker.com/datacenter/dtr/2.2/guides/user/access-dtr/ and changed the deep linkt to the msdn documentation.

### DIFF
--- a/registry/insecure.md
+++ b/registry/insecure.md
@@ -108,3 +108,22 @@ $ update-ca-trust enable
 ```
 
 Restart Docker for the  changes to take effect.
+
+### Windows
+
+In your browser navigate to `https://<dtr-url>/ca` to download the TLS
+certificate used by DTR. Open Windows Explorer, right-click the
+file you've downloaded, and choose **Install certificate**.
+
+Then, select the following options:
+
+* Store location: local machine
+* Check 'place all certificates in the following store'
+* Click 'Browser', and select 'Trusted Root Certificate Authorities'
+* Click 'Finish'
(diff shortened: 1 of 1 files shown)


registry/insecure.md
```diff
@@ -108,3 +108,22 @@ $ update-ca-trust enable
 ```
 
 Restart Docker for the  changes to take effect.
+
+### Windows
+
+In your browser navigate to `https://<dtr-url>/ca` to download the TLS
```
joaofnfernandes: Docker Registry doesn't have this functionality, so in the previous step of this tutorial you are creating the TLS certificates by yourself. Can you rephrase this to say something like:

"Configure Windows to trust your certificate. Open Windows Explorer ..."

registry/insecure.md
```diff
@@ -108,3 +108,22 @@ $ update-ca-trust enable
...
+certificate used by DTR. Open Windows Explorer, right-click the
+file you've downloaded, and choose **Install certificate**.
+
+Then, select the following options:
+
+* Store location: local machine
+* Check 'place all certificates in the following store'
+* Click 'Browser', and select 'Trusted Root Certificate Authorities'
+* Click 'Finish'
+[See also] (https://technet.microsoft.com/en-us/library/cc754841(v=ws.11).aspx#BKMK_addlocal)
```
joaofnfernandes: These links are not rendering properly, you need to remove the space between `[]()`.
Also, there's no reason for having duplicate links, so you can use the sentence of the second url, and the deep link that you have on the first.

Issue #357, opened 2016-10-26
Self-signed certificates setup on Windows


TomHart:
URL: [https://docs.docker.com/registry/insecure.md](https://docs.docker.com/registry/insecure/)

I've follow the "Using self-signed certificates" on a remote Ubuntu server, which works great, however, of course on Windows I don't have a "/etc/docker/certs.d/myregistrydomain.com:5000/ca.crt." folder.

Where about do I need to place the certificate to pacify the error message "Get https://my.registry.com:5000/v1/_ping: x509: certificate signed by unknown authority"?

joaofnfernandes:
From my initial research it seems that you need to add the certs to `%programdata%\docker\certs.d`.
Probably configuring Windows to trust those certificates ([as described here](https://technet.microsoft.com/en-us/library/cc754841(v=ws.11).aspx)) will also work.

But I need to research this a bit more. Please update this issue if you find more info in the meantime.

TomHart:
I've had to drop the project where this was an issue temporarily, but I did get around it, however, I'm not sure which one was the actual fix, or if I needed both, but I added the cert nativley to Windows so it trusted it, and I also installed it in my Docker VM.

mstanleyjones:
@joaofnfernandes I think this was addressed as part of the new content trust docs.

joaofnfernandes:
Yup. Instead of configuring docker, you should configure Windows to trust that CA. That way, other client tools will also trust it.

[Check here how to do that](https://docs.docker.com/datacenter/dtr/2.2/guides/user/access-dtr/#windows)

I'm leaving this issue open since the registry docs need a review to have a page that's essentially a copy of that one.

londoncalling:
Assigning `hackathon` label. 

The goal for the Hackathon is to figure out if there is useful information we can add to the first pages the user mentions ([Test an insecure registry](https://docs.docker.com/registry/insecure/)) to make it more intuitive so that he could have gotten to the right place on his own.

tippexs:
claimed

joaofnfernandes:
#2909 is related to this

joaofnfernandes:
Fixed by #2909 and #2911

PR #2911, opened 2017-04-19
Self-signed certificates setup on Windows (Part 2)

tippexs:
Adjusted the Link to the msdn documentation as did in #357

### Proposed changes
in #357 the Link to the msdn documentation was updated. I changed the links in this articel as well.

#
### Related issues (optional)

 

Signed-off-by: Timo Stark <tippexs91@googlemail.com>

### DIFF
--- a/datacenter/dtr/2.2/guides/user/access-dtr/index.md
+++ b/datacenter/dtr/2.2/guides/user/access-dtr/index.md
@@ -44,6 +44,7 @@ Then, select the following options:
 * Click 'Browser', and select 'Trusted Root Certificate Authorities'
 * Click 'Finish'
 
+[See also] (https://technet.microsoft.com/en-us/library/cc754841(v=ws.11).aspx#BKMK_addlocal).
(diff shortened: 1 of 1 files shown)


datacenter/dtr/2.2/guides/user/access-dtr/index.md
```diff
@@ -44,6 +44,7 @@ Then, select the following options:
 * Click 'Browser', and select 'Trusted Root Certificate Authorities'
 * Click 'Finish'
 
+[See also] (https://technet.microsoft.com/en-us/library/cc754841(v=ws.11).aspx#BKMK_addlocal).
 [Learn more about managing TLS certificates](https://technet.microsoft.com/en-us/library/cc754841(v=ws.11).aspx).
```
joaofnfernandes: There's no reason for having two links to the same place, that will confuse users. You can just update the existing url instead of adding a new `see also`.
Also, when creating urls in markdown you need to do `[]()` without a space between the two.

PR #2915, opened 2017-04-19
Raspbian installation follow-up

mstanleyjones:
### Proposed changes

Refinement for #2893, which was inadvertently merged too soon.

cc/@NonCreature0714

### DIFF
--- a/engine/installation/linux/debian.md
+++ b/engine/installation/linux/debian.md
@@ -168,6 +168,15 @@ from the repository.
 
       > **Warning**: Always audit scripts downloaded from the internet before
       > running them locally.
+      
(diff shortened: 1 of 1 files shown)


jcberthon:
I have corrected some of the statements here in PR #2965 (see also its related PR #2966 ).

As far as I can tell, the Raspbian package does contains the Bash autocompletion file (it is in the .deb file), and when I install it on my Raspberry, completion works.

In addition, there is now a dedicated section for Raspbian, so I decided that the section on Debian should install the specific Debian for ARMHF package rather than the Raspbian variant. Corollary: the repository for Raspbian in the document will install Docker CE Edge, not Stable, whereas with the Debian URL one can chose between Stable and Edge.

PR #2916, opened 2017-04-19
correcting splunk-source parameter

jpooler:


johndmulhausen:
Closing in favor of #2917, which I like a little better, but you'll still get 25 pts @jpooler

Issue #2675, opened 2017-04-07
Splunk Logging Docs: splunk-source


carlvanschie:
[engine/admin/logging/overview.md](https://docs.docker.com/engine/admin/logging/overview/), CC @mstanleyjones 

It looks like splunk-source documentation is a little off as it looks like the splunk-token tag.

jpooler:
#dibs for the hackathon

mstanleyjones:
Sorry @jpooler, this one is back up for grabs!

jpooler:
Added to PR https://github.com/docker/docker.github.io/pull/2916

johndmulhausen:
@jpooler and @constanzekratel both can split 25 points each.

PR #2917, opened 2017-04-19
#2675: Splunk Logging Docs: updated splunk-source description

constanzekratel:
This description had a splunk-token example instead of splunk-source. Went ahead and updated the example. Remove bullet from a non-list item.


johndmulhausen:
Thanks, @constanzekratel :)

PR #2920, opened 2017-04-19
Provide a bit more context around where commands should be executed.

### LINKED ISSUE #995: Differentiate whether each step happens on the local or remote machine
It's not clear which steps are done on the local, and which are done on the remote machines.

Also not clear which machine the resulting files are supposed to live on.

[engine/security/https.md](https://docs.docker.com/engine/security/https/)

clocklear:
### Proposed changes

I added comments and notes that should help the reader understand where certain commands should be executed (whether that's on the Docker daemon's host machine, or the proposed Docker client machine).

### Related issues (optional)

Fixes #995


clocklear:
Not sure why this failed, my changes were pretty simple. :-/

PR #2924, opened 2017-04-19
Doc Hackathon: Issue#1465 https://github.com/docker/docker.github.io/…

### LINKED ISSUE #1465: Specify that all nodes need to be the same exact Docker version for production UCP clusters

constanzekratel:
Added a note saying that the version of Docker CS Engine must be the same for all nodes. Fixes #1465

### DIFF
--- a/datacenter/ucp/2.1/guides/admin/install/index.md
+++ b/datacenter/ucp/2.1/guides/admin/install/index.md
@@ -10,7 +10,7 @@ installed on-premise or on a cloud infrastructure.
 ## Step 1: Validate the system requirements
 
 The first step to installing UCP, is ensuring your
-infrastructure has all the [requirements UCP needs to run](system-requirements.md).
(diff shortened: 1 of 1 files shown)


joaofnfernandes:
Relates to #2885

joaofnfernandes:
Netlify is sad, but there's nothing in this PR that could break the docs. Let's make this go live 🚀

johndmulhausen:
Thanks, @constanzekratel :) (*shakes fist at Netlify*)

Issue #1840, opened 2017-02-17
Remove cadvisor from example host port mapping in swarm services


trapier:
### Problem description

https://docs.docker.com/engine/swarm/services/#/publish-a-services-ports-directly-on-the-swarm-node should not provide an example of running cadvisor in a container as an example if cadvisor is known to cause problems with volumes https://docs.docker.com/engine/admin/troubleshooting_volume_errors/.

### Suggestions for a fix

Use any other app in the swarm services example.

jmaitrehenry:
From slack:
me: Can we simplify the example by using nginx or redis?
johnharris85: ye nginx or tomcat or apache or whatever is probably a good example image for that

t3h2mas:
Claiming

t3h2mas:
https://github.com/docker/docker.github.io/pull/2939

mstanleyjones:
Closed via #2939

PR #2939, opened 2017-04-19
remove cadvisor from swarm example -- #1840

t3h2mas:
Signed-off-by: Thomas Noe <thomas.noe@madwiremedia.com>

### Proposed changes
Use nginx instead of cAdvisor for swarm example

#1840


engine/swarm/services.md
```diff
@@ -394,40 +394,28 @@ option to the `--publish` flag.
...
-port on a given node will show you **that node's** statistics. In practice, this
-is similar to running a single stand-alone cAdvisor container on each node, but
-without the need to manually administer those containers.
+##### Example: Run a `nginx` web server service on every swarm node
+
+[nginx](https://hub.docker.com/_/nginx/) is an open source reverse proxy, load balancer, HTTP cache, and a web server. If you run nginx as a service using the routing mesh, connecting
+to the nginx port on any swarm node will show you the web page for
+(effectively) **a random swarm node** running the service.
+
+The following example runs nginx as a service on each node in your swarm and
```
jmaitrehenry: Is it really necessary for this example? I can't find when having a nginx listening locally would be a good idea.

t3h2mas: I kept that in there to explain the load balancing process. Can you explain what you mean by  > I can't find when having a nginx listening locally would be a good idea.

jmaitrehenry: "good idea" may not be the right word, it's more I don't see a use case of exposing nginx port locally.
And my comment should be on the next line... my bad.

mstanleyjones:
Kicked off Netlify.

I don't think this is as good an example, because `cadvisor` was looking at metrics relating to THAT NODE and it wouldn't make sense to get the metrics of a random load-balanced node. I can understand the reservations about using `cadvisor`. I had them too when I wrote the example. But I don't think there is any benefit in knowing that you are hitting that specific node's `nginx` instance, the way that there was with `cadvisor`. Any thoughts on that?

jmaitrehenry:
@johndmulhausen do you want to update your PR? Or I can make the change if you want.

t3h2mas:
I've used an nginx instance on each node to understand load balancing / test the mesh routing. I was in a hurry trying to get this in at Dockercon. Let me know what changes you'd like to see.

Issue #1978, opened 2017-02-25
Need mapping of pre-1.13 to post 1.13 iptables handling


adamaig:
[engine/userguide/networking/default_network/container-communication.md](https://docs.docker.com/engine/userguide/networking/default_network/container-communication/), CC: @mstanleyjones

It seems like this document is out of sync with the recent changes to the iptables handling in version 1.13. It would be useful to have clear examples of different configuration option sets that yield a system with equivalent rules to the pre-1.13 behavior.

stevenhanna6:
I'll take a shot at this one, claimed

PR #2941, opened 2017-04-20
Explaining the default iptables FORWARD DROP policy

stevenhanna6:
### Proposed changes
Basically explains the default FORWARD DROP policy in iptables, how to change it and a reason you might need to do this. Also explains how a pre1.13  versions might get caught off guard with the setting changed once upgrading.

### Related issues (optional)
Fixes: #1978


stevenhanna6:
Your copyedits look great! is there anything I have to do? It won't let me approve changes from my end if that's what I'm supposed to do. I'm new to this, apologies.

johndmulhausen:
Thanks, @stevenhanna6 this is fantastic work!

Issue #2580, opened 2017-03-31
Improve requirements for HRM network


joaofnfernandes:
https://docs.docker.com/datacenter/ucp/2.1/guides/user/services/use-domain-names-to-access-services/#under-the-hood

Explains that to use the HTTP routing mesh you can either use the default `ucp-hrm` network or create one. But we should show an example command on how to create that network, since that makes it more explicit.

We also need to make it explicit if you can create these networks at any time or you need to create them before enabling the HRM service, or you disable/enable the HRM service.

Rubab-Syed:
I'll take this!

prashant171992:
I'm looking into this.

PR #2943, opened 2017-04-20

prashant171992:
Fixed https://github.com/docker/docker.github.io/issues/2580

### Proposed changes

Updated how ucp-hrm network is created.

### Related issues (optional)

https://github.com/docker/docker.github.io/issues/2580

### DIFF
--- a/datacenter/ucp/2.1/guides/user/services/use-domain-names-to-access-services.md
+++ b/datacenter/ucp/2.1/guides/user/services/use-domain-names-to-access-services.md
@@ -95,8 +95,7 @@ to your services.
 
 The HTTP routing mesh can route to a service, as long as that service:
 
-* Is attached to a network that has the `com.docker.ucp.mesh.http` label. Enabling the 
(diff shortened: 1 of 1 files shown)


datacenter/ucp/2.1/guides/user/services/use-domain-names-to-access-services.md
```diff
@@ -95,8 +95,8 @@ to your services.
 
 The HTTP routing mesh can route to a service, as long as that service:
 
-* Is attached to a network that has the `com.docker.ucp.mesh.http` label. You
-can use the default `ucp-hrm` network or create your own
+* Is attached to a network that has the `com.docker.ucp.mesh.http` label. Enabling the
```
joaofnfernandes: I don't think this addresses #2580. For that, I think you should had a snippet after line 181 that shows how to create an overlay network with the HRM labels and UCP permission labels.

You also need to mention that if you're creating a a new HRM network you need to disable the HRM service first, or disable and enable the HRM service after you create the network. And doing that, requires admin permissions.

prashant171992:
@joaofnfernandes done some changes , please have a look

PR #2948, opened 2017-04-20
fix for a tiny grammar mistake

mtilson:
### Proposed changes

fixed a tiny grammar mistake, why is clear from the commit


johndmulhausen:
Thanks, @mtilson :)

PR #2952, opened 2017-04-20
updated docker unpause

### LINKED ISSUE #2951: show output on docker unpause
### Problem location

https://docs.docker.com/engine/reference/commandline/unpause/

- Other: <DETAILS>

### Suggestions for a fix

as `$ docker unpause my_container` is written as command, we should show its output also

surajnarwade:
resolves #2951

modifies https://docs.docker.com/engine/reference/commandline/unpause/


johndmulhausen:
Sorry, @surajnarwade, but the YAML files are generated from the source at moby/moby. You will need to make these changes there.

PR #2956, opened 2017-04-20
Update delete-service.md

guilhermebr:
`docker service ps` needs the SERVICE-ID as a parameter, and the service don't exist after removed.


johndmulhausen:
Thanks, @guilhermebr :)

PR #2958, opened 2017-04-20
Fix ipv6 netmask

### LINKED ISSUE #933: Mistake in netmask for ipv6
[engine/userguide/networking/default_network/ipv6.md](https://docs.docker.com/engine/userguide/networking/default_network/ipv6/)

I think this line is incorrect: So if you have a /64 for your whole environment use /78 subnets for the hosts and /80 for the containers. This way you can use 4096 hosts with 16 /80 subnets each.

2^12 = 4096
2^4 = 16

I think it should be a /76 subnet: 12 bits = 4096 hosts and 4 bits = 16 /80 subnets

shubheksha:
### Related issues (optional)
Fixes #933


drnybble:
Please review this fix, it is wrong. The fix is to change the /78 to /76. Now it is wrong twice.

shubheksha:
I'll open a new PR, @drnybble! Sorry for the trouble

PR #2961, opened 2017-04-20
Added more info about volumes

### LINKED ISSUE #344: Improve/expand info on volumes, `docker volume` and related concepts
URL: [https://docs.docker.com/engine/tutorials/dockervolumes.md](https://docs.docker.com/engine/tutorials/dockervolumes/)

Can someone expand on `docker volume` and write a separate section on how to manage those since that's the new preferred method of volume management?  I would contribute some edits myself but I'm only getting to know it now.  

Specifically information regarding **backup/restore** and **backups/restores** with **labels** and things like that.

I've managed to sort out this information on my own but someone coming to this guide for the first time might get confused with the focus mainly being on the `docker create -v` and `docker create --name data busybox` and `docker volume create` all of these methods are great and important to keep documented but ideally we should have something set to be the primary / recommended vs deprecated methods.

surajnarwade:
Resolves #344

https://docs.docker.com/engine/tutorials/dockervolumes/

### DIFF
--- a/engine/tutorials/dockervolumes.md
+++ b/engine/tutorials/dockervolumes.md
@@ -37,7 +37,7 @@ Docker therefore *never* automatically deletes volumes when you remove
 a container, nor will it "garbage collect" volumes that are no longer
 referenced by a container.
 
-### Adding a data volume
+### Add a data volume
 
 You can add a data volume to a container using the `-v` flag with the
 `docker create` and `docker run` command. You can use the `-v` multiple times
@@ -54,7 +54,7 @@ This will create a new volume inside a container at `/webapp`.
 > You can also use the `VOLUME` instruction in a `Dockerfile` to add one or
 > more new volumes to any container created from that image.
 
-### Locating a volume
+### Locate a volume
 
 You can locate the volume on the host by utilizing the `docker inspect` command.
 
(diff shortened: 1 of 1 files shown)


engine/tutorials/dockervolumes.md
```diff
@@ -332,6 +332,19 @@ $ docker run --rm --volumes-from dbstore2 -v $(pwd):/backup ubuntu bash -c "cd /
 You can use the techniques above to automate backup, migration and
 restore testing using your preferred tools.
 
+## Listing all volumes
```
mstanleyjones: We try not to use the gerund format (-ing) in titles, so please change this to "List all volumes"

engine/tutorials/dockervolumes.md
```diff
@@ -332,6 +332,19 @@ $ docker run --rm --volumes-from dbstore2 -v $(pwd):/backup ubuntu bash -c "cd /
...
+
+You can list out all volumes which are created.
```
mstanleyjones: Maybe "List all existing volumes:"

engine/tutorials/dockervolumes.md
```diff
@@ -347,6 +360,14 @@ $ docker run --rm -v /foo -v awesome:/bar busybox top
 This command creates an anonymous `/foo` volume. When the container is removed,
 the Docker Engine removes the `/foo` volume but not the `awesome` volume.
 
+If you want to remove all unused volumes, which are unnecessarily occupying the space,
```
mstanleyjones: Maybe something like "To remove unused volumes and free up space:"

londoncalling:
LGTM, it looks like @mstanleyjones  comments are addressed, merging

PR #2965, opened 2017-04-20
Correct Debian for ARMHF installation steps

### LINKED ISSUE #2098: Clarify whether a unique repository can be used both for Debian and Raspbian

jcberthon:
### Proposed changes

My goal is to correct the installation instruction for Debian. The current document mixes up Debian for ARM and Raspbian.

### Related issues (optional)

Fixes #2098 

I was not entirely satisfied by the changes in the PR which fixed the issue. So here is my first contribution which clarifies the Debian part of this document.

### DIFF
--- a/engine/installation/linux/debian.md
+++ b/engine/installation/linux/debian.md
@@ -145,32 +145,11 @@ from the repository.
 
     **armhf**:
 
-    You can choose between two methods for `armhf`. You can use the same method
-    as Debian, setting up the repository and using `apt-get install`, or you can
-    use a convenience script, which requires privileged access, but sets up the
-    repository for you and installs the packages for Bash auto-completion.
-
-    - Setting up the repository directly:
-
-      ```bash
-      $ echo "deb [arch=armhf] https://apt.dockerproject.org/repo \
-          raspbian-jessie main" | \
-          sudo tee /etc/apt/sources.list.d/docker.list
-      ```
-
-    - Using the convenience script:
-
-      ```bash
-      $ curl -sSL https://get.docker.com > install.sh
-
-      $ sudo bash ./install.sh
-      ```
-
-      > **Warning**: Always audit scripts downloaded from the internet before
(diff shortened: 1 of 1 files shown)


NonCreature0714:
I meant *convenience script*

jcberthon:
Hi @NonCreature0714 

I have two Raspberry Pi 2, a Raspbian and Ubuntu variant.

For both, I have been installing the Debian armhf packages. And it works fine.

I could try the convenience script on Ubuntu armhf. Do you think it makes sense? Sadly, I do not have a free SD card at the moment to try and test installing Docker with the convenience script on Debian armhf.

mstanleyjones:
@andrewhsu PTAL

PR #2966, opened 2017-04-20
Raspbian install corrections

### LINKED ISSUE #2098: Clarify whether a unique repository can be used both for Debian and Raspbian

jcberthon:
### Proposed changes

After implementing Pull Request #2965 the Raspbian section was inconsistent.

In addition the Raspbian section had several errors:

- The given URL would install the Edge version of Docker
- The package to install was incorrect (`docker` installs a KDE package, one needs to use `docker-engine`)
- The `docker-engine` package contains the Bash completion setup, so the script installation does not bring any advantages. So I remove any mention of "recommendation" for the script as for Debian and Ubuntu this is also not the recommended installation.

These errors were corrected.

### Related issues

Fixes #2098

Related PR #2965 

It completes the related PR by correcting and making consistent the Raspbian section of the Debian installation instructions.

### DIFF
--- a/engine/installation/linux/debian.md
+++ b/engine/installation/linux/debian.md
@@ -256,8 +256,33 @@ to install.
 
 
 ### Install on Raspian (Raspberry Pi)
->**Warning**: This isn't necessary if you used the recommended
->```bash $ curl -sSL https://get.docker.com | sh ``` command!
+
+You can choose between two methods for `armhf`. You can use the same method
+as Debian, setting up the repository and using `apt-get install`, or you can
+use a convenience script, which requires privileged access, but sets up the
+repository for you and installs the packages for Bash auto-completion.
+
+- Setting up the repository directly:
+
+  ```bash
+  $ echo "deb [arch=armhf] https://apt.dockerproject.org/repo \
+      raspbian-jessie main" | \
+      sudo tee /etc/apt/sources.list.d/docker.list
+  ```
+
+- Using the convenience script:
+
+  ```bash
+  $ curl -sSL https://get.docker.com > install.sh
+
+  $ sudo bash ./install.sh
+  ```
+
+  > **Warning**: Always audit scripts downloaded from the internet before
+  > running them locally.
+      
+  If you use this method, Docker is installed and starts automatically.
+  Skip the rest of this step.
 
 Once you have added the Docker repo to `/etc/apt/sources.list.d/`, you should
 see `docker.list` if you:
--- a/engine/installation/linux/debian.md
+++ b/engine/installation/linux/debian.md
@@ -260,9 +260,9 @@ to install.
 You can choose between two methods for `armhf`. You can use the same method
 as Debian, setting up the repository and using `apt-get install`, or you can
 use a convenience script, which requires privileged access, but sets up the
-repository for you and installs the packages for Bash auto-completion.
+repository for you and start automatically the daemon.
 
-- Setting up the repository directly:
+- Setting up the repository directly (recommended):
 
   ```bash
   $ echo "deb [arch=armhf] https://apt.dockerproject.org/repo \
@@ -312,13 +312,13 @@ installing Docker.
     Use this command to install the latest version of Docker:
 
     ```bash
-    $ sudo apt-get install docker
+    $ sudo apt-get install docker-engine
     ```
-    > **NOTE**: By default, Docker on Raspian is Docker Community Edition, so
-    > there is no need to specify docker-ce.
-
-    > **NOTE**: If ```bash $ curl -sSL https://get.docker.com | sh ``` isn't used,
-    > then docker won't have auto-completion! You'll have to add it manually.
+    > **NOTE**: Currently Docker on Raspian is Docker Community Edition Edge.
+    > The URL is still the same as prior to the introduction of the CE and EE edition.
+    > It is possible that in the future this URL does not work anymore,
+    > then please use the Debian URL. Use also that URL if you want to use the stable
+    > edition.
 
 3.  Verify that Docker is installed correctly by running the `hello-world`
     image.
--- a/engine/installation/linux/debian.md
+++ b/engine/installation/linux/debian.md
@@ -260,9 +260,9 @@ to install.
 You can choose between two methods for Raspbian on `armhf`. You can use the
 same method as Debian, setting up the repository and using `apt-get install`,
 or you can use a convenience script, which requires privileged access, but
-sets up the repository for you and start automatically the daemon.
+sets up the repository for you and automatically start Docker.
 
-- Setting up the repository directly:
+- Set up the repository directly:
 
   ```bash
   $ echo "deb [arch=armhf] https://apt.dockerproject.org/repo \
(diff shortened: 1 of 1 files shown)


NonCreature0714:
There are some minor English grammar issues in the directions. Additionally, I think the "recommend" notes are just opinions which can be rationalized all day, but choices should be left up to the user.

jcberthon:
Hi @NonCreature0714 

Thank you for your feedback. I agree with you about the "recommended" notes, so I removed it to make it unbiased.

I'm not a native English speaker, so I have attempted to correct some of my grammar. Sometimes I decided to rephrase the sentence. Hopefully this is now a bit improved. Fill free to correct me and/or to point me to the sentence I should improve.

mstanleyjones:
I'm finding it a little hard to understand how this fits with #2965. Will they conflict with each other?

jcberthon:
No they are complementary. Each PR is a change on a specific section, not overlapping. 

The current PR is about the Raspbian specific section. It had a few things that could be improved or corrected, and I attempted to do so. 

The other PR is only about the "pure" Debian section and it is enhancing and correcting a few things about running Debian on armhf. 

I've made the 2 PR coherent with each other's. But they could be picked up independently. 

I hope it clarifies my approach.

engine/installation/linux/debian.md
```diff
@@ -255,12 +255,37 @@ To upgrade Docker, first run `sudo apt-get update`, then follow the
...
->**Warning**: This isn't necessary if you used the recommended
->```bash $ curl -sSL https://get.docker.com | sh ``` command!
+### Install on Raspbian (Raspberry Pi)
 
-Once you have added the Docker repo to `/etc/apt/sources.list.d/`, you should
-see `docker.list` if you:
+You can choose between two methods for Raspbian on `armhf`. You can use the
+same method as Debian, setting up the repository and using `apt-get install`,
+or you can use a convenience script, which requires privileged access, but
+sets up the repository for you and start automatically the daemon.
```
mstanleyjones: automatically starts Docker.

engine/installation/linux/debian.md
```diff
@@ -255,12 +255,37 @@ To upgrade Docker, first run `sudo apt-get update`, then follow the
...
+
+- Setting up the repository directly:
```
mstanleyjones: Set up the...

engine/installation/linux/debian.md
```diff
@@ -255,12 +255,37 @@ To upgrade Docker, first run `sudo apt-get update`, then follow the
...
+
+  ```bash
+  $ echo "deb [arch=armhf] https://apt.dockerproject.org/repo \
+      raspbian-jessie main" | \
+      sudo tee /etc/apt/sources.list.d/docker.list
+  ```
+
+- Using the convenience script:
```
mstanleyjones: Use the ...

engine/installation/linux/debian.md
```diff
@@ -287,13 +312,14 @@ installing Docker.
     Use this command to install the latest version of Docker:
 
     ```bash
-    $ sudo apt-get install docker
+    $ sudo apt-get install docker-engine
```
mstanleyjones: I think it is docker-ce

NonCreature0714: Hi @mstanleyjones I'll verify whether it's docker-ce or docker-engine. I just saw all your notes, and like the added clarity.

engine/installation/linux/debian.md
jcberthon: Okay, I will update that. Thanks for the feedback.

engine/installation/linux/debian.md
jcberthon: OK, will do.

engine/installation/linux/debian.md
jcberthon: OK, will do.

engine/installation/linux/debian.md
jcberthon: When you use the "old URL" as it is the case in this section for Raspbian, there is no `docker` (there is but it is that Debian package for KDE) and there is no `docker-ce` it is still called `docker-engine` as it was previously.

My assumption is that this is transitional, in order for people with the old repository to get the updated version. The question is how long this old repo is going to be maintained by the Docker team?

jcberthon:
Hi @mstanleyjones 

I have updated the file with your suggestion. Thank you for your feedback.

*Note: in the coming 3-4 days I will be in a business trip with seldom time or access to internet. So if I do not reply quickly, please excuse me. I will resume any pending tasks at the end of this week probably.*

engine/installation/linux/debian.md
mstanleyjones: You should definitely note that you are not getting `docker-ce` but an older version. @andrewhsu might have more info on the maintenance aspect. This would be  significant downside to using this script.

jcberthon: Actually I checked that when installing `docker-engine` one actually installs Docker CE from the edge channel. So not an older version. For how long this is going to be maintained is unknown to me but should be answered.

Personally, on my Raspberry Pi 2 which is running Raspbian Jessie, I have used the Debian installation instruction for armhf. I did not use the convenience script nor the old URL specific to Raspbian. Therefore I have installed Docker CE and I chose the stable channel. It is now working for roughly a month and it is very stable. I am not sure why we would need a Raspbian dedicated section in the Debian installation, I think we should use the same approach as for Debian on armhf, unless there is a specific package made by Docker for Raspbian. This needs to be confirmed by a Docker responsible.

Why didn't I remove the Raspbian section altogether? Well, I'm obviously not from Docker but a contributor. I do not know if Docker provides a specific Raspbian package or not. I also did not want to delete a complete section without this knowledge.

mstanleyjones:
@andrewhsu confirmed that the convenience script may be deprecated soon and so we should not add it to the docs for now. I'm not going to accept this PR because of that. Thanks for looking into this!

PR #2869, opened 2017-04-18
Fixing Semantic Versioning variables for patch

jennylia:
Fixes #2285
Chatted with @joaofnfernandes today :)

PR #2974, opened 2017-04-20
Improve semantic versioning table

### LINKED ISSUE #2285: no mention of a path from upgrading from 2.2.x to 2.2.3
There currently is no mention of a path from upgrading from 2.2.x to 2.2.3 and mention of whether it is supported. Ran into an issue when upgrading for 2.2.1 to 2.2.3. 

[datacenter/dtr/2.2/guides/admin/upgrade/index.md](https://docs.docker.com/datacenter/dtr/2.2/guides/admin/upgrade/)

jmaitrehenry:
### Proposed changes

Try to add a simplier table about semantic versioning upgrade.
Base on https://github.com/docker/docker.github.io/pull/2869

### Related issues (optional)

Fix #2285

### DIFF
--- a/datacenter/dtr/2.2/guides/admin/upgrade.md
+++ b/datacenter/dtr/2.2/guides/admin/upgrade.md
@@ -18,26 +18,18 @@ support upgrades according to the following rules:
   version. We also strongly recommend upgrading to the latest minor/patch
   version for your major version first.
 
-|From| To| Description| Supported|
-|:----|:---|:------------|----------|
-| x.y.0 | x.y.1 | patch upgrade | yes |
-| x.y.0 | x.y.2 | skip patch version | yes |
-| x.y.2 | x.y.1 | patch downgrade | no |
-| x.y.* | x.y+1.* | minor upgrade | yes |
-| x.y.* | x.y+2.* | skip minor version | no |
-| x.y.* | x.y-1.* | minor downgrade | no |
-| x.*.* | x+2.*.* | skip major version | no |
-| x.*.* | x-1.*.* | major downgrade | no |
-
-In the next table, y and z are the latest minor or patch version  and a,b the
-earliest minor or pacth version 
-
-|From| To| Description| Supported|
-|:----|:---|:------------|----------|
-| x.y.z | x+1.a.b | major upgrade | yes |
-| x.y.z-1 | x+1.a.b+1 | major upgrade from an old version | no |
-| x.y.z | x+1.a+1.b | major upgrade skipping minor version | no |
-
+| Description                          | From  | To        | Supported |
+|:-------------------------------------|:------|:----------|-----------|
+| patch upgrade                        | x.y.0 | x.y.1     | yes       |
+| skip patch version                   | x.y.0 | x.y.2     | yes       |
+| patch downgrade                      | x.y.2 | x.y.1     | no        |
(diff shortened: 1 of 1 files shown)


datacenter/dtr/2.2/guides/admin/upgrade.md
```diff
@@ -20,20 +20,24 @@ support upgrades according to the following rules:
...
+| x.*.* | x-1.*.* | major downgrade | no |
+
+In the next table, y and z are the latest minor or patch version  and a,b the
+earliest minor or pacth version 
+
+|From| To| Description| Supported|
+|:----|:---|:------------|----------|
+| x.y.z | x+1.a.b | major upgrade | yes |
+| x.y.z-1 | x+1.a.b+1 | major upgrade from an old version | no |
+| x.y.z | x+1.a+1.b | major upgrade skipping minor version | no |
```
joaofnfernandes: I like it, but I think that you need to do a lot of thinking when looking at table two. I think we can make it easier by moving the `Description` column to be the first one.

The notation with `a` and `b` is also interesting but it introduces a lot of complexity. Since the first minor and patch versions are always `0`, you can do

|From| To| Description| Supported|
|:-----|:--------|:---------------|---------|
| x.y.z | x+1.0.0 | major upgrade | yes |
| x.y.z | x+1.y+1.z | major upgrade, skipping minor version | no |

The case for "major upgrade from an old version" is a bit special and few users are running that configuration, so we can leave it out.
I think this is a good compromise between accuracy and simplicity. #SimpleDocsFTW

jmaitrehenry: Yeah, I already remove some line because when I read it, it was complicated for me too.
Thanks for your eyes!

jmaitrehenry:
@joaofnfernandes ready for a new review!

datacenter/dtr/2.2/guides/admin/upgrade.md
```diff
@@ -18,22 +18,25 @@ support upgrades according to the following rules:
...
+| patch upgrade | x.y.0 | x.y.1 | yes |
+| skip patch version | x.y.0 | x.y.2 | yes |
+| patch downgrade | x.y.2 | x.y.1 | no |
+| minor upgrade | x.y.* | x.y+1.* | yes |
+| skip minor version | x.y.* | x.y+2.* | no |
+| minor downgrade | x.y.* | x.y-1.* | no |
+| skip major version | x.*.* | x+2.*.* | no |
+| major downgrade | x.*.* | x-1.*.* | no |
+
+In the next table, y and z are the latest minor or patch version  and a,b the
```
joaofnfernandes: Since we're not using `a` and `b` you can remove this sentence, and make this a single table again. After that and padding the table, we're ready to ship.

Pro tip: if you're using Atom, there's this `markdown-table-formatter` that automatically fixes tables when you save.

jmaitrehenry: I check if we have the same kind of plugin for vscode, thanks for the protip!

jmaitrehenry:
@mstanleyjones and @joaofnfernandes ready for a new review :)

joaofnfernandes:
I don't see anything here that would make the build break, and was seeing some errors with the Netlify APIs. Kicking it off again to see if the works this time.

PR #2989, opened 2017-04-21
Fixed networks indentation issues

kvalev:
### Proposed changes

The indentation of several `networks` blocks was malformed, which might leave the impression that the `networks` element is not a top-level element, but a child of `services`.


johndmulhausen:
Thanks, @kvalev, this is great. :)

PR #2991, opened 2017-04-21
compose/django.md: fix Python image version

glarrain:
### Proposed changes

The base image of the Dockerfile in doc 'compose/django.md' was not consistent with the explanation below. Fixed the explanation and added a reference URL.


johndmulhausen:
Yep, thanks @glarrain this is great. :)

Issue #1805, opened 2017-02-17
Using Secrets with DDC from the CLI and Compose file


banjot:
### Problem description
The current user guide docs in DDC for using secrets only talks about the GUI. It doesn't mention the CLI as a way to manage or interface with UCP to manage a secret. 

### Problem location
https://docs.docker.com/datacenter/ucp/2.1/guides/user/secrets/

### Project version(s) affected
UCP 2.1 

### Suggestions for a fix
We should document the CLI commands for secrets and mention how this works with UCP. It's essentially the same content as here: https://docs.docker.com/engine/swarm/secrets/#/example-use-secrets-with-a-service 

We should also include the fields necessary to use secrets from a compose file. Referenced in this forum message: https://forums.docker.com/t/secrets-stacks-and-odd-permissions/28340/6

jmaitrehenry:
I take it!

stevenhanna6:
claimed

PR #2993, opened 2017-04-21
Link to docker secret CLI and docker secret compose information

stevenhanna6:
### Proposed changes

Link to docker secret CLI information for people so people don't have to search.

Added information explaining docker secrets in compose with a basic example

### Related issues (optional)
Fixes: #1805

### DIFF
--- a/engine/swarm/secrets.md
+++ b/engine/swarm/secrets.md
@@ -847,38 +847,7 @@ the information from a Docker-managed secret instead of being passed directly.
 conscious decision, because environment variables can unintentionally be leaked
 between containers (for instance, if you use `--link`).
 
-## Using Docker Secrets in Compose
-
-For this example we will create a simple WordPress site using Docker Secrets in
-a compose file. 
-
-We will be using the short syntax that only specifies the secret 
-name. This will grant the container access to the secret and will mount its value 
-at `/run/secrets/secret_name`.
-
-More information on short and long syntax is can be found at 
-[Compose file version 3 reference](/compose/compose-file/#secrets).
-
-> **Note** The latest mySQL and WordPress images have the Docker Secret functionality 
-> (i.e. ability to read from MYSQL_ROOT_PASSWORD_FILE and MYSQL_PASSWORD_FILE)
-
-Instead of hardcoding the datapase passwords in the compose file, each password 
-will be a Docker secret. Two plain text files will be used for each secret and 
-labeled 'db_password.txt' and 'db_root_password.txt'. Within each file is the actual 
-plain text passwords used to setup the mySQL database.
-
-Notice in the compose file `WORDPRESS_DB_PASSWORD_FILE` is referencing a Docker 
-Secret at `/run/secrets/db_password`
-
-For a service to have access to a secret it must be delcared under `secrets:`. 
-
-You can see the mySQL service is using two secrets under `environment:` 
-and the WordPress service is using just one secret `db_password`.
-
-The actual Docker secrets and their values are declared under a seperate root `secrets:` 
-section. We simply declare each secret and where to find the text we want to associate 
-with that secret. For this case the two secrets point to two seperate text files described earlier.
-
+## Use Secrets in Compose
 
 ```
 version: '3.1'
@@ -922,4 +891,23 @@ secrets:
 volumes:
     db_data:
 ```
+
+This example creates a simple WordPress site using two secrets in
+a compose file.
+
+The keyword `secrets:` defines two secrets `db_password:` and `db_root_password:`.
+
+When deploying, Docker will create these two secrets and populate them with the 
+content from the file specified in the compose file. 
+
+The db service uses both secrets, and the wordpress is using one. 
+
+When you deploy, Docker will mount a file under `/run/secrets/<secret_name>` in the 
+services. These files are never persisted in disk, they're managed in memory
+
+Each service has environment variables to specify where the service should look for 
+that secret data.
+
+More information on short and long syntax for secrets can be found at 
+[Compose file version 3 reference](/compose/compose-file/index.md#secrets).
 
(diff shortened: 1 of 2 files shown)


jmaitrehenry:
On the UCP page, is it helpfull to add a note about the fact that secret created using cli are available and manageable using the UI too?

stevenhanna6:
UCP page already has a blurb at the top "UCP extends the functionality provided by Docker Engine, so you can continue using the same workflows and tools you already use, like the Docker CLI client."

I can't say either way if it needs more information.

engine/swarm/secrets.md
```diff
@@ -846,3 +846,80 @@ the information from a Docker-managed secret instead of being passed directly.
 >**Note**: Docker secrets do not set environment variables directly. This was a
 conscious decision, because environment variables can unintentionally be leaked
 between containers (for instance, if you use `--link`).
+
+## Using Docker Secrets in Compose
```
joaofnfernandes: Two nitpicks:
* We use the normal form. `Use` instead of `Using`
* We only capitalize the first word and product names, and in this case the word "Docker" is not really needed. So it can be `Use secrets in Compose files`

engine/swarm/secrets.md
```diff
@@ -846,3 +846,80 @@ the information from a Docker-managed secret instead of being passed directly.
...
+
+For this example we will create a simple WordPress site using Docker Secrets in
```
joaofnfernandes: When you say `using Docker Secrets`, it seems that you're talking about a product. But this is a feature just like many others so it can be `... using secrets in a Compose file`.

engine/swarm/secrets.md
```diff
@@ -846,3 +846,80 @@ the information from a Docker-managed secret instead of being passed directly.
...
+a compose file. 
+
+We will be using the short syntax that only specifies the secret 
+name. This will grant the container access to the secret and will mount its value 
+at `/run/secrets/secret_name`.
+
+More information on short and long syntax is can be found at 
+[Compose file version 3 reference](/compose/compose-file/#secrets).
+
+> **Note** The latest mySQL and WordPress images have the Docker Secret functionality
```
joaofnfernandes: Is this important? If someone was just learning about secrets, they would probably wonder, what changed, and we don't want to get into that. So I think we can remove this note. #ShortDocsFTW

engine/swarm/secrets.md
```diff
@@ -846,3 +846,80 @@ the information from a Docker-managed secret instead of being passed directly.
...
+name. This will grant the container access to the secret and will mount its value 
+at `/run/secrets/secret_name`.
+
+More information on short and long syntax is can be found at 
+[Compose file version 3 reference](/compose/compose-file/#secrets).
+
+> **Note** The latest mySQL and WordPress images have the Docker Secret functionality 
+> (i.e. ability to read from MYSQL_ROOT_PASSWORD_FILE and MYSQL_PASSWORD_FILE)
+
+Instead of hardcoding the datapase passwords in the compose file, each password
```
joaofnfernandes: Typo. `datapase->database`

engine/swarm/secrets.md
```diff
@@ -846,3 +846,80 @@ the information from a Docker-managed secret instead of being passed directly.
...
+at `/run/secrets/secret_name`.
+
+More information on short and long syntax is can be found at 
+[Compose file version 3 reference](/compose/compose-file/#secrets).
+
+> **Note** The latest mySQL and WordPress images have the Docker Secret functionality 
+> (i.e. ability to read from MYSQL_ROOT_PASSWORD_FILE and MYSQL_PASSWORD_FILE)
+
+Instead of hardcoding the datapase passwords in the compose file, each password 
+will be a Docker secret. Two plain text files will be used for each secret and
```
joaofnfernandes: `Docker secret` -> `secret`

engine/swarm/secrets.md
```diff
@@ -846,3 +846,80 @@ the information from a Docker-managed secret instead of being passed directly.
...
+Secret at `/run/secrets/db_password`
+
+For a service to have access to a secret it must be delcared under `secrets:`. 
+
+You can see the mySQL service is using two secrets under `environment:` 
+and the WordPress service is using just one secret `db_password`.
+
+The actual Docker secrets and their values are declared under a seperate root `secrets:` 
+section. We simply declare each secret and where to find the text we want to associate 
+with that secret. For this case the two secrets point to two seperate text files described earlier.
```
joaofnfernandes: Typo `seperate->separate`

engine/swarm/secrets.md
```diff
@@ -846,3 +846,80 @@ the information from a Docker-managed secret instead of being passed directly.
...
+secrets:
+   db_password:
+     file: db_password.txt
+   db_root_password:
+     file: db_root_password.txt
+
+volumes:
+    db_data:
+```
+
```
joaofnfernandes: The explanation that you have in the top feels a bit confusing. I would put things in this worder

1. compose snippet #ShowMeTheCode
2. Explain that the `secrets` keyword defines two secrets. When deploying, Docker will create two secrets and populate them with the content thats in the file you specify in the compose file. Maybe it's also a good idea to move the whole `secrets` block to the beginning of the snippet to make it obvious to new users.
3. The `db` service is using two secrets, and the `wordpress` one. When you deploy, Docker will mount a file under /run/secrets/<secret_name> in the services. These files are never persisted in disk, they're managed in memory
4. The services have environment variables to specify where the service should look for that secret data.

stevenhanna6:
Looked at your comments and made changes, tell me what you think now.

joaofnfernandes:
I've noticed that the snippet is using `restart:always`, and `docker stack deploy` will complain that the `restart` keyword is not supported and ignores it. We can fix that on a subsequent PR.

Issue #2950, opened 2017-04-20
container_name option


mcctomsk:
Hello.
[compose/compose-file/index.md](https://docs.docker.com/compose/compose-file/), CC @londoncalling 

$ docker-compose --version
```
docker-compose version 1.11.2, build dfed245

```
In documentation you wrote:
> container_name: my-web-container
> Because Docker container names must be unique, you cannot scale a service beyond 1 container if you have specified a custom name. Attempting to do so results in an error.

But when I've tried to use container_name option I've got next warning: 
```
$ docker stack deploy -c docker-compose.yaml --with-registry-auth=true green
Ignoring deprecated options:

container_name: Setting the container name is not supported.
```
Probably there need to add your standard note too:

> This option is ignored when deploying a stack in swarm mode with a (version 3) Compose file.

joaofnfernandes:
Yup. That doc should say that `container_name` is not supported if you're deploying with `docker stack deploy`. We should also validate that this doc is correct for other [keywords that are not supported](https://github.com/moby/moby/blob/master/cli/compose/types/types.go#L34).

londoncalling:
@mcctomsk @joaofnfernandes thanks. From what I can tell, this `container_name` would inherently not make sense with `docker stack deploy` _or_ by v3 at all, so I've worded the note a little differently.

I've submitted a PR ^^ and waiting for some clarification/verify from the Compose developers.

londoncalling:
@mcctomsk thank, fixed by PR #3003, should be published soon.

PR #3003, opened 2017-04-22
update re:container_name not supported in v3 Compose

londoncalling:
### What's changed

- Removed `container_name` option from Compose v3 file reference
- Added mention of `container_name` as one of the options no longer supported v3 under "Upgrading Version 2.x to 3.x"

### Reviewers

@dnephin @shin- Can you verify that this is correct?
@joaofnfernandes pointed to this `.go` file: https://github.com/moby/moby/blob/master/cli/compose/types/types.go#L34. Definitely looks unsupported, but we missed this one in previous iterations of the Compose v3 update.

@joaofnfernandes, per your comments in the issue, the other keywords/options deprecated in v3 or already covered in the [upgrading topic](http://localhost:4000/compose/compose-file/compose-versioning/#version-2x-to-3x), are not in the v3 reference file. 

### Related 

Issue #2950 

### DIFF
--- a/compose/compose-file/compose-versioning.md
+++ b/compose/compose-file/compose-versioning.md
@@ -190,7 +190,7 @@ Designed to be cross-compatible between Compose and the Docker Engine's
 several more.
 
 - Removed: `volume_driver`, `volumes_from`, `cpu_shares`, `cpu_quota`, `cpuset`,
-  `mem_limit`, `memswap_limit`, `extends`, `group_add`. See the [upgrading](#upgrading)
+  `mem_limit`, `memswap_limit`, `extends`, `group_add`, `container_name`. See the [upgrading](#upgrading)
   guide for how to migrate away from these.
 
 - Added: [deploy](index.md#deploy)
--- a/compose/compose-file/compose-versioning.md
+++ b/compose/compose-file/compose-versioning.md
@@ -227,6 +227,7 @@ several options have been removed:
     `deploy`. Note that `deploy` configuration only takes effect when using
     `docker stack deploy`, and is ignored by `docker-compose`.
 
+-   `container_name`: This option has been removed for `version: "3.x"` Compose files.
 -   `extends`: This option has been removed for `version: "3.x"` Compose files.
 -   `group_add`: This option has been removed for `version: "3.x"` Compose files.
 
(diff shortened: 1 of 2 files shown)


dnephin:
This change is not correct. `container_name` is still in the V3 schema https://github.com/docker/compose/blob/master/compose/config/config_schema_v3.0.json#L78 and is supported in docker-compose.

It's ignored in `docker stack deploy` because it doesn't make sense to set the container name when you're creating services not containers.

joaofnfernandes:
@londoncalling sorry for not being clear enough on #2950. The `container_name` is supported in the V3 schema, so I wasn't suggesting we removed it from `compose/compose-file`. I was suggesting that we should include in that article that `docker stack deploy` doesn't support that keyword.

Since `docker-compose up` and `docker stack deploy` only support a subset of the V3 schema, I was also suggesting that we should make clear which properties are supported by which tool. Something like this:

---
### build
Configuration options that are applied at build time.

< sample >

| docker-compose up | docker-stack deploy |
|:--------------------:|:----------------------:|
|  |  | 
---
I'm on the fence about doing this on the `compose/compose-file` article, since we're describing the spec there, and it's up to the tools (docker stack, docker-compose) to implement support for the spec. So maybe the right place to document what they implement or not should be in the tool documentation.

But, most people use `compose/compose-file` as a reference when they are writing compose files, and that article ranks high on Google, so we would be servicing our users better if the information was there, and not hidden in the tool docs.

londoncalling:
@dnephin Thanks I was confused by the listing in https://github.com/moby/moby/blob/master/cli/compose/types/types.go#L34:

```
// DeprecatedProperties that were removed from the v3 format, but their
// use should not impact the behaviour of the application.
var DeprecatedProperties = map[string]string{
	"container_name": "Setting the container name is not supported.",
	"expose":         "Exposing ports is unnecessary - services on the same network can access each other's containers on any port.",
}
```

I've added back `container_name` and simply added the standard not with regard to `docker stack deploy`.

londoncalling:
LGTM, and for now I think this satisfies the original issue, better summarizes `docker stack deploy` unsupported options. Ping me on this thread if you want more changes, but merging to get it published.

joaofnfernandes:
LGTM, and we can always iterate later 😀
It's true that `docker stack` deploy supports a lot of those properties, but V3 also has properties that `docker-compose up` doesn't support. `deploy` is one that comes to my mind.

Adding a table like that would make things a bit more explicit. But again, we don't need to do that on this PR.

londoncalling:
@joaofnfernandes , true, but I think that's covered on the description of `deploy` at: https://docs.docker.com/compose/compose-file/#deploy :

"Specify configuration related to the deployment and running of services. This only takes effect when deploying to a [swarm](https://docs.docker.com/engine/swarm/) with `docker stack deploy`, and is ignored by `docker-compose up` and `docker-compose run`."

PR #3009, opened 2017-04-23
Add NTT Communications Enterprise Cloud machine driver to the list of 3rd-party docker machine drivers

mittz:
### Proposed changes
Added [NTT Communications Enterprise Cloud](https://ecl.ntt.com/en/) driver to the list of 3rd-party docker machine drivers.


mittz:
it's wrong order in the list, i re-make PR. so i close this...

PR #3026, opened 2017-04-24
Updates to multi-process container topic

### LINKED ISSUE #758: Consider removing the topic "Use Supervisor with Docker"
[engine/admin/using_supervisord.md](https://docs.docker.com/engine/admin/using_supervisord/)

This seems to be a total anti-pattern and very much against separation of concerns. I'm quite surprised to see it "officially" endorsed through the documentation.

I appreciate that sometimes folks really wish to do this, but it's often for the wrong reasons and almost never really required. Have you considered more thoroughly educating them in the introduction paragraph to explain the implications of doing this, or more importantly, benefits of properly separating concerns and how they might well be served by Compose?

mstanleyjones:
### Proposed changes

We should more strongly recommend against running multiple processes in a container

We should not specifically advocate for `supervisord` either.

Fixes #758.


engine/admin/multi-process_container.md
```diff
@@ -0,0 +1,89 @@
...
+keywords: docker, supervisor, process management
+redirect_from:
+- /engine/articles/using_supervisord/
+- /engine/admin/using_supervisord/
+title: Run multiple processes in a container
+---
+
+A container's running process is usually the `ENTRYPOINT` and/or `CMD` at the
+end of the `Dockerfile`. It is generally recommended that you separate areas of
+concern by using one container per process, and connect those containers using
```
thaJeztah: We may want to tweak a bit w.r.t. "one process per container"; it may be a bit more nuanced; it's a single "service" per container. That service may fork into multiple _processes_ (for example, Apache starts multiple worker processes). It's ok to have multiple processes, but the concept of Docker is to not have a container be responsible for multiple services / tasks.

Slightly orthogonal: 

The container's main process is responsible for managing all processes that it started. In _some_ cases, the main process isn't well-designed, and doesn't handle "reaping" of child processes well when the container exists. If you're dealing with such a process, you can use the `--init` option, which inserts a tiny init-process in the container as the main process, and handles "reaping" of all processes when the container exits. There's no reason to use a full-fledged init process (Sysvinit, upstart, systemd) for that.

engine/admin/multi-process_container.md
```diff
@@ -0,0 +1,89 @@
...
+If you need to run more than one process within a container, you can accomplish
+this in a few different ways:
+
+- Chain multiple commands together in the `CMD`:
+
+  ```conf
+  FROM ubuntu:latest
+  COPY my_first_process my_first_process
+  COPY my_second_process my_second_process
+  CMD ./my_first_process && ./my_second_process
```
thaJeztah: This probably won't work, because it runs the second process _after_ the first one exits; you probably have to daemonize the first one (`./foobar &`) to keep it running in the background.

I'm actually not sure we should recommend this, as this is a DYI init process, without reaping, and without any management of (e.g.) logging, startup ordering, keeping processes up, etc.

I'd prefer to keep the supervisord page as a _possible_ option to run multiple processes, for those that need it, as long as we reiterate that generally  running a single service in a container is the way to go.

Probably we can come up with some examples how to split those services over multiple containers (e.g. have a `sshd` service container that allows you to rsync files in a volume that's used by another container)

mstanleyjones:
I added your supplementary info, removed the example you didn't like, tweaked the topic name.

thaJeztah:
Thanks! The introduction looks better; Let me call in @tianon for help as well, he may have good suggestions for describing this :heart:

engine/admin/multi-service_container.md
```diff
@@ -0,0 +1,91 @@
...
+  status=$?
+  if [ $status -ne 0 ]; then
+    echo "Failed to start my_second_process: $status"
+    exit $status
+  fi
+
+  # Naive check to see if either of them exited immediately
+  PROCESS_!_STATUS=$(ps aux |grep -q my_first_process)
+  PROCESS_2_STATUS=$(ps aux |grep -q my_second_process)
+  if [ $PROCESS_!_STATUS || $PROCESS_2_STATUS ];
```
tianon: I think these `PROCESS_!_STATUS` instances were intended to be `PROCESS_1_STATUS`. :smile:

engine/admin/multi-service_container.md
```diff
@@ -0,0 +1,91 @@
...
+    echo "One of the processes has already exited."
+    exit -1
+  fi
+  ```
```
tianon: This script, as written, will exit immediately and thus PID 1 will exit and the two services started will be killed.  Is that the intention for this example?  The hacky way I usually see folks solve this is by putting this `ps` logic into a loop or using something silly like `tail -f` to just keep the container alive (`tail -f /dev/null` is the cheekiest example of "keep this container from dying" that I've seen to date :smile:).

engine/admin/multi-service_container.md
```diff
@@ -0,0 +1,91 @@
...
+
+  ```conf
+  FROM ubuntu:latest
+  RUN apt-get update && apt-get install -y supervisor
+  RUN mkdir -p /var/log/supervisor
+  COPY supervisord.conf /etc/supervisor/conf.d/supervisord.conf
+  COPY my_first_process my_first_process
+  COPY my_second_process my_second_process
+  CMD ["/usr/bin/supervisord"]
+  ```
```
tianon: Definitely a fan of how lightweight this example is -- basically, if you want to run multiple processes, that's fine, here's an example of one application that can do so, but getting a "good" configuration for it is Up To You (Docker's not going to help you figure that out because it's Bad). :+1: :heart:

mstanleyjones:
Thanks @tianon I have updated the instructions and made the process check live in a loop, and also left some comments about how painful and hacky that is. ;)

engine/admin/multi-service_container.md
```diff
@@ -0,0 +1,97 @@
...
+  status=$?
+  if [ $status -ne 0 ]; then
+    echo "Failed to start my_second_process: $status"
+    exit $status
+  fi
+
+  # Naive check runs checks once a minute to see if either of the processes exited.
+  # This illustrates part of the heavy lifting you need to do if you want to run
+  # more than one service in a container. The container will exit with an error
+  # if it detects that either of the processes has exited.
```
tianon: This is awesome :heart_eyes:

PR #3030, opened 2017-04-24
Remove the restart:always from engine/swarm/secrets.md

### LINKED ISSUE #3029: Swarm secret use restart:always in a stack file and it's not supported
### Problem description

@joaofnfernandes: I've noticed that the snippet is using restart:always, and docker stack deploy will complain that the restart keyword is not supported and ignores it. We can fix that on a subsequent PR.

### Problem location
engine/swarm/secrets.md

jmaitrehenry:
### Proposed changes

Remove the restart:always from engine/swarm/secrets.md
### Related issues

Fix #3029

### DIFF
--- a/engine/swarm/secrets.md
+++ b/engine/swarm/secrets.md
@@ -857,8 +857,7 @@ services:
      image: mysql:latest
      volumes:
        - db_data:/var/lib/mysql
-     restart: always
-     environment:      
+     environment:
        MYSQL_ROOT_PASSWORD_FILE: /run/secrets/db_root_password
        MYSQL_DATABASE: wordpress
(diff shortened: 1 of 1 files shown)


joaofnfernandes:
Thanks for the fast turnaround on this! Looks good.
Relates to #2993.

As a best practice for the future, don't include issue numbers in your commit messages, but do it in the PR. This makes it easier to navigate between issues and PRs in Github, and also ensures that if your commit gets pushed to other repo, you don't close issues on that repo by accident.

jmaitrehenry:
👍 will do in the futur! Thanks for the tips.

PR #3032, opened 2017-04-24
WIP: Consistent device usage in devicemapper examples

LongLiveCHIEF:
### Proposed changes

The existing docs use inconsistent names for the docker volume, and the thinpool logical volume names.  I've updated the docs to use `docker` consistently as the volume group name, and `docker/thinpool` as the logical volume in all examples.

this makes the docs easier to comprehend, since users don't have to go through multiple times to realize the names of volumes keep switching halfway through the example.

### DIFF
--- a/engine/userguide/storagedriver/device-mapper-driver.md
+++ b/engine/userguide/storagedriver/device-mapper-driver.md
@@ -401,14 +401,16 @@ NAME			   MAJ:MIN RM  SIZE RO TYPE MOUNTPOINT
 xvda			   202:0	0	 8G  0 disk
 └─xvda1			   202:1	0	 8G  0 part /
 xvdf			   202:80	0	10G  0 disk
-├─docker-thinpool_tmeta		   253:0	0	90G  0 lvm
-│ └─docker-202:1-1032-pool 253:2	0	10G  0 dm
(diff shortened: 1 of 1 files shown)


mstanleyjones:
How is this coming along?

mstanleyjones:
This is still listed as WIP and has a merge conflict. Any update?

LongLiveCHIEF:
There was a recent update to [projectatomic/container-storage-setup](https://github.com/projectatomic/container-storage-setup), and I've been working through issues in consistency between manual setup using the current docs, running the latest css, and some issues with selinux enabled hosts.

It's already been a long week with this one, but I hope to be able to take definitive action on this PR by the end of the week.

LongLiveCHIEF:
The issue mentioned in css: [passing -n ftype=1](https://github.com/projectatomic/container-storage-setup/issues/194) hits the nail on the head with the selinux issues. 

Especially running `dind` scenarios, volume creation fails within an outer contain because the given root path for docker on most CentOS/RHEL distro's [isn't formatted correctly when left to docker to create the directory](https://docs.docker.com/engine/userguide/storagedriver/overlayfs-driver/#prerequisites).

LongLiveCHIEF:
After reviewing the changes, I don't think there's any point in rebasing. I think the upstream changes  covered everything I was going to do with this PR.

PR #3045, opened 2017-04-25

scjane:
Repair the inconsistency between subject and pronoun.


johndmulhausen:
Thanks, @scjane :)

PR #3046, opened 2017-04-25
Update index.md

scjane:


johndmulhausen:
I think, post conflict-resolution, that what we have is actually a better read. The comma encourages a pause in the reader's mind. Thanks, though @scjane !

PR #3058, opened 2017-04-25
updating docker version command

londoncalling:
To get the output shown in the example, you need to run `docker --version` (`docker version` gives  you the long form version output)

cc: @johndmulhausen


londoncalling:
LGTM

PR #3083, opened 2017-04-26
Make notes conform to the new style

### LINKED ISSUE #2781: Make notes conform to new style
Follow-up of #2710. Now that have different css classes for notes and warnings, we should clean up existing notes to make them conform to the new style.

You can see an example of the new style at #2710

shubheksha:
### Related issues (optional)
Partially fixes #2781


shubheksha:
@joaofnfernandes, can you take a quick look?

shubheksha:
@joaofnfernandes, I see lots of warnings in the format that's there in this file: https://github.com/docker/docker.github.io/blob/master/datacenter/dtr/2.0/reference/reconfigure.md
Should they use the new style too?

joaofnfernandes:
`datacenter/*/*/reference` are generated from the source code, but the repo for that product is private. So no need to update those files for now.
@shubheksha just to set expectations, we're wrapping up the hackathon and counting points, so this contribution probably won't get you any hackathon points. But a huge thanks for continuing to contribute after the hackathon.

shubheksha:
@joaofnfernandes, I'll revert those changes.
And I didn't expect them to be counted either, just had some free time on my hands 😄 
Just to clarify before I make more changes: for just plain old notes, the only change that needs to be made is to introduce a line break after `**Note:**`, right?

swarm/reference/create.md
```diff
@@ -15,6 +15,8 @@ Later, when you use [`manage`](manage.md) or [`join`](join.md) to create Swarm m
 
 Some documentation also refers to the discovery token as a *cluster_id*.
 
-> Warning: Docker Hub's hosted discovery backend is not recommended for production use. It’s intended only for testing/development.
+> **Warning**:
```
joaofnfernandes: Sorry for not having noticed this before. #2795 is now merged, and it will automatically adds"warning", "info" or "important" before the text. If you look at the result of [your change here](https://deploy-preview-3083--docker-docs.netlify.com/swarm/reference/create/), you'll notice that the note now says "Important: warning: xxxx ".

So to start fixing #2781, we need to do a bit more than just adding a line break. The idea is to make notes a bit more meaningful by adding a small title. Check this example from the bootstrap documentation where they're doing something like this.

screen shot 2017-04-28 at 14 59 44

For this particular article you need to do something like

```
> Not for production
> Docker Hub's hosted discovery backend is not recommended for production use. It’s intended only for testing/development.
```
{:.warning}

That will give you

screen shot 2017-04-28 at 15 02 48

joaofnfernandes:
@shubheksha I'm closing this PR since I didn't get a reply from you. Feel free to open a new one if you feel like tackling the notes style.

PR #3087, opened 2017-04-26
Recommend using services instead of network aliases for HA

### LINKED ISSUE #775: Correct/improve description of name resolution for containers that share the same alias (networking example)
[engine/userguide/networking/work-with-networks.md](https://docs.docker.com/engine/userguide/networking/work-with-networks/)

Docker version: 1.12.3

As the docs says:
>When multiple containers share the same alias, name resolution to that alias will happen to one of the containers (**typically the first container that is aliased**). 

While it is not the case as I tested:
```
>docker network create isolated_nw

>docker run -itd --network=isolated_nw --network-alias=app --name=container1 busybox
>docker run -itd --network=isolated_nw --network-alias=app --name=container2 busybox
>docker run -itd --network=isolated_nw --name=container3 busybox

>docker attach container3
#ping -w4 app
PING app (172.18.0.3): 56 data bytes
64 bytes from 172.18.0.3: seq=0 ttl=64 time=0.091 ms
...

>docker restart container1

>docker attach container3
#ping -w4 app
PING app (172.18.0.2): 56 data bytes
64 bytes from 172.18.0.2: seq=0 ttl=64 time=0.091 ms
...
```

So, name resolution is not resolved to the first alias container but the lastest alias container.

mstanleyjones:
### Proposed changes

#775 mentioned some confusion about network aliases. This seems to have already been fixed,
but I took the opportunity here to add a note recommending the use of services instead of
network aliases for this functionality.

Fixes #775

### DIFF
--- a/engine/userguide/networking/work-with-networks.md
+++ b/engine/userguide/networking/work-with-networks.md
@@ -688,7 +688,7 @@ all environment variables to the linked container, but the `docker network` comm
 has no equivalent. When you connect a container to a network using `docker network`, no
 environment variables can be dynamically shared among containers.
 
-#### Understanding network-scoped aliases
+#### Use network-scoped aliases
 
 Legacy links provide outgoing name resolution that is isolated within the
 container in which the alias is configured. Network-scoped aliases do not allow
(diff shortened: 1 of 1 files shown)


engine/userguide/networking/work-with-networks.md
```diff
@@ -787,6 +787,12 @@ containers connected to that network can access the alias.
 Multiple containers can share the same network-scoped alias within the same
 network. This example illustrates how this works.
 
+> **Note**: Network aliases are a legacy networking feature. Swarm services
```
thaJeztah: Network aliases in itself are not legacy, they're perfectly ok to use (i.e. if my container is named "foobar", but I want to access it by different name(s).

For load-balancing, dnsrr may not always lead to the expected result (e.g. nginx will cache the result so may still always connect to the same container)

mstanleyjones: I made some edits. Maybe @johndmulhausen can work with you on further tweaks needed to get this in.

mstanleyjones:
Addressed the feedback, so merging this for now.

PR #3089, opened 2017-04-26
Fix yml indention

hacfi:
### Proposed changes

Reading the docs at https://docs.docker.com/compose/compose-file/#restartpolicy I noticed the indention was wrong so I fixed it.


johndmulhausen:
Yes, thank you @hacfi !

PR #3090, opened 2017-04-26
Update docker-compose index.md

andyneff:
### Proposed changes

There were a number of things wrong with the `restart` definition section

1. The volumes **note** came after the restart section rather than before
2. The restart command used array notation rather than map. This is confusing to those unfamiliar with yaml and inconsistent with the rest of the documentation
3. Using `docker-compose` myself, saying `restart: no` does not work, but `restart: "no"` does. 

```
ERROR: The Compose file './docker-compose.override.yml' is invalid because:
services.letsencrypt.restart contains an invalid type, it should be a string
```


johndmulhausen:
Thanks, @andyneff, these are great fixes. 👍

PR #3093, opened 2017-04-27
Wrong order of docker compose restart section

Vanuan:
### Proposed changes
Fix restart being placed in the middle of volume section


joaofnfernandes:
Thanks for the contribution 🐳

PR #3097, opened 2017-04-27
Remove copy error syslog TCP+TLS options example

SamLR:
The syslog TCP+TLS example suggests that the relevant options are `syslog-tls-ca-cert=syslog-tls-ca-cert=...`, `syslog-tls-cert=syslog-tls-ca-cert=...` and `syslog-tls-key=syslog-tls-ca-cert=...`. This corrects this example.


johndmulhausen:
Yeesh, looks like someone's copy/paste didn't work out. Thanks, @SamLR, nice catch.

PR #3100, opened 2017-04-27
Update part2.md

99887710:


johndmulhausen:
Shouldn't be necessary. Also, I get this message on Docker for Mac right now:

```
$ docker-machine ip
Error: No machine name(s) specified and no "default" machine exists
```

PR #3111, opened 2017-04-28
Update index.md

scjane:


johndmulhausen:
@scjane I like the second change, but can we leave the use of "their" in there? I think it reads better.

PR #3120, opened 2017-04-28
Update how users download the db file for scanning

sarahpark:


toli:
do you want to include the screenshots that we've been sending through? or do you think it's unnecessary?
maybe just the last screenshot?
image

joaofnfernandes:
Yup, screenshots make it easy to scan the content and find what you need without wasting time. Can you include on @sarahpark ?

sarahpark:
@joaofnfernandes just added screenshots in another commit

PR #3123, opened 2017-04-29
Fix test DB creation in Compose's database.yml

jmdeldin:
### Proposed changes

Without the host and credentials, the `myapp_test` database is not created, and instead, the user receives a PG error and a long backtrace. This commit moves the credentials to a default anchor to share across
development and test environments, so readers can follow the tutorial without a hitch.


mstanleyjones:
@shin- PTAL

mstanleyjones:
Please fix conflicts. @shin- ping :)

shin-:
I'm no rails expert but I think that's correct. LGTM

PR #3132, opened 2017-05-01
Fix overflow issue

jsouth:
http://recordit.co/lATLKXHVVx

http://recordit.co/np2MjR6Yj3

- fix menu overflow (version dropdown)
- add/update dockercon banner


johndmulhausen:
Wonderful, thank you @jsouth !

PR #3144, opened 2017-05-03
Update index.md

scjane:
Should "a" be replaced with "an" here?


johndmulhausen:
Sure, yep. Thanks, @scjane :)

PR #3153, opened 2017-05-03
Typo in redis port on docker-compose.yml example

gbaims:
### Proposed changes

Redis service is forwarding to the wrong port, i believe it was a typo.


johndmulhausen:
You're right, thank you @gbaims

PR #3171, opened 2017-05-05
Update troubleshoot-batch-jobs.md

scjane:
Is it "worker_resurrection"?


johndmulhausen:
Had to check the code; it is indeed. Thanks @scjane :)

PR #3183, opened 2017-05-06
Update part2.md

ruffiem:
### Proposed changes

Pip sample won't install on Mac without the -r flag.


joaofnfernandes:
Makes sense! Thanks for the contribution @ruffiem, let's ship it 🚀

PR #3187, opened 2017-05-06
reduce padding in nav, space items, remove font weight

anagio:
This adds a custom break point for small mobile devices and fixes the nav so the last menu item does not wrap to a new line. Viewing the nav on a Samsung Galaxy S5 the `Samples` menu item wraps to a new line and isn't visible. The font weight was changed so the menu would fit in Firefox without changing the font size.

### DIFF
--- a/_scss/_mobile.scss
+++ b/_scss/_mobile.scss
@@ -198,6 +198,34 @@
     }
 }
 
+@media only screen
+  and (min-device-width: 320px)
+  and (max-device-width: 372px)
+  and (orientation: portrait) {
+    .nav-secondary-tabs {
+      > .container-fluid {
+        padding-left: 5px;
+        padding-right: 5px;
+      }
+    }
+    .tabs {
+      display: -webkit-box;
+      display: -moz-box;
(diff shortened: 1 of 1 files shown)


mstanleyjones:
@jsouth PTAL

johndmulhausen:
This would take the majority of mobile users out of the double-row nav situation (where tabs are visible) and into a single-row nav (where all tabs are under a hamburger). I think this is a bit problematic. What do you think, @jsouth ?

anagio:
To be clear on what this fixes. It adjusts the nav so the `Samples` menu item is in view and not overlapping like seen here.

Another solution would be to place one ore more of the horizontal menu items in to the drop down on mobile. One method would be to simply add additional `li` tags to the drop down and hide them on desktop.

image

jsouth:
@johndmulhausen - I agree. These links should always be visible and not folded into the mobile menu, and always 1 line.

@anagio - This solution is working and resolves the main issue ( samples text wrapping ).

I vote we add this fix for now, and we can refine more.  :)

PR #3200, opened 2017-05-09
Fix IP address in getting started part 5

dkong:
### Proposed changes

Fix getting started part 5 doc.

The docs incorrectly state how to access the web page app.  It should be accessed by either node's IP address and NOT localhost.


joaofnfernandes:
Thanks for the contribution @dkong! Let's ship it 🚀

PR #3204, opened 2017-05-09
Code example in "mode" section is messed up

alesguzik:
For some reason example in "mode" section doesn't render as code. Only difference that I've noticed from other code blocks is 2 newlines between it and preceding text, so probably this should fix formatting. See https://docs.docker.com/compose/compose-file/#mode


joaofnfernandes:
I think #3180 already fixed this problem. Can you still reproduce this?

mstanleyjones:
It's working now.

PR #3211, opened 2017-05-09
Grammar change of is to are

TheOtherRealm:
Grammar change of is to are


johndmulhausen:
"Access ... is" is correct.

PR #3215, opened 2017-05-10
dotnet build fails without sql server dependency

burnettk:
the updates make the docs for building a dot net core app work for me.


burnettk:
thanks, @mstanleyjones ! where's my auto-markdown-preview, github!

PR #3217, opened 2017-05-10
Service create vs run

Vanuan:
This compose section should reference docker service create, not docker run options
https://github.com/moby/moby/issues/32844


Vanuan:
Depends on this: https://github.com/moby/moby/pull/32631/files

PR #3220, opened 2017-05-10
17.06: improve the Docker for Mac networking workaround

djs55:
### Proposed changes

In Docker for Mac 17.05 and earlier, there is no easy way for a container to contact a service running on the host (such as a local HTTP proxy) without knowing the host's IP addresses. The workaround in the docs is to add a new IP alias and use that.

In the upcoming 17.06 release of Docker for Mac, there is a special internal DNS name `docker.for.mac.localhost`. This patch replaces the IP alias workaround with a note about the new DNS name.

This only applies to Docker for Mac 17.06.

### DIFF
--- a/docker-for-mac/networking.md
+++ b/docker-for-mac/networking.md
@@ -78,11 +78,10 @@ There are two scenarios that the above limitations will affect:
 
 #### I want to connect from a container to a service on the host
 
-The Mac has a changing IP address (or none if you have no network access). Our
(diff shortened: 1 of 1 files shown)


mstanleyjones:
I tried changing the base but you need to rebase it locally.

djs55:
@mstanleyjones I've rebased onto `vnext-engine` and changed the base of the PR in github -- hopefully this looks ok.

Issue #3206, opened 2017-05-09
Generate compliance documentation


anweiss:
### Problem description

We currently maintain a host of compliance documentation in our repo here -> https://github.com/docker/compliance. The goal is to autogenerate .md docs for docs.docker.com from the .yaml-based compliance content that adheres to the OpenControl [schema](https://github.com/opencontrol/schemas). Sample compliance file [here](https://github.com/docker/compliance/blob/master/opencontrol/components/DockerEE/component.yaml).

![image](https://cloud.githubusercontent.com/assets/2326106/25862131/f6116ad4-34b4-11e7-98fc-813a801bbec0.png)

We should take an approach that is similar to that which the Moby upstream autogenerates the command-line reference docs -> https://github.com/moby/moby/tree/17.03.x/docs/reference/commandline.

component.yaml -> ... -> .md for docs.docker.com

anweiss:
Tracking in https://github.com/docker/compliance/tree/docs/docs

johndmulhausen:
Thanks for getting this in, @anweiss !

anweiss:
No prob!

PR #3223, opened 2017-05-10
Include compliance documentation and required data sources

anweiss:
### Proposed changes

Include compliance documentation and auto-generated NIST 800-53 reference based on OpenControl-formatted compliance content at https://github.com/docker/compliance. Currently, the yaml data sources are manually copied to `/_data/compliance/`, but once a more formalized process for pulling in external content is established, this can be modified.

### Related issues

Address #3206.


mstanleyjones:
Please don't use `docker` as a keyword. :) Otherwise this LGTM but @johndmulhausen should review.

anweiss:
Ah, ok. No prob. Will update.

compliance/fedramp.md
```diff
@@ -0,0 +1,11 @@
+---
+description: fedramp
+keywords: standards, compliance, security, documentation
```
mstanleyjones: I'm not sure about using the keyword "documentation" in the docs either. WDYT @johndmulhausen

compliance/fedramp.md
```diff
@@ -0,0 +1,11 @@
+---
+description: fedramp
```
mstanleyjones: These descriptions could use some work. It should be a sentence or two, I think.

compliance/fedramp.md
anweiss: I simply referenced some of the existing .md files that have those keywords. Will change as necessary.

mstanleyjones:
LGTM now, even with `documentation` in keywords... I'll let @johndmulhausen have the final say.

johndmulhausen:
Improvements I think are worth making:

- Remove "documentation" as a keyword
- Instead of using includes, just use a layout; see for example how the files under /samples/ use a different layout that processes the information. It creates cleaner shell pages than having to define all these include statements. We should be doing that in the CLI docs but we aren't yet
- These tabs aren't syncing with each other and they probably should: https://deploy-preview-3223--docker-docs.netlify.com/compliance/reference/800-53/cp/#cp-1-contingency-planning-policy-and-procedures
- It seems like the opening salvo paragraphs are a bit jumbled. For example take this: https://deploy-preview-3223--docker-docs.netlify.com/compliance/reference/800-53/ia/#ia-1-identification-and-authentication-policy-and-procedures
... Is it really supposed to use what looks like "put the real information here" markers like `[Assignment: organization-defined frequency]`? Are these from a template that isn't filled out yet? Also, wouldn't this be better formatted as a list with proper indenting? There are both alpha and numeric list items whizzing by in an unbroken stream of text making a big word blob.

anweiss:
Thanks @johndmulhausen ... working on the cleanup now. The NIST 800-53 descriptions are currently being parsed from the included `_data/compliance/nist80053.yml` file, which is based on the OpenControl schemas defined [here](https://github.com/opencontrol/schemas). This was mainly to simplify the parsing code against one schema instead of multiple schemas. However, given that the schema has yet to be finalized and to clean up the description layout, I'll go ahead and instead parse the descriptions based on the original NIST 800-53 Rev 4 xml definition [here](https://nvd.nist.gov/download/800-53/800-53-controls.xml). This should give us better separation of paragraphs and lists.

As far as the `[Assignment: organization-defined frequency]` markers, these are actually part of the NIST 800-53 specification itself and whose contents are replaced by either the organization's specific requirement, or that of another standard that inherits the 800-53 controls (e.g. FedRAMP). I can clarify this in the index docs for the reference.

johndmulhausen:
@anweiss Any news here?

anweiss:
Working through it. Backlogged at the moment.

mstanleyjones:
Ping :)

anweiss:
Sorry, haven't had a chance to look at it. Just pushed some new commits, but running in to some issues with refactoring using layouts per @johndmulhausen's recommendation.

anweiss:
For some reason, I'm unable to pull any of the Liquid variables/data from within the `nist80053.html` layout that I created. It inherits from the `docs` layout. Also getting an Angular module injection error:

```
jquery.min.js:2 Uncaught Error: [$injector:modulerr] http://errors.angularjs.org/1.5.11/$injector/modulerr?p0=Docker&p1=Error%3A…%2Fdbepenphjfofmnjmlacfcdehikakmaap%2Fjs%2Flib%2Fangular.min.js%3A21%3A332)
    at angular.min.js:6
    at angular.min.js:40
    at q (angular.min.js:7)
    at g (angular.min.js:39)
    at fb (angular.min.js:43)
    at c (angular.min.js:21)
    at Gc (angular.min.js:21)
    at se (angular.min.js:20)
    at HTMLDocument.<anonymous> (angular.min.js:323)
    at j (jquery.min.js:2)
```

mstanleyjones:
@johndmulhausen PTAL

anweiss:
Squashed in to single commit

mstanleyjones:
How are you feeling about this PR, @anweiss ?

anweiss:
We've decided to externalize these docs in the compliance repo -> https://github.com/docker/compliance ... will create a new PR that will support fetching these, in a manner similar to that which the CLI docs are pulled in

Issue #3178, opened 2017-05-05
Improve docs on logging options in Compose file reference


londoncalling:
### Problem description

Document logging better, especially per this [comment](https://github.com/docker/compose/issues/1083#issuecomment-299342320)  on how to limit log sizes

### Problem location

https://docs.docker.com/compose/compose-file/#logging

### Related 

[Issue 1083 in docker/compose on Feature: Ability to clear log history](https://github.com/docker/compose/issues/1083)

### Reviewers, (help w/improvements ?)

@thaJeztah @tgjamin @dnephin @shin-

tgjamin:
Pull request adding documentation; #3243

PR #3243, opened 2017-05-11
Add max-size, max-file logging options documentation

tgjamin:
### Proposed changes

Add documentation for logging options max-size and max-file that were previously undocumented. Formatting was checked locally with docker-compose build.

### Related issues (optional)

Fixes: #3178

### DIFF
--- a/compose/compose-file/index.md
+++ b/compose/compose-file/index.md
@@ -764,6 +764,27 @@ Logging options are key-value pairs. An example of `syslog` options:
     options:
       syslog-address: "tcp://192.168.0.42:123"
 
+To limit the amount of logs stored there are two options key-value pairs:
+
+    options:
+      max-size: "200k"
+      max-file: "10"
+
+This example would store log files until the reach a `max-size` of 200kB, and then rotate them.
+The amount of individual log files stored is specified by the `max-file` value. As logs grow
+beyond the max limits, older log files are removed to allow storage of new logs.
+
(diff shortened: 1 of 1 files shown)


londoncalling:
@tgjamin thanks! LGTM

Ping, @shin- @dnephin @nathanleclaire  @thaJeztah for a second look, then we can merge it.

londoncalling:
@tgjamin LGTM, I'm going to merge this and make a few minor copyedits after. Thanks much!

compose/compose-file/index.md
```diff
@@ -764,6 +764,27 @@ Logging options are key-value pairs. An example of `syslog` options:
     options:
       syslog-address: "tcp://192.168.0.42:123"
 
+To limit the amount of logs stored there are two options key-value pairs:
```
thaJeztah: We should be careful documenting this as-is, because these options depend on the logging driver that's in use. For example, `max-file` and `max-size` are only used for the default (`json-file`) logging driver.

I _do_ think this information is useful but;

- we should point out that these options are specific to that driver
- more clearly mention that these options _differ_ per driver, and refer to https://docs.docker.com/engine/admin/logging/overview/ - I see it's being referred to above as well

thaJeztah: ping @londoncalling ^^

londoncalling:
@thaJeztah thanks, I'll create another PR with these notes, and maybe you and @tgjamin can review and / or add to it as appropriate

PR #3251, opened 2017-05-12
Update part2.md

kajbonfils:
### Proposed changes

Tried to follow the demo, but failed when I went to another machine to fetch the image.

Messed around a bit and noticed that I missed the imagename


johndmulhausen:
You must not have uploaded the image anywhere. This probably succeeded because `friendlyhello` was present on your local machine. :)

PR #3258, opened 2017-05-14
Fix for incorrect repo URL

anokun7:
Added "/rhel" in the repo url's.

### DIFF
--- a/engine/installation/linux/rhel.md
+++ b/engine/installation/linux/rhel.md
@@ -84,7 +84,7 @@ the repository.
       [prerequisites](#prerequisites).
 
       ```bash
-      $ sudo sh -c 'echo "<DOCKER-EE-URL>" > /etc/yum/vars/dockerurl'
+      $ sudo sh -c 'echo "<DOCKER-EE-URL>/rhel" > /etc/yum/vars/dockerurl'
       ```
 
     - Store your RHEL version string in `/etc/yum/vars/dockerosversion`.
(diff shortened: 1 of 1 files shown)


mstanleyjones:
This change is incorrect. When you sign up for a Docker EE subscription for RHEL, you get a specific URL for that subscription, that only has RHEL bits in it. Please set up an EE trial subscription if you would like to double-check this.

anokun7:
@mstanleyjones 
Maybe it's a problem on store then, but I did double check this. On store, my specific URL ends in the unique id (`https://storebits.docker.com/ee/linux/sub-<unique-id>`). Also there is a direct discrepancy between instructions on the store versus those in docs for RHEL:

In store (`/rhel` is included):
screen shot 2017-05-15 at 1 09 51 pm

In docs, (no `/rhel`, even though this is a specific docs page for installing on RHEL):
screen shot 2017-05-15 at 1 10 22 pm

mstanleyjones:
The confusion here is that you are using Docker EE for Linux, rather than Docker EE for RHEL. In your case, you get access to all of the Linux bits and your directory structure is a little different. This is a new and not widely-known offering, and I need to update the docs to account for it. But we expect most EE subscribers to subscribe to a specific distribution's bits, in which case the current instructions are correct.

anokun7:
Understood, but that was the only content / link on my store page.

 The store documents the pre-req steps for install & then hands off to the docs for RHEL, it's probably not the case on the distro specific store pages. For other distros, the store does document the command to use to install docker-ee. 

Please feel free to close this PR if it is no longer making any sense to keep open.

mstanleyjones:
I believe that the Store docs tell you to follow the EE instructions but adjust the repo URL. :) I'm sorry for the confusion, and we'd be open to any ideas to make it clearer.

PR #3271, opened 2017-05-15
removed old content from liquid admonitions include file

londoncalling:
### What's changed 

- Removed content from this older version of the admonitions include file because @johndmulhausen since created a more complete version of this solution in #3253. If we decide to use a Liquid solution, or even keep it as an alternative to FontAwesome icons, we should use John's version of the includes file, not mine. So, this will make it easier to merge that newer file, if needed. Otherwise, it's a placeholder which can be used or deleted later. Either way, makes no sense to keep the file as it was with content that didn't fully solve the problem.

### Related

#3253 

### Reviewers fyi

@johndmulhausen @joaofnfernandes @mstanleyjones 

### DIFF
--- a/_includes/content/admonitions/notes.html
+++ b/_includes/content/admonitions/notes.html
@@ -1,12 +1,3 @@
-{% if include.type=="note" %}
-{% assign glyphClass="glyphicon glyphicon-ok-sign" %}
-{% assign blockquoteClass="note-vanilla" %}
-{% else if include.type=="warning"%}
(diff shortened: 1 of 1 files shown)


londoncalling:
LGTM

joaofnfernandes:
Sorry for the late reply, but if we're removing the content on this file, can't we just delete the while file instead?

londoncalling:
@joaofnfernandes yes, I'll delete it, I was being ultra careful since I liked being reminded that there is this other option. I'll delete it.

PR #3275, opened 2017-05-15
fixed code example in Compose startup order per feedback

### LINKED ISSUE #3273: Code Example Clipping in Compose StartUp Order docs
[compose/startup-order.md](https://docs.docker.com/compose/startup-order/), CC @londoncalling 

The last code block on this page doesn't support overrun scrolling and results in a the example being clipped visibly - you can see the entire example in source-code though. 

I think that the markup should be updated to:

```html
<div class="highlighter-rouge">
  <pre class="highlight">
    <code>command: ["./wait-for-postgres.sh", "db", "python", "app.py"]</code>
  </pre>
</div>
```

Rather than what it is currently:

```html
<p>You can use this as a wrapper script as in the previous example, by setting
<code class="highlighter-rouge">command: ["./wait-for-postgres.sh", "db", "python", "app.py"]</code>.</p>
```

Tested in Chrome 58.0.3029.110

londoncalling:
### What's changed

Fixed code example to display properly 

### Related issues

Fixes #3273 

### Reviewers

@mikeantonelli

### DIFF
--- a/compose/startup-order.md
+++ b/compose/startup-order.md
@@ -71,8 +71,11 @@ script:
         >&2 echo "Postgres is up - executing command"
         exec $cmd
 
-    You can use this as a wrapper script as in the previous example, by setting
-    `command: ["./wait-for-postgres.sh", "db", "python", "app.py"]`.
+    You can use this as a wrapper script as in the previous example, by setting:
+
+    ```none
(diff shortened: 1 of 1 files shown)


mikeantonelli:
Looks good @londoncalling

PR #3276, opened 2017-05-15
Updating docs about storage drivers

### LINKED ISSUE #1062: Overlay storage driver documentation should mention compatibility / prerequ
[engine/userguide/storagedriver/overlayfs-driver.md](https://docs.docker.com/engine/userguide/storagedriver/overlayfs-driver/)

This came up in the discussion on https://github.com/docker/docker/issues/29873#issuecomment-271193629. The overlay storage driver documentation does not mention compatible backing filesystems. For example, if the partition is not formatted with `d_type` enabled, issues may occur (in future docker-engine will even refuse to start if this is not enabled).

Also see [requirements on RHEL 7.2](https://access.redhat.com/documentation/en-US/Red_Hat_Enterprise_Linux/7/html/7.2_Release_Notes/technology-preview-file_systems.html, https://github.com/docker/docker/issues/27358, and https://github.com/docker/docker/issues/10294

### LINKED ISSUE #1126: Clarification on storage driver not affecting volumes
This is a follow-up to https://github.com/docker/docker/issues/27215

It seems that __many__ people (going by both that issue, and the blogs and tutorials I've found online for configuring the devicemapper thinpooldev/LVM storage driver and for Docker best practices on RHEL-derivative OSes), myself included, misunderstand the distinction between storage drivers and volume drivers. More specifically, that most of us feel akin to [this comment](https://github.com/docker/docker/issues/27215#issuecomment-252405013):

> The current situation seems a bit anti-POLS to me, though; I told docker to use devicemapper, I expect it to do so.

Would it be possible to please add a clear, stand-out note to the [Docker and the Device Mapper storage driver](https://docs.docker.com/engine/userguide/storagedriver/device-mapper-driver/) documentation, and possibly also the [Select a storage driver](https://docs.docker.com/engine/userguide/storagedriver/selectadriver/) and [Understand images, containers, and storage drivers](https://docs.docker.com/engine/userguide/storagedriver/imagesandcontainers/) pages, clarifying this? Specifically, just clarifying that changing/configuring the storage driver only affects image layers, and not volumes? This seems to be a very common misconception...

### LINKED ISSUE #1444: Add additional information regarding Overlay(2) to documentation
### Problem description

During the investigation of an issue that I reported, it was noted that we are using a non-standard configuration.  In particular, we were running Overlay 2 on a non-standard file system configuration.  Nothing in the documentation indicated that there may be issues.  Therefore, it is suggested that some notes be added to make it clear that there may be some quirks.  The link to the original issue is below.

https://github.com/docker/docker/issues/30531

### Problem location

https://docs.docker.com/engine/userguide/storagedriver/overlayfs-driver/

### Suggestions for a fix

If there are known configurations that may cause issues or are recommended, such as those mentioned in the comments of the referenced bugs, please list them here.  If there would be a minimal list of Linux distros that should be documented, I'd suggest the following:

* RedHat/CentOS
* Debian
* Ubuntu
* Slackware

Any details should be referenced and not embedded in the documentation as we don't want to bloat the documentation too much or have extra maintenance work.

Suggested text:

> For RedHat/CentOS systems, it is recommended that Overlay (2) is used on top of an XFS filesystem.  See the [RedHat documentation](https://access.redhat.com/documentation/en-US/Red_Hat_Enterprise_Linux/7/html/7.2_Release_Notes/technology-preview-file_systems.html) for more information.  Also, see [this thread](https://github.com/docker/docker/issues/27358#issuecomment-254066211) for additional information to ensure the partition is properly formatted.

### LINKED ISSUE #2252: Problems running aufs w/mainline kernel
[engine/installation/linux/ubuntu.md](https://docs.docker.com/engine/installation/linux/ubuntu/)

I am running mainline kernel on ubuntu:

```bash
ubuntu@romano:~$ lsb_release -a
No LSB modules are available.
Distributor ID: Ubuntu
Description:    Ubuntu 16.04.2 LTS
Release:        16.04
Codename:       xenial

ubuntu@romano:~$ uname -r
4.10.0-041000-generic
```

The docs specify installing `linux-image-extra-*` via `uname -r` substitution:

```bash
$ sudo apt-get install \
    linux-image-extra-$(uname -r) \
    linux-image-extra-virtual
```

However this fails because there is currently no package built for the mainline kernel. 

According to stackoverflow (http://askubuntu.com/questions/729157) this kernel does not require `linux-image-extra-*` (unless I'm misreading that comment)

However there is no `aufs` module loaded:

```bash
ubuntu@romano:~$ lsmod | grep aufs
ubuntu@romano:~$ 
```

I'm not sure what the steps are to install `aufs` for mainline, but perhaps the docs should note that this command will fail if you are running mainline kernel. Or perhaps that's just a risk you take running mainline.

### LINKED ISSUE #2694: Problems running aufs w/ linux-aws kernel
### Problem description

Installing Docker on **Ubuntu 16.04 LTS** with the [`linux-aws`](https://insights.ubuntu.com/2017/04/05/ubuntu-on-aws-gets-serious-performance-boost-with-aws-tuned-kernel/) kernel package fails because there is no corresponding `linux-image-extra-$(uname -r)` package for this kernel.

```sh
$ sudo apt-get install linux-image-extra-$(uname -r) \
>                      linux-image-extra-virtual
Reading package lists... Done
Building dependency tree       
E: Unable to locate package linux-image-extra-4.4.0-1012-aws
E: Couldn't find any package by glob 'linux-image-extra-4.4.0-1012-aws'
E: Couldn't find any package by regex 'linux-image-extra-4.4.0-1012-aws'

```

```sh
$ lsb_release -a
No LSB modules are available.
Distributor ID: Ubuntu
Description:  Ubuntu 16.04.2 LTS
Release:  16.04
Codename: xenial
```

```sh
$ uname -r
4.4.0-1012-aws
```

This is similar to an issue described in https://github.com/docker/docker.github.io/issues/2252 (although presumably the mainline is not supported).

The [docs](https://docs.docker.com/engine/installation/linux/ubuntu/#recommended-extra-packages-for-trusty-1404) seem to suggest that `linux-image-extra-*` packages are only required when installing on `Trusty 14.04`.

However, the `aufs-driver` [docs](https://docs.docker.com/engine/userguide/storagedriver/aufs-driver/#configure-docker-with-aufs) include the same instructions, without reference to Ubuntu versions or code names.

This is confusing and probably needs clarification.
* What is the right way to install `aufs` support for Docker on Ubuntu 16.04?
* Are the `linux-image-extra-$(uname -r)` packages required to use `aufs` with Docker? 

### Problem location

URLS:
* https://docs.docker.com/engine/installation/linux/ubuntu/
* https://docs.docker.com/engine/userguide/storagedriver/aufs-driver/#configure-docker-with-aufs

I was installing the latest version of `docker ce` on `Ubuntu 16.04 LTS`, but using the brand new `linux-aws` kernel package. This may affect other versions of Docker on Ubuntu.

### Suggestions for a fix

Not sure if this is a suggestion (or even recommended), but here are my own findings on _successfully_ installing on the `linux-aws` kernel.

The `aufs-driver` [docs](https://docs.docker.com/engine/userguide/storagedriver/aufs-driver/#configure-docker-with-aufs) first suggest to check for `aufs` support:

```sh
$ grep aufs /proc/filesystems

nodev   aufs
```

Then to install the `linux-image-extra-*` packages:

```sh
sudo apt-get install linux-image-extra-$(uname -r) \
  linux-image-extra-virtual
```

Instead, I ran _only_ this command :

```sh
sudo apt-get install linux-image-extra-virtual
```

This also installed `linux-headers-4.4.0-72-generic`, which presumably is what would be installed if I were not running the `linux-aws kernel`.

NOTE: At this point running `grep aufs /proc/filesystems` produces no output.

```sh
$ grep aufs /proc/filesystems

```

From here, I installed `docker-ce` as normal.

After installing Docker, running `grep aufs /proc/filesystems` produces the desired output:

```sh
$ grep aufs /proc/filesystems

nodev aufs
```

And after configuring Docker to use the `aufs` storage driver as described [here](https://docs.docker.com/engine/userguide/storagedriver/aufs-driver/#configuration), everything appears to run as expected. 

```sh
$ sudo docker info

ubuntu@ip-10-150-100-85:~$ sudo docker info
Containers: 2
 Running: 0
 Paused: 0
 Stopped: 2
Images: 1
Server Version: 17.03.1-ce
Storage Driver: aufs
 Root Dir: /var/lib/docker/aufs
 Backing Filesystem: extfs
 Dirs: 5
 Dirperm1 Supported: true
 ...
```

### LINKED ISSUE #2733: Is the aufs graph driver production-ready or not?
* [This article tells that aufs if production-ready](https://docs.docker.com/engine/userguide/storagedriver/selectadriver/#future-proofing)
* [This article tells that aufs is very stable](https://docs.docker.com/engine/userguide/storagedriver/aufs-driver/)
* [And this tells me that aufs might crash my kernel](https://docs.docker.com/engine/reference/commandline/dockerd/#daemon-storage-driver-option)

So it's not really clear if you should be using this in production or not.

**Update**: it seems that aufs is production-ready and we should make this super clear by removing wording that makes you think otherwise.

### LINKED ISSUE #2769: Document compatibility matrix
There's a compatibility matrix at https://success.docker.com/Policies/Compatibility_Matrix but this
information should be front and center for people installing Docker EE and expecting it to be certified.

We should spread this information, to be sure users find it:

* [ ] Migrate compatibility matrix to docs
* [ ] Update install instructions for different distros to point to this page as a system requirement
* [ ] Also review instructions for CS Engine because some users are using that to install
* [ ] Update [select a storage drive](https://docs.docker.com/engine/userguide/storagedriver/selectadriver/) to be consistent/link with the info on this matrix
* [ ] Update https://docs.docker.com/engine/userguide/storagedriver/ to show on the left nav to help people navigate

The last article is also under "User Guide", but might not be the best place for it. If you're using Docker in a production environment you should be aware of this information before installing Docker, and it's more of an admin task to ensure that Docker is properly configured, not a user task.

mstanleyjones:
### Proposed changes

The Storage Driver docs could use some love. This will probably have more issues added to it before I'm done. I will be normalizing the structure of the docs about each storage drive, updating examples, significant rewriting, and other things. This is my project for the week.

Fixes #1062 
Fixes #2733 
Fixes https://github.com/docker/docker/pull/31104
Fixes #2252
Fixes #2694 
Fixes #2733 
Fixes #1062 
Fixes #1444
Fixes #1126
Fixes #2769 

cc/ @justincormack @dmcgowan @jpetazzo @dmp42 @friism  for review.

### DIFF
--- a/engine/installation/linux/centos.md
+++ b/engine/installation/linux/centos.md
@@ -31,14 +31,14 @@ URL when you see the placeholder text `<DOCKER-EE-URL>`.
 To learn more about Docker EE, see
 [Docker Enterprise Edition](https://www.docker.com/enterprise-edition/){: target="_blank" class="_" }.
 
-### OS requirements
-
-To install Docker, you need the 64-bit version of CentOS 7.
-
 In addition, you must use the `devicemapper` storage driver if you use Docker EE
 or CS-Engine. On production systems, you must use `direct-lvm` mode, which
 requires one or more dedicated block devices. Fast storage such as solid-state
 media (SSD) is recommended.
+
+### OS requirements
+
+To install Docker, you need the 64-bit version of CentOS 7.
 
 ### Uninstall old versions
 
--- a/engine/installation/linux/centos.md
+++ b/engine/installation/linux/centos.md
@@ -35,6 +35,11 @@ To learn more about Docker EE, see
 
 To install Docker, you need the 64-bit version of CentOS 7.
 
+In addition, you must use the `devicemapper` storage driver if you use Docker EE
+or CS-Engine. On production systems, you must use `direct-lvm` mode, which
+requires one or more dedicated block devices. Fast storage such as solid-state
+media (SSD) is recommended.
+
 ### Uninstall old versions
 
 Older versions of Docker were called `docker` or `docker-engine`. If these are
@@ -82,10 +87,12 @@ Repository set-up instructions are different for [Docker CE](#docker-ce) and
 
 {% assign download-url-base = "https://download.docker.com/linux/centos" %}
 
-1.  Install `yum-utils`, which provides the `yum-config-manager` utility:
+1.  Install required packages. `yum-utils` provides the `yum-config-manager`
+    utility, and `device-mapper-persistent-data` and `lvm2` are required by the
+    `devicemapper` storage driver.
 
     ```bash
-    $ sudo yum install -y yum-utils
+    $ sudo yum install -y yum-utils device-mapper-persistent-data lvm2
     ```
 
 2.  Use the following command to set up the **stable** repository. You always
@@ -128,10 +135,12 @@ Repository set-up instructions are different for [Docker CE](#docker-ce) and
     $ sudo sh -c 'echo "<DOCKER-EE-URL>" > /etc/yum/vars/dockerurl'
     ```
 
-3.  Install `yum-utils`, which provides the `yum-config-manager` utility:
+3.  Install required packages. `yum-utils` provides the `yum-config-manager`
+    utility, and `device-mapper-persistent-data` and `lvm2` are required by the
+    `devicemapper` storage driver.
 
     ```bash
-    $ sudo yum install -y yum-utils
+    $ sudo yum install -y yum-utils device-mapper-persistent-data lvm2
     ```
 
 4.  Use the following command to add the **stable** repository:
@@ -202,13 +211,27 @@ Repository set-up instructions are different for [Docker CE](#docker-ce) and
     | Docker CE      | `sudo yum install docker-ce-<VERSION>`        |
     | Docker EE      | `sudo yum install docker-ee-<VERSION>`        |
 
-4.  Start Docker.
+4.  Edit `/etc/docker/daemon.json`. If it does not yet exist, create it. Assuming
+    that the file was empty, add the following contents.
+
+    ```json
+    {
+      "storage-driver": "devicemapper"
+    }
+    ```
+
+5.  For production systems, you must use `direct-lvm` mode, which requires you
+    to prepare the block devices. Follow the procedure in the
+    [devicemapper storage driver guide](/engine/userguide/storagedriver/device-mapper-driver.md#configure-direct-lvm-mode-for-production){: target="_blank" class="_" }
+    **before starting Docker**.
+
+6.  Start Docker.
 
     ```bash
     $ sudo systemctl start docker
     ```
 
-5.  Verify that `docker` is installed correctly by running the `hello-world`
+7.  Verify that `docker` is installed correctly by running the `hello-world`
     image.
 
     ```bash
@@ -256,13 +279,27 @@ a new file each time you want to upgrade Docker.
     $ sudo yum install /path/to/package.rpm
     ```
 
-3.  Start Docker.
+3.  Edit `/etc/docker/daemon.json`. If it does not yet exist, create it. Assuming
+    that the file was empty, add the following contents.
+
+    ```json
+    {
+      "storage-driver": "devicemapper"
+    }
+    ```
+
+4.  For production systems, you must use `direct-lvm` mode, which requires you
+    to prepare the block devices. Follow the procedure in the
+    [devicemapper storage driver guide](/engine/userguide/storagedriver/device-mapper-driver.md#configure-direct-lvm-mode-for-production){: target="_blank" class="_" }
+    **before starting Docker**.
+
+5.  Start Docker.
 
     ```bash
     $ sudo systemctl start docker
     ```
 
-4.  Verify that `docker` is installed correctly by running the `hello-world`
+6.  Verify that `docker` is installed correctly by running the `hello-world`
     image.
 
     ```bash
@@ -299,6 +336,9 @@ instead of `yum -y install`, and pointing to the new file.
     ```bash
     $ sudo rm -rf /var/lib/docker
     ```
+
+3.  If desired, remove the `devicemapper` thin pool and reformat the block
+    devices that were part of it.
 
 You must delete any edited configuration files manually.
 
--- a/engine/installation/linux/suse.md
+++ b/engine/installation/linux/suse.md
@@ -34,9 +34,12 @@ Docker Community Edition (Docker CE) is not supported on SLES.
 
 ### OS requirements
 
-To install Docker, you need the 64-bit version of SLES 12.x.
+To install Docker, you need the 64-bit version of SLES 12.x. Docker is not
+supported on OpenSUSE.
 
-Docker is not supported on OpenSUSE.
+The only supported storage driver for Docker on SLES is `btrfs`, which will be
+used by default if the underlying filesystem hosting `/var/lib/docker/` is a
+BTRFS filesystem.
 
 ### Uninstall old versions
 
@@ -58,6 +61,34 @@ It's OK if `zypper` reports that none of these packages are installed.
 
 The contents of `/var/lib/docker/`, including images, containers, volumes, and
 networks, are preserved. The Docker EE package is now called `docker-ee`.
+
+## Configure the btrfs filesystem
+
+If the filesystem which hosts `/var/lib/docker/` is not a BTRFS filesystem,
+you must configure a BTRFS filesystem and mount it on `/var/lib/docker/`:
+
+1.  Format your dedicated block device or devices as a Btrfs filesystem. This
+    example assumes that you are using two block devices called `/dev/xvdf` and
+    `/dev/xvdg`. Double-check the block device names because this is a
+    destructive operation.
+
+    ```bash
+    $ sudo mkfs.btrfs -f /dev/xvdf /dev/xvdg
+    ```
+
+    There are many more options for Btrfs, including striping and RAID. See the
+    [Btrfs documentation](https://btrfs.wiki.kernel.org/index.php/Using_Btrfs_with_Multiple_Devices).
+
+2.  Mount the new Btrfs filesystem on the `/var/lib/docker/` mount point. You
+    can specify any of the block devices used to create the Btrfs filesystem.
+
+    ```bash
+    $ sudo mount -t btrfs /dev/xvdf /var/lib/docker
+    ```
+
+    Don't forget to make the change permanent across reboots by adding an
+    entry to `/etc/fstab`.
+
 
 ## Install Docker
 
@@ -149,13 +180,25 @@ the repository.
     $ sudo zypper install docker-ee-<VERSION_STRING>
     ```
 
-    Start Docker:
+4.  Configure Docker to use the `btrfs` filesystem. Edit the file
+    `/etc/docker/daemon.json` (create it if it does not exist) and add the
+    following contents:
+
+    ```json
+    {
+      "storage-driver": "btrfs"
+    }
+    ```
+
+    Save and close the file.
+
+5.  Start Docker:
 
     ```bash
     $ sudo service docker start
     ```
 
-4.  Verify that Docker EE is installed correctly by running the `hello-world`
+6.  Verify that Docker EE is installed correctly by running the `hello-world`
     image.
 
     ```bash
@@ -200,13 +243,25 @@ need to download a new file each time you want to upgrade Docker.
     $ sudo zypper install /path/to/package.rpm
     ```
 
-    Start Docker:
+4.  Configure Docker to use the `btrfs` filesystem. Edit the file
+    `/etc/docker/daemon.json` (create it if it does not exist) and add the
+    following contents:
+
+    ```json
+    {
+      "storage-driver": "btrfs"
+    }
+    ```
+
+    Save and close the file.
+
+5.  Start Docker:
 
     ```bash
     $ sudo service docker start
     ```
 
-4.  Verify that Docker EE is installed correctly by running the `hello-world`
+6.  Verify that Docker EE is installed correctly by running the `hello-world`
     image.
 
     ```bash
@@ -240,8 +295,10 @@ instead of `zypper install`, and pointing to the new file.
     volumes:
 
     ```bash
-    $ sudo rm -rf /var/lib/docker
+    $ sudo rm -rf /var/lib/docker/*
     ```
+
+    As an alternative, you can unmount and format the `btrfs` filesystem.
 
 You must delete any edited configuration files manually.
 
--- a/engine/userguide/storagedriver/aufs-driver.md
+++ b/engine/userguide/storagedriver/aufs-driver.md
@@ -4,248 +4,231 @@ keywords: 'container, storage, driver, AUFS '
 title: Docker and AUFS in practice
 ---
 
-AUFS was the first storage driver in use with Docker. As a result, it has a
-long and close history with Docker, is very stable, has a lot of real-world
-deployments, and has strong community support. AUFS has several features that
-make it a good choice for Docker. These features enable:
+AUFS a *union filesystem*. The `aufs` storage driver is the default storage
+driver used for managing images and layers on Docker for Ubuntu, and for
+Debian versions prior to Stretch. For Debian Stretch,
+[overlay2](overlayfs-driver.md) is the default.
 
-- Fast container startup times.
-- Efficient use of storage.
-- Efficient use of memory.
+AUFS is the most mature storage driver used by Docker. It provides fast
+container startup time, as well as efficient usage of memory and storage.
 
-Despite its capabilities and long history with Docker, some Linux distributions
- do not support AUFS. This is usually because AUFS is not included in the
-mainline (upstream) Linux kernel.
+If your Linux kernel is version 4.0 or higher, and you use Docker CE, consider
+using the newer [overlay2](overlayfs-driver.md){: target="_blank" class="_" },
+which has potential performance advantages over the `aufs` storage driver.
 
-The following sections examine some AUFS features and how they relate to
-Docker.
+> **Note**: Even though AUFS is used by default if it is present in the Linux
+> kernel, it is not supported on some distributions and Docker editions. See
+> [Prerequisites](#prerequisites) > for more information.
 
-## Image layering and sharing with AUFS
+## Prerequisites
 
-AUFS is a *unification filesystem*. This means that it takes multiple
-directories on a single Linux host, stacks them on top of each other, and
-provides a single unified view. To achieve this, AUFS uses a *union mount*.
+- For Docker CE, AUFS is supported on Ubuntu, and on Debian versions prior to
+  Stretch.
+- For Docker EE, AUFS is supported on Ubuntu.
+- If you use Ubuntu, you need to
+  [install extra packages](/engine/installation/linux/ubuntu.md#recommended-extra-packages-for-trusty-1404){: target="_blank" class="_"}
+  to add the AUFS module to the kernel. If you do not install these packages,
+  you will need to use `devicemapper` on Ubuntu 14.04 (which is not recommended),
+  or `overlay2` on Ubuntu 16.04 and higher, which is also supported.
+- AUFS cannot use the following backing filesystems: `aufs`, `btrfs`, or
+  `ecryptfs`. This means that the filesystem which contains
+  `/var/lib/docker/aufs` cannot be one of these filesystem types.
 
-AUFS stacks multiple directories and exposes them as a unified view through a
-single mount point. All of the directories in the stack, as well as the union
-mount point, must all exist on the same Linux host. AUFS refers to each
-directory that it stacks as a *branch*.
 
-Within Docker, AUFS union mounts enable image layering. The AUFS storage driver
- implements Docker image layers using this union mount system. AUFS branches
-correspond to Docker image layers. The diagram below shows a Docker container
-based on the `ubuntu:latest` image.
+## Configure Docker with the `aufs` storage driver
 
-![](images/aufs_layers.jpg)
+If the AUFS driver is loaded into the kernel when you start Docker, and no other
+storage driver is configured, Docker uses it by default.
 
-This diagram shows that each image layer, and the container layer, is
-represented in the Docker hosts filesystem as a directory under
-`/var/lib/docker/`. The union mount point provides the unified view of all
-layers. As of Docker 1.10, image layer IDs do not correspond to the names of
-the directories that contain their data.
+1.  Use the following command to verify that your kernel supports AUFS.
 
-AUFS also supports the copy-on-write technology (CoW). Not all storage drivers
-do.
+    ```bash
+    $ grep aufs /proc/filesystems
 
-## Container reads and writes with AUFS
+    nodev   aufs
+    ```
 
-Docker leverages AUFS CoW technology to enable image sharing and minimize the
-use of disk space. AUFS works at the file level. This means that all AUFS CoW
-operations copy entire files - even if only a small part of the file is being
-modified. This behavior can have a noticeable impact on container performance,
- especially if the files being copied are large, below a lot of image layers,
-or the CoW operation must search a deep directory tree.
+2.  Check which storage driver Docker is using.
 
-Consider, for example, an application running in a container needs to add a
-single new value to a large key-value store (file). If this is the first time
-the file is modified, it does not yet exist in the container's top writable
-layer. So, the CoW must *copy up* the file from the underlying image. The AUFS
-storage driver searches each image layer for the file. The search order is from
- top to bottom. When it is found, the entire file is *copied up* to the
-container's top writable layer. From there, it can be opened and modified.
+    ```bash
+    $ docker info
 
-Larger files obviously take longer to *copy up* than smaller files, and files
-that exist in lower image layers take longer to locate than those in higher
-layers. However, a *copy up* operation only occurs once per file on any given
-container. Subsequent reads and writes happen against the file's copy already
-*copied-up* to the container's top layer.
+    <truncated output>
+    Storage Driver: aufs
+     Root Dir: /var/lib/docker/aufs
+     Backing Filesystem: extfs
+     Dirs: 0
+     Dirperm1 Supported: true
+    <truncated output>
+    ```
 
-## Deleting files with the AUFS storage driver
+3.  If you are using a different storage driver, then it means that either AUFS
+    is not included in the kernel (in which case a different default driver will
+    be used) or that means that Docker has been explicitly configured to use a
+    different driver. Check `/etc/docker/daemon.json` or the output of
+    `ps auxw |grep dockerd` to see if Docker has been started with the
+    `--storage-driver` flag.
 
-The AUFS storage driver deletes a file from a container by placing a *whiteout
-file* in the container's top layer. The whiteout file effectively obscures the
-existence of the file in the read-only image layers below. The simplified
-diagram below shows a container based on an image with three image layers.
 
-![](images/aufs_delete.jpg)
+## How the `aufs` storage driver works
 
-The `file3` was deleted from the container. So, the AUFS storage driver  placed
-a whiteout file in the container's top layer. This whiteout file effectively
-"deletes" `file3` from the container by obscuring any of the original file's
-existence in the image's read-only layers. This works the same no matter which
-of the image's read-only layers the file exists in.
+AUFS is a *union filesystem*, which means that it layers multiple directories on
+a single Linux host and presents them as a single directory. These directories
+are called _branches_ in AUFS terminology, and _layers_ in Docker terminology.
+The unification process is referred to a a _union mount_.
 
-## Renaming directories with the AUFS storage driver
+The diagram below shows a Docker container based on the `ubuntu:latest` image.
 
-Calling `rename(2)` for a directory is not fully supported on AUFS. It returns
-`EXDEV` ("cross-device link not permitted"), even when both of the source and
-the destination path are on a same AUFS layer, unless the directory has no
-children.
+![Layers of an Ubuntu container](images/aufs_layers.jpg)
 
-So your application has to be designed so that it can handle `EXDEV` and fall
-back to a "copy and unlink" strategy.
+Each image layer, and the container layer, are represented on the Docker host as
+subdirectories within `/var/lib/docker/`. The union mount provides the unified
+view of all layers. The directory names do not directly correspond to the IDs
+of the layers themselves.
 
-## Configure Docker with AUFS
+AUFS uses the Copy-on-Write (CoW) strategy to maximize storage efficiency and
+minimize overhead.
 
-### Prerequisites
+### Example: Image and container on-disk constructs
 
-You can only use the AUFS storage driver on Linux systems with AUFS installed.
-Use the following command to determine if your system supports AUFS.
+The following `docker pull` command shows a Docker host downloading a Docker
+image comprising five layers.
 
 ```bash
-$ grep aufs /proc/filesystems
+$ docker pull ubuntu
 
-nodev   aufs
+Using default tag: latest
+latest: Pulling from library/ubuntu
+b6f892c0043b: Pull complete
+55010f332b04: Pull complete
+2955fb827c94: Pull complete
+3deef3fcbd30: Pull complete
+cf9722e506aa: Pull complete
+Digest: sha256:382452f82a8bbd34443b2c727650af46aced0f94a44463c62a9848133ecb1aa8
+Status: Downloaded newer image for ubuntu:latest
 ```
 
-This output indicates the system supports AUFS. If you get no output, your system does
-not support AUFS. To address this:
+#### The image layers
 
-- Upgrade your host system's kernel to 3.13 or higher. It is recommended to install the
-  kernel headers when you upgrade.
+All of the information about the image and container layers is stored in
+subdirectories of `/var/lib/docker/aufs/`.
 
-- **Ubuntu or Debian**: In addition to updating the kernel if necessary, install the
-  `linux-image-extra-*` packages:
-  
-  ```bash
-  $ sudo apt-get install linux-image-extra-$(uname -r) \
-                         linux-image-extra-virtual
-  ```
+- `diff/`: the **contents** of each layer, each stored in a separate
+  subdirectory
+- `layers/`: metadata about how image layers are stacked. This directory
+  contains one file for each image or container layer on the Docker host. Each
+  file contains the IDs of all the layers below it in the stack (its parents).
+- `mnt/`: Mount points, one per image or container layer, which are used to
+  assemble and mount the unified filesystem for a container. For images, which
+  are read-only, these directories will always be empty.
 
-### Configuration
+#### The container layer
 
-When you have verified that you meet the prerequisites, instruct the Docker daemon to use
-AUFS by starting the Docker daemon with the flag `--storage-driver=aufs`:
+If a container is running, the contents of `/var/lib/docker/aufs/` change in the
+following ways:
 
-```bash
-$ sudo dockerd --storage-driver=aufs &
-```
+> **Note**: If you see files or directories which end in `-init`, these are used
+> to prepare the container to use bind mounts, and should be considered an
+> implementation detail.
 
-To make the change permanent, you can edit the Docker configuration file and add the
-`--storage-driver=aufs` option to the `DOCKER_OPTS` line.
+- `diff/`: Differences introduced in the writable container layer, such as new
+   or modified files.
+- `layers/`: Metadata about the writable container layer's parent layers.
+- `mnt/`: A mount point for each running container's unified filesystem, exactly
+  as it appears from within the container.
 
-```none
-# Use DOCKER_OPTS to modify the daemon startup options.
-DOCKER_OPTS="--storage-driver=aufs"
-```
+## How container reads and writes work with `aufs`
 
-After the daemon starts, verify the default storage driver using the `docker info`
-command:
+### Reading files
 
-```bash
-$ sudo docker info
+Consider three scenarios where a container opens a file for read access with
+overlay.
 
-Containers: 1
-Images: 4
-Storage Driver: aufs
- Root Dir: /var/lib/docker/aufs
- Backing Filesystem: extfs
- Dirs: 6
- Dirperm1 Supported: false
-Execution Driver: native-0.2
-...output truncated...
-```
+- **The file does not exist in the container layer**: If a container opens a
+  file for read access and the file does not already exist in the container
+  layer, the storage driver searches for the file in the image layers,
+  starting with the layer just below the container layer. It is read from the
+  layer where it is found.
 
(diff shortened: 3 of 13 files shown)


mstanleyjones:
cc/ @trapier for review, as well.

All of this is now ready for review except for `zfs`, which I have not gotten to yet.

trapier:
Dunno if it belongs in this PR or not, but when we're creating the direct-lvm thinpool, we create:

- docker/thinpool
- docker/thinpoolmeta

Later on in the graphics and when we're adding a device, the names change to:

- vg-docker/data
- vg-docker/metadata

This lack of continuity decreases re-usability and teaching value of the "adding a device" section.   

Analyzing all support cases for which I've received customer data, the convention (90%) in practice seems to be `docker/thinpool`.  Bringing the doc into line with practice and itself would require rebuilding some graphics.  :-/

mstanleyjones:
@trapier hmm, I don't see `vg-docker` in the devicemapper topic at all anymore. Maybe I updated that after you looked the first time.

engine/userguide/storagedriver/aufs-driver.md
```diff
@@ -4,248 +4,222 @@ keywords: 'container, storage, driver, AUFS '
 title: Docker and AUFS in practice
 ---
 
-AUFS was the first storage driver in use with Docker. As a result, it has a
-long and close history with Docker, is very stable, has a lot of real-world
-deployments, and has strong community support. AUFS has several features that
-make it a good choice for Docker. These features enable:
+AUFS a *union filesystem*. The `aufs` storage driver is the default storage
+driver used for managing images and layers on Docker, for Ubuntu and Debian.
```
thaJeztah: - for Debian, I don't think it's supported anymore in their newer kernels @justincormack ?
- for Ubuntu 16.04, and 14.04 with the newer 4.x kernels installed (which I _think_ is now the default on some installations with 14.04.2), overlay2 is the default

engine/userguide/storagedriver/aufs-driver.md
```diff
@@ -4,248 +4,222 @@ keywords: 'container, storage, driver, AUFS '
...
+> for more information.
 
-The following sections examine some AUFS features and how they relate to
-Docker.
+## Prerequisites
 
-## Image layering and sharing with AUFS
+- For Docker CE, AUFS is supported on Ubuntu and Debian.
+- For Docker EE, AUFS is supported on Ubuntu.
+- If you use Ubuntu 14.04, you may need to
```
thaJeztah: You need to install the extra packages on all versions of Ubuntu if you want to use aufs. For Ubuntu 14.04 (with kernel 3.13) we recommend installing it, because the alternative would be device mapper (which is not ideal). For ubuntu 16.04 we documented the package as "optional", because without it, it defaults to overlay2 which is a better alternative for most.

engine/userguide/storagedriver/aufs-driver.md
```diff
@@ -4,248 +4,222 @@ keywords: 'container, storage, driver, AUFS '
...
+> **Note**: Even though AUFS is the default storage driver, it is not supported
+> on some distributions and Docker editions. See [Prerequisites](#prerequisites)
+> for more information.
 
-The following sections examine some AUFS features and how they relate to
-Docker.
+## Prerequisites
 
-## Image layering and sharing with AUFS
+- For Docker CE, AUFS is supported on Ubuntu and Debian.
```
thaJeztah: For debian support, see my earlier comment (they dropped support for aufs, and no longer offer the required packages for their newer kernels)

engine/userguide/storagedriver/aufs-driver.md
```diff
@@ -4,248 +4,222 @@ keywords: 'container, storage, driver, AUFS '
...
+     Dirperm1 Supported: true
+    <truncated output>
+    ```
 
-Larger files obviously take longer to *copy up* than smaller files, and files
-that exist in lower image layers take longer to locate than those in higher
-layers. However, a *copy up* operation only occurs once per file on any given
-container. Subsequent reads and writes happen against the file's copy already
-*copied-up* to the container's top layer.
+3.  If you are using a different storage driver, that means that Docker has been
```
thaJeztah: If you run on a vanilla kernel (without the kernel-extra package installed), the defaults will be;

- kernel < 3.16: device mapper
- kernel >= 3.16: overlay
- kernel >= 4.x overlay2

engine/userguide/storagedriver/aufs-driver.md
```diff
@@ -4,248 +4,222 @@ keywords: 'container, storage, driver, AUFS '
...
+view of all layers. The directory names do not directly correspond to the IDs
+of the layers themselves.
 
-So your application has to be designed so that it can handle `EXDEV` and fall
-back to a "copy and unlink" strategy.
+AUFS uses the Copy-on-Write (CoW) strategy to maximize storage efficiency and
+minimize overhead.
 
-## Configure Docker with AUFS
+### Example: Image and container on-disk constructs (`overlay`)
```
thaJeztah: use of `overlay` as a term may be confusing here in this context (as this is the `aufs` page)

engine/userguide/storagedriver/aufs-driver.md
justincormack: Yes it was in Debian Jessie but its not in Stretch which is about to be released. Seems to be staying in recent Ubuntu though.

engine/userguide/storagedriver/aufs-driver.md
```diff
@@ -4,248 +4,222 @@ keywords: 'container, storage, driver, AUFS '
...
+If your Linux kernel is version 4.0 or higher, and you use Docker CE, consider
+using the newer [overlay2](overlayfs-driver.md){: target="_blank" class="_" },
+which has potential performance advantages over the `aufs` storage driver.
 
-Despite its capabilities and long history with Docker, some Linux distributions
- do not support AUFS. This is usually because AUFS is not included in the
-mainline (upstream) Linux kernel.
+> **Note**: Even though AUFS is the default storage driver, it is not supported
+> on some distributions and Docker editions. See [Prerequisites](#prerequisites)
+> for more information.
```
justincormack: I think its weird calling it "default", although it is the highest ranked.

engine/userguide/storagedriver/aufs-driver.md
```diff
@@ -4,248 +4,222 @@ keywords: 'container, storage, driver, AUFS '
...
 To summarize some of the performance related aspects already mentioned:
 
-- The AUFS storage driver is a good choice for PaaS and other similar use-cases
- where container density is important. This is because AUFS efficiently shares
-images between multiple running containers, enabling fast container start times
- and minimal use of disk space.
+- The AUFS storage driver is less performant than the `overlay2` driver, but is
+  a good choice for PaaS and other similar use-cases where container density is
+  important. This is because AUFS efficiently shares images between multiple
+  running containers, enabling fast container start times and minimal use of
```
justincormack: Note the alternative here that shares image contents is `btrfs`.

justincormack: (I think thats correct)

engine/userguide/storagedriver/btrfs-driver.md
```diff
@@ -227,179 +106,193 @@ loaded. To verify this, use the following command:
...
-*n* containers accessing the same file require *n* copies to be cached. As a
-result, the `btrfs` driver may not be the best choice for PaaS and other high
-density container use cases.
-
-- **Small writes**. Containers performing lots of small writes (including
-Docker hosts that start and stop many containers) can lead to poor use of Btrfs
- chunks. This can ultimately lead to out-of-space conditions on your Docker
-host and stop working. This is currently a major drawback to use current
-versions of Btrfs.
+  each accessing the same file copies the file into the Docker hosts's memory.
```
justincormack: "each process accessing"

engine/userguide/storagedriver/btrfs-driver.md
```diff
@@ -227,179 +106,193 @@ loaded. To verify this, use the following command:
...
-    If you use the `btrfs` storage driver, closely monitor the free space on
-your Btrfs filesystem using the `btrfs filesys show` command. Do not trust the
-output of normal Unix commands such as `df`; always use the Btrfs native
-commands.
+- **Small writes**. Containers performing lots of small writes (this usage
+  pattern matches what happens when you start and stop many containers in a short
+  period of time, as well) can lead to poor use of Btrfs chunks. This can
+  prematurely fill the Btrfs filesystem and lead to out-of-space conditions on
+  your Docker host. Use `btrfs filesys show` to closely monitor the amount of
+  free space on your Btrfs device.
```
thaJeztah: Perhaps we should refer here to the use of volumes already?

Tangential, but most _write_ issues on our storage driver are not an issue if the containers are "well designed", and are considered immutable (only some writes); any high-traffic data should not live in the container's filesystem

engine/userguide/storagedriver/aufs-driver.md
mstanleyjones: That's a mistake on my part.

engine/userguide/storagedriver/btrfs-driver.md
```diff
@@ -227,179 +106,193 @@ loaded. To verify this, use the following command:
...
+  This reclaims unallocated blocks and helps to prevent the filesystem from
+  filling up unnecessarily. You cannot rebalance a totally full Btrfs
+  filesystem unless you add additional physical block devices to the filesystem.
+  See the
+  [BTRFS Wiki](https://btrfs.wiki.kernel.org/index.php/Balance_Filters#Balancing_to_fix_filesystem_full_errors).
+
+- **Use fast storage**: Solid-state drives (SSDs) provide faster reads and
+  writes than spinning disks.
+
+- **Use volumes for write-heavy workloads**: Volumes provide the best and most
```
thaJeztah: Thinking of which; given that for BTRFS, the whole `/var/lib/docker` directory is BTRFS, volumes (which are stored in `/var/lib/docker/volumes` will still suffer from the performance penalties of BTRFS.

I know something has come up a few times, alternatives are;

- if following these instructions, use `/var/lib/docker/btrfs` as the location to mount the btrfs device (does that work?)
- or we should have an option to specify the location where volumes are stored

engine/userguide/storagedriver/aufs-driver.md
thaJeztah: For Ubuntu 16.04 we basically recommend overlay2 (at least for CE, not sure what EE's plans are)

engine/userguide/storagedriver/btrfs-driver.md
```diff
@@ -6,216 +6,95 @@ title: Docker and Btrfs in practice
...
-As far as Docker on the Linux platform goes, many people see the `btrfs`
-storage driver as a potential long-term replacement for the `devicemapper`
-storage driver. However, at the time of writing, the `devicemapper` storage
-driver should be considered safer, more stable, and more *production ready*.
-You should only consider the `btrfs` driver for production deployments if you
-understand it well and have existing experience with Btrfs.
+- **Docker CE**. For Docker CE, `btrfs` is only supported on Ubuntu.
 
-## Image layering and sharing with Btrfs
+- **Docker EE**: For Docker EE and CS-Engine, `btrfs` is only supported on SLES.
```
mstanleyjones: @dmp42 @friism WDYT?

engine/userguide/storagedriver/btrfs-driver.md
```diff
@@ -6,216 +6,95 @@ title: Docker and Btrfs in practice
...
-5.  Mount the Btrfs volume:
-
-    ```bash
-    $ sudo mount /var/lib/docker
-    ```
+    Don't forget to make the change permanent across reboots by adding an
+    entry to `/etc/fstab`.
 
-    Copy the old Docker contents to `/var/lib/docker`.
+5.  Copy the contents of `/var/lib/docker.bk` to `/var/lib/docker/`.
```
mstanleyjones: I guess it's just so you can go back if something goes horribly wrong.

engine/userguide/storagedriver/aufs-driver.md
thaJeztah: Maybe we should explain the "automatic" selection of storage drivers, which is basically;

- try each driver in order and check if the prerequisites are there / if the driver is supported
- if not, continue with the next driver

Order of drivers is decided based on these;

- minimal configuration required (i.e. if I have btrfs as filesystem, assume the user wants to use that, and accept the choice)
- amount of configuration needed for "best use" (i.e. overlay/overlay2 is above device mapper, because device mapper on loopback is worse than overlay/overlay2. Given that device mapper on direct-lvm always requires manual interaction, it's not selected as default in current docker versions)
- performance and stability; as an extreme: both "overlay" and "vfs" don't require configuration, but performance of "vfs" is not usable, so "overlay" is prefered. Similarly; "overlay2" and "overlay" both require no configuration, but "overlay2" is better performing, so _if_ supported, prefer that over "overlay"

engine/installation/linux/centos.md
```diff
@@ -35,6 +35,11 @@ To learn more about Docker EE, see
 
 To install Docker, you need the 64-bit version of CentOS 7.
 
+In addition, you must use the `devicemapper` storage driver if you use Docker EE
```
thaJeztah: Should we put this under the `Docker EE customers` heading?

engine/installation/linux/centos.md
```diff
@@ -256,13 +279,27 @@ a new file each time you want to upgrade Docker.
...
+3.  Edit `/etc/docker/daemon.json`. If it does not yet exist, create it. Assuming
+    that the file was empty, add the following contents.
+
+    ```json
+    {
+      "storage-driver": "devicemapper"
+    }
+    ```
+
+4.  For production systems, you must use `direct-lvm` mode, which requires you
```
thaJeztah: I'm worried that we re-introduce device mapper in the standard installation steps here. We moved device mapper below overlay and overlay2 exactly to prevent this situation; i.e. running with overlay/overlay2 is highly recommended over using device mapper on loopbacks.

People _will_ skip the "direct-lvm" steps, and run on loopbacks; get really bad performance, and as a result, tons of issues reported.

If device mapper is the only configuration we support for EE, and we need instructions for setting up Docker EE in a configuration that qualifies to get technical support, I think we should have a separate installation guide for Docker EE that explains all the steps to do this, instead of trying to combine EE and CE installation.

engine/userguide/storagedriver/index.md
```diff
@@ -1,29 +1,38 @@
 ---
 description: Learn how select the proper storage driver for your container.
-keywords: container, storage, driver, AUFS, btfs, devicemapper,zvfs
+keywords: container, storage, driver, AUFS, btfs, devicemapper, zvfs, overlay, overlay2
```
thaJeztah: s/btfs/btrfs/
s/zvfs/zfs/

engine/userguide/storagedriver/overlayfs-driver.md
```diff
@@ -4,412 +4,470 @@ keywords: container, storage, driver, OverlayFS
...
+## Prerequisites
 
-Since version 1.12, Docker also provides `overlay2` storage driver which is much
-more efficient than `overlay` in terms of inode utilization. The `overlay2`
-driver is only compatible with Linux kernel 4.0 and later.
+OverlayFS is supported if you meet the following prerequisites:
 
-For comparison between `overlay` vs `overlay2`, please also refer to [Select a
-storage driver](selectadriver.md#overlay-vs-overlay2).
+- **Docker CE only**. OverlayFS is not supported on Docker EE. See
```
thaJeztah: Thinking here as well about distinguishing "supported", and "qualifies for commercial support"

engine/userguide/storagedriver/overlayfs-driver.md
```diff
@@ -4,412 +4,470 @@ keywords: container, storage, driver, OverlayFS
...
+  [Product compatibility matrix](https://success.docker.com/Policies/Compatibility_Matrix){: target="_blank" class="_"}
+  for a list of supported storage drivers for each Docker EE platform.
 
->**Note**: Since it was merged into the mainline kernel, the OverlayFS *kernel
->module* was renamed from "overlayfs" to "overlay". As a result you may see the
-> two terms used interchangeably in some documentation. However, this document
-> uses  "OverlayFS" to refer to the overall filesystem, and `overlay`/`overlay2`
-> to refer to Docker's storage-drivers.
+- Version 4.0 or higher of the Linux kernel. If you use an older kernel, you
+  will need to use the `overlay1` driver, which is not recommended.
```
thaJeztah: s/overlay1/overlay/

engine/userguide/storagedriver/overlayfs-driver.md
```diff
@@ -4,412 +4,470 @@ keywords: container, storage, driver, OverlayFS
...
 
-## Image layering and sharing with OverlayFS (`overlay`)
+- The following backing filesystems are supported:
+  - `ext4`
```
thaJeztah: on RHEL/CentOS, only `xfs` is supported (well, on the current version); I seem to recall RHEL 7.1 only supports `ext4` and RHEL 7.2 only supports `xfs` (ugh)

thaJeztah: (I know there's "no such thing" as RHEL 7.1, only RHEL7, but I think the EE team maintains support for users that decide to stick to older versions)

engine/userguide/storagedriver/overlayfs-driver.md
```diff
@@ -4,412 +4,470 @@ keywords: container, storage, driver, OverlayFS
...
 
 - **Inode limits**. Use of the `overlay` storage driver can cause excessive
-inode consumption. This is especially so as the number of images and containers
- on the Docker host grows. A Docker host with a large number of images and lots
- of started and stopped containers can quickly run out of inodes. The `overlay2`
- does not have such an issue.
+  inode consumption. This is especially true in the presence of a large number
+  of images and containers on the Docker host. The only way to increase the
+  number of inodes available to a filesystem is to reformat it. To avoid running
+  into this issue, it is highly recommended that you use `overlay2` if at all
```
thaJeztah: Not sure if we want to mention this somewhere; RHEL and CentOS support overlay2 on their kernel. However, there is no way for docker to automatically detect this. To user overlay2 on RHEL/CentOS, set the `overlay2.override_kernel_check` storage driver option on the daemon https://docs.docker.com/engine/reference/commandline/dockerd/#storage-driver-options

engine/installation/linux/centos.md
mstanleyjones: Ugh. WDYT @dmp42 @friism

engine/userguide/storagedriver/overlayfs-driver.md
mstanleyjones: I don't know if I should change this. We specifically do not support this configuration on EE and that is what I am saying.

engine/userguide/storagedriver/overlayfs-driver.md
mstanleyjones: For EE, we are specifically only supporting the configs in https://success.docker.com/Policies/Compatibility_Matrix.

engine/userguide/storagedriver/btrfs-driver.md
```diff
@@ -6,216 +6,95 @@ title: Docker and Btrfs in practice
...
+  created inaccessible on the local system. Use `docker save` to save containers,
+  and push existing images to Docker Hub or a private repository, so that you
+  not need to re-create them later.
 
-Subvolumes are natively copy-on-write and have space allocated to them
-on-demand from an underlying storage pool. They can also be nested and snapped.
- The diagram below shows 4 subvolumes. 'Subvolume 2' and 'Subvolume 3' are
-nested, whereas 'Subvolume 4' shows its own internal directory tree.
+- `btrfs` requires a dedicated block storage device such as a physical disk. This
+  block device must be formatted for Btrfs and mounted into `/var/lib/docker/`.
```
mstanleyjones: Addressed this in the latest round.

engine/userguide/storagedriver/btrfs-driver.md
mstanleyjones: Done.

engine/userguide/storagedriver/btrfs-driver.md
mstanleyjones: Well I added a note that you could mount `/var/lib/docker/volumes/` on a different type of FS to mitigate this, for now.

engine/userguide/storagedriver/overlayfs-driver.md
mstanleyjones: I addressed this where we show the edits to the `daemon.json`.

engine/userguide/storagedriver/aufs-driver.md
mstanleyjones: Done in the latest commit.

engine/userguide/storagedriver/device-mapper-driver.md
```diff
@@ -504,196 +466,303 @@ The `Data Space` values show that the pool is 100GB total. This example extends
...
+## Device Mapper and Docker performance
 
-One final point, data volumes provide the best and most predictable
-performance. This is because they bypass the storage driver and do not incur
-any of the potential overheads introduced by thin provisioning and
-copy-on-write. For this reason, you should place heavy write workloads on
-data volumes.
+- **`allocate-on demand` performance impact**:
+
+  The `devicemapper` storage driver uses an `allocte-on-demand` operation to
```
trapier: spelling: `allocate-on-demand`

engine/userguide/storagedriver/device-mapper-driver.md
```diff
@@ -504,196 +466,303 @@ The `Data Space` values show that the pool is 100GB total. This example extends
...
+Use the `mount` command to see the mount-point Docker is using:
 
-    Your volume group may use a different name.
+```bash
+$ mount |grep devicemapper
+/dev/xvda1 on /var/lib/docker/devicemapper type xfs (rw,relatime,seclabel,attr2,inode64,noquota)
+```
 
-2.  Extend the `data` logical volume (LV) `vg-docker/data`
+When you use `devicemapper`, Docker stores image and layer contents in
```
trapier: Images and layer contents are stored in the thinpool, and exposed by mounting them under subdirectories of `/var/lib/docker/devicemapper/`.

engine/userguide/storagedriver/device-mapper-driver.md
```diff
@@ -504,196 +466,303 @@ The `Data Space` values show that the pool is 100GB total. This example extends
...
-    c.  Calculate the real total sectors of the thin pool now.
+3.  Extend the `thinpool/data` logical volume. This command uses 100% of the
+    volume right away, without auto-extend. If you needed to extend the
+    `thinpool/metadata` logical volume instead, substitute it.
 
-        Change the second number of the table info (i.e. the disk end sector) to
-        reflect the new number of 512 byte sectors in the disk. For example, as the
-        new loop size is 200GB, change the second number to 419430400.
+    ```bash
+    $ sudo lvextend  -l+100%FREE -n thinpool/data
```
trapier: This command doesn't work with the `docker/thinpool` naming convention:

```
$ sudo lvresize -l+100%FREE -n thinpool/data
  Volume group "thinpool" not found
  Cannot process volume group thinpool
```

Correcting for the naming convention, this would be:

```
$ sudo lvextend -l+100%FREE -n docker/thinpool_tdata
  Can't resize internal logical volume docker/thinpool_tdata.
```

But that also fails, so the intent is probably:

```
$ sudo lvextend -l+100%FREE -n docker/thinpool
  Size of logical volume docker/thinpool_tdata changed from 100.00 GiB (25599 extents) to 195.00 GiB (49921 extents).
  Logical volume docker/thinpool_tdata successfully resized.
```

engine/userguide/storagedriver/device-mapper-driver.md
```diff
@@ -504,196 +466,303 @@ The `Data Space` values show that the pool is 100GB total. This example extends
...
 
+    Size of logical volume docker/thinpool_tdata changed from 95.00 GiB (24319 extents) to 198.00 GiB (50688 extents).
+    Logical volume docker/thinpool_tdata successfully resized.
+    ```
 
-    d.  Reload the thin pool with the new sector number
+4.  Verify the new thin pool size using the `Data Space Available` field in the
+    output of `docker info`. If you extnded the `thinpool/metadata` logical
```
trapier: Same point on naming here.  Intent is probably `docker/thinpool_meta`:

```
$ lvextend -L+1G -n docker/thinpool_tmeta
  Size of logical volume docker/thinpool_tmeta changed from 2.00 GiB (511 extents) to 3.00 GiB (767 extents).
  Logical volume docker/thinpool_tmeta successfully resized.
```

engine/userguide/storagedriver/btrfs-driver.md
```diff
@@ -6,216 +6,98 @@ title: Docker and Btrfs in practice
...
+  $ sudo cat /proc/filesystems | grep btrfs
 
-![](images/btfs_snapshots.jpg)
+  btrfs
+  ```
 
-Btrfs allocates space to subvolumes and snapshots on demand from an underlying
-pool of storage. The unit of allocation is referred to as a *chunk*, and
-*chunks* are normally ~1GB in size.
+  If support is not included, install the `btrfsprogs` package (SLES) or
```
dmcgowan: This doesn't sound right. Can you have `btrfs` missing from `/proc/filesystems` but still have the module loaded `lsmod | grep btrfs`. My understanding is `/proc/filesystems` returns the support filesystems in the kernel and if it is missing there you need a different kernel. The progs and tools just provide utilities for operating with btrfs since btrfs operations require ioctls which are not supports by the standard fs tools, but they are not required for mounting and using btrfs.

engine/installation/linux/centos.md
mstanleyjones: See #3350 where I tried separating out CE and EE into separate topics.

mstanleyjones:
Addressed the latest feedback and rebased this PR. I don'tknow if we have source for the graphics anymore. Let me see what I can do.

mstanleyjones:
Any pending objections to me merging this? I believe it reflects the current reality.

mstanleyjones:
I'm going to merge this so we can iterate. 🙈

engine/installation/linux/suse.md
```diff
@@ -149,13 +180,25 @@ the repository.
     $ sudo zypper install docker-ee-<VERSION_STRING>
     ```
 
-    Start Docker:
+4.  Configure Docker to use the `btrfs` filesystem. Edit the file
```
friism: @mstanleyjones  Why do people have to configure it if it's used by default already?

engine/userguide/storagedriver/aufs-driver.md
```diff
@@ -1,251 +1,236 @@
...
+     Backing Filesystem: extfs
+     Dirs: 0
+     Dirperm1 Supported: true
+    <truncated output>
+    ```
 
-## Deleting files with the AUFS storage driver
+3.  If you are using a different storage driver, then it means that either AUFS
+    is not included in the kernel (in which case a different default driver will
+    be used) or that means that Docker has been explicitly configured to use a
```
friism: @mstanleyjones that -> it

engine/userguide/storagedriver/aufs-driver.md
```diff
@@ -1,251 +1,236 @@
...
+    different driver. Check `/etc/docker/daemon.json` or the output of
+    `ps auxw |grep dockerd` to see if Docker has been started with the
```
friism: @mstanleyjones  missing space after pipe

engine/userguide/storagedriver/aufs-driver.md
```diff
@@ -1,251 +1,236 @@
...
 
-A container's thin writable layer is stored in a directory under
-`/var/lib/docker/aufs/diff/`. With Docker 1.10 and higher, container IDs no
-longer correspond to directory names. However, the containers thin writable
-layer still exists here and is stacked by AUFS as the top writable layer
-and is where all changes to the container are stored. The directory exists even
- if the container is stopped. This means that restarting a container will not
-lose changes made to it. Once a container is deleted, its thin writable layer
-in this directory is deleted.
+- **Renaming directories**: Calling `rename(2)` for a directory is not fully
```
friism: wat. @mstanleyjones so if one is using aufs, the app will behave differently than with other graph drivers?

engine/userguide/storagedriver/imagesandcontainers.md
```diff
@@ -12,29 +12,51 @@ stores images. Then, you need an understanding of how these images are used by
...
+stacked on top of each other. When you create a new container, you add a new
+writable layer on top of the underlying layers. This layer is often called the
+"container layer". All changes made to the running container, such as writing
+new files, modifying existing files, and deleting files, are written to this thin
+writable container layer. The diagram below shows a container based on the Ubuntu
+15.04 image.
 
-The Docker storage driver is responsible for stacking these layers and
-providing a single unified view.
+The diagram below shows a container with the Ubuntu 15.04 image comprising 4
```
friism: @mstanleyjones is there duplication with previous para here?

engine/userguide/storagedriver/imagesandcontainers.md
```diff
@@ -12,29 +12,51 @@ stores images. Then, you need an understanding of how these images are used by
...
+![Docker image layers](images/container-layers.jpg)
 
-![](images/container-layers.jpg)
+A _storage driver_ handles the details about the way these layers interact with
+each other. Different storage drivers are available, which have advantages
+and disadvantages in different situations.
 
 ### Content addressable storage
 
+If you are already running Docker 1.10 or higher, you can skip this section and
```
friism: @mstanleyjones 1.10 is no longer supported. Can we remove this or archive for clarity?

engine/userguide/storagedriver/selectadriver.md
```diff
@@ -4,40 +4,187 @@ keywords: container, storage, driver, AUFS, btfs, devicemapper,zvfs
...
+When in doubt, the best all-around configuration is to use a modern Linux
+distribution with a kernel that supports the `overlay2` storage driver, and to
+use Docker volumes for write-heavy workloads instead of relying on writing data
+to the container's writable layer.
+
+### Docker for Mac and Docker for Windows
+
+Docker for Mac and Docker for Windows only support `overlay2` `aufs`, or
+`overlay`. However, `overlay` is not recommended. `aufs` is the default in
+stable releases and `overlay2` is the default in Edge releases. In addition,
```
friism: @mstanleyjones something's missing here?

engine/userguide/storagedriver/selectadriver.md
```diff
@@ -4,40 +4,187 @@ keywords: container, storage, driver, AUFS, btfs, devicemapper,zvfs
...
+If your enterprise uses SAN, NAS, hardware RAID, or other shared storage
+systems, they may provide high availability, increased performance, thin
+provisioning, deduplication, and compression. In many cases, Docker can work on
+top of these storage systems, but Docker does not closely integrate with them.
+
+Each Docker storage driver is based on a Linux filesystem or volume manager. Be
+sure to follow existing best practices for operating your storage driver
+(filesystem or volume manager) on top of your shared storage system. For
+example, if using the ZFS storage driver on top of a shared storage system, be
+sure to follow best practices for operating aZFS filesystems on top of that
```
friism: @mstanleyjones "aZFS"?

engine/userguide/storagedriver/selectadriver.md
```diff
@@ -100,149 +213,23 @@ settings in the file, it should have the following contents:
...
-is only compatible with Linux kernel 4.0 and later. For users on a pre-4.0
-kernel or with an existing `overlay` graph, it is recommended to stay
-on `overlay`. For users with at least a 4.0 kernel and no existing or required
-`overlay` graph data, then `overlay2` may be used.
-
-> **Note**
-> `overlay2` graph data will not interfere with `overlay` graph data. However
-> when switching to `overlay2`, the user is responsible for removing
-> `overlay` graph data to avoid storage duplication.
+Before you change the default storage driver on Docker for Mac or Docker for
```
friism: @mstanleyjones why would we ever encourage users to change this?

engine/userguide/storagedriver/aufs-driver.md
thaJeztah: correct; there are some corner cases; not all graph drivers are fully POSIX compliant (outside of docker)

engine/installation/linux/suse.md
mstanleyjones: They don't, really. But they would if their `/var/lib/docker/` wasn't BTRFS or if `/` was not BTRFS, so it made sense to ask them to do it. Customers might ship the `daemon.json` around too.

engine/userguide/storagedriver/aufs-driver.md
mstanleyjones: That really doesn't matter to the shell, but I'll add the space because it looks prettier.

engine/userguide/storagedriver/aufs-driver.md
```diff
@@ -1,251 +1,236 @@
...
 
-When you have verified that you meet the prerequisites, instruct the Docker daemon to use
-AUFS by starting the Docker daemon with the flag `--storage-driver=aufs`:
+If a container is running, the contents of `/var/lib/docker/aufs/` change in the
+following ways:
 
-```bash
-$ sudo dockerd --storage-driver=aufs &
-```
+> **Note**: If you see files or directories which end in `-init`, these are used
```
mstanleyjones: Correct, I thought I added an admonition about that.

engine/userguide/storagedriver/selectadriver.md
mstanleyjones: I have no idea. I only know that it is a possible thing. @thaJeztah any thoughts?

engine/installation/linux/suse.md
friism: ok, would it be worth adding a note that this is only required under certain circumstances?

engine/userguide/storagedriver/aufs-driver.md
friism: @thaJeztah shouldn't that disqualify aufs as a default/recommendation?

engine/userguide/storagedriver/aufs-driver.md
friism: I agree, purely cosmetic

engine/userguide/storagedriver/aufs-driver.md
thaJeztah: @friism what's the alternative? How many bug reports have we had for this?

engine/userguide/storagedriver/selectadriver.md
thaJeztah: There's some corner cases where one could give benefits over the other. I'm not sure we _need_ to document it, other than that it saves us answering questions "and how do I do this on Mac"?

PR #3278, opened 2017-05-16
Added unless-stopped restart policy

zasca:


mstanleyjones:
Is this a new option? When was it introduced?

zasca:
Unfortunately I could not find from which version it was added, but in the official documentation this option is: https://docs.docker.com/engine/reference/run/#restart-policies---restart

PR #3296, opened 2017-05-17
Specify Compose version for Compose file formats

tahia-khan:
### Proposed changes

Explicitly specifying Compose versions for file formats 2.1 and 2.2. 

Context: I had Compose v.1.8.1 and was getting a version error when trying to "docker-compose up" on a 2.1 Compose file. This page states "Version 2 files are supported by **Compose 1.6.0+** ", but I didn't immediately realize that this doesn't cover file formats 2.1 and 2.2.


tahia-khan:
Oops, wrong branch

PR #3297, opened 2017-05-17
Specify Compose version for Compose file formats

tahia-khan:
### Proposed changes

Explicitly specifying Compose versions for file formats 2.1 and 2.2. 

Context: I had Compose v.1.8.1 and was getting a version error when trying to "docker-compose up" on a 2.1 Compose file. This page states "Version 2 files are supported by **Compose 1.6.0+** ", but I didn't immediately realize that this doesn't cover file formats 2.1 and 2.2.

### DIFF
--- a/compose/compose-file/compose-versioning.md
+++ b/compose/compose-file/compose-versioning.md
@@ -171,7 +171,7 @@ between services and startup order.
 ### Version 2.1
 
 An upgrade of [version 2](#version-2) that introduces new parameters only
-available with Docker Engine version **1.12.0+**. Version 2.1 files are 
+available with Docker Engine version **1.12.0+**. Version 2.1 files are
 supported by **Compose 1.9.0+**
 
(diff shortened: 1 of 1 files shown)


londoncalling:
@shin- can you just take a quick look and verify that the proposed updates are accurate, and whether we need to add this kind of clarification elsewhere?

compose/compose-file/compose-versioning.md
```diff
@@ -186,8 +187,9 @@ Introduces the following additional parameters:
 ### Version 2.2
 
 An upgrade of [version 2.1](#version-21) that introduces new parameters only
-available with Docker Engine version **1.13.0+**. This version also allows
-to specify default scale numbers inside the service's configuration.
+available with Docker Engine version **1.13.0+**.  Version 2.1 files are
```
shin-: -2.1 +2.2

PR #3300, opened 2017-05-17
Update Wrong directory for Ubuntu 16.04

petertanath:


mstanleyjones:
I'm not sure this change is correct. Generally, `/usr/bin` is not the right place to save binaries that are not managed by your OS's package manager. WDYT @shin- ?

shin-:
Yes, /usr/local/bin is the correct directory for this. The docs should not
be changed.

PR #3305, opened 2017-05-17
Update part3.md

cschooley:
Following the steps above I spent several hours trying to "Troubleshoot slow Docker swarm performance" as each web server in my stack was taking 'forever' to respond to HTTP requests, but worked just fine when invoked directly. I noticed it was taking almost exactly 30 seconds every time, and when finishing the turorial (thus deploying Redis) I realized that the python apps were waiting for some kind of Redis timeout. After deploying with a working Redis service responses were quick as before.

As somebody new to Docker it would have been good to know that the web app itself (timing out for Redis) was causing the delay not the host system or docker configuration.

### DIFF
--- a/get-started/part3.md
+++ b/get-started/part3.md
@@ -122,6 +122,11 @@ your browser and hit refresh a few times. Either way, you'll see the container
 ID randomly change, demonstrating the load-balancing; with each request, one of
 the five replicas is chosen at random to respond.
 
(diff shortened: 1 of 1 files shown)


mstanleyjones:
Thanks for figuring this out and getting us the feedback! I'm sorry you ran into trouble.

PR #3318, opened 2017-05-18
Update restart.md

scjane:
### Proposed changes

Change "ip address" to "IP address".


johndmulhausen:
It would be nice if we had a better description but the PR looks good and has been stalled 10 days over that. Waving it in.

PR #3345, opened 2017-05-22
Clarify instructions under `create a dockercloud-swarm-role`

JasonMFry:
### Proposed changes

Add clarification to several steps under `create a dockercloud-swarm-role 
role with an embedded policy` to increase clarity and prevent confusion.


londoncalling:
LGTM

PR #3346, opened 2017-05-22
Compose examples

### LINKED ISSUE #3024: Rails project should run bundle before build image
[compose/rails.md](https://docs.docker.com/compose/rails/), CC @londoncalling

In section "Build the project", if use "--skip-bundle" when create project, `Gemfile.lock` will be empty. Event if build image in the next step, because volume do not mount in build process, changes in `Gemfile.lock` will not sync to host. However `Gemfile.lock` should be check into version control (http://bundler.io/v1.14/man/bundle-install.1.html#THE-GEMFILE-LOCK).

My opinion is do not use "--skip-bundle" when create project:

```bash
docker-compose run web rails new . --force --database=postgresql
```

If edit `Gemfile` later, we should run bundle first in order to update `Gemfile.lock`:

```bash
docker-compose run web bundle
```

Then rebuild image:

```bash
docker-compose build
```

### LINKED ISSUE #3337: Not obvious flask defaults to using port 5000
[compose/gettingstarted.md](https://docs.docker.com/compose/gettingstarted/), CC @londoncalling

I was a bit confused when I encountered port 5000 in step 3. I kept looking back to understand why we were referencing port 5000 and where it was "exposed". Didn't realize flask defaulted to port 5000.

Suggested fix to add a note in step 1 that flask defaults to port 5000 like you noted for redis?

londoncalling:
### What's changed

- Improved Compose Rails example per user feedback

- Updated Compose Getting Started example per user feedback

This is first take, I already see things wrong with what I wrote about rebuilding the app, so I'm re-running some tests and will push an updated version of this topic of this soon.

### Related

Fixes #3024 
Fixes #3337

### Reviewers

@chloerei @dnephin @shin- I think this is close, will merge tonight to get the bulk of the changes published, then we can iterate as needed.

@ManoMarks I'm going to do the `docker stack deploy` updates in a separate PR.


londoncalling:
LGTM

PR #3347, opened 2017-05-22
Remove an incorrect step under `create a swarm`

JasonMFry:
### Proposed changes

Remove the instruction to `provide an App name` because that is not 
an option when creating a new swarm on AWS in Docker Cloud.


londoncalling:
LGTM

PR #3355, opened 2017-05-23
Added Instruction for Visual Studio Subscriptions

dstrebel:
For Azure Visual Studio MSDN subscription users will need to enable "programmatic deployments" on the Docker CE VM Marketplace item.


friism:
lgtm

PR #3366, opened 2017-05-24
Add pids_limit to upgrade docs

shin-:
### Proposed changes

The `pids_limit` option added for the 2.1 format in Compose was missing from the upgrade docs

### Related issues (optional)

docker/compose#4792


johndmulhausen:
Thanks, @shin- :)

PR #3378, opened 2017-05-24
Update part4

joeabbey:
Installation of VirtualBox should be in the requirements.

### Proposed changes

Following the tutorial, a user may hit an error by easily overlooking need to have VirtualBox installed.

```
docker-machine create --driver virtualbox myvm1
Creating CA: /Users/joeabbey/.docker/machine/certs/ca.pem
Creating client certificate: /Users/joeabbey/.docker/machine/certs/cert.pem
Running pre-create checks...
Error with pre-create check: "VBoxManage not found. Make sure VirtualBox is installed and VBoxManage is in the path"
```

Adding an explicit requirement for VirtualBox in the "Requirements" section will help a first-time user avoid the error.

### DIFF
--- a/get-started/part4.md
+++ b/get-started/part4.md
@@ -8,6 +8,7 @@ description: Learn how to create clusters of Dockerized machines.
 ## Prerequisites
 
 - [Install Docker version 1.13 or higher](/engine/installation/).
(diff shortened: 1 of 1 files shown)


johndmulhausen:
Fair enough; but added the caveat that Hyper-V users should stick to that :)

PR #3426, opened 2017-05-29
Only `sut` service and its dependents are started

wvanderdeijl:
### Proposed changes

Documentation update to prevent others from trying to diagnose failing tests for hours, like we did


londoncalling:
LGTM with my copy-edits

PR #3431, opened 2017-05-30
Add dateUpdated and datePublished for Google

### LINKED ISSUE #1612: Make Google detect the publish date
[engine/swarm/secrets.md](https://docs.docker.com/engine/swarm/secrets/)

Hi, folks.  I was searching for a guide on Docker secrets, and I was surprised by the rank of this page in Google results:

image

I noticed the date was listed as "Dec 17, 2013", which I thought was odd until I noticed this about midway down the guide:

image

I think Google (and probably other web crawlers) are picking up that "17 Dec 13" in the example output as the publish date of the article (it's the first "date-ish" looking thing), and might be demoting the page in results as being old information.  I'd recommend adding the real publish date higher up in the article near the heading.

(Also, your docs are generally amazing, thank you!)

mstanleyjones:
### Proposed changes

Fixes #1612

Solution found on https://webmasters.stackexchange.com/questions/95792/how-to-get-google-to-display-the-published-date-of-blog-post-in-the-search-resul

### DIFF
--- a/_layouts/docs.html
+++ b/_layouts/docs.html
@@ -49,7 +49,7 @@ else %}{% assign edit_url = "" %}{% endif %} {% break %} {% endif %} {% endfor %
 	<meta property="og:image" content="/favicons/docs@2x.ico"/>
 	<!-- metadata -->
 	<meta property="og:type" content="website"/>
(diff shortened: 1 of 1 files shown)


mstanleyjones:
I don't think this can actually break anything so I'm going to merge it so we can see if it works. :)

PR #3432, opened 2017-05-30
Update webhooks.md to include `tag` in payload

glortho:
### Proposed changes

Show that `tag` is available as part of the `push_data` payload.


londoncalling:
LGTM

PR #3436, opened 2017-05-30
Update docs about starting containers automatically

### LINKED ISSUE #3285: Can we get a supervisord example for automatically restarting services
[engine/admin/host_integration.md](https://docs.docker.com/engine/admin/host_integration/), CC @mstanleyjones

The docs show example for Upstart & SystemD. Although the introduction mentions supervisor, there are no examples for it.

mstanleyjones:
Provide an example for using restart policies, and remove the specific examples for
process managers.

Fixes #3285


engine/admin/start-containers-automatically.md
```diff
@@ -0,0 +1,57 @@
...
+## Use a restart policy
+
+To configure the restart policy for a container, use the `--restart` flag
+when using the `docker run` command. The value of the `--restart` flag can be
+any of the following:
+
+| Flag             | Description                                                                                     |
+|:-----------------|:------------------------------------------------------------------------------------------------|
+| `no`             | Do not automatically restart the container. (the default)                                       |
+| `on-failure`     | Restart the container if it exits due to an error.                                              |
```
thaJeztah: Should we mention "non-zero exit code of the container's process"? (to be more specific about "error")?

engine/admin/start-containers-automatically.md
```diff
@@ -0,0 +1,57 @@
...
+$ docker run -dit --restart unless-stopped redis
+```
+
+## Use a process manager
+
+If restart policies don't suit your needs, such as when processes outside
+Docker depend on Docker containers, you can use a process manager such as
+[upstart](http://upstart.ubuntu.com/),
+[systemd](http://freedesktop.org/wiki/Software/systemd/) or
+[supervisor](http://supervisord.org/) instead.
```
thaJeztah: Supervisor (outside of containers) is not very common to use, so I'd consider removing it here

For the others; we can keep it for now, but I'm certainly "+1" to remove them on the next release or so

engine/admin/start-containers-automatically.md
```diff
@@ -0,0 +1,57 @@
...
+| `unless-stopped` | Restart the container unless it is explicitly stopped or Docker itself is stopped or restarted. |
+| `always`         | Always restart the container if it stops.                                                       |
+
+The following example starts a Redis container and configures it to always
+restart unless it is explicitly stopped or Docker is restarted.
+
+```bash
+$ docker run -dit --restart unless-stopped redis
+```
+
```
thaJeztah: Some "pitfalls" when using restart policies (not sure they should be mentioned in this context, but writing down FYI);

- the restart policy takes effect after the container is started successfully (i.e., after the container is running 10 seconds or more, monitoring is started)
- this means that if a container _fails to start_, it is not re-attempted to start
- of course; this is different with _services_ because then swarm kit will keep retrying to start new tasks
- restart policies do _not_ take effect if you manually stop a container; in that case, the container is not restarted, however it will be restarted when the daemon is restarted. The container is not restarted, because manually stopping a container is a deliberate choice by the user.

engine/admin/start-containers-automatically.md
```diff
@@ -0,0 +1,57 @@
...
+## Use a process manager
```
thaJeztah: Should we mention the `--live-restore` option in the context of this document? i.e. allow keeping the containers running when restarting the daemon?

mstanleyjones:
Not removing `supervisord` from the list for now since it's the one the OP was asking about. Addressed feedback. Thanks for the details!

engine/admin/start-containers-automatically.md
```diff
@@ -0,0 +1,80 @@
...
+```
+
+### Restart policy details
+
+Keep the following in mind when using restart policies:
+
+- A restart policy only takes effect after a container starts successfully. In
+  this case, starting successfully means that the container is up for at least
+  10 seconds and Docker has started monitoring it. This prevents a container
+  which does not start at all from into a restart loop.
```
thaJeztah: missing word "from into"

engine/admin/start-containers-automatically.md
gbarr01: @thaJeztah , per your 4th bullet point, should @mistyhacks 's text on `unless-stopped` include a 2nd `explicitly`? or should the part about `Docker` be removed entirely? 

> Restart the container unless it is explicitly stopped or Docker itself is _explicitly_ stopped or restarted.

gbarr01: I ask because of #6287

thaJeztah: Yes, I think that PR is correct (there have been some corner-cases though, so I guess the behaviour wasn't fully defined for all of those)

gbarr01: I see. Thank you.

PR #3294, opened 2017-05-17
Getstarted tweaks

londoncalling:
### Quick summary

@johndmulhausen I'm walking through the getting started, and here are edits so far, and expansion of some topics. The tutorial is beautifully written, and friendly. My goal is to keep with your style, but make some improvements where I had questions as a guinea pig user.

Adding @mstanleyjones  as a reviewer, for another viewpoint. (Still TBD for my walkthrough are parts 5 stacks, part 6 deploy.)

### What's changed

- copyedits
- added subtopics to longer steps like "Share your image", and showed more example outputs for commands
- more details and a screen snap of showing app in a web browser
- added output for `curl` command (so that this will be recognizable when they watch the asciinema replay of commands)
- added an example of image tagging syntax (the same one used on the asciicinema to better illustrate how repository and tag naming are used, and to map to the asciicinema replay of commands)
- added more example outputs and commands in general, where not having them could potentially confuse someone
- explained logging in to Docker and using Docker ID a little more
- improved the notes a bit, re-styled when they were breaking
- added more info on "cleanup" topics
- clarified Compose and Machine as prerequisites and how to get them
- @mstanleyjones review comments incorporated
- @johndmulhausen, updates per our discussion (reminders and code comments about replacing `username/repo:tag` with user-specific image details, added prerequisites to make sure docker machines are running and swarm is up, screen snaps of visualizer and web page with `redis` service, broke out steps on "Stacks" topic so that it's less likely users skim, and skip steps like creating a `data` directory)
- In part 6 under "create your swarm", fixed links to AWS, and added a link for Azure 

### Reviewers

@johndmulhausen @JimGalasyn @mstanleyjones 

@jkorab, added Docker Machine as a prerequisite on Linux systems in part 4 and subsequent parts, per discussion thread on #3449

PR #3449, opened 2017-05-31
Added prerequisite for docker-machine

jkorab:
Added Docker Machine to prerequisites section, alongside link to install


mstanleyjones:
I don't think this is needed on Docker for Mac or Docker for Windows or Docker EE for Windows Server. PTAL @londoncalling

londoncalling:
@jkorab well, you have a good point, as you would need to install Docker Machine on Linux. (It comes with Docker for Mac and Windows.) But it should be up in the prerequisites. I'm working on a PR with edit suggestions for the Getting Started tutorial as a whole. I'm going to close this PR, but cc you on that other PR, which is #3294. I'll add this change to mention the need for Docker Machine on Linux into that PR.

PR #3453, opened 2017-05-31
Update dockerurl config file URL

rickerc:
The instructions as followed result in an incorrect URL because the platform is not inserted. Update docs to add platform into the URL as required

### Proposed changes

The <DOCKER-EE-URL> does not specify platform so needs it appended to match how the repositories are laid out in Docker's yum repositories

### DIFF
--- a/engine/installation/linux/rhel.md
+++ b/engine/installation/linux/rhel.md
@@ -89,7 +89,7 @@ the repository.
       [prerequisites](#prerequisites).
 
       ```bash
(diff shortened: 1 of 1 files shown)


mstanleyjones:
This change is incorrect, because the installation topic is discussing the Docker EE for RHEL entitlement. If you have the Docker EE for Linux entitlement, Docker Store has some additional instructions for how you need to modify the URL. Apoligies for the confusion.

cc/ @friism

PR #3454, opened 2017-05-31
Rewrite swarm service networking guide

### LINKED ISSUE #2998: Document customizing the ingress network
### Problem description

https://github.com/docker/docker/pull/31714

Docker 17.06 (first available in 17.05 Edge)

mstanleyjones:
- Reorganize and update info
- Add more info about each type of network
- Add info about customizing docker_gwbridge
- Add info about customizing ingress
- Expand on encryption options

Fixes #2998 

PTAL @mavenugo @thajestah @dongluochen @mark-church

### DIFF
--- a/engine/swarm/networking.md
+++ b/engine/swarm/networking.md
@@ -1,77 +1,215 @@
 ---
 description: Use swarm mode networking features
-keywords: guide, swarm mode, swarm, network
-title: Attach services to an overlay network
+keywords: swarm, networking, ingress, overlay, service discovery
+title: Manage swarm service networks
 ---
 
-Docker Engine swarm mode natively supports **overlay networks**, so you can
-enable container-to-container networks. When you use swarm mode, you don't need
-an external key-value store. Features of swarm mode overlay networks include the
-following:
+Docker swarms generate two different kinds of traffic:
 
-* You can attach multiple services to the same network.
-* By default, **service discovery** assigns a virtual IP address (VIP) and DNS
-entry to each service in the swarm, making it available by its service name to
-containers on the same network.
-* You can configure the service to use DNS round-robin instead of a VIP.
+- **Control and management plane traffic**: This includes information about
+  swarm management, such as joining and leaving the swarm. This traffic is
+  always encrypted to protect the integrity of the swarm itself.
 
-In order to use overlay networks in the swarm, you need to have the following
-ports open between the swarm nodes before you enable swarm mode:
+- **Application data plane traffic**: This includes communication between swarm
+  tasks and other tasks, between tasks and containers, and between tasks and
+  external clients.
+
+This topic discusses how to manage the application data for your swarm services.
+For more details about swarm networking in general, see the
+[Docker networking reference architecture](https://success.docker.com/Architecture/Docker_Reference_Architecture%3A_Designing_Scalable%2C_Portable_Docker_Container_Networks){: target="_blank" class="_" }.
+
+The following three network concepts are important to swarm services:
+
+- **Overlay networks** manage communications among the Docker daemons
+  participating in the swarm. You can create overlay networks, in the same way
+  as user-defined networks for standalone containers. You can attach a service
+  to one or more existing overlay networks as well, to enable service-to-service
+  communication. Overlay networks are Docker networks that use the `overlay`
+  network driver.
+
+- The **ingress network** is a special overlay network that facilitates
+  load balancing among a service's nodes. When any swarm node receives a
+  request on a published port, it hands that request off to a module called
+  `IPVS`. `IPVS` keeps track of all the IP addresses participating in that
+  service, selects one of them, and routes the request to it, over the
+  `ingress` network.
+
+  The `ingress` network is created automatically when you initialize or join a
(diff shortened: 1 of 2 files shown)


mstanleyjones:
Pending question: What happens if you update a service's endpoint mode (it seems to be allowed in the CLI).

engine/swarm/networking.md
```diff
@@ -125,179 +126,228 @@ $ docker network inspect my-network
...
 
-2.  Lookup the node where `my-busybox` is running:
+Service task containers connected to an overlay network can communicate with
+each other across it.
 
-    ```bash
-    $ docker service ps my-busybox
+To see which networks a service is connected to, use `docker service ls` to find
+the name of the service, then `docker service ps <service-name>` to list the
+networks. Alternately, to see which task containers are connected to a network,
```
dongluochen: `task container` -> `containers`?

engine/swarm/networking.md
```diff
@@ -125,179 +126,228 @@ $ docker network inspect my-network
...
+Service discovery can work in two different ways, using a virtual IP (VIP) or
+DNS round robin (DNSRR). You can configure this per service.
 
-    You can deduce the container name as `<TASK-NAME>`+`<ID>`. Alternatively,
-    you can run `docker ps` on the node where the task is running.
+- By default, when you attach a service to a network, Docker assigns the service
+  a virtual IP (VIP), which is the "front end" for clients to reach the service.
+  Docker keeps a list of all worker nodes in the service, and routes requests
+  between the client and one of the nodes. Each request from the client might be
+  routed to a different node. This mechanism uses a protocol called `gossip`.
```
dongluochen: > This mechanism uses a protocol called `gossip`.

Gossip is the way how swarm nodes propagate networking changes. It's not directly supporting load balancing. I think this sentence should be removed from here.

engine/swarm/networking.md
```diff
@@ -125,179 +126,228 @@ $ docker network inspect my-network
...
-4.  From inside the busybox container, query the DNS to view the VIP for the
-`my-web` service:
+- If you configure a service to use DNS round-robin (DNSRR) service discovery,
+  there is not a single virtual IP. Instead, Docker sets up DNS entries for the
+  service such that a DNS query for the service name returns a list of IP
+  addresses, and the client connects directly to one of these.
 
-    ```bash
-    $ nslookup my-web
+  You should only use DNSRR service discovery if your service uses an external
```
dongluochen: I think the original sentence is more accurate. "DNS round-robin is useful in cases where you want to use your own load balancer."

engine/swarm/networking.md
```diff
@@ -125,179 +126,228 @@ $ docker network inspect my-network
...
-    Address 1: 10.0.9.2 ip-10-0-9-2.us-west-2.compute.internal
-    ```
+{% include edge_only.md section="feature" %}
 
-    >**Note**: the examples here use `nslookup`, but you can use `dig` or any
-    available DNS query tool.
+Most users never need to configure the `ingress` network, but Docker 17.05 and
+higher allow you to do so. This can be useful if the automatically-chosen subnet
+conflicts with one that already exists on your network, or you need to customize
+other TCP settings such as the MTU.
```
dongluochen: MTU is IP layer. MSS is TCP layer.

engine/swarm/networking.md
```diff
@@ -125,179 +126,228 @@ $ docker network inspect my-network
...
 
-5.  From inside the busybox container, query the DNS using a special query
-&lt;tasks.SERVICE-NAME&gt; to find the IP addresses of all the containers for the
-`my-web` service:
+Customizing the `ingress` network involves removing and recreating it. During
+the time that no `ingress` network exists, existing services continue to
+function but are not load-balanced. This effects services which publish ports,
```
dongluochen: effects -> affects?

engine/swarm/networking.md
```diff
@@ -125,179 +126,228 @@ $ docker network inspect my-network
...
+Most users never need to configure the `ingress` network, but Docker 17.05 and
+higher allow you to do so. This can be useful if the automatically-chosen subnet
+conflicts with one that already exists on your network, or you need to customize
+other TCP settings such as the MTU.
 
-5.  From inside the busybox container, query the DNS using a special query
-&lt;tasks.SERVICE-NAME&gt; to find the IP addresses of all the containers for the
-`my-web` service:
+Customizing the `ingress` network involves removing and recreating it. During
+the time that no `ingress` network exists, existing services continue to
```
dongluochen: Can you remove ingress network if there are services with published-port?

engine/swarm/networking.md
```diff
@@ -125,179 +126,228 @@ $ docker network inspect my-network
...
+function but are not load-balanced. This effects services which publish ports,
+such as a WordPress service which publishes port 80.
 
-    ```bash
-    $ nslookup tasks.my-web
+1.  Inspect the `ingress` network using `docker network inspect`, and stop any
```
dongluochen: There is no `stop` for services. I think you mean `remove`.

mstanleyjones:
Addressed feedback in a new commit.

PR #2884, opened 2017-04-19
Fixes #1683.  Correct misleading information around ports 4789 & 7946

clocklear:
### Proposed changes

Fixed incorrect information around ports 4789 and 7946 in UCP 2.0 / 2.1 documentation

### Related issues (optional)

Fixes #1683

PR #3455, opened 2017-06-01
Update system-requirements.md

Tfindelkind:
changed to TCP and UDP for port 4789.


joaofnfernandes:
Thanks for the contribution. Are you sure you need 4789/TCP? We had that before in the system requirements, but it seems that you don't really need it.
Have you found any issues that were caused by having that port blocked by a firewall? You can check #2884 for more context.

Tfindelkind:
Yep. There are two options than: 1. The installation pre-checks are wrong because these didn't let me install as long TCP 4789 was blocked by firewalld. 2. The port is still needed.

joaofnfernandes:
@alexmavr can you give your input here?

alexmavr:
I didn't know TCP 4789 was required at all. https://docs.docker.com/engine/swarm/ingress/ seems to indicate that it's not needed.

I just noticed that UCP does perform a TCP check for that port though. Can you specify whether this was through UCP or just `docker swarm init`?

joaofnfernandes:
> The installation pre-checks are wrong because these didn't let me install as long TCP 4789 was blocked
Seem to be the UCP pre-install checks that are not allowing you to install UCP if TCP 4789 is not opened

PR #3456, opened 2017-06-01
Remove redundant and incorrect step

rickerc:
The thinpool LVM profile creation step was (partially) in here twice, and the first time was incorrect and results in an unusable LVM profile. Consolidate the steps for correctness.

### Proposed changes

Consolidated steps to produce an LVM profile that doesn't error when applied


mstanleyjones:
I don't think this change is still valid. Please resolve the conflicts and see if there is actually anything that still needs to be changed. This topic has gone through signficant revision lately.

mstanleyjones:
Closed due to inactivity.

PR #3467, opened 2017-06-02
Update builds.md

scjane:


johndmulhausen:
Yes, much better, thank you @scjane !

PR #3476, opened 2017-06-02
Fix typo

djbb7:
### Proposed changes

Fixed a type. It was written 'to', but should be 'two'.


johndmulhausen:
Thanks, @djbb7 :)

PR #3480, opened 2017-06-02
Include example for docker swarm join

shanemgrey:
### Proposed changes

Clarify that the token, ip, and port needed for the command are displayed in the docker swarm init response, showing an example instead of parameters that may leave the reader trying to find the token if they missed it in the terminal.

### DIFF
--- a/get-started/part4.md
+++ b/get-started/part4.md
@@ -69,12 +69,7 @@ this:
 ```
 $ docker-machine ssh myvm1 "docker swarm init"
 Swarm initialized: current node <node ID> is now a manager.
-
-To add a worker to this swarm, run the following command:
-
-  docker swarm join \
-  --token <token> \
-  <ip>:<port>
+...
 ```
 
 > Getting an error about needing to use `--advertise-addr`?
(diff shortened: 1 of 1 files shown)


johndmulhausen:
We don't want to go in this direction. It reads and renders much worse to have a lengthy token in there instead of the format of the swarm command, and the utility of taking out the first code example is lost on me, I'm sorry to say. Thanks for the PR regardless. :)

shanemgrey:
I understand that perspective.  I submitted this because going the the getting started section (being a completely new to docker user), I didn't initially notice where to get the token and IP from and spent time trying to figure it out before reading the directions over again a few times. Likely that's just my problem.  The information is there, but I had hoped to make it more clear for others in the PR.

Maybe the first bit of code I had removed could instead be a full example response with the token and so on instead of the parameters?  Then the second part where I pasted an example could be as it was with the parameters, with the more clear language on which parts to copy from the terminal?

Thank you for the fast response to the PR.  The docs have been excellent. I'm just finding occasionally that there are spots where more knowledge is expected than has already been presented in the tutorial.

PR #3481, opened 2017-06-02
update centos step 4 to include correct path

peytonmcnully:
The DOCKER-EE-URL contains a distribution directory called centos. The docker-ee.repo is located in that dir. This PR updates that path to the correct EE installation path.


mstanleyjones:
This is not true for Docker EE for CentOS, but only CentOS if you have purchased Docker EE for Linux with all of the distributions.

PR #3485, opened 2017-06-03
Reorganize registry deployment guide

### LINKED ISSUE #3433: Document pushing foreign layers
https://github.com/moby/moby/pull/33151

mstanleyjones:
### Proposed changes

Rewrites and reorganization
Also add information about pushing non-distributable layers
into a private registry

Fixes #3433 

Docker 17.06

PTAL @stevvooe @dmcgowan @thaJeztah @cpuguy83

### DIFF
--- a/registry/deploying.md
+++ b/registry/deploying.md
@@ -1,94 +1,240 @@
 ---
 description: Explains how to deploy a registry
 keywords: registry, on-prem, images, tags, repository, distribution, deployment
-title: Deploying a registry server
+title: Deploy a registry server
 ---
 
-You need to [install Docker version 1.6.0 or newer](/engine/installation/index.md).
+Before you can deploy a registry, you need to install Docker on the host.
+A registry is an instance of the `registry` image, and runs within Docker.
 
-## Running on localhost
+This topic provides basic information about deploying and configuring a
+registry. For an exhaustive list of configuration options, see the
+[configuration reference](configuration.md).
 
-Start your registry:
+If you have an air-gapped datacenter, see
+[Considerations for air-gapped registries](#considerations-for-air-gapped-registries).
 
-    docker run -d -p 5000:5000 --restart=always --name registry registry:2
+## Run a local registry
+
+Use a command like the following to start the registry container:
+
+```bash
+$ docker run -d -p 5000:5000 --restart=always --name registry registry:2
+```
 
 You can now use it with docker.
 
-Get any image from the hub and tag it to point to your registry:
+## Copy an image from Docker Hub to your registry
 
-    docker pull ubuntu && docker tag ubuntu localhost:5000/ubuntu
+You can pull an image from Docker Hub and push it to your registry. The
+following example pulls the `ubuntu:16.04` image from Docker Hub and re-tags it
+as `my-ubuntu`, then pushes it to the local registry. Finally, the
+`ubuntu:16.04` and `my-ubuntu` images are deleted locally and the
+`my-ubuntu` image is pulled from the local registry.
 
-... then push it to your registry:
+1.  Pull the `ubuntu:16.04` image from Docker Hub.
 
-    docker push localhost:5000/ubuntu
+    ```bash
+    $ docker pull ubuntu:16.04
+    ```
 
-... then pull it back from your registry:
+2.  Tag the image as `localhost:5000/my-ubuntu`. This creates an additional tag
+    for the existing image.When the first part of the tag is a hostname and
+    port, Docker interprets this as the location of a registry, when pushing.
 
-    docker pull localhost:5000/ubuntu
+    ```bash
+    $ docker tag ubuntu:16.04 localhost:5000/my-ubuntu
+    ```
 
-To stop your registry, you would:
+3.  Push the image to the local registry running at `localhost:5000`:
 
-    docker stop registry && docker rm -v registry
+    ```bash
+    $ docker push localhost:5000/my-ubuntu
+    ```
 
-## Storage
+4.  Remove the locally-cached `ubuntu:16.04` and `localhost:5000/my-ubuntu`
+    images, so that you can test pulling the image from your registry. This
+    does not remove the `localhost:5000/my-ubuntu` image from your registry.
 
-By default, your registry data is persisted as a [docker volume](/engine/tutorials/dockervolumes.md) on the host filesystem. Properly understanding volumes is essential if you want to stick with a local filesystem storage.
+    ```bash
+    $ docker image remove ubuntu:16.04
+    $ docker image remove localhost:5000/my-ubuntu
+    ```
 
-Specifically, you might want to point your volume location to a specific place in order to more easily access your registry data. To do so you can:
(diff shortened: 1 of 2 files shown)


mstanleyjones:
PTAL @friism

registry/deploying.md
```diff
@@ -1,94 +1,240 @@
...
+[configuration reference](configuration.md).
 
-Start your registry:
+If you have an air-gapped datacenter, see
+[Considerations for air-gapped registries](#considerations-for-air-gapped-registries).
 
-    docker run -d -p 5000:5000 --restart=always --name registry registry:2
+## Run a local registry
+
+Use a command like the following to start the registry container:
```
thaJeztah: We should point out that the example here is just for testing / trying out, because;

- it's insecure (there's no TLS, nor authentication in place, and the port is published, so anyone has access to your registry)
- there's no volume used, so images are stored on the container's filesystem; removing the container removes all your images

registry/deploying.md
```diff
@@ -214,15 +394,68 @@ registry:
...
+
+- Certain images, such as the official Microsoft Windows base images, are not
+  distributable. This means that when you push an image based on one of these
+  images to your private registry, the non-distributable layers are **not**
+  pushed, but are always fetched from their authorized location. This is fine
+  for internet-connected hosts, but will not work in an air-gapped set-up.
+
+  In Docker 17.06 and higher, you can configure the Docker daemon to allow
+  pushing non-distributable layers to private registries, in this scenario.
+  **This is only useful in air-gapped set-ups in the presence of
```
friism: _Strictly_ speaking, this is also useful in bandwidth-constrained setups, or setups where one wants to not be dependent on fetching the base layers from MS servers. I have no quarrel with the current formulation though.

registry/deploying.md
```diff
@@ -214,15 +394,68 @@ registry:
...
+
+  3.  Restart the registry if it does not start automatically.
+
+  4.  When you push images to the registries in the list, their
+      non-distributable layers will be pushed to the registry.
+
+      >  **Warning**: Non-distributable artifacts typically have restrictions on
+      > how and where they can be distributed and shared. Only use this feature
+      > to push artifacts to private registries and ensure that you are in
+      > compliance with > any terms that cover redistributing non-distributable
```
friism: bonus `>`

mstanleyjones:
Addressed feedback.

docker-for-mac/networking.md
```diff
@@ -78,11 +78,10 @@ There are two scenarios that the above limitations will affect:
 
 #### I want to connect from a container to a service on the host
 
-The Mac has a changing IP address (or none if you have no network access). Our
-current recommendation is to attach an unused IP to the `lo0` interface on the
-Mac; for example: `sudo ifconfig lo0 alias 10.200.10.1/24`, and make sure that
-your service is listening on this address or `0.0.0.0` (ie not `127.0.0.1`).
-Then containers can connect to this address.
+The Mac has a changing IP address (or none if you have no network access). From
```
stevvooe: This should probably be clarified to OSX, so it is not confused with `MAC Address`.

engine/swarm/networking.md
```diff
@@ -348,6 +348,41 @@ order to delete an existing bridge. The package name is `bridge-utils`.
 
 5.  Initialize or join the swarm.
 
+## Use a separate interface for control and data traffic
```
stevvooe: Is this related to registry deployment?

registry/deploying.md
```diff
@@ -1,94 +1,246 @@
...
+  registry:2
+```
+
+## Storage customization
+
+### Customize the storage location
+
+By default, your registry data is persisted as a
+[docker volume](/engine/tutorials/dockervolumes.md) on the host filesystem.
+Just as any other container, you can use a bind mount instead. The following
```
stevvooe: "a bind mount or volume instead"

registry/deploying.md
```diff
@@ -1,94 +1,246 @@
...
+  --name registry \
+  -v `pwd`/data:/var/lib/registry \
+  registry:2
+```
 
-### Alternatives
+### Customize the storage back-end
 
-You should usually consider using [another storage backend](./storage-drivers/index.md) instead of the local filesystem. Use the [storage configuration options](./configuration.md#storage) to configure an alternate storage backend.
+Instead of the local filesystem, you can use a different storage back-end by
```
stevvooe: You might want to compare this to usage of a volume.

registry/deploying.md
```diff
@@ -1,94 +1,246 @@
...
+    ```
+
+3.  Restart the registry, directing it to use the TLS certificate. This command
+    bind-mounts the `certs/` directory into the container at `/certs/`, and sets
+    environment variables that tell the container where to find the `domain.crt`
+    and `domain.key` file.
+
+    ```bash
+    $ docker run -d \
+      -p 5000:5000 \
```
stevvooe: Can we do this example without port 5000? The section above makes that promise then the example breaks that promise. ;)

registry/deploying.md
```diff
@@ -1,94 +1,246 @@
...
+browser-trusted certificate. For more information on Let's Encrypt, see
+[https://letsencrypt.org/how-it-works/](https://letsencrypt.org/how-it-works/)
+and the relevant section of the
+[registry configuration](configuration.md#letsencrypt).
 
-### Alternatives
+### Use an insecure registry (testing only)
 
-While rarely advisable, you may want to use self-signed certificates instead, or use your registry in an insecure fashion. You will find instructions [here](insecure.md).
+It is possible to use a self-signed certificate, or to use our registry
```
stevvooe: You can still setup verification with a self-signed certificate that isn't used for testing.

registry/deploying.md
```diff
@@ -141,24 +294,45 @@ checks can be directed at "/", which will always return a `200 OK` response.
...
 
-The simplest way to achieve access restriction is through basic authentication (this is very similar to other web servers' basic authentication mechanism).
+The simplest way to achieve access restriction is through basic authentication
+(this is very similar to other web servers' basic authentication mechanism).
+This example uses native basic authentication using `htpasswd` to store the
+secrets.
 
 > **Warning**:
-> You **cannot** use authentication with an insecure registry. You have to [configure TLS first](deploying.md#running-a-domain-registry) for this to work.
+> You **cannot** use authentication with an insecure registry. You must
```
stevvooe: ...with authentication schemes that send credentials as clear text.

mstanleyjones:
@stevvooe the first two bits of feedback you had were for unrelated changes that just indicated this patch needed a rebase. Maybe @londoncalling can make note of them to be changed elsewhere.

registry/deploying.md
mstanleyjones: I don't think so. Line 134 already says it defaults to using a volume.

mstanleyjones:
Addressed feedback and added a service example that uses secrets to store the TLS certificate material.

mstanleyjones:
LGTM, merging into `vnext-engine`.

PR #3495, opened 2017-06-05
Add docker-ee release notes

cpuguy83:
### Proposed changes

Adds a section to the `Enterprise Edition` product manual for release notes.

### DIFF
--- a/_data/toc.yaml
+++ b/_data/toc.yaml
@@ -1229,6 +1229,8 @@ manuals:
   - path: /engine/installation/
     title: Install Docker EE
     nosync: true
+  - path: /enterprise/release-notes/
+    title: Release Notes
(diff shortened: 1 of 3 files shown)


joaofnfernandes:
This looks good. @cpuguy83 in terms of release process the person building/shipping the binaries should also ensure these release notes are updated.
Until now we have been doing this ad-hoc but we need to formalize this process a bit more. Do you have any release checklist on your side where we can add a list item for this?

cpuguy83:
@joaofnfernandes Not as ad-hoc as it may seem, but agree that having a checkbox for release notes would be good.

docker-ee releases have been missing because there haven't been a place for them.

joaofnfernandes:
The builds are taking so long that we're now reaching Netlify's timeouts

PR #3496, opened 2017-06-05
Explain precedence of environment variables

franz-josef-kaiser:
### Proposed changes

* Fix the statement _"`env_file` overrides `environment`"_, which is not true. 
* Give an example using a _NodeJS_ image as the environment has some specifics there like adding it from a `.json` file.
* Explain how this compares to `ARG` and `ENV` in a `Dockerfile`.

### DIFF
--- a/compose/environment-variables.md
+++ b/compose/environment-variables.md
@@ -67,7 +67,7 @@ You can set default values for any environment variables referenced in the Compo
     TAG=v1.5
 
     $ cat docker-compose.yml
-    version: '2.0'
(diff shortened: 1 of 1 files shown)


franz-josef-kaiser:
Can someone explain what is causing a problem for _Netlify_? I am just getting a 404.

joaofnfernandes:
Netlify is sad, but it's nothing to do with the PR. It's that the build is taking too long.
Let's see how it does this time around.

PR #3506, opened 2017-06-06
Remove unsupported link_local_ips param from v3 doc

shin-:
### Proposed changes

This option is not available in the v3 format

### Related issues (optional)

https://github.com/docker/compose/issues/4897


shin-:
cc @dnephin

johndmulhausen:
Hammer of Thor dropped

PR #3514, opened 2017-06-07
Correcting the docs.

clemenko:
Updated the docs to fix the chaining problem.

### DIFF
--- a/datacenter/dtr/2.2/guides/admin/configure/deploy-caches/chaining.md
+++ b/datacenter/dtr/2.2/guides/admin/configure/deploy-caches/chaining.md
@@ -70,8 +70,9 @@ middleware:
                 - https://<asia-cache-url>
           cas:
             - /certs/asia-cache-ca.pem
+            - /certs/dtr-ca.pem
 ```
 
-Since the China cache doesn't need to communicate directly with DTR,
-it only needs to trust the CA certificates for the next hop, in this case
-the CA certificate used by the Asia cache.
+Adding the CA for the upstream DTR is a good idea. 
+Similarly, the China cache will need to be registered with DTR. Please see [deploy a simple cache](https://docs.docker.com/datacenter/dtr/2.2/guides/admin/configure/deploy-caches/#deploy-a-simple-cache) for how to use the API. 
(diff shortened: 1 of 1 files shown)


mstanleyjones:
@joaofnfernandes PTAL

clemenko:
We have a customer that needs the documentation updated asap. This is for a High Value Customer!

joaofnfernandes:
@clemenko I understand that, but do you care to address the feedback?

clemenko:
Updated with feedback.

datacenter/dtr/2.2/guides/admin/configure/deploy-caches/chaining.md
```diff
@@ -70,8 +70,10 @@ middleware:
                 - https://<asia-cache-url>
           cas:
             - /certs/asia-cache-ca.pem
+            - /certs/dtr-ca.pem
 ```
 
-Since the China cache doesn't need to communicate directly with DTR,
-it only needs to trust the CA certificates for the next hop, in this case
-the CA certificate used by the Asia cache.
+Please note that DTR's CA is needed for the downstream cache.
```
mstanleyjones: Please don't say please. :) I think you can just say "DTR's CA is needed...."

datacenter/dtr/2.2/guides/admin/configure/deploy-caches/chaining.md
```diff
@@ -70,8 +70,10 @@ middleware:
...
             - /certs/asia-cache-ca.pem
+            - /certs/dtr-ca.pem
 ```
 
-Since the China cache doesn't need to communicate directly with DTR,
-it only needs to trust the CA certificates for the next hop, in this case
-the CA certificate used by the Asia cache.
+Please note that DTR's CA is needed for the downstream cache. 
+
+Similarly, the China cache will need to be registered with DTR. Please see [deploy a simple cache](/datacenter/dtr/2.2/guides/admin/configure/deploy-caches/#deploy-a-simple-cache) for how to use the API.
```
mstanleyjones: s/will need/needs

No please.

PR #3522, opened 2017-06-07
Grammar fix

JustinCappos:


johndmulhausen:
This is actually on purpose; we are an anti-gerund crowd around here when it comes to headings. :)

PR #3526, opened 2017-06-08
Update syslog documentation

kudabaev:
There is a typo of option:  “syslog=udp://1.2.3.4:1111” should be “syslog-address=udp://1.2.3.4:1111”

### DIFF
--- a/engine/admin/logging/syslog.md
+++ b/engine/admin/logging/syslog.md
@@ -59,7 +59,7 @@ You can set the logging driver for a specific container by using the
 
 ```bash
 docker run \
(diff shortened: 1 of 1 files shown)


joaofnfernandes:
This seems to be right, but while we're at it, @thaJeztah in this page we also have the equivalent config for the `daemon.json` file. Do you mind checking if it is correct, and what other properties are supported for the syslog driver?

thaJeztah:
The `daemon.json` should accept the same options as `--log-opt` (they're parsed by the same code)

PR #3527, opened 2017-06-08
state default is not permitted driver for network

samuelcolvin:
### Proposed changes

Explicitly state that `default` is not a valid driver for compose networks.

### Related issues 

https://github.com/docker/compose/issues/4905

### DIFF
--- a/compose/compose-file/compose-file-v2.md
+++ b/compose/compose-file/compose-file-v2.md
@@ -1083,6 +1083,9 @@ The default driver depends on how the Docker Engine you're using is configured,
 but in most instances it will be `bridge` on a single host and `overlay` on a
 Swarm.
 
+**Note:** although `driver` can be left blank to use the default, explicitly
+settings `driver: default` is invalid.
+
 The Docker Engine will return an error if the driver is not available.
 
     driver: overlay


joaofnfernandes:
@samuelcolvin thanks for the contribution, but I don't think we should merge this since it's negative documentation, in the sense that the docs should tell what is supported, and not the infinite number of things that are not supported.

So in this case, it's implied that if you don't specify a `driver`, the driver used is going to be the default one. Granted that there's a few samples in the same page showing
```
ipam:
  driver: default
```
So maybe we should update those samples and remove the `driver:default`.

ping @shin- for visibility about this inconsistency in the product.

samuelcolvin:
1. I asked by one of the maintainers of docker-compose to make the change https://github.com/docker/compose/issues/4905#issuecomment-306941066
2. The `ipam: driver: default` is a different case: `driver` is a variable in the `ipam` dict not in the root of the network definition. At least that's my understanding, the whole section is fairly confused.

joaofnfernandes:
Thanks for the fast reply. In docker/compose#4905 you said

>The docs here appear to give an example of using default, at the very least they don't make it clear that it's not allowed.

But I could not find any mention of this. As far as I can tell the docs say

> The default driver depends on how the Docker Engine you’re using is configured, but in most instances it will be bridge on a single host and overlay on a Swarm.

This implies that you should use `driver: bridge` or `driver:overlay`. Everything else is not allowed. 
Of course there's a usability issue with the error message that you got, and it seems we're already tracking that at moby/moby#29375

Sorry for making you bounce around multiple repos and spend time creating a PR to them close it. But I hope you understand.

shin-:
@samuelcolvin @joaofnfernandes My bad, I skimmed through the docs and thought we had `driver: default` in some network definition examples. I realize now that these were IPAM definitions.

samuelcolvin:
no problem, I got confused by it too.

PR #3544, opened 2017-06-09
Create deploy-app.md

scjane:

### DIFF
--- a/swarm/swarm_at_scale/deploy-app.md
+++ b/swarm/swarm_at_scale/deploy-app.md
@@ -252,7 +252,7 @@ Now, you can test your application.
 ## Extra Credit: Deployment with Docker Compose
 
 Up to this point, you've deployed each application container individually. This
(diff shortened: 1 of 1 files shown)


swarm/swarm_at_scale/deploy-app.md
```diff
@@ -252,7 +252,7 @@ Now, you can test your application.
 ## Extra Credit: Deployment with Docker Compose
 
 Up to this point, you've deployed each application container individually. This
-can be cumbersome especially because their are several different containers and
+can be cumbersome especially because they are several different containers and
```
joaofnfernandes: Maybe "there are" would be better?

scjane: I also wondered which one is better. Then, I chose "they are". Hard to tell the differences between them. Anyway, I change it to "there are".

PR #3546, opened 2017-06-09
Remove extraneous **warning**

jpetazzo:
The .warning class already prepends "Warning"

### DIFF
--- a/engine/installation/linux/ubuntu.md
+++ b/engine/installation/linux/ubuntu.md
@@ -243,7 +243,6 @@ Docker EE.
     </div>
 
 
-    > **Warning**:
     > If you have multiple Docker repositories enabled, installing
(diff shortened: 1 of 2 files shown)


mstanleyjones:
This won't work. @londoncalling please remind us how to fix these?

londoncalling:
LGTM now. Thanks @jpetazzo . We need to make our docs formats better. Our notes and admonitions "templates" are still a bit in work and not fully consistent. The way you had formatted this originally, you got the colour coding, and the word "Warning" showed up,  but it wasn't bold. I reformatted it with `{: .warning-vanilla}` which gives colour coding, and a little warning icon instead of a full word, then added "Warning" manually in bold. Check out the examples at: https://docs.docker.com/test/#admonitions-notes

PR #3547, opened 2017-06-09
Add bash information about return codes.

lucile-sticky:

### DIFF
--- a/compose/reference/up.md
+++ b/compose/reference/up.md
@@ -51,5 +51,6 @@ flag.
 If you want to force Compose to stop and recreate all containers, use the
 `--force-recreate` flag.
 
-If the command encounters an error, it returns 1.
-If user press `ctrl` + `C`, causing kill of containers, it returns 2.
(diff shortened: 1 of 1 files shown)


joaofnfernandes:
@shin- while this PR might be true today, do we have any assurances that the return codes won't change at any time?

shin-:
Hm, I don't think those numbers are subject to change at this point, but the PR isn't entirely accurate.

Exit code 2 will only be returned when shutdown is forced (Ctrl-C twice). If containers are allowed to stop normally (Ctrl-C once), exit code will be 0.

joaofnfernandes:
@lucile-sticky do you care to update your changes based on @shin- 's feedback

lucile-sticky:
@joaofnfernandes @mstanleyjones section about shell return values edited following @shin- comment!

joaofnfernandes:
@lucile-sticky looks like it's time to ship it 🚀

PR #3554, opened 2017-06-09
Document new DOCKER-USER iptables chain

### LINKED ISSUE #3518: Document filter chain for persistent iptables rules
### Problem description

https://github.com/docker/libnetwork/pull/1675

Docker 17.06

mstanleyjones:
### Proposed changes

Adds just a bit about iptables but not a full treatment for now.

Fixes #3518

Docker 17.06

### DIFF
--- a/engine/userguide/networking/index.md
+++ b/engine/userguide/networking/index.md
@@ -575,11 +575,11 @@ or directly after the network is initialized. Consult the networking
 documentation for your Linux distribution for suggestions about the appropriate
 way to make `iptables` rules persistent.
 
-Through Docker startup, network creation, container creation, and other actions
-Docker dynamically manages `iptables` rules. In Docker 17.06 and higher, you can
-add rules to a new table called `DOCKER-USER` and Docker will not modify these
-rules when it starts or stops. This can be useful if you need to pre-populate
-`iptables` rules relating to containers or services that are not yet running.
+Docker dynamically manages `iptables` rules for the daemon, as well as your
+containers, services, and networks. In Docker 17.06 and higher, you can add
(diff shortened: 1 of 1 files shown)


engine/userguide/networking/index.md
```diff
@@ -550,6 +553,34 @@ in default `bridge` network and the
...
+Typically, `iptables` rules are created by an initialization script or a daemon
+process such as `firewalld`. The rules do not persist across a system reboot, so
+the script or utility must run when the system boots, typically at run-level 3
+or directly after the network is initialized. Consult the networking
+documentation for your Linux distribution for suggestions about the appropriate
+way to make `iptables` rules persistent.
+
+Through Docker startup, network creation, container creation, and other actions
+Docker dynamically manages `iptables` rules. In Docker 17.06 and higher, you can
+add rules to a new table called `DOCKER-USER` and Docker will not modify these
```
thaJeztah: Probably need a slight rewording; docker already wouldn't touch your rules, however, the problem was that it was not possible to reliably put your own rules _before_ Docker's rules. The new feature makes it so that Docker puts it's rules _after_ the rules defined in `DOCKER-USER`.

I'm bad at iptables, so may be good to have someone from networking confirm this.

If we don't have this written down yet (as a follow up), we may want to describe;

- `DOCKER` chain
- `DOCKER-INGRESS` chain
- `DOCKER-ISOLATION` chain
- that `--iptables=false` can disable iptables management (but IIRC, cannot be combined with swarm mode, which always requires some iptables rules to be managed)

mstanleyjones:
I took a stab at re-wording. Thanks for the clarification.

PR #3566, opened 2017-06-12
Fix command to download Docker Compose

ilyar:


mstanleyjones:
@shin- PTAL

shin-:
The current version is correct. This change is unnecessary and redundant with the paragraph above it which mentions using sudo if the user doesn't have write access to `/usr/local/bin`

PR #3579, opened 2017-06-12
fixed small typo: `docker machine` to `docker-machine`

lazarillo:
### Proposed changes

Fixed small typo in documentation: it gave the command `docker machine` instead of `docker-machine`.


johndmulhausen:
This is not correct; `docker-machine` is indeed a separate tool.

PR #3588, opened 2017-06-13
added explanation of Compose files and commands

### LINKED ISSUE #3486: Update Getting started Doc, rename docker-compose.yml to docker-stack.yml
[get-started/part3.md](https://docs.docker.com/get-started/part3/), CC @johndmulhausen 
The docker-compose.yml is usually used with docker-compose, so seeing this name used for docker stack confuses people.
Even I no more know what to think!
Is docker-compose project deprecated? or we can still use it with newest docker version?
can we use the example of docker-compose.yml file shown in this documentation with docker-compose or no?

londoncalling:
### What's changed

- Added an explanation of `docker-compose up` vs. `docker stack deploy` and Compose file naming conventions, per user feedback

- General copyedits

- Resized an image on part 5 (first snap of visualizer)

### Related 

Fixes #3486

### Reviewers

@johndmulhausen @mstanleyjones @joaofnfernandes @Lev09 @ManoMarks @shin- 

### Netlify preview link

https://deploy-preview-3588--docker-docs.netlify.com/get-started/part3/#docker-composeyml


get-started/part3.md
```diff
@@ -103,6 +103,19 @@ This `docker-compose.yml` file tells Docker to do the following:
 - Define the `webnet` network with the default settings (which is a
   load-balanced overlay network).
 
+Notice that we've set the Compose file version to version 3. This essentially
+makes it [swarm mode](/engine/swarm/index.md) compatible. We can make use of the
```
mstanleyjones: I'm not sure about this explanation. I mean, it's kind of true, but it's not complete. Maybe that's ok in this context. Like setting the version to 3 actually uses the version of the Compose API which includes the swarm mode stuff, right? The version specifies the version of the Compose API you can use.

londoncalling: Well, I think @johndmulhausen was trying to limit the amount of complexity/detail in the tutorial for a quick, task-based experience. I'm trying to walk a line here between giving more explanation with regard to the questions about the differences between `docker stack deploy` and `docker-compose up` and file names, without getting overly detailed. 

The `deploy` key with the ability to set a  swarm mode configuration was added in Compose v3, along with support for the `docker stack deploy`command. Taken together, this gives you the ability to create a swarm with a Compose file. I'm not sure that the fact that Compose v3 uses the swarm-enabled Compose API needs to be exposed here, as it's even another layer down, but I can add it if you think the current explanation is misleading in some way.

Any thoughts @ManoMarks ? @shin- @dnephin

joaofnfernandes: Totally agree in keeping complexity low, since this is supposed to be your first interaction with Docker, so at this point I know nothing about compose files, docker-compose, or docker stack deploy.

So this might not be the best place to explain me the difference between all these things, because the answer is going to be complex, and  distract me from deploying my first app.

londoncalling: Makes sense. I'll try to tighten it up a little, not add any more detail, and see what it looks like formatted as a tip.

londoncalling: Makes sense. I'll try to tighten it up a little, not add any more detail, and see what it looks like formatted as a tip.

Lev09: I Agree, this little explanation is less interesting for someone who is just starting to use docker, but for someone who had some experience with older docker versions this is going to be very helpful.

londoncalling:
Thanks all for the feedback, merging this.

Feel free to re-open the related comment if you want more changes with regard to this.

PR #3589, opened 2017-06-13
Add amazonec2 commands that were missing documentation

graingert:
### Proposed changes

Add amazonec2 commands that were missing documentation


graingert:
This also sorts the arguments for easy lookup (for users and when new commands are added)

mstanleyjones:
@kencochrane @FrenchBen is one of you the right one to PTAL?

kencochrane:
This is for machine, going to add @shin- as well.

PR #3590, opened 2017-06-13
Correct example script

kwaio:
### Proposed changes

Previous version could not work as grep -q does not print anything to stdout, only exits with a status code depending on a match was found or not. Hence the variables were always empty and could not be compared with 0

N/A only docs

### Related issues (optional)

N/A


mstanleyjones:
@johndmulhausen please merge this, it's fine.

PR #3615, opened 2017-06-14
Logging technical debt

### LINKED ISSUE #1520: Explain the SIZE column in "docker ps -s" and what "virtual" keyword means
[engine/reference/commandline/ps.md](https://docs.docker.com/engine/reference/commandline/ps/)

Under the SIZE column of the outpout of "docker ps -s". Is it possible to explain what the "virtual" keyword means ?

Example : for a container I am runing I can see 2.045 MB (virtual 540.5 MB)
Does this mean that the container takes 2.045 MB on disk, or 540.5 MB ?

### LINKED ISSUE #2630: Publish ports on "localhost:8080"
[engine/swarm/services.md](https://docs.docker.com/engine/swarm/services/), CC @mstanleyjones

I've been trying the "routing mesh", and under "Publish Ports" the docs say you can try `curl localhost:8080`.  

On my system at least, this just hangs, and I need to curl the network IP address.  I can't find much by googling if the routing mesh should work with localhost or not.  If this is something wrong with my system then apologies!

Containers: 3
 Running: 3
 Paused: 0
 Stopped: 0
Images: 1
Server Version: 17.03.1-ce
Storage Driver: overlay
 Backing Filesystem: xfs
 Supports d_type: true
Logging Driver: json-file
Cgroup Driver: cgroupfs
Plugins:
 Volume: local
 Network: bridge host macvlan null overlay
Swarm: active
 NodeID: csoxe3y8e8ym8hrils4obqs2q
 Is Manager: true
 ClusterID: piv6ehy5tfv3pn9zjw8mitf6m
 Managers: 1
 Nodes: 1
 Orchestration:
  Task History Retention Limit: 5
 Raft:
  Snapshot Interval: 10000
  Number of Old Snapshots to Retain: 0
  Heartbeat Tick: 1
  Election Tick: 3
 Dispatcher:
  Heartbeat Period: 5 seconds
 CA Configuration:
  Expiry Duration: 3 months
 Node Address: 172.28.128.29
 Manager Addresses:
  172.28.128.29:2377
Runtimes: runc
Default Runtime: runc
Init Binary: docker-init
containerd version: 4ab9917febca54791c5f071a9d1f404867857fcc
runc version: 54296cf40ad8143b62dbcaa1d90e520a2136ddfe
init version: 949e6fa
Security Options:
 seccomp
  Profile: default
Kernel Version: 3.10.0-514.el7.x86_64
Operating System: CentOS Linux 7 (Core)
OSType: linux
Architecture: x86_64
CPUs: 1
Total Memory: 992.7 MiB
Name: localhost.localdomain
ID: 7OJC:2HAR:AUFL:S2ZH:QFTU:OVMK:T37X:TPSH:Q2IA:UIKK:IH6J:DRKX
Docker Root Dir: /var/lib/docker
Debug Mode (client): false
Debug Mode (server): false
Registry: https://index.docker.io/v1/
WARNING: bridge-nf-call-ip6tables is disabled
Experimental: false
Insecure Registries:
 127.0.0.0/8
Live Restore Enabled: false

### LINKED ISSUE #3573: Update engine/admin/logging/syslog.md to show all the daemon.json options
The page talks about the flags you can use to configure Syslog when starting the Docker daemon, but it's not obvious which options are supported in the daemon.json file.

Relates to #3526

mstanleyjones:
### Proposed changes

This addresses several issues with `area/engine` and `logging`, though only one of them is actually logging!

- Updated logging driver docs from the code (https://github.com/moby/moby/tree/master/daemon/logger) (Fixes #3573)
- Clarify the routing mesh port is not always on localhost (Fixes #2630)
- Explain the difference between size and virtual size (Fixes #1520)

PTAL @kencochrane @thaJeztah

### DIFF
--- a/engine/admin/logging/awslogs.md
+++ b/engine/admin/logging/awslogs.md
@@ -14,10 +14,25 @@ and Command Line Tools](http://docs.aws.amazon.com/cli/latest/reference/logs/ind
 
 ## Usage
 
-You can configure the default logging driver by passing the `--log-driver`
-option to the Docker daemon:
+To use the `awslogs` driver as the default logging driver, set the `log-driver`
+and `log-opt` keys to appropriate values in the `daemon.json` file, which is
+located in `/etc/docker/` on Linux hosts or
+`C:\ProgramData\docker\config\daemon.json` on Windows Server. For more about
+configuring Docker using `daemon.json`, see
+[daemon.json](/engine/reference/commandline/dockerd.md#daemon-configuration-file).
+The following example sets the log driver to `awslogs` and sets the
+`awslogs-region` option.
 
-    dockerd --log-driver=awslogs
+```json
+{
+  "log-driver": "awslogs",
+  "log-opts": {
+    "awslogs-region": "us-east-1"
+  }
+}
+```
+
+Restart Docker for the changes to take effect.
 
 You can set the logging driver for a specific container by using the
 `--log-driver` option to `docker run`:
@@ -26,7 +41,9 @@ You can set the logging driver for a specific container by using the
 
 ## Amazon CloudWatch Logs options
 
-You can use the `--log-opt NAME=VALUE` flag to specify Amazon CloudWatch Logs logging driver options.
+You can add logging options to the `daemon.json` to set Docker-wide defaults,
+or use the `--log-opt NAME=VALUE` flag to specify Amazon CloudWatch Logs
+logging driver options when starting a container.
 
 ### awslogs-region
 
@@ -77,6 +94,43 @@ $ docker run --log-driver=awslogs \
 > **Note**:
 > Your AWS IAM policy must include the `logs:CreateLogGroup` permission before you attempt to use `awslogs-create-group`.
 
+### aws-datetime-format
+
+The `aws-datetime-format` option defines a multiline start pattern in [Python
+`strftime` format](http://strftime.org). One example of a use case for using
+this format is for parsing output such as a stack dump, which might otherwise
+be logged in multiple entries. The correct pattern allows it to be captured in a
+single entry.
+
+This option always takes precedence if both `awslogs-datetime-format` and
+`awslogs-multiline-pattern` are configured.
+
+For example:
+
+```bash
+$ docker run -it --rm \
+  --log-driver=awslogs \
+  --log-opt awslogs-group=test \
+  --log-opt awslogs-datetime-format='%Y-%m-%d' \
+  awslogtest:latest /test2.sh
+```
+
+### aws-multiline-pattern
+
+The `aws-multiline-pattern` option defines a multiline start pattern using a
+regular expression. This option is ignored if `awslogs-datetime-format` is also
+configured.
+
+For example:
+
+```bash
+$ docker run -it --rm \
+  --log-driver=awslogs \
+  --log-opt awslogs-group=test \
+  --log-opt awslogs-multiline-pattern='^ABCD' \
+  awslogtest:latest /test1.sh
+
+```
 
 ### tag
 
@@ -90,9 +144,9 @@ If not specified, the container ID is used as the log stream.
 {% raw %}
 > **Note**:
 > The CloudWatch log API doesn't support `:` in the log name. This can cause some issues when using the `{{ .ImageName }}` as a tag, since a docker image has a format of `IMAGE:TAG`, such as `alpine:latest`.
-> Template markup can be used to get the proper format. 
+> Template markup can be used to get the proper format.
 > To get the image name and the first 12 characters of the container id, you can use: `--log-opt tag='{{ with split .ImageName ":" }}{{join . "_"}}{{end}}-{{.ID}}'`
-> the output will be something like: `alpine_latest-bf0072049c76` 
+> the output will be something like: `alpine_latest-bf0072049c76`
 {% endraw %}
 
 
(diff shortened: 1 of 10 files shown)


engine/admin/logging/awslogs.md
```diff
@@ -14,10 +14,22 @@ and Command Line Tools](http://docs.aws.amazon.com/cli/latest/reference/logs/ind
 
 ## Usage
 
-You can configure the default logging driver by passing the `--log-driver`
-option to the Docker daemon:
+To use the `awslogs` driver as the default logging driver, set the `log-driver`
+and `log-opt` keys to appropriate values in the `daemon.json` file, which
+is located in `/etc/docker/` on Linux hosts or
+`C:\ProgramData\docker\config\daemon.json` on Windows Server. The following
+example sets the log driver to `awslogs` and sets the `awslogs-region` option.
```
thaJeztah: Perhaps we should link to the `daemon.json` reference in the `dockerd` CLI reference

engine/admin/logging/awslogs.md
```diff
@@ -77,6 +91,38 @@ $ docker run --log-driver=awslogs \
...
+  --log-opt awslogs-group=test \
+  --log-opt awslogs-datetime-format='%Y-%m-%d' \
+  awslogtest:latest /test2.sh
+```
+
+### aws-multiline-pattern
+
+The `aws-multiline-pattern` option defines a multiline start pattern using a
+regular expression. This option is ignored if `awslogs-datetime-format` is also
+configured.
```
thaJeztah: Maybe we should mention what this is useful for, or do you think it's clear from this? (usage can be an application that outputs, e.g. a stack-dump, which would otherwise be logged in multiple entries, and now allows you to capture it in a single log-entry. (could be better examples though)

engine/admin/logging/gcplogs.md
```diff
@@ -4,16 +4,28 @@ keywords: gcplogs, google, docker, logging, driver
...
 
 ## Usage
 
-You can configure the default logging driver by passing the `--log-driver`
-option to the Docker daemon:
+To use the `gcplogs` driver as the default logging driver, set the `log-driver`
+and `log-opt` keys to appropriate values in the `daemon.json` file, which
+is located in `/etc/docker/` on Linux hosts or
+`C:\ProgramData\docker\config\daemon.json` on Windows Server. The following
+example sets the log driver to `gcplogs` and sets the `gcp-meta-name` option.
```
thaJeztah: same here (link)

engine/admin/logging/gelf.md
```diff
@@ -52,15 +69,15 @@ $ docker run \
...
+| `gelf-address`           | The address of the GELF server. `udp` is the only supported URI specifier and you must specify the port.                                                                                                                                                        | `--log-opt gelf-address=udp://192.168.0.42:12201` |
+| `gelf-compression-type`  | The type of compression the GELF driver uses to compress each log message. Allowed values are `gzip`, `zlib` and `none`. The default is `gzip`.                                                                                                                 | `--log-opt gelf-compression-type=gzip`            |
+| `gelf-compression-level` | The level of compression when `gzip` or `zlib` is the `gelf-compression-type`. An integer in the range of `-1` to `9` (BestCompression). Default value is 1 (BestSpeed). Higher levels provide more compression at lower speed.                                 | `--log-opt gelf-compression-level=2`              |
```
thaJeztah: `-1` or `0` is "disable compression" (but not documented here); I  know some people ran into issues with compression (used a lot of resources on high-traffic logging), so may be worth mentioning somewhere

engine/swarm/services.md
```diff
@@ -359,8 +359,12 @@ $ docker service create --name my_web \
...
 Three tasks will run on up to three nodes. You don't need to know which nodes
 are running the tasks; connecting to port 8080 on **any** of the 10 nodes will
-connect you to one of the three `nginx` tasks. You can test this using `curl`
-(the HTML output is truncated):
+connect you to one of the three `nginx` tasks. You can test this using `curl`.
+The following example assumes that `localhost` is one of the swarm nodes. If
+this is not the case, or `localhost` does not resolve to an IP address on your
+host, substitute the host's IP address or resolvable host name.
+
+The HTML output is truncated:
```
thaJeztah: Looks like this change is not related?

engine/userguide/storagedriver/imagesandcontainers.md
```diff
@@ -72,6 +72,39 @@ writable container layer. Each storage driver handles the implementation
...
+disk space:
+
+- Disk space used for log files if you use the `json-file` logging driver. This
+  can be non-trivial if your container generates a large amount of logging data
+  and log rotation is not configured.
+- Volumes and bind mounts used by the container.
+- Disk space used for the container's configuration files, which are typically
+  small.
+- Memory written to disk (if swapping is enabled).
+- Checkpoints, if you're using the experimental checkpoint/restore feature.
```
thaJeztah: Should these changes be in this PR?

kencochrane:
Looks good to me, I had a few of the same comments/suggestions as @thaJeztah so when his comments are addressed, I should be good as well. 

Thanks for cleaning them up.

engine/swarm/services.md
mstanleyjones: Yeah I explain it in the first comment that I just searched for all things that matched "logging" and for some reason those came up so I am fixing them here too. I'll rebase instead of squashing so they will show up as separate commits.

engine/admin/logging/awslogs.md
mstanleyjones: Done.

PR #3617, opened 2017-06-14
Update curl examples to use 'localhost'

### LINKED ISSUE #3612: Update curl examples to reference localhost instead of 0.0.0.0
[engine/swarm/secrets.md](https://docs.docker.com/engine/swarm/secrets/), CC @mstanleyjones 

In working through the examples on the referenced docs, the curl commands given produce errors. One specific example:

```
ubuntu@test-0:~$curl --cacert root-ca.crt https://0.0.0.0:3000/
curl: (51) SSL: certificate subject name (localhost) does not match target host name '0.0.0.0'
ubuntu@test-0:~$
```

This is because the certificate is explicitly created with the common name "localhost" in a previous step.

The correct curl invocations use "localhost" in place of "0.0.0.0". For example

```ubuntu@test-0:~$ curl --cacert root-ca.crt https://localhost:3000/
<!DOCTYPE html>
<html>
<head>
<title>Welcome to nginx!</title>
...```

mstanleyjones:
### Proposed changes

Update curl examples to use `localhost` instead of `0.0.0.0`
Fixes #3612 

@aaronlehmann PTAL

### DIFF
--- a/engine/swarm/secrets.md
+++ b/engine/swarm/secrets.md
@@ -411,7 +411,7 @@ generate the site key and certificate, name the files `site.key` and
     server, and that the correct TLS certificate is being used.
 
     ```bash
-    $ curl --cacert root-ca.crt https://0.0.0.0:3000
+    $ curl --cacert root-ca.crt https://localhost:3000
 
     <!DOCTYPE html>
     <html>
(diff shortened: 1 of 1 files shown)


aaronlehmann:
looks good to me

PR #3644, opened 2017-06-16
Add release notes from upstream repos into the formal docs

mstanleyjones:
### Proposed changes

Previously we did not have release notes for Engine, Compose, Machine or standalone Swarm in our docs at all. We directed people to the upstream projects. This fixes that, using the `CHANGELOG.md` files from those upstream repos as a basis. This also adds a release notes "landing page" which is visible from the **Guides** and **Product Manuals** tabs.

A second commit fixes some unescaped Liquid errors that have crept in.

### Implications

When a Docker CE, Compose, Machine, or Swarm release happens, someone on one of those engineering teams should copy-paste the relevant portion of their `CHANGELOG.md` into the appropriate file in this repo, or raise a PR for a docs person to do it.


_data/toc.yaml
```diff
@@ -157,6 +155,10 @@ guides:
       title: Manage data in containers
   - path: /engine/docker-overview/
     title: Docker overview
+  - path: /release-notes/docker-ce/
```
johndmulhausen: Please don't put Release Notes under "Getting Started"

Why not just make "Release Notes" a node under Guides and have these be sub-topics?

mstanleyjones: Because some of them are already in other places, I was trying to replicate the "other places" idea, but we don't have a "Docker Engine" specific area anymore. I need to put these stubs somewhere or they won't have a left-hand nav at all.

joaofnfernandes:
@dmp42 this seems to be aligned with the process for generating release notes that you were suggesting. The release notes go on a centralized place, and we'll have entries for things that are specific to an edition. E.g. When we release a new patch, you'll have entries for the builder, client ..., but you'll also have entries for Docker for Mac, Windows, AWS, if we made changes that are specific to that edition.

My understanding of your proposal is that we'll manage the change log in a centralized repo (eventually). So while this PR moves the ball forward, we can later use submodules or something else to fetch the release notes from that upstream repo. Editions with UI like Docker for Mac, will also be able to leverage that upstream repo for the release notes they display to users.

_data/toc.yaml
joaofnfernandes: Yup. I only care about release notes if I'm already using the product, not when I'm getting started.

mstanleyjones: I really think people want it during the installation instructions.

joaofnfernandes: Maybe during an upgrade, but not when they're installing for the first time. 

Can you tell me about a time when you needed to read the release notes before understanding what a product does and what problems it solves?

I always go back to this analogy. You have a new fancy car that you believe it's better than any other car out there. Are you going to take me for a test drive and **show me**, or are you going to tell me the story of how you just introduced seats in the latest release, and a steering wheel two versions ago? Do I even care about that story if I'm just learning about your car for the first time?

mstanleyjones: If I have heard of a security problem or I'm concerned that there might be a security bug that I'm exposed to when I install the very first time. As an ops person, I may have a checklist of things that must work or must not be present before installing the first time.

mstanleyjones:
@johndmulhausen PTAL. Now the full release notes are in Product Manuals and the CE and EE release notes are at the bottom of "Get Docker". I kept the landing page so that people who had bookmarked `/release-notes/` would have a place to land. I didn't think it was appropriate to send them to either the CE or EE release notes.

PR #3668, opened 2017-06-20
Fix name of syslog-tls-skip-verify log driver option

philandstuff:
### Proposed changes

The name of this option was previously given as `syslog-tls-skip`, but
the example used `syslog-tls-skip-verify`.  The latter is correct.
This commit fixes the typo.


philandstuff:
Looks like this was independently fixed in baaead7cdd1777f41fe8061bcf9c47badb4637a3

PR #3678, opened 2017-06-21
Add information about userns-remap

### LINKED ISSUE #1366: Improve docker daemon user namespace documentation
### Problem description

The docker daemon configuration instructions  are lacking information on configuring user namespaces. The user namespace section in the reference docs is also getting quite long, so it has been suggested by @thaJeztah to migrate them to the another part of the documentation.

### Problem location

@mstanleyjones suggested adding to the docker daemon admin guide: https://docs.docker.com/engine/admin/

User namespaces were introduced to Docker 1.10 from the experimental build.

### Suggestions for a fix

Copy/Rework some of the namespace documentation from the reference docs  (https://docs.docker.com/engine/reference/commandline/dockerd/) and write an example on using user namespaces.

mstanleyjones:
Add details about `userns-remap`

Fixes #1366

@estesp PTAL

### DIFF
--- a/engine/admin/index.md
+++ b/engine/admin/index.md
@@ -75,6 +75,133 @@ restart Docker. This method works for every Docker platform. The following
 }
 ```
 
+## Isolate containers within a user namespace
+
+Linux namespaces provide isolation for running processes, limiting their access
+to system resources without the running process being aware of the limitations.
+For more information on Linux namespaces, see
+[Linux namespaces](https://www.linux.com/news/understanding-and-securing-linux-namespaces){: target="_blank" class="_" };
+
+The best practice for preventing privilege-escalation attacks from within a
+container is to configure your container's applications to run as unprivileged
+users. For containers whose processes must run as the `root` user within the
+container, you can re-map this user to a less-privileged user on the Docker
+host.
+
+This re-mapping is relatively transparent to the container, but introduces some
+configuration complexity when the container needs access to privileged resources
+on the Docker host, such as bind mounts into areas of the filesystem that the
+system user cannot write to.
+
+### Prerequisites
+
+1.  The username or user ID (and optionally group name or group ID) must exist
+    already. Typically, this means that the relevant entries need to be in
+    `/etc/password` and `/etc/group`, but if you are using a different
+    authentication back-end, this requirement may translate differently.
+
+    To verify this, use the `id` command:
+
+    ```bash
+    $ id testuser
+    
+    uid=1001(testuser) gid=1001(testuser) groups=1001(testuser)
+    ```
+
+2.  If there are any locations on the Docker host where the unprivileged
+    user needs to write, adjust the permissions of those locations
+    accordingly.
+
+3.  The way the namespace remapping is handled on the host is using two files,
+    `/etc/subuid` and `/etc/subgid`. These files are typically managed
+    automatically when you add or remove users or groups, but on a few
+    distributions such as RHEL and CentOS 7.3, you may need to manage these
+    files manually.
+
+    Each file contains three fields: the username or ID of the user, followed by
+    a beginning UID or GID (which is treated as UID or GID 0 within the namespace)
+    and a maxumum number of UIDs or GIDs available to the user. For instance,
+    given the following entry:
+
+    ```none
+    testuser:231072:65536
+    ```
+   
+    This means that user-namespaced processes started by `testuser` will be owne
+    by host UID `231072` (which will look like UID `0` inside the namespace)
+    through 296608 (231072 + 65536). These ranges should not overlap, to ensure
+    that namespaced processes cannot access each other's namespaces.
+
+    After adding your user, check `/etc/subuid` and `/etc/subgid` to see if your
+    user has an entry in each. If not, you need to add it, being careful to
+    avoid overlap.
+
+4.  Enabling `userns-remap` will effectively mask existing image and container
+    layers, as well as other Docker objects within `/var/lib/docker/`. This is
+    because Docker needs to adjust the ownership of these resources and actually
+    stores them in a subdirectory within `/var/lib/docker/`. It is best to enable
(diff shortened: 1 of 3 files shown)


mstanleyjones:
@pvnovarese PTAL

engine/admin/index.md
```diff
@@ -75,6 +75,133 @@ restart Docker. This method works for every Docker platform. The following
...
+    Each file contains three fields: the username or ID of the user, followed by
+    a beginning UID or GID (which is treated as UID or GID 0 within the namespace)
+    and a maxumum number of UIDs or GIDs available to the user. For instance,
+    given the following entry:
+
+    ```none
+    testuser:231072:65536
+    ```
+   
+    This means that user-namespaced processes started by `testuser` will be owne
```
thaJeztah: s/owne/owned/

engine/admin/index.md
```diff
@@ -75,6 +75,133 @@ restart Docker. This method works for every Docker platform. The following
 }
 ```
 
+## Isolate containers within a user namespace
```
thaJeztah: Wondering if;

- this topic should be in the "security" section
- it should be on its own page

If there's a dedicated page for this, we can copy some more of the technical information from https://docs.docker.com/engine/reference/commandline/dockerd/#daemon-user-namespace-options to the page, and remove it from the `dockerd` reference.

The reference docs contain quite some information that's important when using this feature (limitations, how to start a privileged container if userns is enabled, and the current limitation of having a single "global" mapping, instead of custom mappings per container)

If we migrate the content here, we can keep a short section in the daemon docs, but refer to the extended documentation elsewhere

engine/admin/index.md
```diff
@@ -75,6 +75,133 @@ restart Docker. This method works for every Docker platform. The following
...
+[Linux namespaces](https://www.linux.com/news/understanding-and-securing-linux-namespaces){: target="_blank" class="_" };
+
+The best practice for preventing privilege-escalation attacks from within a
+container is to configure your container's applications to run as unprivileged
+users. For containers whose processes must run as the `root` user within the
+container, you can re-map this user to a less-privileged user on the Docker
+host.
+
+This re-mapping is relatively transparent to the container, but introduces some
+configuration complexity when the container needs access to privileged resources
```
thaJeztah: we can remove "privileged" here; also files owned by "regular" users will be inaccessible.

The goal of user-namespaces is to give the user inside the container no access to _anything_ on the host (i.e., if the process _would_ be able to escape the container, it's basically harmless, as it won't be able to do anything).

mstanleyjones:
Addressed great feedback, split this off into its own topic and made some "What next" links in the configuration index, one of them pointing to the security landing page.

mstanleyjones:
Please review at https://deploy-preview-3678--docker-docs.netlify.com/engine/security/userns-remap/

engine/security/userns-remap.md
```diff
@@ -0,0 +1,249 @@
...
+
+This means that `testuser` is assigned a subordinate user ID range of `230172`
+and the next 65536 integers in sequence. UID `231072` is mapped within the
+namespace (within the container, in this case) as UID `0` (`root`). UID `231073`
+is mapped as UID `1`, and so forth. If a process attempts to escalate privilege
+outside of the namespace, the process is running as an unprivileged high-number
+UID on the host, which does not even map to a real user. This means the process
+has no privileges on the host system at all.
+
+> Multiple ranges
```
pvnovarese: should we have an example here?  It's not clear what happens if you have multiple ranges, what the expected behavior should be, or if there is any other user action required to make use of them.

engine/security/userns-remap.md
```diff
@@ -0,0 +1,249 @@
...
+>
+> It is possible to assign multiple subordinate ranges for a given user or group
+> by adding multiple non-overlapping mappings for the same user or group in the
+> `/etc/subuid` or `/etc/subgid` file. In this case, Docker uses only the first
+> five mappings, in accordance with the kernel's limitation of only five entries
+> in `/proc/self/uid_map` and `/proc/self/gid_map`.
```
mstanleyjones: I need more info to work up these examples, I think. @estesp any thoughts?

engine/security/userns-remap.md
```diff
@@ -0,0 +1,249 @@
...
+
+This re-mapping is transparent to the container, but introduces some
+configuration complexity in situations where the container needs access to
+resources on the Docker host, such as bind mounts into areas of the filesystem
+that the system user cannot write to. From a security standpoint, it is best to
+avoid these situations.
+
+## Prerequisites
+
+1.  If you want to use an existing username or user ID, it must already exist.
```
mstanleyjones: OK, I just pushed an amendment that I think addresses this.

engine/security/userns-remap.md
```diff
@@ -0,0 +1,249 @@
...
+
+> **Warning**: Some distributions, such as RHEL and CentOS 7.3, do not
+> automatically add the new group to the `/etc/subuid` and `/etc/subgid` files.
+> You are responsible for editing these files and assigning non-overlapping
+> ranges, in this case. This step is covered in [Prerequisites](#prerequisites).
+{: .warning-vanila }
+
+It is very important that the ranges not overlap, so that a process cannot gain
+access in a different namespace. On most Linux distributions, system utilities
+manage the ranges for you when you add or remove users.
```
mstanleyjones: Talked to @pvnovarese offline and agreed that we can expand the docs on multiple ranges later. I filed #3704 for this.

estesp:
Thanks @mstanleyjones! We can chat about multiple ranges.. its a fairly advanced topic with probably a very small set of potential users.

fabiomartinelli:
I'm one of those potential users, it would be great to get more info and further tested examples about the multiple ranges topic.

mstanleyjones:
@fabiomartinelli can you tell us more about your use  case for using multiple ranges?

PR #3701, opened 2017-06-26
Fix typo in docker-machine completion setup for zsh

ShockN745:
The file that was downloaded was actually the autocompletion for `docker` and **not** `docker-machine`

I tested the "fix" on my local installation: It's working ;)


gabbah:
Under the Mac instructions, the link https://raw.githubusercontent.com/docker/docker-ce/master/components/engine/contrib/completion/bash/docker still returns 404, so anyone using that command will get a file with the contents "404 Not found" in it, instead of a working completion file.
Line 24 and 30.

PR #3702, opened 2017-06-26
Update install to ensure extras repo is enabled

cpuguy83:
The `extras` repos are required for 17.06 which depends on the
`container-selinux` package.

CentOS typically has the repo enabled by default, but in some cases may
be disabled, and RHEL never has it enabled by default.

This is provided by the main Fedora repos and does not require any extra
setting.

ping @mstanleyjones


rogaha:
thanks for putting this together @cpuguy83! /cc @andrewhsu

thaJeztah:
does this resolve https://github.com/docker/for-linux/issues/21 ?

mstanleyjones:
Yep, good catch!

PR #3705, opened 2017-06-26
Final tidying up before 17.06 release

mstanleyjones:
Some housekeeping, and also adding info about the new `testing` repository.

cc/ @andrewhsu


mstanleyjones:
There's one commit coming still. I need to update the upstream references to pull from `docker/docker-ce`.

engine/installation/linux/docker-ce/centos.md
```diff
@@ -77,31 +77,38 @@ from the repository.
     ```
 
 2.  Use the following command to set up the **stable** repository. You always
-    need the **stable** repository, even if you want to install **edge** builds
-    as well.
+    need the **stable** repository, even if you want to install builds from the
+    **edge** or **testing** repositories as well.
```
thaJeztah: I think with the upcoming changes, there's no longer a need to have them both (@andrewhsu?) as stable releases will be pushed to all of _edge_, _stable_, and _testing_

mstanleyjones: Do we really want people to only have Edge and not Stable at all? I think maybe we should think it through and  change it after release.

mstanleyjones:
OK, the upstream fetch script has been updated and the 1.30 API has been added to the TOC. I think this is ready. You'll want to build it locally with `docker build` and test the local build to get a true idea. It looks good to me. I did file https://github.com/docker/cli/pull/241, and we will get some spurious Liquid errors until it's cherry-picked into 17.06.

engine/installation/linux/docker-ce/centos.md
thaJeztah: With "stable" builds now also being pushed to the "edge" channel, basically;

- "stable" only gives you stable releases
- "edge" gives you both "stable" and "edge" releases
- "test" gives you "stable", "edge" and "release-candidates"

All channels should give you the latest version by default (unless "pinned" to a specific version)

We can tweak after release though

PR #3708, opened 2017-06-26
Update documentation for Cloudstor in AWS/Azure for 17.06

ddebroy:
### Proposed changes

Updated persistent volume documentation for AWS/Azure. Mention several examples on how to use cloudstor volumes with local/shared backing in AWS.

17.06

### Related issues (optional)

@friism @kencochrane @FrenchBen

### DIFF
--- a/docker-for-aws/persistent-data-volumes.md
+++ b/docker-for-aws/persistent-data-volumes.md
@@ -8,7 +8,17 @@ title: Docker for AWS persistent data volumes
 
 ## What is Cloudstor?
 
-Cloudstor is a volume plugin managed by Docker. It comes pre-installed and pre-configured in swarms deployed on Docker for AWS. Swarm tasks use a volume created through Cloudstor to mount a persistent data volume that stays attached to the swarm tasks no matter which swarm node they get scheduled or migrated to. Cloudstor relies on shared storage infrastructure provided by AWS to allow swarm tasks to create/mount their persistent volumes on any node in the swarm. In a future release we will introduce support for direct attached storage to satisfy very low latency/high IOPs requirements.
+Cloudstor is a modern volume plugin managed by Docker. It comes pre-installed and pre-configured in Docker Swarms deployed through Docker for AWS. Docker Swarm tasks can use a volume created through Cloudstor to mount a persistent data volume. In Docker for AWS, Cloudstor can be used with two `backing` options: `local` (that uses EBS) and `shared` (that uses EFS) described below.
+
+## Local Cloudstor volumes
+
+ Workloads running in a Docker Service that require access to low latency/high IOPs persistent storage (e.g. a database engine) can use a `local` Cloudstor volume backed by EBS. It is possible to configure the specific type of EBS volume supported by AWS (e.g. gp2, st1, sc1) that the workload requires during volume creation. If the swarm task gets rescheduled to another node within the same availability zone (as the node where the task was running on), Cloudstor takes care of detaching and re-attaching the EBS to the appropriate node. If the swarm task gets rescheduled to a node in a different availability zone (from the node where the task was originally running), Cloudstor will transfer the contents of the EBS volume using a snapshot to the destination availability zone as well as clean up the EBS volume in the original Availability Zone. To minimize the time necessary to create the snapshot to transfer data across availability zones, Cloudstor periodically takes snapshots of the EBS volumes to ensure there is never a large amount of diff that will need to get transferred as part of the snapshot necessary during a cross availability zone task reschedule. Typically the snapshot based transfer process across availability zones takes a few minutes and for extremely write-heavy workloads, this may take several minutes. 
+ 
+ Sharing/mounting the same Cloudstor volume backed by EBS among multiple tasks is not a supported scenario and will lead to data loss. If you need a Cloudstor volume to share data between tasks please read below for EFS backed options. `local` Cloudstor backed by EBS is supported on all AWS regions. It is the default "backing" option if EFS support is not selected during setup/installation or if EFS is not supported in a region.
+
+## Shared Cloudstor volumes 
+
+Workloads running in multiple swarm tasks that need to share data in persistent storage volumes can use a Cloudstor volume backed by EFS. Such a volume and it's contents can be mounted by multiple tasks on multiple swarm nodes simultaneously since EFS makes the data available to all Swarm nodes over NFS. `shared` Cloudstor volumes will only work in those AWS regions where EFS is supported (e.g. Ohio, Oregon, N.Virginia, Sydney). If EFS Support is selected during setup/installation, the default "backing" option for Cloudstor volumes is set to Shared so that EFS is used by default. Please note that `shared` Cloudstor volumes backed by EFS (or even EFS MaxIO) may not be ideal for workloads that have very low latency and high IOPSs requirements. For details, please refer to: http://docs.aws.amazon.com/efs/latest/ug/performance.html
 
 ## Use Cloudstor
 
@@ -22,7 +32,7 @@ f416c95c0dcc        cloudstor:aws               cloud storage plugin for Docker 
 
 The following examples show how to create swarm services that require data persistence using the --mount flag and specifying Cloudstor as the driver.
 
-### Share the same volume between tasks:
+### Share the same volume between tasks (EFS support required and enabled):
 
 ```bash
 docker service create --replicas 5 --name ping1 \
@@ -30,11 +40,21 @@ docker service create --replicas 5 --name ping1 \
     alpine ping docker.com
 ```
 
-Here all replicas/tasks of the service `ping1` share the same persistent volume `sharedvol1` mounted at `/shareddata` path within the container. Docker Swarm takes care of interacting with the Cloudstor plugin to make sure the common backing store is mounted on all nodes in the swarm where tasks of the service are scheduled. Each task needs to make sure they don't write concurrently on the same file at the same time and cause corruptions since the volume is shared.
+The above is supported only in those regions where EFS is supported and EFS support is enabled during deployment of the Cloud Formation template. All replicas/tasks of the service `ping1` share the same persistent volume `sharedvol1` mounted at `/shareddata` path within the container. Docker Swarm takes care of interacting with the Cloudstor plugin to make sure EFS is mounted on all nodes in the swarm where tasks of the service are scheduled. Each task needs to make sure they don't write concurrently on the same file at the same time and cause corruptions since the volume is shared.
 
-With the above example, you can make sure that the volume is indeed shared by logging into one of the containers in one swarm node, writing to a file under `/shareddata/` and reading the file under `/shareddata/` from another container (in the same node or a different node).
+In the above example, you can make sure that the volume is indeed shared by logging into one of the containers in one swarm node, writing to a file under `/shareddata/` and reading the file under `/shareddata/` from another container (in a different node).
 
-### Use a unique volume per task:
+If high IO throughput is desired, set the perfmode parameter as volume-opt to maxio:
+
+```bash
+{% raw %}
+docker service create --replicas 5 --name ping3 \
+    --mount type=volume,volume-driver=docker4x/cloudstor:aws-v{{ edition_version }},source={{.Service.Name}}-{{.Task.Slot}}-vol5,destination=/mydata,volume-opt=perfmode=maxio \
+    alpine ping docker.com
+{% endraw %}
+```
+
+### Use a unique volume per task (using EFS with EFS support present and enabled):
 
 ```bash
 {% raw %}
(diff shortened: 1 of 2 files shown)


docker-for-aws/persistent-data-volumes.md
```diff
@@ -8,7 +8,17 @@ title: Docker for AWS persistent data volumes
 
 ## What is Cloudstor?
 
-Cloudstor is a volume plugin managed by Docker. It comes pre-installed and pre-configured in swarms deployed on Docker for AWS. Swarm tasks use a volume created through Cloudstor to mount a persistent data volume that stays attached to the swarm tasks no matter which swarm node they get scheduled or migrated to. Cloudstor relies on shared storage infrastructure provided by AWS to allow swarm tasks to create/mount their persistent volumes on any node in the swarm. In a future release we will introduce support for direct attached storage to satisfy very low latency/high IOPs requirements.
+Cloudstor is a modern volume plugin managed by Docker. It comes pre-installed and pre-configured in Docker Swarms deployed through Docker for AWS. Docker Swarm tasks as well as regular containers can use a volume created through Cloudstor to mount a persistent data volume. In Docker for AWS, Cloudstor can be used with two `backing` options: `local` (that uses EBS) and `shared` (that uses EFS) described below.
```
friism: "managed" > "built"

friism: "through" > "with"

friism: "can be used with" > "has"

friism: @ddebroy is there well-defined nomenclature for "local" and "shared"? I'm worried about "local" in particular since people might confuse with volumes on the instance itself. I think the word Goelzer uses is "relocatable" and Andrea used "attachable" or something.

docker-for-aws/persistent-data-volumes.md
```diff
@@ -8,7 +8,17 @@ title: Docker for AWS persistent data volumes
...
...ng volume creation. Each `local` Cloudstor volume is backed by a single EBS volume. If a swarm task using a `local` Cloudstor volume gets rescheduled to another node within the same availability zone (as the node where the task was running on), Cloudstor takes care of detaching and re-attaching the backing EBS volume to the target node. If the swarm task gets rescheduled to a node in a different availability zone (from the node where the task was originally running), Cloudstor will transfer the contents of the backing EBS volume using a snapshot to the destination availability zone as well as clean up the EBS volume in the original Availability Zone. To minimize the time necessary to create the snapshot to transfer data across availability zones, Cloudstor periodically takes snapshots of the EBS volumes to ensure there is never a large amount of diff that will need to get transferred as part of the snapshot necessary during a cross availability zone task reschedule. Typically the snapshot based transfer process across availability zones takes a few minutes and for extremely write-heavy workloads (e.g. writing > 20 GB of fresh/new data within minutes), this may take several minutes.
```
friism: so in both cases it takes a "few minutes" or "several minutes"

friism: @ddebroy if you think it's a good idea it might be useful to describe how to constrain a service to a particular AZ to avoid having to incur the overhead of shuttling snapshots

docker-for-aws/persistent-data-volumes.md
```diff
@@ -8,7 +8,17 @@ title: Docker for AWS persistent data volumes
...
+ 
+ Sharing/mounting the same Cloudstor volume backed by EBS among multiple tasks is not a supported scenario and will lead to data loss. If you need a Cloudstor volume to share data between tasks please read below for EFS backed `shared` volume options. `local` Cloudstor backed by EBS is supported on all AWS regions that support EBS. It is the default "backing" option if EFS support is not selected during setup/installation or if EFS is not supported in a region.
```
friism: what happens if I specify "shared" in a region without EFS support? Am I handed an EBS-based volume? That's probably not good...

docker-for-aws/persistent-data-volumes.md
```diff
@@ -22,44 +32,86 @@ f416c95c0dcc        cloudstor:aws               cloud storage plugin for Docker
...
+```
 
-You can use `docker volume ls` to enumerate all volumes created on a node including those backed by Cloudstor. Note that if a swarm service task starts off in a node and has a Cloudstor volume associated and later gets rescheduled to a different node, `docker volume ls` in the initial node will continue to list the Cloudstor volume that was created for the task that no longer executes on the node although the volume is mounted elsewhere. Do NOT prune/rm the volumes that gets enumerated on a node without any tasks associated since these actions will result in data loss if the same volume is mounted in another node (i.e. the volume shows up in the `docker volume ls` output on another node in the swarm). We can try to detect this and block/handle in post-Beta.
+### Use a unique volume per task (using EBS):
 
-### Configure IO performance
+A unique `local` Cloudstor volume backed by a specified type of EBS can be created and mounted for each task in a swarm service using a templatized notation with the `docker service create` CLI. Creation of new EBS volumes typically takes a few minutes.  The volume options (besides `backing=local` indicating `local` EBS backed volumes) are:
```
friism: double space

friism: what does the container see while the ebs volume is provisioning? or is the task not scheduled before the volume is ready? (if that's the case, maybe make a note of that)

docker-for-aws/persistent-data-volumes.md
```diff
@@ -22,44 +32,86 @@ f416c95c0dcc        cloudstor:aws               cloud storage plugin for Docker
...
 
-### List or remove volumes created by Cloudstor
+```bash
+{% raw %}
+docker volume create -d "cloudstor:aws" --opt backing=shared mysharedvol1
+{% endraw %}
+```
 
-You can use `docker volume ls` to enumerate all volumes created on a node including those backed by Cloudstor. Note that if a swarm service task starts off in a node and has a Cloudstor volume associated and later gets rescheduled to a different node, `docker volume ls` in the initial node will continue to list the Cloudstor volume that was created for the task that no longer executes on the node although the volume is mounted elsewhere. Do NOT prune/rm the volumes that gets enumerated on a node without any tasks associated since these actions will result in data loss if the same volume is mounted in another node (i.e. the volume shows up in the `docker volume ls` output on another node in the swarm). We can try to detect this and block/handle in post-Beta.
+### Use a unique volume per task (using EBS):
```
friism: oh, so there's actually kinda three modes: shared, local (single) and local (for each task). Not  critical, but it might be good to very briefly mention that in the intro blurb at the beginning of the document

docker-for-aws/persistent-data-volumes.md
```diff
@@ -22,44 +32,86 @@ f416c95c0dcc        cloudstor:aws               cloud storage plugin for Docker
...
+{% raw %}
+docker volume create -d "cloudstor:aws" --opt ebstype=io1 --opt size=25 --opt iops=1000 --opt backing=local mylocalvol1
+{% endraw %}
+```
+
+Sharing the same `local` Cloudstor volume across multiple tasks of a service or across multiple independent containers is not supported when `backing=local` is specified. Trying to do so in containers across multiple nodes will result in IO errors.
+
+### Use a unique volume per task (using EFS with EFS support present and enabled):
+
+It is possible to use the templatized notation to indicate to Docker Swarm that a unique EFS backed volume be created and mounted for each replica/task of a service. Example:
```
friism: maybe mention why you would do this. (I can't come up with a reason off the top of my hat)

friism: oh, to avoid the overhead of snapshot-shuffles?

mstanleyjones:
Please ping after you have addressed the technical feedback, because I'd like to give this a copy-edit pass.

docker-for-aws/persistent-data-volumes.md
ddebroy: @friism do you think it will be better/clearer to just expose the options for `backing` as `ebs` and `efs` ? That probably clarifies things a lot better instead of `local`/`shared` and having to explain `local` Cloudstor volume backed by EBS and `shared` Cloudstor volume backed by EFS elsewhere in the docs.

friism: If you think it's responsible I'm very +1 on coming up with _our_ nomenclature that describes the semantics and _not_ the implementation. This makes this stuff way more portable, eg. I can deploy a docker-compose file (referencing a "shared" volume) reliably on AWS and Azure (I know there are a lot of pitfalls).

docker-for-aws/persistent-data-volumes.md
ddebroy: @friism the observation was it takes about 2 minutes base for the cross az rescheduling no matter the amount of diffs. Beyond 20 GB of diffs, the snapshot process seems to take greater than 10 minutes and rises for additionals GBs.

friism: cool, I'm not sure that's clear from the current "few minutes" vs. "several minutes" terminology

docker-for-aws/persistent-data-volumes.md
ddebroy: I agree with your point and that's the reason we have various defaults in place so that commands may work across clouds with minimal changes. 

One additional point to consider is if we do want to implement s3 support down the line, we will probably need something other than `shared` to differentiate EFS from s3 although s3 backing will offer `shared` ish semantics.

One possible set of names could be `direct-attached`/`network-attached` which aligns with the usual storage terminology to distinguish SCSI/ATA from NFS/SMB. For s3 we can probably make something up along the lines of `object-attached`. Thoughts?

docker-for-aws/persistent-data-volumes.md
ddebroy: ok will clarify with the above details.

docker-for-aws/persistent-data-volumes.md
ddebroy: You should get an error if you specify `shared` in a region without EFS support. If you don't specify a backing type, that's when you let cloudstor choose for you - you get EFS if EFS is supported and enabled otherwise EBS.

docker-for-aws/persistent-data-volumes.md
ddebroy: sounds good.

docker-for-aws/persistent-data-volumes.md
ddebroy: Task is not scheduled.

docker-for-aws/persistent-data-volumes.md
friism: ah, sorry - I get it know

docker-for-aws/persistent-data-volumes.md
friism: great

docker-for-aws/persistent-data-volumes.md
ddebroy: We specify that users should use EBS only when necessary - specific IOPs/latency requirements that EFS cannot deliver. So if the user has a scenario where the tasks in a service write to say, the same files under the same paths but they don't have stringent IOPs/latency requirements of EBS, they should use this "per-task" EFS option. It also saves on snapshot overhead during reschedules.

friism: I see - just putting it out there, but people might want EBS for this since it's a lot cheaper.

kencochrane:
@ddebroy looks good. You mention it briefly, but we should probably describe how to clean up volumes and snapshots that are left over, and maybe call it out a little more. Because they are created outside of CloudFormation, people will need to manually remove these before they can delete their stack. We should mention that, so that we can point people to the docs when it happens and they email us.

ddebroy:
Addressed review comments so far above. I have added a note on how to get the list of EBS created by the stack that one can manually delete. So far, I have found that the stack deletion goes through fine even with EBS volumes from cloudstor lying around but those volumes just never get deleted.

ddebroy:
@mstanleyjones this is now ready as far as internal tech reviews are concerned.

mstanleyjones:
Thanks, I'm doing copy-editing now. I'll have it finished by GA or very shortly after.

docker-for-aws/persistent-data-volumes.md
```diff
@@ -22,44 +36,88 @@ f416c95c0dcc        cloudstor:aws               cloud storage plugin for Docker
...
 
-### Use a unique volume per task:
+The only option available for EFS currently is `perfmode`. If high IO throughput is desired, please set the `perfmode` volume-opt to `maxio`. Example:
 
 ```bash
 {% raw %}
-docker service create --replicas 5 --name ping2 \
-    --mount type=volume,volume-driver=cloudstor:aws,source={{.Service.Name}}-{{.Task.Slot}}-vol,destination=/mydata \
+docker service create --replicas 5 --name ping3 \
+    --mount type=volume,volume-driver=docker4x/cloudstor:aws-v{{ edition_version }},source={{.Service.Name}}-{{.Task.Slot}}-vol5,destination=/mydata,volume-opt=perfmode=maxio \
```
FrenchBen: Since we alias the plugin to prevent upgrade failures, shouldn't this `volume-driver` remain `cloudstor:aws`?

mstanleyjones:
@ddebroy PTAL at the copyedit to be sure I didn't change any technical stuff in an incorrect way.

docker-for-aws/persistent-data-volumes.md
```diff
@@ -8,58 +8,252 @@ title: Docker for AWS persistent data volumes
...
 
 ```bash
 {% raw %}
-docker service create --replicas 5 --name ping2 \
-    --mount type=volume,volume-driver=cloudstor:aws,source={{.Service.Name}}-{{.Task.Slot}}-vol,destination=/mydata \
-    alpine ping docker.com
+$ docker service create \
+  --replicas 5 \
+  --name ping3 \
+  --mount type=volume,volume-driver=docker4x/cloudstor:aws-v{{ edition_version }},source={{.Service.Name}}-{{.Task.Slot}}-vol5,destination=/mydata,volume-opt=perfmode=maxio \
```
FrenchBen: Same as before - Since we alias the plugin to prevent upgrade failures, shouldn't this `volume-driver` remain `cloudstor:aws`?

ddebroy:
@mstanleyjones  LGTM.

docker-for-aws/persistent-data-volumes.md
```diff
@@ -14,8 +14,8 @@ mode tasks and regular Docker containers can use a volume created with
 Cloudstor to mount a persistent data volume. In Docker for AWS, Cloudstor has
 two `backing` options:
 
-- `relocatable` data volumes are backed by EBS.
-- `shared` data volumes are backed by EFS.
+- `relocatable` data volumes which are backed by EBS.
```
mstanleyjones: I don't think this is really needed. You took a full sentence and turned it into a fragment.

docker-for-aws/persistent-data-volumes.md
```diff
@@ -167,16 +163,14 @@ the following volume options are available:
...
 $ docker service create \
   --replicas 5 \
   --name ping3 \
   --mount type=volume,volume-driver=cloudstor:aws,source={{.Service.Name}}-{{.Task.Slot}}-vol,destination=/mydata,volume-opt=backing=local,volume-opt=size=25,volume-opt=ebstype=gp2 \
   alpine ping docker.com
-{% endraw %}
 ```
 
-This example creates and mounts a distinct Cloudstor volume backed by 25 GB EBS
+The above example creates and mounts a distinct Cloudstor volume backed by 25 GB EBS
```
mstanleyjones: I don't think this change is necessary.

docker-for-aws/persistent-data-volumes.md
```diff
@@ -215,13 +209,11 @@ you already have too many EBS volumes or want to reduce the amount of time it
 takes to transfer volume data across availability zones.
 
 ```bash
-{% raw %}
```
mstanleyjones: You do need these because of the double curly braces below, which will be interpreted as Liquid.

ddebroy:
@mstanleyjones I have addressed/undid the changes you requested.

docker-for-aws/persistent-data-volumes.md
mstanleyjones: That's correct.

PR #3709, opened 2017-06-26
Second open(2) causes copy-up operation

mgenereu:
The documentation incorrectly credits the first, read-only call to open as the trigger of the copy-up operation.  It's the second, read-write call that actually triggers the copy-up operation.

### DIFF
--- a/engine/userguide/storagedriver/overlayfs-driver.md
+++ b/engine/userguide/storagedriver/overlayfs-driver.md
@@ -498,10 +498,10 @@ filesystems:
   container (`upperdir`).
 
   `yum` is known to be affected unless the `yum-plugin-ovl` package is installed.
-  If the `yum-plugin-ovl` package is not available in your distribution (e.g.
-  RHEL/CentOS prior to 6.8 or 7.2), you may need to run `touch /var/lib/rpm/*`
+  If the `yum-plugin-ovl` package is not available in your distribution such as
+  RHEL/CentOS prior to 6.8 or 7.2, you may need to run `touch /var/lib/rpm/*`
   before running `yum install`. This package implements the `touch` workaround
-  referenced above.
+  referenced above for `yum`.
 
 - **rename(2)**: OverlayFS does not fully support the `rename(2)` system call.
   Your application needs to detect its failure and fall back to a "copy and
--- a/engine/userguide/storagedriver/overlayfs-driver.md
+++ b/engine/userguide/storagedriver/overlayfs-driver.md
@@ -489,7 +489,7 @@ filesystems:
   such operation is the *copy-up* operation. Suppose that  your application calls
   `fd1=open("foo", O_RDONLY)` and then `fd2=open("foo", O_RDWR)`. In this case,
   your application expects `fd1` and `fd2` to refer to the same file. However, due
-  to a copy-up operation that occurs after the first calling to `open(2)`, the
+  to a copy-up operation that occurs after the second calling to `open(2)`, the
   descriptors refer to different files.
 
   `yum` is known to be affected unless the `yum-plugin-ovl` package is installed.
--- a/engine/userguide/storagedriver/overlayfs-driver.md
+++ b/engine/userguide/storagedriver/overlayfs-driver.md
@@ -490,12 +490,18 @@ filesystems:
   `fd1=open("foo", O_RDONLY)` and then `fd2=open("foo", O_RDWR)`. In this case,
   your application expects `fd1` and `fd2` to refer to the same file. However, due
   to a copy-up operation that occurs after the second calling to `open(2)`, the
-  descriptors refer to different files.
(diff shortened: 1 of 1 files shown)


mstanleyjones:
@cpuguy83 PTAL

cpuguy83:
This issue occurs when opening a file as RO, and then opening it again as RW. When the 2nd fd is written to, it triggers a copy-up operation, and then the first fd is read from and you end up with a data mismatch because fd1 is pointing to the read-only file in the lower layer, and fd2 is pointing to the new file that was copied up.

mgenereu:
While I was only changing the text to credit the second open as the culprit, @cpuguy83 has provided a deeper explanation of what the second file open does to get all he way to the problem. Want me to expand the patch to include this?

cpuguy83:
It's probably worth it since developers can easily be effected by this in their own applications.

mstanleyjones:
How are you supposed to work around it? When you open the file the second time, do you get a new file descriptor and you're supposed to use that?

cpuguy83:
In this case it's the first fd that's going to have stale data.
The work-around would be to read/write from a single FD, or close/re-open the file after it's been opened for RW.

`yum` suffers from this issue, there's a hack to fix it.

mgenereu:
The hack is to `touch` the files before `open`.  The updated date time causes the copy-up operation to occur and the `open` points to the same file for all following requests.  I'll update the patch.

engine/userguide/storagedriver/overlayfs-driver.md
```diff
@@ -490,12 +490,18 @@ filesystems:
...
+  descriptors refer to different files. The `fd1` continues to reference the file
+  in the image (`lowerdir`) and the `fd2` references the file in the container
+  (`upperdir`). A workaround for this is to `touch` the files which causes the
+  copy-up operation to happen. All subsequent `open(2)` operations regardless of 
+  read-only or read-write access mode will be referencing the file in the 
+  container (`upperdir`).
 
   `yum` is known to be affected unless the `yum-plugin-ovl` package is installed.
   If the `yum-plugin-ovl` package is not available in your distribution (e.g.
   RHEL/CentOS prior to 6.8 or 7.2), you may need to run `touch /var/lib/rpm/*`
```
mstanleyjones: I don't think we support Centos6 at all so can you adjust this?

engine/userguide/storagedriver/overlayfs-driver.md
```diff
@@ -490,12 +490,18 @@ filesystems:
...
-  before running `yum install`.
+  before running `yum install`. This package implements the `touch` workaround
+  referenced above.
```
mstanleyjones: Should you clarify that this workaround only fixes `yum` and that the bug might come up in custom apps or other places?

engine/userguide/storagedriver/overlayfs-driver.md
```diff
@@ -490,12 +490,18 @@ filesystems:
...
-  descriptors refer to different files.
+  descriptors refer to different files. The `fd1` continues to reference the file
+  in the image (`lowerdir`) and the `fd2` references the file in the container
+  (`upperdir`). A workaround for this is to `touch` the files which causes the
+  copy-up operation to happen. All subsequent `open(2)` operations regardless of 
+  read-only or read-write access mode will be referencing the file in the 
+  container (`upperdir`).
 
   `yum` is known to be affected unless the `yum-plugin-ovl` package is installed.
   If the `yum-plugin-ovl` package is not available in your distribution (e.g.
```
mstanleyjones: Can you s/e.g./such as, and then you won't need the parens either?

engine/userguide/storagedriver/overlayfs-driver.md
mgenereu: Cento6 as a docker host or as a docker image?  I took this to mean image.

mgenereu: (I didn't type this part.)

mstanleyjones: Oh you're right. @cpuguy83 is this for centos6 image?

engine/userguide/storagedriver/overlayfs-driver.md
mgenereu: Done.

engine/userguide/storagedriver/overlayfs-driver.md
mgenereu: Done.

engine/userguide/storagedriver/overlayfs-driver.md
mgenereu: This whole section is about OverlayOS problems within the images.

mstanleyjones:
Gently reworded the commit message during squash. Thanks @mgenereu

PR #3715, opened 2017-06-27
Update docker-machine completion link for BASH

### LINKED ISSUE #3700: link to docker completion file returns 404 Not found - needs update
[machine/completion.md](https://docs.docker.com/machine/completion/), CC @londoncalling

ShockN745:
This should now fix #3700

### DIFF
--- a/machine/completion.md
+++ b/machine/completion.md
@@ -21,13 +21,13 @@ Place the completion script in `/etc/bash_completion.d/` as follows:
 *   On a Mac:
 
     ```shell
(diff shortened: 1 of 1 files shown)


mstanleyjones:
This isn't the correct fix. See discussion in #3717. Good news is that in 17.06 these scripts will be installed for you automatically, even on Linux!

mstanleyjones:
@londoncalling can this be closed now?

londoncalling:
@mstanleyjones  yes I think so

londoncalling:
@ShockN745 Thank you for this PR! We ended up fixing this in some other PR's. It's a bit confusing, but as @mstanleyjones points out^^, things should get better going forward with `17.06` and beyond.

PR #3717, opened 2017-06-27
Corrected URL for autocomplete scripts

lichtl:
Corrected URL for autocomplete scripts


thaJeztah:
hm, not sure about this; ideally we'd point to the version that was included as part of a particular release, otherwise the completion scripts don't match that version (i.e., point to a specific _tag_ in the docker/docker-ce repository, like https://github.com/docker/docker-ce/tree/v17.06.0-ce-rc5/components/cli/contrib/completion/bash)

_However_, with docker 17.06 onwards, bash, zsh, and fish completion scripts should be automatically installed when installing Docker, so no action is needed. Completion scripts are also installed as part of the `.deb` and `rpm` packages.

We can refer to the GitHub repositories if someone wants to install the "most up to date" version, but there's still a risk that they don't match the version of docker that's installed

mstanleyjones:
@lichtl can you please modify your PR with the above info, and also adjust your previous ZSH fix, and rebase on `vnext-engine` branch?

lichtl:
ok think I need a new PR on that (just rebased on vnext-engine) Gonna resolve this and submit new PR then.

PR #3724, opened 2017-06-28
add option: experimental

konstruktoid:

### Proposed changes

`option: experimental` is missing on https://docs.docker.com/edge/engine/reference/commandline/dockerd/


mstanleyjones:
@tiborvass @thaJeztah we need to figure out how to generate this file programmatically like the others, so we don't get out of sync like this.

PR #3725, opened 2017-06-28
Update part5.md

Mesqualito:
Small proposal for a possible fix in the documentation. And my first pull request - sorry, if I do anything wrong...

### DIFF
--- a/get-started/part5.md
+++ b/get-started/part5.md
@@ -244,7 +244,7 @@ Redis service.  Be sure to replace `username/repo:tag` with your image details.
     $ docker-machine ssh myvm1 "docker stack deploy -c docker-compose.yml getstartedlab"
     ```
 
-5.  Check the web page at `http://localhost` and you'll see the results of the visitor counter, which is now live and storing information on Redis.
+5.  Check the web page at one of your nodes (e.g. `http://192.168.99.101`) and you'll see the results of the visitor counter, which is now live and storing information on Redis.
 
     ![Hello World in browser with Redis](images/app-in-browser-redis.png)
 


get-started/part5.md
```diff
@@ -244,7 +244,7 @@ Redis service.  Be sure to replace `username/repo:tag` with your image details.
     $ docker-machine ssh myvm1 "docker stack deploy -c docker-compose.yml getstartedlab"
     ```
 
-5.  Check the web page at `http://localhost` and you'll see the results of the visitor counter, which is now live and storing information on Redis.
+5.  Check the web page at one of your nodes (e.g. `http://192.168.99.101`) and you'll see the results of the visitor counter, which is now live and storing information on Redis.
```
mstanleyjones: Actually, you can check on any of your nodes or on `localhost` in this case, because the swarm routing mesh takes care of this for you. Can you tweak the wording to make this more clear?

Mesqualito: I did try to browse localhost, which didn't work in my case - while everything else was exactly working as described. I will repeat all the steps in the very good Get-Started-Documentation and try again today. Thanks for your hint - I will learn to understand the routing mesh.

Mesqualito: Sorry for bothering again: but is my localhost with its local docker-daemon really part of the swarm routing mesh?

I start the services on mydockervm1 (= myvm1) as swarm-leader. So mydockervm1 and mydockervm2 (as worker) are the nodes behind the containers. Port 80 from the containers will be exposed (=> Dockerfile) and the routing mesh will be completed with docker-compose.yml on mydockervm1: ports "80:80" for all nodes in the swarm. But only mydockervm1 (a Windows 10 Hyper-V Virtual-Machine) and mydockervm2 (another machine) are part of the swarm, not localhost.

In part4 it's described as: "Accessing your cluster - You can access your app from the IP address of either myvm1 or myvm2. The network you created is shared between them and load-balancing."

Or do I still missunderstand something?

mstanleyjones:
I think `localhost` being included might be a feature of Docker for Mac or Windows. Maybe you're right and it's better to leave it how you have it.

PR #3730, opened 2017-06-28
Modify the rails command to more gracefully stop

### LINKED ISSUE #989: How to safely stop the images
[compose/rails.md](https://docs.docker.com/compose/rails/)

>  Note: If you stop the example application and attempt to restart it, you might get the following error: web_1 | A server is already running. Check /myapp/tmp/pids/server.pid. One way to resolve this is to delete the file tmp/pids/server.pid, and then re-start the application with docker-compose up.

Unless `Ctrl-C` is not the preferred way to terminate the process, this issue happens 100% of the time (for me\*). Per docker/compose#1393, the workaround is to change the command to the following:

```yaml
command: bash -c "rm -f tmp/pids/server.pid && bundle exec rails s -p 3000 -b '0.0.0.0'"
```

Would there be any opposition if I were to create a PR which amended this tutorial? If not, would the preference be to **A.)** update the `command` line directly and remove the note or **B.)** amend the note to suggest this alternative?

-----------------

_\* — System/environment specs are as follows (macOS):_

```
Version 1.12.5 (14777)
Channel: Stable
3e6f00c1dc
```

mstanleyjones:
Fixes #989

Tweak the command so that if you Ctrl+C you can restart the bundle without having to clean up a temp file

### DIFF
--- a/compose/rails.md
+++ b/compose/rails.md
@@ -53,6 +53,7 @@ to link them together and expose the web app's port.
         image: postgres
       web:
         build: .
+        command: bash -c "rm -f tmp/pids/server.pid && bundle exec rails s -p 3000 -b '0.0.0.0'"
         command: bundle exec rails s -p 3000 -b '0.0.0.0'
         volumes:
           - .:/myapp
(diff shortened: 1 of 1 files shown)


mstanleyjones:
@shin- PTAL

compose/rails.md
```diff
@@ -53,6 +53,7 @@ to link them together and expose the web app's port.
         image: postgres
       web:
         build: .
+        command: bash -c "rm -f tmp/pids/server.pid && bundle exec rails s -p 3000 -b '0.0.0.0'"
```
shin-: My 2c:
Isn't this more obscure / difficult to read? I can see the value of having something that "just works", but I'm concerned this might be taking away from the educational value of being able to see "oh, `command` is just the command I would be running to start my app, cool!"

compose/rails.md
```diff
@@ -53,6 +53,7 @@ to link them together and expose the web app's port.
...
         command: bundle exec rails s -p 3000 -b '0.0.0.0'
```
shin-: If you do go ahead with this change, this line needs to be removed.

PR #3735, opened 2017-06-28
Deprecate some Dockerize topics in favor of library ones

mstanleyjones:
### Proposed changes

The Mongo, Redis, and Couchbase topics in the docs are out of date compared to the ones we are pulling from the samples. Maybe others are too, but I did these first. Changed the TOC and added redirects too.

You won't be able to preview this on Netlify, you'll have to build locally.


mstanleyjones:
I think we should have the old links in place for a little while because people are used to going there. Maybe we can do it in two phases? I guess I don't have Analytics evidence of it, though.

PR #3756, opened 2017-06-30
Update broken link to post install steps Fedora

stevensona:
### Proposed changes

I clicked the link, and it was broken. I updated the link to go to the right page.


mstanleyjones:
Can you please double-check the link? It works for me: https://docs.docker.com/engine/installation/linux/docker-ce/fedora/, go to the bottom and the link URL is https://docs.docker.com/engine/installation/linux/linux-postinstall/, which works for me. Your fix actually breaks it as the file is a directory up from the `fedora.md` which is in the `docker-ce/` subdirectory.

stevensona:
Misty,
I apologize. The link is indeed broken--I just ~~fixed~~ broke a working link. There are 3 links to postinstall steps, and one of the three is different.

PR #3770, opened 2017-06-30
Don't run bundle update in the Dockerfile

mstanleyjones:
This avoids the US-ASCII bug in jekyll-redirect plugin

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,5 @@
 FROM starefossen/github-pages:137
 
-COPY . /usr/src/app/
+COPY . /usr/src/app
 
-CMD jekyll serve -H 0.0.0.0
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -1,5 +1,5 @@
 FROM starefossen/github-pages:137
 
-COPY . /usr/src/app/
+COPY . /usr/src/app
```
thaJeztah: Can you change this back, as it doesn't seem related to this change

mstanleyjones: It is related. One is what @johndmulhausen told me to use and the other one (without the `/`) is what is in the `onbuild` version of the `github-pages` plugin. Probably it doesn't matter but I thought maybe it was best to stick to how they had it.

PR #3771, opened 2017-06-30
New notes for 2.2.6

pdevine:
### Proposed changes

Updated release notes for 2.2.6 (to be released July 3rd, 2017)


joaofnfernandes:
It seems Jenkins is having problems with `git checkout`, and I don't see anything here that could break the build so let's merge.

PR #3772, opened 2017-06-30
Remove OS name from Ubuntu and Debian packages

mstanleyjones:
Relates to https://github.com/docker/docker-ce-packaging/pull/27 and https://github.com/docker/for-linux/issues/50

cc/ @andrewhsu

### DIFF
--- a/engine/installation/linux/docker-ce/debian.md
+++ b/engine/installation/linux/docker-ce/debian.md
@@ -173,7 +173,7 @@ from the repository.
 
     > **Note**: Starting with Docker 17.06, stable releases are also pushed to
     > the **edge** and **testing** repositories.
-    
+
     [Learn about **stable** and **edge** channels](/engine/installation/).
 
 #### Install Docker CE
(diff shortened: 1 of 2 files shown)


engine/installation/linux/docker-ce/debian.md
```diff
@@ -208,16 +208,15 @@ from the repository.
     ```bash
     $ apt-cache madison docker-ce
 
-    docker-ce | {{ minor-version }}.0~ce-0~debian-jessie | {{ download-url-base}} jessie/stable amd64 Packages
+    docker-ce | {{ minor-version }}.0~ce-0~debian | {{ download-url-base}} jessie/stable amd64 Packages
```
thaJeztah: @andrewhsu Is there a way to remove the suffix from older packages as well? The instructions show how to install the _latest_ version, but also explain people can pin to an older version, which won't work because those still have a suffix

mstanleyjones: Yeah that would probably be the most transparent way to do it. But will it break people who have already pinned? Can we keep the packages there with the old names but have them with the new names too?

andrewhsu: These instruction changes look generic enough to be flexible for any package version format in the past and future. I'd prefer not to have to provide multiple names of the same identical package if possible.

mstanleyjones:
@londoncalling his comments direct us to keep the  changes in this PR as opposed to the suggestion that @thaJeztah made, unless I am reading them wrong. Thanks for the review!

PR #3774, opened 2017-07-01
Using a process manager is platform-dependent

InfinityGone:
### Proposed changes

The docs explaining why [using a process manager](https://docs.docker.com/engine/admin/start-containers-automatically/#use-a-process-manager) isn't recommended should either read "because it is platform-dependent" or "because it is not platform-independent". I more or less chose the former at random.


InfinityGone:
Nevermind, it's correctly worded. It's not "platform-dependent" on docker.

PR #3776, opened 2017-07-01
Don't remove `container-selinux` when removing old Docker version

dsanders11:
Unless I'm missing something, `container-selinux` is still a dependency of `docker-ce` so it doesn't make sense to remove it when removing old versions of `docker` when switching to `docker-ce`.


mstanleyjones:
@cpuguy83 PTAL

PR #3798, opened 2017-07-03
Try to update gitmodules to not fail gh-pages

mstanleyjones:
Github Pages is supposed to work with submodules: https://help.github.com/articles/using-submodules-with-pages/

I wonder if maybe we were missing the branch, so this tries to fix that. Let's see.


johndmulhausen:
Worth a shot!

PR #3799, opened 2017-07-03
Document the user for Docker for AWS instances

### LINKED ISSUE #3713: Document EC2 User as "docker@<ec2-instance>"
### Problem description
There is no information about how to log in to the EC2 instance that is the Swarm manager. Typical users are: root, ubuntu, ec2-user. No where in the documentation (unless you look in the CloudFormation Template) is it documented that this has changed to user "docker"

Ergo, ssh-ing to EC2 with:

```
ssh -i <key>.pem ec2-user@<swarm-manager>
```

to _test the installation_ will always return Permission Denied

### Problem location

https://docs.docker.com/docker-for-aws/

https://docs.docker.com/docker-for-aws/

- Other: <DETAILS>

### Suggestions for a fix

Make sure that it is outlined that the user for AWS has been changed to "docker" instead of root/ec2-user/ubuntu.

mstanleyjones:
Fixes #3713 

@friism @FrenchBen @kencochrane PTAL

### DIFF
--- a/docker-for-aws/index.md
+++ b/docker-for-aws/index.md
@@ -165,6 +165,9 @@ Docker for AWS starts with a CloudFormation template that will create everything
 
 The CloudFormation template first creates a new VPC along with subnets and security groups. After the networking set-up completes, two Auto Scaling Groups are created, one for the managers and one for the workers, and the configured capacity setting is applied. Managers start first and create a quorum using Raft, then the workers start and join the swarm one at a time. At this point, the swarm is comprised of X number of managers and Y number of workers, and you can deploy your applications. See the [deployment](deploy.md) docs for your next steps.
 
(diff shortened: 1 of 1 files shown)


friism:
I'm not sure the premise of the but report is correct since the user to use is documented: https://github.com/mstanleyjones/docker.github.io/blob/4c57b5f1e35521a41e7555f8e97c1c63a3af4ee3/docker-for-aws/deploy.md#manager-nodes

... but sure.

mstanleyjones:
To be fair, I think they only used the quickstart, and didn't carefully read the deployment topic you linked to. :)

mstanleyjones:
Amazing idea!

londoncalling:
@mstanleyjones I'm smart.

PR #3580, opened 2017-06-12
added links to Compose release versions, improved matrix description

londoncalling:
### What's new

- Added links to Compose releases page
- Clarified the relationship between Compose releases, Compose file format versions, and Docker Engine releases
- Added a note about how to find information about old Docker versions, Compose releases, and file format versions

@friism @shin- and I discussed this, and we think it doesn't make sense to include such a detailed matrix for older versions of Docker and Compose releases, when that information is available in the release notes. So, what I did as a compromise was try to clarify how these releases and versions related, and suggest how to find details on particular releases in the release notes. I also surfaced the existing matrix a little more by giving it its own heading in the main file reference docs.

### Related

Fixes #3404  

### Reviewers

@friism @shin- @Rots @mstanleyjones 

Signed-off-by: Victoria Bialas <victoria.bialas@docker.com>

PR #3831, opened 2017-07-06
Update compose matrix to include Compose versions

andrerom:
### Proposed changes

Changes the docker compose matrix to also include the version of Docker compose where support for given file format was introduced. To avoid having to crawl true [release notes](https://github.com/docker/compose/releases) every time you wonder about this for updating your minimum requirements and so on when bumping the file format version.

### DIFF
--- a/_includes/content/compose-matrix.md
+++ b/_includes/content/compose-matrix.md
@@ -1,15 +1,15 @@
 This table shows which Compose file versions support specific Docker releases.
 
-| **Compose file format** | **Docker Engine release** |
-|  -------------------    |    ------------------     |
-|      3.3                |       17.06.0+            |
-|      3.2                |       17.04.0+            |
-|      3.1                |       1.13.1+             |
-|      3.0                |       1.13.0+             |
-|      2.2                |       1.13.0+             |
-|      2.1                |       1.12.0+             |
-|      2.0                |       1.10.0+             |
-|      1.0                |       1.9.1.+             |
+| **Compose file format** | **Docker Engine release** | **Docker Compose release** |
+|  -------------------    |    ------------------     |    -------------------     |
+|      3.3                |       17.06.0+            |           1.14.0+          |
+|      3.2                |       17.04.0+            |           1.12.0+          |
+|      3.1                |       1.13.1+             |           1.11.1+          |
+|      3.0                |       1.13.0+             |           1.10.0+          |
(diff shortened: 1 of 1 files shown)


andrerom:
Now the only missing piece of info is what changed between the given smaller releases _(3.0 => 3.1 => 3.2 => 3.3)_, this seems complex as some fetaturs seems ot be added afterwards to given versions.

E.g from recent 1.14.0:
```
## Compose file version 2.2:
- Added the following parameters in service definitions: cpu_count,
cpu_percent, cpus
```

mstanleyjones:
@shin- PTAL

shin-:
See previous efforts in this direction: https://github.com/docker/docker.github.io/pull/3580

The concern to me is that people will see it as a tacit endorsement of (sometimes dramatically) older versions of the software which we no longer support.

londoncalling:
Thanks @shin- for finding the previous thread, and clarifying that here. Let's keep this out of the docs then, as we decided before. It's sending the wrong message both per the complexities of small releases that are skated over per the comments above, and for the larger issue re: unsupported older versions.

andrerom:
Ok, but re-phrasing what I said in the description: _As software vendor, when I bump Compose file version and need to update minimum requirements, not having this is painful_

andrerom:
But maybe @shin- is saying what I was saying in the side note above, given certain features of a compose file version is added _(fixed, improved, forgotten)_ later, I can see why adding this can be largely misleading.

It's basically close to impossible unless you test your compose file defintion across all "supported" _(by the given software vendor)_ docker compose versions.

So I'm fine with this being closed if that is the reason.

Question is if the list of Docker versions have the same flaw, aka are there changes/fixes/improvements in a versioned file format behavior in later versions?

londoncalling:
@shin- and/or @dnephin  can you respond to @andrerom 's last question just above ^^ ?

dnephin:
I'm not sure I understand the question.

If there are bugs in a version of docker-compose or docker/cli, the fix could apply to all version of the Compose file supported by that tool, yes.

andrerom:
@dnephin I was referring to these kind of changes to docker compose format support as found in docker compose v1.4.0:
```
Compose file version 2.2

- Added the following parameters in service definitions: cpu_count,
cpu_percent, cpus
```

But I don't think I have seen similar changes in later releases in docker itself, so I guess that answers my question. _(as in: no the same issue does not apply to docker versions listed here in the matrix)_

PR #3848, opened 2017-07-08
Correct link to linux-postinstall.md

andresmaiden:
The link is incorrect so I put the correct url

### Proposed changes

I correct a simple link, the original file is in /engine/installation/linux/linux-postinstall.md and the actual link is to linux-postinstall.md in the /engine/installation/linux/docker-ce folder so it doesn't exists

### DIFF
--- a/engine/installation/linux/docker-ce/ubuntu.md
+++ b/engine/installation/linux/docker-ce/ubuntu.md
@@ -232,7 +232,7 @@ the repository.
     container runs, it prints an informational message and exits.
 
 Docker CE is installed and running. You need to use `sudo` to run Docker commands.
-Continue to [Linux postinstall](linux-postinstall.md) to allow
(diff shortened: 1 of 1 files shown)


vochicong:
I think that relative linking `../linux-postinstall.md` is better than `/engine/installation/linux/linux-postinstall.md`

andresmaiden:
@vochicong I tried to follow the same path as in https://raw.githubusercontent.com/docker/docker.github.io/master/engine/installation/linux/docker-ce/fedora.md
https://raw.githubusercontent.com/docker/docker.github.io/master/engine/installation/linux/docker-ce/debian.md
Now I notice we have the same problem in
https://raw.githubusercontent.com/docker/docker.github.io/master/engine/installation/linux/docker-ce/centos.md

For me is the same, what ever you prefered...

johndmulhausen:
Fixed elsewhere with the preferred relative link. Thanks!

PR #3850, opened 2017-07-09
correct path to "Linux postinstall"

vochicong:
[Linux postinstall](../linux-postinstall.md) is in the parent folder.


johndmulhausen:
So it is, thanks @vochicong :)

PR #3852, opened 2017-07-09
Update mongodb.md

dm-wyncode:
### Proposed changes

I was double-checking that a mongodb instance running in a Docker container was not exposed to the outside. Even with ufw not allowing port 27017 and with the default net.ip binding for mongodb set to 127.0.0.1 the database was still exposed.

When I started the container with `docker run -p 27017:27017 --name mongo_instance_001 -d my/repo` I was still able to access the database from the outside.

I could no longer access the database from the outside when I started the database with  `docker run -p 127.0.0.1:27017:27017 --name mongo_instance_001 -d my/repo`


johndmulhausen:
Apologies, but we actually just removed this topic in favor of the library reference. Hopefully the replacement topic is useful. Thanks so much for the PR, don't let it discourage you that we closed this one!

PR #3860, opened 2017-07-11
Fix "Installing Ruby" link

asottile:
Not sure if this renders anywhere outside of github, but this fixes the link on https://github.com/docker/docker.github.io


johndmulhausen:
Thanks, @asottile !

PR #3877, opened 2017-07-13
Add CSS for option to create inline images (like whale menu)

londoncalling:
### What's changed

- added CSS class for an inline image

- added tests in Docker for Mac

- updated all occurrences of references to Moby whale in menu to use the new class, but we'll have to change these because the tests show that using the `div` tags inline doesn't work (`div` tags show up in rendered HTML). Using an embedded style directly from within the `img` tag does work, but there is a disconnect when trying to use the class in the `scss` file instead.

### Here's what works, and what doesn't work

Here’s what works:

- Calling the image like this directly as an embedded style: ` `

- Both `display: inline` and `display: auto` work

Here is what doesn’t work:

- When you flip it to call the class from the `scss`, that does not work. It’s just not reading that file, because I tried other tags in there too for the `img` and it doesn't read the values for any of them, but they work if I embed them directly in the file we are testing in.

### Reviewers

@jsouth , run the build locally then go to: http://localhost:4000/docker-for-mac/#preferences and look at the tests under the subhead "**Josh look here**"

### DIFF
--- a/_scss/_utilities.scss
+++ b/_scss/_utilities.scss
@@ -58,7 +58,7 @@ i.fa.fa-outdent {
     padding: 0px 15px 10px 15px;
 }
 
-.image-inline img {
+img.inline {
     display: inline;
 }
 
--- a/_scss/_utilities.scss
+++ b/_scss/_utilities.scss
@@ -58,6 +58,9 @@ i.fa.fa-outdent {
     padding: 0px 15px 10px 15px;
 }
 
+.image-inline img {
+    display: inline;
+}
 
 /*
 * Search *********************************************************************
@@ -259,4 +262,3 @@ div#autocompleteResults span {
 input.gsc-search-button, input.gsc-search-button:hover, input.gsc-search-button:focus {
     display: none!important;
 }
-
--- a/docker-for-mac/index.md
+++ b/docker-for-mac/index.md
@@ -100,10 +100,6 @@ docker-machine version 0.10.0, build 76ed2a6
 
 ## Preferences
 
-This is test calling the class from the `scss` file in the `img` tag (instead of
-embedding the style directly like the previous tests) here, ![whale menu](/docker-for-mac/images/whale-x.png){: .inline} but it does not
-work. This file does not seem to be reading the `scss` file.
-
 Choose ![whale menu](/docker-for-mac/images/whale-x.png){: .inline} ->
 **Preferences** from the menu bar.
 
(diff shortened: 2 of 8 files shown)


_scss/_utilities.scss
```diff
@@ -58,6 +58,9 @@ i.fa.fa-outdent {
     padding: 0px 15px 10px 15px;
 }
 
+.image-inline img {
```
thaJeztah: try changing this to

```css
img.image-inline {
    display: inline;
}
```

The style your currently defined would match an `` element that's a child element of an element that has the `.image-inline` class, for example;

```html
<p class="image-inline"><img src="foobar.png /></p>
```

_scss/_utilities.scss
```diff
@@ -58,6 +58,9 @@ i.fa.fa-outdent {
...
+    display: inline;
```
mstanleyjones: You might experiment with `inline-block` here. Not sure what the difference will be. I think you can put height attributes on `inline-block` things but not `inline`.

docker-for-mac/index.md
```diff
@@ -100,8 +100,36 @@ docker-machine version 0.10.0, build 76ed2a6
 
 ## Preferences
 
-Choose ![whale-x](/docker-for-mac/images/whale-x.png) --> **Preferences** from
-the menu bar.
+### Josh look here
```
mstanleyjones: 😹

docker-for-mac/index.md
```diff
@@ -100,8 +100,36 @@ docker-machine version 0.10.0, build 76ed2a6
...
+sentence ends. This also works!
+
+This is test calling the class from the `scss` file in the `img` tag (instead of
+embedding the style directly like the previous tests) here: <img
+src="/docker-for-mac/images/whale-x.png" class="image-inline" /> but it does not
+work. This file does not seem to be reading the `scss` file.
+
+**END OF IMG AND CSS TESTS**
+
+Choose <img src="/docker-for-mac/images/whale-x.png" class="image-inline" /> ->
```
mstanleyjones: You can keep the Markdown code and still use your class:

```markdown
![whale example](/docker-for-mac/images/whale-x.png){: .image-inline }
```

_scss/_utilities.scss
londoncalling: @mstanleyjones I tried `inline`, `inline-block`, and `auto` (which Josh started with). They all work if they specified directly in the file as `style` but nothing works if I try to call the class in `_scss/_utilities.scss`

londoncalling:
@jsouth @thaJeztah @mstanleyjones this should do it, @thaJeztah please approve. This was a group effort! Thanks!

londoncalling:
@thaJeztah thanks! I'll look into making this changes soon, will merge this for now to get the fix for the line breaks out

PR #3878, opened 2017-07-13
Update json-file.md

matglas:
### Proposed changes
I change `-–` to `--` as the `–` does not work.


johndmulhausen:
Thanks, @matglas -- I bet someone's autocorrect got in the way on that one. :)

PR #3883, opened 2017-07-13
Update reconfigure.md

scjane:
Add an escape character (\) to the cell to prevent wrongly generating a new column.


johndmulhausen:
Thanks, @scjane :)

johndmulhausen:
cc @joaofnfernandes

PR #3884, opened 2017-07-13
Update restore.md

scjane:


johndmulhausen:
Nice backslash escape fix; definite improvement. cc @joaofnfernandes so he sees that. Thanks, @scjane :)

scjane:
It seems that some words is missing at this fix version:

1.  In the "Files changed" tab, click "View the whole file at version db50d70" at the upper right corner.
2.  See the whole file. Compared with its original version, "<mountpoint>" is not displayed as shown in the following figure

words-missing

I don't know why.

joaofnfernandes:
@scjane The files under `datacenter/dtr/2.2/reference/` are automatically generated from code. This means that on the next release of DTR your changes will not be persisted.
Since this project is close sourced, there's no way for you to fix things directly but I can port some of these changes for you.

PR #3907, opened 2017-07-17
Fix path to requirements.txt

akmalabbasov:


johndmulhausen:
@akmalabbasov  I'm pretty sure this is not necessary; can you provide repro steps that produce the error this fixes?

akmalabbasov:
@johndmulhausen you're right, sorry by bad :)

PR #3924, opened 2017-07-19
Note default max-size and max-file for json-file

trapier:
The json-file logger does not rotate log files by default.  This
behavior is worthy of documentation.


trapier:
rebased on latest master.

PR #3929, opened 2017-07-20
Update binaries.md

scjane:


mstanleyjones:
This will need to be rebased after #3977 is merged.

PR #3932, opened 2017-07-20
added Compose topics on caching options for performance

### LINKED ISSUE #3921: No documentation for volume caching directives
[compose/compose-file/index.md](https://docs.docker.com/compose/compose-file/), CC @londoncalling 

The compose file reference doesn't seem to have been updated with the caching directives as introduced by the resolution of this issue: https://github.com/docker/for-mac/issues/1592#issuecomment-300150553

londoncalling:
### What's changed

- Added topic with example on caching options to Compose version 3 file reference
- Cross referenced it with the details on these options in the Docker for Mac topic on performance tuning

### Questions for Reviewers

@shin- @dnephin @dsheets @kkrauth @yallop Please let me know if these additions to the docs make sense, and are in the right place. Also, I assume that if one of these options is specified on a volume (`consistent`, `cached`, or `delegated`) and the platform is _*not*_ Mac, the option is ignored. Is that true? If so, we should probably add that here, right?

Also let me know If you want me to add this to Compose version 2 file reference.

### Related

Fixes #3921 

### DIFF
--- a/compose/compose-file/index.md
+++ b/compose/compose-file/index.md
@@ -1431,7 +1431,8 @@ services:
 On Docker 17.04 CE Edge and up, including 17.06 CE Edge and Stable, you can
 configure container-and-host consistency requirements for bind-mounted
 directories in Compose files to allow for better performance on read/write of
-volume mounts. These options address issues specific to `osxfs` file sharing, and therefore are only applicable on Docker for Mac.
+volume mounts. These options address issues specific to `osxfs` file sharing,
+and therefore are only applicable on Docker for Mac.
 
 The flags are:
 
--- a/compose/compose-file/index.md
+++ b/compose/compose-file/index.md
@@ -1426,9 +1426,49 @@ services:
         constraints: [node.role == manager]
 ```
 
+#### Performance tuning for volume mounts (Docker for Mac)
+
+On Docker 17.04 CE Edge and up, including 17.06 CE Edge and Stable, you can
+configure container-and-host consistency requirements for bind-mounted
+directories in Compose files to allow for better performance on read/write of
+volume mounts. These options address issues specific to `osxfs` file sharing, and therefore are only applicable on Docker for Mac.
+
+The flags are:
+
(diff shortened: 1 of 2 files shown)


londoncalling:
@dnephin well, I thought from my read of the long thread on issue #1592 that at least two options were available (`cached` and/or `delegated` since right now they do the same thing, and `consistent`). If this is not the case, I can save this until they are implemented. (Also, maybe my reading of the thread was influenced by the ping for not having it in the docs.)

dnephin:
`consistent` is the default, so I'm sure it works, I know `delegated` was added at some point, but I'm not sure which version.

@dsheets can tell us more

dsheets:
The options available in the short form are `delegated`, `cached`, and `consistent`. The option `default` is also available in the long form (`consistency=default`). Currently there are two behaviors that are observable: `cached` and `consistent`. `default` is the same as `consistent` (but can be overridden by `cached` or `delegated` where `consistent` cannot). `cached` and `delegated` currently behave the same but will not in the future. These options change behavior in Docker for Mac but will soon be available to volume plugins and mount point plugins on all platforms.

londoncalling:
Great! Thanks @dnephin and @dsheets , it sounds like to me then that I can
merge the current updates and then can update again when these options are
available on other platforms besides Docker for Mac

compose/compose-file/index.md
```diff
@@ -1426,9 +1426,49 @@ services:
         constraints: [node.role == manager]
 ```
 
+#### Caching options for volume mounts (Docker for Mac)
+
+On Docker 17.04 CE Edge and up, including 17.06 CE Edge and Stable, you can
+configure container-and-host consistency requirements for bind-mounted
+directories in Compose files to allow for better performance on read/write of
+volume mounts. These options address issues specific to `osxfs` file sharing, and therefore are only applicable on Docker for Mac.
```
JimGalasyn: Nit: Maybe wrap this line at 80 chars.

PR #3936, opened 2017-07-20
Update upgrade notes for Docker4Azure

ddebroy:
### Proposed changes

Update the notes for upgrading Docker4Azure to describe the new utility/wrapper script around the docker commandline to invoke the upgrade container.

### Related issues (optional)
https://github.com/docker/editions/pull/1245

@FrenchBen @mstanleyjones

### DIFF
--- a/docker-for-azure/upgrade.md
+++ b/docker-for-azure/upgrade.md
@@ -42,6 +42,18 @@ $ docker run \
   docker4x/upgrade-azure:17.06.0-ce-azure1
 ```
 
+If you are already on a version more recent than 17.06.1 CE or 17.07.0 CE, you can use the below script to initiate the upgrade:
+
+```bash
+$ upgrade.sh YY.MM.X-ce-azureC
+```
+
+For example, to upgrade from 17.06.1 CE edge to 17.07.0 CE edge, you can use:
+
+```bash
+$ upgrade.sh 17.07.0-ce-azure1
+```
+
 This initiates a rolling upgrade of the Docker swarm. Service state is maintained during and after the upgrade. Appropriately scaled services should not experience downtime during an upgrade. Single containers which are not part of services (for example, containers started with `docker run`) are **not** preserved during an upgrade. This is because they are not Docker  services but are known only to the individual Docker engine where they are running.
 
 ## Monitoring
--- a/docker-for-azure/upgrade.md
+++ b/docker-for-azure/upgrade.md
@@ -42,19 +42,19 @@ $ docker run \
   docker4x/upgrade-azure:17.06.0-ce-azure1
 ```
 
-If you are already on a version more recent than 17.06.1 CE or 17.07.0 CE, you can use the below script to initiate the upgrade:
+If you are already on a version more recent than 17.06.1 CE or 17.07.0 CE, you can use `upgrade.sh` to initiate the upgrade:
 
 ```bash
 $ upgrade.sh YY.MM.X-ce-azureC
 ```
 
-For example, to upgrade from 17.06.1 CE edge to 17.07.0 CE edge, you can use:
+This example upgrades Docker for Azure from 17.06.1 CE Edge to 17.07.0 CE Edge:
 
 ```bash
 $ upgrade.sh 17.07.0-ce-azure1
 ```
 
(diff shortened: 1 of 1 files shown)


docker-for-azure/upgrade.md
```diff
@@ -42,6 +42,18 @@ $ docker run \
...
+If you are already on a version more recent than 17.06.1 CE or 17.07.0 CE, you can use the below script to initiate the upgrade:
+
+```bash
+$ upgrade.sh YY.MM.X-ce-azureC
+```
+
+For example, to upgrade from 17.06.1 CE edge to 17.07.0 CE edge, you can use:
+
+```bash
+$ upgrade.sh 17.07.0-ce-azure1
```
FrenchBen: Getting the exact build may be difficult, since we don't advertise them here:
https://docs.docker.com/docker-for-azure/release-notes/
It would be ideal if we could do: `upgrade 17.07.0` and we figure out the latest version.

The script could then lookup the proper tag via:
`docker run --rm docker4x/awscli:latest s3api --no-sign-request get-object --bucket editions-us-east-1 --key azure/stable/17.06.0/Docker.tmpl docker.out | jq -r '.Metadata | .edition_version'`

ddebroy: I think updating the  the latest version somewhere in release notes or this upgrade page would be ideal to keep users aware. I think having `upgrade.sh` from a past version try to compute any aspect of a future version will break if the logic it's relying on is not adhered to in the future version. For example, if a decision is made in the future to change the bucket `editions-us-east1` for any legitimate reason, the command line above will break.

docker-for-azure/upgrade.md
```diff
@@ -42,6 +42,18 @@ $ docker run \
   docker4x/upgrade-azure:17.06.0-ce-azure1
 ```
 
+If you are already on a version more recent than 17.06.1 CE or 17.07.0 CE, you can use the below script to initiate the upgrade:
```
mstanleyjones: s/below script/`upgrade.sh` script/

docker-for-azure/upgrade.md
```diff
@@ -42,6 +42,18 @@ $ docker run \
...
+
+```bash
+$ upgrade.sh YY.MM.X-ce-azureC
+```
+
+For example, to upgrade from 17.06.1 CE edge to 17.07.0 CE edge, you can use:
```
mstanleyjones: This example upgrades Docker for Azure from 17.06.1 CE Edge to 17.07.0 CE Edge:

Also, you should probably make it clear where the `upgrade.sh` script is expected to be, rather than just calling it from the current working directory.

docker-for-azure/upgrade.md
```diff
@@ -42,6 +42,18 @@ $ docker run \
...
+
+```bash
+$ upgrade.sh 17.07.0-ce-azure1
+```
+
 This initiates a rolling upgrade of the Docker swarm. Service state is maintained during and after the upgrade. Appropriately scaled services should not experience downtime during an upgrade. Single containers which are not part of services (for example, containers started with `docker run`) are **not** preserved during an upgrade. This is because they are not Docker  services but are known only to the individual Docker engine where they are running.
```
mstanleyjones: Is there any way to customize the rolling upgrade, when using the `upgrade.sh` script?

Nit: double space between "Docker" and "services".

Is there a way to preserve `docker run` containers during an upgrade? Perhaps by setting a restart policy on them?

docker-for-azure/upgrade.md
ddebroy: Done

docker-for-azure/upgrade.md
ddebroy: Updated the note. Since `upgrade.sh` is in /usr/bin which is set in PATH, `ash` will always be able to locate it. The full path is not necessary.

docker-for-azure/upgrade.md
ddebroy: Fixed the double spacing. Unfortunately the restart policy doesn't take effect for upgrades since we bring up fresh VMs/docker hosts.

ddebroy:
@mstanleyjones the `docker run` containers are not handled by upgrade since their state in the local disk is completely wiped out when a fresh new VM with a fresh disk is brought up as part of upgrade. All the state around services is retrieved through swarm when the node joins swarm and individual containers are not part of the swarm state.

PR #3940, opened 2017-07-21
Update ubuntu.md

jgeyser:


mstanleyjones:
I prefer the longer form, and experts will already know they can use the shorter form. Thanks, though!

jgeyser:
The id used is also still wrong.

Supposed to be 6D085F96

jgeyser:
@mstanleyjones ^^

mstanleyjones:
Please double-check again, I am pretty sure it's correct: https://docs.docker.com/engine/installation/linux/docker-ee/ubuntu/#install-using-the-repository

Note that the key for Docker EE is different from the one for Docker CE. It is likely that you actually have both keys installed, if you started out with Docker CE. You can use `apt-key list` to show all keys.

PR #3941, opened 2017-07-21
Update suse.md

cherylq26:


mstanleyjones:
This is already being addressed in #3977. Thanks!

PR #3942, opened 2017-07-21
Update ubuntu.md

jgeyser:


jgeyser:
apt-key command was wrong and pointed to wrong id
/ubuntu needed when adding repositories

mstanleyjones:
Thanks, but these are being addressed as part of bigger changes in #3977.

PR #3943, opened 2017-07-21
Update rhel.md

raviupad:


mstanleyjones:
Thanks, but this change is being incorporated as part of bigger changes in #3977.

PR #3945, opened 2017-07-21
Add docs for new docker for aws ELB HTTPS protocol feature

kencochrane:
### Proposed changes

I added some documentation for a new feature that is added to docker for aws 17.07.0 related to the AWS ELB

Signed-off-by: Ken Cochrane <kencochrane@gmail.com>

/cc @FrenchBen @ddebroy @friism

### DIFF
--- a/docker-for-aws/load-balancer.md
+++ b/docker-for-aws/load-balancer.md
@@ -95,13 +95,11 @@ $ docker service create \
 
 ### HTTPS vs SSL load balancer protocols
 
-The elastic load balancer allows you to configure which protocol your listener will support. Originally, only `TCP` and `SSL` was supported. This was limiting because when you use `TCP` or `SSL` it doesn't include the HTTP headers that most applications use to determine the requestor's original IP address. `TCP` and `SSL` does support the [Proxy Protocol](http://docs.aws.amazon.com/elasticloadbalancing/latest/classic/enable-proxy-protocol.html#proxy-protocol), but most applications aren't setup to support it.
+Docker for AWS version 17.07.0 and later also support the `HTTPS` listener protocol when using ACM certificates.
 
-Starting with `Docker for AWS 17.07.0` it is now possible to specify if you want `HTTPS` or `SSL` as the load balancer protocol when using an ACM certificate. If you do not specify the protocol it will default to `SSL (TCP)` so that it is backwards compatible with older versions. The only valid options are `HTTPS` or `SSL`, if you put any other value, it will default to `SSL`.
+Use the `HTTPS` protocol if your app relies on checking the [X-Forwarded-For](http://docs.aws.amazon.com/elasticloadbalancing/latest/classic/x-forwarded-headers.html) header for resolving the client IP address. Note that the client IP is also available with `SSL` by using the [Proxy Protocol](http://docs.aws.amazon.com/elasticloadbalancing/latest/classic/enable-proxy-protocol.html#proxy-protocol), but many apps and app frameworks don't support this.
 
-Using `HTTPS` will send along some extra headers including the IP address of the client in the [X-Forwarded-For](http://docs.aws.amazon.com/elasticloadbalancing/latest/classic/x-forwarded-headers.html) header.
-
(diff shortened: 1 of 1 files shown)


docker-for-aws/load-balancer.md
```diff
@@ -93,14 +93,55 @@ $ docker service create \
   yourname/your-image:latest
 ```
 
+### HTTPS vs SSL load balancer protocols
+
+The elastic load balancer allows you to configure which protocol your listener will support. Originally, only `TCP` and `SSL` was supported. This was limiting because when you use `TCP` or `SSL` it doesn't include the HTTP headers that most applications use to determine the requestor's original IP address. `TCP` and `SSL` does support the [Proxy Protocol](http://docs.aws.amazon.com/elasticloadbalancing/latest/classic/enable-proxy-protocol.html#proxy-protocol), but most applications aren't setup to support it.
```
friism: I don't think the docs should have the history of the product embedded, but it's great we cover the trade-offs between the two options and proxy-protocol stuff.

docker-for-aws/load-balancer.md
```diff
@@ -93,14 +93,55 @@ $ docker service create \
...
+
+Starting with `Docker for AWS 17.07.0` it is now possible to specify if you want `HTTPS` or `SSL` as the load balancer protocol when using an ACM certificate. If you do not specify the protocol it will default to `SSL (TCP)` so that it is backwards compatible with older versions. The only valid options are `HTTPS` or `SSL`, if you put any other value, it will default to `SSL`.
```
friism: I think ideally we should error on invalid input. A user (not having read these docs) providing invalid input and getting back `SSL` might surmise that there's a bug in the product.

kencochrane: The problem is there is no place to show the error, it is in the l4controller on the master node so it is hidden. Better to default to something and still work, then break silently and be confusing for the person.

friism: ok, I see - +1

docker-for-aws/load-balancer.md
kencochrane: Open to suggestions, but I feel it is important to let people know how it used to work and how it works now, so they are not confused with the change. We can remove this once it becomes a standard feature in a stable release.

docker-for-aws/load-balancer.md
```diff
@@ -93,14 +93,55 @@ $ docker service create \
...
+com.docker.aws.lb.arn="arn:...@HTTPS:443"
+```
+
+#### A SSL (TCP) listener on port 443
+
+```none
+com.docker.aws.lb.arn="arn:...@443"
+```
+
+#### A SSL (TCP) listener on port 443
```
FrenchBen: Move with the above header, as an equivalent to `@SSL:443`

docker-for-aws/load-balancer.md
friism:
```AWS Elastic Load Balancer supports different kind of listener protocols. Docker for AWS defaults to using the SSL (TCP) protocol.
Docker for AWS version 17.07.0 and later also support the HTTPS listener protocol when using ACM certificates.

Use the HTTPS protocol if your app relies on checking the `X-Forwarded-For` header for resolving the client IP address. Note that the client IP is also available with `SSL` by using the [Proxy Protocol](http://docs.aws.amazon.com/elasticloadbalancing/latest/classic/enable-proxy-protocol.html#proxy-protocol), but many apps and app frameworks don't support this.

The only valid options are `HTTPS` and `SSL`. Specifying any other value will cause `SSL` to be selected. For backwards compatibility the default protocol is `SSL`.
```

kencochrane: thanks

kencochrane:
@friism better?

friism:
lgtm!

mstanleyjones:
This relies upon a version of Docker that is not yet released, and needs to be based on the `vnext-engine` branch. Give me a few minutes to make sure that branch is ready, though.

PR #3954, opened 2017-07-24
Fix type

bolshakov:
### Proposed changes

Just fixed small typo -- must instead of mut.


johndmulhausen:
Thanks, @bolshakov :)

PR #3958, opened 2017-07-24
Update netlify URL's for CLI reference and API

thaJeztah:
Please verify if I didn't make copy/paste mistakes here 😅 

ping @mstanleyjones @johndmulhausen PTAL

### DIFF
--- a/netlify.sh
+++ b/netlify.sh
@@ -4,33 +4,36 @@
 # Replicates the non-archive functionality of the Dockerfile
 
 # First build master
-ENGINE_SVN_BRANCH="branches/17.03.x"
-ENGINE_BRANCH="17.03.x"
+ENGINE_SVN_BRANCH="branches/17.06"
+ENGINE_BRANCH="17.06"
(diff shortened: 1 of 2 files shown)


netlify.sh
```diff
@@ -4,33 +4,36 @@
 # Replicates the non-archive functionality of the Dockerfile
 
 # First build master
-ENGINE_SVN_BRANCH="branches/17.03.x"
-ENGINE_BRANCH="17.03.x"
+ENGINE_SVN_BRANCH="branches/17.06"
```
mstanleyjones: Not sure what this new file is for, can you explain it?

thaJeztah: This is not something new, only changed 17.03.x to 17.06

johndmulhausen:
Just checked and `netlify.sh` doesn't run in Netlify itself, so we should delete it. We *could have* made Netlify use this script when creating previews, but that would create a situation where Netlify was doing all the upstream work on every PR. Given that we are having timeout problems as it is, I'm probably going to just delete `netlify.sh`

PR #3959, opened 2017-07-24
updated key concept

AlicanAkkus:
docker service, will run on the manager or worker. Last file contains only worker.

### DIFF
--- a/engine/swarm/key-concepts.md
+++ b/engine/swarm/key-concepts.md
@@ -44,7 +44,7 @@ worker.
 
 ## Services and tasks
 
-A **service** is the definition of the tasks to execute on the worker nodes. It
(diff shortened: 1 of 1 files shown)


mstanleyjones:
@aaronlehmann do you think this is accurate and worth doing? It seems to me that a node will only run services as part of its worker capacity, even if it happens to also be a manager node.

aaronlehmann:
SGTM, I think the change is good because it avoids the implication that services only run on worker (as opposed to manager) nodes.

We might want to separately explain elsewhere that managers do everything that workers do, including running service tasks, if we don't make that clear already.

mstanleyjones:
Yep, we do, right above that section!

PR #3962, opened 2017-07-24
fixed linktext

pvnovarese:
the link to CE instructions said "CE for CentOS" even though the link pointed to the Ubuntu page.  Also "Docker" in that link text was not capitalized.


johndmulhausen:
Thanks, @pvnovarese :)

PR #3965, opened 2017-07-24
Remove v2-registry-auth image

mstanleyjones:
Put it back upstream instead

Depends upon https://github.com/docker/distribution/pull/2345 being merged and cherry-picked into the current distribution release.

PTAL @stevvooe


stevvooe:
LGTM

PR #3985, opened 2017-07-26
Adding compose Variable substitution

dark0dave:
Missing Documentation.

### Proposed changes

Adding compose Variable substitution documentation from docker compose git.

### DIFF
--- a/compose/compose-file/compose-file-v2.md
+++ b/compose/compose-file/compose-file-v2.md
@@ -1269,7 +1269,43 @@ refer to it within the Compose file:
 
 ## Variable substitution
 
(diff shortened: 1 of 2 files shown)


compose/compose-file/compose-file-v2.md
```diff
@@ -1269,7 +1269,40 @@ refer to it within the Compose file:
 
 ## Variable substitution
 
-{% include content/compose-var-sub.md %}
```
mstanleyjones: Not sure about removing included content and adding non-included content here. Can you explain it?

dark0dave: File does not exist....

PR #3988, opened 2017-07-26
Environment setup is needed before SSH works on Windows 10 using HyperV

willharrison:
### Proposed changes

I've been going through the docs today trying to get everything set up and was unable to ssh into my docker machine. This command fixes the issue.

### DIFF
--- a/get-started/part4.md
+++ b/get-started/part4.md
@@ -71,7 +71,14 @@ You now have two VMs created, named `myvm1` and `myvm2` (as `docker-machine ls`
 shows). The first one will act as the manager, which executes `docker` commands
 and authenticates workers to join the swarm, and the second will be a worker.
 
-You can send commands to your VMs using `docker-machine ssh`. Instruct `myvm1`
+Before sending commands to these VMs, you must set up the environment for the Docker client.
+
+```shell
+$ docker-machine env myvm1
+$ docker-machine env myvm2
+```
+
+Now you can send commands to your VMs using `docker-machine ssh`. Instruct `myvm1`
 to become a swarm manager with `docker swarm init` and you'll see output like
 this:
 


mstanleyjones:
@shin- PTAL

shin-:
That shouldn't actually be necessary.

londoncalling:
@willharrison , agree with @shin- , it is not necessary to run `docker-machine env <machine-name>` in order to send `ssh` commands directly to the machine. I can see how you might think this, but they are two different paradigms or ways of interacting with the virtual machine.

In the first case (where you run the `docker-machine env <machine-name>` sequence), you are setting up a Docker specific environment in your  local shell to work as if you are logged into the Docker virtual machine. When you close that shell, that environment disappears.

In the second case (sending `ssh` commands to the docker virtual machine), you are using `ssh`. Alternatively, you _could_ do a direct `ssh` login. But you don't need to set shell environment variables to use SSH. (More on using `ssh` with Docker Machine is in the Docker Machine CLI reference: [docker-machine ssh](https://docs.docker.com/machine/reference/ssh/)

Here, I've created two machines per the Getting Started tutorial:

```
$ docker-machine ls
NAME    ACTIVE   DRIVER       STATE     URL                         SWARM   DOCKER        ERRORS
myvm1   -        virtualbox   Running   tcp://192.168.99.100:2376           v17.06.0-ce   
myvm2   -        virtualbox   Running   tcp://192.168.99.101:2376           v17.06.0-ce   
```

I can send Linux commands via SSH:

```
$ docker-machine ssh myvm1 "pwd"
/home/docker

$ docker-machine ssh myvm1 "ls"
log.log
```

Or log in directly via SSH and type commands:

```
 docker-machine ssh myvm1
                        ##         .
                  ## ## ##        ==
               ## ## ## ## ##    ===
           /"""""""""""""""""\___/ ===
      ~~~ {~~ ~~~~ ~~~ ~~~~ ~~~ ~ /  ===- ~~~
           \______ o           __/
             \    \         __/
              \____\_______/
 _                 _   ____     _            _
| |__   ___   ___ | |_|___ \ __| | ___   ___| | _____ _ __
| '_ \ / _ \ / _ \| __| __) / _` |/ _ \ / __| |/ / _ \ '__|
| |_) | (_) | (_) | |_ / __/ (_| | (_) | (__|   <  __/ |
|_.__/ \___/ \___/ \__|_____\__,_|\___/ \___|_|\_\___|_|
Boot2Docker version 17.06.0-ce, build HEAD : 0672754 - Thu Jun 29 00:06:31 UTC 2017
Docker version 17.06.0-ce, build 02c1d87
docker@myvm1:~$ ls
log.log

```

So, try out just sending the `ssh` commands (or doing the direct `ssh` login) without setting up the `docker-machine` environment variables and see if that works. It should work (regardless of your system OS).

londoncalling:
Closing this for now, because there is nothing to update on the tutorial with regard to this. We _could_ clarify that you don't need to set `docker-machine env <machine-name>` but this may just add more complexity or raise questions that will confuse people.

PR #4029, opened 2017-07-30
part3: `docker stack ps` needs the stack name

fasfsfgs:
### Proposed changes

Adding stack name to the `docker stack ps` command.

When following along the steps at part3, I got stock when trying to do `docker stack ps`. It outputs the following:
```
"docker stack ps" requires exactly 1 argument(s).
See 'docker stack ps --help'.

Usage:  docker stack ps [OPTIONS] STACK

List the tasks in the stack
```


fasfsfgs:
Nvm. It was simply linking to a previous complete command.

Issue #4044, opened 2017-07-31
The prerequisites link is broken: https://docs.docker.com/engine/installation/linux/docker-ee/#prerequisites


apuchitnis:
[engine/installation/linux/docker-ee/ubuntu.md](https://docs.docker.com/engine/installation/linux/docker-ee/ubuntu/), CC @mstanleyjones

londoncalling:
@mstanleyjones thinks this broke when #3981 was merged. @johndmulhausen can you revert #3981, and I will update all links with `#index.md<some-subtopic>` to have full paths instead of relative, if you think this will fix that other problem?

londoncalling:
Fixed in #4051

PR #4055, opened 2017-08-01
fixed links to on-page installs

londoncalling:
### What's changed

In Get Docker install topics, fixed links to on-page install instructions for CE and EE, which required two things:

- match link path to target topic name (e.g., `install-docker-ce` or `install-docker-ee`,  instead of `install-docker`)

- rename duplicate named lower level subtopic, which was causing conflict and breaking the links even when they had the right target topic name (e.g., `Install Docker` instead of `Install Docker CE` or `Install Docker EE`)

### Related

- Marginally related to #4044

### DIFF
--- a/engine/installation/linux/docker-ce/centos.md
+++ b/engine/installation/linux/docker-ce/centos.md
@@ -114,7 +114,7 @@ from the repository.
 
     [Learn about **stable** and **edge** builds](/engine/installation/).
 
-#### Install Docker
+#### Install Docker CE
 
 1.  Update the `yum` package index.
 
--- a/engine/installation/linux/docker-ce/centos.md
+++ b/engine/installation/linux/docker-ce/centos.md
@@ -10,7 +10,7 @@ title: Get Docker CE for CentOS
 
 To get started with Docker CE on CentOS, make sure you
 [meet the prerequisites](#prerequisites), then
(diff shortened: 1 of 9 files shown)


engine/installation/linux/docker-ce/centos.md
```diff
@@ -114,7 +114,7 @@ from the repository.
 
     [Learn about **stable** and **edge** builds](/engine/installation/).
 
-#### Install Docker CE
+#### Install Docker
```
mstanleyjones: This change conflicts with the  change in line 13, I think.

londoncalling:
@mstanleyjones the problem is that there are two titles on the page with the same name, so this will prevent any links to either of those titles from working. Should we just not link to those subtopics? Or we can use anchor tags. I'll try that.

mstanleyjones:
No, I think the link should be to for instance, https://docs.docker.com/engine/installation/linux/docker-ee/ubuntu/#install-docker-ee

(try it in your browser, it will work)

londoncalling:
@mstanleyjones okay, this seems to be working fine now. I don't know why the links seemed not to work before. Please double check on Netlify and also make sure I reverted the subtopic names back properly. I think I did, double-checking again. Thanks!

PR #4065, opened 2017-08-01
Expand on meaning of support for Docker CE

### LINKED ISSUE #3905: Clarify what "supported" means for storage drivers
The documentation on storage drivers has a table indicating which storage drivers are supported (it's under the "Supported storage drivers per Linux distribution" header).

In my humble opinion, it is a bit confusing, because it doesn't say what "supported" means. All drivers are available on all platforms (as far as I know), but they are not _recommended_ (right?). For instance, we don't _support_ `btrfs` on any platform on Docker CE; but people can still use it (it's still compiled in). So, what does "supported" mean? That we provide support? Not really, since that's Docker CE – we don't have commercial support on it.

Tentatively, I think I'd go for a paragraph like the following one, below the compatibility table:

_Keep in mind that "supported storage driver" means that this distro/driver combination is in our automated regression tests, is used by a large number of users, and that if you find a reproducible issue, it will most likely be fixed very quickly. If the driver that you want to use is not "supported" according to this table, it doesn't mean that it's not available. You can run it at your own risk. You can even (and are encouraged to!) report issues if you find any. But for technical reasons, it will be harder for us to help you with the "unsupported" combinations. Yes, this means that `btrfs` is only "suppoted" in Docker EE, for instance on SLES. But you can still run it on your favorite distro if you need to!_

[engine/userguide/storagedriver/selectadriver.md](https://docs.docker.com/engine/userguide/storagedriver/selectadriver/), CC @mstanleyjones

mstanleyjones:
Fixes #3905 

cc/ @jpetazzo @dmp42 @friism

### DIFF
--- a/engine/userguide/storagedriver/selectadriver.md
+++ b/engine/userguide/storagedriver/selectadriver.md
@@ -84,17 +84,32 @@ For Docker CE, only some configurations are tested, and your operating system's
 kernel may not support every storage driver. In general, the following
 configurations work on recent versions of the Linux distribution:
 
-| Linux distribution   | Supported storage drivers            |
-|----------------------|--------------------------------------|
-| Docker CE on Ubuntu  | `aufs`, `devicemapper`, `overlay2` (Ubuntu 14.04.4 or later, 16.04 or later), `overlay`, `zfs` |
-| Docker CE on Debian  | `aufs`, `devicemapper`, `overlay2` (Debian Stretch), `overlay`                                 |
-| Docker CE on CentOS  | `devicemapper`                                                                                 |
-| Docker CE on Fedora  | `devicemapper`, `overlay2` (Fedora 26 or later, experimental), `overlay` (experimental)        |
+| Linux distribution  | Recommended storage drivers                                                                    |
+|:--------------------|:-----------------------------------------------------------------------------------------------|
+| Docker CE on Ubuntu | `aufs`, `devicemapper`, `overlay2` (Ubuntu 14.04.4 or later, 16.04 or later), `overlay`, `zfs` |
+| Docker CE on Debian | `aufs`, `devicemapper`, `overlay2` (Debian Stretch), `overlay`                                 |
+| Docker CE on CentOS | `devicemapper`                                                                                 |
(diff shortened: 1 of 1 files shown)


engine/userguide/storagedriver/selectadriver.md
```diff
@@ -84,18 +84,33 @@ For Docker CE, only some configurations are tested, and your operating system's
...
+> that is available for your platform. For instance, you can use `btrfs` with
+> Docker CE, even though it is not recommended on any platform for Docker CE,
+> and you do so at your own risk.
+>
+> The recommendations in the table above are based on automated regression
+> testing and the configurations that are known to work for a large number of
+> users. If you use a recommended configuration and find a reproducible issue,
+> it is likely to be fixed very quickly. If the driver that you want to use is
+> not recommended according to this table, you can run it at your own risk. You
+> can and should still report any issues you run into. However, for technical
```
friism: > However, for technical
> reasons, it will be more challenging to provide assistance for problems with
> storage drivers we do not recommend for your platform

I would change that to "However, fixes and improvements to non-recommended storage drivers are lower priority"

friism:
lgtm

engine/userguide/storagedriver/selectadriver.md
mstanleyjones: Updated.

jpetazzo:
👍

engine/userguide/storagedriver/selectadriver.md
```diff
@@ -84,18 +84,33 @@ For Docker CE, only some configurations are tested, and your operating system's
...
+> and you do so at your own risk.
+>
+> The recommendations in the table above are based on automated regression
+> testing and the configurations that are known to work for a large number of
+> users. If you use a recommended configuration and find a reproducible issue,
+> it is likely to be fixed very quickly. If the driver that you want to use is
+> not recommended according to this table, you can run it at your own risk. You
+> can and should still report any issues you run into. However, such issues will
+> have a lower priority than issues encountered when using a recommended
+> configuration..
```
jpetazzo: There is a double dot here

mstanleyjones: Thanks, I fixed that as a follow-up commit.

PR #4069, opened 2017-08-01
Clarify that the docker group is created automatically

### LINKED ISSUE #2397: Unclear if docker group is auto-created in post-install (CentOS 7)
[engine/installation/linux/linux-postinstall.md](https://docs.docker.com/engine/installation/linux/linux-postinstall/)

I am researching Docker as part of a class project. While working through the Post-installation steps for Linux, I noticed that the docker group was already present on my CentOS 7 VM. Is the docker group created during the install?

mstanleyjones:
Fixes #2397


mstanleyjones:
Good catch, fixed!

PR #4085, opened 2017-08-03
Update release-notes.md

scjane:
### Proposed changes

Zero-fill numbers to two digits in "release date".


mstanleyjones:
I don't think we have this standard around zero-filling dates. @johndmulhausen WDYT?

johndmulhausen:
We don't really have the standard but it doesn't hurt anything and the sentence casing is an improvement. :) Thanks, @scjane 👍

PR #4089, opened 2017-08-03
Add general topic about pruning objects

### LINKED ISSUE #4000: Document the ability to exclude volumes from pruning
See https://github.com/docker/docker-ce/pull/109

mstanleyjones:
Fixes #4000 

I needed to document that volumes are not automatically pruned by `docker system prune` anymore as of 17.06.1, and I realized we don't have a general topic about pruning objects, so I wrote one. This is against `master` since the behavior change is actually in 17.06.1 and 17.07.0.

### DIFF
--- a/engine/admin/pruning.md
+++ b/engine/admin/pruning.md
@@ -1,0 +1,177 @@
+---
+description: Pruning unused objects
+keywords: pruning, prune, images, volumes, containers, networks, disk, administration, garbage collection
+title: Prune unused Docker objects
+---
+
+Docker takes a conservative approach to cleaning up unused objects (often
+referred to as "garbage collection"), such as images, containers, volumes, and
+networks: these objects are generally not removed unless you explicitly ask
+Docker to do so. This can cause Docker to use extra disk space. For each type of
+object, Docker provides a `prune` command. In addition, you can use `docker
+system prune` to clean up multiple types of objects at once. This topic shows
+how to use these `prune` commands.
+
+## Prune images
+
+The `docker image prune` command allows you to clean up unused images. By
+default, `docker image prune` only cleans up _dangling_ images. A dangling image
+is one that is not tagged and is not referenced by any container. To remove
+dangling images:
+
+```bash
+$ docker image prune
+
+WARNING! This will remove all dangling images.
+Are you sure you want to continue? [y/N] y
+```
+
+To remove all images which are not used by existing containers, use the `-a`
+flag:
+
+```bash
+$ docker image prune -a
+
+WARNING! This will remove all images without at least one container associated to them.
+Are you sure you want to continue? [y/N] y
+```
+
+By default, you are prompted to continue. To bypass the prompt, use the `-f` or
+`--force` flag.
+
+You can limit which images are pruned using filtering expressions with the
+`--filter` flag. For example, to only consider images created more than 24
+hours ago:
+
+```bash
+$ docker image prune -a --filter "until=24h"
+```
+
+Other filtering expressions are available. See the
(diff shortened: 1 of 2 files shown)


engine/admin/pruning.md
```diff
@@ -0,0 +1,187 @@
...
+removed unless you explicitly ask Docker to do so. This can cause Docker to use
+extra disk space. For each type of object, Docker provides a `prune` command.
+In addition, you can use `docker system prune` to clean up multiple types of
+objects at once. This topic shows how to use these `prune` commands.
+
+## Prune images
+
+The `docker image prune` command allows you to clean up unused images. By
+default, `docker image prune` only cleans up _dangling_ images. A dangling
+image is one that is not tagged and is not referenced by any running container.
```
thaJeztah: remove `running` here, stopped containers also count

engine/admin/pruning.md
```diff
@@ -0,0 +1,187 @@
...
+Other filtering expressions are available. See the
+[`docker image prune` reference](/engine/reference/commandline/image_prune.md)
+for more examples.
+
+## Prune containers
+
+When you stop a container, it is not automatically removed unless you started
+it with the `--rm` flag. To see all containers on the Docker host, including
+stopped containers, use `docker ps -a`. You may be surprised how many containers
+exist, especially on a development system! A stopped container's read-only and
```
thaJeztah: the `read-only` layers are not removed (the read-only layers are what makes up an "image")

engine/admin/pruning.md
```diff
@@ -0,0 +1,187 @@
...
+WARNING! This will remove all volumes not used by at least one container.
+Are you sure you want to continue? [y/N] y
+```
+
+By default, you are prompted to continue. To bypass the prompt, use the `-f` or
+`--force` flag.
+
+By default, all unused volumes are removed. You can limit the scope using
+the `--filter` flag. For instance, the following command only removes
+volumes older than 24 hours:
```
thaJeztah: volumes don't (yet) support the `until` filter, you could use an example using labels, e.g. 

prune all volumes that don't have the "keep" label;

```
docker volume prune --filter label!=keep
```

thaJeztah: (the label filter is supported by all object types)

engine/admin/pruning.md
```diff
@@ -0,0 +1,187 @@
...
+
+By default, you are prompted to continue. To bypass the prompt, use the `-f` or
+`--force` flag.
+
+By default, all unused objects are removed. You can limit the scope using
+the `--filter` flag. For instance, the following command only removes
+objects older than 24 hours:
+
+```bash
+$ docker system prune --filter "until=24h"
```
thaJeztah: Use the `label` filter here; the "until" filter cannot be used with volumes, so will produce an error; https://github.com/docker/docker-ce/pull/154

thaJeztah: There's also discussion about the validity of filters on "system prune"; https://github.com/docker/cli/pull/311#issuecomment-314489773 (perhaps we should remove the filtering section here altogether)

engine/admin/pruning.md
```diff
@@ -0,0 +1,187 @@
+---
+description: Pruning unused objects
+keywords: pruning, prune, images, volumes, containers, networks, disk, administration
```
thaJeztah: We should have "garbage collection", "garbage collection" as keyword.

Perhaps even use it in the title ("Garbage collecting - pruning unused Docker objects")

mstanleyjones:
Addressed feedback.

PR #4100, opened 2017-08-04
[compose up] Rewrite exit codes paragraph

shin-:
### Proposed changes

Improved accuracy and fixed grammar

### DIFF
--- a/compose/reference/up.md
+++ b/compose/reference/up.md
@@ -51,6 +51,6 @@ flag.
 If you want to force Compose to stop and recreate all containers, use the
 `--force-recreate` flag.
 
-If the command encounters an error, it returns `1`.
-If user press `ctrl` + `C`, causing stop of the containers, it returns `0`.
-If user press `ctrl` + `C` two time, forcing kill of containers, it returns `2`.
+If the process encounters an error, the exit code for this command is `1`.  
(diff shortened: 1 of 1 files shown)


cjerdonek:
Thanks. I believe `SIGTERM` is [handled the same](https://github.com/docker/compose/blob/f9aaa72c54957ee834141bdf7f7ec9b656f26697/compose/cli/signals.py#L17) as `SIGINT`, so you may want to add that.

cjerdonek:
Also, if the process is interrupted quickly using `SIGINT` (`ctrl` + `C`), the exit code is 1 and not 0.

compose/reference/up.md
```diff
@@ -51,6 +51,6 @@ flag.
 If you want to force Compose to stop and recreate all containers, use the
 `--force-recreate` flag.
 
-If the command encounters an error, it returns `1`.
-If user press `ctrl` + `C`, causing stop of the containers, it returns `0`.
-If user press `ctrl` + `C` two time, forcing kill of containers, it returns `2`.
+If the process encounters an error, the exit code for this command will be `1`.
```
mstanleyjones: Can you replace all the future tense like "will be" with "is"?

shin-:
Updated :+1:

PR #4102, opened 2017-08-05
Convert Library references from submodule to upstream

johndmulhausen:
Pulling in Library refs via shell script and using Jekyll collections will save a lot of grief over using the submodule system. 

To test, run `docker build -t doctest .` and then `docker run -p 4000:4000 -ti doctest`

Netlify can complete builds created with this system; generation time on this PR was 7.35 minutes, well within the timeout range of 15 minutes.

`jekyll build` run locally from a normal authoring clone (which grabs nothing from upstream) is 3.89 minutes.

`docker build` run locally from this branch (which grabs everything from upstream) finishes generation in 5.4 minutes.

This also ensures that our copy of "library" is up to date and kills the dependency between shell files and upstream files, which was always bound to break. The library is very active and now we can just use what they have, whatever that is.


mstanleyjones:
LGTM though I feel like the script could use some improvements like using a [heredoc](http://tldp.org/LDP/abs/html/here-docs.html) instead of all the echoes. Maybe @thaJeztah can provide some input since he's really good at this. Though you may not be able to use the `heredoc` if you are on Alpine and stuck with `/bin/sh`...

johndmulhausen:
@mstanleyjones Let's think about what we want to do here and iterate. Merging since the guts work and we need the infra fix.

johndmulhausen:
Thanks for the feedback!

PR #4115, opened 2017-08-07
Update btrfs-driver.md

thehappyhippo:
Removed useless 1st whole copy process.

### DIFF
--- a/engine/userguide/storagedriver/btrfs-driver.md
+++ b/engine/userguide/storagedriver/btrfs-driver.md
@@ -65,8 +65,8 @@ This procedure is essentially identical on SLES and Ubuntu.
     the contents of `/var/lib/docker/`:
 
     ```bash
-    $ sudo cp -au /var/lib/docker /var/lib/docker.bk
-    $ sudo rm -rf /var/lib/docker/*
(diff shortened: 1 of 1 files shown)


engine/userguide/storagedriver/btrfs-driver.md
```diff
@@ -65,8 +65,8 @@ This procedure is essentially identical on SLES and Ubuntu.
     the contents of `/var/lib/docker/`:
 
     ```bash
-    $ sudo cp -au /var/lib/docker /var/lib/docker.bk
-    $ sudo rm -rf /var/lib/docker/*
+    $ sudo mv /var/lib/docker /var/lib/docker.bk
+    $ sudo mkdir  /var/lib/docker/
```
mstanleyjones: I don't think so, because the new directory won't have the right permissions.

thehappyhippo: The rights depend on the filesystem of the mounted share and not on the mount point. I've done it this way without any problems.

mstanleyjones:
And we've done it the other way without problems. If the user configured their system such that the `docker` user owned the directory, or they are using userns-remapping, your method might not work for them. I appreciate your contribution but I don't think it's the right way to go right now.

PR #4120, opened 2017-08-08
Update part5.md

scjane:


scjane:
Sorry, I made a mistake. This should be "five" instead of "5". The rule described in the documentation is "Write out numbers in body text and titles from one to ten. From 11 on, use numerals." I confused them.
Can I change it back?

mstanleyjones:
Sure, I'll revert it.

mstanleyjones:
Actually I can't revert it, you'll need to create a new PR.

PR #4131, opened 2017-08-09
Update part3.md

scjane:


scjane:
Sorry, I made a mistake. This should be "five" instead of "5". The rule described in the documentation is "Write out numbers in body text and titles from one to ten. From 11 on, use numerals." I confused them.
Can I change it back?

mstanleyjones:
You'll need to make a new PR. Sorry!

PR #4139, opened 2017-08-09
Update on quotactl in default seccomp profile docs

pmoust:
Updated the description of reasons why `quotactl` is blocked by the
default seccomp profile.

Ref: http://man7.org/linux/man-pages/man2/quotactl.2.html

Rel: https://github.com/moby/moby/pull/34445

Signed-off-by: Panagiotis Moustafellos <pmoust@elastic.co>


justincormack:
I don't think we need to document this, the change does not affect the default set of seccomp rules at all, only the rules if `CAP_SYS_ADMIN` is used.

PR #3967, opened 2017-07-25
Better examples to show Compose file structure, expand/collapse accordion

londoncalling:
### What's changed

- added more context to examples for `build` and `volume` sub-options to `service`
- added back in example Compose file at the top of the page to provide overall context
- fixed a link

### Related 

Fixes #1349 
Fixes #3955 

### Reviewers fyi

@shin- @MicheleMannino @Capricorn1

Signed-off-by: Victoria Bialas <victoria.bialas@docker.com>

PR #4142, opened 2017-08-09
Fix Compose file samples for volumes and other

londoncalling:
### What's changed

Updated sample for Volumes per feedback on [#5092](https://github.com/docker/compose/issues/5092)

### Related

Issue [#5092](https://github.com/docker/compose/issues/5092)

PR where some of these mistakes were introduced: #3967

### Reviewers

@shin- @noamcore

### DIFF
--- a/compose/compose-file/index.md
+++ b/compose/compose-file/index.md
@@ -1509,37 +1509,31 @@ files](#volumes-for-services-swarms-and-stack-files).
 > a named volume and references it from each service's `volumes` list. This replaces `volumes_from` in earlier versions of the Compose file format. (See [Docker Volumes](/engine/userguide/dockervolumes.md) and
 [Volume Plugins](/engine/extend/plugins_volume.md) for general information on volumes.)
 
-This example shows a named volume (`db-data`) being used by the `postgres` service, and a mounted volume for a single service (under the `redis` service).
+This example shows a named volume (`db-data`) being used by the `postgres`
+service, and a mounted volume for a single service (under the `redis` service).
 
 ```none
-version: "3"
-
+version: "3.2"
 services:
-
   web:
-    nginx:alpine
+    image: nginx:alpine
     ports:
-    - "80:80"
-
-  postgres:
-    image: postgres:9.4
+      - "80:80"
     volumes:
-      - db-data:/var/lib/db
+      - type: volume
+        source: mydata
+        target: /data
+        volume:
+          nocopy: true
+      - type: bind
+        source: ./static
+        target: /opt/app/static
 
-  backup:
-    image: postgres:9.4
-    volumes:
-      - db-data:/var/lib/backup/data
-
-  redis:
-    image: redis
-    ports:
-      - "6379:6379"
-    volumes:
-      - ./data:/data
+networks:
+  webnet:
 
 volumes:
-  db-data:
+  mydata:
 ```
 
 #### Short syntax
--- a/compose/compose-file/index.md
+++ b/compose/compose-file/index.md
@@ -1506,7 +1506,7 @@ files](#volumes-for-services-swarms-and-stack-files).
 
 > **Note**: The top-level
 > [volumes](#volume-configuration-reference) key defines
-> a named volume and references it from each service's `volumes` list. This replaces `volumes_from` in earlier versions of the Compose file format. See [Manage data in Docker](/engine/admin/volumes/index.md) and [Volume
+> a named volume and references it from each service's `volumes` list. This replaces `volumes_from` in earlier versions of the Compose file format. See [Use Volumes](/engine/admin/volumes/volumes.md) and [Volume
 Plugins](/engine/extend/plugins_volume.md) for general information on volumes.
 
 This example shows a named volume (`mydata`) being used by the `web` service,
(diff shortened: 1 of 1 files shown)


compose/compose-file/index.md
```diff
@@ -1509,37 +1509,31 @@ files](#volumes-for-services-swarms-and-stack-files).
...
+version: "3.2"
 services:
-
   web:
-    nginx:alpine
+    image: nginx:alpine
     ports:
-    - "80:80"
-
-  postgres:
```
mstanleyjones: Why did you take away the postgres service? Or am I confused?

compose/compose-file/index.md
```diff
@@ -1509,37 +1509,31 @@ files](#volumes-for-services-swarms-and-stack-files).
 > a named volume and references it from each service's `volumes` list. This replaces `volumes_from` in earlier versions of the Compose file format. (See [Docker Volumes](/engine/userguide/dockervolumes.md) and
 [Volume Plugins](/engine/extend/plugins_volume.md) for general information on volumes.)
 
-This example shows a named volume (`db-data`) being used by the `postgres` service, and a mounted volume for a single service (under the `redis` service).
+This example shows a named volume (`db-data`) being used by the `postgres`
```
mstanleyjones: This no longer matches the volume name.

compose/compose-file/index.md
```diff
@@ -1506,42 +1506,54 @@ files](#volumes-for-services-swarms-and-stack-files).
...
 > **Note**: The top-level
 > [volumes](#volume-configuration-reference) key defines
-> a named volume and references it from each service's `volumes` list. This replaces `volumes_from` in earlier versions of the Compose file format. (See [Docker Volumes](/engine/userguide/dockervolumes.md) and
-[Volume Plugins](/engine/extend/plugins_volume.md) for general information on volumes.)
+> a named volume and references it from each service's `volumes` list. This replaces `volumes_from` in earlier versions of the Compose file format. See [Use Volumes](/engine/admin/volumes/volumes.md) and [Volume
+Plugins](/engine/extend/plugins_volume.md) for general information on volumes.
 
-This example shows a named volume (`db-data`) being used by the `postgres` service, and a mounted volume for a single service (under the `redis` service).
+This example shows a named volume (`mydata`) being used by the `web` service,
+and a mounted volume for a single service (under the `visualizer` service).
```
mstanleyjones: The second one is a bind mount. It's always a bind mount if the first part (the source) is a path instead of just a name.

shin-:
I think there's potential to make the example shorter so we don't inadvertently bury the relevant information (removing `ports` and `networks` entries). I also added an example of the "old", string format for mounting a named volume with the ` - "dbdata:/var/lib/postgresql/data"` line.

```yaml
version: "3.2"
services:
  web:
    image: nginx:alpine
    volumes:
      - type: volume
        source: mydata
        target: /data
        volume:
          nocopy: true
      - type: bind
        source: ./static
        target: /opt/app/static

  db:
    image: postgres:latest
    volumes:
      - "/var/run/postgres/postgres.sock:/var/run/postgres/postgres.sock"
      - "dbdata:/var/lib/postgresql/data"
volumes:
  mydata:
  dbdata:
```

compose/compose-file/index.md
```diff
@@ -1506,42 +1506,51 @@ files](#volumes-for-services-swarms-and-stack-files).
...
-  db-data:
+  mydata:
 ```
 
+> **Tips**:
+>
+> * The visualizer service in the above example is borrowed from the
+> [Get Started tutorial](/get-started/index,nd) and just used here
+> to demonstrate mounted volumes. It is explained in [Part
+> 5](/get-started/part5.md#add-a-new-service-and-redeploy) of that topic.
```
shin-: With the change I suggested, this point does not apply anymore, I think?

compose/compose-file/index.md
```diff
@@ -1506,42 +1506,51 @@ files](#volumes-for-services-swarms-and-stack-files).
...
+      - "dbdata:/var/lib/postgresql/data"
 
-  redis:
-    image: redis
-    ports:
-      - "6379:6379"
-    volumes:
-      - ./data:/data
+networks:
+  webnet:
```
shin-: Should remove this as it's unused in this example

compose/compose-file/index.md
```diff
@@ -1506,42 +1506,51 @@ files](#volumes-for-services-swarms-and-stack-files).
...
 
 volumes:
-  db-data:
+  mydata:
```
shin-: need an entry for `dbdata` which we use on line 1534

compose/compose-file/index.md
```diff
@@ -1533,21 +1537,12 @@ services:
...
 
-> **Tips**:
->
-> * The visualizer service in the above example is borrowed from the
-> [Get Started tutorial](/get-started/index,nd) and just used here
-> to demonstrate mounted volumes. It is explained in [Part
-> 5](/get-started/part5.md#add-a-new-service-and-redeploy) of that topic.
->
-> * See [Use Volumes](/engine/admin/volumes/volumes.md) and [Volume
+> **Note**: See [Use Volumes](/engine/admin/volumes/volumes.md) and [Volume
```
mstanleyjones: Use sentence case in the link :)

ivan-c:
I think the [Long Syntax section](https://github.com/docker/docker.github.io/blob/master/compose/compose-file/index.md#long-syntax-3) still uses the old examples (array, not mapping)

Kitanotori:
How am I supposed to mount directory from the host as a read-only volume with that syntax?

hoybin:
Is there a solution to this problem?

PR #4145, opened 2017-08-10
Update docker-overview.md

scjane:

### DIFF
--- a/engine/docker-overview.md
+++ b/engine/docker-overview.md
@@ -1,6 +1,6 @@
 ---
 description: Docker explained in depth
-keywords: docker, introduction, documentation, about, technology,  understanding
+keywords: docker, introduction, documentation, about, technology, understanding
 redirect_from:
 - /introduction/understanding-docker/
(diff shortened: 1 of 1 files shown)


engine/docker-overview.md
```diff
@@ -267,8 +267,8 @@ the future, Docker may support other container formats by integrating with
 technologies such as BSD Jails or Solaris Zones.
 
 ## Next steps
-- Read about [Installing Docker Engine](installation/index.md#installation).
-- Get hands-on experience with the [Get Started With Docker](getstarted/index.md)
+- Read about [Install Docker Engine](installation/index.md#installation).
```
mstanleyjones: This should be just "installing Docker" as the link text. We are preferring "Docker" over "Docker Engine" now. You don't need to mirror the actual topic's title exactly if you can fit it into the sentence.

scjane: Thank you. I restored it to “Installing". What about the link text in the following item? Should it also be "Getting started with Docker"?

PR #4177, opened 2017-08-12
Add clarification on env_file / environment interaction

shin-:
### Proposed changes

Add clarification on the interaction between the  `env_file` and `environment` sections

### Related issues (optional)

https://github.com/docker/compose/issues/1171

### DIFF
--- a/compose/compose-file/compose-file-v2.md
+++ b/compose/compose-file/compose-file-v2.md
@@ -352,8 +352,9 @@ Add environment variables from a file. Can be a single value or a list.
 If you have specified a Compose file with `docker-compose -f FILE`, paths in
 `env_file` are relative to the directory that file is in.
 
-Environment variables specified in [environment](#environment) _override_
-these values.
+Environment variables declared in the [environment](#environment) section
+_override_ these values &ndash; this holds true even if those values are
+empty or undefined.
 
     env_file: .env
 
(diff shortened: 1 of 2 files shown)


jejacks0n:
I know I'm dragging up an old commit here, but I'm curious if anybody can clarify why environment takes precedence over the file instead of the inverse? Before opening an issue I figured I'd ask so I can understand it better.

mistyhacks:
It's a common UNIX pattern to override behavior by setting a temporary environment variable, since it goes away when the session goes away (such as when you close your terminal). You can also set an environment variable that will only be used for the duration of a command, like:

```
PATH="/opt/bin:$PATH" which weirdcommand
```

This will run the `which` command with the custom `$PATH` variable, and then unset it immediately. Maybe that helps to understand the way that UNIX people tend to think about what the order of precedence should be.

jejacks0n:
Thanks @mistyhacks! I don't fully understand why that would influence the order, but I understand the use case described. I also understand this isn't the commit that did anything other than describe it. =)

I wanted to allow devs to setup different env vars, and the most elegant way would be to "gitignore" the `env_file` and define the defaults in the `environment` (in this case it's in a docker-compose.yml). This commit just clarifies how it works, but it's a bit of a bummer that what I describe isn't possible because of the order.

I'm not sure I fully understand, but I just wanted to explain why I asked. Thank you again for answering.

PR #4179, opened 2017-08-12
Fix credential-spec documentation

thaJeztah:
The values that are provided in the docker-compose file should be relative, not absolute.

ping @friism @vsaraswat PTAL


friism:
Thanks @thaJeztah ❤️ 

cc @brandonRoyal

PR #4191, opened 2017-08-14
Update get-started-macvlan.md

scjane:


johndmulhausen:
Yes indeed, @scjane :)

PR #4205, opened 2017-08-15
Update dotnetcore to 2.0

kendrahavens:
### Proposed changes

Update suggested Dockerfile to .NET 2.0 images


kendrahavens:
@mstanleyjones We've had a major version release moving .NET Core to 2.0 so I've updated the images referenced here.

kendrahavens:
Thank you!

PR #4227, opened 2017-08-16
Add known issue to DTR

joaofnfernandes:
/cc @vikstrous for visibility


vikstrous:
That's not exactly accurate... It's very specifically when you hit next on the users page, then switch tabs and then go back. @kyhy please confirm

kyhy:
Both of you are correct. Simple description might be more useful for the user.

joaofnfernandes:
Does the problem go away if the user refreshes the page?

vikstrous:
yes

PR #4245, opened 2017-08-17
use Docker as canonical source

friism:
cc @patricklang


friism:
Closing, re-opening with fixed update too

PR #4247, opened 2017-08-17
Update docker-ee.md

friism:
use docker inc for package source and make update more succinct.


friism:
I tested this on new machine, including upgrading from 17.03 to 17.06

PR #739, opened 2016-11-29
Fix access denied

Varentsov:
Fix error: `-bash: /etc/bash_completion.d/docker-compose: access denied`

PR #4248, opened 2017-08-17
compose - fix permissions issue for curl

and800:
### Proposed changes

Removed output redirections from curl commands, which used to cause permission issues, even if used
with sudo.
Prepended sudo to these cmds - this doesn't lower guide's generality, sudo is already used there anyway


and800:
see #739 as similar fix

PR #4254, opened 2017-08-17
Document problem with Oracle Linux and selinux

mstanleyjones:
cc/ @friism @andrewhsu


Djelibeybi:
I'm not sure this is going to resolve your issue, though. Whether or not SElinux is enabled won't change the package dependencies. Yum doesn't check for SElinux at all during its run.

PR #4255, opened 2017-08-17
Update Compose quick-start to show static loaded image, then volume mount

### LINKED ISSUE #4176: The quick start guide results in weird code placement [x-post docker/compose]
[compose/gettingstarted.md](https://docs.docker.com/compose/gettingstarted/), CC @londoncalling

Originally docker/compose#1154

> **Problem**
>
> On Quick Start section under Create a Docker image, the Dockerfile copies over the current directory into the image and the Docker-compose.yml file mounts the current directory inside of the container, both actions result in code being under /code directory. For beginners this example makes more complicated to understand how the data got there.
>
> **Solution**
>
> Break up the example into two parts, explicitly showing how to both actions end up having the /code files inside of the running container.

londoncalling:
### What's changed

- Reorganized Compose Getting Started to first show building and running a static image, then in a different step, updating the Compose file to add a host mount as a volume

### Related

- Fixes #4176

- Original issue [docker/compose#1154](https://github.com/docker/compose/issues/1154)

### Reviewers

@shin- @crosbymichael @dnephin 


compose/gettingstarted.md
```diff
@@ -98,73 +96,142 @@ This Compose file defines two services, `web` and `redis`. The web service:
...
+    redis_1  | 1:M 17 Aug 22:11:10.483 # Server initialized
+    redis_1  | 1:M 17 Aug 22:11:10.483 # WARNING you have Transparent Huge Pages (THP) support enabled in your kernel. This will create latency and memory usage issues with Redis. To fix this issue run the command 'echo never > /sys/kernel/mm/transparent_hugepage/enabled' as root, and add it to your /etc/rc.local in order to retain the setting after a reboot. Redis must be restarted after THP is disabled.
+    web_1    |  * Debugger PIN: 330-787-903
+    redis_1  | 1:M 17 Aug 22:11:10.483 * Ready to accept connections
+    ```
+
+    Compose pulls a Redis image, builds an image for your code, and start the
+    services you defined. In this case, the code is statically copied into the image at build time.
+
+2.  Enter `http://0.0.0.0:5000/` in a browser to see the application running.
```
joaofnfernandes: Is there any reason to list the `http://0.0.0.0:5000` option first? I think it's more intuitive if we talked about `http://localhost:5000` first.

compose/gettingstarted.md
```diff
@@ -98,73 +96,142 @@ This Compose file defines two services, `web` and `redis`. The web service:
...
+    If you're using Docker on Linux natively, then the web app should
+    now be listening on port 5000 on your Docker daemon host.
+    If `http://0.0.0.0:5000` doesn't resolve, you can also
+    try `http://localhost:5000`.
+
+    If you're using Docker Machine on a Mac, use  
+    `docker-machine ip MACHINE_VM` to get the IP address of your
+    Docker host. Then, open `http://MACHINE_VM_IP:5000` in a browser.
+
+    You should see a message in your browser saying:
```
joaofnfernandes: A screenshot of the browser would be cool!

compose/gettingstarted.md
```diff
@@ -98,73 +96,142 @@ This Compose file defines two services, `web` and `redis`. The web service:
...
 
-        $ docker-compose up
-        Pulling image redis...
-        Building web...
-        Starting composetest_redis_1...
-        Starting composetest_web_1...
-        redis_1 | [8] 02 Jan 18:43:35.576 # Server started, Redis version 2.8.3
-        web_1   |  * Running on http://0.0.0.0:5000/
-        web_1   |  * Restarting with stat
+4.  Switch to another terminal window, and type `docker image ls` to
```
joaofnfernandes: I didn't try this myself, but do I really need to open another terminal window? Won't `docker-compose up` run in the background and allow me to run other commands?

compose/gettingstarted.md
```diff
@@ -98,73 +96,142 @@ This Compose file defines two services, `web` and `redis`. The web service:
...
-   browser.
+5.  Stop the application, either by running `docker-compose down` from
+within your project directory in the second terminal, or CTRL+C in the
+original terminal where you started the app.
 
-   You should see a message in your browser saying:
+## Step 5: Edit the Compose file to add a host mount as a volume
 
-   `Hello World! I have been seen 1 times.`
+Edit `docker-compose.yml` in your project directory to add a volume mount for the `web` service:
```
joaofnfernandes: For consistency instead of "volume mount", consider using "bind mount".
@mstanleyjones created https://docs.docker.com/engine/admin/volumes/ that makes clear the difference between bind mounts and volumes. If we use the terms consistently people will understand these concepts faster.

compose/gettingstarted.md
```diff
@@ -98,73 +96,142 @@ This Compose file defines two services, `web` and `redis`. The web service:
...
-containers](../engine/tutorials/dockervolumes.md).
-
 ## Step 4: Build and run your app with Compose
 
-1. From your project directory, start up your application.
+1.  From your project directory, start up your application.
+
+    ```
+    $ docker-compose up
+    Creating network "composetest_default" with the default driver
```
mstanleyjones: Please add an extra line between the input and output

compose/gettingstarted.md
```diff
@@ -98,73 +96,142 @@ This Compose file defines two services, `web` and `redis`. The web service:
...
+    web_1    |  * Debugger PIN: 330-787-903
+    redis_1  | 1:M 17 Aug 22:11:10.483 * Ready to accept connections
+    ```
+
+    Compose pulls a Redis image, builds an image for your code, and start the
+    services you defined. In this case, the code is statically copied into the image at build time.
+
+2.  Enter `http://0.0.0.0:5000/` in a browser to see the application running.
+
+    If you're using Docker on Linux natively, then the web app should
```
mstanleyjones: or using Docker for Mac or Docker for Windows (right?)

compose/gettingstarted.md
londoncalling: I'm not sure, that's the way they had it. Yeah I'll change that.

compose/gettingstarted.md
londoncalling: @joaofnfernandes it won't run it in the background without the `-d` option, which we show them later. A few of the tutorials are written like this, maybe I'll have a look and introduce `-d` sooner than was originally done.

compose/gettingstarted.md
londoncalling: @mstanleyjones the extra line doesn't exist on the terminal .. I can add it, but it isn't representative of what the user will see. Leaving for now, we can chat about it tomorrow. Samples in Compose, d4mac, d4win, Machine show the input / output as it actually displays.

londoncalling:
@shin- I think this is about ready to merge, but would like you to have a look and make sure it's okay with you now

shin-:
:+1: Thank you, looks great

PR #4291, opened 2017-08-18
Update docker commands

gbarr01:
Updated several commands such as `docker ps` to `docker container ls`, `docker rmi` to `docker image rm <image id>`, and so on. Also, added the command to remove an image locally (under **Pull and run the image from the remote repository**) so that the output makes sense.

### DIFF
--- a/get-started/part2.md
+++ b/get-started/part2.md
@@ -213,11 +213,11 @@ docker run -d -p 4000:80 friendlyhello
 
 You get the long container ID for your app and then are kicked back to your
 terminal. Your container is running in the background. You can also see the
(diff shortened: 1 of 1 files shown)


get-started/part2.md
```diff
@@ -318,6 +318,10 @@ If the image isn't available locally on the machine, Docker will pull it from
 the repository.
 
 ```shell
+docker image rm <image id>
```
mstanleyjones: Worth stating that you can pass it an image name or image:tag?

gbarr01: I'm not sure it's worth redoing the video. There are only 2 commands that could be updated: (1) docker ps [1:50] and docker stop <hash> [1:56].

PR #4294, opened 2017-08-19
Add information on how install cloudstor for Azure

### LINKED ISSUE #4266: Docker volume with Azure
### Problem description

The documentation page for share volume with cloudstor on azure may have more informations, like how to configure the Account storage on Azure and what is supported or not.

### Problem location

- I couldn't find the information I wanted. I expected to find it near the following URL: https://docs.docker.com/docker-for-azure/persistent-data-volumes

I testing on CoreOS alpha with docker 17.05.0-ce and docker4x/cloudstor:17.05.0-ce-azure2 and 

### Suggestions for a fix

We can add a part about how to install the plugin, how to create the storage account, add a note about `Secure transfer required` option not supported on linux.

jmaitrehenry:
### Proposed changes

Add how to install and configure the plugin

### Related issues (optional)

Fixes #4266

I could add a Note saying the share volume is not supporting some file lock for now and a SQLite database will fail to acquire locks and is not usable.


mstanleyjones:
@ddebroy PTAL

ddebroy:
Thanks for the PR with the screen shot @jmaitrehenry 

I don't think the Docker4Azure(or Docker4AWS) sections is the right place for out-of-band cloudstor installation documentation. The cloud editions already come with cloudstor pre-installed and the steps here are not necessary for users. Perhaps we can relocate independent cloudstor installation instructions to an independent area.

We also need a note and pointer to something along the lines of `swarm-exec` to make sure users get the plugin installed on all nodes of the swarm rather than just on a single node.

+ @friism

ddebroy:
@mstanleyjones Do you think we can add the above info in a fresh page under https://docs.docker.com/engine/extend/plugins_volume/ ?

mstanleyjones:
@ddebroy nope, I don't think so, because that's about V1 plugins, right? That content lives in the `docker/cli` repo, as well. We just pull it in from there.

ddebroy:
Good point .. how about here then: https://docs.docker.com/engine/extend/#installing-and-using-a-plugin . Can that section be restructured to also have a pointer to `Cloudstor` along with `sshfs`?

mstanleyjones:
Maybe **instead of** the `sshfs` plugin, since I think that one is just an unofficial hack. WDYT?

ddebroy:
That is fine but note that we do not have an immediate plan to open source `cloudstor`. So just for installation demonstration (especially with parameters), we can point to `cloudstor` installation examples instead of `sshfs`. But creation examples need to be based on something open source.

jmaitrehenry:
@mstanleyjones, @ddebroy I could create a new PR for that and close this one if you want, just let me know if I can go ahead with this issue.

mstanleyjones:
Yes, if you could create a new PR with an example not using `cloudstor`, that would be great. If I am understanding correctly. I'll close this one for now.

PR #4308, opened 2017-08-21
Fix broken link

### LINKED ISSUE #4305: 404 at /datacenter/dtr/2.3/admin/upgrade/
URL: /datacenter/dtr/2.3/admin/upgrade/

There are no instructions for upgrading to DTR 2.3.

joaofnfernandes:
Fixes #4305


joaofnfernandes:
@mstanleyjones don't forget that for UCP and DTR we're doing versioning by copy-pasting 😥
If we use an absolute path, once we start working on the next version, we would have to comb all the articles and fix their links.

PR #4312, opened 2017-08-21
Add develop section, reorganize TOC (WIP)

mstanleyjones:
@tiborvass PTAL, this is for the unified API / SDK

This adds a top-level "Develop with Docker" section to the TOC, with two sections so far: one targeted at Dockerizing apps and one targeted at users of the API / SDK. We can probably fatten up the first category using existing examples.


johndmulhausen:
New TOC

```
- Get Docker
- Get started
- MTA
- Runtime
- Networking
- Storage
- Develop with Docker
  - Develop with the Docker API
  - Develop apps with Docker
- Security
  - Standards and Compliance
- Run your app in production
- Open source at Docker
  - Contributing to docs
```

cpuguy83:
In "Overview" there seems to be a quickstart section, and then there's a "Quickstart" page.

I wonder if we could have an example to pull it all together.
Like fetching trust data from notary and verifying an image and then using that to do a `docker run` equivalent.

develop/api/index.md
```diff
@@ -0,0 +1,223 @@
...
+## Available APIs and SDKs
+
+The following table links you straight to the API and SDK documentation for each
+Docker component with SDK or API coverage.
+
+| Component       | HTTP API | Python SDK | Go SDK |
+|:----------------|:---------|:-----------|:-------|
+| Moby core       |          |            |        |
+| Docker registry |          |            |        |
+| Notary          |          |            |        |
```
dnephin: empty table, I guess this will be filled in before merge?

develop/api/index.md
```diff
@@ -0,0 +1,223 @@
...
+scale Docker apps and solutions quickly and easily, no matter what languages or
+infrastructures your organization uses. The API and SDKs provide programmatic
+access to core Docker functions (previously Docker Engine), as well as Docker
+Registry and Notary. APIs and SDKs for more components will be added in the
+future.
+
+The Docker API is a RESTful API accessed by an HTTP client such as `wget` or
+`curl`, or the HTTP library which is part of most modern programming languages.
+Where possible, SDKs are provided for Python and Go developers, for even more
+convenience than using the API directly.
```
dnephin: I would drop this last sentence. I think it's better explained by the following section and "where possible" isn't really meaningful. It's possible in all cases, we just haven't done the work yet.

develop/api/index.md
mstanleyjones: @tiborvass will help me with this.

develop/api/index.md
```diff
@@ -0,0 +1,223 @@
...
+| Docker registry |          |            |        |
+| Notary          |          |            |        |
+
+## Install the SDK
+
+To install the SDK, use one of the following commands:
+
+| SDK    | Installation command                              |
+|:-------|:--------------------------------------------------|
+| Go     | `dep ensure --add golang.docker.io/docker@v1.0.0` |
```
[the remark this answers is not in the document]
mstanleyjones: This is just what we had before, Tibor is going to update this with whatever the new way is.

develop/api/index.md
mstanleyjones: This was to try to capture how Tibor pitched this to me. We can work on wording together.

mstanleyjones:
@cpuguy83 we definitely can. For the first cut we were trying to reuse as much existing material as possible. We certainly don't want to go backward in user experience, compared to what we have now.

mdlinville:
Nope, this PR is years old and it's been a long time since I contributed to this code base.

PR #4315, opened 2017-08-21
Update instructions for Raspberry Pi

### LINKED ISSUE #4292: Raspberry Pi Stretch -
[engine/installation/linux/docker-ce/debian.md](https://docs.docker.com/engine/installation/linux/docker-ce/debian/), CC @mstanleyjones

The recommended "Install on Raspbian (Raspberry Pi)" script (curl -sSL https://get.docker.com | sh) isn't updated yet for the recent Raspbian Stretch image (Aug. 17). The "Install using the repository" steps work and the note could be updated to reflect that.

mstanleyjones:
Fixes #4292

Update instructions for Raspberry Pi to account for Stretch, consolidate instructions to use the recommended packages, a little light reorganization of the topic.

cc/ @andrewhsu  @shin-

### DIFF
--- a/engine/installation/linux/docker-ce/debian.md
+++ b/engine/installation/linux/docker-ce/debian.md
@@ -69,7 +69,7 @@ You can install Docker CE in different ways, depending on your needs:
 - Most users
   [set up Docker's repositories](#install-using-the-repository) and install
   from them, for ease of installation and upgrade tasks. This is the
-  recommended approach.
+  recommended approach, except for Raspbian.
 
 - Some users download the DEB package and
   [install it manually](#install-from-a-package) and manage
@@ -78,6 +78,7 @@ You can install Docker CE in different ways, depending on your needs:
 
 - In testing and development environments, some users choose to use automated
   [convenience scripts](#install-using-the-convenience-script) to install Docker.
+  This is currently the only approach for Raspbian.
 
 ### Install using the repository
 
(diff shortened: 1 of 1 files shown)


gbarr01:
@mstanleyjones, why don't we use a version number for Debian Stretch? For example, `Stretch 9.1 (stable) / Raspbian Stretch`.

mstanleyjones:
@gbarr01 I asked about this and it's just Debian convention. I think that Stretch is maybe the whole `9` line. See https://www.debian.org/releases/.

gbarr01:
@mstanleyjones , I asked about the convention because our entries for Jessie and Wheezy have version numbers.

mstanleyjones:
@seemethere PTAL

seemethere:
So unfortunately for raspbian, the latest release of 17.07.0-ce introduced a bug highlighted in docker/swarmkit#2294 which led us to using the installation script to install an older version of Docker (17.05.x-ce) that does not have this bug.

So installation for Raspbian proper is really only supported in the get.docker.com / test.docker.com script. 

Also, the base url should be different for debian / raspbian.

Debian: https://download.docker.com/linux/debian
Raspbian: https://download.docker.com/linux/raspbian

My suggestion for right now would be to say that installation for raspbian is only currently supported through the convenience script and leave at that until we can fix the swarmkit bug.

seemethere:
Also I guess I can't make a comment on lines that haven't changed but this line:

`{% assign download-url-base = "https://download.docker.com/linux/debian" %}`

Should probably be changed to

`{% assign download-url-base = "https://download.docker.com/linux/$(. /etc/os-release | echo "$ID")" %}`

So that it can accommodate both `raspbian` and `debian`.

mstanleyjones:
@seemethere That's a Liquid (docs-build-time) variable, and doesn't have access to what is on the user's Docker host...

seemethere:
@mstanleyjones I understand that, but it'd be fine to leave that in since it is a shell command that can be run on the users system and is platform agnostic.

It's actually what is used in the get.docker.com install script as well.

So using it in that format should work no matter what:

Example:
```shell
curl -fsSL https://download.docker.com/linux/$(. /etc/os-release; echo "$ID")/gpg | sudo apt-key add -
```

Would actually run (on debian):
```shell
curl -fsSL https://download.docker.com/linux/debian/gpg | sudo apt-key add -
```

PR #4317, opened 2017-08-22
Update garbage-collection.md

scjane:

### DIFF
--- a/registry/garbage-collection.md
+++ b/registry/garbage-collection.md
@@ -7,27 +7,17 @@ title: Garbage collection
 As of v2.4.0 a garbage collector command is included within the registry binary.
 This document describes what this command does and how and why it should be used.
 
-## What is Garbage Collection?
-
-From [wikipedia](https://en.wikipedia.org/wiki/Garbage_collection_(computer_science)):
-
-"In computer science, garbage collection (GC) is a form of automatic memory management. The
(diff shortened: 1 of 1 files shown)


registry/garbage-collection.md
```diff
@@ -72,15 +72,15 @@ collection. Layer `a` had one reference removed but will not be garbage
...
+After garbage collection has been run, manifest `A` and its blobs remain.
 
 ```
 A -----> a
     \--> b
 ```
 
 
-## How Garbage Collection works
+## How Garbage Collection works?
```
mstanleyjones: No `?` needed here

scjane: Is the question mark (?) unnecessary in heading "Why Garbage Collection?", too? 
Is the question mark (?) required only if the heading is a complete sentence, like "How does garbage collection work?"? 
A little confused.

mstanleyjones: "How garbage collection works" is a sentence fragment. I would not have used something like "Why garbage collection" either, I probably would have called it "About garbage collection". Feel free to change it.

scjane: I think "About garbage collection" is similar to heading "What is garbage collection?", which already exists on this page. So, can we change "Why garbage collection" to "Why garbage collection is necessary"?

scjane:
I see the reorganized headings and content. Thanks.

PR #4321, opened 2017-08-22
Update compatibility.md

scjane:

### DIFF
--- a/registry/compatibility.md
+++ b/registry/compatibility.md
@@ -45,14 +45,6 @@ support the new version.
 
 ## Registry v2.3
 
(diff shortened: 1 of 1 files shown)


registry/compatibility.md
```diff
@@ -45,15 +45,15 @@ support the new version.
 
 ## Registry v2.3
 
-### Manifest Push with Docker 1.9 and Older
+### Manifest push with Docker 1.9 and older
```
mstanleyjones: Can you swap the order so that the newest is first?

scjane: Order adjusted. Please review again. Thanks.

PR #4330, opened 2017-08-23
Update get-started.md

starriv:
Remove instructions that do not support HyperKit drivers


mstanleyjones:
@shin- PTAL

shin-:
That docker-machine is not maintained or vetted by us, so I'd caution against including it in the official documentation.

mstanleyjones:
That was my feeling too. @starriv sorry about this, but we can only document the bits that we distribute. You might add this content to the Xhyve documentation if you have access, or to your own blog or public Gist.

PR #4332, opened 2017-08-23
nginx does not support bcrypt when using auth_basic

vidarl:
Registry only support bcrypt when using basic auth.
However, nginx does *not* support bcrypt, only md5....

Without this change you'll get error in the nginx log like
```[crit] 5#5: *1 crypt_r() failed (22: Invalid argument)```

Ref. https://stackoverflow.com/questions/31833583/nginx-gives-an-internal-server-error-500-after-i-have-configured-basic-auth

### DIFF
--- a/registry/recipes/nginx.md
+++ b/registry/recipes/nginx.md
@@ -146,7 +146,7 @@ Review the [requirements](/registry/recipes/index.md#requirements), then follow 
 3.  Create a password file `auth/nginx.htpasswd` for "testuser" and "testpassword".
 
     ```bash
-    $ docker run --rm --entrypoint htpasswd registry:2 -Bbn testuser testpassword > auth/nginx.htpasswd
+    $ docker run --rm --entrypoint htpasswd registry:2 -bn testuser testpassword > auth/nginx.htpasswd
     ```
 
 4.  Copy your certificate files to the `auth/` directory.


mstanleyjones:
@stevvooe PTAL

stevvooe:
nginx actually supports whatever is supported by libc `crypt_r` (the error message even makes this clear :) ). Certain operating systems (debian, if I remember correctly) omit support, making it seem like nginx lacks support.

This change either needs to be rolled back or a note needs to be made on how to test for this. `bcrypt` (or similar strong hash) should be used on modern deployments.

stevvooe:
https://github.com/nginxinc/docker-nginx/issues/29 explains the issue in more detail.

vidarl:
Thanx for the info @stevvooe. I was just trying to get the example described in the doc working...

So, based on your input I guess the correct fix is to use nginx:alphine instead of nginx:1.9.

I can make a new PR for that ( which also enables bcrypt again )

mstanleyjones:
@vidarl that would be appreciated. Thanks for the review, @stevvooe and sorry I didn't wait for it before I merged.

stevvooe:
@vidarl I've been through this same thing (I may have written this...). It is up to you. A simple note indicating that there are options might be enough. Include me on the PR and I can help out.

@mstanleyjones No worries!

vidarl:
So to conclude, the example in `registry/recipes/nginx.md` creates the password hashes using bcrypt, while the libc in the nginx image being used doesn't support bcrypt.
This PR fixed the problem by changing hashes from bcrypt to md5, while the more correct way (as pointed out by Stephen ) would be to switch to a image with a libc which supports bcrypt.

Therefore I am reverting this fix and change to nginx:alpine image in #4489

PR #4348, opened 2017-08-23
updates to version variables for compose and machine

londoncalling:
### What's changed

- replaced Compose `site: compose_current` site-wide variable in `_config.yml` with `composeversion` on the only page where it's called

- added a page-level variable for `machineversion`

- updated the values for both to match the current released verisons showing on the associated release pages ([Compose releases](https://github.com/docker/compose/releases) currently at `1.15.0` and [Machine releases](https://github.com/docker/machine/releases/) currently at `0.12.2`)

### Reviewers

@shin- @FrenchBen 

### DIFF
--- a/compose/install.md
+++ b/compose/install.md
@@ -5,6 +5,8 @@ title: Install Docker Compose
 toc_max: 2
 ---
 
+{% assign composeversion = '0.15.0' %}
+
 You can run Compose on macOS, Windows and 64-bit Linux.
 
 ## Prerequisites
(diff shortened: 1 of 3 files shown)


compose/install.md
```diff
@@ -5,6 +5,8 @@ title: Install Docker Compose
 toc_max: 2
 ---
 
+{% assign composeversion = '1.15.0' %}
```
joaofnfernandes: It would be cool if you moved these version definitions to the `_config.yml` file. That way, when  multiple products were released, you could go into a single file and update all the versions for the products.
With the definition being done on this file, you need to scavenger for the files that need to be updated on a release.

mstanleyjones:
@joaofnfernandes people were not updating the Compose one because they didn't realize it was there. It's easier to maintain if it is in the actual install file, where people will be looking when the version seems out of date, I think. The point of this PR is to move them out of the config.yml.

londoncalling:
@joaofnfernandes for now, I think it works better on the particular page. It was getting lost in the site-wide file, and the only place these are being used right now are in their particular install files.
 
If we start using these Compose and Machine version numbers everywhere, then, yes, we should move them back to be site-wide variables.

PR #4352, opened 2017-08-24
Update awslogs.md

scjane:

### DIFF
--- a/engine/admin/logging/awslogs.md
+++ b/engine/admin/logging/awslogs.md
@@ -49,7 +49,7 @@ logging driver options when starting a container.
 
 The `awslogs` logging driver sends your Docker logs to a specific region. Use
 the `awslogs-region` log option or the `AWS_REGION` environment variable to set
-the region.  By default, if your Docker daemon is running on an EC2 instance
+the region. By default, if your Docker daemon is running on an EC2 instance
 and no region is set, the driver uses the instance's region.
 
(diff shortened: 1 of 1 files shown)


engine/admin/logging/awslogs.md
```diff
@@ -168,7 +168,7 @@ The following `strftime` codes are supported:
 | `%p` | AM or PM.                                                        | AM       |
 | `%M` | Minute as a zero-padded decimal number.                          | 57       |
 | `%S` | Second as a zero-padded decimal number.                          | 04       |
-| `%L` | Milliseconds as a zero-padded decimal number.                    | 123      |
+| `%L` | Millisecond as a zero-padded decimal number.                    | 123      |
```
mstanleyjones: Please revert this one.

scjane: Whould you please give me some clue why the plural form ("milliseconds") should be used here? Thanks.

mstanleyjones: Because you're almost never going to set it to `1`. It is technically a "number of milliseconds". That would be the most correct way to say it, but it's a bit wordy.

scjane: Thanks. Then, what about the next item "microsecond" in the table? Should it also use the plural form?

mstanleyjones: Yes, good catch! 👍

PR #4370, opened 2017-08-25
Note about installing required dependencies

frumioj:
Added a note about required dependencies in the case of manual installation of the rpm from local file.

### Proposed changes

I added a couple of notes to mention the need to install required dependencies for Docker, manually from files, if Docker is itself installed manually from a file.

### Related issues (optional)

None known.


mstanleyjones:
I need to think about this. It's really outside the scope of Docker. It also needs some copyediting, but I first need to think about the approach.

mstanleyjones:
I don't think we are going to accept this approach. We did add info about how you might need to grab external dependencies, but we can't really tell people how to do their host OS administration. Thank you for your submission!

PR #4378, opened 2017-08-25
Distinguish tasks from containers

gbarr01:
The command, `docker stack ps getstartedlab`, does not list containers but tasks. I added a flow that helps the user understand tasks and find associated containers so that running `curl http://localhost` makes sense.

### DIFF
--- a/get-started/part3.md
+++ b/get-started/part3.md
@@ -155,12 +155,15 @@ Get the service ID for the one service in our application:
 docker service ls
 ```
 
-Swarm runs tasks that, in this case, spawn containers. Tasks have state and
-their own IDs:
+Swarm for Docker runs tasks that spawn containers. Tasks have state and their
+own IDs:
 
 ```shell
 docker service ps <service>
 ```
+
+>**Note**: SwarmKit tasks do not need to be containers (you can redefine them),
+> but for Docker they are.
 
 Let's inspect one task and limit the ouput to container ID:
 
@@ -203,7 +206,8 @@ Docker will do an in-place update, no need to tear the stack down first or kill
 any containers.
 
 Now, re-run `docker container ls -q` to see the deployed instances reconfigured.
-For example, if you scaled up the replicas, more containers are started.
+If you scaled up the replicas, more tasks, and hence, more containers, are
+started.
 
 ### Take down the app and the swarm
 
--- a/get-started/part3.md
+++ b/get-started/part3.md
@@ -155,15 +155,14 @@ Get the service ID for the one service in our application:
 docker service ls
 ```
 
-Swarm for Docker runs tasks that spawn containers. Tasks have state and their
-own IDs:
+Docker Swarm runs tasks that spawn containers. Tasks have state and their own
+IDs:
 
 ```shell
 docker service ps <service>
 ```
 
->**Note**: SwarmKit tasks do not need to be containers (you can redefine them),
-> but for Docker they are.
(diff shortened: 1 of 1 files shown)


mstanleyjones:
@nishanttotla PTAL

get-started/part3.md
```diff
@@ -146,10 +146,36 @@ Now let's run it. You have to give your app a name. Here, it is set to
 docker stack deploy -c docker-compose.yml getstartedlab
 ```
 
-See a list of the 5 containers you just launched:
+Our single service stack is running 5 container instances of our deployed image on one host. Let's investigate.
```
mstanleyjones: Nit: We try to wrap lines around 80 characters. In Atom you can wrap a paragraph by doing Cmd+Ctrl+Q on Mac. Not sure on Linux.

get-started/part3.md
```diff
@@ -146,10 +146,36 @@ Now let's run it. You have to give your app a name. Here, it is set to
...
+
+Get the service ID for the one service in our application:
+
+```shell
+docker service ls
+```
+
+Swarm runs tasks that, in this case, spawn containers. Tasks have state and their own IDs:
```
mstanleyjones: Why only "in this case"?

get-started/part3.md
```diff
@@ -174,8 +200,7 @@ docker stack deploy -c docker-compose.yml getstartedlab
 Docker will do an in-place update, no need to tear the stack down first or kill
 any containers.
 
-Now, re-run the `docker stack ps` command to see the deployed instances reconfigured. For example, if you scaled up the replicas, there will be more
-running containers.
+Now, re-run `docker container ls -q` to see the deployed instances reconfigured. For example, if you scaled up the replicas, there will be more running containers.
```
mstanleyjones: Maybe "more containers are started."

get-started/part3.md
nishanttotla: Maybe this could be "in the case of Docker", with a clarification added somewhere (maybe a footnote) that tasks need not be containers in SwarmKit.

get-started/part3.md
nishanttotla: Here it might be useful to make the distinction and say "more tasks are started, which means you will see more running containers"

get-started/part3.md
```diff
@@ -146,10 +146,41 @@ Now let's run it. You have to give your app a name. Here, it is set to
...
+Our single service stack is running 5 container instances of our deployed image
+on one host. Let's investigate.
+
+Get the service ID for the one service in our application:
+
+```shell
+docker service ls
+```
+
+Swarm for Docker runs tasks that spawn containers. Tasks have state and their
```
nishanttotla: It is my opinion that we could make this line clearer with something like
"Swarm for Docker runs tasks, which have state and their own IDs. In the case of Docker, each task is a container."

cc @mstanleyjones

get-started/part3.md
```diff
@@ -146,10 +146,41 @@ Now let's run it. You have to give your app a name. Here, it is set to
...
+own IDs:
+
+```shell
+docker service ps <service>
+```
+
+>**Note**: SwarmKit tasks do not need to be containers (you can redefine them),
```
nishanttotla: Since SwarmKit hasn't been mentioned in this document before, should there be context about it? Something like "Docker Swarm is built using a project called SwarmKit", followed by the above.

mstanleyjones:
@gbarr01 can you try to address @nishanttotla comments? You can add additional commits into your branch and then push the branch.

mstanleyjones:
LGTM, let's hope we get a clean Netlify build.

PR #4396, opened 2017-08-28
Remove ro userns restriction

tych0:
As of
https://github.com/opencontainers/runc/commit/66eb2a3e8fc930e1bb6703561152edf5ab550bff
in runc, this is no longer true (in fact, it was never true; the problem
was a bug in runc, not a kernel check). Let's remove it.

Signed-off-by: Tycho Andersen <tycho@docker.com>

### DIFF
--- a/engine/security/userns-remap.md
+++ b/engine/security/userns-remap.md
@@ -251,9 +251,6 @@ The following standard Docker features are incompatible with running a Docker
 daemon with user namespaces enabled:
 
 - sharing PID or NET namespaces with the host (`--pid=host` or `--network=host`).
-- A `--read-only` container filesystem. This is a Linux kernel restriction
-  against remounting an already-mounted filesystem with modified flags when
-  inside a user namespace.
 - external (volume or storage) drivers which are unaware or incapable of using
   daemon user mappings.
 - Using the `--privileged` mode flag on `docker run` without also specifying


tych0:
Oh, whoops. I forgot this was based on master :)

mstanleyjones:
Which Docker version is this targeting? It's now based on `vnext-engine` which right now represents `17.07`.

mstanleyjones:
@cpuguy83 PTAL

tych0:
On Tue, Aug 29, 2017 at 02:27:35PM -0700, Misty Stanley-Jones wrote:
> Which Docker version is this targeting? It's now based on `vnext-engine` which right now represents `17.07`.

Oof my bad, no, I think it's vnext-next-engine, then; nobody has done
a version bump in moby/moby for this fix yet. I'll do that and link it
here.

Which branch should I target it at?

mstanleyjones:
Well, 17.07 just went out and I rebased `vnext-engine` on `master`. You can target at `vnext-engine` for now, but please make a note that we don't know when it will land in Docker yet and please add WIP to the title so we won't merge it by accident.

tych0:
Looks like this just went in: https://github.com/moby/moby/pull/34687, so now it really is legit to be targeted against vnext-engine. I've rebased it on that to get rid of all the other stuff that snuck in, so I think this is good to go.

PR #4397, opened 2017-08-28
improve formatting

friism:
cc @FrenchBen


johndmulhausen:
Looks good, thanks @friism !

PR #4419, opened 2017-08-29
Fix download url for Machine's  command completion

### LINKED ISSUE #4138: Wrong install instruction for machine's bash completion
The download instructions in https://docs.docker.com/machine/completion/ are wrong:
```bash
curl -L https://raw.githubusercontent.com/docker/docker-ce/master/components/cli/contrib/completion/bash/docker > /etc/bash_completion.d/docker
```
Instead of installing bash completion for Machine, the code downloads the current development version of Docker's bash completion over the installed version of Docker's completion. This will introduce inconsistencies, e.g. completion for unreleased features.

albers:
Fixes #4138

### DIFF
--- a/machine/completion.md
+++ b/machine/completion.md
@@ -23,13 +23,13 @@ Place the completion script in `/etc/bash_completion.d/` as follows:
 *   On a Mac:
 
     ```shell
-    curl -L https://raw.githubusercontent.com/docker/machine/v{{machineversion}}/contrib/completion/bash/docker-machine.bash > `brew --prefix`/etc/bash_completion.d/docker-machine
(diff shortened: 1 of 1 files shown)


machine/completion.md
```diff
@@ -4,6 +4,8 @@ keywords: machine, docker, orchestration, cli, reference
 title: Command-line completion
 ---
 
+{% assign machineversion = '0.12.2' %}
```
albers: Is there a way to define variables upstream? This variable is also defined in _machine/install-machine.md_

londoncalling: We can use a variable in the files, that's as automated as we've gotten right now. I'll add that as a part of your PR.

londoncalling: @albers we can define variables in the `config.yml` but they were getting lost in some cases (i.e., people didn't know they were there), so what you did on these works great for now.

PR #4420, opened 2017-08-29
Fix download url for Compose's command completion

### LINKED ISSUE #4137: Wrong install URL for compose's bash completion
Compose and its bash completion must match so that the correct set of features is completed.
If you install a development version of the bash completion script alongside a released version of Compose, you might end up with competion suggesting unsupported features.

The current download link in https://docs.docker.com/compose/completion/ (https://raw.githubusercontent.com/docker/compose/master/contrib/completion/bash/docker-compose) points to the current development version.

albers:
Fixes #4137 

This PR changes the urls from `master` to the tag specified in the template variable `{{composeversion}}`. This notion is also used in the install instructions of Compose and Machine, so it will look familiar to users.

See also #4419 for the corresponding fix for Machine.

### DIFF
--- a/compose/completion.md
+++ b/compose/completion.md
@@ -4,6 +4,8 @@ keywords: fig, composition, compose, docker, orchestration, cli, reference
 title: Command-line completion
 ---
 
+{% assign composeversion = '1.15.0' %}
+
 Compose comes with [command completion](http://en.wikipedia.org/wiki/Command-line_completion)
 for the bash and zsh shell.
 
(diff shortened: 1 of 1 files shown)


compose/completion.md
```diff
@@ -4,6 +4,8 @@ keywords: fig, composition, compose, docker, orchestration, cli, reference
 title: Command-line completion
 ---
 
+{% assign composeversion = '1.15.0' %}
```
albers: Is there a way to define variables upstream? This variable is also defined in compose/install.md

londoncalling: @albers I can add a variable to these files on your PR

PR #4440, opened 2017-08-30
fix example typo, --it should be -it

twoyao:
original command example will fail:
```
➜  ~ docker run -d \
  --it \
  --name sshfs-container \
  --volume-driver vieux/sshfs \
  --mount src=sshvolume,target=/app,volume-opt=sshcmd=test@node2:/home/test,volume-opt=password=testpassword \
  nginx:latest
unknown flag: --it
```


albers:
LGTM. `-it` is short for `-i -t`. `--it` is a non-existent option.

PR #4447, opened 2017-08-30
simplify FAQ

friism:
### Proposed changes

Some of the info in the FAQ no longer makes sense when presented as documentation for Docker products.

Also assorted other clarifications

### DIFF
--- a/engine/faq.md
+++ b/engine/faq.md
@@ -3,54 +3,14 @@ description: Most frequently asked questions.
 keywords: faq, questions, documentation,  docker
 redirect_from:
 - /engine/misc/faq/
-title: Docker Engine frequently asked questions (FAQ)
+title: Docker frequently asked questions (FAQ)
 ---
 
-If you don't see your question here, feel free to submit new ones to
-<docs@docker.com>.  Or, you can fork [the
-repo](https://github.com/moby/moby) and contribute them yourself by editing
-the documentation sources.
+### Does Docker run on Linux, macOS, and Windows?
 
+You can run both Linux and Windows programs and excutables in Docker containers. The Docker platform runs natively on Linux (on x84-64, ARM and many other CPU architectures) and on Windows (x86-64).
 
-### How much does Engine cost?
-
-Docker Engine is 100% free. It is open source, so you can use it without paying.
-
-### What open source license are you using?
-
-We are using the Apache License Version 2.0, see it here:
-[https://github.com/moby/moby/blob/master/LICENSE](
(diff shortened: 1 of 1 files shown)


engine/faq.md
```diff
@@ -3,54 +3,14 @@ description: Most frequently asked questions.
...
-
-### How do containers compare to virtual machines?
-
-Containers and virtual machines (VMs) are complementary. VMs excel at providing
-extreme isolation (for example with hostile tenant applications where you need
-the ultimate break out prevention). Containers operate at the process level,
-which makes them very lightweight and perfect as a unit of software delivery.
-While VMs take minutes to boot, containers can often be started in less than a
-second.
+Docker Inc. has products that let you build and run containers on Linux, Windows and OS X.
```
londoncalling: How about macOS instead of OS X?

londoncalling:
LGTM, with exception of one small comment on changing OS X to macOS.

londoncalling:
@friism I'm going to merge this, I'm just going to fix some broken links I found first. Not anything that you did here, it looks like they were there already. I'll do this soon.

friism:
awesome, thanks @londoncalling

dsheets:
It would be great to have an operating system/container OS compatibility matrix for the question "does Docker run on...?" that shows that macOS and Linux can run Linux containers and Windows can run Windows and Linux containers. WDYT?

friism:
@dsheets I agree that would be nice, but I'm not sure it necessarily belongs in the FAQ (it seems more like an introductory-type item). If we add it here, it would just be because the FAQ used to have a long-winded explanation.

londoncalling:
@dsheets @friism Here are some options for pages it might make sense to add such a matrix. What do you think:

[Top-level Engine install page](https://docs.docker.com/engine/installation/)  - this might make the most sense, and then we can have a brief topic on the other pages linking back to this matrix

[Docker for Mac install page](https://docs.docker.com/docker-for-mac/install/)

[Docker for Windows install page](https://docs.docker.com/docker-for-windows/install/) - this already has [this topic](https://docs.docker.com/docker-for-windows/install/#about-windows-containers-and-windows-server-2016) which could be revised to include or link to the matrix if it is elsewhere

[Get started with Docker for Mac](https://docs.docker.com/docker-for-mac/) 

[Get started with Docker for Windows](https://docs.docker.com/docker-for-windows/)  - this already has this topic on [switching between Windows and Linux containers](https://docs.docker.com/docker-for-windows/#switch-between-windows-and-linux-containers), where we could link back to that matrix if that topic is elsewhere

PR #4463, opened 2017-08-31
Propose an automated update script

drAlberT:
I wrote a little script to handle automatic update to latest release .. hope it could be usefull

### DIFF
--- a/compose/install.md
+++ b/compose/install.md
@@ -109,42 +109,16 @@ by step instructions are also included below.
     ```bash
     sudo -i
     curl -L
-https://github.com/docker/compose/releases/download/$dockerComposeVersion/docker-compose-`uname
--s`-`uname -m` -o /usr/local/bin/docker-compose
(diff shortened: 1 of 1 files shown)


mstanleyjones:
I appreciate you writing up the script, but I'm not sure about this. If you wanted to propose the script itself for `docker/compose` and then it got accepted, then maybe we could use it. As it is now, we can't document a script based off someone's fork. It's a security problem. @shin- what do you think?

drAlberT:
How have I to propose it in `docker/compose` ? Which is the right place for it?

drAlberT:
An alternative could be to simply insert the script code in the README? So anybody can simply C&P it and use .. it would be secure enough IMHO

mstanleyjones:
You can check in https://github.com/docker/compose and look in the contributor guidelines there. It doesn't make sense to put a script like this in the readme in the docs repo, because that readme is all about how to build the docs themselves. Thanks!

PR #4464, opened 2017-08-31
centos/rhel/oracle/fedora install: fixes

kolyshkin:
This commit improves engine installation documentation for CentOS,
RHEL, Fedora and Oracle Linux. There are a few fixes and
simplifications. In particular:

1. ~~Fixed incorrect GPG key fingerprint for docker-ee centos and rhel repos.
I have verified keys for Fedora (they are correct) but not Oracle Linux
-- can someone please check it?~~ v2: no longer required.

2. ~~`yum makecache` does not lead to gpg keys installation -- move the
text telling about gpg key to after 'yum install' and modify it
accordingly.~~  Fix a statement about when GPG keys are presented for verification
    (when you install a package from a recently added repo).

3. Unlike for Debian/Ubuntu, which require `apt update`, with both yum
and dnf, `yum makecache` is not required as it is performed
automatically, so remove this step for the sake of simplicity.
Removing this step also fixes the following problem on Fedora 26:
```
	# dnf makecache fast
	...
	dnf makecache: error: argument timer: invalid choice: 'fast' (choose from 'timer')
```
4. As source repos are disabled by default, there is no need to specify
the architecture when running `yum list docker-[ce]e`. The comment
explaining it is also removed.

5. In Docker EE installation instructions, user is required to input
 `$DOCKERURL` twice. Use `$(cat /etc/yum/vars/dockerurl)`
for the second time to avoid it and simplify the process.

6. An extra `>` character in the middle of the line was removed.

### DIFF
--- a/_includes/ee-linux-install-reuse.md
+++ b/_includes/ee-linux-install-reuse.md
@@ -138,8 +138,8 @@ You can install Docker EE in different ways, depending on your needs:
     $ sudo yum -y install docker-ee
     ```
 
-    If this is the first time you have refreshed the package index since adding
-    the Docker repositories, you will be prompted to accept the GPG key, and
+    If this is the first time you are installing a package from a recently added
+    repository, you will be prompted to accept the GPG key, and
     the key's fingerprint will be shown. Verify that the fingerprint matches
     `{{ gpg-fingerprint }}` and if so, accept the key.
 
@@ -148,11 +148,8 @@ You can install Docker EE in different ways, depending on your needs:
     This example uses the `sort -r` command to sort the results by version
     number, highest to lowest, and is truncated.
 
-    > **Note**: This `yum list` command only shows binary packages. To show
-    > source packages as well, omit the `.x86_64` from the package name.
-
     ```bash
-    $ sudo yum list docker-ee.x86_64  --showduplicates | sort -r
+    $ sudo yum list docker-ee  --showduplicates | sort -r
 
     docker-ee.x86_64         {{ minor-version }}.ee.2-1.el7.{{ linux-dist }}          docker-ee-stable-17.06
     ```
@@ -223,9 +220,7 @@ To upgrade Docker EE:
     Docker 17.03.x to Docker 17.06.x),
     [add the new repository](#set-up-the-repository){: target="_blank" class="_" }.
 
-2.  Run `sudo yum makecache fast`.
-
-3.  Follow the
+2.  Follow the
     [installation instructions](#install-docker), choosing the new version you
     want to install.
 
--- a/engine/installation/linux/docker-ce/centos.md
+++ b/engine/installation/linux/docker-ce/centos.md
@@ -136,8 +136,8 @@ from the repository.
     > which may not be appropriate for your stability needs.
     {:.warning}
 
-    If this is the first time you have refreshed the package index since adding
-    the Docker repositories, you will be prompted to accept the GPG key, and
+    If this is the first time you are installing a package from a recently added
+    repository, you will be prompted to accept the GPG key, and
     the key's fingerprint will be shown. Verify that the fingerprint is
     correct, and if so, accept the key. The fingerprint should match
     `060A 61C5 1B55 8A7F 742B  77AA C52F EB6B 621E 9F35`.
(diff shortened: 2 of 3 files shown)


kolyshkin:
@mstanleyjones can you please take a look?

mstanleyjones:
Can you please rebase your changes, after #4556 was merged?

kolyshkin:
> Can you please rebase your changes

Rebased to current master, commit message updated accordingly (e.g. commit no longer fixes GPG keys).

kolyshkin:
PR description modified accordingly. Local time: 3:24am :)

_includes/ee-linux-install-reuse.md
```diff
@@ -121,7 +121,7 @@ You can install Docker EE in different ways, depending on your needs:
     ```bash
     $ sudo yum-config-manager \
         --add-repo \
-        "$DOCKERURL/{{ linux-dist-url-slug }}/docker-ee.repo"
+        $(cat /etc/yum/vars/dockerurl)/docker-ee.repo"
```
mstanleyjones: I think my way is better now. :)

_includes/ee-linux-install-reuse.md
```diff
@@ -134,8 +134,8 @@ You can install Docker EE in different ways, depending on your needs:
     $ sudo yum -y install docker-ee
     ```
 
-    If this is the first time you have refreshed the package index since adding
-    the Docker repositories, you will be prompted to accept the GPG key, and
+    If this is the first time you are installing a package from a recently added
+    repository, you will be prompted to accept the GPG key, and
```
mstanleyjones: I'm not sure this change really adds much.

_includes/ee-linux-install-reuse.md
```diff
@@ -144,11 +144,8 @@ You can install Docker EE in different ways, depending on your needs:
     This example uses the `sort -r` command to sort the results by version
     number, highest to lowest, and is truncated.
 
-    > **Note**: This `yum list` command only shows binary packages. To show
-    > source packages as well, omit the `.x86_64` from the package name.
-
     ```bash
-    $ sudo yum list docker-ee.x86_64  --showduplicates | sort -r
+    $ sudo yum list docker-ee  --showduplicates | sort -r
```
mstanleyjones: Is this just in case you actually are running s390x?

_includes/ee-linux-install-reuse.md
kolyshkin: OK, here's a longer explanation why I added this.

The `$DOCKERURL` is something you get from a web page, and during step 2 you save it to a file. The steps are:
 - copy (or type) the command from step 2 from docs to a terminal, without pressing Enter
 - copy the EE URL repo from Docker EE registration page
 - remove $DOCKERURL from the command, paste the above-copied URL instead
 - make sure everything looks good, press Enter.

Now, the same sequence is needed during step 6 here, but it can be avoided as we already saved the URL to a file and we can reuse it right here. This is what this hunk does, hope my explanation makes sense.

_includes/ee-linux-install-reuse.md
kolyshkin: The reason is, originally this text was right after `yum makecache` (which refreshed the package index), and now it is (correctly) moved to after `yum install`. Also, refreshing the package index doesn't really lead to asking a user to accept a GPG key, so the statement is just incorrect.

_includes/ee-linux-install-reuse.md
kolyshkin: That, too (or any other non-x86_64 architecture supported, like arm, aarch64, ppc64 etc.)

Also, as I explain in commit message, since source repos are disabled (by default), the only packages one will see will be docker-ee.x86_64. Therefore, specifying architecture is redundant, there is no need to.

kolyshkin:
@mstanleyjones can you PTAL?

_includes/ee-linux-install-reuse.md
mstanleyjones: Your explanation makes sense but the steps in the published file are no longer the same as what you outline.

kolyshkin: Oh, now I see. Yeah, makes total sense. Will remove this hunk.

kolyshkin:
I have rebased the patchset and addressed your last comment. Perhaps you can merge it now and work on splitting the ce install instructions later, so I won't have to rebase again?

PR #4467, opened 2017-08-31
Fix Windows support dump to use version number

gitamiller:
### Proposed changes

Change the docker support dump to use ucp-dsinfo-win:2.2.2
vs
ucp-dsoinfo-win

as the latter uses the latest tag which doesn't exist.


joaofnfernandes:
Sorry for taking so long to check this. Thanks for the contribution 🚀

PR #4477, opened 2017-08-31
Add AWS faq for large clusters

FrenchBen:
### Proposed changes

Document best practice in order to deploy larger clusters on AWS

/cc @friism @kencochrane @ddebroy

### DIFF
--- a/docker-for-aws/faqs.md
+++ b/docker-for-aws/faqs.md
@@ -151,6 +151,9 @@ Benchmark of 3 Managers (m4.large) + 200 workers (t2.medium):
 	* Scaling: 20 workers -> 200 workers via ASG = ~15mins
 
 
+> **Note**: During a Stack upgrade, you will need to match the Auto-Scaling Group worker count, otherwise AWS will scale it back down (aka type 200 workers in the input box)
+
+
 ## Where do I report problems or bugs?
 
 Send an email to <docker-for-iaas@docker.com> or post to the [Docker for AWS](https://github.com/docker/for-aws) GitHub repositories.
--- a/docker-for-aws/faqs.md
+++ b/docker-for-aws/faqs.md
@@ -137,6 +137,20 @@ For more information about adding an SSH key pair to your account, please refer 
 
 All container logs are aggregated within [AWS CloudWatch](https://aws.amazon.com/cloudwatch/).
 
+## Best practice to deploy a large cluster
+
+When deploying a cluster of more than 20 workers, it can take a very long time for AWS to deploy all of the instances (1+hrs).
+It is best to deploy a cluster of 20 workers then scale it up in the Auto-Scaling Group (ASG) once it's been deployed.
+
+Benchmark of 3 Managers (m4.large) + 200 workers (t2.medium):
+
+* Deploying (~3.1hrs)
+	* Deployment: 3 Managers + 200 workers = ~190mins
+* Scaling (~35mins)
+	* Deployment: 3 Managers + 20 workers = ~20mins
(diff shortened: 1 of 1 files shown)


docker-for-aws/faqs.md
```diff
@@ -137,6 +137,20 @@ For more information about adding an SSH key pair to your account, please refer
 
 All container logs are aggregated within [AWS CloudWatch](https://aws.amazon.com/cloudwatch/).
 
+## Best practice to deploy a large cluster
+
+When deploying a cluster of more than 20 workers, it can take a very long time for AWS to deploy all of the instances (1+hrs).
+It is best to deploy a cluster of 20 workers then scale it up in the Auto-Scaling Group (ASG) once it's been deployed.
```
kencochrane: We shouldn't recommend that they change the ASG directly, this could cause problems on upgrades. They should update the cloudformation stack, and change the number of workers. This is safer.

FrenchBen: The problem is the time difference between the deployments. What's the point of having the ASG if they can't change it there directly. 
Especially since we recommend scaling that way:
https://docs.docker.com/docker-for-aws/scaling/#scaling-workers

kencochrane: If you do an update on the cloud formation stack, it is still going to change the ASG behind the scenes but it keeps the stack in sync with the change, and this will prevent issues in the future, if they decide to upgrade. 

We should update the docs to remove the second option, to remove confusion.

friism: @kencochrane what's the exact failure scenario? This should work fine, no?

 1. Start swarm with 20 workers
 2. Scale up to 200 workers
 3. Apply an stack update (making sure that the worker count is still correct)

kencochrane: @friism I believe that would work, but it would depend on the person who is updating the stack to make sure the worker count value is set the same as the ASG value. If they don't do that, they could be surprised.

friism: yes, I agree that's a bit precarious, and we should call that out in the docs. But it seems like a more reliable approach than trying to vault a 200-node swarm into existence, encountering timeouts and other problems.

wdyt?

kencochrane: OK, I just tried it out, this is what I found.

If I create a stack with 0 workers. I then changed the ASG manually to 3, waited for the 3 workers to join the swarm. Then I ran a stack update, and set the worker to 3. The stack was updated, and the 3 worker nodes didn't change. This confirmed what you were saying @friism 

I then tested the following. I started a stack, set the workers to 0. I manually changed the ASG to 3 workers, and waited for them to join the swarm. I then updated the stack and set the worker count to 2. This resulted in the ASG to remove one of the nodes so that I was down to 2 workers to match what the stack update had. Which is what I was expecting.

I'm OK with leaving it the way it is, and then we just update the docs, but I'm afraid that people don't read the docs and they might get burned. Not much we can do about that, but just wanted to point it out.

friism: Thanks for testing @kencochrane. I agree we should point out that upgrade pitfall

FrenchBen: Essentially you need to set your stack update to the number currently running, correct?
I can add that to the docs

kencochrane: Yup

mstanleyjones:
What's the status on this?

FrenchBen:
@mstanleyjones I added the last stack update mention that Ken pointed out - If he's ok with it, we can merge.

PR #4481, opened 2017-08-31
Clarify --replica-id format

adamancini:
### Proposed changes

The DTR replica ID must be in hexadecimal format, and must be 12 characters long.  Update documentation to reflect this limitation.


mstanleyjones:
Unless this change is only for DTR 2.1, you should also update the docs for other DTR versions. cc/ @joaofnfernandes

joaofnfernandes:
@adamancini this file is generated from the same source as `docker/dtr --help`, so we need to make this change upstream, otherwise your change will be lost in the next DTR release.
I'm closing this PR but I've put a PR upstream that introduces the equivalent change.

PR #4491, opened 2017-09-01
Indentation fix

santiph:
### Proposed changes
Indentation in the code example for Volumes > Long syntax is wrong, those volumes should be part of the services definition, and not as part of top level volumes key
https://docs.docker.com/compose/compose-file/#long-syntax-3

### DIFF
--- a/compose/compose-file/index.md
+++ b/compose/compose-file/index.md
@@ -1622,6 +1622,12 @@ services:
       - type: bind
         source: ./static
         target: /opt/app/static
+        
+networks:
+  webnet:
+
+volumes:
+  mydata:
 ```
 
 > **Note:** The long syntax is new in v3.2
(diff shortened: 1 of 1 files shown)


compose/compose-file/index.md
```diff
@@ -1610,18 +1610,24 @@ services:
...
 volumes:
-  - type: volume
-    source: mydata
-    target: /data
-    volume:
-      nocopy: true
-  - type: bind
-    source: ./static
-    target: /opt/app/static
+  mydata:
```
mstanleyjones: Don't you also want to give it access to the bind mount?

gbarr01: @mstanleyjones , I think it has access as is. See:

* https://github.com/docker/compose/issues/5092#issuecomment-321365861
* https://docs.docker.com/compose/compose-file/#volumes

mstanleyjones: Ahh I see, you don't have to explicitly grant the service access to the bind mount.

PR #4492, opened 2017-09-01
Update volumes.md

puchrojo:
Only a minor error   --it -> -it
unknown flag: --it

I get a message error if I try to reproduce it.  
WARN[0000] `--volume-driver` is ignored for volumes specified via `--mount`. Use `--mount type=volume,volume-driver=...` instead. 
docker: Error response from daemon: create sshvolume: invalid option key: "sshcmd".

docker -v
Docker version 17.06.1-ce, build 874a737

Maybe the syntax are changed?


johndmulhausen:
Indeed, thanks @puchrojo 👍

PR #4502, opened 2017-09-04
Update Docker Compose version to 1.16.1

albers:
This PR updates the install docs for Compose and its completion to point to the current version.
See https://github.com/docker/compose/releases/


mstanleyjones:
@shin- oops we need to remember to do this next time. 👼

PR #4529, opened 2017-09-06
Extra package requirements for RHEL

mstanleyjones:
Addresses https://github.com/docker/for-linux/issues/20#issuecomment-326329939

cc/ @friism @alecbenzer

### DIFF
--- a/engine/installation/linux/docker-ee/rhel.md
+++ b/engine/installation/linux/docker-ee/rhel.md
@@ -255,6 +255,16 @@ If you cannot use the official Docker repository to install Docker EE, you can
 download the `.rpm` file for your release and install it manually. You will
 need to download a new file each time you want to upgrade Docker EE.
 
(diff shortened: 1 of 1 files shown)


friism:
lgtm, thx @mstanleyjones

AlecBenzer:
I think the statement on line 76 should also be amended? 

https://github.com/docker/docker.github.io/pull/4529/files#diff-d68b03b9222bc72efb04d43ffca30c1bR76

The "Install from package" installation method is no longer "useful in situations such as installing Docker on air-gapped systems with no access to the internet".

mstanleyjones:
Yes it is! If you need to do it that way you need to get the dependency (and any other uninstalled dependencies) some other way. One very typical way for RHEL is to have local mirrors of the important repos. I'm not going to tell RHEL admins how to do their jobs.

PR #4532, opened 2017-09-07
Update index.md

wfhu:
add emphasis about automatic service discovery on the default network and user-defined networks

### Proposed changes

emphasis text in the document, specifically add emphasis about automatic service discovery on the default network and user-defined networks.

I really really think this section should be emphasised, It is very important


wfhu:
I think this part of the document is pretty important. Most of my colleagues here are confused when they try use Docker service:  why one service can not be reach by another service just by service name? They searched a lot and then found out this information. So I think  it is necessary to emphasize these sentences.


Please kindly consider accept my pull request!

PR #4570, opened 2017-09-08
security: remove blurb about userns conflicting with r/o (#4396)

chengjinshengtaichanyejituan:
As of
https://github.com/opencontainers/runc/commit/66eb2a3e8fc930e1bb6703561152edf5ab550bff
in runc, this is no longer true (in fact, it was never true; the problem
was a bug in runc, not a kernel check). Let's remove it.

Signed-off-by: Tycho Andersen <tycho@docker.com>


chengjinshengtaichanyejituan:
没有授权所以关闭

PR #4572, opened 2017-09-08
Update content cache configs with new style configuration

hinshun:
### Proposed changes

Content caches can discover upstream DTR so there is no longer any need to specify `originhost` and `upstreamhosts` separately. There is a new style to declare `upstreams` and that is just an array of externally-qualified URLs. Content cache for `docker/dtr:2.3.x` onwards still support the old format but we should update the docs.


vikstrous:
@joaofnfernandes PTAL

joaofnfernandes:
Thanks for the contribution @hinshun 🚀

PR #4580, opened 2017-09-11
Typo in example command

EvdokimovN:
Configuration file should be extended with new key-valye instead of overwritten by it


joaofnfernandes:
Thanks for the contribution 🐳

PR #4592, opened 2017-09-12
volume identifier should be a path

gasp:
otherwise the data dir would vanish between build and run time
Docker version 17.06.2-ce, build cec0b72
docker-compose version 1.14.0, build c7bdf9e

### DIFF
--- a/compose/wordpress.md
+++ b/compose/wordpress.md
@@ -40,7 +40,7 @@ Compose to set up and run WordPress. Before starting, you'll need to have
        db:
          image: mysql:5.7
          volumes:
-           - db_data:/var/lib/mysql
+           - ./db_data:/var/lib/mysql
          restart: always
          environment:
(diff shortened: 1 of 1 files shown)


compose/wordpress.md
```diff
@@ -40,7 +40,7 @@ Compose to set up and run WordPress. Before starting, you'll need to have
        db:
          image: mysql:5.7
          volumes:
-           - db_data:/var/lib/mysql
+           - ./db_data:/var/lib/mysql
```
mstanleyjones: This is incorrect. If the first field is a name rather than a path, this indicates that it is a **volume**, rather than a **bind mount**. See https://docs.docker.com/engine/admin/volumes/.

gasp: You are right @mstanleyjones, sorry for the bad PR

PR #4595, opened 2017-09-12
ucp: correct agent config change command

trapier:


trapier:
Thanks to @hannah-docker for reporting.

PR #4606, opened 2017-09-12
Differentiate Edge and Stable channels

### LINKED ISSUE #4520: Release notes should clearly discern between edge and stable channel
[release-notes/docker-ce.md](https://docs.docker.com/release-notes/docker-ce/), CC @johndmulhausen

Like most of the end users I'm on the stable channel, and today I got the 17.06.2 update from the apt repository.
I checked the release notes and found that the [proxy support has been enhanced](https://github.com/docker/cli/pull/93), but unfortunately I hadn't noticed the version mismatch.. After a few attempts, I realized that it was not gonna work, as that feature belongs to the 17.07 which is only available on the edge channel, while I'm on the stable, like I said.

So I think that the release notes document should clearly discern the release notes for the different channels in two separate sections of the document.

BTW, please note that the release notes for 17.06.2-ce are missing.

alb-i986:


johndmulhausen:
Closing due to staleness, non-responsiveness, and merge conflicts.

PR #4611, opened 2017-09-13
Indent services in compose file

### LINKED ISSUE #4610: Compose example has wrong identation
[compose/compose-file/index.md](https://docs.docker.com/compose/compose-file/#compose-file-structure-and-examples), CC @gbarr01

The section 'Compose file structure and examples' has a wrong example. The service definitions need to be children of `services:`, and the example has them as siblings.

In this screenshot `redis` and `db` should have one extra level of indentation.

gbarr01:
Indent services, redis and db, in docker-compose.yml file, so that they are children of "services" rather than siblings. 

Fixes #4610

### DIFF
--- a/compose/compose-file/index.md
+++ b/compose/compose-file/index.md
@@ -55,63 +55,65 @@ services:
     deploy:
       placement:
         constraints: [node.role == manager]
-vote:
-  image: dockersamples/examplevotingapp_vote:before
-  ports:
-    - 5000:80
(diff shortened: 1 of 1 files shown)


gbarr01:
Also indented vote, result, etc., per [dockersamples](https://github.com/dockersamples/example-voting-app/blob/master/docker-compose-javaworker.yml).

Issue #4618, opened 2017-09-13
ucp support command does not support --log-driver argument


trapier:
### Problem description
```
$ docker run --rm -it docker/ucp:2.2.2 support --log-driver 
Incorrect Usage.                                    

Docker Universal Control Plane Tool                 

   support - Create a support dump for this UCP node                                                    

USAGE:                                              
   docker run --rm \                                
        --name ucp \                                
        -v /var/run/docker.sock:/var/run/docker.sock \                                                  
        --log-driver none \                         
        docker/ucp \                                
        support [command options] > docker-support.tgz                                                  
                                                    
DESCRIPTION:                                        
                                                    
This command creates a support dump file for this node, and prints it to stdout.                        


OPTIONS:                                            
   --debug, -D  Enable debug mode                   
   --jsonlog    Produce json formatted output for easier parsing                                        
                                                    
FATA[0000] flag provided but not defined: -log-driver                   
```

### Problem location

https://docs.docker.com/datacenter/ucp/2.2/guides/get-support/#from-the-cli

### Project version(s) affected
UCP: 2.2

### Suggestions for a fix
```
diff --git a/datacenter/ucp/2.2/guides/get-support.md b/datacenter/ucp/2.2/guides/get-support.md
index 181eb9f5df..7e46c807f9 100644
--- a/datacenter/ucp/2.2/guides/get-support.md
+++ b/datacenter/ucp/2.2/guides/get-support.md
@@ -35,8 +35,9 @@ and run:
 docker container run --rm \
   --name ucp \
   -v /var/run/docker.sock:/var/run/docker.sock \
+  --log-driver none \
   {{ page.ucp_org }}/{{ page.ucp_repo }}:{{ page.ucp_version }} \
-  support --log-driver none > docker-support.tgz
+  support > docker-support.tgz
 ```

PR inbound shortly.

trapier:
Thanks to @darwin-traver for reporting.

PR #4619, opened 2017-09-13
ucp: fix support dump --log-driver arp placement

trapier:
fixes: #4618


joaofnfernandes:
Thanks for the contribution @trapier 🤘

PR #4622, opened 2017-09-13
Add info about adding shared folders to Toolbox

### LINKED ISSUE #4334: Integrate note about shared volumes location in docker toolbox
[toolbox/toolbox_install_windows.md](https://docs.docker.com/toolbox/toolbox_install_windows/)

I ran in the issue when installing the docker toolbox that the vm wasn't able to find my shared volume. The reason was that according to the README of boot2docker only the directory `C:\Usres` is mapped by default to the vm. see also https://github.com/boot2docker/boot2docker/blob/af8b089c8c528ed08ad2074c83f1756a5aeaa4ef/README.md#virtualbox-guest-additions.  However, my workspace was under `C:` only. 

A note about this behavior/restriction might avoid others to run into the same problem.

mstanleyjones:
Fixes #4334 

cc/ @FrenchBen for technical review

Specifically, I only included the command-line method on Mac, and I almost feel like I should remove it there too since it's limited, compared to adding the shared folder within VirtualBox.

### DIFF
--- a/toolbox/toolbox_install_mac.md
+++ b/toolbox/toolbox_install_mac.md
@@ -118,33 +118,35 @@ Virtual Box VM, it maintains its configuration between uses.
 
     The terminal does a number of things to set up Docker Quickstart Terminal for you.
 
-        Last login: Sat Jul 11 20:09:45 on ttys002
-        bash '/Applications/Docker Quickstart Terminal.app/Contents/Resources/Scripts/start.sh'
-        Get http:///var/run/docker.sock/v1.19/images/json?all=1&filters=%7B%22dangling%22%3A%5B%22true%22%5D%7D: dial unix /var/run/docker.sock: no such file or directory. Are you trying to connect to a TLS-enabled daemon without TLS?
-        Get http:///var/run/docker.sock/v1.19/images/json?all=1: dial unix /var/run/docker.sock: no such file or directory. Are you trying to connect to a TLS-enabled daemon without TLS?
-        -bash: lolcat: command not found
+    ```
+    Last login: Sat Jul 11 20:09:45 on ttys002
+    bash '/Applications/Docker Quickstart Terminal.app/Contents/Resources/Scripts/start.sh'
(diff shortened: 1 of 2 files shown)


tianon:
If using vbox, it does enumerate all shares and automount!  https://github.com/boot2docker/boot2docker/blob/29dc611f82f134f5a0c5e038ccc7bf114e6da5af/rootfs/rootfs/etc/rc.d/vbox#L40

mstanleyjones:
@tianon Is there anything we need to tweak in the content we merged in this PR, to account for that?

tianon:
It uses the "share name" as the directory for auto mounting inside the VM,
and ignores the automount setting entirely (it automounts everything), so
yeah it's probably worth describing the oddity. 😅

mstanleyjones:
Can you do the automounting on the command line or do you have to do it from within the VirtualBox UI? If you can do it in the command line, what are the relevant options? Can you give an example command?

londoncalling:
@mstanleyjones [This Stack Overflow page](https://stackoverflow.com/questions/30040708/how-to-mount-local-volumes-in-docker-machine) might be helpful. Scroll down to the posts starting with "Add Virtual Machine Mount Point (part 1)"

tianon:
@mstanleyjones yeah, it's available via `docker-machine create`:
```console
$ docker-machine create --help --driver virtualbox | grep share
   --virtualbox-no-share      Disable the mount of your home directory [$VIRTUALBOX_NO_SHARE]
   --virtualbox-share-folder  Mount the specified directory instead of the default home location. Format: dir:name [$VIRTUALBOX_SHARE_FOLDER]
```

Example usage would be something like this:
```console
$ docker-machine create --driver virtualbox --virtualbox-share-folder C:\www:var/www test
```
(which will create `/var/www` inside the container with the contents of `C:\www` from the host mounted)

tianon:
(Or one can use `VBoxManage` as noted by @londoncalling's Stack Overflow link)

PR #4634, opened 2017-09-15
Flesh out swarm key concepts

### LINKED ISSUE #2048: key-concepts: add link to SwarmKit
[engine/swarm/key-concepts.md](https://docs.docker.com/engine/swarm/key-concepts/), CC: @mstanleyjones 
In the first paragraph, it is said that:

>... are built using **SwarmKit**

I think we could make either a small introduction to what is SwarmKit (so that people don't think it's black magic), to at least link the [github repo](https://github.com/docker/swarmkit) for SwarmKit.

Maybe I should fork & make a PR?

mstanleyjones:
Fixes #2048 

cc/ @nishanttotla

### DIFF
--- a/engine/swarm/key-concepts.md
+++ b/engine/swarm/key-concepts.md
@@ -11,16 +11,19 @@ orchestration features of Docker Engine 1.12.
 
 The cluster management and orchestration features embedded in the Docker Engine
 are built using [swarmkit](https://github.com/docker/swarmkit/). Swarmkit is a
-separate project which implements Dockr's orchestration layer and is used
+separate project which implements Docker's orchestration layer and is used
 directly within Docker.
 
 A swarm consists of multiple Docker hosts which run in **swarm mode** and act as
-managers (to manage membership and delegation) and workers, which run
-[swarm services](key-concepts.md#services-and-tasks). When you create a service,
-you define its optimal state (number of replicas, network and storage resources
+managers (to manage membership and delegation) and workers (which run
+[swarm services](key-concepts.md#services-and-tasks)). A given Docker host can
+be a manager, a worker, or perform both roles. When you create a service, you
+define its optimal state (number of replicas, network and storage resources
 available to it, ports the service exposes to the outside world, and more).
-Docker works to maintain that optimal state. For instance, if a worker node
-becomes unavailable, Docker schedules that node's tasks on other nodes.
+Docker works to maintain that desired state. For instance, if a worker node
+becomes unavailable, Docker schedules that node's tasks on other nodes. A _task_
+is a running container which is part of a swarm service and managed by a swarm
+manager, as opposed to a standalone container.
 
 One of the key advantages of swarm services over standalone containers is that
 you can modify a service's configuration, including the networks and volumes it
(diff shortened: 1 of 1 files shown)


engine/swarm/key-concepts.md
```diff
@@ -10,19 +10,36 @@ orchestration features of Docker Engine 1.12.
...
-[services](key-concepts.md#services-and-tasks). The Docker Engine CLI and API
-include commands to manage swarm nodes (e.g., add or remove nodes), and deploy
-and orchestrate services across the swarm.
-
-When you run Docker without using swarm mode, you execute container
-commands. When you run the Docker in swarm mode, you orchestrate services. You can run swarm services and standalone containers on the same Docker instances.
-
-## What is a node?
+are built using [swarmkit](https://github.com/docker/swarmkit/). Swarmkit is a
+separate project which implements Dockr's orchestration layer and is used
```
nishanttotla: typo: `Dockr`

engine/swarm/key-concepts.md
```diff
@@ -10,19 +10,36 @@ orchestration features of Docker Engine 1.12.
...
+directly within Docker.
+
+A swarm consists of multiple Docker hosts which run in **swarm mode** and act as
+managers (to manage membership and delegation) and workers, which run
```
nishanttotla: This language is a little confusing. Act as managers and workers doesn't make it clear that nodes could have either of these two roles.

engine/swarm/key-concepts.md
```diff
@@ -10,19 +10,36 @@ orchestration features of Docker Engine 1.12.
...
+[swarm services](key-concepts.md#services-and-tasks). When you create a service,
+you define its optimal state (number of replicas, network and storage resources
```
nishanttotla: The state is not "optimal". It is simply a desired state.

engine/swarm/key-concepts.md
```diff
@@ -10,19 +10,36 @@ orchestration features of Docker Engine 1.12.
...
+available to it, ports the service exposes to the outside world, and more).
+Docker works to maintain that optimal state. For instance, if a worker node
+becomes unavailable, Docker schedules that node's tasks on other nodes.
```
nishanttotla: Are "tasks" defined before this? It may not be clear what tasks mean otherwise.

engine/swarm/key-concepts.md
```diff
@@ -10,19 +10,36 @@ orchestration features of Docker Engine 1.12.
...
+
+One of the key advantages of swarm services over standalone containers is that
+you can modify a service's configuration, including the networks and volumes
+it is connected to, without the need to restart the service. Docker will
```
nishanttotla: Technically, updating a service does restart it i.e. restart all its tasks. So while the service object will stay in the store and that you don't need to delete, the service does restart. The main feature wrt updating is that we can do rolling updates in a controlled manner to prevent downtime etc.

engine/swarm/key-concepts.md
```diff
@@ -10,19 +10,36 @@ orchestration features of Docker Engine 1.12.
...
+update the configuration and recreate the service tasks for you, in a
+configurable and managed way.
+
+When Docker is running in swarm mode, you can still run standalone containers
+on any of the Docker hosts participating in the swarm, as well as swarm
+services.
```
nishanttotla: Here perhaps it's important to make clear that services can only be created by sending requests to manager nodes. Containers on a given node can be created by sending requests to the Docker API at any node.

mstanleyjones:
@nishanttotla I addressed your feedback. Thanks!

PR #4635, opened 2017-09-15
Explain placement constraints and preferences

### LINKED ISSUE #2241: Lacking documentation for swarm service placement constraints
### Problem description
In https://docs.docker.com/engine/swarm/services/#control-service-scale-and-placement:

```
Service constraints let you set criteria for a node to meet before the scheduler deploys a service to the node. You can apply constraints to the service based upon node attributes and metadata or engine metadata. For more information on constraints, refer to the docker service create CLI reference.
```

However the CLI reference at https://docs.docker.com/engine/reference/commandline/service_create/ only mentions:

```
--constraint	 	Placement constraints
```

I had to go to https://github.com/docker/swarmkit/blob/fb654e9afa426579dc0ce14bf51a80db2c18286e/manager/constraint/constraint.go

### Problem location
URLs included above

I don't think any versions had placement constraint docs

### Suggestions for a fix
Make sure they are documented

mstanleyjones:
Fixes #2241 

cc/ @nishanttotla @aaronlehmann @stevvooe for technical review

### DIFF
--- a/engine/swarm/services.md
+++ b/engine/swarm/services.md
@@ -2,6 +2,7 @@
 description: Deploy services to a swarm
 keywords: guide, swarm mode, swarm, service
 title: Deploy services to a swarm
+toc_max: 4
 ---
 
 Swarm services uses a *declarative* model, which means that you define the
@@ -470,7 +471,38 @@ To create a service with access to Docker-managed secrets, use the `--secret`
 flag. For more information, see
 [Manage sensitive strings (secrets) for Docker services](secrets.md)
 
-### Control service scale and placement
+### Control service placement
+
+Swarm services provide a few different ways for you to control scale and
+placement of services on different nodes.
+
+- You can specify whether the service needs to run a specific number of replicas
+  or should run globally on every worker node. See
+  [_replicated or global services](#replicated-or-global-services).
+
+- You can configure the service's
+  [CPU or memory requirements](#reserve-memory-or-cpus-for-a-service), and the
+  service will only run on nodes which can meet those requirements.
+
+- You can configure the service to run only on nodes with specific (arbitrary)
+  metadata set, and **cause the deployment to fail if appropriate nodes do not
+  exist**. For instance, you could specify that your service should only run on
+  nodes where an arbitrary label `pci_compliant` is set to `true`. This is
+  referred to as [_placement constraints_](#placement-constraints).
+
+- A [_placement preference_](#placement-preferences) is a little different from
+  a plaacement constraint. You can give each node an arbitrary label with a
+  range of values, and spread your service's tasks across those nodes using an
+  algorithm (currently only `spread`, which tries to place them evenly). For
+  instance, if you label each node with `rack` and a value from `1-10`, then
+  specify a placement preference keyed on `rack`, then service tasks will be
+  placed as evenly as possible across all nodes with the label `rack`.
+
+  Unlike constraints, placement preferences are best-effort, and a service will
+  not fail to deploy if no nodes can satisfy the preference.
+
+
+#### Replicated or global services
 
 Swarm mode has two types of services: replicated and global. For replicated
 services, you specify the number of replica tasks for the swarm manager to
@@ -507,17 +539,7 @@ service based upon node attributes and metadata or engine metadata. For more
 information on constraints, refer to the `docker service create`
 [CLI reference](/engine/reference/commandline/service_create.md).
 
-Use placement preferences to divide tasks evenly over different categories of
-nodes. An example of where this may be useful is balancing tasks between
-multiple datacenters or availability zones. In this case, you can use a
-placement preference to spread out tasks to multiple datacenters and make the
-service more resilient in the face of a localized outage. You can use
-additional placement preferences to further divide tasks over groups of nodes.
-For example, you can balance them over multiple racks within each datacenter.
-For more information on constraints, refer to the `docker service create`
-[CLI reference](/engine/reference/commandline/service_create.md).
-
-### Reserve memory or CPUs for a service
+#### Reserve memory or CPUs for a service
 
 To reserve a given amount of memory or number of CPUs for a service, use the
 `--reserve-memory` or `--reserve-cpu` flags. If no available nodes can satisfy
@@ -525,7 +547,7 @@ the requirement (for instance, if you request 4 CPUs and no node in the swarm
 has 4 CPUs), the service remains in a pending state until a node is available to
 run its tasks.
 
-#### Out Of Memory Exceptions (OOME)
+##### Out Of Memory Exceptions (OOME)
 
 If your service attempts to use more memory than the swarm node has available,
 you may experience an Out Of Memory Exception (OOME) and a container, or the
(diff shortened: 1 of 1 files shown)


engine/swarm/services.md
```diff
@@ -472,6 +473,29 @@ flag. For more information, see
...
+
+- You can configure the service to run only on nodes with specific (arbitrary)
+  metadata set, and **cause the service to fail if appropriate nodes do not
+  exist**. This is referred to as
+  [_placement constraints_](#placement-constraints).
+
+- You can configure a replicated service to try to run only on nodes with
+  an arbitrary label set, but to **be able to run even if no appropriate nodes
+  are available**. This is referred to as
+  [_placement preferences_](#placement-preferences).
```
aaronlehmann: That's not really what placement preferences are. They're about trying to balance tasks equally between groups of nodes with certain labels, for example a datacenter label. There is some technical documentation here: https://github.com/docker/swarmkit/blob/master/design/topology.md

engine/swarm/services.md
```diff
@@ -603,6 +641,14 @@ $ docker service create \
   redis:3.0.6
 ```
 
+The syntax is a little different from the previous example. It is a key-value
+pair where the key refers to a placement strategy (currently, `spread` is the
+only supported strategy), and the value is a string using dotted notation, where
+each field refers to a category of metadata. In this case, the service will be
+spread across nodes with the node label `region` set to `east`. If you left off
+the last field, the service would be spread across nodes which had `region` set
+to any value.
```
aaronlehmann: This last part doesn't seem right. Where does `east` enter the picture? What is "the last field" referring to?

engine/swarm/services.md
```diff
@@ -507,41 +531,55 @@ service based upon node attributes and metadata or engine metadata. For more
...
-If your service attempts to use more memory than the swarm node has available,
-you may experience an Out Of Memory Exception (OOME) and a container, or the
-Docker daemon, might be killed by the kernel OOM killer. To prevent this from
-happening, ensure that your application runs on hosts with adequate memory and
-see
-[Understand the risks of running out of memory](/engine/admin/resource_constraints.md#understand-the-risks-of-running-out-of-memory).
+In the following example, you constrain a service to only run on nodes with
+the [label](engine/swarm/manage-nodes.md#add-or-remove-label-metadata) `region`
+set to `east`, using `--constraint` flag with an equality operator
+(`==` or `!=`). You can also use the `constraint` key in a `docker-compose.yml`
```
thaJeztah: perhaps link to the constraint key in the docker-compose file reference for discoverability?

engine/swarm/services.md
```diff
@@ -507,41 +531,55 @@ service based upon node attributes and metadata or engine metadata. For more
...
+run on the same node, or each node will only run one replica, or that some
+nodes won't run any replicas.
 
-Swarm services allow you to use resource constraints, placement preferences, and
-labels to ensure that your service is deployed to the appropriate swarm nodes.
+```bash
+$ docker service create \
+  --name my-nginx \
+  --replicas 5 \
+  --constraint region==east \
```
thaJeztah: Should we have an additional example showing that multiple constraints can be combined (e.g. adding `region!=foo` or `type!=development`), or too much in-depth for this topic?

stevvooe:
@aaronlehmann :crossed_fingers:

mstanleyjones:
Updated, PTAL again. cc/ @cpuguy83

engine/swarm/services.md
```diff
@@ -2,6 +2,7 @@
 description: Deploy services to a swarm
 keywords: guide, swarm mode, swarm, service
 title: Deploy services to a swarm
+toc_max: 4
 ---
 
 Swarm services uses a *declarative* model, which means that you define the
```
gbarr01: Swarm services use

engine/swarm/services.md
```diff
@@ -470,7 +471,38 @@ To create a service with access to Docker-managed secrets, use the `--secret`
...
 
-### Control service scale and placement
+### Control service placement
+
+Swarm services provide a few different ways for you to control scale and
+placement of services on different nodes.
+
+- You can specify whether the service needs to run a specific number of replicas
+  or should run globally on every worker node. See
+  [_replicated or global services](#replicated-or-global-services).
```
gbarr01: _replicated > Replicated

engine/swarm/services.md
```diff
@@ -470,7 +471,38 @@ To create a service with access to Docker-managed secrets, use the `--secret`
...
+  service will only run on nodes which can meet those requirements.
+
+- You can configure the service to run only on nodes with specific (arbitrary)
+  metadata set, and **cause the deployment to fail if appropriate nodes do not
+  exist**. For instance, you could specify that your service should only run on
+  nodes where an arbitrary label `pci_compliant` is set to `true`. This is
+  referred to as [_placement constraints_](#placement-constraints).
+
+- A [_placement preference_](#placement-preferences) is a little different from
+  a plaacement constraint. You can give each node an arbitrary label with a
```
gbarr01: plaacement > placement

engine/swarm/services.md
mstanleyjones: I did add a note that you can specify multiple constraints, and I do have an example with multiple preferences. Not sure. I suppose I could add a second example.

engine/swarm/services.md
```diff
@@ -470,7 +471,38 @@ To create a service with access to Docker-managed secrets, use the `--secret`
...
+  range of values, and spread your service's tasks across those nodes using an
+  algorithm (currently only `spread`, which tries to place them evenly). For
+  instance, if you label each node with `rack` and a value from `1-10`, then
+  specify a placement preference keyed on `rack`, then service tasks will be
+  placed as evenly as possible across all nodes with the label `rack`.
+
+  Unlike constraints, placement preferences are best-effort, and a service will
+  not fail to deploy if no nodes can satisfy the preference.
+
```
gbarr01: Suggestion: 

> Placement preferences let you apply an arbitrary label with a range of values to each node, and spread your service's tasks across those nodes using an algorithm (currently only `spread`), which tries to place them evenly. For instance, if you label each node with `rack` and a value from `1-10`, then specify a placement preference keyed on `rack`, then service tasks are placed as evenly as possible across all nodes with the label `rack`.
> 
> Unlike constraints, placement preferences are best-effort, and a service do not fail to deploy if no nodes can satisfy the preference.

engine/swarm/services.md
```diff
@@ -470,7 +471,38 @@ To create a service with access to Docker-managed secrets, use the `--secret`
...
+- You can configure the service's
+  [CPU or memory requirements](#reserve-memory-or-cpus-for-a-service), and the
+  service will only run on nodes which can meet those requirements.
+
+- You can configure the service to run only on nodes with specific (arbitrary)
+  metadata set, and **cause the deployment to fail if appropriate nodes do not
+  exist**. For instance, you could specify that your service should only run on
+  nodes where an arbitrary label `pci_compliant` is set to `true`. This is
+  referred to as [_placement constraints_](#placement-constraints).
+
```
gbarr01: Suggestion: 

> Placement constraints let you configure the service to run only on nodes with specific (arbitrary) metadata set, and cause the deployment to fail if appropriate nodes do not exist. For instance, you can specify that your service should only run on nodes where an arbitrary label `pci_compliant` is set to `true`.

engine/swarm/services.md
```diff
@@ -537,11 +559,52 @@ see
...
+`pci_compliant` set to `true`. If no appropriately-labelled nodes are available,
+deployment will fail. The `--constraint` flag uses an equality operator
+(`==` or `!=`). It is possible that all services will run on the same node, or
+each node will only run one replica, or that some nodes won't run any replicas.
+
+```bash
+$ docker service create \
+  --name my-nginx \
+  --replicas 5 \
+  --constraint region==east \
```
gbarr01: Should this be: `--constraint pci_compliant==true`? or `constraint label==pci_compliant`? Not sure of the syntax but don't see anything about pci_compliant in this example.

engine/swarm/services.md
```diff
@@ -537,11 +559,52 @@ see
...
+  nginx
+```
+
+You can also use the `constraint` service-level key in a `docker-compose.yml`
+file.
+
+You can specify multiple placement constraints, and they are `AND`ed together.
```
gbarr01: Suggestion: 
You can specify multiple placement constraints by conjoining them with `AND`.

mstanleyjones: That's not what I mean. I mean that if you specify multiple ones, you will only deploy onto nodes that meet all of them.

mstanleyjones:
Thanks @gbarr01, I incorporated your feedback.

engine/swarm/services.md
mstanleyjones: The text just got out of sync wth the example. Fixed.

mstanleyjones:
Feedback incorporated. Merged. Thanks all!

engine/swarm/services.md
```diff
@@ -470,7 +471,40 @@ To create a service with access to Docker-managed secrets, use the `--secret`
...
+- [Placement constraints](#placement-constraints) let you configure the service
+  to run only on nodes with specific (arbitrary) metadata set, and cause the
+  deployment to fail if appropriate nodes do not exist. For instance, you can
+  specify that your service should only run on nodes where an arbitrary label
+  `pci_compliant` is set to `true`.
+
+- [Placement preferences](#placement-preferences) let you apply an arbitrary
+  label with a range of values to each node, and spread your service's tasks
+  across those nodes using an algorithm. Currently, the only supported algorithm
+  is `spread`, which  which tries to place them evenly. For instance, if you
```
cpuguy83: `which  which`

engine/swarm/services.md
```diff
@@ -470,7 +471,40 @@ To create a service with access to Docker-managed secrets, use the `--secret`
...
+  label each node with a label `rack` which has a value from 1-10, then specify
+  a placement preference keyed on `rack`, then service tasks are placed as
+  evenly as possible across all nodes with the label `rack`, after taking other
+  placement constraints, placement preferences, and other node-specific
+  limitations into account.
```
cpuguy83: Also note that the preference is just ranking a node higher than other nodes.
Swarm will still spread to nodes that do not match any existing preference if all nodes with a preference are already scheduled (or unavailable for new tasks that matter) but more replicas were requested.
This is because swarm's spread strategy schedules for high availability and the preference just says "prefer these nodes over others", but HA is still the most important thing... hope that makes sense.

cpuguy83: An example:

If you have `N` nodes with the `rack` label (and then some others), and `N+1` replicas requested, the `+1` will be scheduled on a node that doesn't already have the service on it if there is one regardless of the `rack` label existing.

PR #4641, opened 2017-09-15
Docker service registry not running correctly (Replicas 0/1)

jonashackt:
As it is said in the text before "The example bind-mounts `/mnt/registry` on the swarm node to `/var/lib/registry/` within the container." the example uses a bind-mount instead of a volume-mount (see difference here in the docs: https://docs.docker.com/engine/swarm/services/#give-a-service-access-to-volumes-or-bind-mounts). As bind-mounts are only applied with `type=bind` configuration parameter of `--mount`, it has to be added here so that the `swarm-registry` service is able to start.

Additionally, the bind-mount relies on the pre-existing src-directory `/mnt/registry`, which should also be added to the documentation - see important notes "If you bind mount a host path into your service’s containers, the path must exist on every swarm node." in https://docs.docker.com/engine/swarm/services/#give-a-service-access-to-volumes-or-bind-mounts.

### Proposed changes

Updated the docs.

### Related issues (optional)

https://github.com/moby/moby/issues/25912
https://github.com/moby/moby/issues/34532


jonashackt:
Resolved the conflict and changed `src` into `src` :)

PR #4657, opened 2017-09-18
remove the recommendation of 7 nodes.

clemenko:
removing the recommendation of 7 managers nodes based guidance from engineering. they claimed 7 manager nodes was less performative then 5.

### Proposed changes

Guidance that came out from a customer issue.


thaJeztah:
The bullets below the table already point out that adding more managers will degrade performance; https://github.com/docker/docker.github.io/pull/4657/files#diff-050b630391e33fbc9fc7bb3018848054R34 (I see max 7 managers is also mentioned there btw)

PR #4661, opened 2017-09-19
Add Windows example for running Prometheus

### LINKED ISSUE #4206: Prometheus samples aren't Windows friendly
[engine/admin/prometheus.md](https://docs.docker.com/engine/admin/prometheus/), CC @mstanleyjones 

For Docker for Windows users, two things stand out. The first is the path to /tmp/prometheus.yml, which doesn't work in the command:
``` 
 docker service create --replicas 1 --name my-prometheus \
    --mount type=bind,source=/tmp/prometheus.yml,destination=/etc/prometheus/prometheus.yml \
    --publish 9090:9090/tcp \
    prom/prometheus ```
 the path should be c:\tmp\promethus.yml. I have verified this works.

Also the copy/paste of the commands in this doesn't work for some reason. Worth exploring further. I reverted all commands to a single line with no breaks and they worked.

mstanleyjones:
Fixes #4206

Added more info to guide Windows users, added tabbed output for running on Linux/Mac or Windows.

Best to wait for Netlify and view the output in place. There is a little hiccup where the tab groups are not exactly 1:1 for Mac (in the first tab set, Mac is in the second tab, while in the second one, Mac is in the first group) but I feel like it's OK.

cc/ @stevvooe

### DIFF
--- a/engine/admin/prometheus.md
+++ b/engine/admin/prometheus.md
@@ -50,9 +50,10 @@ Docker now exposes Prometheus-compatible metrics on port 9323.
 In this example, Prometheus runs as a Docker container on the same host.
 
 Copy one of the following configuration files and save it to
-`/tmp/prometheus.yml`. This is a stock Prometheus configuration file, except
-for the addition of the Docker job definition at the bottom of the file.
-Docker for Mac and Docker for Windows need a slightly different configuration.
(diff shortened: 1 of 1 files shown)


stevvooe:
LGTM

PR #4670, opened 2017-09-19
Update example to use official consul image

### LINKED ISSUE #2908: Use official consul docker image in tutorial
[engine/userguide/networking/get-started-overlay.md](https://docs.docker.com/engine/userguide/networking/get-started-overlay/), CC @mstanleyjones
Hello,
Is it possible to have this tutorial with the official consul image?
This tutorial actually work with progrium/consul image wich has an old version of consul.
I tried to do it myself without success.

mstanleyjones:
Fixes #2908


mstanleyjones:
@stevvooe Is there any reason this example has to use Docker machine and standalone swarm? I feel like maybe it should be updated, in a follow-up PR.

mstanleyjones:
Latest commit removes swarm-mode stuff from this topic, is more explicit about this only being for swarm classic, renames the file (and adds a redirect), and reorgs the TOC slightly. @thaJeztah @nishanttotla  PTAL

mstanleyjones:
Merging now, we may need some follow-up if some of these options are deprecated.

PR #4682, opened 2017-09-20
Repair note on deferred deletion

### LINKED ISSUE #4679: Wrong sentence about devmapper deferred deletion
[engine/userguide/storagedriver/device-mapper-driver.md](https://docs.docker.com/engine/userguide/storagedriver/device-mapper-driver/), CC @mstanleyjones

Currently, the documentation says:

> **Note**: The deferred deletion option, `dm.use_deferred_deletion=true`,
> is not yet supported on RHEL, CentOS, or Ubuntu 14.04 when using the
> default kernel version 3.18.

This is very confusing as there is no default kernel version 3.18 for any of those platforms.

I found the original commit (https://github.com/docker/docker.github.io/pull/2266/commits/4d09936aefa3a86b96cbf57b4636e0187a04442c) and it says:

> **Note**: The deferred deletion option, `dm.use_deferred_deletion=true`, is not yet supported
> on CentOS, RHEL, or Ubuntu 14.04 when using the default kernel.  Support was added in the
> upstream kernel version 3.18.

Now, this makes the perfect sense! Please correct the sentence (the way it was stated in the original sounds correct and clear).

gbarr01:
Updated the note text as recommended per the original commit: 
https://github.com/docker/docker.github.io/commit/4d09936aefa3a86b96cbf57b4636e0187a04442c

Fixes #4679

### DIFF
--- a/engine/userguide/storagedriver/device-mapper-driver.md
+++ b/engine/userguide/storagedriver/device-mapper-driver.md
@@ -312,7 +312,7 @@ assumes that the Docker daemon is in the `stopped` state.
     ```bash
     $ sudo lvs -o+seg_monitor
 
-    LV       VG     Attr       LSize  Pool Origin Data%  Meta%  Move Log Cpy%Sync Convert Monitor  
+    LV       VG     Attr       LSize  Pool Origin Data%  Meta%  Move Log Cpy%Sync Convert Monitor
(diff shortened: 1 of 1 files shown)


kolyshkin:
FWIW LGTM

PR #4684, opened 2017-09-21
Update networking.md

skymon:
changed --p to -p

### Proposed changes

Changed the documentation for networking in Docker for Mac, because the --p option resulted in "unknown flag --p".


johndmulhausen:
Thanks, @skymon :)

PR #4688, opened 2017-09-21
Update part2.md

ms03001620:
按照操作步骤，在这里会有2个相同id的镜像。如果不加-f 会提示（image is referenced in multiple repositories）导致操作失败


johndmulhausen:
This should not be necessary unless you're using a really generic name that collides with something you did somewhere else on your local machine. Let's keep it simple here.

PR #4689, opened 2017-09-21
Update RAM recommendation for UCP managers

binman-docker:
### Proposed changes

Added recommendation of 8GB of RAM for manager nodes. This is was recommended for our systems by the UCP team. 4GB is dangerously low for a medium to large cluster running a large number of services and tasks.

### DIFF
--- a/datacenter/ucp/2.0/guides/installation/system-requirements.md
+++ b/datacenter/ucp/2.0/guides/installation/system-requirements.md
@@ -15,7 +15,8 @@ all nodes must have:
 
 * Linux kernel version 3.10 or higher
 * CS Docker Engine version 1.12.1
-* 2.00 GB of RAM
+* 8.00 GB of RAM for manager nodes or nodes running DTR
+* 4.00 GB of RAM for worker nodes
 * 3.00 GB of available disk space
 * A static IP address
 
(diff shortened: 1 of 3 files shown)


joaofnfernandes:
I agree with this, but I'd like @vsaraswat 's opinion.

vsaraswat:
At a start I think it's fine. I'd ask @vikstrous or @pdevine what DTR recommendation is (if it is 8GB then I agree).

One q: Should we backport this fix to the UCP 2.1 docs and 2.0 docs as well?

joaofnfernandes:
Yes, I would backport this to UCP 2.1 and 2.0. I can take care of that.

mghazizadeh:
FWIW, recent testing of a patch for the snapshot size also shows that 4 GB is too low of a recommendation for a medium size cluster. cc @davidwilliamson

vsaraswat:
Mehdi does that include workers? Or still just managers?

-Vivek

davidwilliamson:
@vsaraswat I think we should limit the updated recommendation for managers only. 

The memory-related issues I saw pertained to managers running UCP (i.e., user services restricted to worker nodes)

The requirements for workers depend on what services the user plans on running, how many workers they deploy, and possibly what node constraints they use for deployments. I'm not sure we can make a blanket recommendation for workers.

vikstrous:
rethinkdb scales to use as much ram as it can for memory caches, so the more ram you give DTR, the better it'll perform. It needs at least 2-4 GB to be able to run scans though. It can run on less if you don't use security scanning.

mghazizadeh:
Just managers

PR #4697, opened 2017-09-22
test repo is called test, not testing

smartbit:
search `docker-ce-testing` and replace with `docker-ce-test`


smartbit:
You're welcome.

mstanleyjones:
Thanks! I looked over the other Docker CE installation docs and found that the Ubuntu doc also had this mistake, so I added a commit that fixes it too.

PR #4706, opened 2017-09-24
Update part5.md

khlumzeemee:
I couldn't start redis and was getting the following error message `invalid mount config for type "bind": bind source path does not exist"`.
Using docker-compose config I got the full path that docker was expecting, created the folder and could finally start the redis service.

### DIFF
--- a/get-started/part5.md
+++ b/get-started/part5.md
@@ -231,6 +231,8 @@ Redis service. Be sure to replace `username/repo:tag` with your image details.
     ```shell
     docker-machine ssh myvm1 "mkdir ./data"
     ```
+    
+    You should also create the `data` folder in the folder containing your `docker-compose` file.
 
 3.  Make sure your shell is configured to talk to `myvm1` (examples are [here](part4.md#configure-a-docker-machine-shell-to-the-swarm-manager)).
 


johndmulhausen:
We do tell people to do that just beforehand. :)

khlumzeemee:
Hi, I don't think it is written anywhere in the tutorial, that's why I offered to add it.
But if you think otherwise, I would appreciate if you could point out where you tell people to do that, I couldn't find it.

johndmulhausen:
Search for "mkdir"

khlumzeemee:
Hi @johndmulhausen , the `mkdir` is preceded by an `ssh`, which means that it creates the directory on the virtual machine.
But you need to create two directories, one on your local machine and one on the virtual one.
Then this configuration maps your local directory to the one on the virtual machine.
```
volumes:
      - /home/docker/data:/data
```
Anyway, I see that you already closed the request, so I'm not going to insist anymore.
I lost some time with this and simply wanted to prevent other people from facing the same `invalid mount config for type "bind": bind source path does not exist"` error message. If they do then they can always refer to this conversation 😄

johndmulhausen:
Hi @khlumzeemee - I'm sorry if we're not making this clear in the tutorial but the intent of creating the virtual machines is to have a cluster that is totally separate from your host. There should be no interaction between your local machine's file system and the virtual machine's.

PR #4702, opened 2017-09-22
Docker CE 17.09 release prep

mstanleyjones:
- Update release notes (also reorg to make Edge more obvious, Fixes #4520)
- Add 1.32 API
- Update CLI ref yaml for Edge and Stable
- Update variables

PR #4711, opened 2017-09-25
Add info about deprecated and experimental options to CLI template

mstanleyjones:
Capture new `deprecated` and `experimental` info in CLI flags.

This is WIP because it doesn't work yet. @johndmulhausen @allejo please help! I tried changing the tables to HTML to help eliminate problems to do with newlines. I think it's really close to working, I must be just missing something really obvious. Right now, you will want to apply this patch on top of  https://github.com/docker/docker.github.io/pull/4702/ which is the only place where the YAML has these extra items in it. Sorry it's a little bit hard to test.

### DIFF
--- a/_includes/cli.md
+++ b/_includes/cli.md
@@ -10,6 +10,33 @@
 {{ site.data[include.datafolder][include.datafile].short }}
 {% endif %}
 
+{% if site.data[include.datafolder][include.datafile].min_api_version %}
+
+**The client and daemon API must both be at least
+{{ site.data[include.datafolder][include.datafile].min_api_version }}
+to use this command.** Use the `docker version` command on the client to check
+your client and daemon API versions.
+
+{% endif %}
+
+{% if site.data[include.datafolder][include.datafile].deprecated %}
+
+> This command is deprecated.
+>
+> It may be removed in a future Docker version.
+{: .warning }
+
+{% endif %}
+
+{% if site.data[include.datafolder][include.datafile].experimental %}
+
+> This command is experimental.
+>
+> It should not be used in production environments.
+{: .important }
+
+{% endif %}
+
 {% if site.data[include.datafolder][include.datafile].usage %}
 
 ## Usage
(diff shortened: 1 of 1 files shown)


mstanleyjones:
@johndmulhausen This is not quiiiiite working yet. See https://github.com/docker/docker.github.io/blob/7f2fcb9e820d7b011d9dde562c2362f81c7664ec/_data/engine-cli-edge/docker_container_start.yaml how the `--checkpoint` flag is supposed to be `experimental` but in the rendered output it is still showing as `stable` 😭 Any thoughts? You should be able to actually build and this in the PR now because it's against `vnext-engine` which has the updated YAML files.

_includes/cli.md
```diff
@@ -23,20 +50,60 @@
...
+    <td>Description</td>
+  </tr>
+</thead>
+<tbody>
+{% for option in site.data[include.datafolder][include.datafile].options %}
+  <tr>
+    <td markdown="span">`--{{ option.option }}{% if option.shorthand %} , -{{ option.shorthand }}{% endif %}`</td>
+    <td markdown="span">{% if option.default_value and option.default_value != "[]" %}`{{ option.default_value }}`{% endif %}</td>
+    <td markdown="span">
+    {% if option.deprecated == 'true' and option.experimental == 'true' %}
```
allejo: If the YAML file is as such,

```yaml
deprecated: false
experimental: true
```

Those values would be treated as booleans but in the Liquid, aren't you checking them as strings? Either remove the quotes from `'true'` or just do `{% if option.deprecated and option.experimental %}` since they're booleans.

mstanleyjones:
Thanks @thaJeztah !

We can tweak the styles in a follow-up, there is something a little funny about the alignment, and it is possible we should put the `experimental` or `deprecated` status of the flags at the beginning of their descriptions.

mstanleyjones:
Thanks @allejo you were absolutely right!

Issue #4723, opened 2017-09-26
Starting Redis outside of management node results in bind exception


jasonmevans:
[get-started/part5.md](https://docs.docker.com/get-started/part5/), CC @johndmulhausen

First, thanks for the very thorough, and well-written tutorial. I've found it extremely helpful getting acquainted with Docker.

I was following the instructions verbatim, and was having great success up to the point where I was deploying the Redis service. After creating the `~/data` folder in myvm1, checking `docker-machine ls` for myvm1 being active and running the `docker stack deploy -c docker-compose.yml getstartedlab` I noticed my Redis service was failing to start with the error *invalid mount config for type "bind": bind source path does not exist*. After Googling around for a while I determined that the issue was that the Redis service was not finding the data folder I had created on the manager node; however, it was there. I tried several times to tear down the service, restart the service and even tried creating a data folder at `/data`, but had no luck. 

Then, I remembered while watching the screen recording at the bottom of the tutorial page that you actually started the getstartedlab stack by scp-ing the docker-compose.yml file into the master, and running the `docker stack deploy ...` command directly from there. I gave that method a shot, and was successful! I believe what was happening was that the `docker-machine env myvm1` command was causing the `docker stack deploy ...` command to actually look for the data folder in my MacBook home folder INSTEAD of in the management node.

Can you confirm/refute my diagnosis, and if correct suggest how to get the posted docs to work as written? If it is not actually possible to execute the `docker stack deploy ...` command from outside of the management node for the reason I have proposed could you please update the tutorial to reflect that?

Thanks very much!
Jason Evans

johndmulhausen:
Yep that's the issue! You can actually avoid doing it the `scp` way; @londoncalling is punching up this page so it's a little less confusing, and also `scp` is Linux-only. But the essential takeaway you have: that you have to direct the manager to run the `docker stack deploy` command is correct. :)

Will review how clear we are being about where the command needs to run with @londoncalling

londoncalling:
@jasonmevans you are exactly right. We are trying to find a way to make the commands work natively on both Mac and Windows. We'll address this in the docs today one way or the other. More soon.

jasonmevans:
@johndmulhausen @londoncalling I really appreciate you guys looking at this so quickly. I just tested out what I suspected was happening, and yeah it binds to the data folder in the working folder in my Mac file system. Kinda neat, actually! Docker is quite powerful. 

[...]

PR #4725, opened 2017-09-27
Stop gap measure to fix data persistence part of the tutorial

londoncalling:
### What's changed

- Added back in `docker-machine scp` and `docker-machine ssh` commands in part 5 to get the the data persistence step with the `redis` service to work. With the current Compose file, when you run  the `docker stack deploy` command on the local host with `myvm1` active, Docker looks for the `~./data` directory on the local host, not on the remote machine, `myvm1`.

- Added an override to the `redis` command in the Compose file to get the data to persist.

### Related 

Addresses #4723, but we will leave the issue open to further research getting this to work without `scp` if possible.

Related PR is #4675

### Reviewers

@shin- @friism @johndmulhausen @jasonmevans @mstanleyjones

### DIFF
--- a/get-started/part5.md
+++ b/get-started/part5.md
@@ -232,10 +232,7 @@ Redis service. Be sure to replace `username/repo:tag` with your image details.
 
     - The placement constraint you put on the Redis service, ensuring that it
       always uses the same host.
-    - The volume you created that lets the container access `./data` (on the host)
-      as `/data` (inside the Redis container). While containers come and go, the
-      files stored on `./data` on the specified host will persist, enabling
-      continuity.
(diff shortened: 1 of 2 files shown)


get-started/part5.md
```diff
@@ -85,7 +85,7 @@ with the following. Be sure to replace `username/repo:tag` with your image detai
       visualizer:
         image: dockersamples/visualizer:stable
         ports:
-          - "8080:8080"
+          - "8080:8080":
```
friism: Is this colon intentional?

mstanleyjones: I can't speak for whether it is intentional but I can confirm that it doesn't actually break YAML.

londoncalling: it's not intentional, I believe. I'll remove it. Working with @shin-  on this problem today.

PR #4733, opened 2017-09-27
Fix problems in the SDK examples

mstanleyjones:
cc/ @tiborvass 

Without this fix, the Python example in the Overview is still shown even though the Go tab is active, and there is another small mistake in one of the examples in the Examples page.


tiborvass:
LGTM

PR #4740, opened 2017-09-28
standalone swarm: consul listen on public ip

jwfang:
consul (maybe caused by latest changes) default listens on 127.0.0.1, so can't access by outside.


johndmulhausen:
Thanks, @jwfang - looks like this is indeed the case.

PR #4743, opened 2017-09-28
nginx image doesn't have *ip* command

jwfang:
use *ip* from *busybox*, and join *web* container network namespace


jwfang:
below are my result, maybe different nginx image ?

```bash
jamesf@xenial:~/src/libnetwork$` docker images nginx
nginx               latest              1e5ab59102ce        25 hours ago        108MB
jamesf@xenial:~/src/libnetwork$ docker run -it --rm nginx bash
root@9de50c6b26b9:/# which ip
root@9de50c6b26b9:/# echo $?
1
root@9de50c6b26b9:/# type -t ip
root@9de50c6b26b9:/# echo $?
1
```

mstanleyjones:
cc/ @thaJeztah

thaJeztah:
Yes, looks like `ip` is not available in the current nginx image (possibly because they switched to a newer base image distro); 

```bash
$ docker run --rm nginx sh -c 'which ip'
# (no output0
```

Using an old version of nginx:

```bash
$ docker run --rm nginx:1.9 sh -c 'which ip'
/sbin/ip
```

depending on the context; `busybox` or `nginx:alpine` would work (both use busybox);

```bash
$ docker run --rm nginx:alpine sh -c 'which ip'                                                                                                                                                                                                                                                                                                                    /sbin/ip

$ docker run --rm busybox sh -c 'which ip'                                                                                                                                                                                                                                                                                                                    /sbin/ip
```

If we want to keep the examples a bit more lightweight, we could change all nginx images here to `nginx:alpine`

PR #4752, opened 2017-09-29
Update plan-installation.md

scjane:


johndmulhausen:
The first use of the FQDN should be spelled out and then have the acronym in parenthesis afterwards. Just jumping in with FQDN isn't a good idea.

scjane:
Should the acronym given in parentheses for the first time on a page also use the plural form if its full name is plural? That is, is it necessary to change "(FQDN)" to "(FQDNs)“ in the following sentence on this page?

"Decide if you want to use short hostnames, like engine01, or Fully Qualified Domain Names (FQDN), like engine01.docker.vm."

PR #4760, opened 2017-09-29
Use the handy Django runserver shortcut

maxg203:
### Proposed changes

I noticed whilst reading a Django guide that there is a more concise shortcut whereby `0:8000` is equivalent to `0.0.0.0:8000`.

### DIFF
--- a/compose/django.md
+++ b/compose/django.md
@@ -71,7 +71,7 @@ and a `docker-compose.yml` file. (You can use either a `.yml` or `.yaml` extensi
         image: postgres
       web:
         build: .
-        command: python3 manage.py runserver 0.0.0.0:8000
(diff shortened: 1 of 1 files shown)


maxg203:
Interesting, I was of the perspective that many people do not know that the `0:8000` shortcut exists, so it would be nice to see it more widely used in my opinion and to me it is also more concise albeit *slightly* less explicit.

Thanks!

mstanleyjones:
I agree with @thaJeztah that the current docs are more clear. In addition to his comments, it's unclear what happens if you are using IPv6, among other potential points of confusion. If this is a feature that Django users find useful, let them learn about it in the official Django docs. :) Thanks for your contribution, though!

PR #4761, opened 2017-09-29
Use a sample screen shot that uses the default port 80.

kenhtanaka:
### Proposed changes

Added a screen shot called app80-in-browser.png to the get-started tutorial part3. The new image is based on app-in-browser.png but has the port 4000 removed to match the text and configuration of the docker-compose.yml in part3. part3.md uses this new image instead.


get-started/part3.md
```diff
@@ -194,7 +194,7 @@ docker container ls -q
 You can run `curl http://localhost` several times in a row, or go to that URL in
 your browser and hit refresh a few times.
 
-![Hello World in browser](images/app-in-browser.png)
```
mstanleyjones: Shouldn't you also remove the other image? Or is it used somewhere else too?

londoncalling:
@mstanleyjones I like the change, and I like the idea of leaving the existing image there for the moment. 

I'll do more cleanup after this is merged when I've made sure the existing image isn't used elsewhere, but it's a good change. Thanks @kenhtanaka

londoncalling:
@mstanleyjones just to follow up on this, the other image is used in part 2, where we _do_ want to show `localhost:400`, so @kenhtanaka 's PR is perfect as written. Thanks!

PR #4790, opened 2017-10-03
Add code to enable keeping archive.js in master

mstanleyjones:
Follow-up to #4787

This adds logic in `archive.js` that suppresses the butterbar if we are actually in the current version, as opposed by the canonical `archives.json` file in the online published docs. This means we don't have to have a commit like #4787 every time we add a new archive.

It also adds a `site.engine_version` variable to `_config.yml`. I needed it for my logic, but we can rationalize use of version variables and have them all in the top-level file if we want. I'll do that for Engine in a follow-up PR.

Testing this is awkward. You need to check out the PR, download the https://docs.docker.com/js/archives.json into the `js/` folder (replacing the other one) and then do a local build. At first, `master` build won't have a butterbar and you can see a console log that it is suppressed. If you then edit the `archives.json` and change which entry is `current` then it will have the butterbar. Please please don't commit this `archives.json` anywhere but just use it for testing.

cc/ @allejo who helped me.

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -11,7 +11,13 @@ permalink: pretty
 safe: false
 lsi: false
 url: https://docs.docker.com
-keep_files: ["v1.4", "v1.5", "v1.6", "v1.7", "v1.8", "v1.9", "v1.10", "v1.11", "v1.12", "v1.13", "v17.03","v17.06"]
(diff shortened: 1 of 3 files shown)


mstanleyjones:
Actually it's easy to check one of the two conditions (no butterbar) because that works in the Netlify preview.

PR #4798, opened 2017-10-04
Hide things for archives

mstanleyjones:
Relies on and augments #4790. Logical changes:

- Default to assuming it's an archive
- If it's an archive, hide the search div, the normal archive drop-down (because we have the butterbar one), the "Edit this page" link, and the ratings widget.
- If it's not an archive, show the above things.

The reason we default to it being an archive is that it's less visually flickery to hide elements by default and then show them, than to show them and then hide them. I tested both ways. I think this is pretty clean.

Of course it breaks if JS is disabled, but so does the whole site.

To test what happens if it's an archive, set `isArchive = true` in line 296 of `js/docs.js` and do `jekyll serve`.

### DIFF
--- a/js/docs.js
+++ b/js/docs.js
@@ -292,6 +292,26 @@ window.onload = function() {
     var group = $(this).attr('data-group');
     $('.nav-tabs > li > a[data-group="'+ group +'"]').tab('show');
   })
+
+  // isArchive is set by logic in archive.js
+  if ( isArchive == false ) {
+    console.log("Showing content that should only be in the current version.");
+    // Hide elements that are not appropriate for archives
+    // PollDaddy
+    $('#ratings-div').css("visibility","visible");
+    console.log("Ratings widget shown.");
+    // Archive drop-down
+    $('.ctrl-right .btn-group').css("visibility","visible");
+    console.log("Archive widget shown.");
+    // Swarch
+    $('.search-form').css("visibility","visible");
+    console.log("Search widget shown.");
+    // Page edit link
+    $('.feedback-links li').first().css("visibility","visible");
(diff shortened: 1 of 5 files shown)


js/docs.js
```diff
@@ -292,6 +292,26 @@ window.onload = function() {
...
+    console.log("Archive widget shown.");
+    // Swarch
+    $('.search-form').css("display","block");
+    console.log("Search widget shown.");
+    // Page edit link
+    $('.feedback-links li').first().css("display","block");
+    console.log("Page edit link shown.");
+  } else {
+    console.log("Keeping non-applicable elements hidden.");
+  }
```
thaJeztah: An alternative could be to have a css class like;

```css
.is-archive .hide-on-archive { display: none }
```

Then, add the `.is-archive` on a top-level element on the page (`<body class="is-archive">` or `<html class="is-archive">`.

Instead of finding all the elements that should _possibly_ be hidden, you can just remove the `.is-archive` class from the body, and those elements become visible (reducing the risk that elements may/may not be present in a specific version of the archive, and reduces maintenance of the script)

thaJeztah: (each element that needs to be hidden should then have a `class="hide-on-archive"` class)

mstanleyjones: I don't really know enough to implement this. Feel free to commit into this PR if you want to try it, after I address the other feedback.

mstanleyjones:
@johndmulhausen PTAL, I changed it to `visibility`.

mstanleyjones:
Oops I did not mean to actually click that. 🙈 let me know if I should revert.

PR #4805, opened 2017-10-05
Typo-Fix

bxtp4p:
### Proposed changes

Changed `login` to `logic` in the first paragraph. I don't think webhooks are there to support custom logins.


johndmulhausen:
Thanks, @bxtp4p :)

PR #4812, opened 2017-10-05
Clarify that Compose .env support does not work with docker stack deploy

### LINKED ISSUE #3654: Compose file reference doc references `.env` file support
See https://github.com/moby/moby/issues/29133#issuecomment-309429376

https://docs.docker.com/compose/compose-file/#variable-substitution

We should either remove the reference to `.env` from this document or clarify that it's only a `docker-compose` feature.

mstanleyjones:
Fixes #3654 

cc/ @dnephin


mstanleyjones:
Moved and also fixed my typo.

mstanleyjones:
Aha, I misunderstood and thought that whole included file only applied to `docker-compose up`. Fixed.

PR #4813, opened 2017-10-05
Don't print defaults from the YAML in cases where the default is unset

mstanleyjones:
Fixes https://github.com/docker/cli/issues/591


mstanleyjones:
For now, excluding `default` and `map[]`, though we may be able to use the latter at a later time, if it is applied consistently to flags that can be passed more than once.

mstanleyjones:
Merging for now. This de-clutters the view and makes actual interesting defaults stand out.

PR #4838, opened 2017-10-07
Make the link in wordpress.md point to the right engine doc

axilleas:
### Proposed changes

Change to the correct link. Although there's a redirect, it's better to link to to the right place.


johndmulhausen:
Thanks, @axilleas :)

PR #4841, opened 2017-10-08
Translating to pt-BR

Michelhsn:
adding brazilian portuguese translation to documentation.


johndmulhausen:
I am so sorry to do this because I very much appreciate the effort, but pt-BR is not currently a supported language in which we are prepared to officially offer documentation. Please don't let this stop you from providing this information to users where you are able to help them, it's just not something we can publish in the official docs right now.

PR #4851, opened 2017-10-09
"DTR needs that a majority" -> "DTR requires that a majority"

benjamin-issa:
### Proposed changes

Changed the phrase "needs that a majority" to "requires that a majority" in the docker dtr documentation because the existing wording is unclear.


johndmulhausen:
Better! Thanks, @benjamin-issa :)

benjamin-issa:
:)

PR #4855, opened 2017-10-10
Update index.md

scjane:


johndmulhausen:
Yikes, good catch, thanks @scjane 👍

johndmulhausen:
Yikes, good catch, thanks @scjane 👍

PR #4862, opened 2017-10-10
Update license-your-installation.md

scjane:


johndmulhausen:
I think "Settings" is in bold because that's the label text in the UI, so this one stays as is. Thanks though!

scjane:
@johndmulhausen Yes, "Settings" should be in bold, but "page" should not. So, I moved it out of the bold characters :)
page in bold

PR #4863, opened 2017-10-10
Documentation: Remove reference to PUBLISH Dockerfile directive in networking docs

deployable:
### Proposed changes

Remove reference to PUBLISH Dockerfile directive in networking documentation as PUBLISH does not exist in the Dockerfile reference https://docs.docker.com/engine/reference/builder/


johndmulhausen:
Good catch. Just so there's a note about what's possible when using config files, I think we should talk about the `EXPOSE` directive in Dockerfiles here, which are probably what the original writer meant.

thaJeztah:
EXPOSE doesn't publish, so likely needs some extra words in that case

mstanleyjones:
EXPOSE seems pretty useless really. I've been de-emphasizing it. There is PORTS though.

PR #4864, opened 2017-10-10
Update s3.md

scjane:
Should it be "IAM policy" rather than "AIM policy"?


johndmulhausen:
Yep, thanks @scjane 👍

PR #4865, opened 2017-10-10
Update index.md

scjane:
Should "an" be used here?


johndmulhausen:
It should, thanks @scjane 👍

PR #4869, opened 2017-10-10
Cross-linking between compose and configs/secrets

### LINKED ISSUE #4789: Link together Compose's and Engine's documentation relevant to "config"
(Apologies if this has been discussed and nixed before, but...)

It would be great to link Compose's documentation about `config` [1] from this section. E.g., "If you are using Compose, you can define and use configurations from your Compose files, as explained in [1]."

[1] https://docs.docker.com/compose/compose-file/#short-syntax

[engine/swarm/configs.md](https://docs.docker.com/engine/swarm/configs/), CC @mstanleyjones

mstanleyjones:
Fixes #4789 

cc/ @jpetazzo


jpetazzo:
Yay, thanks!

PR #4891, opened 2017-10-11
Add end-to-end backup and upgrade of Docker EE

### LINKED ISSUE #1725: How to create and end-to-end backup of DDC and how to restore
#1722 will talk about creating backups for DTR
#1724 will talk about creating backups for UCP
[These docs ](https://docs.docker.com/engine/swarm/admin_guide/#/back-up-the-swarm) talk how to create a swarm backup

Create an article that leverages all this content and explains how to set up an end-to-end disaster recovery plan for DDC as a whole.

### LINKED ISSUE #1744: Document end-to-end upgrade of DDC
@mbentley commented

### Problem description
When upgrading DDC as a whole platform, moving from Engine 1.12, UCP 2.0.x, and DTR 2.1.x to Entiner 1.13, UCP 2.1.x, and DTR 2.2.x, what is the recommended upgraded path?  What can be done in parallel?  What must be done serially?

This was specifically requested by a user but it really should be documented in our docs for everyone to see.

/cc @vsaraswat @pvnovarese 

### Problem location
n/a

n/a

### Suggestions for a fix
Docs overarching for DDC; specific sections referencing the proper upgrade order in the Engine, UCP, and DTR docs sections.

---

@vsaraswat commented on

Here's the latest:
Engine: Recommend sequential updates, however 1.11 to 1.13 directly is possible
UCP: 1.1.x to 2.1.x is directly possible
DTR: Enforces upgrades from minor version (so only 2.1.x to 2.2.x). 

Looks like we need two things: A DDC upgrade doc (which covers all cases), as well as a special case Engine 1.11 to 1.13 customer success document.

The first one should cover upgrade process of all components--Engine, UCP, DTR--in that order, and require sequential updates of minor versions (eg. DTR 2.1.x to DTR 2.2.x, or Engine 1.12 to 1.13). Best practice is to upgrade using the latest patches  each time as well.

However--several of our customers are upgrading from Engine 1.11 to 1.13 (UCP 1.1.x to UCP 2.1.x) and skipped Engine 1.12. We know that it is okay to upgrade directly from Engine and UCP in this specific use case, and it fact may be preferable if we are worried about stability issues in 1.12. So, we should make a customer success document that specifies doing the jump directly from 1.11 to 1.13 without sequential upgrades, and stress that is okay **only for this specific case**. DTR will still need sequential updates however.

I can work on the 2nd customer success doc. The first one is probably a good candidate for docs day

joaofnfernandes:
Fixes #1725
Fixes #1744


joaofnfernandes:
@JimGalasyn take a look and let me know if anything is confusing or missing.

PR #4892, opened 2017-10-11
Add pointer to latest API ref

### LINKED ISSUE #4775: Link to API reference from main API/SDK page
[develop/sdk/index.md](https://docs.docker.com/develop/sdk/), CC @johndmulhausen

This is the top result for "docker api" on Google, but there is no link to the actual API reference from this page which was the thing I was looking for. Perhaps worth putting a link to https://docs.docker.com/engine/api/v1.32/ in the left hand bar and as a section in this page.

Confusingly, clicking on "reference", then "Docker Engine API", then "Overview" takes me back to this page, again taking me away from the reference which was the thing I actually wanted.

mstanleyjones:
Fixes #4775 

cc/ @bfirsh

### DIFF
--- a/_data/toc.yaml
+++ b/_data/toc.yaml
@@ -1,3 +1,5 @@
+---
+---
 horizontalnav:
 - title: Guides
   path: /
@@ -606,10 +608,12 @@ reference:
     title: Get started
   - path: /engine/api/sdks/
     title: SDKs
-  - path: /engine/api/v1.32/
-    title: v1.32 Reference
-  - sectiontitle: Previous API versions
+  - path: /engine/api/latest/
+    title: v{{ latest_stable_docker_engine_api_version }} Reference
+  - sectiontitle: API reference by version
     section:
+    - path: /engine/api/v1.32/
+      title: v1.32 Reference
     - path: /engine/api/version-history/
       title: Version history overview
     - path: /engine/api/v1.31/
--- a/develop/sdk/index.md
+++ b/develop/sdk/index.md
@@ -44,10 +44,16 @@ go get github.com/docker/docker/client
 
 [Read the full Docker Engine Python SDK reference](https://docker-py.readthedocs.io/).
 
+## View the API reference
+
+You can
+[view the reference for the latest version of the API](/engine/api/latest/)
+or [choose a specific version](/engine/api/version-history/).
+
 ## Versioned API and SDK
 
-The version of the Docker API you should use depends upon the version of your
-Docker daemon and Docker client.
+The version of the Docker Engine API you should use depends upon the version of
+your Docker daemon and Docker client.
 
 A given version of the Docker Engine SDK supports a specific version of the
 Docker Engine API, as well as all earlier versions. If breaking changes occur,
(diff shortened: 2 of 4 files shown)


_data/toc.yaml
```diff
@@ -1,3 +1,5 @@
+---
```
mstanleyjones: This is a hack to get Liquid to expand the variable at line 611, and doesn't seem to have any bad side effect.

_data/toc.yaml
```diff
@@ -606,10 +608,12 @@ reference:
     title: Get started
   - path: /engine/api/sdks/
     title: SDKs
-  - path: /engine/api/v1.32/
-    title: v1.32 Reference
-  - sectiontitle: Previous API versions
+  - path: /engine/api/latest/
+    title: v{{ latest_stable_docker_engine_api_version }} Reference
```
mstanleyjones: The latest stable API will always be featured prominently

_data/toc.yaml
```diff
@@ -606,10 +608,12 @@ reference:
...
+  - sectiontitle: API reference by version
     section:
+    - path: /engine/api/v1.32/
```
mstanleyjones: When we add an API version we only have to put it here. This is no longer "previous" but "all" versioned APIs. The one above is always just "latest", which will also be in this list by number.

develop/sdk/index.md
```diff
@@ -44,10 +44,16 @@ go get github.com/docker/docker/client
 
 [Read the full Docker Engine Python SDK reference](https://docker-py.readthedocs.io/).
 
+## View the API reference
+
+You can
```
mstanleyjones: This is the link @bfirsh asked for.

engine/api/latest/index.html
```diff
@@ -0,0 +1,6 @@
+---
```
mstanleyjones: This file handles the redirect from `/engine/api/latest/` to the real latest version, so that the user always lands on a versioned page.

johndmulhausen: I was thinking if you'd rather not use a redirect you could potentially have this file do:

`{% include_relative {{ latest_stable_docker_engine_api_version }}/index.md %}`

Otherwise this approach LGTM!

thaJeztah: The redirect works, but the API doc itself is a 404 (may be expected; https://deploy-preview-4892--docker-docs.netlify.com/engine/api/v1.32/)

Should we use `latest` or `current` ?

mstanleyjones: Yes, that's expected. I think `latest` is clearer, but I'll wait for what y'all think.

I really don't mind using a redirect. I'd rather the user end up on a versioned page rather than have them bookmark `latest` and then wonder why it changed underneath them.

thaJeztah: I'm +1 for using a redirect as well; it also helps, as people automatically get a "permalink" if they bookmark, or share.

bfirsh:
Yay! ✨

thaJeztah:
@bfirsh It was temporarily reverted as there was an issue, but  I saw a new PR is being worked on

bfirsh:
Oh yes, haha. Glad to see it's on it's way though. ;)

PR #4893, opened 2017-10-11
encrypted attachable networks work

programmerq:
The documentation said that it does not work. It does indeed work, so here is an update that reflects that.

### DIFF
--- a/engine/userguide/networking/overlay-security-model.md
+++ b/engine/userguide/networking/overlay-security-model.md
@@ -35,32 +35,12 @@ automatically rotate the keys every 12 hours.
 
 ## Swarm mode overlay networks and unmanaged containers
 
-Because the overlay networks for swarm mode use encryption keys from the manager
-nodes to encrypt the gossip communications, only containers running as tasks in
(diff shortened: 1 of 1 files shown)


engine/userguide/networking/overlay-security-model.md
```diff
@@ -35,32 +35,12 @@ automatically rotate the keys every 12 hours.
...
+9s1p1sfaqtvaibq6yp7e6jsrt
 ```
 
-Because [swarm mode](../../swarm/index.md) is an optional feature, the Docker
-Engine preserves backward compatibility. You can continue to rely on a
-third-party key-value store to support overlay networking if you wish.
-However, switching to swarm-mode is strongly encouraged. In addition to the
-security benefits described in this article, swarm mode enables you to leverage
-the substantially greater scalability provided by the new services API.
+Just like services that are attached to an encrypted network, regular containers can also have the advantage of encrypted traffic when attached to a network created this way.
```
mstanleyjones: Do you want to point out that the containers are still just plain containers and only managed by their specific Docker host? Wouldn't want people to get confused about this.

PR #4897, opened 2017-10-11
Add pointer to latest API ref

mstanleyjones:
cc/ @johndmulhausen take 2

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -27,7 +27,7 @@ ENV ENGINE_BRANCH="17.06.x"
 ENV DISTRIBUTION_SVN_BRANCH="branches/release/2.6"
 ENV DISTRIBUTION_BRANCH="release/2.6"
 
-RUN sh md_source/_scripts/fetch-upstream-resources.sh \
+RUN ./md_source/_scripts/fetch-upstream-resources.sh \
 	&& jekyll build -s md_source -d target --config md_source/_config.yml \
 	&& rm -rf target/apidocs/layouts \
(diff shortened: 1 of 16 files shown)


mstanleyjones:
How in the world did this work before? I can't get it to work now. Weird.

cc/ @allejo thoughts?

allejo:
@mstanleyjones So in YAML, `---` has a special meaning, it's how to define the "start" of a document. I did some testing and my guess is that two lines of `---` in a YAML data file is treated as two "documents" within the same file and not as Jekyll's Front Matter (possibly because there shouldn't be any Liquid templates inside the `_data` folder, just data?). As for how it probably worked the first time, something might have been cached in the process and it just seemed to work.

mstanleyjones:
It works now! 💃 

To test, you need to check out the branch, run `docker build -t mytest .`, run the container locally, then look at the TOC.

allejo:
This is what I see

https://deploy-preview-4897--docker-docs.netlify.com/reference/

image

mstanleyjones:
@allejo can you do `jekyll clean` and try again? It is working for me:

screen shot 2017-10-16 at 11 08 24 am

mstanleyjones:
And that screenshot made me see that we are using title case in that TOC entry. I'll update that.

mstanleyjones:
OK, had to tweak the Dockerfile a bit more but all the tests are now running cleanly and we probably won't break things. Merging!

mstanleyjones:
@bfirsh 🙌

PR #4900, opened 2017-10-12
Update scan-images-for-vulnerabilities.md

scjane:


johndmulhausen:
Indeed, a step was missed. Thanks, @scjane 👍 cc @joaofnfernandes

PR #4902, opened 2017-10-12
Update reconfigure.md

scjane:


johndmulhausen:
Wow, much better, many good catches here. Thanks, @scjane 👍 cc @joaofnfernandes

PR #4907, opened 2017-10-12
Psql exits with \q not \l

danidee10:
### Proposed changes

Psql exits with \q not \l without this, the sample startup script with "wait forever" even though Postgres is up.

### DIFF
--- a/compose/startup-order.md
+++ b/compose/startup-order.md
@@ -63,7 +63,7 @@ script:
         shift
         cmd="$@"
 
-        until psql -h "$host" -U "postgres" -c '\l'; do
+        until psql -h "$host" -U "postgres" -c '\q'; do
           >&2 echo "Postgres is unavailable - sleeping"
           sleep 1
         done


danidee10:
I also whipped out a simple Python script showing how you could do this without `psql` in python. 

The base Python (and a lot programming languages if not all) docker image don't have postgres and therefore `psql` installed. Personally i wouldn't want to install postgres in my web app's docker image just to test a database connection. I'll rather find a way to do it purely from the Language itself.

I believe it can be of great use if this Script is also added to the documentation

#!/usr/bin/env python

"""Test if Postgres instance is up and running"""

```python
import sys
import subprocess
import time
import psycopg2

host = sys.argv[1]
cmd = sys.argv[2:]

while(True):
    try:
        con = psycopg2.connect(
            "dbname='xxx' user='xxx' host=%s password='xxx" % host
        )
        print('postgres is up, Executing command...')
        con.close()
        subprocess.call(cmd)
    except psycopg2.OperationalError as error:
        print('Postgres is not ready yet, sleeping for a sec...', error)
        time.sleep(1)
        continue
    break
```

mstanleyjones:
@shin- PTAL

joaofnfernandes:
There might be a bit of confusion here. When you say:

> It's not an exact Replica, but it even does a better job because it actually connects to the Database

When the sample runs `psql -h "$host" -U "postgres" -c '\l'` it is listing all schemas that exist in that Postgres deployment. And for that it needs to connect to the database, right?
If your concern is with the exit code of `-c`, I also checked the `psql` and probably you're OK either by quitting, listing schemas, or running some other command.

```
-c <command>
Specifies that psql is to execute the given command string, command.
This option can be repeated and combined in any order with the -f option.
When either -c or -f is specified, psql does not read commands from standard input;
instead it terminates after processing all the -c and -f options in sequence.
```

I agree with you. Few images will have `psql` installed, but the same happens with Python too. If I want to use Alpine as my base image, Python won't be available. It also seems that `psycopg2` is not a standard library, so you'd have to install that dependency too.
In the end, this logic is always going to be specific to the database you're trying to deploy, so I think this doc page should give you some pointers, and maybe some examples, but it should not be tied to any specific database.

@danidee10 I'm going to close this PR, but let me know if you have any ideas on how to improve this doc while still keeping it generic.

The docs where also a bit sparse in some aspects, so I had to [create my own sample](https://github.com/joaofnfernandes/compose-startup) to test your suggestions. Maybe we could leverage it in the docs?

danidee10:
Thanks for the response @joaofnfernandes 

> When the sample runs psql -h "$host" -U "postgres" -c '\l' it is listing all schemas that exist in that Postgres deployment. And for that it needs to connect to the database, right?

It doesn't actually connect to a specific "database" listing the Schema only shows that that the Postgres instance is up and running (Which is fine in most cases). But a particular database you might be interested in might be down.
It's possible for a Postgres Database to be listed even though it's down (Not accepting connections for some weird reason)
The only way to be 100% sure is to actually connect to the said database. A simple way to do that would be:

 `psql -c "\c $database"`

Also i can confirm that `psql` doesn't quit with the `\l` command (I tried it again to be sure). The `\l` command expects input from the user, so it doesn't actually follow what the man pages say. The behaviour is only true when you execute commands that don't expect any input from the user. So i believe my PR is still valid because it solves this.

> In the end, this logic is always going to be specific to the database you're trying to deploy, so I think this doc page should give you some pointers, and maybe some examples, but it should not be tied to any specific database.

Honestly I don't think that's possible. Lol :smile:  (For an example not to be tied to any database or Technology) The docs itself is already specific to Postgres.

joaofnfernandes:
If you're seeing `\l` hang, then let's merge your PR. Thanks for investing the time to explain this tom me 🚀

PR #4959, opened 2017-10-13
Add info about network: host

### LINKED ISSUE #4593: network "host" is missing
[compose/compose-file/index.md](https://docs.docker.com/compose/compose-file/), CC @gbarr01

With >= 17.06 it is possible to use the Host Network within Containers. This is also true in swarm mode.

See:
https://github.com/docker/compose/issues/5039 
https://github.com/moby/moby/issues/25873#issuecomment-319109840

On this page, this information is missing.
The sections
```
network-mode:
"This option is ignored when deploying a stack in swarm mode with a (version 3) Compose file."
```
and
https://docs.docker.com/compose/compose-file/#network-configuration-reference
keep silent about that.

mstanleyjones:
Fixes #4593 

@shin- PTAL

### DIFF
--- a/compose/compose-file/compose-file-v2.md
+++ b/compose/compose-file/compose-file-v2.md
@@ -1415,6 +1415,49 @@ refer to it within the Compose file:
         external:
           name: actual-name-of-network
 
+#### host or none
+
+Use the host's networking stack, or no networking. Equivalent to
+`docker run --net=host` or `docker run --net=none`. Only used if you use
+`docker stack` commands. If you use the `docker compose` command,
+use [network_mode](#network_mode) instead.
+
+The syntax for using built-in networks like `host` and `none` is a little
+different. Define an external network with the name `host` or `none` (which
+Docker has already created automatically) and an alias that Compose can use
+(`hostnet` or `nonet` in these examples), then grant the service access to that
+network, using the alias.
+
+```yaml
+version: 2
+
+services:
+  web:
+    ...
+    networks:
+      hostnet: {}
+
+networks:
+  hostnet:
+    external:
+      name: host
+```
+
+```yaml
+version: 2
+
+services:
+  web:
+    ...
+    networks:
+      nonet: {}
+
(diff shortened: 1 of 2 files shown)


shin-:
@dnephin should probably take a look.

dnephin:
There was some support added for `network: host` but it does not use this field. It should use something like this, which I believe is also supported in v2:

```yaml
services:
  web:
    ...
    networks: [hostnet]

networks:
  hostnet:
    external:
      name: host
    
```

mstanleyjones:
Thanks @dnephin I have attempted to update the info to be accurate. Can you take another look?

compose/compose-file/compose-file-v2.md
```diff
@@ -1415,6 +1415,49 @@ refer to it within the Compose file:
         external:
           name: actual-name-of-network
 
+#### host or none
+
+Use the host's networking stack, or no networking. Equivalent to
+`docker run --net=host` or `docker run --net=none`. Only used if you use
+`docker stack` commands. If you use the `docker compose` command,
```
dnephin: `docker-compose` (not `docker compose`)

Since this is the reference for v2, and v2 is not supported by `docker stack` commands, do we need this here? It seems like it's only appropriate for the v3 reference.

compose/compose-file/index.md
```diff
@@ -1942,6 +1942,45 @@ discovery](https://github.com/docker/labs/blob/master/networking/A3-overlay-netw
 networking concepts lab on the [Overlay Driver Network
 Architecture](https://github.com/docker/labs/blob/master/networking/concepts/06-overlay-networks.md).
 
+#### host or none
+
+Use the host's networking stack, or no networking. Equivalent to
+`docker run --net=host` or `docker run --net=none`. Only used if you use
+`docker stack` commands. If you use the `docker compose` command,
```
dnephin: `docker-compose`

PR #4965, opened 2017-10-13
Deploying Registry: Fixed ports on HTTPS

### LINKED ISSUE #4963: Ports incorrect on deploying a registry with TLS
### Problem description

The docker cli/daemon uses port `443` on default when tagging a image with a registry without specifying a port. like:

```bash
$ docker tag alpine registry.someserver.com/alpine
$ docker push registry.someserver.com/alpine
``` 
pushes to `registry.someserver.com:443` but the docs suggests it uses port `80`. 

### Problem location

URL: https://docs.docker.com/registry/deploying/#get-a-certificate at point 3. and 4.

I've tested it on a blank client and host each on ubuntu 16.04 running docker 17.09-0-ce

### Suggestions for a fix

We could either change the example for tagging and pushing under point 4. or change the port in the example under point 3. to `443` which is the default for HTTPS. I've created a PR for the second solution.

AnianZ:
### Related issues (optional)

Fixes #4963

### DIFF
--- a/registry/deploying.md
+++ b/registry/deploying.md
@@ -210,17 +210,17 @@ If you have been issued an _intermediate_ certificate instead, see
 3.  Restart the registry, directing it to use the TLS certificate. This command
     bind-mounts the `certs/` directory into the container at `/certs/`, and sets
     environment variables that tell the container where to find the `domain.crt`
-    and `domain.key` file. The registry runs on port 80.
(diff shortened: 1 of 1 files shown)


AnianZ:
Relates to issue #4963

mstanleyjones:
Thanks!

PR #4978, opened 2017-10-16
Update stack-yaml-reference.md

scjane:

### DIFF
--- a/docker-cloud/apps/stack-yaml-reference.md
+++ b/docker-cloud/apps/stack-yaml-reference.md
@@ -93,7 +93,7 @@ deployment_strategy: high_availability
 ```
 
 ## devices
-List of device mappings. Use the same format as the `--device` docker client create option.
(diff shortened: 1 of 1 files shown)


docker-cloud/apps/stack-yaml-reference.md
```diff
@@ -93,7 +93,7 @@ deployment_strategy: high_availability
 ```
 
 ## devices
-List of device mappings. Uses the same format as the `--device` docker client create option.
+List of device mappings. Use the same format as the `--device` docker client create option.
```
mstanleyjones: This is incorrect. I know it's an incomplete sentence, but so is "List of device mappings". :) English is weird.

scjane: @mstanleyjones I reverted it, and changed the other one similar to this to "Uses", too. But it's still a little weird. Please review it. Thanks.

PR #4989, opened 2017-10-16
env proxy format.

### LINKED ISSUE #4970: proxy spec is wrong
[get-started/part2.md](https://docs.docker.com/get-started/part2/), CC @johndmulhausen 
This doc uses a weird syntax for setting a proxy using ENV.

It asks to define host:/port (so in my example here it would be "http_proxy 192.168.x.z:/3128")
Normal and/or correct is the following.
ENV http_proxy http://192.168.x.z:3128
ENV https_proxy http://192.168.x.z:3128

konstruktoid:

### Proposed changes

Modify section about proxies.

### Related issues (optional)

Fixes #4970

### DIFF
--- a/get-started/part2.md
+++ b/get-started/part2.md
@@ -92,9 +92,9 @@ CMD ["python", "app.py"]
 > proxy servers:
 >
 > ```conf
-> # Set proxy server, replace host:/port with values for your servers
-> ENV http_proxy host:/port
-> ENV https_proxy host:/port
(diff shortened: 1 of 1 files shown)


konstruktoid:
Believe it to be fixed.

mstanleyjones:
Somehow you still had the original commit and added an empty one and a merge one. Not sure how! Anyway I checked out your branch, removed the empty commits, and amended the original commit to change the commit message. Should be fine now after Netlify builds.

PR #4991, opened 2017-10-16
improve links from v2.1

gotcha:
- fix link_local_ips and pids_limit
- more specific links to volume and network labels, volume name

### Proposed changes

improve links from v2.1, some where broken, other too fuzzy


londoncalling:
@gotcha LGTM, thanks for the fixes!

Issue #4871, opened 2017-10-11
Outdated docker-machine release notes 


jamshid:
### Problem description

Outdated release notes seem to be on https://docs.docker.com/release-notes/docker-machine/ (0.10.0), newer ones on https://github.com/docker/machine/releases/ (0.12.2). 

### Project version(s) affected

docker-machine 0.12.2

### Suggestions for a fix

Would be good for release notes to be complete everywhere they're shown and to be available in one consistent place (across docker projects and flavors).

FWIW I prefer github.

mstanleyjones:
@shin- I thought you were going to submit PRs for these? Did this slip through the cracks?

gbarr01:
@mstanleyjones , would you confirm that this is done?

mstanleyjones:
Yep! Should be published shortly.

PR #5012, opened 2017-10-17
Add Machine 0.13.0 release notes

shin-:
### Proposed changes

Updated Docker Machine release notes

0.13.0

### Related issues (optional)

#4871


mstanleyjones:
There is one more change that needs to be made with these now. Update the `machine_version` variable in the top-level `_config.yaml` file. Thanks! (this will also need to be done for Compose now)

shin-:
:+1: updated

PR #5047, opened 2017-10-19
Add alt text to images which are missing it

### LINKED ISSUE #4998: Add alt text to all images which are missing it
### Problem description

For accessibility, it is good for all images to have alt text.

Do a search for the following string and add the alt text into the empty brackets:

`
![](
`

For instance, an image with missing alt text might look like:

```
![](foo.png)
```

And to fix it, you'd change it to something like:

```
![an example of foo](foo.png)
```

it's best if the alt text is useful to someone who can't actually see the image, but may be using a screen reader or other assistive text. So don't just repeat the image name or something like that.

This is not a difficult task but will take some time due to considering the right alt text to use for each image.

bhavin192:
### Proposed changes
Add alt text to images wherever it is missing.
Still WIP
### Related issues (optional)
Fixes #4998

### DIFF
--- a/apidocs/overview.md
+++ b/apidocs/overview.md
@@ -12,4 +12,4 @@ DTR repositories, permissions, and settings.
 To get access to interactive documentation, in your **DTR UI**, click
 on the **top-right menu** and choose **API docs**.
 
-![](dtr-api-overview.png)
+![API docs page, DTR UI](dtr-api-overview.png)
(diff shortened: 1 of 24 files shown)


mstanleyjones:
This is looking great so far! 👏

mistyhacks:
Let us know when you would like this reviewed. 🥇

mistyhacks:
@bhavin192 this is great and I am going to go ahead and merge it. Please feel free to open a new PR for any other images that need alt text. Thank you so much for making our docs more accessible!

PR #5050, opened 2017-10-19
Document that overlay2 is now the default for ubuntu

### LINKED ISSUE #4141: Update storage-driver for promotion of overlay2
### Problem description

https://github.com/moby/moby/pull/34430 updated the list of storage-drivers and promoted overlay2 over aufs.

As a result, aufs will no longer be selected automatically if the required kernel modules are installed to support it. For comparison;

Before this PR (taking Ubuntu 16.04 as example, because it always has kernel 4+)

- When installing docker on a Ubuntu 16.04 machine with the `linux-image-extra` package installed (which provides `aufs`), docker would select `aufs` as storage driver.
- When installing docker on a Ubuntu 16.05 machine _without_ the `linux-image-extra` package installed, docker would select `overlay2`

After this PR, `overlay2` will be selected as the default, _unless_ an existing `aufs` directory was found (so users that are currently running `aufs` will stay on `aufs` unless they do a "factory reset" and wipe out their `/var/lib/docker` directory)

### Problem location

Most likely this section needs an update; https://docs.docker.com/engine/userguide/storagedriver/selectadriver/

This change will be included in the next Docker release that is using the current master of the moby/moby repository (17.08)

mstanleyjones:
Fixes #4141 

This will need follow-up to the Ubuntu EE install instructions when the next version of EE comes out. cc/ @JimGalasyn

### DIFF
--- a/engine/userguide/storagedriver/aufs-driver.md
+++ b/engine/userguide/storagedriver/aufs-driver.md
@@ -4,20 +4,14 @@ keywords: 'container, storage, driver, AUFS '
 title: Use the AUFS storage driver
 ---
 
-AUFS a *union filesystem*. The `aufs` storage driver is the default storage
-driver used for managing images and layers on Docker for Ubuntu, and for
-Debian versions prior to Stretch. For Debian Stretch,
-[overlay2](overlayfs-driver.md) is the default.
+AUFS a *union filesystem*. The `aufs` storage driver was previously the default
+storage driver used for managing images and layers on Docker for Ubuntu, and for
+Debian versions prior to Stretch. If your Linux kernel is version 4.0 or higher,
+and you use Docker CE, consider using the newer
+[overlay2](overlayfs-driver.md){: target="_blank" class="_" }, which has
+potential performance advantages over the `aufs` storage driver.
 
-AUFS is the most mature storage driver used by Docker. It provides fast
-container startup time, as well as efficient usage of memory and storage.
-
-If your Linux kernel is version 4.0 or higher, and you use Docker CE, consider
-using the newer [overlay2](overlayfs-driver.md){: target="_blank" class="_" },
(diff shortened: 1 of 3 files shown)


mstanleyjones:
Yep! That's old info. I mean it is "most mature" but that's not really relevant anymore.

engine/userguide/storagedriver/aufs-driver.md
```diff
@@ -4,7 +4,7 @@ keywords: 'container, storage, driver, AUFS '
 title: Use the AUFS storage driver
 ---
 
-AUFS a *union filesystem*. The `aufs` storage driver was previously the default
+AUFS is a *union filesystem*. The `aufs` storage driver was previously the default
```
mstanleyjones: That's not an article, it's a verb! 🤓

gbarr01: Ha.

engine/userguide/storagedriver/selectadriver.md
```diff
@@ -33,20 +30,22 @@ this decision, there are three high-level factors to consider:
...
       environments, because `loopback-lvm`, while zero-configuration, has very
       poor performance.
 
-  The selection order is defined in Docker's source code. You can see the order
-  for Docker 17.03 by looking at
-  [the source code](https://github.com/moby/moby/blob/v17.03.1-ce/daemon/graphdriver/driver_linux.go#L54-L63).
-  For a different Docker version, change the URL to that version.
+  The selection order is defined in Moby's source code. You can see the order
+  by looking at
+  [the source code](https://github.com/docker/docker-ce/blob/master/components/engine/daemon/graphdriver/driver_linux.go#L54-L63).
```
thaJeztah: Might be good to change this link to point to a specific commit or tag, as "master" can change at any time

PR #5077, opened 2017-10-23
Update git submodules to be able to fetch from China

magic-chenyang:
When I use the command "git clone - recursive", my project will misreport, stating "Failed to connect to go.googlesource.com port 443: Connection timed out". In mainland China will appear this problem, I suggest to use "https://github.com/golang/net" instead of the previous site, I have success in the local.


joaofnfernandes:
@aduermael can you take a look at 7a5c4f3c1494fc2246863d739b13160bddbe8666 (disregard the other commits, since it seems that this PR is using an old base)

johndmulhausen:
@magic-chenyang I don't have a problem with the change to the submodules but this PR is full of changes to over 200 other files. Feel free to rebase and send another PR that only changes the submodules.

magic-chenyang:
@johndmulhausen  It's all right

PR #5080, opened 2017-10-23
Update docker-overview.md

anilgkurian:
### Proposed changes

Since 'container run' is 'container create' + 'container start' the documentation is confusing when we say "You can create, **run**, stop, move, or delete a container using the Docker API or CLI."


joaofnfernandes:
@anilgkurian thanks for helping make the docs better. Let's ship it 🚀

PR #5099, opened 2017-10-24
Improve the link to source code for storage drivers

mstanleyjones:
Follow-up to #5050, improves links to source  code for storage drivers.

cc/ @thaJeztah 

(This will drift if the line number ranges change but I think it's close enough)

### DIFF
--- a/engine/userguide/storagedriver/selectadriver.md
+++ b/engine/userguide/storagedriver/selectadriver.md
@@ -36,11 +36,11 @@ this decision, there are three high-level factors to consider:
       environments, because `loopback-lvm`, while zero-configuration, has very
       poor performance.
 
-  The selection order is defined in Moby's source code. You can see the order
+  The selection order is defined in Docker's source code. You can see the order
   by looking at
-  [the source code](https://github.com/docker/docker-ce/blob/master/components/engine/daemon/graphdriver/driver_linux.go#L54-L63).
-  The order may vary slightly by version. Select the branch for the version you
-  use to verify the storage driver order.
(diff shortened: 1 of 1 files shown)


engine/userguide/storagedriver/selectadriver.md
```diff
@@ -36,11 +36,13 @@ this decision, there are three high-level factors to consider:
...
 
-  The selection order is defined in Moby's source code. You can see the order
+  The selection order is defined in Docker's source code. You can see the order
   by looking at
-  [the source code](https://github.com/docker/docker-ce/blob/master/components/engine/daemon/graphdriver/driver_linux.go#L54-L63).
-  The order may vary slightly by version. Select the branch for the version you
-  use to verify the storage driver order.
+  [the source code for {{ site.docker_ce_stable_version }}](https://github.com/docker/docker-ce/blob/{{ site.docker_ce_stable_version }}/components/engine/daemon/graphdriver/driver_linux.go#L54-L63)
+  or
+  [the source code for {{ site.docker_ee_version }}](https://github.com/docker/docker-ee/blob/{{ site.docker_ee_version }}/components/engine/daemon/graphdriver/driver_linux.go#L54-L63).
```
thaJeztah: This one won't work, because that's a private repository. Note that the line numbers may of course change between releases (so the easy way would be to just link to a commit, and explain it's just for illustration)

mstanleyjones: Oh, I didn't realize the repo was private and the link worked for me when I tried it because I have access. :/

PR #5100, opened 2017-10-24
Increase max TOC level for installation topics

mstanleyjones:
cc/ @JimGalasyn @thaJeztah @friism


mstanleyjones:
See https://deploy-preview-5100--docker-docs.netlify.com/engine/installation/linux/docker-ee/centos/ for example.

PR #5103, opened 2017-10-24
Fix archive butterbar margin-top

### LINKED ISSUE #4837: Archive bar obscured on 17.03, 17.06
https://docs.docker.com/v17.06/engine/docker-overview/

You can see it peeking out from under the top nav. It might be gnarly to fix this because the top nav bar is set to `position:fixed`; you might have to resort to frames or something.

mstanleyjones:
Fixes #4837 for 17.06


mstanleyjones:
cc/ @seemethere any ideas?

seemethere:
Appears to be failing at this line:

https://github.com/mstanleyjones/docker.github.io/blob/88d98ef3a9fa6cd76228baa2021a2cb9c885f3fd/Jenkinsfile#L18

`test` directory does not seem to exist within this revision. Maybe it needs a rebase?

seemethere:
Yeah this `Jenkinsfile` doesn't seem to be built to work with the particular base branch you're trying to push to.

mstanleyjones:
@jsouth this has the JS / CSS fixes we suggested (I think). PTAL

mstanleyjones:
Jenkins failure will be solved by an upstream commit that hasn't hit yet.

PR #5089, opened 2017-10-24
Document order option in update_config

hairyhenderson:
### Proposed changes

Adding some documentation for the `order` option in `update_config` in the Docker Compose v3.4 format - new in Docker CE 17.09.

### Related issues (optional)

Fixes #5088

Signed-off-by: Dave Henderson <dhenderson@gmail.com>

Issue #5088, opened 2017-10-24
Missing docs for new update_config order option


hairyhenderson:
### Problem description

In Docker 17.09, support was added for a new `order` option in the `update_config` section in the Docker Compose v3.4 file.

See https://github.com/docker/cli/pull/360 for details.

### Problem location

- I couldn't find the information I wanted. I expected to find it near the following URL: https://docs.docker.com/compose/compose-file/#update_config

hairyhenderson:
Here's a fix for this: https://github.com/docker/docker.github.io/pull/5089

londoncalling:
@shin- @gbarr01 I'm working on a contribution to @hairyhenderson PR, but as I look at it more closely, we need to add the following:

- add a section in [about versions and upgrading](https://docs.docker.com/compose/compose-file/compose-versioning/#version-33) under `Versioning` about `Version 3.4` (I'm doing this now)

- add examples and mention of all the new `v3.4` supported parameters to their associated topics in the [Compose file reference](https://docs.docker.com/compose/compose-file/).

shin-:
Version 3.4 was introduced in Docker Engine 17.09.0, but it hasn't made it yet in a Compose release. Those changes would have ideally been made at that time by whoever is responsible for documenting Engine features.

I did have a versioning section written out and ready to go in `vnext-compose` for when we release Compose 1.17.0 (https://github.com/docker/docker.github.io/pull/5019), as well as documentation for extension fields.
- For the fields in the `build` section, they're no-ops with `docker stack` (it ignores the `build` section entirely), but they should be documented in the [v2 reference](https://docs.docker.com/compose/compose-file/compose-file-v2/#build) and examples can probably be copied over.
- Volume name is also documented in the [v2 reference](https://docs.docker.com/compose/compose-file/compose-file-v2/#name).

mstanleyjones:
Oh no, maybe I should revert that merge....

hairyhenderson:
@mstanleyjones since the fix has been reverted, can this bug be reopened? It's _not ideal_ that the option remains undocumented for Swarm users...

mstanleyjones:
It is documented in the `vnext-compose` branch, which will be merged with `master` when Compose is released. cc/ @shin- in case I am misunderstanding.

hairyhenderson:
Ok, just to be clear, I didn't run into the lack of documentation while using `docker-compose`, but while using `docker stack deploy`.

[...]

PR #5105, opened 2017-10-25
Revert "Document order option in update_config"

mstanleyjones:
Reverts docker/docker.github.io#5089

### DIFF
--- a/compose/compose-file/compose-versioning.md
+++ b/compose/compose-file/compose-versioning.md
@@ -251,20 +251,6 @@ Introduces the following additional parameters:
 - [`configs`](/compose/compose-file/index.md#configs)
 - [deploy `endpoint_mode`](/compose/compose-file/index.md#endpointmode)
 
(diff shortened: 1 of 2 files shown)


hairyhenderson:
@mstanleyjones was something wrong with #5089? why does it need to be reverted?

hairyhenderson:
oh, I see the comment https://github.com/docker/docker.github.io/issues/5088#issuecomment-339169200 now...

thaJeztah:
hm, yes, this is getting confusing now, as the specs in the docker/cli repository are considered the source of truth for the format, but the docs are tied to the docker compose release.

/cc @dnephin

dnephin:
The specs in `docker/cli` are the source of truth for v3.x, the specs in `docker/compose` are the source of truth for v1 and v2.x. 

I don't know how the docs should be versioned, it doesn't necessarily fit the release cycle of either product.

PR #5109, opened 2017-10-25
Avoid Liquid warning

DirtyF:
This page contains curly braces and Liquid interpreter generates a warning at build:

```
Liquid Warning: Liquid syntax error (line 14): [:dot, "."] is not a valid expression in "{{.Architecture}}" in datacenter/ucp/2.2/guides/admin/install/architecture-specific-images.md
```

> Jekyll processes all Liquid filters in code blocks
> If you are using a language that contains curly braces, you will likely need to place {% raw %} and {% endraw %} tags around your code.


mstanleyjones:
Added @londoncalling as a reviewer because she just noticed this yesterday too.

PR #5114, opened 2017-10-25
Minor edit to docs

miketwo:
### Proposed changes

Using verbose args in example makes it easier to grok


joaofnfernandes:
@miketwo let's ship this 🚀

PR #5145, opened 2017-10-27
Fix daemon command, and improve order of examples

thaJeztah:
Spotted this when @jpetazzo linked to this article (I think he's planning to make some improvements to this document 👍 🤗 )

Updated the example to use `dockerd` instead of `docker -d` (which is deprecated, and I think no longer functional).

Also moved the `daemon.json` example before the command-line flag example, because using a daemon.json configuration file is the preferred way to configure the daemon, and allows changing this configuration without _restarting_ the daemon (just `systemctl reload docker` / sending a `SIGHUP` should do)


thaJeztah:
ping @mistyhacks 😇

PR #5149, opened 2017-10-27
Clarify where signing and verification happen

### LINKED ISSUE #2715: Documentation: unclear how verification is happening
There is a lack of documentation (section) which describes in details how signature verification is happening and where.

It, probably, should mention:

1. Metadata containing signature and public key of signing file;
2. Initial verification happening in Notary;
3. If verification in notary doesn't succeed then image is NOT loaded;
4. If initial verification succeeds then the image is loaded and verified against SHA. 

It is also unclear if there is a second-time verification of target machine to verify signature again in addition to SHA on the image.

Originally from here: https://github.com/docker/notary/issues/1132

### LINKED ISSUE #2716: Documentation: it is unclear where image is signed originaly
Neither the architecture doc, not TUF make it clear that image is signed on originating machine, NOT in notary server.

Moved from: https://github.com/docker/notary/issues/1131

mistyhacks:
Fixes #2715 
Fixes #2716 

cc/ @riyazdf @endophage PTAL

### DIFF
--- a/engine/security/trust/content_trust.md
+++ b/engine/security/trust/content_trust.md
@@ -11,7 +11,7 @@ operates on. You use Docker Engine to push and pull images (data) to a public or
 gives you the ability to verify both the integrity and the publisher of all the
 data received from a registry over any channel.
 
-## Understand trust in Docker
+## About trust in Docker
 
 Content trust allows operations with a remote Docker registry to enforce
 client-side signing and verification of image tags. Content trust provides the
@@ -29,6 +29,11 @@ Once content trust is enabled, image publishers can sign their images. Image con
 ensure that the images they use are signed. Publishers and consumers can be
 individuals alone or in organizations. Docker's content trust supports users and
 automated processes such as builds.
+
+When you enable content trust, signing occurs on the client before push and
+verification happen on the client after pull if you use Docker CE. If you use
+Docker EE with UCP, and you have configured UCP to require images to be signed
+before deploying, signing is verified by UCP.
 
 ### Image tags and content trust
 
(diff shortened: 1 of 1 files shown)


engine/security/trust/content_trust.md
```diff
@@ -30,6 +30,11 @@ ensure that the images they use are signed. Publishers and consumers can be
 individuals alone or in organizations. Docker's content trust supports users and
 automated processes such as builds.
 
+When you enable content trust, signing occurs on the client before push and
+verification happen on the client after pull if you use Docker CE. If you use
```
riyazdf: `happens`

engine/security/trust/content_trust.md
```diff
@@ -30,6 +30,11 @@ ensure that the images they use are signed. Publishers and consumers can be
 individuals alone or in organizations. Docker's content trust supports users and
 automated processes such as builds.
 
+When you enable content trust, signing occurs on the client before push and
```
endophage: Technically signing happens after push. This ensures we sign the thing that got accepted by the registry.

mstanleyjones: So we allow the image to be pushed, then the registry tell us that it got pushed, then the client signs the **local**  notion of it and re-pushes that? Could it change after pushing but before signing such that the client and server have different things they think are the same?

endophage: It could change, but what the server thinks is irrelevant. Notary signing is designed such that a completely malicious registry _still_ can't make me trust something I shouldn't. The main benefit of signing after pushing is in case something goes wrong in the push.

PR #5157, opened 2017-10-27
WIP: Modify Rails example to use secrets

### LINKED ISSUE #3118: Update Rails sample to use secrets
Rails 5.1 supports encrypted secrets, so we can update https://docs.docker.com/compose/rails/
to leverage secrets

mistyhacks:
Fixes #3118 

@shin- I need some advice on how to get the database secret into the Rails config file here. Is the best thing to hand-craft the Rails config file with the secret that we create and then store the whole Rails config as another secret? It seems messy for rolling the secret.

cc/ @riyazdf

### DIFF
--- a/compose/rails.md
+++ b/compose/rails.md
@@ -5,8 +5,8 @@ title: "Quickstart: Compose and Rails"
 ---
 
 This Quickstart guide will show you how to use Docker Compose to set up and run
-a Rails/PostgreSQL app. Before starting, you'll need to have [Compose
-installed](install.md).
+a Rails/PostgreSQL app. Before starting, you'll need to have
+[Compose installed](install.md).
 
 ### Define the project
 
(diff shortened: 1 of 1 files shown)


compose/rails.md
```diff
@@ -16,50 +16,86 @@ dependencies, you'll need to define exactly what needs to be included in the
...
 
 You'll need an empty `Gemfile.lock` in order to build our `Dockerfile`.
 
-    touch Gemfile.lock
+```bash
+touch Gemfile.lock
+```
+
+Next, create a [secret](/engine/swarm/secrets.md) to store the database
+password. This example uses a semi-random string using the `openssl` command,
```
riyazdf: super nit: I think we used the phrasing "relatively random" to get around having to explain pseudo-randomness in the original secrets docs

compose/rails.md
```diff
@@ -16,50 +16,86 @@ dependencies, you'll need to define exactly what needs to be included in the
...
-        volumes:
-          - .:/myapp
-        ports:
-          - "3000:3000"
-        depends_on:
-          - db
+to link them together and expose the web app's port. In addition, the secret
+used to connect to the database is stored in a
+[secret](/engine/swarm/secrets.md) so that it does not need to be stored in
+clear text. The Rails configuration is also stored in a secret
```
riyazdf: add closing punctuation here?  (Either `:` or `.`)

compose/rails.md
```diff
@@ -16,50 +16,86 @@ dependencies, you'll need to define exactly what needs to be included in the
...
-      web:
-        build: .
-        command: bundle exec rails s -p 3000 -b '0.0.0.0'
-        volumes:
-          - .:/myapp
-        ports:
-          - "3000:3000"
-        depends_on:
-          - db
+to link them together and expose the web app's port. In addition, the secret
```
riyazdf: "the secret [...] is stored in a secret" is a little confusing when reading.

Maybe this sentence should instead be something describing the compose directive, like: "In addition, the database password secret is specified in a `secrets` section so that it does not need to be stored in clear text" ?

riyazdf:
@mstanleyjones: as for your question about rolling the secrets into config, perhaps we can do something similar to the [nginx example](https://docs.docker.com/engine/swarm/configs/#advanced-example-use-configs-with-a-nginx-service) ?

PR #5170, opened 2017-10-30
Added reference to the proxy setup in part 2 to part 1 of tutorial

LawnboyMax:
### Proposed changes

Added reference to the next part of the tutorial in case there is an issue running `docker run hello-world`. This was an issue for me. As soon as I couldn't run the command successfully I went on searching what might cause the problem. Would have saved me quite some time if I knew that the next part contains info on proxy settings.


joaofnfernandes:
Let's ship it 🚀

PR #5172, opened 2017-10-30
Run only signed images change

vsaraswat:
From https://github.com/docker/orca/issues/10298#issuecomment-340592917 it turns out the "run only signed images" feature only works by selecting teams in the `docker-datacenter` org.  This became an issue in UCP 2.2 which has multiple org support.

Fixing this is very unlikely in a patch release, so we should doc in the UI immediately.

@joaofnfernandes


joaofnfernandes:
Not sure what's going with Netlify, but I gave it another nudge to see if it builds this time.

PR #5184, opened 2017-11-01
fix command

galdiolo:
The command returns the following error:

> no such service: getstartedlab


joaofnfernandes:
Thanks for fixing this!

PR #5193, opened 2017-11-02
Fix service name

bsjcho:
### Proposed changes

Updated to correct service name when attempting to list tasks.
`docker service ps getstartedlab` returns `no such service: getstartedlab`
`docker service ps getstartedlab_web` has the expected behaviour of listing all tasks.

screenshot 2017-11-02 00 00 16


joaofnfernandes:
Thanks for fixing it. Let's ship this 🚀

PR #5210, opened 2017-11-03
Update running_ssh_service.md

jjvein:
the initialized status of `PermitRootLogin prohibit-password` is off as `#PermitRootLogin prohibit-password`

### DIFF
--- a/engine/examples/running_ssh_service.md
+++ b/engine/examples/running_ssh_service.md
@@ -16,7 +16,7 @@ FROM ubuntu:16.04
 RUN apt-get update && apt-get install -y openssh-server
 RUN mkdir /var/run/sshd
 RUN echo 'root:screencast' | chpasswd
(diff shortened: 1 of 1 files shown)


joaofnfernandes:
If I create an image from this Dockerfile

```
FROM ubuntu:16.04

RUN apt-get update && apt-get install -y openssh-server
```

And then run it, here's what I get:

```
docker run ubuntu:ssh cat /etc/ssh/sshd_config | grep -i permit

PermitRootLogin prohibit-password
PermitEmptyPasswords no
# the setting of "PermitRootLogin without-password".
```

This means that the entry on that file is `PermitRootLogin`, and not `#PermitRootLogin`, so your suggestion would not work 😢
Because of this, I'm closing this issue, but let me know if you disagree. I'm sure you'll have tons of other opportunities to help fix the docs 🐳

PR #5211, opened 2017-11-03
UCP 224 Installer has new --data-path-addr flag

ollypom:
### Proposed changes
The --data-path-addr flag was added to the UCP installer in UCP 2.2.4


joaofnfernandes:
Thanks for keeping the docs fresh 🥑. Let's ship it!

PR #5214, opened 2017-11-03
Update volumes.md

WhiteHalmos:
### Proposed changes

Fix bind-mounts link. It was pointing to "bind-mounts-md" and that doesn't exist.


joaofnfernandes:
Thanks for keeping the docs free of broken links 🙌

PR #5222, opened 2017-11-06
Add info about selinux limitations for IBM Z

mistyhacks:
cc/ @JimGalasyn @mideidocker @andrewhsu @friism


mistyhacks:
If you don't understand what is going on with the Liquid, I create the note and then use it three times in the topic, where I think the reader will need to see it in case they don't read from top to bottom (nobody does). Or you can wait for the Netlify preview to render and go to `Get Docker` > `Docker EE` -> `RHEL`

PR #5229, opened 2017-11-06
Updated include file for Compose to use code fences

### LINKED ISSUE #5226: Extension fields flaw
Hi there,

I just noticed a tiny flaw on the Docker compose documentation side in the Extension fields section:

https://docs.docker.com/compose/compose-file/compose-file-v2/#extension-fields

Markdown seems broken in for 3rd code block.
And the indention of 4th code block is wrong.

Kind regards

londoncalling:
### What's changed

- Reformatted Compose extensions examples to use code fences

### Related

- Fixes #5226 

### Reviewers

@qoomon Thanks for catching this. It was in an include file, so affects multiple versions of the Compose file reference. 

These changes should address the issue with code block formatting, but I'm not sure I see where the indent is wrong on the last code block. Can you let me know? 

### DIFF
--- a/_includes/content/compose-extfields-sub.md
+++ b/_includes/content/compose-extfields-sub.md
@@ -2,28 +2,33 @@ It is possible to re-use configuration fragments using extension fields. Those
 special fields can be of any format as long as they are located at the root of
 your Compose file and their name start with the `x-` character sequence.
 
-    version: '2.1'
-    x-custom:
-      items:
-        - a
-        - b
-      options:
-        max-size: '12m'
-      name: "custom"
+```none
+version: '2.1'
+x-custom:
+  items:
+    - a
+    - b
+  options:
+    max-size: '12m'
+  name: "custom"
+```
 
 The contents of those fields will be ignored by Compose, but they can be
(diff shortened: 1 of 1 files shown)


_includes/content/compose-extfields-sub.md
```diff
@@ -2,28 +2,33 @@ It is possible to re-use configuration fragments using extension fields. Those
...
 
-    version: '2.1'
-    x-custom:
-      items:
-        - a
-        - b
-      options:
-        max-size: '12m'
-      name: "custom"
+```none
```
shin-: not a big deal, but this could be \`\`\`yaml instead

londoncalling: Hey thanks @shin- I'll fix that.

mistyhacks: There are some instances when the yaml codefence breaks Atom syntax highlighting, but mostly when the YAML has pseudo-stuff in it (like `...` to show that output is truncated). It never breaks rendering of the actual output but can be annoying when editing the file. In those cases you will notice that syntax highlighting gets weird and you can selectively change those few back to `none`.

PR #5231, opened 2017-11-07
Make archive easier

mistyhacks:
Adds a script and layout for dealing with archives

Updated instructions in https://docker.atlassian.net/wiki/spaces/DOCSTEAM/pages/133081331/Docs+release+and+archiving+process to reflect using this script. @JimGalasyn @johndmulhausen can you please try this out, aside from the pushing part?

### DIFF
--- a/_layouts/archive-redirect.html
+++ b/_layouts/archive-redirect.html
@@ -1,0 +1,14 @@
+---
+layout: docs
+---
+<html>
+<head><title>{{ page.prod_title }}</title></head>
+<body>
+<h1>{{ page.prod_title }}</h1>
+<p>You are viewing an archive branch of the
+  <a href="https://docs.docker.com/">Docker documentation</a>, but
+  {{ prod_title }} documentation is not published in the current archive.
+  <a href="https://docs.docker.com/{{ page.prod_url }}/" target="_blank">Go to https://docs.docker.com/{{ page.prod_url }}/</a>
+  in a new tab.</p>
+</body>
+</html>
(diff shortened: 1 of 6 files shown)


Dockerfile
```diff
@@ -1,11 +1,16 @@
 FROM docs/docker.github.io:docs-base
 
+# IF THIS IS NOW AN ARCHIVE: remove everything in this file (including the line above) and replace it with the below (and uncomment):
+# (see https://docker.atlassian.net/wiki/spaces/DOCSTEAM/pages/133081331/Docs+release+and+archiving+process):
+#
+# FROM starefossen/github-pages:137
```
johndmulhausen: Soon to be 147 just FYI; https://github.com/docker/docker.github.io/pull/5241

There's not a hard dependency between that PR and this one but it might be prudent to just change it to 147

Dockerfile
```diff
@@ -1,11 +1,16 @@
 FROM docs/docker.github.io:docs-base
 
+# IF THIS IS NOW AN ARCHIVE: remove everything in this file (including the line above) and replace it with the below (and uncomment):
+# (see https://docker.atlassian.net/wiki/spaces/DOCSTEAM/pages/133081331/Docs+release+and+archiving+process):
```
axilleas: This is not accessible to non-members I presume? Is it possible maybe to use a link that's open to the public? :)

mistyhacks: Hi @axilleas! The archiving process for docs is a Docker-internal process, because we need access to some systems and tools that are not available externally, as well as access to Docker release engineering etc. Sorry about the internal-only link, though.

axilleas: Sure, no problem, if there's no other way!

Issue #5235, opened 2017-11-07
17.06.2-ee-4 note Serialize IP allocation docker/libnetwork#1788 links to wrong repo


adamancini:
`Serialize IP allocation docker/libnetwork#1788` links to https://github.com/docker/swarmkit/pull/1788  but I think it should link to https://github.com/docker/libnetwork/pull/1788

[enterprise/17.06/index.md](https://docs.docker.com/enterprise/17.06/), CC @johndmulhausen

thaJeztah:
Will be addressed by https://github.com/docker/docker.github.io/pull/5236

PR #5236, opened 2017-11-07
Correcting PR URL in 17.06.2-ee-4 notes

ericsmalling:
### Proposed changes
Linked URL of PR was incorrect - fixed it.

fixes https://github.com/docker/docker.github.io/issues/5235


joaofnfernandes:
Let's ship it then 🚀
@thaJeztah keep in mind that the source of truth for these release notes is docker/docker-ee (private repo). I don't think in this case we need to send a PR upstream, but are you pinged for reviews on that repo / does it make sense for you to be pinged in the future?

PR #5238, opened 2017-11-07
update telemetry docs

friism:
cc @rogaha


rogaha:
Thanks for updating it @friism!

PR #5241, opened 2017-11-08
Speed up site rendering for authors

johndmulhausen:
There are multiple vectors affecting site performance and here I'm trying to take on many of them:

1. tree.html is being called hundreds of times per build and each call is a recursive loop. This is replaced by JS logic that reads a jsonified version of the TOC at load time (note: not page ready time), injecting the TOC HTML using pure JS, no jQuery. This is faster-than-ajax processing as it does not involve waiting for jQuery to load and realize the the document is ready, nor does it involve making XmlHttpRequests asynchronously; liquid pre-prints a JSON object with the TOC in it and pure JS immediately rips it apart and injects it into the HTML as soon as it's read. The HTML is built up in an Array that is appended to using `push()` (as string concatenation does more memory/CPU work), and this Array is `join()`ed at injection time. It's as fast as can possibly be done with JS, and I think there is no difference between this and the current site. Here's the JSON TOC: https://deploy-preview-5241--docker-docs.netlify.com/js/toc.js -- poke around on Netlify and see the speed for yourself.

2. An outdated version of GH Pages is being used; this upgrades us from 137 > 147

3. We're hand-rolling the sitemap.xml file, when there's a perfectly good plugin for it included with GH Pages. 

4. Our use of plugins and other such bells and whistles is creating bottlenecks because plugins are too stupid to know which file got updated; now there is `_config_authoring.yml`, which takes out all plugins, skips applying several site configurations, skips most of the per-page ("scoped") variable setting, skips sitemap generation altogether, and skips any attempts to process Jekyll collections, all of which massively improves the incremental build performance when editing individual files.

Naturally, the Dockerfile/production build process will use the full `_config.yml` that has these all turned on.

To use: 
```
jekyll serve --config _config_authoring.yml
```

## Editing a single page

Before these changes, a full site rebuild was triggered, ~300 seconds, depending on my machine

Now: 9 seconds on my Windows machine. Haven't tested Mac yet.

## Full site rebuild

Before: ~300 seconds (using `_config.yml`)

Now: 148.168 seconds on my Windows machine (using `_config_authoring.yml`). Haven't tested Mac yet.

Edit: Have tested on Mac, build time is 51 seconds on my Macbook Pro!

Unfortunately incremental building seems not to work due to reasons beyond what this PR addresses but with a 50 second build time we are significantly reducing refresh time while maintaining status quo on incremental builds.

Will file a jekyll bug asking for help once this PR is merged. At the point where we're using the latest github pages with no plugins, it should be working.


johndmulhausen:
WIP: Still have to do procedurally-generated content (library docs, glossary, the weird ENV/shell-injected thing Misty added to the API refs, etc).

mistyhacks:
Is this going to be the normal way to build too? Or just for local authoring builds?

johndmulhausen:
Just for local authoring builds; regular/production builds and Netlify builds will still run plain `jekyll serve` which has "all the trimmings."

mistyhacks:
I just tried this and get no left-hand nav at all. Is that by design? It is certainly fast but ...

mistyhacks:
```Uncaught ReferenceError: renderNav is not defined
    at toc.js:2```

mistyhacks:
I think I see the problem. Testing and will push a fix here if I am right.

mistyhacks:
I just pushed a commit that fixes left-hav rendering in all but the glossary. I introduced a bug there and I'm not 100% sure how to fix it. PTAL @johndmulhausen

johndmulhausen:
@mistyhacks 
Not that I saw the issue you were seeing beforehand, but I also can't repro the glossary bug you're seeing and neither can Netlify? https://deploy-preview-5241--docker-docs.netlify.com/glossary/

Are you sure you aren't encountering caching issues?

mistyhacks:
@johndmulhausen I saw the issue when I built locally (using `bundle exec jekyll serve --incremental --config _config_authoring.yml`  and then viewed the docs and looked at the console. Before I put all of the JS logic into the external JS file, I was getting errors because the JS file relied on logic that was in docs.js which was not yet loaded. Putting all the logic into the external JS file fixed most of it.

johndmulhausen:
Unfortunately incremental building seems not to work due to reasons beyond what this PR addresses but with a 50 second build time we are significantly reducing refresh time while maintaining status quo on incremental builds.

Will file a jekyll bug asking for help once this PR is merged. At the point where we're using the latest github pages with no plugins, it should be working.

johndmulhausen:
Let's do this!

PR #5244, opened 2017-11-08
Updated volume create to use name option.

Jatrott:

### DIFF
--- a/engine/admin/volumes/volumes.md
+++ b/engine/admin/volumes/volumes.md
@@ -97,7 +97,7 @@ container.
 **Create a volume**:
 
 ```bash
-$ docker volume create my-vol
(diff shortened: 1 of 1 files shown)


engine/admin/volumes/volumes.md
```diff
@@ -97,7 +97,7 @@ container.
 **Create a volume**:
 
 ```bash
-$ docker volume create my-vol
+$ docker volume create --name my-vol
```
thaJeztah: hm, not this isn't correct.

are you perhaps running an old version of docker? the `--name` flag is now hidden, and only kept for backward compatibility; the new format is to use the name as is documented here

mistyhacks:
Closed as out of date. Please let us know if you have questions.

PR #5248, opened 2017-11-08
remove note that's no longer accurate

friism:
cc @mistyhacks


vieux:
should we move the notes 15 lines down and say `overlay` is only available in CE ? only `overlay2` is available in EE

PR #5250, opened 2017-11-08
add placement pref note about null labels

BretFisher:
### Proposed changes

Nearly a cut and paste from SwarmKit topology.md to prevent confusion about missing labels on nodes when setting placement preferences for spread.

### Related issues (optional)

https://github.com/moby/moby/issues/35433#issuecomment-342862546


BretFisher:
Thanks @mistyhacks !

PR #5252, opened 2017-11-09
Add detail on how to allow hosts to communicate via macvlan.

PaulSchulz:
### Proposed changes

These details are eluded to in lots of places when I searched for Docker Networking and Macvlan interfaces, but I couldn't find anything specific. 

With this change I am able to communicate between the host and the docker container.


mistyhacks:
Thanks! I did a copyedit to correct some Markdown problems and prevent note stacking, and also lightly copyedited surrounding text (typos you didn't introduce, etc).

PR #5259, opened 2017-11-09
Update work-with-networks.md

JimCodified:
Corresponds to issue #34873: clarifies that the IP address used for binding should be an actual IP on the user's host, and not literally 172.23.0.1

### Proposed changes

Edited the wording to clarify that the IP address used in `network create` example must be an IP address available on the user's host, and not literally 172.23.0.1. 

### Related issues (optional)

Fixes https://github.com/moby/moby/issues/34873


mistyhacks:
See my edit to the description and also my edit to the commit message -- please don't mention issues or PRs by number in the *commit message* as this will ping the issue (or in this case, a hypothetical issue in THIS repo, since you didn't provide the full URL to the Moby issue) and possibly close issues in other repos, erroneously, each time the commit makes it into a new branch.

PR #5266, opened 2017-11-09
Add new debian & ubuntu edge support

gbarr01:
Add Debian Buster and Ubuntu Artful to list of supported OSes in edge only

### DIFF
--- a/engine/installation/linux/docker-ce/debian.md
+++ b/engine/installation/linux/docker-ce/debian.md
@@ -26,9 +26,9 @@ and distributions for different Docker editions, see
 To install Docker CE, you need the 64-bit version of one of these Debian or
 Raspbian versions:
 
-- Buster (edge only)
-- Stretch (stable) / Raspbian Stretch
-- Jessie 8.0 (LTS) / Raspbian Jessie
+- Buster 10 (edge only)
+- Stretch 9 (stable) / Raspbian Stretch
+- Jessie 8 (LTS) / Raspbian Jessie
 - Wheezy 7.7 (LTS)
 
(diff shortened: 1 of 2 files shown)


engine/installation/linux/docker-ce/ubuntu.md
```diff
@@ -29,6 +29,7 @@ To learn more about Docker EE, see
 To install Docker CE, you need the 64-bit version of one of these Ubuntu
 versions:
 
+- Artful 17.10 (edge only)
```
thaJeztah: It will be in stable as well once the next stable is there (so at that point we can remove that part)

engine/installation/linux/docker-ce/debian.md
```diff
@@ -26,6 +26,7 @@ and distributions for different Docker editions, see
 To install Docker CE, you need the 64-bit version of one of these Debian or
 Raspbian versions:
 
+- Buster (edge only)
```
thaJeztah: Wondering if we should add versions to all of these;

- Buster 10 (or `Debian 10 "Buster"`? Not sure what the preferred notation is)
- Stretch 9
- Jessie 8 (LTS)
- Wheezy 7.7 (LTS)

Frankly; we should start thinking of removing Wheezy (as it's getting old in the tooth), and it won't even support docker without installing a custom kernel

gbarr01: I agree .. but there has been debate about that. I'll just do it. More clarity is always better than less.

thaJeztah: Ignore the "deprecate wheezy" part, that's just me rambling 😅 - plans were already in place for that, but will be early next year probably

PR #5269, opened 2017-11-10
fix "List all images" active first

MwumLi:


joaofnfernandes:
Thanks for spotting this!

PR #5272, opened 2017-11-10
Use valid CPU limits

### LINKED ISSUE #5271: CPU limit is set too low
### Problem description

Docker compose documentation uses 0.01% as a reservation value 

`0.0001 CPU time reserve`

```
reservations:
    cpus: '0.0001'
```
But using it produces the following error:

```
failed to create service servicename: Error response from daemon: rpc error: code = InvalidArgument desc = invalid cpu value 0.0001: Must be at least 0.001
```

### Problem location

Files is: `compose/compose-file/index.md`
Under the section **resources**.

[Direct link](https://github.com/docker/docker.github.io/blob/master/compose/compose-file/index.md#resources)

docker-compose version 1.17.0
Docker version 17.10.0-ce, build f4ffd2511c

### Suggestions for a fix

Use 0.001 as a reservation value and increase the limit value accordingly or drop the reservations section.

AlexandreRio:
### Proposed changes

Use 50% and 25% CPU limits instead of invalid 0.001

### Related issues (optional)

Fixes #5271

### DIFF
--- a/compose/compose-file/index.md
+++ b/compose/compose-file/index.md
@@ -659,8 +659,8 @@ Each of these is a single value, analogous to its [docker service
 create](/engine/reference/commandline/service_create.md) counterpart.
 
 In this general example, the `redis` service is constrained to use no more than
-50M of memory and `0.001` (0.1%) of available processing time (CPU), and has
-`20M` of memory and `0.0001` CPU time reserved (as always available to it).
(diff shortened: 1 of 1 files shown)


londoncalling:
LGTM, thanks @AlexandreRio , @thaJeztah , merging

PR #5273, opened 2017-11-10
Fix "virtual" explanation for "docker ps -s"

### LINKED ISSUE #5270: Explain what "virtual" keyword means in "docker ps -s"
### Problem description
 [Container size on disk](https://docs.docker.com/engine/userguide/storagedriver/imagesandcontainers/#container-size-on-disk)
`virtual size: the amount of data used for the read-only image data used by the container. `

for my understanding, the virtual size will not change if run  `docker ps -s ` , but when I created new file inside container, the virtual size was changed accordingly.

Step to reproduce
```
Dockers-MacBook-Pro-3:~ docker$ docker run -d nginx
Dockers-MacBook-Pro-3:~ docker$ docker ps -s
CONTAINER ID        IMAGE                                COMMAND                  CREATED             STATUS              PORTS                    NAMES                   SIZE
af0288fead8b        nginx                                "nginx -g 'daemon ..."   8 seconds ago       Up 7 seconds        80/tcp                   inspiring_montalcini    2B (virtual 108MB)
Dockers-MacBook-Pro-3:~ docker$ docker exec -it af0288fead8b /bin/sh
# dd if=/dev/zero of=/test.log bs=1M count=100
100+0 records in
100+0 records out
104857600 bytes (105 MB, 100 MiB) copied, 0.122115 s, 859 MB/s
Dockers-MacBook-Pro-3:~ docker$ docker ps -s
CONTAINER ID        IMAGE                                COMMAND                  CREATED             STATUS              PORTS                    NAMES                   SIZE
af0288fead8b        nginx                                "nginx -g 'daemon ..."   4 minutes ago       Up 4 minutes        80/tcp                   inspiring_montalcini    105MB (virtual 213MB)
```

this also happened in `docker container ls -s`

I have already check [#1520](https://github.com/docker/docker.github.io/issues/1520)

xinfengliu:
### Proposed changes

Copy from Slack #docs conversation:

Xinfeng Liu [16:35] 
Hi, 
from https://github.com/moby/moby/blob/master/daemon/getsize_unix.go#L39   
and https://docs.docker.com/engine/userguide/storagedriver/imagesandcontainers/#container-size-on-disk  `virtual size: the amount of data used for the read-only image data used by the container. `  The source codes and docs are not consistent,  I want to confirm which is correct?
Also please see new issue for this: https://github.com/docker/docker.github.io/issues/5270 (edited)

Sebastiaan van Stijn [17:19] 
@xinfeng was just reading it this morning (didn't reply on that issue yet); yes, I think an error slipped into the docs: `size:` is the size used by the writable layer of the container, and `virtual size` is "writable layer + read-only layers (AKA "the image")". To get the size of "the image" on disk, you'd have to do "`virtual size - size`". To calculate the total size on disk for `N` containers running `image:foo`; the size would be: `SUM(size of containers)` + `1 x "virtual size - size"`

### Related issues (optional)
Fixes #5270

### DIFF
--- a/engine/userguide/storagedriver/imagesandcontainers.md
+++ b/engine/userguide/storagedriver/imagesandcontainers.md
@@ -81,7 +81,8 @@ command. Two different columns relate to size.
   each container
 
 - `virtual size`: the amount of data used for the read-only image data
-  used by the container. Multiple containers may share some or all read-only
+  used by the container plus the container's writable layer `size`. 
+  Multiple containers may share some or all read-only
   image data. Two containers started from the same image share 100% of the
   read-only data, while two containers with different images which have layers
   in common share those common layers. Therefore, you can't just total the
@@ -90,8 +91,9 @@ command. Two different columns relate to size.
 
 The total disk space used by all of the running containers on disk is some
 combination of each container's `size` and the `virtual size` values. If
-multiple containers have exactly the same `virtual size`, they are likely
-started from the same exact image.
+multiple containers started from the same exact image, the total size on disk for 
+these containers would be SUM (`size` of containers) plus one container's 
+(`virtual size`- `size`).
 
 This also does not count the following additional ways a container can take up
 disk space:


thaJeztah:
Thanks!

Also added some examples in https://github.com/docker/docker.github.io/issues/5270#issuecomment-343428138 (in case we want a full example to show - but may not be needed)

PR #5274, opened 2017-11-10
Correct sync_schedule in ucp-description-file.md

axeal:
The sync_schedule parameter in the auth.ldap section of the UCP Configuration file contains a seconds field, whilst the existing documentation claims this is omitted. Should explicitly call out inclusion of seconds field here, as this is usually omitted in a standard CRON entry.


joaofnfernandes:
According to this you need to always include the seconds field, but it needs to be set to zero.


```
docker run -it --rm docker/ucp:2.2.4 example-config

...
    # The scheduled time for automatic LDAP sync jobs. Can be specified in CRON
    # table entry format (with a seconds field always set to 0):
    #
    #   0 (minutes) (hours) (day of month) (month of year) (day of week, optional)
    #
    # Can also be specified as a predefined scheduling definition: '@hourly',
    # '@daily', or '@weekly'.
    # Default is '@hourly' if empty or omitted.
    sync_schedule = "@daily"
```

Thanks for your contribution @axeal. I'll push a change to your PR so make it clear that the field always needs to be set to zero.

PR #5275, opened 2017-11-10
Do not split the content inside backticks

axilleas:
Adding newlines inside backticks breaks Markdown. Page affected: https://docs.docker.com/machine/examples/aws/#step-3-run-docker-commands-on-the-new-instance.

| **Before** | **After** |
| :----: | :-----: |
| screenshot-docs docker com-2017-11-10-15-16-35  | screenshot-localhost 4000-2017-11-10-15-16-06 |

### DIFF
--- a/machine/examples/aws.md
+++ b/machine/examples/aws.md
@@ -182,7 +182,7 @@ other machine):
     eval $(docker-machine env -u)
     docker container ls -a
     ```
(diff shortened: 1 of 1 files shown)


joaofnfernandes:
Looks good! Let's ship it 🚀

PR #5276, opened 2017-11-10
clarify backup relationship between ucp and swarm

trapier:
Current formulation over conflates the concepts of "UCP" and "swarm".
Especially in the context of backup and restore, it is important to note their
differences.


trapier:
Would have just filed an issue, but I think it's more productive to make arguments by example.  Please reword, carry, or otherwise massage as needed.

trapier:
cc @JimGalasyn.

PR #5286, opened 2017-11-12
correct a typo

maxstreichert:
this command is false, since the the correct name of the service is getstartedlab_web


maxstreichert:
sorry, my mistake. I mixed up with the docker service command. For docker stack it's correct.

PR #5291, opened 2017-11-13
17.11 docs release

gbarr01:


_data/toc.yaml
```diff
@@ -1275,10 +1277,22 @@ reference:
     section:
     - path: /edge/engine/reference/commandline/trust/
       title: docker trust
+    - path: /edge/engine/reference/commandline/trust_key/
+      title: docker trust key
+    - path: /edge/engine/reference/commandline/trust_key_generate/
```
mistyhacks: These need to be "under" `docker trust key`. Same issues with `trust signer` sub-subcommands. @riyazdf this is not a pattern I've seen before. Is it OK for us to have this extra level of sub-command depth? cc/ @thaJeztah

riyazdf: @mistyhacks: I'm fine with another directory level if it's preferred

PR #5292, opened 2017-11-13
DTR patch releases

joaofnfernandes:
/cc @vikstrous, @pdevine if you want to review this.


vikstrous:
We need to update the release dates and add one more fix to 2.4.1. See the private tracker for details.

PR #5298, opened 2017-11-13
Improve script to reset user password

### LINKED ISSUE #5237: ucp password reset documentation incorrect if ucp installed from existing config with debug enabled
### Problem description
```
[vagrant@node01 ~]$ docker exec -it ucp-auth-api enzi \
>   "$(docker inspect --format '{{ index .Args 0 }}' ucp-auth-api)" \
>   passwd -i
Admin Username: admin
Admin Password:
Confirm Admin Password:
DEBU[0010] connecting to db ...
DEBU[0010] connecting to DB Addrs: [db.enzi]
FATA[0010] unable to connect to database cluster: gorethink: dial tcp: lookup db.enzi on 192.168.121.1:53: no such host
```

When debug is enabled, index 0 is incorrect here:

```
[vagrant@node01 ~]$ docker inspect --format '{{ index .Args 0 }}' ucp-auth-api
--debug
```

enzi is expecting:

```
[vagrant@node01 ~]$ docker inspect --format '{{ index .Args 1 }}' ucp-auth-api
--db-addr=192.168.121.206:12383
```

### Problem location

https://docs.docker.com/datacenter/ucp/2.2/guides/access-control/recover-a-user-password/

UCP 2.2.4

### Suggestions for a fix
Change the copy-pastable instructions to hunt for `--db-addr` and not rely on the position of this argument, or use some other means to determine the IP address.

joaofnfernandes:
Updates the script to reset a UCP admin password so that it works independently of the environment setup. I could not get a solution with Grep that was portable between OSX and Ubuntu, that's why I ended up with this small perl script.
/cc @trapier for a review.

Fixes #5237

### DIFF
--- a/datacenter/ucp/2.2/guides/access-control/recover-a-user-password.md
+++ b/datacenter/ucp/2.2/guides/access-control/recover-a-user-password.md
@@ -26,7 +26,7 @@ node managed by UCP, and run:
 ```none
 {% raw %}
 docker exec -it ucp-auth-api enzi \
-  "$(docker inspect --format '{{ .Args}}' ucp-auth-api | perl -ne 'print "$1\n" if /(--db-addr=(\d+[.:]?)+)/')" \
+  $(docker inspect --format '{{range .Args}}{{if eq "--db-addr=" (printf "%.10s" .)}}{{.}}{{end}}{{end}}' ucp-auth-api) \
   passwd -i
 {% endraw %}  
 ```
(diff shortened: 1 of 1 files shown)


trapier:
Anecdote: `perl` not installed by default on centos vagrant libvirt box.

```
$ docker inspect --format '{{ .Args}}' ucp-auth-api | perl -ne 'print "$1\n" if /(--db-addr=(\d+[.:]?)+)/'
-bash: perl: command not found
```

works everywhere(?):
```
docker inspect --format '{{range .Args}}{{if eq "--db-addr=" (printf "%.10s" .)}}{{.}}{{end}}{{end}}' ucp-auth-api
--db-addr=192.168.121.206:12383
```

joaofnfernandes:
That's way better since it only relies on Docker. I tried to do something similar before going for perl, but I couldn't find how to do substrings with Go templates. Now I know`printf("%.10s", s)` 😄

PR #5310, opened 2017-11-14
[WIP] remove windows manual instructions

friism:
This is already covered here: https://docs.docker.com/engine/installation/windows/docker-ee/#use-a-script-to-install-docker-ee


veyalla:
@friism Please don't delete these instructions. These for the latest Docker CE bits (17.09). We are referencing this in our documentation that is going live tomorrow. If this page is gone, there will no instructions for our users to install our product. This is needed for Windows IoT Core where Docker for Windows GUI installer will not work...

PR #5312, opened 2017-11-15
adds hostname to bash prompts

kphunter:
### Proposed changes

Adding hostname prompts makes it easier to see that you need to login to different machines to inspect the task more closely.


engine/swarm/swarm-tutorial/inspect-service.md
```diff
@@ -18,7 +18,7 @@ the Docker CLI to see details about the service running in the swarm.
     To see the details on the `helloworld` service:
 
     ```bash
-    $ docker service inspect --pretty helloworld
+    manager1$ docker service inspect --pretty helloworld
```
mistyhacks: I like the intent but please use a syntax like `[manager1] $`.

PR #5321, opened 2017-11-16
Add a note regarding retrieving log messages when tty is enabled

ripcurld0:
### Proposed changes

When running a container with tty enabled and setting the journald
logging driver, you may see "blob data" instead of the real log
message.

That's because tty appends "\r" to the end of the log line,
which isn't stripped by jorunald unless you specify the "--all" flag.

### Related issues (optional)
related to https://github.com/moby/moby/pull/35416 and https://github.com/moby/moby/pull/22450

### DIFF
--- a/engine/admin/logging/journald.md
+++ b/engine/admin/logging/journald.md
@@ -100,6 +100,17 @@ to return the log messages in JSON format.
 $ sudo journalctl -o json CONTAINER_NAME=webserver
 ```
 
+### View logs for a container with a TTY enabled
+
+If TTY is enabled on a container you may see `[10B blob data]` in the output
+when retrieving log messages.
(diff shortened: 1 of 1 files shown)


engine/admin/logging/journald.md
```diff
@@ -100,6 +100,17 @@ to return the log messages in JSON format.
 $ sudo journalctl -o json CONTAINER_NAME=webserver
 ```
 
+### Note regarding Tty enabled containers
+
+When Tty is enabled on a container you may see `[10B blob data]` in the output
```
thaJeztah: Perhaps: "If a TTY is enabled"

ripcurld0:
Thank you! 🙌

PR #5336, opened 2017-11-18
Use COPY statement instead of ADD

jottr:


mistyhacks:
OK, you can use `ADD` or `COPY`. Can you explain why `COPY` is better here?

thaJeztah:
Generally `COPY` is the recommended instruction if you don't need the additional magic of `ADD` (automatic extraction, or pulling from remote servers)

PR #5341, opened 2017-11-20
Repair broken link

### LINKED ISSUE #5337: 404 at /engine/swarm/networking/@configure-service-discovery
What is wrong with the below page?

I get to the missing page through [this page](https://docs.docker.com/engine/swarm/ingress/#without-the-routing-mesh).

screenclip

screenclip

URL: /engine/swarm/networking/@configure-service-discovery

gbarr01:
The the link to config service discovery was using @ and not #. 

Fixes #5337

### DIFF
--- a/engine/swarm/ingress.md
+++ b/engine/swarm/ingress.md
@@ -244,7 +244,7 @@ single virtual IP. Instead, Docker sets up DNS entries for the service such that
 a DNS query for the service name returns a list of IP addresses, and the client
 connects directly to one of these. You are responsible for providing the list of
 IP addresses and ports to your load balancer. See
-[Configure service discovery](networking.md@configure-service-discovery).
+[Configure service discovery](networking/#configure-service-discovery).
 
 ## Learn more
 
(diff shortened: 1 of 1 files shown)


engine/swarm/ingress.md
```diff
@@ -244,7 +244,7 @@ single virtual IP. Instead, Docker sets up DNS entries for the service such that
 a DNS query for the service name returns a list of IP addresses, and the client
 connects directly to one of these. You are responsible for providing the list of
 IP addresses and ports to your load balancer. See
-[Configure service discovery](networking.md@configure-service-discovery).
```
mistyhacks: Leave it as is but change the `@` to `#` please. :) John has some magic script that does the right thing with the `.md` suffixes.

gbarr01: @mistyhacks ... are you saying, remove the `/`? Because that's the only difference...

mistyhacks: The problem is that the `@` is supposed to be a `#`...  so change

```
-[Configure service discovery](networking.md@configure-service-discovery).
```

to

```
-[Configure service discovery](networking.md#configure-service-discovery).
```

PR #5351, opened 2017-11-21
Improve archival dockerfile 1.12

mistyhacks:
This is in preparation for a larger overhaul of the `docs-base` Dockerfile. This will make the final archival image only have the built HTML and the Nginx config.

To test, check out this branch and build the Dockerfile locally, then run the image at port 4000.

If you like this I will propagate it to the other archive branches.

**This will change the contract of the Docker archive image, such that the docs will now live in `/etc/nginx/html`**. This is important for `docs-base` which will need to change as it goes multi-stage.


Dockerfile
```diff
@@ -1,3 +1,10 @@
...
+FROM starefossen/github-pages:147 AS build_docs
+COPY . /source
+RUN jekyll build --source /source --destination /site
 
-CMD bundle update && bundle exec jekyll serve -d /_site --watch -H 0.0.0.0 -P 4000
+# Reset with nginx
+FROM nginx:alpine
+COPY nginx.conf /etc/nginx/nginx.conf
+COPY --from=build_docs /site /usr/src/app/target
+CMD echo "Docker docs are viewable at:" && echo "http://0.0.0.0:4000" && nginx -g 'pid /tmp/nginx.pid;'
```
thaJeztah: Probably needs `exec "nginx -g 'pid /tmp/nginx.pid;'"` at the end, so that `nginx` becomes the main process.

Do you know where the `/tmp/nginx.pid` came from? The default `CMD` of the `nginx:alpine` image is:

```Dockerfile
CMD ["nginx", "-g", "daemon off;"]
```

So, this should do the same:

```Dockerfile
CMD echo -e "Docker docs are viewable at:\nhttp://0.0.0.0:4000"; exec nginx -g 'daemon off;'
```

Dockerfile
```diff
@@ -1,3 +1,10 @@
...
+# Build the docs
+FROM starefossen/github-pages:147 AS build_docs
+COPY . /source
+RUN jekyll build --source /source --destination /site
 
-CMD bundle update && bundle exec jekyll serve -d /_site --watch -H 0.0.0.0 -P 4000
+# Reset with nginx
+FROM nginx:alpine
+COPY nginx.conf /etc/nginx/nginx.conf
+COPY --from=build_docs /site /usr/src/app/target
```
thaJeztah: If we copy to `/usr/share/nginx/html`, there's even less customisation needed in the default configuration 👍

nginx.conf
```diff
@@ -0,0 +1,18 @@
...
+  server {
+    root /usr/src/app/target;
+    port_in_redirect off;
+    location / {
+      include /etc/nginx/mime.types;
+    }
+    listen 4000;
+    error_page 403 404 /404.html;
+  }
+}
```
thaJeztah: Comparing with the default nginx.conf; https://github.com/nginxinc/docker-nginx/blob/3ba04e37d8f9ed7709fd30bf4dc6c36554e578ac/mainline/alpine/nginx.conf#L2-L32 customisations are:

```
port_in_redirect off;
listen 4000;
error_page 403 404 /404.html;
```

We could try putting those in a file, and adding to the image at; `/etc/nginx/conf.d/000-default.conf` (and the main configuration file will include that file).

mistyhacks:
Word of warning, this will be a breaking change when it hits and likely `docs-base` will be weird until we get this propagated across all branches and get `docs-base` updated....

_config.yml
```diff
@@ -11,7 +11,7 @@ safe: false
 lsi: false
 url: https://docs.docker.com/v1.12
 keep_files: ["v1.4", "v1.5", "v1.6", "v1.7", "v1.8", "v1.9", "v1.10", "v1.11"]
-
+exclude: ["Dockerfile", "Gemfile", "Gemfile.lock", "hooks", "nginx-overrides.conf", "_scripts"]
 gems:
```
thaJeztah: This something we need to change?

```
Configuration file: /source/_config.yml
       Deprecation: The 'gems' configuration option has been renamed to 'plugins'. Please update your config file accordingly.
            Source: /source
       Destination: /site
 Incremental build: enabled
```

mistyhacks: Maybe eventually, but right now, changing it breaks Netlify. It's only a deprecation for now. Seems to be GH Pages being more opinionated than the generic Jekyll config.

Dockerfile
```diff
@@ -1,3 +1,11 @@
...
+# Build the docs
+FROM starefossen/github-pages:147 AS build_docs
+COPY . /source
+RUN jekyll build --source /source --destination /site
 
-CMD bundle update && bundle exec jekyll serve -d /_site --watch -H 0.0.0.0 -P 4000
+# Reset with nginx, so we don't get docs source in the image
+FROM nginx:alpine
+# Override some nginx conf -- this gets added to default nginx conf
+COPY nginx-overrides.conf /etc/nginx/conf.d/000-default.conf
```
thaJeztah: Okay; found some issues; I forgot there's a default override file already in `/etc/nginx/conf.d/default.conf`. That one defines the "home" directory, but only for "localhost".

So, we should overwrite that file, and change the directory to copy the HTML to to `/usr/share/nginx/html`.

Here's a diff with the changes;

```patch
diff --git a/Dockerfile b/Dockerfile
index 8432103650..b1f497dcb6 100644
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,6 +6,7 @@ RUN jekyll build --source /source --destination /site
 # Reset with nginx, so we don't get docs source in the image
 FROM nginx:alpine
 # Override some nginx conf -- this gets added to default nginx conf
-COPY nginx-overrides.conf /etc/nginx/conf.d/000-default.conf
-COPY --from=build_docs /site /etc/nginx/html/
-CMD echo -e "Docker docs are viewable at:\nhttp://0.0.0.0:4000"; exec nginx -g 'daemon off;'
\ No newline at end of file
+COPY nginx-overrides.conf /etc/nginx/conf.d/default.conf
+RUN rm /usr/share/nginx/html/*
+COPY --from=build_docs /site /usr/share/nginx/html
+CMD echo -e "Docker docs are viewable at:\nhttp://0.0.0.0:4000"; exec nginx -g 'daemon off;'
diff --git a/nginx-overrides.conf b/nginx-overrides.conf
index 55ab5d2e0a..e05e1edd8b 100644
--- a/nginx-overrides.conf
+++ b/nginx-overrides.conf
@@ -2,4 +2,9 @@ server {
   port_in_redirect off;
   listen 4000;
   error_page 403 404 /404.html;
-}
\ No newline at end of file
+
+  location / {
+    root   /usr/share/nginx/html;
+    index  index.html index.htm;
+  }
+}
```

Dockerfile
```diff
@@ -1,3 +1,11 @@
-FROM starefossen/github-pages:onbuild
+# Build the docs
+FROM starefossen/github-pages:147 AS build_docs
```
thaJeztah: One more thought: what if we created a `docs-builder` image, that is basically just:

```Dockerfile
FROM starefossen/github-pages:147
```

And change these Dockerfiles to:

```Dockerfile
FROM docs-builder:latest
```

Then, whenever we want to bump the version of Jekyll, we can just push that image, and all archive versions will use the new Jekyll version when rebuilding.

Using `:latest` may not be "best practice", but in this case would save a bit of maintenance :smile:

thaJeztah: We could stash the `nginx-override.conf` file in there as well, so that we can manage that one from a central location, and `COPY --from` that to the nginx image

thaJeztah: Or even: if we have other configuration files that we want to share between "master" and the archives (idk, perhaps there's some Jekyll configuration that needs to be kept in sync), we could have a `docs/docker.github.io:config` image that contains those files, then we can

```Dockerfile
FROM docs/docker.github.io:config AS configuration

...


COPY --from=docs-configuration /config/some-file /destination/
```

Dockerfile
mistyhacks: This also requires a change in #5367 and #5369 due to the paths changing.

Gemfile
```diff
@@ -1,3 +1,3 @@
 source "https://rubygems.org"
 
-gem "github-pages", "112" #Update me once in a while: https://github.com/github/pages-gem/releases
+gem "github-pages", "147" #Update me once in a while: https://github.com/github/pages-gem/releases
```
thaJeztah: Should this one be in `docs/docker.github.io:docs-builder` as well? Looks like it's tied to the version of github-pages.

Actually; looking at the Dockerfile for [starefossen/github-pages](https://hub.docker.com/r/starefossen/github-pages/), We may not even need it at all: https://github.com/Starefossen/docker-github-pages/blob/147/Dockerfile#L6-L16

mistyhacks: You need the gem for when it's running in a random environment, like when someone just does `bundle exec jekyll serve` from the checked-out repo, or for Netlify. In fact, the Netlify builds on the older branches were failing because there was no Gemfile (now they are failing for other reasons I can't figure out). The Gemfile that lives in the source is different from the Gemfile imported from the github-pages Docker image (and probably should be ignored by Docker for that reason, but that is a call for a different chunk of work).

thaJeztah: Ahhhhh, right, yes that makes sense 👍

Dockerfile
```diff
@@ -1,3 +1,16 @@
...
-CMD bundle update && bundle exec jekyll serve -d /_site --watch -H 0.0.0.0 -P 4000
+# Build the docs from this branch
+COPY . /source
+RUN jekyll build --source /source --destination /site
+
+# Reset to bare Nginx image
+FROM nginx:alpine
+
+# Copy the Nginx config
+COPY --from=builder /conf/nginx-overrides.conf /etc/nginx/conf.d/default.conf
```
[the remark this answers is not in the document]
mistyhacks: I took your advice and separated out the config.

Dockerfile
```diff
@@ -1,3 +1,18 @@
...
-CMD bundle update && bundle exec jekyll serve -d /_site --watch -H 0.0.0.0 -P 4000
+# Build the docs from this branch
+COPY . /source
+RUN jekyll build --source /source --destination /site
+
+# Reset to bare Nginx image
+FROM nginx:alpine
+
+# Copy the Nginx config from the docs-config image
+COPY --from=docs/docker.github.io:docs-config /conf/nginx-overrides.conf /etc/nginx/conf.d/default.conf
```
mistyhacks: @thaJeztah Is this the right way not to bust the cache due to a config change?

thaJeztah: Yes, because the nginx-configuration should not have an effect on the content of the HTML

thaJeztah:
Hm, there's something wrong in this branch (not new); the menu isn't clickable for me; same for the current online version; https://docs.docker.com/v1.12/ 

Could be this?

screen shot 2017-12-22 at 16 20 53

thaJeztah:
Yes, that's it; fixing the URL for the metadata.txt unblocks the menu;

```patch
diff --git a/js/menu.js b/js/menu.js
index 9f83c1e3fc..e3b4ef9c74 100644
--- a/js/menu.js
+++ b/js/menu.js
@@ -195,7 +195,7 @@ function hookupTOCEvents()
 }
 
 jQuery(document).ready(function(){
-    $.getJSON( "/metadata.txt", function( data ) {
+    $.getJSON( "/v1.12/metadata.txt", function( data ) {
       metadata = data;
       hookupTOCEvents();
     });
```

We should look into that; may even consider removing search from the archives, or at least make sure that reading the metadata does not break the menu

mistyhacks:
So, we actually do disable the search in the archives now, by just hiding it. I'll look into the recent archives for how to disable this part of it too.

PR #5352, opened 2017-11-21
[WIP] Integrate UCP with multiple registries

joaofnfernandes:
/cc @jlhawn, @bkauf for a review

### DIFF
--- a/datacenter/ucp/2.2/guides/admin/configure/integrate-with-multiple-registries.md
+++ b/datacenter/ucp/2.2/guides/admin/configure/integrate-with-multiple-registries.md
@@ -1,0 +1,60 @@
+---
+title: Integrate with multiple registries
+description: Integrate UCP with multiple registries
+keywords: trust, registry, integrate, UCP, DTR
+---
+
+Universal Control Plane can pull and run images from any image registry,
+including Docker Trusted Registry and Docker Store.
+
+If your registry uses globally-trusted TLS certificates, everything works
+out of the box, and you don't need to configure anything.
+If your registries use self-signed certificates or use your own Certificate
+Authority to sign the public key certificates, you need to configure UCP
+to trust those repositories.
+
+## Trust Docker Trusted Registry
+
+To configure UCP to trust a DTR deployment, you need to update the
+[UCP system configuration]() to include one entry for each DTR deployment:
+
+```
+[[registries]]
+  host_address = "dtr.example.org"
+  ca_bundle = """
+-----BEGIN CERTIFICATE-----
+...
+-----END CERTIFICATE-----"""
+
+[[registries]]
+  host_address = "internal-dtr.example.org:444"
+  ca_bundle = """
+-----BEGIN CERTIFICATE-----
+...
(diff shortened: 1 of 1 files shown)


datacenter/ucp/2.2/guides/admin/configure/integrate-with-multiple-registries.md
```diff
@@ -0,0 +1,60 @@
...
+---
+
+Universal Control Plane can pull and run images from any image registry,
+including Docker Trusted Registry and Docker Store.
+
+If your registry uses globally-trusted TLS certificates, everything works
+out of the box, and you don't need to configure anything.
+If your registries use self-signed certificates or use your own Certificate
+Authority to sign the public key certificates, you need to configure UCP
+to trust those repositories.
```
jlhawn: > ... to trust those *registries*.

datacenter/ucp/2.2/guides/admin/configure/integrate-with-multiple-registries.md
```diff
@@ -0,0 +1,60 @@
...
+
+[[registries]]
+  host_address = "internal-dtr.example.org:444"
+  ca_bundle = """
+-----BEGIN CERTIFICATE-----
+...
+-----END CERTIFICATE-----"""
+```
+
+You only need to include the port section, if your DTR deployment is running
```
jlhawn: looks like the comma is a typo.

datacenter/ucp/2.2/guides/admin/configure/integrate-with-multiple-registries.md
```diff
@@ -0,0 +1,60 @@
...
+DTR_CA_URL=$DTR_URL/ca
+
+# Create the registry configuration and save it it
+cat <<EOL > trust-dtr.toml
+
+[[registries]]
+  # host address should not contain protocol or port if using 443
+  host_address = "$(echo $DTR_URL | sed 's/^https\{0,1\}:\/\///' | sed 's/:443$//')"
+  ca_bundle = """
+$(curl -sk $DTR_CA_URL)"""
```
jlhawn: This script could be simplified by just having a `DTR_ADDR` variable which is just the hostname [and optional port]. Then just put `https://${DTR_ADDR}/ca` in the `curl` command and you wont need to use the subshell and pipes for the `host_address`.

joaofnfernandes: From a scripting perspective your suggestion is way better, but I would prefer to keep the variable as a full url that users can copy-paste from their browsers.

In the past the docs and DTR bootrapper had the concept of `url` vs `domain/hostname` and users were totally confused by it. When users looked at `domain name` they were not even sure what they should put into that field.
The way we addressed that was by always accepting a url (even when it doesn't make sense), and by stripping the protocol from it.

But if you have a better way to achieve that, I'm up for suggestions. This sed stuff is hard to digest, specially if you want it to be platform independent. 😞

PR #5355, opened 2017-11-21
Improve archival dockerfile v1.13

mistyhacks:
Same work as #5351 but for 1.13.


thaJeztah:
Same problem on this branch (and already the same problem online; https://docs.docker.com/v1.13/) due to `/metadata.txt` being referenced by the script, the menu is not working

PR #5358, opened 2017-11-21
Improve archival dockerfile 1.11

mistyhacks:
Same work as #5351 but for v1.11 (which already has only HTML in the source).


mistyhacks:
Queried Netlify about this one.

_config.yml
```diff
@@ -1 +1 @@
-exclude: [src]
+exclude: ["src","highlight","hooks"]
```
thaJeztah: Same in this PR; "highlight" should not be excluded

PR #5359, opened 2017-11-21
Improve archival dockerfile v1.10

mistyhacks:
Same work as #5351 but for v1.10 (which already has only HTML in the source).


_config.yml
```diff
@@ -1 +1 @@
-exclude: [src]
+exclude: ["src","highlight","hooks"]
```
thaJeztah: "highlight" should not be excluded here (or we need to update the HTML files to remove references to it)

PR #5361, opened 2017-11-22
Improve archival dockerfile 1.8

mistyhacks:
Same work as #5351 but for v1.8 (which already has only HTML in the source).


_config.yml
```diff
@@ -1 +1 @@
-exclude: [src]
+exclude: ["src","highlight","hooks"]
```
thaJeztah: "highlight" should not be excluded here; it contains the javascript that we still use for this version

PR #5365, opened 2017-11-22
Improve archival dockerfile 1.5

mistyhacks:
Same work as #5351 but for v1.5 (which already has only HTML in the source).


Dockerfile
```diff
@@ -1,3 +1,16 @@
...
-CMD jekyll serve -d /_site --watch -H 0.0.0.0 -P 4000
+# Build the docs from this branch
+COPY . /source
+RUN jekyll build --source /source --destination /site
+
+# Reset to bare Nginx image
+FROM nginx:alpine
+
+# Copy the Nginx config
+COPY --from=builder /conf/nginx-overrides.conf /etc/nginx/conf.d/default.conf
```
thaJeztah: We could use the same as https://github.com/docker/docker.github.io/pull/5366/files#r152847830 here, but up to you :smile:

PR #5366, opened 2017-11-22
Improve archival dockerfile 1.4

mistyhacks:
Same work as #5351 but for v1.4 (which already has only HTML in the source).

PR #5357, opened 2017-11-21
Improve archival dockerfile v17.06

mistyhacks:
Same work as #5351 for 17.06

PR #5356, opened 2017-11-21
Improve archival dockerfile v17.03

mistyhacks:
Same work as #5351 for v17.03

PR #5367, opened 2017-11-22
Use multistage builds for docs-base

mistyhacks:
Depends on #5366 #5365 #5364 #5363 #5361 #5360 #5359 #5358 #5357 #5356 #5355 #5351
Temporarily blocks #5231 

There is a chicken-egg problem with this PR and its dependencies on Docker Cloud with auto-deploy. It's also hard to test this. You really need to apply each of the depends to its target branch locally, build and tag those Dockerfiles, and then build this one using the locally-tagged images, to get a sense if it works. There will need to be a modification to the `master` build too, because we are changing the contract of where the static HTML provided by `docs-base` lives.

It is possible that with multi-stage, we won't even need `docs-base` anymore. We might just need a `master-with-archives` and a `master-lite` that doesn't have archives. We may be able to do this in the same Dockerfile by short-circuiting? @thaJeztah WDYT?


Dockerfile
```diff
@@ -1,75 +1,76 @@
...
+RUN bash /usr/bin/fix_archives.sh ${TARGET} 'v1.12'
+
+COPY --from=archive_v1.13 /etc/nginx/html ${TARGET}/v1.13
+RUN bash /usr/bin/fix_archives.sh ${TARGET} 'v1.13'
+
+COPY --from=archive_v17.03 /etc/nginx/html ${TARGET}/v17.03
+RUN bash /usr/bin/fix_archives.sh ${TARGET} 'v17.03'
+
+COPY --from=archive_v17.06 /etc/nginx/html ${TARGET}/v17.06
+RUN bash /usr/bin/fix_archives.sh ${TARGET} 'v17.06'
```
thaJeztah: Ok for a follow-up, but what do you think about adding the _current_ documentation as an "archive" as well? This allows using perma-links to sections in the documentation for the current release

.Dockerignore
```diff
@@ -0,0 +1 @@
+_site
```
thaJeztah: can you rename this file to be lowercase (`.dockerignore`)?

Also; looks like we no longer need Git during build, so we can now ignore `.git`, `.github`, and likely the `tests` directory (There's some other files we don't need, but excluding them likely won't improve performance);

```dockerignore
.git
.github
_site
tests
```

thaJeztah: Actually, this is the `docs-base` branch, so only `.git` needs to be added to `.dockerignore`. Doing so makes a big difference though:

Before:

```bash
$ docker build -t docs:base .
Sending build context to Docker daemon  541.2MB
...
```

After:

```bash
$ docker build -t docs:base .
Sending build context to Docker daemon  13.82kB
...
```

Dockerfile
mistyhacks: It's an interesting idea for later, but could wreck SEO if we did it wrong.

thaJeztah: We may need [`canonical` headers](https://moz.com/blog/canonical-url-tag-the-most-important-advancement-in-seo-practices-since-sitemaps) to prevent the duplicate content penalty (same for all archives)

mistyhacks:
Not sure why Netlify is even trying to build this. It's not relevant.

nginx-overrides.conf
```diff
@@ -0,0 +1,5 @@
+server {
+  port_in_redirect off;
+  listen 4000;
+  error_page 403 404 /404.html;
```
thaJeztah: Can you check my comments on https://github.com/docker/docker.github.io/pull/5351 w.r.t. the configuration file, and the target directory for the html (`/usr/share/nginx/html`)?

Dockerfile
```diff
@@ -1,75 +1,76 @@
...
+## to the version, to make a new archive
 
 # This index file gets overwritten, but it serves a sort-of useful purpose in
 # making the docs/docs-base image browsable:
+COPY index.html ${TARGET}/
 
-COPY index.html target
+# Reset with nginx again, so we don't get scripts or extra apps in the final image
+FROM nginx:alpine
+COPY --from=builder /conf/nginx-overrides.conf /etc/nginx/conf.d/000-default.conf
```
thaJeztah: can you change `000-default.conf` to `default.conf` here? I later noticed there's an existing `default.conf` that we need to overwrite.

Dockerfile
```diff
@@ -1,75 +1,76 @@
...
+COPY --from=docs_base ${TARGET} ${TARGET}
```
thaJeztah: `$TARGET` is not set in this stage, so we need to set it here, or just use fixed values

mistyhacks:
Addressed feedback here.

johndmulhausen:
@mistyhacks @thaJeztah Is this PR in conflict with https://github.com/docker/docker.github.io/pull/5389 ? Which one should we consider canonical?

mistyhacks:
No, i've integrated his change into mine. Need to get back into this and find my place.

Dockerfile
```diff
@@ -1,75 +1,39 @@
...
-
-# Serve the site (target), which is now all static HTML
-
-CMD echo "Docker docs are viewable at:" && echo "http://0.0.0.0:4000" && nginx -g 'pid /tmp/nginx.pid;'
+# Start over with nginx:alpine
+FROM nginx:alpine AS docs_base
+
+# Set TARGET to the default location where Nginx serves files from
+# (and where they are stored in all the archives)
+ENV TARGET=/usr/share/nginx/html
```
thaJeztah: nit: we could change this to an `ARG`, then the `TARGET` env-file isn't persisted in the image

Dockerfile
```diff
@@ -1,75 +1,39 @@
...
+
+# For each of these blocks
+# - Get an archive's HTML
+# - Run fix-archives.sh, which calls scripts to fix relative links, minify
```
thaJeztah: oh, actually; this comment is no longer current; we're now only copying the archives, nothing else (no processing)

mistyhacks:
We don't need this one anymore!!! 💃 🕺

PR #5369, opened 2017-11-22
Use multistage build for master

mistyhacks:
Depends on #5367 and **all its dependencies!**

We don't need to install Bash anymore (we get it from docs-base). We don't need to compress the number of layers anymore by complicated Bash logic.

I clean things up at the very end to ensure that the final image only contains static HTML and the Nginx config.


Dockerfile
```diff
@@ -1,35 +1,87 @@
...
+
+# BUILD OF MASTER DOCS IS NOW DONE!
+# Reset to alpine so we don't get any docs source or extra apps
+FROM nginx:alpine
+
+# Set the target again
+ENV TARGET=/usr/share/nginx/html
+
+# Get the built docs output from the previous step
+COPY --from=builder ${TARGET} ${TARGET}
```
thaJeztah: We should move this one after the archives, so that the archives copy step can be cached, and a change in master doesn't require updating the archives.

PR #5382, opened 2017-11-23
Remove some temp-files

thaJeztah:
Looks like these were accidentally committed :smile:

ping @mistyhacks


thaJeztah:
Also removed some images that were not referenced

thaJeztah:
Actually; wondering now if we should rename these back to` .md`, and create a /source directory (e.g.) so that we can re-generate the docs

mistyhacks:
Not unless we can guarantee that all the .md files are represented and that the contents of them is synced with the latest actual output. I think it's risky. I like the idea of removing them.

PR #5383, opened 2017-11-23
[v1.5] Remove tempfiles and unused images

thaJeztah:
ping @mistyhacks


thaJeztah:
Actually; wondering now if we should rename these back to .md, and create a /source directory (e.g.)

PR #5384, opened 2017-11-23
[v1.6] Remove tempfiles

thaJeztah:


thaJeztah:
Actually; wondering now if we should rename these back to `.md`, and create a `/source` directory

PR #5389, opened 2017-11-25
[WIP] More multistage build

thaJeztah:
Based on https://github.com/docker/docker.github.io/pull/5367 - commit messages contain description of each change, but more explanation follow :smile:

Also see the last commit for some testing for follow-up improvements 👍


thaJeztah:
Doing a quick comparison with this and https://github.com/docker/docker.github.io/pull/5367, each one doing three runs of;

```
time docker build --no-cache -t docs/docker.github.io:docs-base .
```

On https://github.com/docker/docker.github.io/pull/5367 :

```
Successfully tagged docs/docker.github.io:docs-base
      371.86 real         0.24 user         0.17 sys

Successfully tagged docs/docker.github.io:docs-base
      366.03 real         0.25 user         0.18 sys

Successfully tagged docs/docker.github.io:docs-base
      367.75 real         0.23 user         0.13 sys
```

And on this branch:

```
Successfully tagged docs/docker.github.io:docs-base
      261.24 real         0.21 user         0.16 sys

Successfully tagged docs/docker.github.io:docs-base
      250.81 real         0.20 user         0.10 sys

Successfully tagged docs/docker.github.io:docs-base
      254.16 real         0.20 user         0.17 sys
```

Of course, that's not taking into account time to _pull_ the (archive) images

thaJeztah:
And size-wise:

on https://github.com/docker/docker.github.io/pull/5367

```
REPOSITORY              TAG                 IMAGE ID            CREATED              SIZE
docs/docker.github.io   docs-base           68248db99856        44 seconds ago       1.42GB
```

Ad on this branch:

```
REPOSITORY              TAG                 IMAGE ID            CREATED             SIZE
docs/docker.github.io   docs-base           ae92cc3968fe        4 minutes ago       1.36GB
```

johndmulhausen:
For the layperson:

The previous docs-base image looped through the various archive branches in github and ran `jekyll build` on each of them using a full-fat GitHub Pages docker image, to compile the final site you see at docs.docker.com.

However we also build docker images of each archive version. You can do, for example, `docker run -ti -p 4000:4000 docs/docker.github.io:v1.13` and see the old docs there - already built, having gone through the Jekyll process. 

What this PR does is take advantage of that fact by using Docker's new multi-stage build process. It pulls each archive image from `docs/docker.github.io:*` and copies the HTML output from there. This should save quite a bit of time.

In addition, this PR's final "stage" in the multi-stage build only copies forward the HTML output from both master (the current docs) and the aforementioned archive docs, leaving behind the GitHub Pages environment that was run on the current docs. This means that the final image is just static files and nginx -- which is not only smaller but passes Docker Security Scanning with a clean slate. 

Finally, the PR uses minification to reduce the size of the final static assets (see the `compress_assets.sh` script).

mistyhacks:
Closed in favor of #5369

PR #5390, opened 2017-11-25
Fix incorrect redis volumes

c-dy:
Redis will not start with the existing config. You can see in the terminal recap that the volumes config should in fact be ".data:/data".


c-dy:
Actually, now that I've tried with a swarm on aws, it requires the full /home/docker/data version. Locally it would not work. Ah well, closing this.

PR #5403, opened 2017-11-28
Update ubuntu.md

knil-sama:
### Proposed changes

Add /dists/ subdirectory for the download key command


knil-sama:
command was correct

PR #5424, opened 2017-12-01
Update fedora.md

javabrett:
Added Fedora 26, which is served from https://download.docker.com/linux/fedora/26/ .

### Proposed changes
Added Fedora 26 to list of supported versions.  It is available for download.


johndmulhausen:
Looks good, thanks @javabrett, cc @mistyhacks

PR #5427, opened 2017-12-01
Logentries log driver update

ikarpovich:
### Proposed changes

Documentation updated to include `line-only` log option

### Related issues (optional)

Described in moby/moby#31726 together with later changes from moby/moby#35626


joaofnfernandes:
It seems that the option shipped with 17.09, so let's merge this 🚀

PR #5433, opened 2017-12-03
Pointing wrong URI to prometheus alert

asleea88:
### Proposed changes
It seems to be pointing wrong or legacy URI(https://prometheus.io/docs/alerting/rules/)
The [URI](https://prometheus.io/docs/alerting/rules/) returns "Page Not found".
I have found an alternative URI(https://prometheus.io/docs/alerting/overview/) for Prometheus alert.

I just exchange 'https://prometheus.io/docs/alerting/rules/' to 'https://prometheus.io/docs/alerting/overview/'


johndmulhausen:
Yep, looks good @asleea88, thanks a lot! 👍

PR #5436, opened 2017-12-04
Reason for adding dotnet runtime image

Pothulapati:
Adding a runtime image on top of sdk causes confusion as the sdk can perform all the operations. Using the runtime image reduces the deployments size.

Thank You


johndmulhausen:
I think the logic is self-explanatory in this Dockerfile so long as people are familiar with multi-stage builds. But thanks for the PR, @Pothulapati :)

Issue #5263, opened 2017-11-09
No Minimum Spec listed for Windows Node.  


ekitagawa:
[datacenter/ucp/2.2/guides/admin/install/system-requirements.md](https://docs.docker.com/datacenter/ucp/2.2/guides/admin/install/system-requirements/), CC @JimGalasyn  

Could we have MSR listed for Windows worker nodes?

joaofnfernandes:
@ekitagawa you're right. We'll follow-up on this one.

friism:
I think the current (4GB RAM) is adequate. I'm adding a note about disk space.

joaofnfernandes:
Closed by #5437

PR #5437, opened 2017-12-04
update system requirements

friism:
Addresses https://github.com/docker/docker.github.io/issues/3656

### DIFF
--- a/datacenter/ucp/2.2/guides/admin/install/system-requirements.md
+++ b/datacenter/ucp/2.2/guides/admin/install/system-requirements.md
@@ -29,14 +29,7 @@ You can install UCP on-premises or on a cloud provider. Common requirements:
  
 Note that Windows container images are typically larger than Linux ones and for that reason, you should consider provisioning more local storage for Windows nodes and for DTR setups that will store Windows container images.
 
-Also, make sure the nodes are running one of these operating systems:
-
-* A maintained version of CentOS 7. Archived versions aren't supported or tested
-* Red Hat Enterprise Linux 7.0, 7.1, 7.2, 7.3, or 7.4
-* Ubuntu 14.04 LTS or 16.04 LTS
-* SUSE Linux Enterprise 12
-* Oracle Linux 7.3
-* Windows Server 2016 or Windows Server 1709 (1709 will only support Docker EE Basic)
+Also, make sure the nodes are running an [operating system support by Docker EE](https://success.docker.com/Policies/Compatibility_Matrix).
 
 For highly-available installations, you also need a way to transfer files
 between hosts.
--- a/datacenter/ucp/2.2/guides/admin/install/system-requirements.md
+++ b/datacenter/ucp/2.2/guides/admin/install/system-requirements.md
@@ -9,23 +9,34 @@ Before installing, be sure your infrastructure has these requirements.
 
 ## Hardware and software requirements
 
-You can install UCP on-premises or on a cloud provider. To install UCP,
-all nodes must have:
+You can install UCP on-premises or on a cloud provider. Common requirements:
 
 * [Docker Enterprise Edition](/engine/installation/index.md) version 17.06 or higher
 * Linux kernel version 3.10 or higher
+* A static IP address
+
+### Minimum requirements
+
 * 8GB of RAM for manager nodes or nodes running DTR
 * 4GB of RAM for worker nodes
 * 3GB of free disk space
-* A static IP address
+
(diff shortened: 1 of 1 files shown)


datacenter/ucp/2.2/guides/admin/install/system-requirements.md
```diff
@@ -9,23 +9,34 @@ Before installing, be sure your infrastructure has these requirements.
...
+### Minimum requirements
+
 * 8GB of RAM for manager nodes or nodes running DTR
 * 4GB of RAM for worker nodes
 * 3GB of free disk space
-* A static IP address
+
+### Recommended production requirements
+
+ * 16GB of RAM for manager nodes or nodes running DTR
```
pdevine: 16 GB should work with any large instance.  For POCs, we can recommend 2 GB of RAM

datacenter/ucp/2.2/guides/admin/install/system-requirements.md
```diff
@@ -9,23 +9,34 @@ Before installing, be sure your infrastructure has these requirements.
...
+ * 4 vCPUs for manager nodes or nodes running DTR
+ * 25-100GB of free disk space
+ 
+Note that Windows container images are typically larger than Linux ones and for that reason, you should consider provisioning more local storage for Windows nodes and for DTR setups that will store Windows container images.
 
 Also, make sure the nodes are running one of these operating systems:
 
-* A maintained version of CentOS 7. Archived versions aren't supported or tested.
+* A maintained version of CentOS 7. Archived versions aren't supported or tested
 * Red Hat Enterprise Linux 7.0, 7.1, 7.2, 7.3, or 7.4
```
pdevine: This should be 7.2+ IMO.  I don't think anyone is testing w/ 7.0 or 7.1, and the base kernels in those don't even work well w/ docker (I don't think DTR will even install).

johndmulhausen:
Page in preview: https://deploy-preview-5437--docker-docs.netlify.com/datacenter/ucp/2.2/guides/admin/install/system-requirements/

cc @JimGalasyn and @joaofnfernandes so they know to carry this forward to UCP 2.5

datacenter/ucp/2.2/guides/admin/install/system-requirements.md
joaofnfernandes: The compatibility matrix says that we support 7.1, but I'm not sure if it is correct. Either way the two docs need to be consistent.
https://success.docker.com/Policies/Compatibility_Matrix

datacenter/ucp/2.2/guides/admin/install/system-requirements.md
joaofnfernandes: @pdevine does it make sense to mention DTRs running the security scanner separately?

datacenter/ucp/2.2/guides/admin/install/system-requirements.md
```diff
@@ -9,23 +9,34 @@ Before installing, be sure your infrastructure has these requirements.
...
 * 3GB of free disk space
-* A static IP address
+
+### Recommended production requirements
+
+ * 16GB of RAM for manager nodes or nodes running DTR
+ * 4 vCPUs for manager nodes or nodes running DTR
+ * 25-100GB of free disk space
+ 
+Note that Windows container images are typically larger than Linux ones and for that reason, you should consider provisioning more local storage for Windows nodes and for DTR setups that will store Windows container images.
```
joaofnfernandes: @friism does it make sense to also mention that admins should configure Docker EE to store logs using an external system? By default I think that the Daemon doesn't delete logs, which means that in a production system with intense usage, your logs will also take a significant amount of disk space.

datacenter/ucp/2.2/guides/admin/install/system-requirements.md
friism: I've updated to point to compat matrix instead of duplicating

datacenter/ucp/2.2/guides/admin/install/system-requirements.md
friism: @joaofnfernandes I think that's a worthy recommendation, but I also want to keep this doc tight and focused on system requirements. Otherwise it'll just morph into a kitchen-sink guide with assorted random recommendations.

joaofnfernandes:
@friism does this contribution also close #5263?

friism:
ping @ekitagawa - I think Windows is ok for now

PR #5440, opened 2017-12-04
Adding reference to Hardware Requirement

ekitagawa:
DTR has higher requirement than regular worker nodes, and it is somehow listed on UCP section instead of here.  Adding reference.


joaofnfernandes:
I've updated your suggestion to not include the `https://` component of the link. Thanks for the contribution 🙌

PR #5444, opened 2017-12-05
Fix typos in https://docs.docker.com/get-started/part4/

buiquangduc:
### Proposed changes

Today I read this article: https://docs.docker.com/get-started/part4/ and notice that there is a typo in ```commands``` text.
image

Anyway, thanks for all these awesome articles on https://docs.docker.com.


joaofnfernandes:
Thanks for helping keep the docs free of typos! Let's ship it 🚀

PR #5448, opened 2017-12-06
Remove unimplemented bullet

mvetsch:
Since there are "Deprecated" images on docker hub with the "Official Repository" flag, this bullet might should be removed. 

Django was not patched for more than a year: 
"This image is officially deprecated in favor of the standard python image, and will receive no further updates after 2016-12-31 (Dec 31, 2016). Please adjust your usage accordingly"

https://hub.docker.com/r/library/django/


johndmulhausen:
For non-deprecated images it is still the case, so we'd like to leave this. Thanks!

PR #5450, opened 2017-12-06
dtr: promote current_issues in rethinkcli docs

trapier:
This section of the docs is about troubleshooting and table quorum is a common
issue so we should make the `current_issues` table most obvious.


joaofnfernandes:
Awesome 🚀

PR #5453, opened 2017-12-06
Add instructions to activate zsh completion in Docker for Mac

OmeGak:
### Proposed changes

I simply added instructions on how to set up zsh completion in Docker for Mac because they were missing.

### Related issues

Related to https://github.com/docker/for-mac/issues/1948


joaofnfernandes:
I think you could have split this in two sections like
```
## Install shell completion
### Bash
### Zsh
```

But this looks good, so let's ship it and if you feel like it, file another PR to implement my suggestion 😄

PR #5456, opened 2017-12-07
Fix parameter links in compose-versioning.md

lachlancooper:


joaofnfernandes:
This is really interesting... I expected the relative urls to work as they are, but I've confirmed that they're not currently working 😒
Thanks for stopping and fixing this, let's ship it 🚀

PR #5458, opened 2017-12-07
Update imagesandcontainers.md

slawiko:
### Proposed changes

This PR just removes spaces in imagesandcontainers.md for better read experience. 

Here is an example why these spaces are ugly
example

### DIFF
--- a/engine/userguide/storagedriver/imagesandcontainers.md
+++ b/engine/userguide/storagedriver/imagesandcontainers.md
@@ -235,9 +235,9 @@ layers are the same.
     ```bash
     $ docker images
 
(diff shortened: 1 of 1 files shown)


johndmulhausen:
See: https://deploy-preview-5458--docker-docs.netlify.com/engine/userguide/storagedriver/imagesandcontainers/

slawiko:
@johndmulhausen I think it's better. 

> Run the `docker ps` command to verify the 5 containers are running.

this point don't look nice, but there is no possibility to make it nicer

mistyhacks:
@slawiko Thanks for the reformatting. I don't know if you are aware, but you can scroll long code blocks horizontally. It isn't as nice on mobile, but on a system with a mouse or trackpad it works well.

PR #5466, opened 2017-12-07
Note about setting ipv4/ipv6 not working in swarm

new-guy:
### Proposed changes

I spent a few hours yesterday working under the assumption that it was possible to set a static IP address for in a swarm, but it turns out it isn't.  Hopefully this will help prevent others from making the same mistake I did.


mistyhacks:
@shin- PTAL

shin-:
"Docker" needs to be capitalized, otherwise LGTM

PR #5467, opened 2017-12-08
clarify EE release schedule

friism:


johndmulhausen:
@friism  - wouldn't you rather just delete two of the quarterly green checks for Docker EE, rather than remove the column entirely?

friism:
@johndmulhausen im not sure we know with 100% certainty in what month future ee versions will arive

PR #5469, opened 2017-12-08
Navigation under "Reference" and "Manuals," Registry warning, "Latest" UCP/DTR

johndmulhausen:
- Navigation under "Reference" and "Manuals" is revamped 
- /registry has an admonition clarifying that this is NOT "Docker Trusted Registry"
- "Latest" UCP/DTR variables in _config.yml makes linking to `{{ site.dtr_version }}` possible, as with the various Engine variables.

### DIFF
--- a/_data/toc.yaml
+++ b/_data/toc.yaml
@@ -207,6 +207,7 @@ guides:
           title: Overview
         - path: /develop/sdk/examples/
           title: SDK and API examples
+
 - sectiontitle: Configure networking
   section:
   - path: /engine/userguide/networking/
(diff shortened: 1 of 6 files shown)


_data/toc.yaml
```diff
@@ -544,783 +546,805 @@ guides:
...
-    - path: /engine/api/v1.20/
-      title: v1.20 reference
-    - path: /engine/api/v1.19/
-      title: v1.19 reference
-    - path: /engine/api/v1.18/
-      title: v1.18 reference
-
-- sectiontitle: Docker Enterprise Edition references
-  section:
-  - title: Truste
```
mistyhacks: Really?

PR #5474, opened 2017-12-09
Fixed a typo

hKaspy:
Fixed a minor typo, looks like the text vas coppied from Bind Mounts and the anchor was not edited accordingly.


johndmulhausen:
Thanks, @hKaspy 👍

PR #5477, opened 2017-12-10
Update django.md

chan24092005:
this command should run with `sudo`, otherview it will be fail with this error:
```
ERROR: Couldn't connect to Docker daemon at http+docker://localunixsocket - is it running?
```
Solution refer from https://github.com/docker/compose/issues/1214.


johndmulhausen:
Thanks to the link for context, @chan24092005 :) 👍

PR #5480, opened 2017-12-10
Recommend alpine base image

tjamet:
Hi,

I noticed that the recommended image is debian while I notice more and more recommendations using alpine as a base image.
This PR suggests users to use the alpine base image, allowing building smaller images.

Please consider this PR as a question opening discussion upon what are the latest recommendations regarding base images.

### Proposed changes

Suggest users to use alpine base image by default.


johndmulhausen:
Definitely a better recommendation. Thanks, @tjamet 👍

PR #5481, opened 2017-12-11
Add multi stage build best practices

tjamet:
Hi,

May I propose to add a best practice section regarding multi stage builds. Suggesting users best practices to benefit build cache, not rebuilding unnecessary things, for building applications while keeping very small final images.

### DIFF
--- a/engine/userguide/eng-image/dockerfile_best-practices.md
+++ b/engine/userguide/eng-image/dockerfile_best-practices.md
@@ -69,6 +69,47 @@ drastically reduce the size of your final image, without the need to
 jump through hoops to reduce the number of intermediate layers or remove
 intermediate files during the build.
 
+Images being built by the final stage only, you can most of the time benefit
+both the build cache and minimize images layers.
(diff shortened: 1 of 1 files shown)


joaofnfernandes:
It seems that we already have some information about multi-stage builds in this article on line 64. Do you care to move your changes closer to that, instead of putting them in the section `The Dockerfile instructions`?

engine/userguide/eng-image/dockerfile_best-practices.md
```diff
@@ -167,6 +167,53 @@ the command string itself will be used to find a match.
 Once the cache is invalidated, all subsequent `Dockerfile` commands will
 generate new images and the cache will not be used.
 
+### Optimizing [multi-stage builds](multistage-build.md)
```
mistyhacks: No links or Markdown in titles

tjamet: done

PR #5499, opened 2017-12-13
No need to mention a fix to a test failure

alb-i986:
If I understand correctly, it's just a fix to a test failure.
Therefore I don't think it's relevant to mention in the product release notes.


mistyhacks:
We don't edit these release notes, but pull them from the upstream repo intact. So we won't be making this change here. If you really feel like it should be removed, then raise it in `docker/docker-ce` under the appropriate branch and then ping us here to update the release notes after it is merged there. Thanks!

alb-i986:
Ping
https://github.com/docker/docker-ce/pull/363

Cheers

alb-i986:
Hey ya. Pinging again.

PR #5504, opened 2017-12-13
Mac: FAQ: Disk usage in raw format

akimd:
### Proposed changes

Describe disk usage of Raw format.

This is a new feature of Docker for Mac.  Questions and PRs have
already been made about this, and we expect more of them.

This currently applies only to Edge releases, there's no stable with this feature yet.  But I wouldn't wait to make this public, we already had issues with this (e.g., https://github.com/docker/pinata/issues/8358)

### DIFF
--- a/docker-for-mac/faqs.md
+++ b/docker-for-mac/faqs.md
@@ -195,36 +195,39 @@ Starting with High Sierra with APFS (Apple Filesystem) enabled, Docker
 uses the "raw" format (`Docker.raw`), otherwise it uses the Qcow2
 format (`Docker.qcow2`).
 
-#### `Docker.raw` consumes an insane amount of disk space!
+#### Docker.raw consumes an insane amount of disk space!
 
 This is an illusion.  Docker uses the raw format on Macs running the
-Apple Filesystem (APFS).  APFS supports sparse files, i.e., files
-where long series of "zeros" are compressed.  Calling `ls` is
-misleading, because it reveals the logical size of the file, not its
-physical size; be sure to use `-s` to display the physical size:
+Apple Filesystem (APFS).  APFS supports _sparse files_, which compress
+long runs of zeroes representing unused space.  The output of `ls` is
+misleading, because it lists the logical size of the file rather than
+its physical size. To see the physical size, add the `-ks` switch; to
+see the logical size in human readable form, add `-lh`:
 
 ```bash
-..ker/Data/com.docker.driver.amd64-linux $ ls -klsh Docker.raw
+$ cd ~/Library/Containers/com.docker.docker/Data/com.docker.driver.amd64-linux
+$ ls -klsh Docker.raw
 2333548 -rw-r--r--@ 1 akim  staff    64G Dec 13 17:42 Docker.raw
 ```
 
-here, the logical size (or max space allowed) is 64GB, but only
-2,333,548KB are physically consumed.
+In this listing, the logical size is 64GB, but the logical size is
+only 2.3GB.
 
-Alternatively, you may use GNU ls or `du` (disk usage):
+Alternatively, you may use `du` (disk usage):
 
 ```bash
-..ker/Data/com.docker.driver.amd64-linux $ gls -slh Docker.raw
-2,3G -rw-r--r-- 1 akim staff 64G Dec 13 17:44 Docker.raw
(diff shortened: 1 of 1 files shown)


docker-for-mac/faqs.md
```diff
@@ -184,7 +184,43 @@ For full information on adding server and client side certs, see [Adding
...
+#### Qcow2 or Raw?
+
+By default Docker for Mac stores containers and images in a file
+`Docker.raw` or `Docker.qcow2` in the directory
+`~/Library/Containers/com.docker.docker/Data/com.docker.driver.amd64-linux`.
+Starting with High Sierra with APFS (Apple Filesystem) enabled, Docker
+uses the "raw" format (`Docker.raw`), otherwise it uses the Qcow2
+format (`Docker.qcow2`).
+
+#### `Docker.raw` consumes an insane amount of disk space!
```
mistyhacks: No markdown in headings, please

docker-for-mac/faqs.md
```diff
@@ -184,7 +184,43 @@ For full information on adding server and client side certs, see [Adding
...
+
+This is an illusion.  Docker uses the raw format on Macs running the
+Apple Filesystem (APFS).  APFS supports sparse files, i.e., files
```
mistyhacks: reword to avoid the "i.e.": APFS supports _sparse files_, which compress long runs of zeroes representing unused space. The output of `ls` is misleading, because it lists the logical size of the file rather than its physical size. To see the physical size, add the `-s` switch:

docker-for-mac/faqs.md
```diff
@@ -184,7 +184,43 @@ For full information on adding server and client side certs, see [Adding
...
+where long series of "zeros" are compressed.  Calling `ls` is
+misleading, because it reveals the logical size of the file, not its
+physical size; be sure to use `-s` to display the physical size:
+
+```bash
+..ker/Data/com.docker.driver.amd64-linux $ ls -klsh Docker.raw
```
mistyhacks: Simplify the prompt to just `$`

Also you don't need all these flags, just use `-k` or `-h`. Maybe explain that `-l` allows the long listing and `-h` uses a human-readable size.

docker-for-mac/faqs.md
```diff
@@ -184,7 +184,43 @@ For full information on adding server and client side certs, see [Adding
...
+2333548 -rw-r--r--@ 1 akim  staff    64G Dec 13 17:42 Docker.raw
+```
+
+here, the logical size (or max space allowed) is 64GB, but only
```
mistyhacks: In this listing, the logical size (the maximum possible size of the file) is 64 GB, but the logical size is only 2.33 GB.

docker-for-mac/faqs.md
```diff
@@ -184,7 +184,43 @@ For full information on adding server and client side certs, see [Adding
...
+2,333,548KB are physically consumed.
+
+Alternatively, you may use GNU ls or `du` (disk usage):
```
mistyhacks: I'd remove this whole part. This implies a lot (that they have Homebrew or Macports, that they installed gnu ls, that they did not tell it to override the system ls). We don't need to tell them how to use extra tools they may or may not have installed.

docker-for-mac/faqs.md
```diff
@@ -184,7 +184,43 @@ For full information on adding server and client side certs, see [Adding
...
+```bash
+..ker/Data/com.docker.driver.amd64-linux $ gls -slh Docker.raw
+2,3G -rw-r--r-- 1 akim staff 64G Dec 13 17:44 Docker.raw
+..ker/Data/com.docker.driver.amd64-linux $ du -h Docker.raw
+2,2G	Docker.raw
+```
+
+#### How do I reduce the size of Docker.qcow2?
 
 By default Docker for Mac stores containers and images in a file
```
mistyhacks: Update this text.

docker-for-mac/faqs.md
akimd: Hi.

My point was to show the directory we're in.  So I guess I'll add a `cd` before.

I can sure teach the meaning of the flags, but I believed that what matters is how it's displayed, and this is not a documentation for `ls`.

docker-for-mac/faqs.md
akimd: So I'll leave just `du`, ok.

PR #5508, opened 2017-12-13
Trivial typo fix, but it stood out

daniel347x:


joaofnfernandes:
Thanks for helping keep the docs updated 🚀

PR #5517, opened 2017-12-14
Add kubectl context instructions and tighten text

### LINKED ISSUE #5512: Update D4M Kubernetes docs with kubectl context info
Users must uninstall minikube to use K8s with D4M.

gbarr01:
Users with `kubectl` already installed and pointing to minikube or a gke cluster must change context to `docker-for-desktop`.

Fixes #5512

### DIFF
--- a/docker-for-mac/index.md
+++ b/docker-for-mac/index.md
@@ -297,7 +297,7 @@ choose to discard or not apply changes when asked.
 
 ### Kubernetes
 
-**Kubernetes is only available in the private beta of Docker for Mac 17.12 CE Edge through the [Docker Beta program](https://beta.docker.com/). To access beta builds, you must be signed in with your Docker ID within Docker for Mac: select ![whale menu](/docker-for-mac/images/whale-x.png){: .inline} -> Sign in / Create Docker ID from the menu bar.**
+**Kubernetes is only available in Docker for Mac 17.12 CE Edge to participants in the [Docker Beta program](https://beta.docker.com/). To access beta builds, you must be signed in with your Docker ID within Docker for Mac: select ![whale menu](/docker-for-mac/images/whale-x.png){: .inline} -> Sign in / Create Docker ID from the menu bar.**
 
 If you are participating in the Docker beta program, you can access Docker for
 Mac 17.12 CE Edge, which includes a standalone Kubernetes server that runs on
--- a/docker-for-mac/index.md
+++ b/docker-for-mac/index.md
@@ -309,9 +309,11 @@ pointing to some other environment, such as minikube or a gke cluster, be sure
 to change context so that kubectl is pointing to `docker-for-desktop`:
 
 ```
-kubectl config view
+kubectl config get-contexts
 kubectl config use-context docker-for-desktop
 ```
+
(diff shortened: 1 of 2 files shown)


docker-for-mac/index.md
```diff
@@ -297,15 +297,23 @@ choose to discard or not apply changes when asked.
 
 ### Kubernetes
 
-**Kubernetes is only available if you are part of the private beta for Docker for Mac 17.12. To access beta builds, you must be signed in within Docker for Mac using your Docker ID.**
+**Kubernetes is only available in the private beta of Docker for Mac 17.12 CE Edge through the [Docker Beta program](https://beta.docker.com/). To access beta builds, you must be signed in with your Docker ID within Docker for Mac: select ![whale menu](/docker-for-mac/images/whale-x.png){: .inline} -> Sign in / Create Docker ID from the menu bar.**
```
mistyhacks: ...to participants in the private beta.... (maybe?) The deal is that 17.12 CE Edge is available to everyone, but everyone won't be able to enable Kubernetes, I think. Maybe verify that.

docker-for-mac/index.md
```diff
@@ -297,15 +297,23 @@ choose to discard or not apply changes when asked.
...
 
 If you are participating in the Docker beta program, you can access Docker for
 Mac 17.12 CE Edge, which includes a standalone Kubernetes server that runs on
 your Mac, so that you can test deploying your Docker workloads on Kubernetes.
+
 The Kubernetes client command, `kubectl`, is included and configured to connect
-to the local Kubernetes server.
+to the local Kubernetes server. If you have `kubectl` already installed and
+pointing to some other environment, such as minikube or a gke cluster, be sure
```
mistyhacks: Should `minikube` be in backticks? Should GKE be capitalized?

docker-for-mac/index.md
```diff
@@ -297,15 +297,23 @@ choose to discard or not apply changes when asked.
...
+to change context so that kubectl is pointing to `docker-for-desktop`:
+
+```
+kubectl config get-contexts
+kubectl config use-context docker-for-desktop
+```
 
-To log in with your Docker ID, select ![whale menu](/docker-for-mac/images/whale-x.png){: .inline} -> **Sign in / Create Docker ID** from the menu bar.
+If you installed `kubectl` with homebrew and encounter a symlink error, remove `/usr/local/bin/kubectl`.
```
mistyhacks: Check branding on Homebrew. I think maybe they prefer to capitalize it. There are other ways to install stuff on Mac, such as Mac Ports. So maybe make this a bit more generic.

docker-for-mac/index.md
gbarr01: @mistyhacks , are you saying that this is not enough? 
> Kubernetes is only available in the private beta of Docker for Mac 17.12 CE Edge through the [Docker Beta program](https://beta.docker.com/).

gbarr01: The section just below could use the clarification. Perhaps that is what you meant.

mistyhacks: It just reads weird to me. I think it is only available in this specific D4M build which you can only access if you are in the private beta program. Not sure how to convey that.

Issue #5177, opened 2017-10-31
stale entries in .NOT_EDITED_HERE.yaml


arehman199:
### Problem description
.NOT_EDITED_HERE.yaml references following non-existent paths that have been removed from the repository for different reasons:

```bash
user@machine:/path/to/dev/repos/docker.github.io$ ls engine/deprecated.md engine/extend registry/configuration registry/spec compliance
ls: cannot access 'engine/deprecated.md': No such file or directory
ls: cannot access 'engine/extend': No such file or directory
ls: cannot access 'registry/configuration': No such file or directory
ls: cannot access 'registry/spec': No such file or directory
ls: cannot access 'compliance': No such file or directory
user@machine:/path/to/dev/repos/docker.github.io$
```

### Suggestions for a fix
Should I go ahead and send a PR to remove these entries from  .NOT_EDITED_HERE.yaml?

arehman199:
e.g. 554b2c99679cb2e1334313483943a0b8be9767d2 removes registery/spec

johndmulhausen:
The point of `not_edited_here` is that it isn't in the repo any longer; the idea is that this file can help point people who click "edit this page" to the proper GitHub repo so that editing works correctly. However I'm noticing that doesn't work right now... Fixing.

PR #5530, opened 2017-12-15
Fix for not-edited-here Edit this Page link

johndmulhausen:
Using `path` instead of `url` will actually get us the matching we want.


johndmulhausen:
Fixes bad "Edit this Page" links per discussion in #5177

PR #5535, opened 2017-12-15
Add legacy link warning

### LINKED ISSUE #5137: links instruction to be deprecated in compose doc
[compose/compose-file/index.md](https://docs.docker.com/compose/compose-file/), CC @gbarr01 
Hello. i  think that you need to mention that `link` instruction is deprecated since it is already mentionned in docker networking documentation 
https://docs.docker.com/engine/userguide/networking/default_network/dockerlinks/

gbarr01:
Fixes #5137

### DIFF
--- a/compose/compose-file/index.md
+++ b/compose/compose-file/index.md
@@ -1098,7 +1098,7 @@ It's recommended that you use reverse-DNS notation to prevent your labels from c
 
 ### links
 
->**Warning**: >The `--link` flag is a deprecated legacy feature of Docker. It
+>**Warning**: >The `--link` flag is a legacy feature of Docker. It
 may eventually be removed. Unless you absolutely need to continue using it, we
 recommend that you use [user-defined networks](/engine/userguide/networking//#user-defined-networks)
 to facilitate communication between two containers instead of using `--link`.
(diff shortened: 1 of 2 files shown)


mistyhacks:
@thaJeztah are you OK with this wording?

thaJeztah:
Oh, you just opened a can of worms :joy:

- In the _default_ bridge network, `--link` creates an actual, uni-directional link between two containers, and has all the problems described (including restarting the target container breaking the link, and no option to re-attach) - **this is the "legacy" functionality**
- When using custom networks (irregardless the driver, could be bridge, overlay), the `--link` option is not creating an actual "link", but creates a container-scoped DNS alias for the linked container  (i.e. `container-a --link foo:container-b`; now `container-a` can access `container-b` both through its name (`container-b`), _and_ through the alias `foo`. That alias is _only_ known to `container-a`, whereas setting `docker run --network=mynetwork --network-alias=foo --name=container-a myimage` does the same, but makes the `foo` alias known to _all containers_ connected to the `my-network` network. **this usage of `--link` is not deprecated/legacy**

mistyhacks:
LGTM now, don't think we need to go into the level of detail but thanks for reiterating the background, @thaJeztah

PR #5546, opened 2017-12-16
Replace /abc.md links with /abc/ in glossary

### LINKED ISSUE #5288: Links within glossary are #term-name, should be #term<space>name
In glossary, links to multiple-word terms are sometimes given as `#term-name`, while they should be `#term<space>name`.

For example, in [glossary/swarm](https://docs.docker.com/glossary/?term=swarm) link to the definition of "swarm mode" is broken. See the [source](https://github.com/docker/docker.github.io/blob/cd44af55718728b9ca1aa835b9fd9090534f69be/_data/glossary.yaml#L277):

```
  A [swarm](/engine/swarm/) is a cluster of one or more Docker Engines running in [swarm mode](#swarm-mode).
```
Should probably be this:
```
  A [swarm](/engine/swarm/) is a cluster of one or more Docker Engines running in [swarm mode](#swarm mode).
```
Probable solutions:
* If changing links to the right form is enough, I can do it myself.
* Or maybe the docs builder should add both `#term-name` and `#term name` anchors to each term.

---

Files: [glossary.md](https://docs.docker.com/glossary/), [glossary.yaml](https://github.com/docker/docker.github.io/blob/master/_data/glossary.yaml).

 CC @johndmulhausen

NickVolynkin:
Fix other broken links in glosary.

### Proposed changes

Some links in glossary were still broken after a [recent commit](https://github.com/docker/docker.github.io/commit/97a1001fe91e94fff8b57bef0c6106dfe58784bd). This repairs them.

A typo was introduced in that commit, also fixed.

Fixes #5288

/CC @johndmulhausen

### DIFF
--- a/_data/glossary.yaml
+++ b/_data/glossary.yaml
@@ -29,7 +29,7 @@ cgroups: |
 collection: |
   A collection is a group of swarm resources that Docker EE uses for role-based
   access control. Collections enable organizing permissions for resources like
-  nodes, services, containers, volumes, networks, and secrets. [Learn how to manage collections](/datacenter/ucp/2.2/guides/access-control/manage-access-with-collections.md).
+  nodes, services, containers, volumes, networks, and secrets. [Learn how to manage collections](/datacenter/ucp/2.2/guides/access-control/manage-access-with-collections/).
 Compose: |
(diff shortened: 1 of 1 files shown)


johndmulhausen:
Thanks, @NickVolynkin 👍

PR #5558, opened 2017-12-18
Document network diagnostic tool

### LINKED ISSUE #5555: Review daemon config networking diagnosis feature for docs impact
New daemon config for networking diagnosis moby/moby#35677

mistyhacks:
Document the network diagnostic tool introduced in moby/moby#35677

Fixes #5555 

cc/ @fcrisciani

Not sure if we should really be documenting the CLI at all. It's not great that the API isn't documented in Swagger.

I'm also not 💯 clear on the use  case for this or the type of info you'd use it to get.

### DIFF
--- a/engine/userguide/networking/network-debug.md
+++ b/engine/userguide/networking/network-debug.md
@@ -1,0 +1,212 @@
+---
+description: Learn to use the built-in network debugger to debug overlay networking problems
+keywords: network, troubleshooting, debug
+title: Debug overlay or swarm networking issues
+---
+
+Docker CE 17.12 and higher introduce a network debugging tool designed to help
+debug issues with overlay networks and swarm services running on Linux hosts.
+When enabled, a network diagnostic server listens on the specified port and
+provides diagnostic information. The network debugging tool should only be
+started to debug specific issues, and should not be left running all the time.
+
+Information about networks is stored in a database, which can be examined using
+the API. The database consists of one table per network.
+
+The Docker API exposes endpoints to query and control the network debugging
+tool. CLI integration is provided as a preview, but the implementation is not
+yet considered stable and commands and options may change without notice.
+
+## Enable the diagnostic tool
+
+The tool currently only works on Docker hosts running on Linux. Repeat these
+steps for each node participating in the swarm.
+
+1.  Set the `network-diagnostic-port` to a port which is free on the Docker
+    host, in the `/etc/docker/daemon.json` configuration file.
+
+    ```json
+    “network-diagnostic-port”: <port>
+    ```
+
+2.  Get the process ID (PID) of the `dockerd` process. It is the second field in
+    the output, and is typically a number from 2 to 6 digits long.
+
+    ```bash
+    $ ps aux |grep dockerd | grep -v grep
+    ```
+
+3.  Reload the Docker configuration without restarting Docker, by sending the
+    `HUP` signal to the PID you found in the previous step.
+
+    ```bash
+    kill -HUP <pid-of-dockerd>
+    ```
+
+A message like the following will appear in the Docker host logs:
+
+```none
+Starting the diagnose server listening on <port> for commands
+```
+
+## Disable the diagnostic tool
+
+Repeat these steps for each node participating in the swarm.
+
+1.  Remove the `network-diagnostic-port` key from the `/etc/docker/daemon.json`
+    configuration file.
+
+2.  Get the process ID (PID) of the `dockerd` process. It is the second field in
+    the output, and is typically a number from 2 to 6 digits long.
+
+    ```bash
+    $ ps aux |grep dockerd | grep -v grep
+    ```
+
+3.  Reload the Docker configuration without restarting Docker, by sending the
+    `HUP` signal to the PID you found in the previous step.
+
+    ```bash
+    kill -HUP <pid-of-dockerd>
+    ```
+
+A message like the following will appear in the Docker host logs:
+
+```none
+Disabling the diagnose server
+```
+
+## Access the diagnostic tool's API
+
+The network diagnostic tool exposes its own RESTful API. To access the API,
+send a HTTP request to the port where the tool is listening. The following
+commands assume the tool is listening on port 2000.
+
+Examples are not given for every endpoint.
+
+### Get help
+
+```bash
+$ curl localhost:2000/help
+
+OK
+/updateentry
+/getentry
+/gettable
+/leavenetwork
+/createentry
+/help
+/clusterpeers
+/ready
+/joinnetwork
+/deleteentry
+/networkpeers
+/
+/join
+```
+
+### Join or leave the swarm
+
+```bash
+$ curl localhost:2000/join?members=ip1,ip2,...
+```
+
+```bash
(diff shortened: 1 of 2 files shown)


engine/userguide/networking/network-debug.md
```diff
@@ -0,0 +1,212 @@
...
+---
+
+Docker CE 17.12 and higher introduce a network debugging tool designed to help
+debug issues with overlay networks and swarm services running on Linux hosts.
+When enabled, a network diagnostic server listens on the specified port and
+provides diagnostic information. The network debugging tool should only be
+started to debug specific issues, and should not be left running all the time.
+
+Information about networks is stored in a database, which can be examined using
+the API. The database consists of one table per network.
```
fcrisciani: `The database consists of one table per network.` would remove this, it is actually 2 per network so far, but not really of interest

fcrisciani: `The database consists of one table per network.` I would remove this, the number is actually 2 at the moment, but is not really relevant

engine/userguide/networking/network-debug.md
```diff
@@ -0,0 +1,212 @@
...
+/clusterpeers
+/ready
+/joinnetwork
+/deleteentry
+/networkpeers
+/
+/join
+```
+
+### Join or leave the swarm
```
fcrisciani: Join leave the network database cluster

engine/userguide/networking/network-debug.md
```diff
@@ -0,0 +1,212 @@
...
+```bash
+$ curl localhost:2000/<method>?nid=<network id>&tname=<table name>&key=<key>[&value=<value>]
+```
+
+## Access the diagnostic tool's CLI
+
+The CLI is provided as a preview and is not yet stable. Commands or options may
+change at any time.
+
+The CLI executable is called `debugClient` and is made available using a
```
fcrisciani: right now is called `diagnosticClient`

engine/userguide/networking/network-debug.md
mistyhacks: How are you supposed to know what the table name is that you want to dump? I couldn't figure this out.

fcrisciani: the idea is that the majority of the people will use the client that abstracts this detail and deserialize the structures of the database.
The tables are the `endpoint_table` and the `overlay_peer_table`

mistyhacks:
I think one of my questions got hidden addressing your feedback, @fcrisciani . I don't understand how the database tables are named. Can I give the user a clue, since they need to specify the table name in some of the commands?

mistyhacks:
I think this is ready. PTAL again.

engine/userguide/networking/network-debug.md
```diff
@@ -0,0 +1,219 @@
...
+title: Debug overlay or swarm networking issues
+---
+
+Docker CE 17.12 and higher introduce a network debugging tool designed to help
+debug issues with overlay networks and swarm services running on Linux hosts.
+When enabled, a network diagnostic server listens on the specified port and
+provides diagnostic information. The network debugging tool should only be
+started to debug specific issues, and should not be left running all the time.
+
+Information about networks is stored in a database, which can be examined using
```
ddebroy: `stored in a database`: how about a little more details around the networkdb:
1. What kind of information is stored?
2. Mention that the database is shared/clustered across all nodes.

engine/userguide/networking/network-debug.md
```diff
@@ -0,0 +1,219 @@
...
+/createentry
+/help
+/clusterpeers
+/ready
+/joinnetwork
+/deleteentry
+/networkpeers
+/
+/join
+```
```
ddebroy: Shouldn't the `leave` endpoint be listed here too?

engine/userguide/networking/network-debug.md
```diff
@@ -0,0 +1,219 @@
...
+
+### Join or leave the network database cluster
+
+```bash
+$ curl localhost:2000/join?members=ip1,ip2,...
```
ddebroy: It might be good to clarify `ip1`, `ip2`. Maybe add a note that they are the IPs of the nodes in the swarm? Ideal would be an example with a `docker node ls` and `docker node inspect` showing the IP addresses.

engine/userguide/networking/network-debug.md
```diff
@@ -0,0 +1,219 @@
...
+| -v            | Enable verbose output.                          |
+
+### Access the CLI
+
+The CLI is provided as a container that needs to run using privileged mode.
+
+1.  To run the container, use a command like the following:
+
+    ```bash
+    $ docker container run --name net-diagnostic -d --privileged --network host fcrisciani/network-diagnostic
```
ddebroy: How about we push the containerized tool to `docker` org in hub?

mistyhacks:
Most of these are not really docs issues. @fcrisciani can you please provide some more info here?

cc/ @thaJeztah

engine/userguide/networking/network-debug.md
```diff
@@ -0,0 +1,219 @@
...
+When enabled, a network diagnostic server listens on the specified port and
+provides diagnostic information. The network debugging tool should only be
+started to debug specific issues, and should not be left running all the time.
+
+Information about networks is stored in a database, which can be examined using
+the API.
+
+The Docker API exposes endpoints to query and control the network debugging
+tool. CLI integration is provided as a preview, but the implementation is not
+yet considered stable and commands and options may change without notice.
```
thaJeztah: We should add a big fat warning that this tool should be used with care, as (IIUC) incorrect use of the tool can destroy/damage the network (it's not a read-only tool), and also expose information about your cluster's configuration that should be kept private (so don't expose this API outside of the host).

engine/userguide/networking/network-debug.md
```diff
@@ -0,0 +1,219 @@
...
+
+    ```json
+    “network-diagnostic-port”: <port>
+    ```
+
+2.  Get the process ID (PID) of the `dockerd` process. It is the second field in
+    the output, and is typically a number from 2 to 6 digits long.
+
+    ```bash
+    $ ps aux |grep dockerd | grep -v grep
```
thaJeztah: If running with systemd (perhaps it's ok to assume that's the case), it's possible to use `systemctl reload docker`

engine/userguide/networking/network-debug.md
thaJeztah: Agreed; but that can be handled separate from this PR (and this reference updated in a follow-up)

engine/userguide/networking/network-debug.md
thaJeztah: Yes, I think that's useful (at least for one of the examples)

mistyhacks:
@fcrisciani We have decided not to doc this officially. Please use the contents of this PR to develop a README for https://github.com/docker/libnetwork/. After you have done so, please ping me to link to the readme from the `dockerd` doc in `docker/cli` repo.

mistyhacks:
@fcrisciani Please also include cautions about potential to harm the network, in the upstream `README`.

PR #5572, opened 2017-12-19
Update journald logging driver options docs

william-richard:
### Proposed changes

I added more specific docs for the journald logging driver options, since I was confused about how they behaved.  I also fixed a minor typo in the gcp logging driver docs.

### Related issues (optional)

N/A

### DIFF
--- a/engine/admin/logging/journald.md
+++ b/engine/admin/logging/journald.md
@@ -50,20 +50,20 @@ $ docker run --log-driver=journald ...
 
 ## Options
 
-You can use the --log-opt NAME=VALUE flag to specify these additional Journald Logging driver options:
+Use the `--log-opt NAME=VALUE` flag to specify additional `journald` logging driver options.
 
 | Option | Required | Description |
 | -- | -- | -- |
-| `tag` | optional | Specify template to set CONTAINER_TAG value in journald logs. Refer to [log tag option documentation](https://docs.docker.com/engine/admin/logging/log_tags/) to customize the log tag format |
+| `tag` | optional | Specify template to set `CONTAINER_TAG` value in journald logs. Refer to [log tag option documentation](/engine/admin/logging/log_tags/) to customize the log tag format |
 | `label` | optional | Comma-separated list of keys of labels, which should be included in message, if these labels are specified for the container. |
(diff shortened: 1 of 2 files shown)


engine/admin/logging/journald.md
```diff
@@ -50,24 +50,29 @@ $ docker run --log-driver=journald ...
...
+    --log-opt env=TEST \
+    --env "TEST=false" \
+    --label location=west \
+    your/application
+```
 
-`env-regex` is similar to and compatible with `env`. Set it to a regular
-expression to match logging-related environment variables. It is used for
-advanced [log tag options](log_tags.md).
+This configuration also directs the driver to include in the payload the label location, and the environment variable ENV.  If the `--env "TEST=false"` or `--label location=west` arguments were omitted, the corresponding key would not be set in the journald log.
```
mistyhacks: The environment variable ENV? Don't you mean the environment variable `TEST` or `location`?

engine/admin/logging/journald.md
```diff
@@ -50,24 +50,29 @@ $ docker run --log-driver=journald ...
...
 
-Use the `--log-opt NAME=VALUE` flag to specify additional `journald` logging
-driver options.
+You can use the --log-opt NAME=VALUE flag to specify these additional Journald Logging driver options:
 
-### `tag`
+| Option | Required | Description |
+| -- | -- | -- |
+| `tag` | optional | Specify template to set CONTAINER_TAG value in journald logs. Refer to [log tag option documentation](https://docs.docker.com/engine/admin/logging/log_tags/) to customize the log tag format |
+| `label` | optional | Comma-separated list of keys of labels, which should be included in message, if these labels are specified for the container. |
```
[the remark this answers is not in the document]
william-richard: Heh yeah, this is a big part of what I found confusing too.  This is the same language used on the gcp docs, so once we figure out some nicer wording for it here, I'll change it there too.  

What do you think about this wording instead?
> Comma-separated list of label keys.  If a label key is present on a container, the key and value of that label will be included in any logs generated by that container.

engine/admin/logging/journald.md
william-richard: Yep, sorry, you're right.

PR #5574, opened 2017-12-20
WIP: Fix menu overflow

### LINKED ISSUE #5207: Version selector breaks at 1132px
[engine/swarm/networking.md](https://docs.docker.com/engine/swarm/networking/)

At this breakpoint the version selector overlays the Edit This Page link and makes it unclickable. There is another weird breakpoint up to 1153px wide, when it gets normal. cc/ @jsouth

### LINKED ISSUE #5511: docker enterprise edition menu item is not clickable
### Problem description
screenshot

### Problem location
every page

n/a

### Suggestions for a fix
Test on different screen sizes?

jsouth:
Fixes #5511
Fixes #5207 

However, I would like to fine tune the actions on viewport rescaling a bit more. 

More to come  :)


jsouth:
@johndmulhausen 

- update -
Still noticing issues in some browsers. Needs further refinement.

jsouth:
@mistyhacks - Thank you for the above comments/details!

Going to take another crack at this 🌟

mistyhacks:
1025px - 1141px is weird because the top bar is still small (not the hero one) but its background is transitioning like the hero one does. I think this is not what we want. @jsouth

mistyhacks:
@jsouth looks like you need to rebase on `msater`.

jsouth:
menu

jsouth:
@mistyhacks - The hamburger menu is utilizing the default bootstrap breakpoints. We can override, but need to double check to be sure it won't effect anything negatively. will circle back!

jsouth:
@mistyhacks - Would it be possible to merge what we have, and i"ll do some follow up fixes?
The current PR addresses the breakage for the version selector, and would like to get this out soon if possible. 

please advise :)

mistyhacks:
Sure, that sounds fine. I'm not sure why the hamburger menu is not behaving if it is just using the default Bootstrap breakpoints. I would prefer for us not to have to override the Bootstrap defaults in anyway. We can debug that 7-pixel gap in a follow-up!

PR #5599, opened 2017-12-21
Fixing some formatting and command params

artberger:
### Proposed changes

For D4IC docs: Corrected some formatting issues that were causing code blocks not to render. Adding some params to a couple commands. Also removed an FAQ that is no longer relevant.


artberger:
fyi @joaofnfernandes

artberger:
Thank you! ☃️ 🎉 

And thanks for the `raw` tip!

PR #5607, opened 2017-12-23
Improve Engine API documentation

thaJeztah:
This PR adds a couple of optimisations to the Engine API documentation;

### 1. Share ReDoc JavaScript between API versions

The redoc JavaScript is a big file, and there's no need to keep a separate copy for each version of the API. This patch removes the copy per-version and moves the script to a central location.

Saves +/ 12MB in size, and sharing the file improves browser caching (i.e. only having to load the file once)

### 2.  Use central stylesheet for API reference

Not a performance optimisation, but having a central CSS allows easier modification of styles for all API versions, instead of having to modify each API version's HTML separately.

### 3. Update ReDoc to v1.19.3

Several bug fixes and improvements; doesn't hurt to update :smile:

### 4.  Hide "request type" badges in menu

Related to the ReDoc update; ReDoc v1.13 added "request type" badges in the navigation menu. Although this information _could be_ useful, it's also a bit noisy, and we didn't show these before, so changing the CSS to hide them. Easy to add back if we think it's useful

With badges:

screen shot 2017-12-23 at 16 33 30

And without:

screen shot 2017-12-23 at 16 33 51

### 5. Add `.map` file for ReDoc

Having a .map file allows debugging script issues in the ReDoc JavaScript (which is minified).

It's a big file, but may be useful to have.

ping @mistyhacks @johndmulhausen


mistyhacks:
Works fine with a local build. I'm going to merge this because it cuts way down on the size of our image. 💃

mistyhacks:
Realized that it shouldn't be expected to work in Netlify anyway because it hasn't downloaded the Swagger files.

PR #5613, opened 2017-12-24
Document the `mode` and `max-buffer-size` log options

skuenzli:
Document how and why to configure the `mode` and `max-buffer-size` log options.

Documentation based on:
* implementation and comments of the [Ring Logger](https://github.com/moby/moby/blob/master/daemon/logger/ring.go#L49)
* conversation in implementing [PR #28762](https://github.com/moby/moby/pull/28762)

### Proposed changes

Document how and why to configure the `mode` and `max-buffer-size` log options because they are very useful in creating a robust container host configuration.

N/A - these options were released in Docker 17.04

### Related issues (optional)

Documents changes introduced in #28762

### DIFF
--- a/engine/admin/logging/overview.md
+++ b/engine/admin/logging/overview.md
@@ -108,7 +108,7 @@ The `max-buffer-size` log option controls the size of the ring buffer used for i
 The following example starts an Alpine container with log output in non-blocking mode and a 4 megabyte buffer:
 
 ```bash
-$ docker run --log-opt mode=non-blocking --log-opt max-buffer-size=4m alpine sh
+$ docker run -it --log-opt mode=non-blocking --log-opt max-buffer-size=4m alpine ping 127.0.0.1
 ```
 
 ### Use environment variables or labels with logging drivers
--- a/engine/admin/logging/overview.md
+++ b/engine/admin/logging/overview.md
@@ -89,6 +89,26 @@ json-file
 {% endraw %}
 ```
 
+## Configure the delivery mode of log messages from container to log driver
+
+Docker provides two modes for delivering messages from the container to the log driver since Docker 17.04:
+1. (default) direct, blocking delivery from container to driver
+2. non-blocking delivery that stores log messages in an intermediate per-container ring buffer
+
+The non-blocking mode of message delivery is useful for ensuring that applications will not block when the logging sub-system is unable to keep up with message throughput.
+
+*WARNING*: the oldest messages in memory *will be dropped* when the buffer is full, but this is often preferable to blocking the log-writing process of an application.  
+
+The `mode` log option controls whether to use the `blocking` (default) or `non-blocking` message delivery.
+
+When `mode` is set to `non-blocking`, the `max-buffer-size` log option may be used to control the size of the ring buffer used for intermediate message storage.  `max-buffer-size` defaults to 1 megabyte.
+
+The following example starts an Alpine container with log output in non-blocking mode and a 4 megabyte buffer:
+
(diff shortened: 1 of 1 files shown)


allingeek:
The example might be more useful if the command is something that perpetually generates log data. Maybe a `ping` command.

mistyhacks:
@dnephin PTAL

engine/admin/logging/overview.md
```diff
@@ -89,6 +89,26 @@ json-file
...
 
+## Configure the delivery mode of log messages from container to log driver
+
+Docker provides two modes for delivering messages from the container to the log driver since Docker 17.04:
+1. (default) direct, blocking delivery from container to driver
+2. non-blocking delivery that stores log messages in an intermediate per-container ring buffer
+
+The non-blocking mode of message delivery is useful for ensuring that applications will not block when the logging sub-system is unable to keep up with message throughput.
+
+*WARNING*: the oldest messages in memory *will be dropped* when the buffer is full, but this is often preferable to blocking the log-writing process of an application.
```
mistyhacks: This should be an admonition, which means leading each line with `>`. In addition, to get the warning style, add `{:.warning}` on a line by itself (with no leading `>`) before line 101.

engine/admin/logging/overview.md
```diff
@@ -89,6 +89,26 @@ json-file
 {% endraw %}
 ```
 
+## Configure the delivery mode of log messages from container to log driver
+
+Docker provides two modes for delivering messages from the container to the log driver since Docker 17.04:
+1. (default) direct, blocking delivery from container to driver
```
mistyhacks: Use a bullet list rather than a numbered list, as order does not matter here.

engine/admin/logging/overview.md
```diff
@@ -89,6 +89,26 @@ json-file
...
+2. non-blocking delivery that stores log messages in an intermediate per-container ring buffer
+
+The non-blocking mode of message delivery is useful for ensuring that applications will not block when the logging sub-system is unable to keep up with message throughput.
```
mistyhacks: Maybe "for preventing applications from blocking when..."

allingeek: I'd replace all of line 98 with: 

> This message delivery mode will prevent applications from blocking due to logging back pressure. Applications - especially network applications - will likely fail in unexpected ways when STDERR and STDOUT IO blocks.

skuenzli: Thank you for your feedback Misty and Jeff.

I adjusted the docs to use a more active voice and to try and match the suggested style.  I referred to test.md to try and get the warning admonition correct.

While preparing these updates, I found documentation of these features by cpuguy83 (author) in PR https://github.com/docker/docker.github.io/pull/1482

It looks like those docs were merged to the `vnext-engine` branch in 144f9ab and then were removed somewhere along the way.

Misty, if you would prefer to use/restore those docs, it's fine with me.  I just want the options documented :)

engine/admin/logging/overview.md
```diff
@@ -89,6 +89,28 @@ json-file
 {% endraw %}
 ```
 
+## Configure the delivery mode of log messages from container to log driver
+
+Docker provides two modes for delivering messages from the container to the log driver since Docker 17.04:
```
mistyhacks: Please remove mentions of specific versions that are already very old and unsupported. :)

skuenzli:
Great - thank you Misty.  I removed the reference to 17.04.  Hopefully everything is in order now.

PR #5616, opened 2017-12-26
/compose/install/#alternative-install-options

paulvi:
/compose/install/#alternative-install-options  link does not work, so add "tab above" words


paulvi:
I think link worked before. And may work.

Anyway, if I click and it does not navigate, I can read "above tab", and understand that this is on this page.

PR #5635, opened 2017-12-27
Update getting started guide with retry-loop and short explanation

### LINKED ISSUE #5631: Update Redis example to fix possible race condition
See https://github.com/docker/docker.github.io/pull/5602#issuecomment-354110949

@shin- you were nominated by @thaJeztah as someone who might be able to update this example.

shin-:
### Proposed changes

Added basic (unoptimized) retry loop to redis request in getting started guide, and a short insert explaining why it is useful.

### Related issues (optional)

Fixes #5631

### DIFF
--- a/compose/gettingstarted.md
+++ b/compose/gettingstarted.md
@@ -26,21 +26,50 @@ Define the application dependencies.
 
 2.  Create a file called `app.py` in your project directory and paste this in:
 
+        import time
+
+        import redis
         from flask import Flask
-        from redis import Redis
+
 
         app = Flask(__name__)
-        redis = Redis(host='redis', port=6379)
+        cache = redis.Redis(host='redis', port=6379)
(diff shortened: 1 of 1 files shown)


shin-:
There's a few things I don't mention for the sake of brevity:

- This implementation is naive. It doesn't do exponential backoffs, or [jitter](https://aws.amazon.com/blogs/architecture/exponential-backoff-and-jitter/), which one would want to look into for high-scale, production-ready applications.

- This implementation doesn't handle errors that aren't of the specific `redis.exceptions.ConnectionError` type. As a result, other exceptions, like timeouts, that could potentially be handled with a retry aren't.

I feel these would belong in a separate "resilience guide" or some such.

thaJeztah:
Thanks so much for fixing this so fast!

PR #5643, opened 2017-12-30
Update centos.md

huangsen365:
Configure the Docker service to start automatically at boot time.


mistyhacks:
This is already covered in the non-distro-specific topic https://docs.docker.com/engine/installation/linux/linux-postinstall/#configure-docker-to-start-on-boot which is linked to from the CentOS topic, so I am going to close this PR. Thanks!

PR #5648, opened 2018-01-02
Systemd examples keep required [Service] header like in the example above it

selckin:
Systemd examples keep required [Service] header like in the example above it


adamancini:
Section header is required in systemd unit override files.  LGTM

PR #5649, opened 2018-01-02
Specify 0.0.0.0 as alternative address

paulrougieux:
### Proposed changes
I explained that the app can also  be accessed at http://0.0.0.0/.

On my Debian machine,  `run -p 80:80 usename/get-started:part2` starts an app accessible at http://localhost/ but for some reason  `docker stack deploy -c docker-compose.yml lab` is not accessible at localhost. The browser returns an "Unable to connect" error.

The app is accessible at http://0.0.0.0/ in both cases, when started with `docker run` and  with `docker stack deploy`.

### DIFF
--- a/get-started/part3.md
+++ b/get-started/part3.md
@@ -186,6 +186,9 @@ the previous command (`docker container ls -q`).
 > address later in the tutorial. For now, the visitor counter isn't working
 > for the same reason; we haven't yet added a service to persist data.
 
+> Unable to connect to localhost?
+> 
+> Check also http://0.0.0.0/
 
 ## Scale the app
 


mistyhacks:
@gbarr01 PTAL as you are working on the Getting Started topics.

gbarr01:
@mistyhacks , I've already updated the doc as part of my get-started updates so I'm going to close this PR. 

@paulrougieux , thanks for the PR. For reference, this is not a Docker issue but something you can fix in your [hosts](https://stackoverflow.com/questions/20345132/127-0-0-1-is-accessible-working-but-localhost-not-accessible-not-working) file.

PR #5650, opened 2018-01-02
Update TOC for Docker Store

gbarr01:
- Update TOC with entries for new Docker Store docs (BYOL and Publisher FAQ).
- Rearrange TOC to mirror upcoming changes to Docker Store docs.

cc: @KickingTheTV


_data/toc.yaml
```diff
@@ -2964,13 +2964,21 @@ manuals:
...
-  - path: /docker-store/faq/
-    title: Consumer FAQs
+    title: About Docker Store
+  - sectiontitle: Docker Store FAQs
+    section:
+    - path: /docker-store/customer_faq/
+      title: Customer FAQs
+    - path: /docker-store/publisher_faq/
+      title: Publisher FAQs
+  - sectiontitle: Publisher
```
mistyhacks: I think this title should be "Publish content" or something like that. Or I suppose Publisher Guide maybe. @johndmulhausen for opinion

gbarr01: What do you think @KickingTheTV?

gbarr01: @mistyhacks , what about `For Publishers`? This section covers more than content (and there are plans to add more sub-sections).

gbarr01:
@mistyhacks , thoughts on why this is failing?

mistyhacks:
It is saying that the tag doesn't exist but it does --- https://cloud.docker.com/app/docs/repository/docker/docs/docker.github.io/tags/docs-builder maybe a race condition? You can do an amend and force-push in the PR branch and it will make Jenkins run again. The Jenkins tests are not show-stoppers for merging, though.

PR #5651, opened 2018-01-02
Move /engine/installation to /install

mistyhacks:
Make installation topics a top-level citizen

I also adjusted a few other things along the way.


thaJeztah:
Earlier discussion on this https://github.com/moby/moby/pull/19844#issuecomment-194906466

But at the time, I proposed using `<product>/install`

mistyhacks:
UCP and DTR already have their own install docs, as do the companion commands like `docker-compose`.

PR #5653, opened 2018-01-02
Update part4.md

alstas:
Add clarification about `docker stack deploy` command that it takes time to update all services

### Proposed changes

When I did this tutorial first time and did not know anything about Docker and its services, it took me a time to realize that when you do `docker stack deploy` it takes time to your services be updated especially when you have several of them and in a swarm. 

I followed the instructions in tutorial and when decided to modify my code and push changes to repository, so it turned out that I pushed latest updated image to docker repo and did `docker stack deploy` and went by URL to check my changes took effect, but there was nothing updated (I guess that some of the services were already updated, but I had a chance to be routed to not updated ones). Then I thought that I have to push with another tag and that docker do not update if see nothing has been changed in settings. So, I tried to push with a new tag, and I saw that no updates again were shown after `docker stack deploy`.  Thus, after some playing with it I found that you can check ps of services and check if they were restarted and updated and that Docker really do deploy changes even if the repo and the tag are the same. 

I hope it will give some clarification and saves time for new users that do Get Started tutorial and want to check deployment of changes.


johndmulhausen:
@alstas I think the warning is a little dire; how about "it may take a few seconds" (in my experience, that's plenty of time) rather than "takes time?"

alstas:
@johndmulhausen I agree with you, that sound better

PR #5661, opened 2018-01-03
Also modify JS files when normalizing links

mistyhacks:
Without this, the left-hand and top TOC don't work in the 17.09 archive, which generates those TOCs purely from JS.

### DIFF
--- a/scripts/normalize-links.sh
+++ b/scripts/normalize-links.sh
@@ -27,6 +27,7 @@ fi
 # - src=
 # - href=
 # followed by an optional double quote
+# The goal is to change URLs to all be absolute links starting at /
 
 printf "Cleaning up $VER"
 
--- a/scripts/create-permalinks.sh
+++ b/scripts/create-permalinks.sh
@@ -23,6 +23,14 @@ if ! [ -d "$TARGET" ]; then
 fi
 
 # Create permalinks for archived versions
+# first the HTML, then JS which is used to build the TOC in 17.09+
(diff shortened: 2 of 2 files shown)


scripts/normalize-links.sh
```diff
@@ -28,23 +28,25 @@ fi
 # - href=
 # followed by an optional double quote
 
+# We had to add the js checks for 17.09 and higher when we use JS for the TOC
+
 printf "Cleaning up $VER"
 
 # Fix relative links for archive
-printf "."; find ${TARGET} -type f -name '*.html' -print0 | xargs -0 sed -i 's#\(src\|href\)=\("\{0,1\}\)\(http\|https\)://\(docs\|docs-stage\).docker.com/#\1=\2/#g'
+printf "."; find ${TARGET} -type f \( -name '*.html' -or -name '*.js' \) -print0 | xargs -0 sed -i 's#\(src\|href\)=\("\{0,1\}\)\(http\|https\)://\(docs\|docs-stage\).docker.com/#\1=\2/#g'
```
thaJeztah: Hm don't think, this will work, because the paths are in JSON, and not `src=` / `href=`;

```
"path":"/engine/tutorials/networkingcontainers/"
```

mistyhacks: Ahh, good catch

johndmulhausen:
Because you're dealing with different source patterns and looking for different files, I'd just do a separate statement for `toc.js` rather than complicating the `*.html` pass. You're on the right path!

PR #5663, opened 2018-01-04
use force yes when install packages.

shockinglee:
use --force-yes incase dependencies in need.

### Proposed changes

add --force-yes when using apt to install openssh-server.

### Related issues (optional)

authentication required when dependencies missing like python.

### DIFF
--- a/engine/examples/running_ssh_service.md
+++ b/engine/examples/running_ssh_service.md
@@ -13,7 +13,7 @@ quick access to a test container.
 ```Dockerfile
 FROM ubuntu:16.04
 
-RUN apt-get update && apt-get install -y openssh-server
+RUN apt-get update && apt-get install -y openssh-server --force-yes
 RUN mkdir /var/run/sshd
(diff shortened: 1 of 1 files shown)


engine/examples/running_ssh_service.md
```diff
@@ -13,7 +13,7 @@ quick access to a test container.
 ```Dockerfile
 FROM ubuntu:16.04
 
-RUN apt-get update && apt-get install -y openssh-server
+RUN apt-get update && apt-get install -y openssh-server --force-yes
```
thaJeztah: Can you explain why this is needed? I tried this example, and it works fine without this option

mistyhacks: Yes, this is the same as the `-y` already provided.

shockinglee: Sorry that, I used the different base image, I copy the code, I missed some dependencies in that image, so I need to add `--force-yes` to confirm the installation of the dependencies.

PR #5664, opened 2018-01-04
Try again to fix JS-generated TOC in archives

mistyhacks:
This is not working yet. The toc.js looks right but it makes the left-hand nav not render at all.


mistyhacks:
To test this, you need to check out this PR branch and use the `README` to build all three of the base images, then apply the patch in #5666 to the 17.09 archive (it is raised against `master`, and build and deploy that using the `docker build` command. Sorry it's so round-about, but it's a problem that won't actually show up except in an archive where we are using the JS TOC.

PR #5669, opened 2018-01-05
Use Jenkins nodes that support multi-stage

mistyhacks:
All our Jenkins jobs are failing. This hopefully fixes that.

cc/ @seemethere


mistyhacks:
Done. I guess if this PR passes we know it works.

PR #5673, opened 2018-01-05
Release notes for 17.12.0-ce-mac45 (Edge)

gtardif:
cc @gbarr01, cc @mistyhacks

### DIFF
--- a/docker-for-mac/release-notes.md
+++ b/docker-for-mac/release-notes.md
@@ -469,6 +469,18 @@ events or unexpected unmounts.
 
 ## Edge Release Notes
 
+### Docker Community Edition 17.12.0-ce-mac45 2018-01-05 (Edge)
+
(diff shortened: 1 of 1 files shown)


gtardif:
OK, I didn't understand your concern at first. In Docker for Desktop when there is a Stable release we also do an Edge release with the same number, and the stable release has different release notes from the Edge one (including changes from previous Edge releases from since the last Stable). 
In this case we'll have separate release notes as well ; for Stable 17.12.0 it is here: https://github.com/docker/pinata/blob/stable-17.12.x/mac/CHANGELOG.
The stable release notes do not mention Kubernetes. Since the kube feature was presented as an edge-specific feature in docs and emails sent to users, this should be OK.
There will be a PR for the Stable release notes going along with the Stable release.

Issue #5685, opened 2018-01-08
Issue with polling processes in multi-service container script


bvwells:
### Problem description

The example script for running multiple services within a container has an issue. Sometimes the processes which have been launched from the entry point script can take a few seconds to start up and for the process tree to be updated to reflect the process name. Due to this, polling straight after the process has been launched can lead to the process not being found and the script exiting with a non-zero exit code.

### Problem location

- I saw a problem on the following URL: https://docs.docker.com/engine/admin/multi-service_container/

### Project version(s) affected

N/A - Purely documentation.

### Suggestions for a fix

Wait after processes are launched to poll to see whether they have exited.

bvwells:
Fixed in PR #5686.

L-Hudson:
Closing this ticket due to its age, and the impending refactor. If you think this is in error, feel free to reopen. Thanks!

PR #5686, opened 2018-01-08
Wait after processes launched to poll to see whether exited

bvwells:
### Proposed changes

This is to address issue #5685. Wait after processes are launched to poll to see whether they have exited.

N/A

### Related issues (optional)

N/A

### DIFF
--- a/engine/admin/multi-service_container.md
+++ b/engine/admin/multi-service_container.md
@@ -58,7 +58,7 @@ this in a few different ways.
   # if it detects that either of the processes has exited.
   # Otherwise it will loop forever, waking up every 60 seconds
 
-  while /bin/true; do
+  while sleep 60; do
     ps aux |grep my_first_process |grep -q -v grep
     PROCESS_1_STATUS=$?
     ps aux |grep my_second_process |grep -q -v grep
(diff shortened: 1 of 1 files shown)


bvwells:
Here's a nice article which describes what happens when using ps to monitor sub-processes which have been forked from a shell script.

https://aweirdimagination.net/2015/02/06/child-process-not-in-ps/

mistyhacks:
I think this will only run once, rather than continuing to run, and thus changes the whole intent of the script. Am I missing something? By the way, as stated this is meant to be an extremely naive script and is just shown for illustration purposes. This is not a recommended approach at all.

bvwells:
Hi @mistyhacks. Thanks for the review. Sadly I think you are missing something :-). Give it a go, it's a trick a friend showed me. The script will still execute infinitely, but will sleep before each execution of the loop body. Try the simple script:

```
#!/bin/bash

while sleep 1; do
    echo 'hello'
done
```

I completely agree that this is not a recommended approach, but I still feel that if you are going to show a script it should be bug free. The other alternative is to completely remove the script.

mistyhacks:
Cool! I like it. I played around with it. Thanks for showing that to me.

bvwells:
No problems! I have to say I thought the same thing when I saw this. :-)
Many thanks for the approval!

PR #5689, opened 2018-01-08
Fix invisibility of "Edit this page" link

### LINKED ISSUE #5688: "Edit this page" link hidden by default
### Problem description

As stated. See https://github.com/docker/docker.github.io/issues/5670#issuecomment-355770943 .

### Problem location

- I saw a problem on the following URL: https://docs.docker.com/get-started/

sampablokuper:
### Proposed changes

If there is a good reason for this element to be hidden, I am not aware of it.

### Related issues (optional)

Fixes #5688.

### DIFF
--- a/_layouts/docs.html
+++ b/_layouts/docs.html
@@ -247,7 +247,7 @@
 								<div class="feedback-links">
 									<ul>
 										{% if edit_url != "" %}
-										<li style="visibility: hidden"><a href="{{ edit_url }}"><i class="fa fa-pencil-square-o" aria-hidden="true"></i> Edit this page</a></li>{% endif %}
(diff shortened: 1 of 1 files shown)


johndmulhausen:
It's invisible by default, and made visible by JavaScript if you are viewing the page in a non-archive URL (i.e. from `/` and not from `/v17.09/`)

johndmulhausen:
cc @mistyhacks

sampablokuper:
@johndmulhausen wrote:
> It's invisible by default, and made visible by JavaScript if you are viewing the page in a non-archive URL (i.e. from `/` and not from `/v17.09/`)

Why not just make it visible by default, and have it link to the revision of the source file from which the web page being viewed was rendered? Doing it as I've just described would:

- remove the wrong and dangerous expectation that visitors use Javascript;
- facilitate patches against all branches of the docs, not just the current branch, reducing the total number of errors in the docs while still permitting the patches to be propagated to more recent branches.

PR #5705, opened 2018-01-09
update option values

fretscha:
- only lowercase availability zones are accepted in CLI options 
- only image short-names are accepted in CLI options

### Proposed changes
Updated documentation. Changes default values and examples to values which work in real life.


greut:
What happens when the documentation PR get merged way earlier than the docker-machine ones...

PR #5728, opened 2018-01-12
Remove gradientswitch

sampablokuper:
### Proposed changes

This PR:

- solves an accessibility issue (moving/changing backgrounds ... ouch), and

- solves a client browser resource usage issue (on some web browsers, CSS transitions - especially continuously cycling ones - disproportionately increase CPU and/or memory utilisation).

### DIFF
--- a/_scss/_layout.scss
+++ b/_scss/_layout.scss
@@ -22,61 +22,13 @@
 */
 
 .header {
-    animation: gradientswitch 80s infinite;
-    -webkit-animation: gradientswitch 80s infinite;
+    background: #0087C9;
     height: $global-header-height;
     position: relative;
     z-index: 1;
     transition: all 0.1s ease;
 }
 
-@keyframes gradientswitch {
-    0% {
-        background: #0087C9;
-    }
-    16% {
-        background: #EF4A53;
-    }
-    32% {
-        background: #FFB463;
-    }
-    49% {
-        background: #33D5D4;
-    }
-    65% {
-        background: #254356;
-    }
-    81% {
-        background: #5B4097
-    }
-    100% {
-        background: #0087C9
-    }
-}
-
-@-webkit-keyframes gradientswitch {
-    0% {
-        background: #0087C9;
-    }
-    16% {
-        background: #EF4A53;
-    }
-    32% {
-        background: #FFB463;
-    }
-    49% {
-        background: #33D5D4;
-    }
-    65% {
-        background: #254356;
-    }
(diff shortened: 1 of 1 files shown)


mistyhacks:
I think this work is already being done in #5574 . Maybe you can work with @jsouth there to make sure your concerns are addressed there. Thanks!

sampablokuper:
@mistyhacks and @jsouth, #5574 is titled "Fix menu overflow". The gradientswitch my PR removes has nothing to do with menu overflows, AFAICT. If the gradientswitch is being addressed in #5574, then #5574 is not a granular PR and should be broken up into a set of PRs that address one issue each.

My PR above *is* a granular PR, and addresses just one issue: it removes the gradientswitch. You can think of it as one member of the set of PRs mentioned above.

Please reconsider your decision not to merge it. Thank you.

mistyhacks:
Look, I understand, but I happen to know that @jsouth is removing the gradient as part of his work in that PR, because I'm working with him and already asked him to do it (he leads the Docker web design team). We appreciate your help but sometimes (through no fault of your own) you may not have all of the info. I'm sorry we seem to be saying no to you a lot. It's not our intention. I am sure that @jsouth would appreciate you working with him in his PR and even committing into it.

jsouth:
@sampablokuper - thanks for working on this, and apologies for the slight mix-up.

As @mistyhacks mentioned, I have removed this in #5574.

Happy to collaborate with @sampablokuper anytime on future issues.

sampablokuper:
@jsouth, thanks for your positive remark :)

@mistyhacks, thanks for the offer to commit to #5574, but as I mentioned, IMO that was not a suitable place to address this issue. As such, I had no desire to address it there. (And besides, jsouth had already addressed it there, so I'm not sure what I would have committed anyhow.) As to your point about not having the inside info: I realise that "drive-by contributors" like me don't have that info, but I would counter that if all PRs, bug reports, and commits, are made to be suitably granular and well-described, then there needn't *be* any insider info (except maybe about a handful of super-sensitive matters, such as responsibly-disclosed security issues that are under embargo). This should result in less overhead for your team, or at least, a more productive kind of overhead: policing PRs/issues/etc for granularity instead of having to write comments as a workaround for the lack of granularity. This in turn might mean more time available to focus on making Docker's docs even better than they already are as a learning tool, and maybe also to make them more respecting of visitors' desires for security ;)

Both of you: thanks for fixing the issue, anyhow :)

PR #5735, opened 2018-01-13
Document 'start_period' option in healthcheck

goddva:
### Proposed changes

Updated documentation with the option 'start_period' under 'healthcheck', as it was missing from the documentation.


mistyhacks:
@shin- PTAL

PR #5741, opened 2018-01-13
fix alpine command from ash to sh

smueller18:
### Proposed changes

comand to print network information from inside the alpine container was wrong


goddva:
@smueller18 the command is not wrong, Ash is Almquist Shell. If your command failed, I guess its because you dont have IPv6.

See if this works for you: 
`docker run -it alpine ash -c "ip -4 addr show dev eth0; ip -4 route show"`

smueller18:
sorry for that, checked it on an ipv6 ready host and it worked

PR #5712, opened 2018-01-09
Fix broken links to installation instructions

sampablokuper:
### Proposed changes

Replace `.md` with `/`.

### Related issues (optional)

Fixes #5711 .

PR #5742, opened 2018-01-14
Remove .md extensions that break links

sampablokuper:
### Proposed changes

The navigation links at the start of the content on pages such as https://docs.docker.com/get-started/part2/ have broken links (at least, with JavaScript disabled). The links go to e.g. https://docs.docker.com/get-started/part2.md but should instead go to https://docs.docker.com/get-started/part2 or https://docs.docker.com/get-started/part2/ .

### Related issues (optional)

Similar to #5712.

### DIFF
--- a/get-started/nav.html
+++ b/get-started/nav.html
@@ -1,8 +1,8 @@
 <ul class="pagination">
-  <li {% if include.selected=="1"%}class="active"{% endif %}><a href="index.md">1: Orientation</a></li>
-  <li {% if include.selected=="2"%}class="active"{% endif %}><a href="part2.md">2: Containers</a></li>
-  <li {% if include.selected=="3"%}class="active"{% endif %}><a href="part3.md">3: Services</a></li>
-  <li {% if include.selected=="4"%}class="active"{% endif %}><a href="part4.md">4: Swarms</a></li>
-  <li {% if include.selected=="5"%}class="active"{% endif %}><a href="part5.md">5: Stacks</a></li>
-  <li {% if include.selected=="6"%}class="active"{% endif %}><a href="part6.md">6: Deploy your app</a></li>
+  <li {% if include.selected=="1"%}class="active"{% endif %}><a href="index">1: Orientation</a></li>
+  <li {% if include.selected=="2"%}class="active"{% endif %}><a href="part2">2: Containers</a></li>
+  <li {% if include.selected=="3"%}class="active"{% endif %}><a href="part3">3: Services</a></li>
+  <li {% if include.selected=="4"%}class="active"{% endif %}><a href="part4">4: Swarms</a></li>
+  <li {% if include.selected=="5"%}class="active"{% endif %}><a href="part5">5: Stacks</a></li>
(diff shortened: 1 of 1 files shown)


mistyhacks:
I don't know if we will be able to accept this because lots and lots of aspects of the site require JS. @johndmulhausen PTAL

sampablokuper:
@mistyhacks [wrote](https://github.com/docker/docker.github.io/pull/5742#issuecomment-358089551):
> I don't know if we will be able to accept this because lots and lots of aspects of the site require JS.

Let's consider this logically.

Unless your JS is doing something silly (or is [as cunning as a fox who's just been appointed Professor of Cunning at Oxford University](http://www.imdb.com/title/tt0096548/quotes)), then this PR shouldn't break it, so the PR should be accepted.

If your JS *is* doing something silly, then the JS should be fixed (but that's a separate issue), and this PR should be accepted.

mistyhacks:
The reason we use links to the `.md` files and have our JS transform them is to preserve the ability for people to be able to browse our Markdown source files directly on Github and for cross-file linking to work there, with Github's auto-rendering of Markdown.

sampablokuper:
@mistyhacks [wrote](https://github.com/docker/docker.github.io/pull/5742#issuecomment-358723588):

> The reason we use links to the `.md` files and have our JS transform them is to preserve the ability for people to be able to browse our Markdown source files directly on Github and for cross-file linking to work there, with Github's auto-rendering of Markdown.

Your stated rationale does not apply to my PR above, because:

- My PR above only modifies an HTML *include* file (i.e. an HTML fragment), not a Markdown document that is rendered to a browsable page by GitHub's renderer. See this for yourself [here](https://github.com/docker/docker.github.io/blob/master/get-started/nav.html) if you don't believe me.

- Nor would my PR above affect the rendering, on GitHub, of the Markdown documents that include that file, because GitHub's Markdown renderer is not Liquid-aware and so does not fetch or render such includes. Again, if you don't believe me, see e.g. these two pages where the Liquid template calls the *include* affected by my PR above: https://github.com/docker/docker.github.io/blob/master/get-started/index.md and https://github.com/docker/docker.github.io/blob/master/get-started/part2.md .

mistyhacks:
I do totally believe you for this PR. I was just giving you context in general, since you had expressed extreme skepticism about this particular chunk of JS we use and why we are unwilling to stop using the `.md` links in general in our docs.

sampablokuper:
@mistyhacks [wrote](https://github.com/docker/docker.github.io/pull/5742#issuecomment-359061029):
> I do totally believe you for this PR. I was just giving you context in general, since you had expressed extreme skepticism about this particular chunk of JS we use and why we are unwilling to stop using the `.md` links in general in our docs.

I remain sceptical about the justification for JS on Docker's documentation website (and indeed on most websites that use JS), have provided sound reasons for my concerns, and would still be grateful if Docker would reconsider its stance on the matter.

But please can my PRs at least be addressed on their own merits?

Anyway, thanks for merging this :)

PR #5755, opened 2018-01-16
Update compose reference for name fields

dnephin:
Ref: https://github.com/docker/cli/issues/811

Attempt to clarify the behaviour of these fields, and why you might want to use them. Also document that `external.name` is deprecated.

cc @shin- I'm not sure if they are also deprecated in 2.x

cc @mistyhacks


shin-:
>  I'm not sure if they are also deprecated in 2.x

Not explicitly at the moment, but it might be worth doing.

PR #5771, opened 2018-01-18
spacing + config.json -> daemon.json

idi-ot:
### Proposed changes
Again, not sure if that's actually a typo, but in the previous guides that I've read, the config file was always referred to as `daemon.json` and not `config/json`


mistyhacks:
Yeah that is a big typo! There is also a `~/.docker/config.json` that configures the Docker client, but this one is about the daemon.

PR #5780, opened 2018-01-19
add spacing before grep command

idi-ot:
add spacing before grep command


mistyhacks:
Cosmetic, but sure why not?

PR #5788, opened 2018-01-19
Various copyedits to reduce future tense, wordiness, and use of 'please'

mistyhacks:
Where possible, simplify language to use simple present tense for clarity. I need to do a careful review before merging this as it is a lot of changes.


_includes/install-script.md
```diff
@@ -48,9 +48,9 @@ adding your user to the "docker" group with something like:
 
   sudo usermod -aG docker your-user
 
-Remember that you will have to log out and back in for this to take effect!
+Remember to log out and back in for this to take effect!
```
thaJeztah: This should also be updated in the actual install script

mistyhacks:
Let's do this thing.

PR #5795, opened 2018-01-22
link to blog post about wait service for use in Compose

afeld:
### Proposed changes

I stumbled upon this blog post when trying to find examples of `wait-for-it.sh` - it's an interesting Compose-specific approach that hadn't occurred to me before, which doesn't require modification to the other containers or start commands. Figured it might be interesting to others.


mistyhacks:
Please rebase to resolve conflicts.

@shin- PTAL

shin-:
:-1: 

I'm glad people find this helpful, but the official doc shouldn't link to a medium post presenting a solution that relies on some random, unverified Docker image.

PR #5801, opened 2018-01-22
Update how to get support

### LINKED ISSUE #5618: Recommended changes on guidance to obtain technical support
[datacenter/ucp/2.2/guides/get-support.md](https://docs.docker.com/datacenter/ucp/2.2/guides/get-support/), CC @JimGalasyn 

in this set of instructions

> If you need help, you can file a ticket via:
> 
> Email
> Docker support page
> Be sure to use your company email when filing tickets.
> 
> 

I recommend some changes.

Tech support is trying to deprecate email support, since such cases come in with no entitlement, nor real knowledge of the customer, which can delay our response to the customer by an arbitrary amount of time. We also strongly prefer paid customers (i.e. EE customers) use the support portal, or, barring that, the Account Support Form.

I recommend changing that paragraph to something like

> If you need help, and you are an Authorized Support Contact, please submit a new case via the [Docker support page](https://support.docker.com/). If you are not an Authorized Support Contact, please notify a member of your organization who is one. If you are an Authorized Support Contact, and are unable to submit a new case, please use our [Account Support Form](https://success.docker.com/support), using your company email.

joaofnfernandes:
Fixes #5618

### DIFF
--- a/datacenter/dtr/2.4/guides/support.md
+++ b/datacenter/dtr/2.4/guides/support.md
@@ -5,12 +5,15 @@ description: Your Docker subscription gives you access to prioritized support. Y
 keywords: support, help
 ---
 
-Your Docker Data Center, or Docker Trusted Registry subscription gives you
-access to prioritized support. The service levels depend on your subscription.
+Your Docker Enterprise Edition subscription gives you access to prioritized
+support. The service levels depend on your subscription.
 
-If you need help, you can file a ticket via:
+Before reaching out to Docker Support, make sure you're listed as an authorized
+support contact for your account. If you're not listed as an authorized
+support contact, find a person that is, and ask them to open a case with
+Docker Support in your behalf.
 
-* [Email](mailto:support@docker.com)
-* [Docker support page](https://support.docker.com/)
-
-Be sure to use your company email when filing tickets.
+You can open a new support case in the [Docker support page](https://support.docker.com/).
+If you're unable to submit a new case in the support page, fill-in the
+[Docker account support form](https://success.docker.com/support) using your
+company email.
--- a/datacenter/dtr/2.4/guides/support.md
+++ b/datacenter/dtr/2.4/guides/support.md
@@ -10,10 +10,10 @@ support. The service levels depend on your subscription.
 
 Before reaching out to Docker Support, make sure you're listed as an authorized
 support contact for your account. If you're not listed as an authorized
-support contact, find a person that is, and ask them to open a case with
+support contact, find a person who is, and ask them to open a case with
 Docker Support in your behalf.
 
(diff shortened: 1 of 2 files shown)


datacenter/dtr/2.4/guides/support.md
```diff
@@ -5,12 +5,15 @@ description: Your Docker subscription gives you access to prioritized support. Y
...
 
-Your Docker Data Center, or Docker Trusted Registry subscription gives you
-access to prioritized support. The service levels depend on your subscription.
+Your Docker Enterprise Edition subscription gives you access to prioritized
+support. The service levels depend on your subscription.
 
-If you need help, you can file a ticket via:
+Before reaching out to Docker Support, make sure you're listed as an authorized
+support contact for your account. If you're not listed as an authorized
+support contact, find a person that is, and ask them to open a case with
```
mistyhacks: s/that/who

People always get "who" and things always get "that".

Can you give them a quick link to the place where they can see if they are an authorized contact?

datacenter/dtr/2.4/guides/support.md
```diff
@@ -5,12 +5,15 @@ description: Your Docker subscription gives you access to prioritized support. Y
...
+Docker Support in your behalf.
 
-* [Email](mailto:support@docker.com)
-* [Docker support page](https://support.docker.com/)
-
-Be sure to use your company email when filing tickets.
+You can open a new support case in the [Docker support page](https://support.docker.com/).
```
mistyhacks: s/in/from (or at?)

datacenter/dtr/2.4/guides/support.md
```diff
@@ -5,12 +5,15 @@ description: Your Docker subscription gives you access to prioritized support. Y
...
+If you're unable to submit a new case in the support page, fill-in the
```
mistyhacks: "fill in" is two separate words. Why wouldn't they be able to use the support page?

s/email/email address. But this seems like a weird thing to say. Maybe "using an email address associated with the subscription"?

datacenter/dtr/2.4/guides/support.md
joaofnfernandes: I don't think we have something like that. From a user standpoint is not great that users have to learn how to navigate their internal orgs to get to the right person.
But from a security standpoint it's actually makes sense no to provide such tools since an attacker could gain information that otherwise they would not have access to.

datacenter/dtr/2.4/guides/support.md
joaofnfernandes: This was suggested by the support team. This provides an alternative way for users to open a case if the support page is down, or the designated point of contact cannot be found.

PR #5817, opened 2018-01-24
Clarify overlay2 support in Docker EE RHEL / CentOS

clemenko:
Removing the confusion around rhel 7.4 and overlay2.

### Proposed changes

I updated the language around the use of overlay to reflect the change in the support matrix. This was a complaint from a customer this morning.

### DIFF
--- a/install/linux/docker-ee/centos.md
+++ b/install/linux/docker-ee/centos.md
@@ -28,17 +28,27 @@ Docker CE users should go to
 
 ### OS requirements
 
-To install Docker EE, you need a maintained version of CentOS 7. Archived
-versions aren't supported or tested.
+To install Docker EE, you need the 64-bit version of {{ linux-dist-long }}
+running on `x86_64`.
 
-The `centos-extras` repository must be enabled. This repository is enabled by
-default, but if you have disabled it, you need to
-[re-enable it](https://wiki.centos.org/AdditionalResources/Repositories){: target="_blank" class="_" }.
+In addition, you must use the `overlay2` or `devicemapper` storage driver.
+Beginning with Docker EE 17.06.2-ee-5 the `overlay2` storage driver is the
+recommended storage driver.
 
-In addition, you must use the `overlay2` or `devicemapper` storage driver if you
-use Docker EE. On production systems using `devicemapper`, you must use
-`direct-lvm` mode, which requires one or more dedicated block devices. Fast
-storage such as solid-state media (SSD) is recommended.
+The following limitations apply:
+
+**OverlayFS**:
+
+- The `overlay2` storage driver is only supported on CentOS 7 systems
+  using version 3.10.0-693 or high of the kernel.
+- If `selinux` is enabled, the `overlay2` storage driver is only supported on
+  RHEL 7.4 or higher.
+
+**Devicemapper**:
+
+- On production systems using `devicemapper`, you must use `direct-lvm` mode,
+  which requires one or more dedicated block devices. Fast storage such as
+  solid-state media (SSD) is recommended.
 
 ### Uninstall old versions
 
--- a/install/linux/docker-ee/rhel.md
+++ b/install/linux/docker-ee/rhel.md
@@ -39,7 +39,8 @@ The following limitations apply:
 
 **OverlayFS**:
 
-- The `overlay2` storage driver is only supported on RHEL 7.2 or higher.
+- The `overlay2` storage driver is only supported on RHEL 7.2 or higher systems
+  using version 3.10.0-693 or high of the kernel.
 - If `selinux` is enabled, the `overlay2` storage driver is only supported on
   RHEL 7.4 or higher.
 
--- a/storage/storagedriver/overlayfs-driver.md
+++ b/storage/storagedriver/overlayfs-driver.md
@@ -15,7 +15,9 @@ storage driver as `overlay` or `overlay2`.
 
 > **Note**: If you use OverlayFS, use the `overlay2` driver rather than the
 > `overlay` driver, because it is more efficient in terms of inode utilization.
-> To use the new driver, you need version 4.0 or higher of the Linux kernel.
+> To use the new driver, you need version 4.0 or higher of the Linux kernel,
+> unless you are a Docker EE user on RHEL or CentOS, in which case you need
+> version 3.10.0-693 or higher of the kernel and to follow some extra steps.
 >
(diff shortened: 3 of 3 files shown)


engine/userguide/storagedriver/overlayfs-driver.md
```diff
@@ -14,6 +14,7 @@ storage driver as `overlay` or `overlay2`.
 > **Note**: If you use OverlayFS, use the `overlay2` driver rather than the
 > `overlay` driver, because it is more efficient in terms of inode utilization.
 > To use the new driver, you need version 4.0 or higher of the Linux kernel.
+> Or RHEL/Centos 7.4 or higher.
```
mistyhacks: This should be a comma, not a complete new sentence. What version of the kernel does 7.4 ship with? Is it lower than 4.0? If not, this extra clause seems unnecessary.

engine/userguide/storagedriver/overlayfs-driver.md
```diff
@@ -27,7 +28,7 @@ OverlayFS is supported if you meet the following prerequisites:
   
 - The `overlay` driver is allowed but not recommended for Docker CE.
 
-- Version 4.0 or higher of the Linux kernel. If you use an older kernel, you
+- Version 4.0 or higher of the Linux kernel, or RHEL/Centos 3.10.0-693 or higher. If you use an older kernel, you
```
mistyhacks: Is that a kernel version or a RHEL / Centos version? I'm guessing the former, so maybe clarify. I think the acronym is RHCK (Red Hat compatible kernel).

engine/userguide/storagedriver/overlayfs-driver.md
```diff
@@ -74,19 +75,7 @@ Before following this procedure, you must first meet all the
...
-      "storage-driver": "overlay2"
-    }
-    ```
-
-    If you need to use the legacy `overlay` driver, specify it instead.
-
-    To use `overlay2` on CentOS (Docker CE only), you must also set the storage
-    option `overlay2.override_kernel_check`. In this case the `daemon.json`
-    would look like this:
+    that the file was empty, add the following contents for `overlay2`.
```
mistyhacks: Pretty sure you'd only want the override thing if you are running a kernel prior to 4.0, so it's probably dangerous to recommend users to have it all the time. @cpuguy83 PTAL

engine/userguide/storagedriver/overlayfs-driver.md
```diff
@@ -97,6 +86,8 @@ Before following this procedure, you must first meet all the
     }
     ```
 
+    If you need to use the legacy `overlay` driver, specify it instead.
```
mistyhacks: If you use the `overlay` driver and you specify that kernel check override, what happens?

engine/userguide/storagedriver/overlayfs-driver.md
clemenko: Rhel/Centos 7.4 ships with the correct kernel that supports overlay2 of 3.10.0-693.

clemenko: The comma makes sense. The Rhel compatible kernel is 3.10.0-693.

engine/userguide/storagedriver/overlayfs-driver.md
clemenko: 3.10.0-693 is the correct kernel for Rhel/Centos 7.4.

engine/userguide/storagedriver/overlayfs-driver.md
clemenko: The kernel check override is ONLY for overlay2.

engine/userguide/storagedriver/overlayfs-driver.md
thaJeztah: The kernel-version-check option is no longer needed in current docker versions (at least 17.12, and the upcoming EE release; https://github.com/moby/moby/pull/35527)

clemenko: but it IS required today...

thaJeztah: With the improved detection, this means that overlay2 is selected by default on;

- Ubuntu, Debian running on kernel 4.x
- CentOS 7
- RHEL 7 (depending on kernel version and backing filesystem)


Running on XFS without `d_type` support is no longer supported, and will now cause docker to _skip_ overlay; existing installs will still run (but still produce an error) to allow users to migrate their data and fix the issue; that will change in a future version, when it will be a fatal error, and docker will refuse to start

clemenko: Great. we can change this doc when it is released...

mistyhacks: Let's not guide ALL users in overriding the kernel checks when ONLY the users using RHEL on an old kernel need it.

thaJeztah: @clemenko but it's been released ... in december

mistyhacks: Correct, these docs are for the latest stable Engine, which is CE 17.12.

mistyhacks:
I committed into your PR what I consider the appropriate way to document this. PTAL @clemenko and let me know if it's OK to merge. @thaJeztah PTAL too.

engine/userguide/storagedriver/overlayfs-driver.md
```diff
@@ -82,22 +94,26 @@ Before following this procedure, you must first meet all the
     }
     ```
 
-    If you need to use the legacy `overlay` driver, specify it instead.
-
-    To use `overlay2` on CentOS (Docker CE only), you must also set the storage
-    option `overlay2.override_kernel_check`. In this case the `daemon.json`
-    would look like this:
+    > **Note**: RHEL and CentOS users on Docker EE 17.06
```
thaJeztah: Erm, sorry, just had a brain wave, but for the current Docker EE 17.06.2 release we don't support overlay2 (it's not in the the support matrix); the upcoming release has the kernel support check added, and will add overlay2 to the support matrix.

So, I don't think we should mention this, because this describes how to configure the daemon to use a graph driver that's not yet in the support matrix

clemenko:
@mistyhacks Now that your team has approved supporting of Overlay2 can we push this change through ASAP?

PR #5818, opened 2018-01-25
Mention Docker package repositories contain docker-compose

breadcentric:
### Proposed changes

Docker Compose can be installed using the package repositories mentioned on the Docker install page.

--

### Related issues (optional)

--

### DIFF
--- a/compose/install.md
+++ b/compose/install.md
@@ -97,7 +97,7 @@ Docker Compose. To do so, follow these steps:
 <div id="linux" class="tab-pane fade" markdown="1">
 ### Install Compose on Linux systems
 
-*[The package repository set up for Docker instalation](https://docs.docker.com/engine/installation/linux/docker-ce/ubuntu/#upgrade-docker-ce) contains a docker-compose package. If you installed Docker this way, simply run * `sudo apt-get install docker-compose`.
+*[The package repository set up for Docker instalation](https://docs.docker.com/engine/installation/linux/docker-ce/ubuntu/#upgrade-docker-ce) contains a docker-compose package. If you installed Docker this way, simply run* `sudo apt-get install docker-compose`.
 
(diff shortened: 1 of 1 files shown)


compose/install.md
```diff
@@ -97,6 +97,8 @@ Docker Compose. To do so, follow these steps:
 <div id="linux" class="tab-pane fade" markdown="1">
 ### Install Compose on Linux systems
 
+*[The package repository set up for Docker instalation](https://docs.docker.com/engine/installation/linux/docker-ce/ubuntu/#upgrade-docker-ce) contains a docker-compose package. If you installed Docker this way, simply run* `sudo apt-get install docker-compose`.
```
mistyhacks: nit typo: "instalation"

@shin- PTAL and make sure this is accurate -- do we have `docker-compose` in our official CE repos?

shin-: To my knowledge, we don't - and the version in the default distro packages tend to be severely outdated. @andrewhsu can confirm.

mistyhacks:
Not recommending that users use unsanctioned third-party packages of `docker-compose`.

breadcentric:
@mistyhacks thank you for your comment. I was wondering if it would make sense to note this in the documentation. The packages are in a repository which is a standard documented way to install Docker and I have already talked to people who assumed it was just undocumented for Docker Compose.
Regardless of this, thanks for clarifying.

mistyhacks:
But it's not there. For instance, here is Docker's Ubuntu repository for Xenial:

https://download.docker.com/linux/ubuntu/dists/xenial/pool/stable/amd64/

That's the only official way to install Docker CE on Ubuntu, and as you can see, there are no `docker-compose` packages there. If there are `docker-compose` packages in Ubuntu's OS repositories, then Ubuntu's packagers are doing it themselves (or someone is) and not coordinating with Docker on it.

PR #5819, opened 2018-01-25
Document Kubernetes support for Win

mistyhacks:
For 18.2 RC1.


_includes/kubernetes-mac-win.md
```diff
@@ -0,0 +1,176 @@
...
+  {% assign kubectl-path = "/usr/local/bin/kubectl" %}
+
+{% elsif platform == "windows" %}
+  {% assign product = "Docker for Windows" %}
+
+  {% capture min-version %}{{ product }} 18.02 CE Edge{% endcapture %}
+
+  {% capture version-caveat %}
+**Kubernetes is only available in Docker for Windows 18.02 CE Edge users who
+participants in the [Docker Beta program](https://beta.docker.com/). To access
```
gtardif: "Kubernetes is only available in Docker for Windows 18.02 CE Edge. ", we don't do a private beta for the windows kube integration (since we already got feedback from the mac users. No need to log in.

docker-for-mac/index.md
```diff
@@ -302,6 +302,9 @@ support is not included in Docker for Mac 17.12 CE Stable. To find out
 more about Stable and Edge channels and how to switch between them, see
 [General configuration](/docker-for-mac/#general).
 
+The addition of Kubernetes support increases the total binary size by about
```
gtardif: I misundestood your question on slack ; 300 MB is actually the size of the download when you activate Kubernetes ; the size of the D4Mac / Win install is increased by about 100 MB. (Not sure we need to mention this here).

docker-for-windows/index.md
```diff
@@ -707,6 +707,63 @@ Check out these [Docker Cloud topics](/docker-cloud/index.md) to learn more:
 Need a direct link to Cloud? [Take me to Docker
 Cloud](https://cloud.docker.com/){: target="_blank" class="_" }.
 
+### Kubernetes
+
+**Kubernetes is only available in Docker for Windows 18.02 CE Edge users who
```
gtardif: Same as above, this is available for Edge users, no beta program & no need to log in.

docker-for-windows/index.md
```diff
@@ -707,6 +707,63 @@ Check out these [Docker Cloud topics](/docker-cloud/index.md) to learn more:
...
+participants in the [Docker Beta program](https://beta.docker.com/). To access
+beta builds, you must be signed in with your Docker ID within Docker for Windows.
+Select ![whale menu](/docker-for-windows/images/whale-x.png){: .inline} -> Sign in / Create Docker ID
+from the menu bar.**
+
+The addition of Kubernetes support increases the total binary size by about
```
gtardif: Same as for mac : 300 MB is the size of the download when enabling Kubernetes ; maybe to mention below next to "an internet connection is required"

PR #5820, opened 2018-01-25
Remove/destroy DTR replica

zaggash:
I think the actual documentation can be confusing.
In fact if we want to completely remove a DTR cluster, we can use the destroy command, it is fine.

But, if we just want to remove a replica from a cluster, it is better to use the remove command which is safer and cleaner.
The destroy command do not send a notification to the RethinkDB on the others replicas which can be dangerous and leads to unhealthy DTR clusters.

### DIFF
--- a/datacenter/dtr/2.4/guides/admin/install/uninstall.md
+++ b/datacenter/dtr/2.4/guides/admin/install/uninstall.md
@@ -5,7 +5,13 @@ keywords: dtr, install, uninstall
 ---
 
 Uninstalling DTR can be done by simply removing all data associated with each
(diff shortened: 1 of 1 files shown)


zaggash:
My modifications might need rewording.
Please correct me :)

joaofnfernandes:
Thanks for the contribution @zaggash! It seems that we have information on how to remove replicas in a [separate article ](https://docs.docker.com/datacenter/dtr/2.4/guides/admin/configure/set-up-high-availability/#remove-existing-replicas), but we never linked the two together.

If you feel like doing that, you can open a separate PR 🙌

PR #5848, opened 2018-01-27
google cloud auth command changed

sabarirangan:


mistyhacks:
It didn't change. Your command is actually a different command (https://cloud.google.com/sdk/gcloud/reference/auth/login vs https://cloud.google.com/sdk/gcloud/reference/auth/application-default/login). Do you have any info on why the second one would be preferred?

mistyhacks:
Lack of response

PR #5852, opened 2018-01-28
Explain how `volumes` and `devices` are merged

klaussner:
### Proposed changes

I've added a short explanation and an example about how `volumes` and `devices` are merged if more than one Compose configuration file is used. The current version of the "Adding and overriding configuration" section only has an example for `environments` and it's unclear how entries are merged if they don't have a `key=value` format.

### Related issue

Fixes docker/compose#2633


mistyhacks:
@shin- PTAL

PR #5873, opened 2018-01-31
Update troubleshoot.md

KaishM:
Added "How to enable Virtualization in Windows 10"


mistyhacks:
I don't think we can accept this. Every hardware manufacturer uses slightly different techniques to get into the BIOS and every BIOS has these settings in slightly different places. The only thing we can do is say to consult your hardware provider's documentation.

PR #5877, opened 2018-01-31
Update part 3

gbarr01:
Part of the get-started update


gbarr01:
Have another PR for the getting started update.

Issue #5878, opened 2018-01-31
Missing syntax for configs


mbentley:
[engine/swarm/configs.md](https://docs.docker.com/engine/swarm/configs/), CC @mistyhacks

I am not sure if I do not see it but I am not seeing where the fields are documented for utilizing configs.  I expect that there are settings for source, target, UID, GID, and Mode for the file name where the config appears but I can not seem to find it from googling or searching through the docs.

mbentley:
Looks like I found something that worked from a blog about secrets and adapted it for configs:

`--config-add "source=com.docker.license-2,target=/etc/ucp/docker.lic,uid=65534,gid=65534,mode=292"`

mistyhacks:
You need to use a 4-digit mode and take into account any umask set in the container. You can specify `uid` and `gid` as names instead of numbers if you want. Why does world need to read the license?

mistyhacks:
Narrative docs fix merged and upstream PR submitted, so closing this bug.

PR #5879, opened 2018-01-31
Add info about setting config options

mistyhacks:
Relates to https://github.com/docker/docker.github.io/issues/5878

cc/ @mbentley


mbentley:
Looks great, thanks for the quick work as always!

PR #5914, opened 2018-02-02
Favor docker <object> <verb> format of commands

### LINKED ISSUE #3031: Updating Docs for new CLI?
### Problem description

Are there plans to migrate the docs to use the new CLI commands in all cases? I feel like new users should be exposed to the new commands right away (since it's new and more intuitive) but I don't want to spend time updating if it's not wanted.

### Problem location

Example: https://docs.docker.com/get-started/ "docker run hello-world" vs "docker container run hello-world"

1.13 and later

mistyhacks:
Fixes #3031 

This is not a partial fix but it cleans up the most vague of the old-format commands.

### DIFF
--- a/compose/install.md
+++ b/compose/install.md
@@ -216,7 +216,7 @@ Alternatively, if you're not worried about keeping them, you can remove them.
 Compose just creates new ones.
 
 ```bash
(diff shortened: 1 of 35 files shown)


mistyhacks:
@gbarr01 feel free to merge if you are +1

PR #5919, opened 2018-02-05
fixed links in DTR overview page

michha:
Fixed links in Efficiency section

Live link: https://docs.docker.com/datacenter/dtr/2.4/guides/


joaofnfernandes:
Thanks for helping keep the docs correct 🙌

PR #5924, opened 2018-02-06
Fixed tar options and path for restoring images

cherylq26:
### Proposed changes

`-z` option for tar is used to filter through gzip. This should be excluded for the restore as it was not used during the backup.
The path is also fixed as it recreated the entire path within the volumes folder.


joaofnfernandes:
@cherylq26 Thanks for making our docs better 🙌
Let's ship it 🚀

PR #5926, opened 2018-02-06
17.12 release notes

FrenchBen:
### Proposed changes

This updates the different releases to show the Stable and Edge releases, which took some time to publish. 

* 17.12 Stable
* 18.01 Edge

### DIFF
--- a/docker-for-aws/release-notes.md
+++ b/docker-for-aws/release-notes.md
@@ -27,9 +27,16 @@ title: Docker for AWS release notes
 
 ## Stable channel
 
-### 17.09 CE
+### 17.12 CE
 
 {{aws_blue_latest}}
+
+Release date: 1/9/2018
+
+- Docker Engine upgraded to [Docker 17.12.0 CE](https://github.com/docker/docker-ce/releases/tag/v17.12.0-ce){: target="_blank" class="_"}
+- Kernel patch to mitigates Meltdown attacks ( CVE-2017-5754) and enable KPTI
+
+### 17.09 CE
 
 Release date: 10/6/2017
 
--- a/docker-for-aws/release-notes.md
+++ b/docker-for-aws/release-notes.md
@@ -38,9 +38,7 @@ Release date: 1/9/2018
 - Docker Engine upgraded to [Docker 17.12.0 CE](https://github.com/docker/docker-ce/releases/tag/v17.12.0-ce){: target="_blank" class="_"}
 - Kernel patch to mitigates Meltdown attacks ( CVE-2017-5754) and enable KPTI
 
-**notes**
-
-There is currently an issue in LinuxKit that prevents containers from [starting after a machine reboot](https://github.com/moby/moby/issues/36189){: target="_blank" class="_"}.
+> **Note** There is currently an issue in LinuxKit that prevents containers from [starting after a machine reboot](https://github.com/moby/moby/issues/36189){: target="_blank" class="_"}.
 
 ### 17.09 CE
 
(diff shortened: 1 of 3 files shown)


docker-for-aws/upgrade.md
```diff
@@ -36,6 +36,8 @@ Single containers started (for example) with `docker run -d` are
 **not** preserved during an upgrade. This is because they're not Docker Swarm
 objects, but are known only to the individual Docker engines.
 
+_NIB: Please note that if EFS encryption is enabled, upgrade will [recreate the EFS volume](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/using-cfn-updating-stacks-update-behaviors.html){: target="_blank" class="_"}_
```
netflash: it's not correct. It WILL RECREATE the volumes no matter what. See details https://github.com/docker/for-aws/issues/138

FrenchBen: @netflash thanks for the clarification - I missed this part of the comment: `Not having an "Encrypted" property or having it set to false makes a big difference`
Should we simply remove encryption to prevent this odd behavior?

netflash: Removing that would be the very simple solution.
If you still want to keep it for "Encrypted = true", you may want to deal with the template the way I suggested in the original ticket.

FrenchBen: I'll update the docs to reflect this change, and the next release will remove encryption as well

docker-for-aws/upgrade.md
```diff
@@ -36,8 +36,6 @@ Single containers started (for example) with `docker run -d` are
 **not** preserved during an upgrade. This is because they're not Docker Swarm
 objects, but are known only to the individual Docker engines.
 
-_NIB: Please note that if EFS encryption is enabled, upgrade will [recreate the EFS volume](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/using-cfn-updating-stacks-update-behaviors.html){: target="_blank" class="_"}_
```
netflash: I suggest you should still have something here. As current template breaks things

FrenchBen: Done!

docker-for-aws/release-notes.md
```diff
@@ -25,12 +25,25 @@ title: Docker for AWS release notes
...
+### 17.12 CE
 
 {{aws_blue_latest}}
 
+Release date: 1/9/2018
+
+- Docker Engine upgraded to [Docker 17.12.0 CE](https://github.com/docker/docker-ce/releases/tag/v17.12.0-ce){: target="_blank" class="_"}
+- Kernel patch to mitigates Meltdown attacks ( CVE-2017-5754) and enable KPTI
+
+**notes**
```
mistyhacks: This needs to be a Markdown note.

docker-for-azure/release-notes.md
```diff
@@ -26,10 +26,21 @@ title: Docker for Azure Release Notes
...
+### 17.12 CE
 
 {{azure_blue_latest}}
 
+Release date: 1/9/2018
+
+- Docker Engine upgraded to [Docker 17.12.0 CE](https://github.com/docker/docker-ce/releases/tag/v17.12.0-ce){: target="_blank" class="_"}
+- Kernel patch to mitigates Meltdown attacks ( CVE-2017-5754) and enable KPTI
+
+**notes**
```
mistyhacks: Needs to be a Markdown note

mistyhacks:
Added @gbarr01 since she'll be doing review for these now.

PR #5928, opened 2018-02-06
Update product version references

sgerrand:
### Proposed changes

Update the version strings for the latest GA releases of Docker Engine, Compose, and Machine.

### DIFF
--- a/docker-for-mac/index.md
+++ b/docker-for-mac/index.md
@@ -37,13 +37,13 @@ Run these commands to test if your versions of `docker`, `docker-compose`, and
 
 ```shell
 $ docker --version
-Docker version 17.03.0-ce, build 60ccb22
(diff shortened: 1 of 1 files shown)


mistyhacks:
@gbarr01 consider using the site-scoped variables for these so they are always up to date.

sgerrand:
Sure, that makes sense. Will do.

gbarr01:
Hm, I wonder if it makes more sense to remove that stuff. The user can check for the latest versions in the [release notes](https://docs.docker.com/docker-for-mac/release-notes/).

sgerrand:
Thanks @gbarr01. 🙇

gbarr01:
No problem, @sgerrand ... seems useful to keep around. Let's see what happens when I merge. :)

PR #5936, opened 2018-02-06
Restructure and rewrite network content

### LINKED ISSUE #1692: documentation: iptables and firewalld issues
@yongshin commented

The customer was frustrated because he didn't understand how Docker natively changes your iptables config: https://docs.docker.com/engine/userguide/networking/default_network/container-communication/#communicating-to-the-outside-world . It is possible to run docker without changing iptables but I don't know if its possible to do for UCP. Either way, the customer wishes this assumption that Docker will change your iptable config is documented: 

> I just reinstalled UCP on cloud01 again. Later, I realized that Docker, or UCP, will dynamically add firewall rules to the firewall. I realized that if I restart iptables/Netfilter, iptables will also remove any Firewall rules, which will lead to connectivity issues. As a workaround, I need to make sure that I restart the Docker daemon after iptables is restarted, every time.

The same thing can be said with firewalld since firewalld uses iptables underneath the hood.

### LINKED ISSUE #1895: Request for more in-depth guidance on IPv6
### Problem description

Current docs only describe how to setup IPv6 for _default_ docker network. That is not helpful in most of cases as generally you'd use either docker-compose or manually created networks for various deployments. It also doesn't cover swarm at all.
Additionally, only one use case is covered by the current docs - assigning IPv6 to docker containers.
There's no info on e.g. how to setup IPv6 access _from_ container to external IPv6 network without assigning IPv6 addresses to said containers.

### Problem location

- I saw a problem on the following URL: https://docs.docker.com/engine/userguide/networking/default_network/ipv6/
- I couldn't find the information I wanted. I expected to find it near the following URL: https://docs.docker.com/engine/userguide/networking/ (I guess?)
- Other: more in-depth guide on IPv6 would be quite nice

All of them

### Suggestions for a fix

Would be really good to have a docs that describe how to configure IPv6

### LINKED ISSUE #2379: Instructions fail for making the docker engine listen on a tcp port (as well as a unix socket)
[engine/admin/systemd.md](https://docs.docker.com/engine/admin/systemd/)

I have an installation on a raspberry pi (using raspbian jessie), and am trying to follow instructions to make the docker engine listen on a tcp port (as well as a unix socket).  As this section suggests, I created the `/etc/docker/daemon.json` file and added a `"hosts": ["tcp://0.0.0.0:4243", "unix:///var/run/docker.sock"]` entry.  However, this approach fails with a message that says that there is an entry in the config file trying to override the flags passed to the daemon on startup.  I think this is because `/lib/systemd/system/docker.service` starts the service with the the flag `-H fd://`

It is not clear the correct approach here, but the documentation just doesn't match. I am worried that editing `/lib/systemd/system/docker.service` to remove the flags will just get overwritten on the next upgrade.  Perhaps the debian package should provide a small daemon.json file rather than provide flags on the start

### LINKED ISSUE #2639: Request | Networking: Add example to enable multiple IPs and restrict single port
Restricting access to services on public accessible servers is important.
 
> Docker’s forward rules permit all external source IPs by default. To allow only a specific IP or network to access the containers, insert a negated rule at the top of the DOCKER filter chain. For example, to restrict external access such that only source IP 8.8.8.8 can access the containers, the following rule could be added:
> `$ iptables -I DOCKER -i ext_if ! -s 8.8.8.8 -j DROP`

I would like to see some more details when running multiple servers with multiple services on each:

1. How to restrict access to multiple IPs?
Use case: database cluster where 3 servers connect to each other

2. How to restrict one service/port and leave others open?
Use case: one server with public web server and restricted database (clustered, not "linking")

It would be great if you could provide simple `iptables` examples.

[engine/userguide/networking/default_network/container-communication.md](https://docs.docker.com/engine/userguide/networking/default_network/container-communication/), CC @mstanleyjones

### LINKED ISSUE #4794: example inconsistent with documentation.
[engine/userguide/networking/default_network/custom-docker0.md](https://docs.docker.com/engine/userguide/networking/default_network/custom-docker0/), CC @mstanleyjones

You say "and must be a subset of the bridge IP range" but in your example above that, it's not.

mistyhacks:
This is a huge rewrite of the entire networking area of the docs.
Fixes #1692 
Fixes #1895 
Fixes #2379 
Fixes #2639
Fixes #4794


config/containers/container-networking.md
```diff
@@ -0,0 +1,65 @@
+---
+title: Container networking
+description: How networking works from the container's point of view
+keywords: networking, container, standalone
+redirect_from:
+- /engine/userguide/networking/configure-dns/
+- /engine/userguide/networking/default_network/binding/
+---
+
+The type of network a container uses, whether it is a [brudge](bridges.md), an
```
bbinet: s/brudge/bridge/

mistyhacks: Thanks, this will be fixed in #5945. As this content is now live, please file new issues when you see problems.

PR #5939, opened 2018-02-07
Windows toolbox migration instructions

gtardif:
Instructions for manually migrating data from Toolbox to Docker for Windows.
cc @gbarr01


gbarr01:
What do you think, @gtardif ?

gtardif:
LGTM

PR #5955, opened 2018-02-08
HRM routing mesh on Windows Server

abjinugu:
Note:- Docker on Windows Server doesn't support Swarm Mode Routing Mesh, which means the hrm service is exposed only through the manager node's ip. Therefore DNS needs to configured so that all the host names ('wordpress.example.org' in this example) point to manager nodes and not the worker nodes.


abjinugu:
please ignore this one

PR #5966, opened 2018-02-08
Add known issue to UCP 2.2.5

joaofnfernandes:


khudgins:
Even with the CLI, if it's targeted at UCP and not a single Docker node, users should run into the same problems.

uday-shetty:
@joaofnfernandes: as @khudgins mentioned this will still not work. The problem is with enforcing RBAC on volumes in UCP 2.2.5 for 3rd party plugin providers, since volume plugin API doesn't support label feature in Docker Engine.

joaofnfernandes:
Got it. Is there any workaround that allows UCP users to create volumes from 3rd-parties?
If the plugin API doesn't allow you to label the volume, this means that users really need to use the local driver. Right?
So what should we tell users to do?

uday-shetty:
@joaofnfernandes we need to add a note with positive spin as it's not an issue with plugin vendors since labels for volumes are not supported in Docker Engine or added to plugin helper APIs

 "enforcing RBAC for volumes in UCP 2.2.5 will not work with 3rd party volume plugins. Customers using 3rd party volume plugins planning to upgrade UCP should use next planned version of UCP 2.2.6  which will provide an option to turn off enforcement of RBAC for volumes"

@khudgins @marcusmartins -- please review and approve it to be added to UCP 2.2.5 release notes

khudgins:
I'd be a bit more explicit in the wording for how to manage versions:

"Customers using 3rd party volume plugins planning to upgrade UCP should skip UCP 2.2.5 and instead use the upcoming release of 2.2.6 which will provide an option to turn off enforcement of RBAC for volumes."

uday-shetty:
@khudgins SGTM

PR #5974, opened 2018-02-09
ENV layer persistence clarification

foodebeer:
Hi John,

I came across a Slack conversation that may be relevant for the Dockerfile ENV section. Basically from what I understand the following happens:
When you use ENV, docker build creates a new intermediate layer that persists the environment variable. When you try to unset it in the next layer, the lower layer still keeps the environment variable.

I tried to explain this.

Credits go to @church and @clemenko.

### DIFF
--- a/develop/develop-images/dockerfile_best-practices.md
+++ b/develop/develop-images/dockerfile_best-practices.md
@@ -461,30 +461,46 @@ Similar to having constant variables in a program (as opposed to hard-coding
 values), this approach lets you change a single `ENV` instruction to
 auto-magically bump the version of the software in your container.
 
-Because a new intermediate layer is created after running the ENV command, the environment variable is effectively persisted.
-If you try to unset the variable afterwards, then the set value of the lower lying layer is used instead:
-Dockerfile
-`FROM alpine
+Each `ENV` line creates a new intermediate layer, just like `RUN` commands. This
+means that even if you unset the environment variable in a future layer, it
+still persists in this layer and its value can be dumped. You can test this by
+creating a Dockerfile like the following, and then building it.
+
+```Dockerfile
+FROM alpine
 ENV ADMIN_USER="mark"
 RUN echo $ADMIN_USER > ./mark
 RUN unset ADMIN_USER
-CMD sh`
+CMD sh
+```
 
-Build and run: `$ docker run --rm -it test sh` shows:
-`/ # echo $ADMIN_USER
+```bash
+$ docker run --rm -it test sh echo $ADMIN_USER
+
 mark
-`
+```
 
-If the environment variable needs to be unset afther the command, or its content should not be serialized, use:
-Dockerfile
-`FROM alpine
-RUN export ADMIN_USER="mark"; echo $ADMIN_USER > ./mark;  unset ADMIN_USER
-CMD sh`
+To prevent this, and really unset the environment variable, use a `RUN` command
+with shell commands, to set, use, and unset the variable all in a single layer.
+You can separate your commands with `;` or `&&`. If you use the second method,
+and one of the commands fails, the `docker build` also fails. This is usually a
+good idea. Using `\` as a line continuation character for Linux Dockerfiles
+improves readability. You could also put all of the commands into a shell script
+and have the `RUN` command just run that shell script.
 
-Build and run: `$ docker run --rm -it test sh` shows:
-`/ # echo $ADMIN_USER
+```Dockerfile
+FROM alpine
+RUN export ADMIN_USER="mark" \
+    && echo $ADMIN_USER > ./mark \
+    && unset ADMIN_USER
+CMD sh
+```
 
-`
+```bash
+$ docker run --rm -it test sh echo $ADMIN_USER
+
+```
+
 
 ### ADD or COPY
 
--- a/develop/develop-images/dockerfile_best-practices.md
+++ b/develop/develop-images/dockerfile_best-practices.md
@@ -461,6 +461,31 @@ Similar to having constant variables in a program (as opposed to hard-coding
 values), this approach lets you change a single `ENV` instruction to
 auto-magically bump the version of the software in your container.
 
+Because a new intermediate layer is created after running the ENV command, the environment variable is effectively persisted.
+If you try to unset the variable afterwards, then the set value of the lower lying layer is used instead:
+Dockerfile
+`FROM alpine
+ENV ADMIN_USER="mark"
+RUN echo $ADMIN_USER > ./mark
+RUN unset ADMIN_USER
+CMD sh`
+
+Build and run: `$ docker run --rm -it test sh` shows:
+`/ # echo $ADMIN_USER
+mark
+`
+
+If the environment variable needs to be unset afther the command, or its content should not be serialized, use:
+Dockerfile
+`FROM alpine
+RUN export ADMIN_USER="mark"; echo $ADMIN_USER > ./mark;  unset ADMIN_USER
+CMD sh`
+
+Build and run: `$ docker run --rm -it test sh` shows:
+`/ # echo $ADMIN_USER
+
+`
+
 ### ADD or COPY
 
 - [Dockerfile reference for the ADD instruction](/engine/reference/builder.md#add)


mistyhacks:
I'd like to copyedit this with you, just give me a couple hours.

ChristianKniep:
@foodebeer @mistyhacks I am sorry, but IMHO it's not quite correct that `ENV` writes a filesystem layer.

```Dockerfile
FROM alpine
ENV ADMIN_USER="mark"
RUN echo $ADMIN_USER > ./mark
RUN unset ADMIN_USER
CMD sh
```
This has two layers:
```bash
$ docker inspect test|jq '.[] |.RootFS.Layers'
[
  "sha256:cd7100a72410606589a54b932cabd804a17f9ae5b42a1882bd56d263e02b6215",
  "sha256:09fd1cd8e789b055f962a2a18403be0a016ed44f57853172c3cb9789cb49c8f3"
]
```

The first one is the `alpine` layer and the second one comes from the first `RUN`, the second one does not change the file-system, thus it does not result in a layer.
```bash
$ docker inspect alpine|jq '.[] |.RootFS.Layers'
[
  "sha256:cd7100a72410606589a54b932cabd804a17f9ae5b42a1882bd56d263e02b6215"
]
$ docker history test
IMAGE               CREATED             CREATED BY                                      SIZE                COMMENT
9df871acb23e        5 minutes ago       /bin/sh -c #(nop)  CMD ["/bin/sh" "-c" "sh"]    0B
59e0acb5af42        5 minutes ago       /bin/sh -c unset ADMIN_USER                     0B
cd1dae97f276        5 minutes ago       /bin/sh -c echo $ADMIN_USER > ./mark            5B
af93bab4f3f4        5 minutes ago       /bin/sh -c #(nop)  ENV ADMIN_USER=mark          0B
3fd9065eaf02        4 weeks ago         /bin/sh -c #(nop)  CMD ["/bin/sh"]              0B
<missing>           4 weeks ago         /bin/sh -c #(nop) ADD file:093f0723fa46f6cdb…   4.15MB
```

`ENV` will be stored in the meta-data part of the Manifest:

```
$ docker inspect test|jq '.[] |.Config.Env'
[
  "PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin",
  "ADMIN_USER=mark"
]
```

`unset`/`export` within a `RUN` results in a change within the instantiated container for this particular `RUN` phase. For the next `RUN` as well as for an instantiation using `docker run` the environment will be rest to what is defined in the `Dockerfile`.

ChristianKniep:
In addition the proposed way of dealing with sensitive data using `export` and `unset` is insecure, because `docker history` will still expose the `ADMIN_USER` value.

AFAIK there is no easy way to solve this, one might `COPY` in a file with the password, `rm` the file at the end of the `Dockerfile` and use `--squash` to merge  all layers together. The file containing the password should be in `.gitignore` though, so that it has to be created at build-time.
Doing this `history` will still show that a file was added, but since it is not part of the merged `layer` it is not recoverable.

Downside is that layering is not a thing for this Dockerfile then.

thaJeztah:
@ChristianKniep is correct here that, since the switch to a content-addressable store in Docker 1.10, `ENV` no longer creates a new image _layer_, but they still have to be taken into account when using them for secrets (which should be discouraged), because they are exposed in `docker history`.

So, even though environment variables and build-args are not persisted in an image _layer_ (unless you explicitly write them to a layer), they are always preserved in the image's history. In the output of `docker history` with the `--no-trunc` option, the value of build-args is prefixed for each `RUN`:

```bash
docker history --no-trunc localhost:5000/myimage

IMAGE                                                              CREATED              CREATED BY                                                                                          SIZE                COMMENT
34a1bb5f2b67202c97559f7880eef8b68ddf89608ad25ce5bae00c510ab31a7c   About a minute ago   /bin/sh -c #(nop) ARG ALSO_SECRET=                                                                  0 B                 
51077539b6b8cd79ae33d6490c0e9d85d310a59e2eff0309e35856da3bc1e79a   About a minute ago   /bin/sh -c #(nop) ENV TOP_SECRET=                                                                   0 B                 
1746a99de8a37338ecc6f7b2aae31b62d65093964d2955a755adf9f4dc7daa00   About a minute ago   |1 ALSO_SECRET=mykey /bin/sh -c echo hello world                                                    0 B                 
743e6740c3bd2f83b1e7598058cec8501dc2aaaadc15e28f84262f76517bf156   About a minute ago   /bin/sh -c #(nop) ARG ALSO_SECRET=mykey                                                             0 B                 
caaf0aa25ada17db4f761a68daaa4764496ec7725129ad6cfa5db65f1010c83a   About a minute ago   /bin/sh -c #(nop) ENV TOP_SECRET=password                                                           0 B                 
ae3a22f6ae7b7c8f7efa55bd5c7ee0322ab6ef201b9d746932b0546bbb8c60c9   2 weeks ago          /bin/sh -c #(nop)  CMD ["sh"]                                                                       0 B                 
6b9456aadd7e7a928746af7ccf0f25d6fc43e2b4813cbe9988ff35b0bbeca8c2   2 weeks ago          /bin/sh -c #(nop) ADD file:413866e909560e078243ce3a4fe1ec87fa5c966a7d47d7f1d9448a2dcbc26a5a in /    1.146 MB            
```

Note that build-args are _not_ designed for passing secrets; they are explicitly _not_ encoded/encrypted to prevent people from _assuming_ they are safe to be used for that.




_Before_ Docker 1.10, every step in a Dockerfile would produce a full image, including an image layer (even if the step would not make changes to the filesystem).

As an example, take the following Dockerfile:

```Dockerfile
FROM busybox
ENV TOP_SECRET=password
ARG ALSO_SECRET=mykey
RUN echo hello world
ENV TOP_SECRET=""
ARG ALSO_SECRET=""
```

## Situation in Docker 1.9 and under

In Docker 1.9 and under, every step in a Dockerfile produced an _image_, and each image produced has a "link" to its parent image. All images produced during build are kept both in the local cache, but are also _pushed_ when pushing to a registry, and _pulled_ when pulling the image. This is something to be aware of, because this means that every intermediate step is made available through the registry (which could be Docker Hub).

To see this in action, start a local registry ("mimicking" Docker Hub):

```bash
docker run -d --name registry -p 5000:5000 registry:2
```

Then, build the Dockerfile;

```bash
docker build -t localhost:5000/myimage -<<EOF
FROM busybox
ENV TOP_SECRET=password
ARG ALSO_SECRET=mykey
RUN echo hello world
ENV TOP_SECRET=""
ARG ALSO_SECRET=""
EOF
```

After building completed, you'll see that each step in the build produced an image:

```bash
docker images -a
REPOSITORY               TAG                 IMAGE ID            CREATED             VIRTUAL SIZE
localhost:5000/myimage   latest              34a1bb5f2b67        5 seconds ago       1.146 MB
<none>                   <none>              51077539b6b8        6 seconds ago       1.146 MB
<none>                   <none>              1746a99de8a3        6 seconds ago       1.146 MB
<none>                   <none>              743e6740c3bd        7 seconds ago       1.146 MB
<none>                   <none>              caaf0aa25ada        7 seconds ago       1.146 MB
<none>                   <none>              6b9456aadd7e        2 weeks ago         1.146 MB
busybox                  latest              ae3a22f6ae7b        2 weeks ago         1.146 MB
...
```

Those images are chained together using the `Parent` property of the image:

```bash
docker inspect --format '{{.Parent}}' localhost:5000/myimage 
51077539b6b8cd79ae33d6490c0e9d85d310a59e2eff0309e35856da3bc1e79a

docker inspect --format '{{.Parent}}' 51077539b6b8cd79ae33d6490c0e9d85d310a59e2eff0309e35856da3bc1e79a
1746a99de8a37338ecc6f7b2aae31b62d65093964d2955a755adf9f4dc7daa00

(and so on)
```

To see that all images are also preserved when pushing to a registry;


First, push the image to a registry:

```bash
docker push localhost:5000/myimage

The push refers to a repository [localhost:5000/myimage] (len: 1)
34a1bb5f2b67: Pushed 
51077539b6b8: Pushed 
1746a99de8a3: Pushed 
743e6740c3bd: Pushed 
caaf0aa25ada: Pushed 
6b9456aadd7e: Pushed 
latest: digest: sha256:b5179c1c03ed77e54c8e2c59fb3a73da39f1e6f65fc8df296600289750a9811d size: 9822
```

Then _remove_ the local image:

```bash
docker rmi localhost:5000/myimage

Untagged: localhost:5000/myimage:latest
Deleted: 34a1bb5f2b67202c97559f7880eef8b68ddf89608ad25ce5bae00c510ab31a7c
Deleted: 51077539b6b8cd79ae33d6490c0e9d85d310a59e2eff0309e35856da3bc1e79a
Deleted: 1746a99de8a37338ecc6f7b2aae31b62d65093964d2955a755adf9f4dc7daa00
Deleted: 743e6740c3bd2f83b1e7598058cec8501dc2aaaadc15e28f84262f76517bf156
Deleted: caaf0aa25ada17db4f761a68daaa4764496ec7725129ad6cfa5db65f1010c83a
```

And verify that the local image, including intermediate images, is gone:

```bash
docker images -a

REPOSITORY          TAG                 IMAGE ID            CREATED             VIRTUAL SIZE
busybox             latest              ae3a22f6ae7b        2 weeks ago         1.146 MB
<none>              <none>              6b9456aadd7e        2 weeks ago         1.146 MB
...
```

Now, _pull_ the image again:

```bash
docker pull localhost:5000/myimage

Using default tag: latest
latest: Pulling from myimage
cdc89926aaf4: Pull complete 
c0b499e046f3: Pull complete 
7ae99a511ded: Pull complete 
bf20b6e3d471: Pull complete 
5db359c9855c: Pull complete 
da8c7ceebae6: Pull complete 
e37f333db5f4: Pull complete 
Digest: sha256:b5179c1c03ed77e54c8e2c59fb3a73da39f1e6f65fc8df296600289750a9811d
Status: Downloaded newer image for localhost:5000/myimage:latest
```

And see that after pulling, all the intermediate images are back:

```bash
docker images -a

REPOSITORY               TAG                 IMAGE ID            CREATED             VIRTUAL SIZE
localhost:5000/myimage   latest              e37f333db5f4        23 minutes ago      1.146 MB
<none>                   <none>              da8c7ceebae6        23 minutes ago      1.146 MB
<none>                   <none>              5db359c9855c        23 minutes ago      1.146 MB
<none>                   <none>              bf20b6e3d471        23 minutes ago      1.146 MB
<none>                   <none>              7ae99a511ded        23 minutes ago      1.146 MB
<none>                   <none>              c0b499e046f3        2 weeks ago         1.146 MB
<none>                   <none>              6b9456aadd7e        2 weeks ago         1.146 MB
<none>                   <none>              cdc89926aaf4        2 weeks ago         1.146 MB
busybox                  latest              ae3a22f6ae7b        2 weeks ago         1.146 MB
...
```

And the intermediate images contain the environment variable (as well as build-args);

```bash
docker inspect --format '{{json .Config.Env}}' bf20b6e3d471

["PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin","TOP_SECRET=password"]


docker inspect --format '{{json .ContainerConfig.Cmd}}' bf20b6e3d471

["/bin/sh","-c","#(nop) ARG ALSO_SECRET=mykey"]
```

## Situation in Docker 1.10 and up

Docker 1.10 switched to a new, content-addressable store for images. In the new store, each image is self-dependent, and no longer requires parent images to be pushed or pulled; where older versions used the parent image chain to get the associated image layer(s), docker 1.10 and up keeps a list of image layers that is used by an image.

Using the same steps as above to illustrate how this differs;


Start a local registry:

```bash
docker run -d --name registry -p 5000:5000 registry:2
```

Build the Dockerfile;

```bash
docker build -t localhost:5000/myimage -<<EOF
FROM busybox
ENV TOP_SECRET=password
ARG ALSO_SECRET=mykey
RUN echo hello world
ENV TOP_SECRET=""
ARG ALSO_SECRET=""
EOF
```

After building completed, you'll see that each step in the build produced an image:

```bash
docker images -a
REPOSITORY               TAG                 IMAGE ID            CREATED             SIZE
localhost:5000/myimage   latest              39137fd587e0        53 seconds ago      1.15MB
<none>                   <none>              46cd462bd350        53 seconds ago      1.15MB
<none>                   <none>              244a3b366f7a        54 seconds ago      1.15MB
<none>                   <none>              cb5c0706d847        55 seconds ago      1.15MB
<none>                   <none>              d89744054384        55 seconds ago      1.15MB
busybox                  latest              5b0d59026729        2 weeks ago         1.15MB
...
```

Even the parent chain is present locally:

```bash
docker inspect --format '{{.Parent}}' localhost:5000/myimage 
sha256:46cd462bd350a29535f37911dc8f1e110fd10e89c7b525cd1ce68be82baa5caa

docker inspect --format '{{.Parent}}' sha256:46cd462bd350a29535f37911dc8f1e110fd10e
89c7b525cd1ce68be82baa5caa
sha256:244a3b366f7a47449c3120a14e65c0529bfedf6ef44dba5e4b33f2a0437ed0a5
```

#### So, what's different?

The difference is subtle, but should become more clear later on.

First of all, the image now has a `Layers` property:

```bash
docker inspect --format '{{json .RootFS.Layers}}' localhost:5000/myimage

["sha256:4febd3792a1fb2153108b4fa50161c6ee5e3d16aa483a63215f936a113a88e9a"]
```

As mentioned earlier; the content-addressable store produces self-contained images; the `.RootFS.Layers` property is part of that, as the image itself knows about all the layers it needs, without needing the "parent" chain to get that information. In the example output above, only a _single_ layer was produced by the Dockerfile.


While the `.Parent` chain is present _locally_; that chain is only used for the build-cache. Intermediate images are only available on the machine from where the build was performed, but are never pushed to a registry, nor pulled from a registry.

To see this, first push the image to a registry, and notice that only a single layer (and image) is pushed:

```bash
docker push localhost:5000/myimage
The push refers to repository [localhost:5000/myimage]
4febd3792a1f: Pushed 
latest: digest: sha256:449b96344e7f4f31a1b34954586ca51f239892a16d7dba8695abb6822852ae88 size: 527
```

Now, remove the local image:

```bash
docker rmi localhost:5000/myimage

Untagged: localhost:5000/myimage:latest
Untagged: localhost:5000/myimage@sha256:449b96344e7f4f31a1b34954586ca51f239892a16d7dba8695abb6822852ae88
Deleted: sha256:39137fd587e057f0c40dd80bd9cffb880a21a5e6b657994a8b02747daa529f3b
Deleted: sha256:46cd462bd350a29535f37911dc8f1e110fd10e89c7b525cd1ce68be82baa5caa
Deleted: sha256:244a3b366f7a47449c3120a14e65c0529bfedf6ef44dba5e4b33f2a0437ed0a5
Deleted: sha256:d89744054384e216a609c35f67ecb92f0041e982237f5117480f38a94b797d7a
Deleted: sha256:cb5c0706d84718bf3ac9e80988430269f9eb80fc55399071250a274ade051f5a
```

And verify that the local image (and all intermediate images) is gone:

```bash
docker images -a

busybox             latest              5b0d59026729        2 weeks ago         1.15MB
...
```


Pull the image again from the registry, and notice only a single image (and, in this case, layer) is pulled:

```bash
docker pull localhost:5000/myimage

Using default tag: latest
latest: Pulling from myimage
57310166fe88: Already exists 
Digest: sha256:449b96344e7f4f31a1b34954586ca51f239892a16d7dba8695abb6822852ae88
Status: Downloaded newer image for localhost:5000/myimage:latest
```

Checking the local image cache shows that only the tagged image was pulled, and none of the intermediate images:

```bash
docker images -a

REPOSITORY               TAG                 IMAGE ID            CREATED             SIZE
localhost:5000/myimage   latest              39137fd587e0        11 minutes ago      1.15MB
busybox                  latest              5b0d59026729        2 weeks ago         1.15MB
...
```

Inspecting the image shows that the `.Parent` chain is gone (i.e., it's not pushed/pulled from a registry);

```bash
docker inspect --format '{{json .Parent}}' localhost:5000/myimage 
""
```

Only the final value for the Env var is preserved in the image configuration (note that in this example, no env-var was written to disk: if the Dockerfile wrote the env-var to disk, that would still be persisted in the final image);

```bash
docker inspect --format '{{json .Config.Env}}' localhost:5000/myimage

["PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin","TOP_SECRET="]
```


Be aware though, that even though the intermediate images themselves are not pushed/pulled, `docker history` still preserves information _about_ those intermediate steps:

```bash
docker history --no-trunc localhost:5000/myimage
IMAGE                                                                     CREATED             CREATED BY                                                                                          SIZE                COMMENT
sha256:39137fd587e057f0c40dd80bd9cffb880a21a5e6b657994a8b02747daa529f3b   24 minutes ago      /bin/sh -c #(nop)  ARG ALSO_SECRET=                                                                 0B                  
<missing>                                                                 24 minutes ago      /bin/sh -c #(nop)  ENV TOP_SECRET=                                                                  0B                  
<missing>                                                                 24 minutes ago      |1 ALSO_SECRET=mykey /bin/sh -c echo hello world                                                    0B                  
<missing>                                                                 24 minutes ago      /bin/sh -c #(nop)  ARG ALSO_SECRET=mykey                                                            0B                  
<missing>                                                                 24 minutes ago      /bin/sh -c #(nop)  ENV TOP_SECRET=password                                                          0B                  
<missing>                                                                 2 weeks ago         /bin/sh -c #(nop)  CMD ["sh"]                                                                       0B                  
<missing>                                                                 2 weeks ago         /bin/sh -c #(nop) ADD file:413866e909560e078243ce3a4fe1ec87fa5c966a7d47d7f1d9448a2dcbc26a5a in /    1.15MB              
````


(The `<missing>` labels in the `IMAGE` column show that the image itself is not present)

## How to keep steps out of your image's history?

First of all, let me repeat that build-args were _not_ designed for secrets


### Do you need a secret _in_ your Dockerfile?

A frequent use of secrets in a Dockerfile is to clone a private git repository, for example (using OAuth token, just for illustration; there's different ways to use credentials)

```Dockerfile
FROM foobar
ARG OAUTH_TOKEN

# Note: use git pull, not git clone to prevent the OAuth token
# from being stored on disk (see https://github.com/blog/1270-easier-builds-and-deployments-using-git-over-https-and-oauth)
RUN mkdir source \
 && cd source \
 && git init \
 && git pull https://${OAUTH_TOKEN}@github.com/foobar/private-repo.git

RUN make build # (for example)
```

The above approach has a number of downsides:

- The `git pull` (or `git clone`) needs to have access to a secret in order to get access to the repository
- The `git pull` is non-deterministic; Docker has not enough information to determin if the cache needs to be invalidated. In the above example, it will use the build cache, even if the `git pull` in the cache is outdated.
- The build requires networking access (which may not be desirable).


Instead of performing a `git clone` or `git pull` as a step in the Dockerfile, consider adding the Dockerfile to the git repository, and perform cloning the repsitory _before_ starting the build;

```Dockerfile
FROM foobar
RUN mkdir source \
 && cd source 

# Copy source files
COPY . .

RUN make build # (for example)
```

Now, to build;


```bash
git clone https://github.com/foobar/private-repo.git .
docker build -t myimage .
```

Advantages:

- Secrets/credentials never enter the Docker build
- The Dockerfile is locked-in together with the source code, and
- The build-cache is automatically invalidated if any file changed in the git repository (Docker calculates a checksum of the files that are in the build context to determin if the cache should be used)

### If you _need_ secrets in your Dockerfile

In some cases, secrets may still be needed during build. Consider using a multi-stage build to keep the secrets out of your final image:


```Dockerfile
FROM foobar AS build_stage
ARG MY_SECRET
COPY ./id_rsa /root/id_rsa
RUN make foo

FROM alpine

# Only copy the produced artifacts
COPY --from=build_stage /build-artifacts/my-binary /usr/local/bin/my-binary
```

In the example above, the final build stage only contains the build artifact; environment variables, build-args, and files containing secrets are not copied, so won't be persisted in the final image.


This approach works well, but take into account that: the intermediate images in the local cache _do_ contain the secrets. These images are not pushed to a registry (unless you explicitly push them), but if you do not control the host on which the image is built (which may be the case in a shared environment), so the secrets will be available to anyone that has access to the build server.

PR #5980, opened 2018-02-10
Update bridge.md

gore28akshay:
Spelling mistake.

### Proposed changes
Current line : User-defined **brides** provide automatic DNS resolution between containers .
Changed line : User-defined **bridges** provide automatic DNS resolution between containers.


joaofnfernandes:
`User-defined brides` 🤣
Thanks for fixing it!

PR #5981, opened 2018-02-10

gore28akshay:
### Proposed changes
Current :- docker network connect alpine-net alpine4
Changed :- docker network connect bridge alpine4 

Container alpine4 is already connected to network alpine-net. It has to be connected to bridge.


joaofnfernandes:
Thanks for the contribution @gore28akshay!

PR #5982, opened 2018-02-11
add info about arm64 repo

gidder:
Added info about arm64 repo for Debian.


gbarr01:
@arm64b ...

arm64b:
>On the docker-ce/debian page, we provide docker run commands for x86_64 and armhf. Does arm64 have a unique command or is it the same as x86_64?

Hello @gbarr01 , on AArch64, if the image of `hello-world` has multi-arch support, we can `docker run` it just as `x86_64` does: `sudo docker run hello-world`, if the `hello-world` image is non-multiarch, then we need use this one: `sudo docker run arm64v8/hello-world`. So the question is depended on the docker image itself: multi-arch or non-multiarch.  Does this answer make sense for you ❓

gbarr01:
Hi, @arm64b, it makes perfect sense. Thanks so much.

PR #5984, opened 2018-02-11
Fix #5219: Change unsafe ARG example in compose file docs.

### LINKED ISSUE #5219: Compose file reference has an unsafe example leaking a secret password into persistent Dockerfile history.
### Problem description

Docker's compose file reference has an example with a secret password in a Dockerfile. Meanwhile, the Dockerfile reference warns against this practice.

### Problem location

[Compose file reference](https://docs.docker.com/compose/compose-file/#args):

> First, specify the arguments in your Dockerfile:
> 
>     ARG buildno
>     ARG password
>     
>     RUN echo "Build number: $buildno"
>     RUN script-requiring-password.sh "$password"

[Dockerfile reference](https://docs.docker.com/engine/reference/builder/#arg):

> **Warning**: It is not recommended to use build-time variables for passing secrets like github keys, user credentials etc. Build-time variable values are visible to any user of the image with the `docker history` command.

I dunno. At least v17.09 (current as of this writing).

### Suggestions for a fix

Come up with a better example in the compose-file reference =) I'm too new at this to dare provide any suggestions.

bjorm:
### Proposed changes
Change example passing a password using ARG instructions to passing a commit hash instead.

The docker builder docs [1] advise not to pass any kind of secret to the build process using ARG since it will be retrievable using `docker history`.
[1] https://docs.docker.com/engine/reference/builder/#arg

### DIFF
--- a/compose/compose-file/compose-file-v2.md
+++ b/compose/compose-file/compose-file-v2.md
@@ -165,10 +165,10 @@ build process.
 First, specify the arguments in your Dockerfile:
 
     ARG buildno
-    ARG password
+    ARG gitcommithash
 
     RUN echo "Build number: $buildno"
-    RUN script-requiring-password.sh "$password"
+    RUN echo "Based on commit: $gitcommithash"
 
 Then specify the arguments under the `build` key. You can pass either a mapping
 or a list:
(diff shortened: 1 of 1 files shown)


bjorm:
The jenkins build failed because some operation timed out. Is there a way to run it again without pushing another commit?

gbarr01:
Hi, @shin-  PTAL. Do you approve of the commit?

As @bjorm points out, we have this warning here:  https://docs.docker.com/engine/reference/builder/#arg

> Warning: It is not recommended to use build-time variables for passing secrets like github keys, user credentials etc. Build-time variable values are visible to any user of the image with the docker history command.

But we also have this example here: 

https://docs.docker.com/compose/compose-file/
https://docs.docker.com/compose/compose-file/compose-file-v2/

> Add build arguments, which are environment variables accessible only during the build process.
> 
> First, specify the arguments in your Dockerfile:
> 
> ```
> ARG buildno
> ARG password
> ```
> 
> RUN echo "Build number: $buildno"
> RUN script-requiring-password.sh "$password"
> Then specify the arguments under the build key. You can pass either a mapping or a list:
> 
> ```
> build:
>   context: .
>   args:
>     buildno: 1
>     password: secret
> 
> build:
>   context: .
>   args:
>     - buildno=1
>     - password=secret
> ```
> 
> You can omit the value when specifying a build argument, in which case its value at build time is the value in the environment where Compose is running.
> 
> ```
> args:
>   - buildno
>   - password
> ```

PR #5988, opened 2018-02-12
Update network-tutorial-overlay.md

sify21:
### Proposed changes

`--global` changed to `--mode globel`
I'm using docker-ce-17.09.1, `--global` is not recognized in `docker service create` command.


joaofnfernandes:
Thanks for keeping the docs correct @sify21 .
Let's ship it 🚀

PR #5993, opened 2018-02-12
Update docker-ee.md

carlfischer1:
Restore Docker install docs for Windows Server to their original form using DockerMsftProvider for consistency with Microsoft docs. DockerProvider will be used for Insider/Preview builds only.


joaofnfernandes:
This seems to be correct, but I'd like @johnstep's thumbs on this.

carlfischer1:
paging Dr. @johnstep

joaofnfernandes:
Thanks for the follow-up. Let's ship it!

PR #5997, opened 2018-02-13
Fix typos (`8021.q` -> `802.1q`)

jjlin:
### Proposed changes

Global search and replace of `8021.q` with `802.1q`.


joaofnfernandes:
@jjlin thanks for fixing the docs!
Let's ship it 🚀🌏🚗

PR #6001, opened 2018-02-14
Update mirror.md

AkihiroSuda:
It was not obvious that additional configuration is needed for the official Registry image.

cc @dmcgowan @stevvooe

### DIFF
--- a/registry/recipes/mirror.md
+++ b/registry/recipes/mirror.md
@@ -59,6 +59,8 @@ be configured to use the `filesystem` driver for storage.
 
 The easiest way to run a registry as a pull through cache is to run the official
 Registry image.
+At least, you need to specify `proxy.remoteurl` within `/etc/docker/registry/config.yml`
+as described in the following subsection.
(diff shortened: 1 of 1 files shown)


joaofnfernandes:
@AkihiroSuda thanks for helping make the docs better!

Your change seems a bit redundant given that a paragraph below the docs tell you that you need to include that snippet in the config file.
So I'm wondering what's the core problem you're trying to address? Is it the fact that's not obvious where the "config" file resides in the filesystem? Is it that `username` and `password` are optional?

I'm asking because we might need to do some structural changes to that doc page. If find the content a bit confusing. One of the reasons is that we use the words "mirror" and "cache" interchangeably.
People seem to agree that the article is confusing because the page has 20 👍, 51 👎

AkihiroSuda:
The problem is that `The easiest way to run a registry as a pull through cache is to run the official
  Registry image.` looks as if no configuration is needed at all

PR #6003, opened 2018-02-14
Fix typo

giancarloromeo:
Wrong docker-compose filename.


joaofnfernandes:
`docker-compose-yml` 😂
Thanks for fixing it @giancarloromeo. Let's ship it 🚀

PR #6009, opened 2018-02-14
Updated customer FAQ

nandhini915:
Clarify docker login via CLI

### Proposed changes

Added a Q&A regarding Docker Login. We've received customer feedback where customers end up doing `docker login store.docker.com` which is incorrect. 
We're adding this Q&A so customers get clarity on what is expected of them when pulling certified content.


joaofnfernandes:
Let's ship it! @nandhini915 as a best practice for next time, try keeping headings a bit shorter 😄

PR #6026, opened 2018-02-16
Ensure custom build hooks consume DOCKERFILE_PATH

RJacksonm1:
I have a custom build hook to pass in build arguments on a repository from which I build two Docker images (tags), from two Dockerfiles.

My second build, however, with the custom Dockerfile, was not being built correctly. It was being built with the default `Dockerfile`, instead of my specified custom one.

I quickly figured out the custom build hook was to blame because it doesn't specify which Dockerfile it should build from, but there was no mention in the documentation of an environmental variable I can consume to get the custom "Dockerfile location" value. By dumping the env vars during the build process, I can there is an environmental variable with that value: `DOCKERFILE_PATH`.

This PR updates the documentation accordingly, to:

1) Expose the fact this environmental variable exists, and explains its purpose
2) Make sure its consumed in the example custom build hook, so any copy-pasters find themselves already supporting custom Dockerfiles, rather than struggling with the same problem

### DIFF
--- a/docker-cloud/builds/advanced.md
+++ b/docker-cloud/builds/advanced.md
@@ -19,6 +19,7 @@ processes and do not affect your service's run environment.
 * `SOURCE_COMMIT`: the SHA1 hash of the commit being tested.
 * `COMMIT_MSG`: the message from the commit being tested and built.
 * `DOCKER_REPO`: the name of the Docker repository being built.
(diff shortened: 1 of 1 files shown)


joaofnfernandes:
@KickingTheTV can you confirm that we set the `DOCKERFILE_PATH` environment variable during the build stage?

joaofnfernandes:
Thanks! Let's ship it 🚀

PR #5828, opened 2018-01-25
Update part 2

gbarr01:
Update Get Started part 2 with a slightly new example and tightened text. 

Fixes part of #5807

PR #6030, opened 2018-02-16
Hostname mismatch

arjun-urs:
Hostname from the browser should match the curl output.


gbarr01:
Thanks, @arjun-urs , I agree. I'm working on updating The Get Started Guide. Actually, I'm removing that part altogether because it's redundant (from part 1). More on this soon: https://github.com/docker/docker.github.io/pull/5828/files

PR #6032, opened 2018-02-19
Remove adjtimex from seccomp blocked syscall list.

brthor:
### Proposed changes

I removed adjtimex from the list of blocked syscalls by the default profile in docker.
It is whitelisted in the current default seccomp profile.

https://github.com/moby/moby/blob/master/profiles/seccomp/default.json#L58

### Related issues (optional)

https://github.com/moby/moby/issues/33126


thaJeztah:
ping @tonistiigi @justincormack

justincormack:
The related issue is closed; the syscall was added if you have `CAP_SYS_TIME`, is that not ok? What is the issue you have?

justincormack:
Well it is blocked unless you have `CAP_SYS_TIME` enabled, which is not on by default.

brthor:
@justincormack 

But it is no longer blocked by default in the syscomp profile, which is what most readers are expecting the table in this documentation to describe.

PR #6051, opened 2018-02-22
Mention that rootdirectory prefix  has to be pre-created

xueshanf:
Like bucket, to use a prefix to separate data, the `bucketname/<prefix>` needs to be pre-created, otherwise, you get 503 error when starting registry.


joaofnfernandes:
@stevvooe I'd like a thumbs up from you before accepting this contribution.

L-Hudson:
@nandhini915 can you give this a look so we can resolve? Thank you!

nandhini915:
lgtm.

ahh-docker:
Reviewed with @bermudezmt. Merging.

PR #6054, opened 2018-02-23
Update swift.md

scjane:


joaofnfernandes:
Ahhhh... nothing better than hand-baked html ☕️
Thanks for fixing this!

scjane:
@joaofnfernandes It's my pleasure to fix them when learning Docker docs🐸

PR #6056, opened 2018-02-23

LydellManganti:
Fix Step 7 Detach from alpine1 instead of alpine2. Previous step is to attach to alpine1.


joaofnfernandes:
Thanks for the contribution @LydellManganti. Let's ship it 🚀

Issue #6015, opened 2018-02-15
Get Docker EE for CentOS page references RHEL version


arueth:
[install/linux/docker-ee/centos.md](https://docs.docker.com/install/linux/docker-ee/centos/), CC @johndmulhausen 

The statement:
```
If selinux is enabled, the overlay2 storage driver is only supported on RHEL 7.4 or higher.
```
should reference the CentOS version, not RHEL.

joaofnfernandes:
@dmp42 we have two contradictory statements. The [compatibility matrix](https://success.docker.com/article/Compatibility_Matrix) states that for CentOS we only support the devicemapper driver.
But the [CentOS install docs](https://docs.docker.com/install/linux/docker-ee/centos/) states that overlay2 is now supported and the preferred driver. Which one is correct?

Also do we have any idea on what version of CentOS you need to have if you're using overlay2+selinux?

---
Here's what the docs are saying
<img width="623" alt="1" src="https://user-images.githubusercontent.com/1525937/36320715-efcde9d8-12fb-11e8-8d04-64f8ae126d6f.png">

gbarr01:
Fixed by https://github.com/docker/docker.github.io/pull/6363

PR #6063, opened 2018-02-23
Remove outdated support statement on overlayfs

cpuguy83:


joaofnfernandes:
@cpuguy83 according to the [compatibility matrix](https://success.docker.com/article/Compatibility_Matrix), you need devicemapper for CentOS and Oracle Linux.

If we're removing that statement, we also need to fix the compat matrix. I feel that we already circled around on this issue a couple of times, so who should we ping to gather some consensus around this?

This relates to #6015

cpuguy83:
@joaofnfernandes https://github.com/docker/docker.github.io/pull/5817/files

PR #6069, opened 2018-02-24
Update runmetrics.md

LydellManganti:
Update docker stats link, it current get "we can't find that page" -> https://docs.docker.com/config/reference/commandline/stats/
Update it to be /engine/reference/commandline/stats.md which should go to https://docs.docker.com/engine/reference/commandline/stats/


gbarr01:
Thank you!

PR #6082, opened 2018-02-26
Fix double-hyphen typographic bug

### LINKED ISSUE #4178: Headers transform two dashes (--) into an en-dash, breaking anchors
### Problem description

Originally reported by @iblancasa  in https://github.com/moby/moby/issues/34209 (thanks!)

The documentation has some "smart transform" enabled, resulting in double dashes in headers to be transformed to an em-dash. As a result, long flags on the command line (which start with `--`) are presented as an en-dash (`–`)

For example, the `--rm` flag is an option with _two_ dashes; the anchor ("fragment") for this _should be_ `#clean-up--rm`, however it looks like there's a configuration issue with the tooling generating the docs, and it replaces `--` with an en-dash (`–`), resulting in `#clean-up-rm` as anchor.

screen shot 2017-08-12 at 13 06 09

This causes both the link in the inline TOC (here: https://docs.docker.com/engine/reference/run/#operator-exclusive-options) to not work, but also is confusing because it now looks like the option is `-rm` (single dash), which is incorrect:

```bash
$ docker run -rm busybox
unknown shorthand flag: 'r' in -rm
See 'docker run --help'.
```

We had similar issues with the old tooling in the past, but could fix that by changing the configuration (see https://github.com/docker/docs-base/issues/66, https://github.com/docker/docs-base/issues/189) which was fixed by updating the configuration https://github.com/docker/docs-base/pull/190

### Problem location

https://docs.docker.com/engine/reference/run/#operator-exclusive-options

### Suggestions for a fix

Change the configuration to not transform double-dashes into an en-dash. For the odd occasion and en-dash _is_ needed, it should just be entered as such :smile:

edit: it's actually an en dash, not an em dash :smile:

mistyhacks:
Signed-off-by: Misty Stanley-Jones <misty@apache.org>

Fixes #4178 

This introduces two changes in the `_config.yml`. One change is a tweak to Kramdown's config, and the other is an update to the version of Github Pages that is used for builds invoked by `jekyll build` (local builds, Netlify, etc). For our actual live deploy, we actually already use the latest version of the Github Pages gem, but don't yet have the tweaked config. The changes to `_config.yml` will need to be backported back to archive branches, possibly all the way back to 17.03, if you care about fixing the typography problem in those.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -35,5 +35,9 @@ source "https://rubygems.org"
 # gem install bundler
 # bundle install
 
-gem "github-pages", "172"
+# This only affects interactive builds (local build, Netlify) and not the
+# live site deploy, which uses the Dockerfiles found in the publish-tools
+# branch.
+
+gem "github-pages", "175"
 gem 'wdm' if Gem.win_platform?
--- a/_config.yml
+++ b/_config.yml
@@ -2,6 +2,7 @@ name: Docker Documentation
 markdown: kramdown
 kramdown:
   input: GFM
+  gfm_quirks: [paragraph_end, no_auto_typographic]
   html_to_native: true
   hard_wrap: false
   syntax_highlighter: rouge


mistyhacks:
@johndmulhausen @gbarr01 @kolyshkin PTAL

kolyshkin:
(Note the `no_auto_typographic` is the option we need. I forgot why there is a need to also add `paragraph_end`, had to look it up now to refresh my memory. The story is, this option is a default value of `gfm_quirks` (i.e. if qfm_quirks is unset, it defaults to paragraph_end), so once `qfm_quirks` is set, `paragraph_end` disappears unless set explicitly. At least that's my understanding of how things work.)

kolyshkin:
Good news is it seems to work! For example, compare https://deploy-preview-6082--docker-docs.netlify.com/engine/reference/commandline/run/#capture-container-id---cidfile (the header looks right, i.e. `--cidfile`) vs https://docs.docker.com/engine/reference/commandline/run/#capture-container-id-cidfile (doesn't look right, i.e. `–cidfile`). I declare this a huge win for humanity!

Bad news is I can't fully check it using netlify, as e.g. https://deploy-preview-6082--docker-docs.netlify.com/engine/reference/run/ gives 404 for whatever reason.

mistyhacks:
That's expected. Netlify doesn't run the scripts that fetch remote resources to be used in the builds.

kolyshkin:
LGTM

kolyshkin:
> For our actual live deploy, we actually already use the latest version of the Github Pages gem

Apparently it's not the case as the image (`docs/docker.github.io:docs-builder`, see https://hub.docker.com/r/docs/docker.github.io/tags/, tag docs-builder) was last updated 2 months ago, which, by looking into https://github.com/Starefossen/docker-github-pages/commits/master, corresponds to v172, while we need v175+). It seems it is needed to trigger a rebuild; @johndmulhausen can you please?

kolyshkin:
In the meantime I filed https://github.com/Starefossen/docker-github-pages/issues/36 so that https://github.com/docker/docker.github.io/blob/docs-base/Dockerfile could use a tag/version.

gbarr01:
@mistyhacks , is this ready for squash and merge?

mistyhacks:
I believe so. I can't do it anymore though.

kolyshkin:
@johndmulhausen can you please trigger a rebuild of docs/docker.github.io:docs-builder as per https://github.com/docker/docker.github.io/pull/6082#issuecomment-368721332? Otherwise https://github.com/docker/docker.github.io/issues/4178 is not yet fixed

joaofnfernandes:
Just leaving a breadcrumb here where you can find more info about the two options now being used with Kramdown - https://kramdown.gettalong.org/parser/gfm.html#options

PR #6083, opened 2018-02-26
Fix broken links to postinstall documentation 

russellcardullo:
### Proposed changes

In the distribution specific instructions for installing docker-ee, the links to the post-installation steps are broken.  For instance, see https://docs.docker.com/install/linux/docker-ee/ubuntu/ halfway through the page at the end of the section `INSTALL DOCKER EE`.  This fixes the broken links to point to the correct location.

N/A

### Related issues (optional)

N/A


joaofnfernandes:
Reusable content should always use absolute paths since you don't know where that snippet is going to be rendered. Thanks for fixing this @russellcardullo. Let's ship it!

PR #6088, opened 2018-02-27
docs-base Dockerfile: fix apk command

kolyshkin:
1. It does not make sense to run apk many times.

2. Use --no-cache flag to avoid doing update, as per https://github.com/gliderlabs/docker-alpine/blob/master/docs/usage.md#disabling-cache


kolyshkin:
The netlify failure does not seem to be related

joaofnfernandes:
Yup. Netlify is failing because this branch doesn't have any files for generating a static website.

PR #6104, opened 2018-02-27
Add point version to ce release variables

gbarr01:
In _config.yml, ce release version variables use major.minor versions only (17.12). Here, I add the point version too (17.12.0) so that we can use a variable in `docker --version`.

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -18,9 +18,9 @@ exclude: ["_scripts", "apidocs/layouts", "Gemfile", "hooks"]
 # Component versions -- address like site.docker_ce_stable_version
 # You can't have - characters in these for non-YAML reasons
 
-docker_ce_stable_version: "17.12"
 latest_stable_docker_engine_api_version: "1.36"
-docker_ce_edge_version: "18.02"
+docker_ce_stable_version: "17.12.0"
+docker_ce_edge_version: "18.02.0"
 docker_ee_version: "17.06"
 compose_version: "1.19.0"
 machine_version: "0.13.0"


joaofnfernandes:
I don't think this is what you want to do.
Let's check where one of those variabels is being used (with a few results removed):

```
 grep -r 'docker_ce_stable_version' .

./_layouts/docs.html:	var dockerVersion = 'v{{ site.docker_ce_stable_version }}';
./_layouts/docs.html:	var dockerVersion = 'v{{ site.docker_ce_stable_version }}';

./docker-for-mac/index.md:Docker version {{ site.docker_ce_stable_version }}, build c97c6d6
./install/linux/docker-ce/centos.md:    docker-ce.x86_64            {{ site.docker_ce_stable_version }}.ce-1.el7.centos             docker-ce-stable
./install/linux/docker-ce/debian.md:    docker-ce | {{ site.docker_ce_stable_version }}.0~ce-0~debian | https://download.docker.com/linux/debian jessie/stable amd64 Packages
./install/linux/docker-ce/fedora.md:    docker-ce.x86_64  {{ site.docker_ce_stable_version }}.0.fc26                              docker-ce-stable
./install/linux/docker-ce/ubuntu.md:    docker-ce | {{ site.docker_ce_stable_version }}.0~ce-0~ubuntu | {{ download-url-base }} xenial/stable amd64 Packages
```

The first two results are the version dropdown, so we don't want to show the path number there
screen shot 2018-02-27 at 14 21 34

In the remaining uses of that variable we use the variable assuming it just contains major.minor, and then hardcode the patch.

This means that for patch releases we should not change the variables in `_config.yml`, but we might have to update those hard-coded patch numbers that are currently `0` to `1`, given that the version we're bumping from `17.12.0->17.12.1`.

Can you ask the release team to do a pass on those `/install/...` pages and tell us if we need to updated those articles?

PR #6105, opened 2018-02-27
Added required line to update TLS version

sethreidnz:
### Proposed changes

Documentation change for Docker Compose on Windows

Github now only supports TLS1.2 and you need to run the following command to make sure these instructions work. Please let me know if my phrasing is not good and I can update to make it work


joaofnfernandes:
@sethreidnz thanks for improving our docs. I think everything looks good, and like that the command only affects the current Powershell session, but I'd like @johnstep thumbs up on this one.

joaofnfernandes:
Haven't heard back, so let's merge it 😄

PR #6113, opened 2018-02-28
Update port explanation

doc75:
It was mention a map of port 80:80 instead of 4000:80 as shown in the command lines

### Proposed changes

Fix a typo from 80 to 4000 so that explanation matches the example

### DIFF
--- a/get-started/part2.md
+++ b/get-started/part2.md
@@ -220,7 +220,7 @@ $ curl http://localhost:4000
 
 This port remapping of `4000:80` is to demonstrate the difference
 between what you `EXPOSE` within the `Dockerfile`, and what you `publish` using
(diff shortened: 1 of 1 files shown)


joaofnfernandes:
@doc75 thanks for your contribution!
I think you misunderstood what that paragraph is trying to tell. It says that right now we're doing "--publish 4000:80`, so you need to access the service using `http://localhost:4000`. Then it says that in other snippets ahead, instead of doing that, we'll be doing `--publish 80:80`, so that readers know that they should use `http://localhost:80` or just `http://localhost`, when they see that.

For that reason I don't think it makes sense to incorporate your change, and I'll close this PR 😢
Feel free to comment if you disagree or if you have any ideas on how to improve this.

PR #6116, opened 2018-02-28
Fix small typo

caervs:
### Proposed changes

Noticed an 'an' which I believe was meant to be an 'and'


joaofnfernandes:
Thanks @caervs!

PR #6124, opened 2018-03-01
DO-NOT-MERGE update the docs for using host-network

rdxmb:
with syntax before, there is
`WARN[0000] network hostnet: network.external.name is deprecated in favor of network.name `

tested with docker-compose-file-version 3.5
Server Version: 17.12.0-ce

in swarm mode

### DIFF
--- a/compose/compose-file/index.md
+++ b/compose/compose-file/index.md
@@ -2066,8 +2066,7 @@ services:
 
 networks:
   hostnet:
-    external:
-      name: host
+    name: host
 ```
 
 ```yaml
@@ -2079,8 +2078,7 @@ services:
 
 networks:
   nonet:
-    external:
-      name: none
+    name: none
 ```
 
 ### driver_opts
--- a/compose/compose-file/index.md
+++ b/compose/compose-file/index.md
@@ -2066,6 +2066,7 @@ services:
 
 networks:
   hostnet:
+    external: true
     name: host
 ```
 
@@ -2078,6 +2079,7 @@ services:
 
 networks:
   nonet:
+    external: true
     name: none
 ```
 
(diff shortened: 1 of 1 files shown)


rdxmb:
Hm, the warning disappears, but without using 'external' the swarm does create a network named
`
<STACK>_host`

which is overlay. So what about the deprecation?

gbarr01:
Hi, @rdxmb  ... let's see if I understand you:

* You are referring to this page/section: https://docs.docker.com/compose/compose-file/#host-or-none
* You're in swarm mode, using Compose, and you want to use the host networking stack. But it's failing. 
* You think that the use of `external` in this section may be wrong because of the deprecation note down the page.

```
networks:
  hostnet:
    external:
      name: host
```

@shin-  are the Compose examples under "[HOST OR NONE](https://docs.docker.com/compose/compose-file/#host-or-none)" accurate for v3.5? For reference: #4593 / #4959

shin-:
The new recommend syntax would be:
```yaml
networks:
  hostnet:
    external: true
    name: host
```

rdxmb:
@gbarr01 

> You are referring to this page/section: https://docs.docker.com/compose/compose-file/#host-or-none
exactly.

> You're in swarm mode, using Compose, and you want to use the host networking stack. But it's failing. You think that the use of external in this section may be wrong because of the deprecation note down the page.

Not exactly. It's not failing, its just a deprecation warning:
`WARN[0000] network hostnet: network.external.name is deprecated in favor of network.name`

networks:
  hostnet:
    external:
      name: host

@shin-   it seems to work. Thanks.

https://docs.docker.com/compose/compose-file/#network-configuration-reference

rdxmb:
PR changed.

PR #6126, opened 2018-03-01
Update _config.yml

gbarr01:
Update assignments


joaofnfernandes:
I manually went to Netlify and the build is passing, so I'll merge this to prevent new issues from pinging John.

PR #6129, opened 2018-03-01
fixing a command's description

jtgasper3:
### Proposed changes

Updated documentation because the command does not create a key, but generates a signed cert based on the CSR.


joaofnfernandes:
Makes sense, so let's ship it 🚀

PR #6133, opened 2018-03-02
Update rails.md

leymannx:
Avoid running `docker-compose run web rake db:create` every time you restart/rebuild the app.

### Proposed changes

I updated the rails tutorial to advice to mount a volume from the host into the DB container. To have the DB synced to the host and then be taken into account when restarting/rebuilding the containers. We should avoid running `docker-compose run web rake db:create` every time the container is restarted/rebuilt. Instead an existing DB should be initialized and the devs continue their development with running `docker-compose run web rake db:migrate` as they need.

### DIFF
--- a/compose/rails.md
+++ b/compose/rails.md
@@ -50,6 +50,8 @@ to link them together and expose the web app's port.
     services:
       db:
         image: postgres
+        volumes:
+          - ./tmp/db:/var/lib/postgresql/data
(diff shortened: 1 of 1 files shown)


joaofnfernandes:
Ye, I think it makes sense to persist the database state using a volume.
Thanks for the contribution, let's ship it!

PR #6138, opened 2018-03-03
Add shutdown order information for depends_on

msgeissler:
### Proposed changes

_depends_on_ also affects shutdown order but the documentation and examples never mention that. A user might jump to the conclusion that it only works when starting containers.


shin-:
Technically accurate!

PR #6140, opened 2018-03-03
Fix typo related to dot

adursun:


gbarr01:
Hi, @adursun ... actually, that's how we punctuate in the  U.S.  From: 

https://www.quickanddirtytips.com/education/grammar/how-to-use-quotation-marks

> Quotation Marks with Commas and Periods
> The most common question people ask about quotation marks is whether periods and commas go inside or outside, and the answer depends on where your audience lives because in American English we always put periods and commas inside quotation marks, but in British English periods and commas can go inside or outside (kind of like the American rules for question marks and exclamation points). I use this memory trick: Inside the US, inside the quotation marks. Here are some examples:
> 
> “Don’t underestimate me,” she said with a disarmingly friendly smile.
> 
> I can never remember how to spell “bureaucracy.”
> 
> Don’t get confused when you see this handled differently in The Economist or on the BBC website; just remember that it’s different in those publications because the British do it differently.

gbarr01:
It's strange, I know. Thank you for  the submission.

PR #6153, opened 2018-03-06
Dockerfile: bump gh-pages to v177

kolyshkin:
This is needed for the following reasons:

1. gh-pages >= 174 is required to fix the issue #4178 (together with https://github.com/docker/docker.github.io/pull/6082)

2. Not using a tag leads to pulling some unknown version of the image, which might or might not work.

3. Not using a tag means the image is never rebuilt.

NOTE: it was not possible to use a tag before, as recent gh-pages releases were not tagged on dockerhub; this is now fixed (see https://github.com/Starefossen/docker-github-pages/issues/36).

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM starefossen/github-pages
+FROM starefossen/github-pages:177
 
 # This is the source for docs/docs-base. Push to that location to ensure that
 # the production image gets your update :)


kolyshkin:
Netlify CI failure is unrelated (there's nothing to deploy in this branch).

kolyshkin:
@gbarr01 @joaofnfernandes PTAL

Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM starefossen/github-pages
+FROM starefossen/github-pages:177
```
thaJeztah: I think this needs to be updated in https://github.com/docker/docker.github.io/tree/publish-tools

joaofnfernandes: If you're bumping the Dockerfile to use GH Pages 177, could you also bump the Gemfile used for local `jekyll build`? - https://github.com/docker/docker.github.io/blob/master/Gemfile

kolyshkin: I admin I don't fully understand how things are working here. This branch [README.md](https://github.com/docker/docker.github.io/blob/docs-base/README.md#reasons-to-update-this-branch) says:
> Reasons to update this branch
> ...
> Updating the GitHub Pages version used in production to a newer release

... but the branch you mentioned also pulls gh-pages. I guess I'll have two PRs and let the more knowledgeable people decide which one to take.

kolyshkin: https://github.com/docker/docker.github.io/pull/6159

kolyshkin: > could you also bump the Gemfile used for local jekyll build? - https://github.com/docker/docker.github.io/blob/master/Gemfile

It's a separate branch so it requires a separate PR. Here: https://github.com/docker/docker.github.io/pull/6167

kolyshkin:
@joaofnfernandes PTAL

gbarr01:
@joaofnfernandes , can we merge this PR and #6167?

gbarr01:
It seems we have a new related issue: #6189

kolyshkin:
> It seems we have a new related issue: #6189

I can't see how some random segfault can be related; left some constructive comments in that issue (and I think you can close it).

PR #6155, opened 2018-03-06
 Call out go version required to get client

RobertKielty:
### Proposed changes
If you have an old version of go installed on your machine the instruction to 
`go get github.com/docker/docker/client` fails. 

ref https://forums.docker.com/t/possible-doc-bug-for-go-client/47374?u=rkielty

I think that go 1.8 is the required minimum based on the following :
* https://forums.docker.com/t/possible-doc-bug-for-go-client/47374/2?u=rkielty Thanks David Maze!
* I have tested this with go 1.6.2 where it failed as described above
* I have tested this with go 1.10 where it succeed

I cannot find a statement where the required minimum go version to get the client is called out
I checked the above doc along with 
* https://godoc.org/github.com/docker/docker/client
* https://github.com/moby/moby/blob/master/client/README.md

### DIFF
--- a/develop/sdk/index.md
+++ b/develop/sdk/index.md
@@ -29,6 +29,9 @@ installed and coexist together.
 ```bash
 go get github.com/docker/docker/client
 ```
+The client requires a recent version of Go. Run `go version` and ensure that you 
+are running at least version 1.8 of Go
+
 
 [Read the full Docker Engine Go SDK reference](https://godoc.org/github.com/docker/docker/client).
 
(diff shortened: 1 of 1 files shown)


RobertKielty:
I found a reference to a golang version in this commit, Bump Golang to 1.9.4

[Bump Golang to 1.9.4 - caeab268430 ](https://github.com/moby/moby/commit/caeab268430a033fedd27c53be16758ac1a0f71e)

So I suppose that given that caeab268430 references a vulnerability being fixed this change should bump to 1.9.4?

L-Hudson:
Sorry this took so very long. Good to go.

PR #6159, opened 2018-03-06
bump gh-pages to v177

kolyshkin:
This is needed for the following reasons:

1. gh-pages >= 174 is required to fix the issue #4178 (together with https://github.com/docker/docker.github.io/pull/6082)

2. Not using a tag leads to pulling some unknown version of the image, which might or might not work.

3. Not using a tag means the image is never rebuilt.

NOTE: using a tag was not possible before, as recent gh-pages releases were not tagged on dockerhub; this is now fixed (see https://github.com/Starefossen/docker-github-pages/issues/36).

While at it, fix apk invocation in Dockerfile.build.


kolyshkin:
@joaofnfernandes @gbarr01 PTAL

Dockerfile.builder
```diff
@@ -5,14 +5,14 @@ RUN go get -d github.com/tdewolff/minify/cmd/minify \
  && go build -v -o /minify github.com/tdewolff/minify/cmd/minify
 
 # Set the version of Github Pages to use for each docs archive
-FROM starefossen/github-pages:latest
+FROM starefossen/github-pages:177
```
thaJeztah: sgtm, but surprised if `:latest` would be older than `:177` 🤔 
possibly the image wasn't rebuilt for a while, so being explicit is good to do

Dockerfile.builder
```diff
@@ -5,14 +5,14 @@ RUN go get -d github.com/tdewolff/minify/cmd/minify \
...
 
 # Get some utilities we need for post-build steps
-RUN apk update && apk add bash wget subversion gzip
+RUN apk add --no-cache bash wget subversion gzip
```
thaJeztah: :heart:

Dockerfile.builder
kolyshkin: It was last rebuilt two months ago, so it's at v173 or so. The problems with using `latest` are listed in the changelog message; in particular it's never going to change => no clean way to trigger an auto-rebuild (surely we can add a comment or fix a typo but that is oh so wrong thing to do).

kolyshkin:
@joaofnfernandes PTAL

PR #6164, opened 2018-03-07
Fix filename check-config.sh

Trejjam:
### Proposed changes

Docs typo
[install/linux/linux-postinstall.md](https://docs.docker.com/install/linux/linux-postinstall/), CC @gbarr01


joaofnfernandes:
Thanks for the contribution @Trejjam 🐳

PR #6175, opened 2018-03-08
add a missing command into summary

venumurthy:
the docker swarm init is missing from the summary of all commands in this section.

### DIFF
--- a/get-started/part3.md
+++ b/get-started/part3.md
@@ -244,6 +244,7 @@ it runs, using the same command that launched the service:
 Some commands to explore at this stage:
 
 ```shell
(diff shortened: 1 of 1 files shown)


joaofnfernandes:
@venumurthy thanks for trying to improve the docs!
That list is not intended to be a comprehensive list of commands that exist in Docker. It's just some commands that would can explore at your own after completing the part 3 of this tutorial.

Right at the beginning of part 3, we've told users to run `docker swarm init`, so I think it doesn't make sense for us to mention that command on the list below.
For that reason I'm closing your pull request, but I'm sure you'll have tons of other opportunities to make our docs better 🐳

PR #6178, opened 2018-03-08
added missing 'end block mark' which made broken block

sio4:
### Proposed changes

There is missing end block mark (```) for one of code blocks in page and it make broken page of reference manual for `scp` subcommand of `docker-machine`.


joaofnfernandes:
Thanks for keeping our docs easy to read!

sio4:
Thanks, Because I need to read it better, :-)

PR #6183, opened 2018-03-09
Update multi-service_container.md

datenhahn:
Negative exitcodes are not allowed, see:

http://tldp.org/LDP/abs/html/exitcodes.html

exit takes only integer args in the range 0 - 255

### DIFF
--- a/config/containers/multi-service_container.md
+++ b/config/containers/multi-service_container.md
@@ -68,7 +68,7 @@ this in a few different ways.
     # If they are not both 0, then something is wrong
     if [ $PROCESS_1_STATUS -ne 0 -o $PROCESS_2_STATUS -ne 0 ]; then
       echo "One of the processes has already exited."
-      exit -1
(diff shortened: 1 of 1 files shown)


joaofnfernandes:
Ohhhh, this is an interesting one @datenhahn 😀 Bash seems to support `exit -1`. Even the article you mention has an example where that is used:

```
bash -c 'exit -1'; echo $?
255
```

Now... I'm not sure when `255` should be returned or why use -1 instead of explicitly 255.
Also, this behavior is specific to Bash (which is OK for the snippet in the docs page). Sh doesn't seem to accept negative exit codes:

```
/bin/sh -c 'exit -1'
exit: line 1: Illegal number: -1
```

@thaJeztah I would like your opinion on whether we should merge this contribution or keep the snippet as is.

joaofnfernandes:
Then let's shiiiiiiiiip it 🚀

PR #6187, opened 2018-03-09
Fix capitalisation of WordPress in 3 places

davejamesmiller:


joaofnfernandes:
Thanks for fixing this 🐳

PR #6188, opened 2018-03-10
Fixed broken dtr_version rendering

omegamormegil:
### Proposed changes

Commands were failing because the DTR version number wasn't rendering. I fixed it on this page by replacing "dtr_version" with "page.dtr_version" in three places.


joaofnfernandes:
Thanks for fixing this!

PR #6193, opened 2018-03-12
Configuration for mysql

omkar-dsd:
Configurations needed for setting up mysql database with django for Docker. Almost similar to postgreSQL, just two small tweaks needed.

### DIFF
--- a/compose/django.md
+++ b/compose/django.md
@@ -79,6 +79,30 @@ and a `docker-compose.yml` file. (You can use either a `.yml` or `.yaml` extensi
         depends_on:
           - db
     ```
+    
+    For `mysql` add the following cofiguration to the file.
+    
+    ```none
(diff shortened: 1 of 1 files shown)


ahh-docker:
@shin- PTAL. Thanks.

shin-:
I'm not sure it's a good idea - this adds bloat to a page that intends to serve as a quick, simple introduction to the setup. The database this page suggests (postgres) is in line with the Django community's consensus on what database works best with Django [1][2]. I would be in favor of leaving the page as is.

[1] https://stackoverflow.com/questions/9540154/which-database-engine-to-choose-for-django-app
[2] https://www.reddit.com/r/django/comments/54luat/what_is_the_best_db_to_use_with_django/

paigehargrave:

PR #6195, opened 2018-03-12
Fix events JSON blob format

benwhite-deltas:
### Proposed changes

The example JSON blob given doesn't match up with the values I experience when running `docker-compose events` myself, so I'm updating the documentation to save someone else some time. In addition, the events are listed in the API reference, so I've added a link to there for convenience.

### DIFF
--- a/compose/reference/events.md
+++ b/compose/reference/events.md
@@ -31,4 +31,4 @@ format:
 }
 ```
 
-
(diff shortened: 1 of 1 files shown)


joaofnfernandes:
Thanks for including the link, it makes total sense! Let's ship it 🚀

PR #6208, opened 2018-03-12
Issue fix #6163

### LINKED ISSUE #6163: Kali (Rolling) Linux Install Fix
[install/linux/docker-ce/debian.md](https://docs.docker.com/install/linux/docker-ce/debian/), CC @gbarr01 

In steps 3 and 4 of "*Set Up the Repository*", Kali Linux users will run into issues because of the auto-generated commands. I circumvented these by

Step 3:
Changing:
`$ curl -fsSL https://download.docker.com/linux/$(. /etc/os-release; echo "$ID")/gpg | sudo apt-key add -` 

to

`$ curl -fsSL https://download.docker.com/linux/debian/gpg | sudo apt-key add -` 

And in Step 4 I used (as super user):
`echo "deb [arch=amd64] https://download.docker.com/linux/debian jessie stable" > /etc/apt/sources.list.d/docker.list `

Just FYI

spikel46:
### Proposed changes

Added additional documentation for non-official debian-based operating systems in install/linux/docker-ce/debian.md

### Related issues (optional)

Fixes #6163

### DIFF
--- a/install/linux/docker-ce/debian.md
+++ b/install/linux/docker-ce/debian.md
@@ -164,7 +164,6 @@ from the repository.
 
     To also add the **edge** repository, add `edge` after `stable` on the last
     line of the command.
-    To add an unsupported operating Debian-based operating system, you will need to know which distribution you have, such as `jessie`.
 
     <ul class="nav nav-tabs">
       <li class="active"><a data-toggle="tab" data-target="#x86_64_repo">x86_64 / amd64</a></li>
--- a/install/linux/docker-ce/debian.md
+++ b/install/linux/docker-ce/debian.md
@@ -94,7 +94,7 @@ from the repository.
 
 #### Set up the repository
 
-{% assign download-url-base = 'https://download.docker.com/linux/$(. /etc/os-release; echo "$ID")' %}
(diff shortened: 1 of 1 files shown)


install/linux/docker-ce/debian.md
```diff
@@ -164,6 +164,7 @@ from the repository.
 
     To also add the **edge** repository, add `edge` after `stable` on the last
     line of the command.
+    To add an unsupported operating Debian-based operating system, you will need to know which distribution you have, such as `jessie`.
```
joaofnfernandes: If a distribution is not supported we should not even talk about it in the docs or provide workarounds. So can you remove this line before we can merge this contribution?

Whenever the docs provide workarounds for unsupported configs, users are confused on whether something is supported or not and will end up deploying an unsupported config. then when something breaks they will be completely disappointed with the product, even though the product never supported that config.

We had situations like this in the past, and I would like to avoid that at all costs.

spikel46:
Yeah absolutely!

joaofnfernandes:
Let's ship it 🚀

thaJeztah:
Thanks for your contribution :+1:

PR #6210, opened 2018-03-13
Jekyll: don't put {% raw %} directives in pre blocks

akimd:
### Proposed changes

Replace all occurrences of

    ```foo
    {% raw %}
    bar
    {% endraw %}
    ```

(which generates spurious empty lines in the rendered pre block) with

    {% raw %}
    ```foo
    bar
    ```
    {% endraw %}

Also, fix some occurrences where the raw section is too large and
prevented interpretation of Jekyll directives.

This is the syntax used in the documentation of Jekyll itself:

https://raw.githubusercontent.com/jekyll/jekyll/master/docs/_docs/templates.md

FTR, done with two perl substitutions:

    '^([\t ]*```[^\n]*
    )([ \t]*\{% raw %\}[^\n]*
    )' '$2$1'

and

    '^([ \t]*\{% endraw %\}[^\n]*
    )([\t ]*```[^\n]*
    )' '$2$1'

and manually tweaks.  A mechanical check would be most useful.

Here is an example of raw sections with spurious empty lines:

capture d ecran 2018-03-12 a 16 37 44


joaofnfernandes:
Thanks for this @akimd, let's ship it!

PR #6221, opened 2018-03-14
Update Registry documentation to use latest version 2.6

vegasbrianc:
Bump the registry version to 2.6

### DIFF
--- a/registry/index.md
+++ b/registry/index.md
@@ -48,7 +48,7 @@ The Registry is compatible with Docker engine **version 1.6.0 or higher**.
 
 Start your registry
 
-    docker run -d -p 5000:5000 --name registry registry:2
+    docker run -d -p 5000:5000 --name registry registry:2.6
 
 Pull (or build) some image from the hub
 


joaofnfernandes:
For the official images, the convention is if you specify
* `M.m.p` - you'll always get that specific patch
* `M.m` - you'll get the latest available patch for that minor release
* `M` - you'll get the latest available `m.p` for that major version
* `latest` - you'll the the latest available `M.m.p`

Since `2.6` is the latest minor release for the `2.x` series, that also means that at least for the time being `registry:2` and `registry:2.6` are exactly the same thing.

```
registry            2                   d1fd7d86a825        2 months ago        33.3MB
registry            2.6                 d1fd7d86a825        2 months ago        33.3MB
```

So I would prefer leaving the documentation as it currently is. That way when/if we ship `2.7` we don't need to update the docs for users to get that new minor version.

For that reason I'm closing this contribution 😭but feel free to comment if you disagree.

vegasbrianc:
@joaofnfernandes Thanks for the update. Somehow I misread 2 and 2.6 thinking they were different sizes. I appreciate the clarification! 👍

joaofnfernandes:
No worries, and thanks for trying to keep the docs up to date!

PR #6226, opened 2018-03-14

khmel-aleh:
### Proposed changes

Without this line, it is not fully clear how the postgres is passed by the first parameter


joaofnfernandes:
Well spotted. Let's ship it 🚀

Issue #6230, opened 2018-03-15
Restart policies differs


dkurzaj:
Hello,

I've found that the **--restart** option can take different values from where you read it in the documentation :

[engine/reference/commandline/run.md](https://docs.docker.com/engine/reference/commandline/run/), CC @gbarr01 

[config/containers/start-containers-automatically.md](https://docs.docker.com/config/containers/start-containers-automatically/), CC @gbarr01 

Or maybe are those pages talking about different things.

Thank you !

Dorian

gbarr01:
Thanks, @dkurzaj .   

@joaofnfernandes, 

This docs page: https://docs.docker.com/engine/reference/commandline/run/#restart-policies---restart

is generated from this cli page: https://github.com/docker/cli/blob/c19b272c09274c18e2a1eeafe4b495d3fecd837f/docs/reference/commandline/run.md#restart-policies---restart

The docs page does not match and I'm trying to figure out why. I just committed a tiny change by adding two newlines ...  if this doesn't work, perhaps you can help me debug.

thaJeztah:
This looks to already have been fixed through https://github.com/docker/cli/pull/909; that fix is included in the upcoming 18.03 release https://github.com/docker/docker-ce/commit/767a8f62271a7a65b635a98e339a773bb01fd881

gbarr01:
This was fixed in:  https://github.com/docker/docker.github.io/pull/6234

Issue #5642, opened 2017-12-30
Docs for run seem to be incorrect


SystemDisc:
[engine/reference/run.md](https://docs.docker.com/engine/reference/run/), CC @mistyhacks

The following line seems to contradict other mentions of `-d` and `--rm` in the docs.

> --rm=false: Automatically remove the container when it exits (incompatible with -d)

thaJeztah:
Ah, yes, this used to be a conflicting option, but this changed in Docker 1.13, when the `--rm` was moved from the client to the daemon (see https://github.com/moby/moby/pull/20848)

gbarr01:
Disregard the reference ... different page

gbarr01:
For some reason, the docs don't update when we generate the yaml from docker-ce.  

docker/cli and  docker/docker-ce match. But not docs: 

https://github.com/docker/cli/blob/master/docs/reference/run.md#clean-up---rm
https://github.com/docker/docker-ce/blob/master/components/cli/docs/reference/run.md#clean-up---rm
https://docs.docker.com/engine/reference/run/#clean-up---rm

mistyhacks:
Unless something has changed, you need to regenerate the YAML files and add them to the right place in the `_data` directory. It's in the docs for releasing. And you don't want them YAML files to be built against `docker-ce` `master` branch, but the branch you're trying to release.

gbarr01:
Thanks,  @mistyhacks , nothing has changed and that's the issue. This doc is not being properly updated during YAML generation (against the proper branch).

thaJeztah:
Oh, this is `reference/run.md`, not `reference/commandline/run.md`; it's a separate file and not auto-generated

gbarr01:
A ha. What's the protocol? Copy/paste from docker-ce?

thaJeztah:
I think it’s pulled in from the docker-ce repo, so we should backport it to the 18.03 branch in that repo (if it’s fixed in docker/cli)

gbarr01:
Source page: https://raw.githubusercontent.com/docker/docker-ce/master/components/cli/docs/reference/run.md

thaJeztah:
This should be fixed by https://github.com/docker/cli/pull/957 (backported to the 18.03 docs through https://github.com/docker/docker-ce/pull/519)

gbarr01:
[...]

PR #6234, opened 2018-03-15
Attempt to fix cmd run page

gbarr01:
The  commandline run page does not mirror the cli commadline run page and it should. 

Added two newlines (to match some of the other files) to see if it works. 

Reference: 
#6230 
#5642 (different page but possibly related)

### DIFF
--- a/engine/reference/commandline/run.md
+++ b/engine/reference/commandline/run.md
@@ -3,6 +3,7 @@ datafolder: engine-cli
 datafile: docker_run
 title: docker run
 ---
+
 <!--
 Sorry, but the contents of this page are automatically generated from
 Docker's source code. If you want to suggest a change to the text that appears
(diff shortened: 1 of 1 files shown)


gbarr01:
@joaofnfernandes , I see no change in the netlify preview ... should I given that this page is generated?

thaJeztah:
These files are coming from the docker/cli repository, and a YAML file is generated from those files, however, I think the source of truth is the https://github.com/docker/docker-ce repository (from which the YAML files are generated)

https://github.com/docker/docker.github.io/issues/6230  looks to already have been fixed through https://github.com/docker/cli/pull/909; that fix is included in the upcoming 18.03 release https://github.com/docker/docker-ce/commit/767a8f62271a7a65b635a98e339a773bb01fd881

I'll open a quick PR for the other one; also noticed an anchor that needs updating

thaJeztah:
Opened https://github.com/docker/cli/pull/957 to fix; we can cherry-pick that into the 18.03 release branch so that it goes out with the 18.03 release

gbarr01:
Thanks,  @thaJeztah , the reference files from cli to docker-ce to docs match now.

PR #6243, opened 2018-03-16
Clarify the webhook's definiton in webhooks.md

cloudmelon:
Here is necessary to clarify the real definition of webhook which is a request with POST action to a specified url ( when this happen, the url going to return the result, in this context, is when we do 'docker push'  to submit a image to registry, it fires the automatically build

### DIFF
--- a/docker-hub/webhooks.md
+++ b/docker-hub/webhooks.md
@@ -13,8 +13,8 @@ Configure webhooks on `https://hub.docker.com/r/<USERNAME>/<REPOSITORY>/~/settin
 
 ![Create Webhook](/docker-hub/images/webhooks.png)
 
-With your webhook, you specify a target URL to POST to. Docker Hub POSTs Docker Hub POSTs
-the URL with the following payload:
+With your webhook, you specify a target URL to POST to. Docker Hub POSTs the URL
+with the following payload:
 
 ```json
 {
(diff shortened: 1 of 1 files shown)


docker-hub/webhooks.md
```diff
@@ -6,14 +6,14 @@ title: Webhooks for automated builds
...
-repository. Docker Hub webhooks fire when an image is built in, or a new tag
+repository. Webhook is a POST request sent to a defined URL which provides the service. Docker Hub webhooks fire when an image is built in, or a new tag
 is added to, your automated build repository.
 
 Configure webhooks on `https://hub.docker.com/r/<USERNAME>/<REPOSITORY>/~/settings/webhooks/`.
 
 ![Create Webhook](/docker-hub/images/webhooks.png)
 
-With your webhook, you specify a target URL to POST to. Docker Hub POSTs
+With your webhook, you specify a target URL to POST to. URL that accepts data from the Webhook, stores this POST request in JSON or XML format. Docker Hub POSTs
```
joaofnfernandes: With the change you've made to lines 7-9 this whole line feels a bit redundant, so I think we can remove the whole thing and just leave something like "Docker Hub sends a POST to your URL, containing the following payload:"

What do you think?

cloudmelon:
Sounds great. Thank you for your suggestions. It seems clear now. 

Melony Qin

joaofnfernandes:
@cloudmelon I was asking if you can implement my suggestion, otherwise I don't think we should integrate your change 😭

cloudmelon:
hi Joao, sorry I had a long work day, I just got back home. I updated the
doc with you suggest and I committed it. Thank you very much for your
suggestion :)

joaofnfernandes:
Thanks for your contribution! You had a duplicate sentence, so I've fixed it for you so that we can ship this 😄

PR #6245, opened 2018-03-16
Contribution link, hyper-v clarify, bootcamp

valgaze:
### Proposed changes

- Fix contribution link on main readme (broken)

- Clarify Hyper-V & Virtualization are different concepts, call out Windows10 Pro, provide alternative

- Add obscure Windows on Mac Bootcamp procedure to enable virtualization

### DIFF
--- a/docker-for-windows/troubleshoot.md
+++ b/docker-for-windows/troubleshoot.md
@@ -267,6 +267,15 @@ script](https://github.com/moby/moby/issues/24388).
 
 ### Virtualization
 
+In order for Docker for Windows to function properly your machine needs:
+
+1. [Hyper-V](https://docs.microsoft.com/en-us/windows-server/virtualization/hyper-v/hyper-v-technology-overview) installed and working
+
+2. Virtualization enabled
+
+Note: If your OS is lower than Windows 10 PRO, you can use [docker-toolbox](https://docs.docker.com/toolbox/toolbox_install_windows/)
+
+
 #### Hyper-V
 
 Docker for Windows requires a Hyper-V as well as the Hyper-V Module for Windows
@@ -281,8 +290,8 @@ Virtualization needs to be enabled in the BIOS. The steps to do so are Vendor
 specific, but typically the BIOS option is called `Virtualization Technology
 (VTx)` or similar.
 
-Once Hyper-V is enabled, it shows up as such on "Turn Windows features on or
-off".
+From the START menu enter in "Turn Windows features on or
+off" and hit enter and verify Hyper-V is enabled and has a checkmark:
 
 ![Hyper-V on Windows features](images/hyper-v-enable-status.png)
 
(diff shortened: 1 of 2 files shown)


docker-for-windows/troubleshoot.md
```diff
@@ -267,6 +267,15 @@ script](https://github.com/moby/moby/issues/24388).
 
 ### Virtualization
 
+In order for Docker for Windows to function properly your machine needs:
+
+1. [Hyper-V](https://docs.microsoft.com/en-us/windows-server/virtualization/hyper-v/hyper-v-technology-overview) installed and working
+
+2. Virtualization enabled
+
+Note: If your OS is lower than Windows 10 PRO, you can use [docker-toolbox](https://docs.docker.com/toolbox/toolbox_install_windows/)
```
joaofnfernandes: I don't think we should include this note here. If I were a new user I would think that this is a way for me to install Docker for Windows if I have an version lower than 10 Pro. But that is not the case. Toolbox is a completely different tech stack than Docker for Windows.

Also, [the Docker for Windows install page](https://docs.docker.com/docker-for-windows/install/#what-to-know-before-you-install) already explains that if you don't have the minimum requirements, you can use Docker Toolbox instead, and I think that's the right place for this info.

docker-for-windows/troubleshoot.md
```diff
@@ -281,8 +290,8 @@ Virtualization needs to be enabled in the BIOS. The steps to do so are Vendor
 specific, but typically the BIOS option is called `Virtualization Technology
 (VTx)` or similar.
 
-Once Hyper-V is enabled, it shows up as such on "Turn Windows features on or
-off".
+From the START menu enter in "Turn Windows features on or
```
joaofnfernandes: Replace "START" with "Start".
You also have too many "and" in this sentence. Can you update it to use some commas instead?

docker-for-windows/troubleshoot.md
```diff
@@ -316,6 +325,12 @@ Also, see the user-reported issue [Unable to run Docker for Windows on Windows
 10 Enterprise](https://github.com/docker/for-win/issues/74).
 
 
+### Virtualization on Windows Bootcamp (Mac)
```
joaofnfernandes: This section basically explains how to use macOS Bootcamp. I think we should not include this in our docs, since they should focus on what's specific to Docker, and this information can be found anywhere else.

It might seem that you're helping users but down the line adding non-critical information to this doc will make it harder for users to troubleshoot when they really have problems.

docker-for-windows/troubleshoot.md
```diff
@@ -267,6 +267,15 @@ script](https://github.com/moby/moby/issues/24388).
 
 ### Virtualization
 
+In order for Docker for Windows to function properly your machine needs:
+
+1. [Hyper-V](https://docs.microsoft.com/en-us/windows-server/virtualization/hyper-v/hyper-v-technology-overview) installed and working
```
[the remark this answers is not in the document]
valgaze: I'm presuming full stop means period-- I'm going to ignore since these are not proper sentences requiring a period

docker-for-windows/troubleshoot.md
valgaze: That's a good point, that info can live elsewhere. Removed

docker-for-windows/troubleshoot.md
valgaze: Sure

docker-for-windows/troubleshoot.md
valgaze: I was solving my own use-case problem here & figured it'd be useful for others-- definitely see your point, this doc is already a bit unwieldy as is. I'll remove and stuck folks can search around for this slightly esoteric Windows-on-Mac-running-bootcamp scenario =)

Filed an issue on bootcamp+docker-for-windows specifically to get the issue off of this pr: https://github.com/docker/docker.github.io/issues/6465

valgaze:
@joaofnfernandes Changes per remarks & rebased to master

PR #6246, opened 2018-03-16
Add docker ps for old Docker versions

gbarr01:
- Remove prompt signs
- Add numbering
- Tidy text

### DIFF
--- a/get-started/index.md
+++ b/get-started/index.md
@@ -131,7 +131,8 @@ and higher.
 Ensure that you have a supported version of Docker:
 
 ```shell
-$ docker --version
+docker --version
+
 Docker version 17.12.0-ce, build c97c6d6
 ```
 
@@ -139,7 +140,8 @@ Run `docker version`(without `--`) or `docker info` to view even more details
 about your docker installation:
 
 ```shell
-$ docker info
+docker info
+
 Containers: 0
  Running: 0
  Paused: 0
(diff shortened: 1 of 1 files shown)


get-started/index.md
```diff
@@ -175,19 +177,22 @@ This message shows that your installation appears to be working correctly.
...
 
 ```shell
-$ docker container ls --all
+docker container ls --all
+
 CONTAINER ID     IMAGE           COMMAND      CREATED            STATUS
 54f4984ed6a8     hello-world     "/hello"     20 seconds ago     Exited (0) 19 seconds ago
 ```
 
+> If you have docker-ce 1.12 or lower, run `docker ps --all`. For commands introduced in docker-ce 1.13, see [Introducing Docker 1.13](https://blog.docker.com/2017/01/whats-new-in-docker-1-13/){: target="_blank" class="_"}.
```
joaofnfernandes: I'm OK with this note, but I think a good assumption to make is that users going through the getting started are just starting with Docker. That means we can assume they have the latest stable version of Docker, and we don't have to put in a bunch of notes explaining things that are relevant for users running an older version.

gbarr01: That was my thinking ... I have an idea.

gbarr01:
Actually, @joaofnfernandes , I agree. Version 1.12 is far too old and we make it clear that a supported version should be used. I have other small changes however that I would like to incorporate.

joaofnfernandes:
I'm 👍 on all the other changes here.

PR #6260, opened 2018-03-19
Bump Engine, UCP, DTR

joaofnfernandes:
This change updates the release notes and version variables to release:
* EE Engine 17.06.2-ee-7
* UCP 2.2.6
* DTR 2.4.3

It seems that I can't add a couple of folks as reviewers. We'll update this in the future, but for the time being, /ping @andrewhsu @jose-bigio @corbin-coleman @ry4nz @RobbKistler @gbarr01

### DIFF
--- a/datacenter/dtr/2.4/guides/release-notes.md
+++ b/datacenter/dtr/2.4/guides/release-notes.md
@@ -11,9 +11,13 @@ known issues for each DTR version.
 You can then use [the upgrade instructions](admin/upgrade.md),
 to upgrade your installation to the latest release.
 
-## Version 2.4.2
+## Version 2.4.3 (2018-03-19)
 
-(13 February 2018)
+**Security**
+
+* Dependencies updated to consume upstream CVE patches.
+
+## Version 2.4.2 (13 February 2018)
 
 **Security notice**
 
@@ -29,9 +33,7 @@ potentially disclosed due to the vulnerability.
 Use the `--log-driver=none` option for `docker run` when running a DTR backup, HA
 cluster join or dumpcerts.
 
-## 2.4.1
-
-(20 November 2017)
+## 2.4.1 (20 November 2017)
 
 **Bug fixes**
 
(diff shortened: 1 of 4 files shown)


datacenter/ucp/2.2/guides/release-notes.md
```diff
@@ -10,6 +10,60 @@ known issues for the latest UCP version.
 You can then use [the upgrade instructions](admin/install/upgrade.md), to
 upgrade your installation to the latest release.
 
+## Version 2.2.6 (2018-03-19)
```
ry4nz: in past releases, we put the date in a separate line, can you change it for consistency?

joaofnfernandes: UCP and DTR were the exception rather than the rule. All other release notes were using this format, so let's standardize on that.

I've updated the release notes for 2.2.x to start using that format.

datacenter/dtr/2.4/guides/release-notes.md
```diff
@@ -11,9 +11,13 @@ known issues for each DTR version.
 You can then use [the upgrade instructions](admin/upgrade.md),
 to upgrade your installation to the latest release.
 
-## Version 2.4.2
+## Version 2.4.3 (2018-03-19)
```
gbarr01: Why the change in date format?

joaofnfernandes:
@JimGalasyn you're right, but let's fix that upstream.
@andrewhsu @jose-bigio @corbin-coleman Moving forward we should end each change log entry with a full stop.

As an example:

```
* Create labels when volume exists only remotely. [moby/moby#34896](https://github.com/moby/moby/pull/34896)
```

datacenter/dtr/2.4/guides/release-notes.md
joaofnfernandes: The majority of components (Engine, compose, ...) were already using the format
```
## version (date)
```

UCP and DTR were among a few components that were using a different format. So we're standardizing on this format and moving forward all components should follow the same convention.

PR #6271, opened 2018-03-20
Add using of PGPASSWORD for psql

KenanBek:
Given example is not complete at least for official postgres image. I have added setting of psql password over POSTGRES_PASSWORD env variable.


joaofnfernandes:
Makes sense. Let's ship it 🚀

PR #6273, opened 2018-03-21
Update part2.md

xubaifuCode:


joaofnfernandes:
@xubaifuCode Can you explain the intent of your change?
At that point in the tutorial you actually have built an image, but called it `friendlyhello`. So now we're teaching people how to re-tag it so that it can be pushed to Docker Hub.

Because of that I don't think your change makes sense, but would like to hear your thoughts.

joaofnfernandes:
I'm closing this contribution due to inactivity, but feel free to comment.

PR #6280, opened 2018-03-22
Fix highlighted line in GitHub link to storage driver selection

TheLastProject:
### Proposed changes

The GitHub link for the storage driver selection order links to a few lines below the actual line that should be referred to. This pull request makes GitHub highlight the line that actually contains the storage driver selection order as intended by the documentation.

As the link will always refer to the location for the latest CE stable, this may break again in a newer stable release, but at least this fixes it for now.

### DIFF
--- a/storage/storagedriver/select-storage-driver.md
+++ b/storage/storagedriver/select-storage-driver.md
@@ -41,7 +41,7 @@ this decision, there are three high-level factors to consider:
 
   The selection order is defined in Docker's source code. You can see the order
   by looking at
-  [the source code for Docker CE {{ site.docker_ce_stable_version }}](https://github.com/docker/docker-ce/blob/{{ site.docker_ce_stable_version }}/components/engine/daemon/graphdriver/driver_linux.go#L53)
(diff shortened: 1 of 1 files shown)


storage/storagedriver/select-storage-driver.md
```diff
@@ -41,7 +41,7 @@ this decision, there are three high-level factors to consider:
 
   The selection order is defined in Docker's source code. You can see the order
   by looking at
-  [the source code for Docker CE {{ site.docker_ce_stable_version }}](https://github.com/docker/docker-ce/blob/{{ site.docker_ce_stable_version }}/components/engine/daemon/graphdriver/driver_linux.go#L54-L63)
+  [the source code for Docker CE {{ site.docker_ce_stable_version }}](https://github.com/docker/docker-ce/blob/{{ site.docker_ce_stable_version }}/components/engine/daemon/graphdriver/driver_linux.go#L53)
```
gbarr01: Thanks, @TheLastProject ... actually, it looks like it should be L50. Do you agree? 

https://github.com/docker/docker-ce/blob/18.03/components/engine/daemon/graphdriver/driver_linux.go#L50

gbarr01: Thanks. Yes, we probably need to change this. Anyway. LGTM.

PR #6287, opened 2018-03-22
Fix docs for "unless-stopped" restart policy

adw1n:
### Proposed changes
This part:
>or Docker itself is stopped or restarted.

is not accurate. When you restart docker daemon containers with `unless-stopped` policy are also restarted, as long as they haven't been explicitly stopped before the docker daemon restart.

Please read correct examples of documentation for `unless-stopped` restart policy:
* https://github.com/moby/moby/blob/766d9edf39b79e11018bb0f55056b5bd915ddf3c/api/swagger.yaml#L314
* https://github.com/moby/moby/blob/766d9edf39b79e11018bb0f55056b5bd915ddf3c/api/types/container/host_config.go#L279

If you are still in doubt please read the MR that added `unless-stopped` restart policy:
https://github.com/moby/moby/pull/15348
and the inital ticket for it:
https://github.com/moby/moby/issues/11008

### DIFF
--- a/config/containers/start-containers-automatically.md
+++ b/config/containers/start-containers-automatically.md
@@ -28,7 +28,7 @@ any of the following:
 |:-----------------|:------------------------------------------------------------------------------------------------|
 | `no`             | Do not automatically restart the container. (the default)                                       |
 | `on-failure`     | Restart the container if it exits due to an error, which manifests as a non-zero exit code.     |
-| `unless-stopped` | Restart the container unless it is explicitly stopped or Docker itself is stopped or restarted. |
+| `unless-stopped` | Restart the container unless it is explicitly stopped.                                          |
 | `always`         | Always restart the container if it stops.                                                       |
 
 The following example starts a Redis container and configures it to always


gbarr01:
Hang on. @thaJeztah .... the original text came directly from the cli and docker-ce docs: 

https://github.com/docker/cli/blob/master/docs/reference/commandline/run.md
https://github.com/docker/docker-ce/blob/18.03/components/cli/docs/reference/commandline/run.md

> Restart the container unless it is explicitly stopped or Docker itself is stopped or restarted.

adw1n:
@gbarr01 this one: https://github.com/docker/cli/blob/master/docs/reference/run.md#restart-policies---restart is correct
>Always restart the container regardless of the exit status, but do not start it on daemon startup if the container has been put to a stopped state before.

gbarr01:
@adw1n, actually, this seems to be the correct doc:

https://github.com/docker/cli/blob/master/docs/reference/commandline/run.md#restart-policies---restart

The commit in `reference/commandline/run.md` was much later than `reference/run.md`:

https://github.com/docker/cli/commit/df39b014
https://github.com/docker/cli/commit/d281b72a
https://github.com/docker/cli/commit/e9ce688d

The question is, why do they differ? The restart policies in `reference/run.md` should match those in `reference/commandline/run.md`.

cc: @thajeztah

gbarr01:
@adw1n , I'm going to revert this commit. If I'm wrong, I can reapply. But I think the real problem is that the 2 CLI docs need to be in sync. Sorry for the confusion and thanks for the PR.

PR #6289, opened 2018-03-23
Fix "Run the registry as a service" link

ntzm:


joaofnfernandes:
Thanks for fixing this link!

PR #6294, opened 2018-03-23
Use printf instead of echo

dilipkumar2k6:
`echo` adds a newline feed for the given input, therefore, it doesn't work.

### DIFF
--- a/engine/swarm/secrets.md
+++ b/engine/swarm/secrets.md
@@ -161,7 +161,7 @@ real-world example, continue to
     secret from, is set to `-`.
 
     ```bash
-    $ echo "This is a secret" | docker secret create my_secret_data -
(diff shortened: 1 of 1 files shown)


joaofnfernandes:
Well caught @dilipkumar2k6, this is the kind of thing that will drive people crazy trying to understand where the problem is with their password.

@thaJeztah do you have any preference between `printf` of `echo -n` to make sure the secret is created without a trailing `\n`?

Once we merge this we should also do a follow-up PR to update the [CLI docs](https://github.com/docker/cli/blob/master/docs/reference/commandline/secret_create.md), to make sure they are consistent.

joaofnfernandes:
Let's merge this since others can benefit from this change right away. I've opened docker/cli#979 so that we can also fix this in the CLI reference docs.

PR #6296, opened 2018-03-24
Copyedit phase 1

gbarr01:
Clean up the page, phase 1. 

cc: @akimd


docker-for-windows/index.md
```diff
@@ -12,32 +12,29 @@ redirect_from:
...
 
 ```none
-PS C:\Users\Docker> docker --version
-Docker version 17.03.0-ce, build 60ccb22
+PS C:\Users\Gordon> docker --version
+Docker version 17.12.0-ce, build c97c6d6
 
-PS C:\Users\Docker> docker-compose --version
-docker-compose version 1.11.2, build dfed245
+PS C:\Users\Gordon> docker-compose --version
```
akimd: I would argue that it is not important to show the version of all the components.  I would leave only `docker version` or `docker --version`.

docker-for-windows/index.md
```diff
@@ -48,148 +45,78 @@ and make sure `docker` commands are working properly.
...
-    Version:      17.03.0-ce
-    API version:  1.26 (minimum version 1.12)
-    Go version:   go1.7.5
-    Git commit:   3a232c8
-    Built:        Tue Feb 28 07:52:04 2017
-    OS/Arch:      linux/amd64
-    Experimental: true
-    ```
-
-    Here is an example of command output for `docker info`.
```
akimd: Good thing you got rid of this.

docker-for-windows/index.md
```diff
@@ -48,148 +45,78 @@ and make sure `docker` commands are working properly.
...
-    PS C:\Users\jdoe> docker run -it ubuntu bash
-
-    Unable to find image 'ubuntu:latest' locally
-    latest: Pulling from library/ubuntu
-    5a132a7e7af1: Pull complete
-    fd2731e4c50c: Pull complete
-    28a2f68d1120: Pull complete
-    a3ed95caeb02: Pull complete
-    Digest: sha256:4e85ebe01d056b43955250bbac22bdb8734271122e3c78d21e55ee235fc6802d
+    PS C:\Users\Gordon> docker run -it ubuntu bash
```
akimd: I would add `--rm` `docker run --rm -it ...`.  Unless of course you depend on it for the removal of the containers below.

docker-for-windows/index.md
```diff
@@ -48,148 +45,78 @@ and make sure `docker` commands are working properly.
...
-    PS C:\Users\jdoe> docker run -d -p 80:80 --name webserver nginx
+    PS C:\Users\Gordon> docker container ls --all
+    CONTAINER ID      IMAGE            COMMAND       CREATED            STATUS                        PORTS      NAMES
+    64b5205999ed      ubuntu           "bash"        2 minutes ago      Up 2 minutes                             laughing_kowalevski
+    45f77eb48e78      hello-world      "/hello"      3 minutes ago      Exited (0) 3 minutes ago                 relaxed_sammet
     ```
 
-    This downloads the `nginx` container image and start it. Here is the
-    output of running this command in a powershell.
+6.  In the first PowerShell tab, press `Ctrl+c` to stop the Ubuntu conatiner.
```
akimd: That really works? (Ctrl-c)?  I would recommend typing `exit` instead.

And container, not conatiner.  We should have spell-checking phases in the checks.

docker-for-windows/index.md
```diff
@@ -203,98 +130,58 @@ and make sure `docker` commands are working properly.
...
-    webserver, type: `docker container stop webserver` and start it again with `docker
-    start webserver`.
-
-    To stop and remove the running container with a single command, type:
-    `docker container rm -f webserver`. This removes the container, but not the
-    `nginx` image. You can list local images with `docker image ls`. You might
-    want to keep some images around so that you don't need to pull them again
-    from Docker Hub. To remove an image you no longer need, use `docker image rm` followed by an image ID or image name. For example, `docker image rm nginx`.
-
-**Want more example applications?** [Get Started](/get-started/) and [Samples](/samples) are great places to start.
```
akimd: No more samples?

docker-for-windows/index.md
```diff
@@ -203,98 +130,58 @@ and make sure `docker` commands are working properly.
...
-  right-click an empty portion of the taskbar, select Settings, and choose
-  display options through taskbar settings for notifications.
->
-> ![whale pinned to taskbar](images/whale-taskbar.png)
+![Showing hidden apps in the taskbar](images/whale-icon-systray.png){:width="300px"}
 
-![Showing hidden apps in the taskbar](images/whale-systray.png)
+The Settings dialog provides options to allow Docker auto-start, automatically
+check for updates, share local drives with Docker containers, enable VPN
+compatibility, manage CPUs and memory Docker uses, restart Docker, or perform a
```
akimd: I would have written CPU, but I am not a native.

docker-for-windows/index.md
```diff
@@ -307,75 +194,58 @@ access the mounted volumes.
...
 
 To share the drive, allow connections between the Windows host machine and the
 virtual machine in Windows Firewall or your third party firewall software. You
 do not need to open port 445 on any other network. By default, allow connections
-to 10.0.75.1 port 445 (the Windows host) from 10.0.75.2 (the virtual machine).
-If the firewall rules appear to be open, consider [reinstalling the File and
-Print Sharing service on the virtual network
-adapter.](http://stackoverflow.com/questions/42203488/settings-to-windows-firewall-to-allow-docker-for-windows-to-share-drive/43904051#43904051)
+to `10.0.75.1` port 445 (the Windows host) from `10.0.75.2` (the virtual machine).
+If the firewall rules appear to be open, consider [reinstalling the File and Print Sharing service on the virtual network adapter](http://stackoverflow.com/questions/42203488/settings-to-windows-firewall-to-allow-docker-for-windows-to-share-drive/43904051#43904051).
```
akimd: You don't need to keep the title, you can just stick to `https://stackoverflow.com/questions/43904051`.

akimd: Or `https://stackoverflow.com/a/43904051/1353549`, which is what you get if you click `Share` on that answer.

docker-for-windows/index.md
```diff
@@ -468,23 +340,25 @@ If `Experimental` is `true`, then Docker is running in experimental mode, as
 shown here. (If `false`, Experimental mode is off.)
 
 ```bash
-PS C:\Users\Vicky> docker version
```
akimd: That's a lot of cruft to maintain between releases.  I would rather recommend
```
docker info -f '{{.ExperimentalBuild}}'
```
(or maybe single quotes is really PowerShell does not support single quotes).

docker-for-windows/index.md
```diff
@@ -493,23 +367,21 @@ You can set up your own [registries](/registry/introduction.md) on the **Basic**
 Daemon settings.
 
 As an alternative to using [Docker Hub](https://hub.docker.com/) to store your
-public or private images or [Docker Trusted
-Registry](/datacenter/dtr/2.1/guides/index.md), you can use Docker to set up
-your own insecure [registry](/registry/introduction.md). Add URLs for insecure
-registries and registry mirrors on which to host your images. (See also, [How do
-I add custom CA certificates?](faqs.md#how-do-i-add-custom-ca-certificates) and [How do I add client certificates?](faqs.md#how-do-i-add-client-certificates) in
-the FAQs.)
+public or private images or [Docker Trusted Registry](/datacenter/dtr/2.1/guides/index.md),
```
akimd: Is it ok to pin a version number?  Shouldn't we rather point to the latest doc?

docker-for-windows/index.md
```diff
@@ -307,75 +194,58 @@ access the mounted volumes.
...
-the MobyVM, or use a [data
-volume](/engine/tutorials/dockervolumes.md#data-volumes)
-(named volume) or [data
-container](/engine/tutorials/dockervolumes.md#creating-and-mounting-a-data-volume-container).
-There are a number of issues with using host-mounted volumes and network paths
-for database files. see the troubleshooting topic on [Volume mounts from
-host paths use a nobrl option to override database
-locking](troubleshoot.md#volume-mounts-from-host-paths-use-a-nobrl-option-to-override-database-locking).
+ * If possible, avoid volume mounts from the Windows host, and instead mount on
+ the MobyVM, or use a [data volume](/engine/tutorials/dockervolumes.md#data-volumes)
```
akimd: I think we should avoid such a name (MobyVM), and keep a single way to denote it.  Below, there is "Linux VM", and I think that's better.  Or just "VM".

docker-for-windows/index.md
```diff
@@ -593,11 +464,9 @@ because they do not apply to Windows containers:
...
 ### Kubernetes
 
-**Kubernetes is only available in Docker for Windows 18.02 CE Edge.** Kubernetes
-support is not included in Docker for Windows 18.02 CE Stable.
-
-To find out more about Stable and Edge channels and how to switch between them,
-see [General configuration](#general).
+[Kubernetes on Docker for Windows](/docker-for-windows/kubernetes/){: target="_blank" class="_"}
+is available in
+[18.02 Edge (win50)](/docker-for-windows/edge-release-notes/#docker-community-edition-18020-ce-rc1-win50-2018-01-26){: target="_blank" class="_"} and higher.
```
akimd: Our anchors are really bad, we should add simpler links in there and use them.  We could for instance have simply `#win50`.

docker-for-windows/index.md
```diff
@@ -793,17 +655,14 @@ Check out these [Docker Cloud topics](/docker-cloud/index.md) to learn more:
...
 ## Where to go next
 
-* Try out the walkthrough at [Get Started](/get-started/).
+* Try out the walkthrough at [Get Started](/get-started/){: target="_blank" class="_"}.
 
 * Dig in deeper with [Docker Labs](https://github.com/docker/labs/) example walkthroughs and source code.
 
-* For a summary of Docker command line interface (CLI) commands,
-see the [Docker CLI Reference Guide](/engine/reference/commandline/docker.md).
+* For a summary of Docker command line interface (CLI) commands, see [Docker CLI Reference Guide](/engine/api.md){: target="_blank" class="_"}.
```
akimd: If we don't depend on the abbreviation CLI, we should not introduce it.

docker-for-windows/index.md
```diff
@@ -793,17 +655,14 @@ Check out these [Docker Cloud topics](/docker-cloud/index.md) to learn more:
...
 
-* Check out the blog post
-  [Introducing Docker 1.13.0](https://blog.docker.com/2017/01/whats-new-in-docker-1-13/).
+* Check out the blog post, [What’s New in Docker 17.06 Community Edition (CE)](https://blog.docker.com/2017/07/whats-new-docker-17-06-community-edition-ce/){: target="_blank" class="_"}.
```
akimd: That's the latest?  We should have a constant link to the latest version, it's too much work to update IMHO.

thaJeztah:
Had a quick look at the images, and looks like some space can be saved by losslessly optimizing them;

screen shot 2018-03-26 at 11 13 42

I also see that some of the images are no longer referenced, so can be deleted;

- I see there's an existing `kubernetes/kubernetes-enable.png` - perhaps we can use that one instead or replace it with the new screenshot
- Looks like `settings-proxy.png` replaces `proxies.png` - could you remove the old one? (but check if it's used in other locations)
- `config-menu-solo.png` no longer seems to be used
- Should `docker-cloud-menu.png` (and the Docker Cloud paragraph) be removed? (probably for another PR, as there's more mentions in the docs)
- `docker-daemon_advanced.png` is no longer used (looks to be replaced by `settings-daemon-advanced.png`) so can be removed
- `docker-daemon_basic` is no longer used (looks to be replaced by `settings-daemon-basic.png`) so can be removed
- `whale-systray.png` is no longer used so can be removed


I create optimised versions for the images that are added in this PR; if you can replace them and amend your commit (so that the non-optimised versions don't get into git) 👍 

[optimized-this-pr.zip](https://github.com/docker/docker.github.io/files/1847197/optimized-this-pr.zip)

Also optimised versions of the other images, but some of those can be removed altogether (as mentioned above);

[optimized-other.zip](https://github.com/docker/docker.github.io/files/1847199/optimized-other.zip)

akimd:
@thaJeztah All this is really good, but it should be addressed by tools.  I was also amazed to see that we don't check link, not even inner links.  I had a typo in one link, and it's only because I checked the page that the error was caught.

BTW, I would be happy if the repo kept the original images, and the web site used the compressed ones.  I feel uncomfortable throwing away the originals.

Also, again, Mac and Windows should be kept in sync.

thaJeztah:
> BTW, I would be happy if the repo kept the original images, and the web site used the compressed ones. I feel uncomfortable throwing away the originals.

This is lossless optimisation, I think it's good to do this as a step _before_ committing to the repository to prevent it from ballooning. This repository is already huge, so I don't see a lot of benefit in making it even bigger.

w.r.t. link-checking, yes, I agree; we had this in the past (when the docs were using Hugo), but I think checks were removed when we switched to using Jekyll. 

I think link-generation is currently broken in general as well, and all links are actually rewritten with JavaScript (e.g. view this page with javascript disabled https://docs.docker.com/develop/, and see that all links point to URLs ending in  `.md`)

akimd:
Ah, l had not realized in was lossless.  💯

docker-for-windows/index.md
gbarr01: @akimd , my terminal does not return the prompt ... not sure why not.

docker-for-windows/index.md
gbarr01: less is more. agree.

docker-for-windows/index.md
gbarr01: I agree ... but that's how those docs are being designed. It causes problems with linking, I know.

docker-for-windows/index.md
gbarr01: wilco

docker-for-windows/index.md
gbarr01: agree

docker-for-windows/index.md
gbarr01: agree... like the Quote syntax at the top. We never use that syntax on this page. We should have an example, or put it somewhere else.

docker-for-windows/index.md
gbarr01: Let's remove it.

gbarr01:
Thanks, @akimd , @thaJeztah ! I agree 100%. That's why I called it phase 1. Not only can we get rid of unused images, but I also think we don't need _that_ many images in  general.  

@akimd ... I'm wondering, given your comment about sharing the mac/win pages, maybe we can start from scratch ... use %includes% for the shared text. It would be faster than editing these pages.

akimd:
I dunno.  I don't want either to dive into a refactoring hell.  I'm more inclined to do that incrementally.  But if you feel you have enough resources to go from scratch, then sure!

thaJeztah:
Yes, we can remove the unused images in a follow-up, but it would help to update the new images with their optimised version in this PR (i.e., update the commit). The problem with doing it _afterwards_ is that it may make the images smaller, but the old version would still be in the Git history, so the repository as a while would still contain those bytes.

docker-for-windows/index.md
gbarr01: wilco

docker-for-windows/index.md
gbarr01: Just wondering ... @akimd, perhaps we should call it what it is ... `MobyLinuxVM`? The Internets seems to use that.

docker-for-windows/index.md
gbarr01: It seems that you cannot run an interactive session with PowerShell? The "Command Prompt" terminal works. But you cannot stop or remove all containers in Command Prompt (but PowerShell works). 

How do I run an interactive session with PowerShell? Do I need to do something different? 

cc: @akimd @thaJeztah

docker-for-windows/index.md
akimd: Hi Gwen.  I've been discussing with the other guys from d4d, and the current status is:
- MobyVM is really a misnomer.  The name Moby was actually taken from it (which caused
  a lot of confusion) for some other project.  So we should definitely stay away from it.
- If we were to be precise, today it's LinuxKit VM.  But then again, in the future things
  might change, and we prefer to stick to the name of the feature/service (it's a Linux VM) rather
  than the name of the implementation.
- That being said, on Windows the name of the VM is exposed to the users when they
  deal with HyperV.  And to avoid making upgrades harder on the users, the original name
  was kept (MobyLinuxVM).  We might want to move away from this in the future.
  But only die hard users might see it.

So I suggest that we stick to Linux VM on both architectures.

gbarr01: Thanks for the backstory, @akimd. Very helpful.  I ended up using Linux VM in the end.

gbarr01:
@thaJeztah , thanks again for taking the time to compress and clean the images. We got rid of many.

docker-for-windows/index.md
gbarr01: Solved. I was running PowerShell ISE, not PowerShell: https://github.com/docker/for-win/issues/223

PR #6300, opened 2018-03-25
Update centos.md

ksabhnan:
Text doesnt match version from command output

### DIFF
--- a/install/linux/docker-ce/centos.md
+++ b/install/linux/docker-ce/centos.md
@@ -175,7 +175,7 @@ from the repository.
 
     > **Note**: The version string is the package name plus the version up to
     > the first hyphen. In the example above, the fully qualified package name
-    > is `docker-ce-17.06.1.ce`.
+    > is `docker-ce-18.03.ce`.
 
     ```bash
     $ sudo yum install <FULLY-QUALIFIED-PACKAGE-NAME>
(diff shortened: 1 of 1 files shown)


install/linux/docker-ce/centos.md
```diff
@@ -175,7 +175,7 @@ from the repository.
 
     > **Note**: The version string is the package name plus the version up to
     > the first hyphen. In the example above, the fully qualified package name
-    > is `docker-ce-17.06.1.ce`.
```
joaofnfernandes: If you notice, on line `164` we are using a variable - `{{ site.docker_ce_stable_version }}`, but when the variable is set in the `_config.yml` file we don't include the patch number there.

When you look at the live docs it shows

```
docker-ce.x86_64            18.03.ce-1.el7.centos             docker-ce-stable
```

But it should actually be

```
docker-ce.x86_64            18.03.0.ce-1.el7.centos             docker-ce-stable
```

Would you mind updating your PR so that line `164` is 
```
{{ site.docker_ce_stable_version }}.0.ce...
```
and also update your change to line `178` to be `18.03.0.ce`?

ksabhnan: Done

joaofnfernandes:
Let's ship it 🚀

PR #6309, opened 2018-03-26
Update overlayfs-driver.md

stshow:
The docker info output shows that the driver is overlay. However, in the path that follows it states that the output confirms that overlay2 is in use. This is likely to cause confusion as the docker info output should state overlay2 if that is what is configured.


joaofnfernandes:
Let's ship it 🚀

PR #6311, opened 2018-03-26
Update password to comply with password rules

ChristianKniep:
The password needs an upper-case and a number to comply. Like that it fails and for a beginner it might not be obvious why - even thought it will print an error; but it's just one line.

### Proposed changes

The password needs some upper-case and a number to not fail while the SQL server starts.


joaofnfernandes:
Thanks for fixing it. Let's ship it!

PR #6318, opened 2018-03-27
correct example for --aux-address  parameter

xiaods:


joaofnfernandes:
Thanks for the contribution!

PR #6334, opened 2018-03-28
Remove Gated BYOL

gbarr01:
Address https://github.com/docker/store-product/issues/1729

cc: @KickingTheTV


KickingTheTV:
@gbarr01 We need to remove everything including and below this section: https://docs.docker.com/docker-store/byol/#gated-byol

Specifically the following sections:

Gated BYOL
API reference for ISVs
Group product keys

KickingTheTV:
The first paragraph needs to change to the following:

Bring Your Own License (BYOL) allows customers with existing software licenses to easily migrate to the containerized version of the software you make available on Docker Store.

To see and access an ISV's BYOL product listing, customers simply subscribe to the product with their Docker ID. We call this Ungated BYOL.

ISVs can use the Docker Store/Hub as an entitlement and distribution platform for their various audiences:

- Existing customers that want their licensed software made available as Docker containers.
- New customers that are only interested in consuming their software as Docker containers.

PR #6335, opened 2018-03-28
Update links to AppArmor's wiki

dalg24:
### Proposed changes

Fix dead link that gave
```
# Empty page
You are not allowed to create wiki pages
```


dalg24:
Turns out AppArmor's Globbing Syntax link is also outdated.

NOTE: the anchor `#AppArmor_globbing_syntax` does not seem to do its job.  I could not figure out how to make it work.

PR #6337, opened 2018-03-29
__docker_machine_ps1 is not found

macknight:
Fix __docker_machine_ps1: command not found


macknight:
I find a solution in https://github.com/docker/machine/issues/2529

joaofnfernandes:
Let's ship it!

PR #6340, opened 2018-03-29
Update upgrade.md

lenestopage:


joaofnfernandes:
Can you give a bit more context into what you're trying to achieve?
Are you adding this note because previous versions of UCP allowed the TLS connection to be downgraded to TLS 1.1, and you want to make sure users know that they can configure UCP to never downgrade the UCP connection?

If that's the case why did you chose to add that information as a note that's part of the upgrade process? Is it because when you upgrade from a previous version, UCP won't use the new default (force TLS connections to a 1.2)?

joaofnfernandes:
Ping @lenestopage for a reply, otherwise we'll close this contribution due to inactivity.

joaofnfernandes:
I'm closing this one due to inactivity 😢

PR #6344, opened 2018-03-29
Remove unsupported strftime code

NotMrSteve:
Remove the `%L` option as it is not supported by `awslogs-datetime-format`.

### DIFF
--- a/config/containers/logging/awslogs.md
+++ b/config/containers/logging/awslogs.md
@@ -169,7 +169,6 @@ The following `strftime` codes are supported:
 | `%p` | AM or PM.                                                        | AM       |
 | `%M` | Minute as a zero-padded decimal number.                          | 57       |
 | `%S` | Second as a zero-padded decimal number.                          | 04       |
-| `%L` | Milliseconds as a zero-padded decimal number.                    | 123      |
 | `%f` | Microseconds as a zero-padded decimal number.                    | 000345   |
 | `%z` | UTC offset in the form +HHMM or -HHMM.                           | +1300    |
 | `%Z` | Time zone name.                                                  | PST      |


joaofnfernandes:
@thaJeztah I'd like your input on this one.
As far as I can tell, there is no `%L` option in either in [AWS CloudWatch](https://docs.aws.amazon.com/AmazonCloudWatch/latest/logs/AgentReference.html) or [Python's datetime package](https://docs.python.org/2/library/datetime.html#strftime-and-strptime-behavior).

But [looking at the driver](https://github.com/moby/moby/blob/master/daemon/logger/awslogs/cloudwatchlogs.go#L162-L210), it seems that we always convert the directives into a regex first. If that's the case we should not merge this contribution.

thaJeztah:
Yes, I think this should be supported (also looking at the driver code) 

@NotMrSteve have you tried using it? Did it cause an error / problem?

NotMrSteve:
I tried it and it didn't work when running a container on AWS ECS, using the ECS-Optimized Amazon Linux AMI (2017.09.k). Not sure what version of docker that AMI is using. The rest of the pattern matched absolutely fine, as soon as I added the `%L` it stopped matching. I haven't tried it anywhere else.

It's possible this is a bug in the code rather than the documentation - maybe add to the documentation to clarify that `%L` is intentionally used and not in `strftime` standard? This comment is what led me to assume it was a documentation issue: "The following `strftime` codes are supported:" (https://github.com/docker/docker.github.io/pull/6344/files#diff-dba43446f630f91349e3973091c3bb88R154)

NotMrSteve:
Just found the issue: Looking at the driver, `%L` matches `\.\d{3}` (https://github.com/moby/moby/blob/785b3e32876e10d638d9b1bae00e3f340a2f47ec/daemon/logger/awslogs/cloudwatchlogs.go#L209). This means it's requires a leading `.` at the start of the pattern which is not documented. This doesn't match the `%f` behaviour, which matches `\d{6}`.

The time part I am trying to match doesn't have a leading `.`, as the format is:
`00:00:00,789`

thaJeztah:
@NotMrSteve could you perhaps open an issue with details in https://github.com/moby/moby/issues ?

NotMrSteve:
Done. Thanks! See https://github.com/moby/moby/issues/36796

PR #6360, opened 2018-04-02
Adding link to ICCS docs

artberger:
We had some users find the D4IC docs and think that it was the same
thing as the IBM Cloud container service docs. To address this, I added:
* Link on the index page to the ICCS docs
* An FAQ in case they are looking for containers in Kubernetes clusters


artberger:
@joaofnfernandes FYI

PR #6361, opened 2018-04-02
Update s3.md

rcarmstrong:
I am unsure if this problem is unique to trying to apply this policy to a bucket using Terraform, but I was unable to use the policy straight from the docs to configure my storage backend's bucket permissions. The error I would receive from Terraform is as follows;

```
Error putting S3 policy: MalformedPolicy: Missing required field Principal
```

I recreated the policy using the AWS policy generator and it added the principal section.

### DIFF
--- a/registry/storage-drivers/s3.md
+++ b/registry/storage-drivers/s3.md
@@ -187,22 +187,30 @@ Amazon S3 or S3 compatible services for object storage.
 
 ## S3 permission scopes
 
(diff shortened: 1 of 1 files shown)


ahh-docker:
@mark-church PTAL. Thanks.

L-Hudson:
Closing for lack of response from anyone. If you feel the ticket needs to be reopened, just let us know.

rcarmstrong:
Bummer

Issue #6322, opened 2018-03-27
Please describe that ppc64 is big endian


yongshin:
[install/linux/docker-ee/rhel.md](https://docs.docker.com/install/linux/docker-ee/rhel/), CC @gbarr01 

There seems to be some confusion for people installing EE on Power thinking ppc64 is what we support. Could we be more explicit and add the following the the note about ppc64le?

```
If you run `uname -a` and see the results of the command include `ppc64` 
instead of `ppc64le` this means you are running RHEL in big endian mode. 
```

justincormack:
the title of this issue should read "Please describe that ppc64 is little endian"!!!

gbarr01:
Hi, @yongshin , I can add it, but we do say: 

> To install Docker EE, you need ... ppc64le (IBM Power) architectures.

yongshin:
I know this might be tricky to accommodate in that page, but really people think they can install on pp64, we just need to be explicit and say that if you see 'ppc64' that is RHEL in big endian mode and we don't support it. Otherwise they will still try to install even though they see it on works on 'ppc64le'.

gbarr01:
Fixed by https://github.com/docker/docker.github.io/pull/6363

PR #6363, opened 2018-04-03
Address ppc64le PRs; Clean up EE install pages

### LINKED ISSUE #6343: RHEL Power Installs require a different repo for extra rpms
[install/linux/docker-ee/rhel.md](https://docs.docker.com/install/linux/docker-ee/rhel/), CC @gbarr01 

In step 5 of Install using the repository is actually different for IBM power installs because of the different repo names RHEL uses for little endian. The customer had issues because the container-selinux package was out of date and the latest container-selinux package was out of date. 

```
subscription-manager repos --enable=rhel-7-for-power-le-extras-rpms
```

For posterity's sake, this is the workaround the customer went through to install the latest version of container-selinux:  

```
yum-config-manager --enable extras
subscription-manager repos --enable=rhel-7-for-power-le-extras-rpms
yum makecache fast
yum -y install container-selinux
```

gbarr01:
- Clarified the Docker EE on IBM Power only supports `ppc64le`
- Cleaned up Docker EE install pages for RHEL, CentOS, and Oracle
- Restructured and changed headings (to be more intuitive)

Requesting a tech review, especially in regard to old text (and possible links to new headings):

- Is the Oracle SELinux warning on Docker EE 17.06.1 still relevant? 
- If using `overlay2`, do you need to install `device-mapper-persistent-data` and `lvm2`? 
- Can IBM Power users install in the Cloud? If so, what `extras` RHEL repo should they use for the `container-selinux` package?
- Does Store or Success or any other product/site link to a doc anchor? Did anything break?

Fixes #6343 and #6322 and #6015 and #6369 

cc: @yongshin   @justincormack

### DIFF
--- a/_includes/ee-linux-install-reuse.md
+++ b/_includes/ee-linux-install-reuse.md
@@ -10,73 +10,61 @@ Usage: {% include ee-linux-install-reuse.md section="ee-install-intro" %}
 {% endcomment %}
 
 
-
 {% if section == "ee-install-intro" %}
 
-To get started with Docker EE on {{ linux-dist-long }}, make sure you
-[meet the prerequisites](#prerequisites), then
-[install Docker](#install-docker-ee).
+There are two ways to install and upgrade [Docker Enterprise Edition (Docker EE)](https://www.docker.com/enterprise-edition/){: target="_blank" class="_" }
+on {{ linux-dist-long }}:
 
-{% elsif section == "ee-url-intro" %}
+- [YUM repository](#repo-install-and-upgrade): Set up a Docker repository and install Docker EE from it. This is the recommended approach because installation and upgrades are managed with YUM and easier to do.
 
-To install Docker Enterprise Edition (Docker EE), you need to know the Docker EE
-repository URL associated with your trial or subscription. These instructions
-work for Docker EE for {{ linux-dist-long }} and for Docker EE for Linux, which
-includes access to Docker EE for all Linux distributions. To get this
-information:
+- [RPM package](#package-install-and-upgrade): Download the {{ package-format }} package, install it manually, and manage upgrades manually. This is useful when installing Docker EE on air-gapped systems with no access to the internet.
 
-- Go to [https://store.docker.com/my-content](https://store.docker.com/my-content).
-- Each subscription or trial you have access to is listed. Click the **Setup**
-  button for **Docker Enterprise Edition for {{ linux-dist-long }}**.
-- Copy the URL from the field labeled
-  **Copy and paste this URL to download your Edition**.
+{% if linux-dist == "rhel" or linux-dist == "oraclelinux" %}
+Docker Community Edition (Docker CE) is _not_ supported on {{ linux-dist-long }}.
+{% endif %}
+{% if linux-dist == "centos" %}
+For Docker Community Edition on {{ linux-dist-cap }}, see [Get Docker CE for CentOS](/install/linux/docker-ce/centos.md).
+{% endif %}
 
-Use this URL when you see the placeholder text `<DOCKER-EE-URL>`.
+{% elsif section == "find-ee-repo-url" %}
 
-To learn more about Docker EE, see
-[Docker Enterprise Edition](https://www.docker.com/enterprise-edition/){: target="_blank" class="_" }.
+To install Docker EE, you will need the URL of the Docker EE repository associated with your trial or subscription:
+
+1.  Go to [https://store.docker.com/my-content](https://store.docker.com/my-content){: target="_blank" class="_" }. All of your subscriptions and trials are listed.
+2.  Click the **Setup** button for **Docker Enterprise Edition for {{ linux-dist-long }}**.
+3.  Copy the URL from **Copy and paste this URL to download your Edition** and save it for later use.
+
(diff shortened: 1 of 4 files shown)


_includes/ee-linux-install-reuse.md
```diff
@@ -10,83 +10,69 @@ Usage: {% include ee-linux-install-reuse.md section="ee-install-intro" %}
...
-[install Docker](#install-docker-ee).
-
-{% elsif section == "ee-url-intro" %}
-
-To install Docker Enterprise Edition (Docker EE), you need to know the Docker EE
-repository URL associated with your trial or subscription. These instructions
-work for Docker EE for {{ linux-dist-long }} and for Docker EE for Linux, which
-includes access to Docker EE for all Linux distributions. To get this
-information:
+There are two ways to install and upgrade [Docker Enterprise Edition (Docker EE)](https://www.docker.com/enterprise-edition/){: target="_blank" class="_" }
```
joaofnfernandes: test

yongshin:
@gbarr01 the only place to likely deploy Power for EE on cloud is IBM Cloud. Also it defaults to overlay2 and shows a warning when running it. I would assume that is fine since that is our recommendation on RHEL and Ubuntu.

install/linux/docker-ee/rhel.md
```diff
@@ -7,63 +7,50 @@ redirect_from:
 - /engine/installation/linux/rhel/
 - /engine/installation/linux/docker-ee/rhel/
 title: Get Docker EE for Red Hat Enterprise Linux
-toc_max: 4
 ---
 
 {% assign linux-dist = "rhel" %}
+{% assign linux-dist-cap = "RHEL" %}
```
joaofnfernandes: You don't really need to define a new variable for this. You can use the existing one and use a filter to capitalize or upcase the value. For that delete this, where you are using this variable, use this instead:

```
{{ linnux-dist | upcase }}
```

[You can learn more here](https://shopify.github.io/liquid/filters/upcase/)

_includes/ee-linux-install-reuse.md
```diff
@@ -131,120 +127,84 @@ You can install Docker EE in different ways, depending on your needs:
...
-3.  If you need to use `devicemapper`, follow the procedure in the
-    [devicemapper storage driver guide](/engine/userguide/storagedriver/device-mapper-driver.md#configure-direct-lvm-mode-for-production){: target="_blank" class="_" }
-    **before starting Docker**. For production systems using `devicemapper`,
-    you must use `direct-lvm` mode,
-    which requires you to prepare the block devices.
-{% endif %}
+3.  Start Docker:
 
-4.  Start Docker.
+    > If using `devicemapper`, ensure it is properly configured before starting Docker, per the [storage guide](/storage/storagedriver/device-mapper-driver/){: target="_blank" class="_" }.
```
joaofnfernandes: I've noticed a couple of notes like this. Make sure you use the new style

```
> Note title
>
> line 1
> line 2
```

install/linux/docker-ee/oracle.md
```diff
@@ -6,90 +6,90 @@ redirect_from:
...
 
-{% include ee-linux-install-reuse.md section="ways-to-install" %}
-
-### Install using the repository
-
-Before you install Docker EE for the first time on a new host machine, you need
-to set up the Docker repository. Afterward, you can install and update Docker EE
-from the repository.
+{% capture selinux-warning %}
+> Docker EE cannot install on {{ linux-dist-long }} with SELinux enabled!
```
joaofnfernandes: I think we should use the same convention for note titles as we use for regular headings: no punctuation at the end.

_includes/ee-linux-install-reuse.md
gbarr01: @joaofnfernandes , it takes up so much space. I thought this was one of 3 styles. ;)

joaofnfernandes: The first style should be considered legacy. It works because we have a lot of stuff written in the past using that style. Whenever we fix something or write a new piece we should be using the new style.

I also think it's a good thing that the new style takes a lot of space. Whenever you write a note it should feel ugly to you as a writer, so that we don't end up with docs with a ton of notes spread all over the place.

install/linux/docker-ee/rhel.md
gbarr01: Actually, @joaofnfernandes , this solution doesn't work for CentOS and Oracle Linux (OL).

_includes/ee-linux-install-reuse.md
gbarr01: I agree that too many notes is an eyesore and makes a page unreadable. I disagree that all notes should be as large and as ugly as possible. Within procedures/tasks, if I need a note, I do my best to keep it crisp and on one line. We can discuss offline.

clnperez:
Can anyone comment specifically as to why the line recommending that selinux be disabled when installing on Power was added? If the container-selinux package is installed, shouldn't this be okay? Or was there some bug not referenced here? I'm kind of wondering if that was a typo, and should be z/s390x instead, due to other mentions that selinux isn't supported for z.

gbarr01:
@clnperez , definitely looks like a typo. Fixing ...

PR #6368, opened 2018-04-03
Adding "windows" label to Windows node.

ekitagawa:
Additional steps to add label to Windows Nodes.  This will allow admin to filter Windows node when creating a service like this.
`docker service create --constraint engine.labels.os==windows --name winonly microsoft/nanoserver`  The same steps are on this article: https://success.docker.com/article/how-can-i-assign-a-service-to-windows-nodes-only


joaofnfernandes:
@ekitagawa, @JimGalasyn I thought swarm would check the `platform.os` listed on the image manifest, and if it was Windows (you're trying to deploy a Windows image), then swarm would make sure to schedule the service on a Windows node.

Do you know if that really happens? If that's the case you don't really need to add constrains to your services.

joaofnfernandes:
This has been stalled for a while, and it doesn't hurt to add labels to your nodes. Let's merge it. I've filed #6690 to bring this to UCP 3.0 docs too.

ekitagawa:
so sorry for the delay @joaofnfernandes... Yes, in an ideal world, task assignment does check the platform.  but we had an issue like this https://github.com/docker/orca/issues/8591 so it's good to add the label.  thank you for merging this!

PR #6372, opened 2018-04-04
Add a note for the DNS settings in docs.

floatingpurr:
### Proposed changes

A little modification in docs to describe DNS config. Reason: a lot of people struggle against problems with the DNS configuration when they try to build their first app following the docs.

n/a

### Related issues (optional)

n/a

### DIFF
--- a/get-started/part2.md
+++ b/get-started/part2.md
@@ -84,43 +84,6 @@ ENV NAME World
 CMD ["python", "app.py"]
 ```
 
-> Are you behind a proxy server?
->
-> Proxy servers can block connections to your web app once it's up and running.
-> If you are behind a proxy server, add the following lines to your
-> Dockerfile, using the `ENV` command to specify the host and port for your
-> proxy servers:
->
-> ```conf
-> # Set proxy server, replace host:port with values for your servers
-> ENV http_proxy host:port
-> ENV https_proxy host:port
-> ```
->
-> Add these lines before the call to `pip` so that the installation succeeds.
-
-
-> Setting the DNS server
->
-> DNS misconfgiurations can generate problems with `pip`. You need to set your 
-> own DNS server address to make `pip` work properly. You might want 
-> to change the DNS settings of the Docker daemon. You can edit (or create) the 
-> configurarion file at `/etc/docker/daemon.json` with the `dns` key, as following:
->
-> ```json
->{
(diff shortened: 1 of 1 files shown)


joaofnfernandes:
Thanks for putting in the time to fix this @floatingpurr! I know that this is affecting some users but this getting started is becoming a monster of its own, riddled with tons of notes and caveats.
If we continue going down this path, the get started will be basically a list of assorted notes.

I'm probably guessing that you're using docker on Linux (not docker for mac or windows). If that's the case can you tell me a bit more about the distro where you're experiencing this problem?

We've also been having discussions about reviewing this getting started for a couple of reasons. If we do that, the revised version should give you all the resources you need upfront, an not rely on package managers or on fetching resources from the internet. That would spare people from these proxying/DNS problems.

floatingpurr:
@joaofnfernandes you are welcome. I see. I thought about that fix just to complete the caveat _"Are you behind a proxy server?"_, since I think the problem with DNS is very common for Docker users on Linux.

In particular, I'm on Ubuntu 16.04.4 LTS (GNU/Linux 4.4.0-116-generic x86_64).

joaofnfernandes:
@floatingpurr I thought you might have another distro, but if you're running into this issue with Ubuntu, I'm kind of concerned since a lot of folks will be using it too 😩

I think we should integrate your contribution if you're willing to make some changes to it:
* Try to merge the two notes about proxies and DNS into a single note. If you think making the DNS changes will make things work if you're behind a proxy, you can delete the note about the proxy.
* The note title should make clear that if you're not using Linux you can skip reading the note. Maybe something like "Troubleshoot problems on Linux". This is important because if Docker for Mac/Win are having problems they might drive themselves crazy trying to use this fix and it's not going to work for.

floatingpurr:
@joaofnfernandes, I think that such a problem is pretty common. At least for Debian/Ubuntu [users](https://stackoverflow.com/questions/28668180/cant-install-pip-packages-inside-a-docker-container-with-ubuntu/43178937
).

I did not delete the proxy part but I merged the two notes. Besides, I moved the new note after the description of the build command, since it's where that kind of problems usually arises.

Hope it helps! : )

joaofnfernandes:
Thanks for putting in the work to get this across the finish line 🏁
Let's ship it!

floatingpurr:
It was a pleasure!

PR #6373, opened 2018-04-04
Clean config docs, don't share /var/lib/docker

### LINKED ISSUE #5773: tmpfs topic references --volume instead of --tmpfs
[engine/admin/volumes/tmpfs.md](https://docs.docker.com/engine/admin/volumes/tmpfs/), CC @mistyhacks

Where it reads 

_The --mount syntax is more verbose than **-v** or **--volume**_

shouldn't it be

_The --mount syntax is more verbose than **--tmpfs**_

?

It looks like the sentence was a copy-paste from volumes or bind mounts topics. Although it does not affect the understanding of the topic, it may confuse some readers.

### LINKED ISSUE #6267: Include an explicit statement in the docs advising against moving /var/lib/docker to an NFS share
### Problem description

Per [this](https://github.com/moby/moby/issues/22539#issuecomment-217316284) comment in the upstream (CC @thaJeztah), we should have an explicit statement in the documentation that tells users not to move `/var/lib/docker` to an NFS share; more specifically, telling users not to set the Docker daemon `--data-root` flag to an NFS share.

### Problem location and fix suggestions

We should include a statement prohibiting this somewhere on the [About storage drivers](https://docs.docker.com/storage/storagedriver/) page along with an explanation as to why one shouldn't do so. We should also include a similar statement next to the `--data-root` daemon flag docs somewhere in [this](https://github.com/docker/cli/blob/master/docs/reference/commandline/dockerd.md#run-multiple-daemons) section.

joaofnfernandes:
This change cleans some explanations on how volume works and how to configure the Docker daemon.
It also make it clear that the Docker engine data file, which by default is `/var/lib/docker` should be dedicated for a single Docker daemon. You should not [share that directory between multiple daemons](https://www.youtube.com/watch?v=LMMUYQp_XRA).

I've split this PR in multiple commits to make it easier to review.

* Fixes #6267
* Fixes #5773


gbarr01:
@joaofnfernandes , do you think it addresses any of the other storage issues? https://github.com/docker/docker.github.io/issues?q=is%3Aopen+is%3Aissue+label%3Aarea%2Fstorage

storage/volumes.md
```diff
@@ -362,6 +362,25 @@ $ docker container rm nginxtest
...
+applications to store files on a cloud object storage system like Amazon S3.
+Another option is to develop your applications to use volumes, but create the
+volume using a driver that supports writing files to an external storage system
+like NFS or Amazon S3.
+
+Volume drivers allow you to abstract to underlying storage system from the
+application logic. As an example if your services are using a volume with
+an NFS driver, you can update the service to use a volume with a different
+driver and start storing data in the cloud, without having to make any changes
+the the application logic.
```
gbarr01: remove one "the"

gbarr01: This is stylistic (so take it or leave it), but a tighter way to express these paragraphs might be: 

> There are several ways to achieve this when developing your applications. One is to store files on a cloud object storage system like Amazon S3. Another is to create volumes with a driver that supports writing files to an external storage system like NFS or Amazon S3.
> 
> Volume drivers allow you to abstract the underlying storage system from the application logic. For example, if your services use a volume with an NFS driver, you can update the services to use a different driver (say, to store data in the cloud), without changing the application logic.

storage/index.md
```diff
@@ -19,10 +19,11 @@ are some downsides:
...
-Docker host: _volumes_, _bind mounts_, or _`tmpfs` volumes_. When in doubt,
-volumes are almost always the right choice. Keep reading for more information
-about each mechanism for mounting data into containers.
+Docker has two options for containers to store files in the host machine, so
+that the files are persisted even after the container stops: _volumes_, and
+_bind mounts_. If you're running Docker on Linux you can also use a _tmpfs mount_.
+
+Keep reading for more information about these two ways of persisting data.
 
 ## Choose the right type of mount
```
gbarr01: @joaofnfernandes , down the page (and not part of your edit), can you fix the warning note. It looks malformed. It's part of the **bind mounts** section.

storage/storagedriver/index.md
```diff
@@ -397,33 +396,6 @@ started a new container, container start times and disk space used would be
...
-operate at native host speeds. You can mount any number of data volumes into a
-container. Multiple containers can also share one or more data volumes.
-
-The diagram below shows a single Docker host running two containers. Each
-container exists inside of its own address space within the Docker host's local
-storage area (`/var/lib/docker/...`). There is also a single shared data volume
-located at `/data` on the Docker host. This is mounted directly into both
-containers.
-
-![Shared volume across containers](images/shared-volume.jpg)
```
gbarr01: I find this graphic useful. Why are you deleting it?

config/daemon/index.md
```diff
@@ -85,6 +108,35 @@ documentation. Some places to go next include:
...
+
+By default this directory is:
+
+* `/var/lib/docker` on Linux.
+* `C:\ProgramData\docker` on Windows.
+
+You can configure the Docker daemon to use a different directory, using the
+`data-root` configuration option.
+
+Since all the state of a Docker daemon is kept on this directory, make sure
```
gbarr01: Since the state...

config/daemon/index.md
```diff
@@ -85,6 +108,35 @@ documentation. Some places to go next include:
...
+you use a dedicated directory for each daemon. If two daemons share the same
+directory, as an example using an NFS share, you might get errors that are
+difficult to troubleshoot.
+
```
gbarr01: If two daemons share the same directory, for example, an NFS share, you might ...

storage/tmpfs.md
```diff
@@ -6,24 +6,30 @@ redirect_from:
...
-non-persistent application state. An example might be a temporary one-time
-password that the container's application creates and uses as-needed.
-
-To give the container access to the data without writing it anywhere
-permanently, you can use a `tmpfs` mount, which is only stored in the host
-machine's memory (or swap, if memory is low). When the container stops, the
-`tmpfs` mount is removed. If a container is committed, the `tmpfs` mount is not
-saved.
+[Volumes](volumes.md) and [bind mounts](bind-mounts.md) allow you to share files
+between the host machine and container. This allows you to persist data even
```
gbarr01: "allows you" is used twice back to back. maybe this? 

> Volumes and bind mounts let you share files between the host machine and container so that you can persist data even after the container is stopped.

storage/tmpfs.md
```diff
@@ -6,24 +6,30 @@ redirect_from:
...
+after the container is stopped.
+
+If you're running Docker on Linux, you have a third option: `tmpfs` mounts.
+When you create a container with a `tmpfs` mount, the container has a way to
+create files outside the container's writable layer, just like it happens with
+volumes and bind mounts.
```
gbarr01: This is strange: "just like it happens with ...". Maybe, 

> When you create a container with a `tmpfs` mount, the container can create files outside the container's writable layer (as it can with volumes and bind mounts).

PR #6374, opened 2018-04-04
Remove references to production; Clean text phase 1

gbarr01:
- Removed text about CE on prod systems
- Tightened text in Install  CE sub-section

There is much more to do on these pages ...

cc: @JustinINevill

### DIFF
--- a/install/linux/docker-ce/centos.md
+++ b/install/linux/docker-ce/centos.md
@@ -131,32 +131,30 @@ from the repository.
 
 #### Install Docker CE
 
-1.  Install the latest version of Docker CE, or go to the next step to install a
-    specific version.
+1.  Install the _latest version_ of Docker CE, or go to the next step to install a specific version:
 
     ```bash
     $ sudo yum install docker-ce
     ```
 
-    > **Warning**: If you have multiple Docker repositories enabled, installing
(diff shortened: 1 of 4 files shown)


install/linux/docker-ce/centos.md
```diff
@@ -131,56 +131,51 @@ from the repository.
...
+    `060A 61C5 1B55 8A7F 742B 77AA C52F EB6B 621E 9F35`, and if so, accept it.
+
+    > Got multiple Docker repositories?
+    >
+    > If you have multiple Docker repositories enabled, installing
     > or updating without specifying a version in the `yum install` or
     > `yum update` command always installs the highest possible version,
     > which may not be appropriate for your stability needs.
-    {:.warning}
+    {:.warning-vanilla}
```
joaofnfernandes: This notation applies a CSS class, so you're basically creating

```
<blockquote class="warning-vanilla">
```

But we don't have any CSS declaration for that class. If you want this note to render in blue, just delete the `{...}`

I see that this class is also being used in other places, so consider doing a search/replace.

gbarr01:
Merging knowing full well that this page needs more cleanup, including the notes, no matter what style you prefer.

PR #6383, opened 2018-04-06
less is more - avoid cat to grep something.

gionniboy:
### Proposed changes
Changed seccomp docs: avoid cat to grep boot config file.

reference: http://porkmail.org/era/unix/award.html


teknoraver:
Seems legit.
While you're at it, replace backticks, which are deprecated, with $()

Issue #6381, opened 2018-04-05
SLAT (Second Level Adress Translation) hardware feature not clearly defined as prerequisite for Docker for Windows


docbenton:
[docker-for-windows/troubleshoot.md](https://docs.docker.com/docker-for-windows/troubleshoot/), CC @gbarr01 

after installing Docker for Windows  ( Linux container chosen)
engine fails to start with famous popup : 
"An error occurred. Hardware assisted virtualization and data execution protection must be enabled in the BIOS. See https://docs.docker.com/docker-for-windows/troubleshoot/#virtualization-must-be-enabled"

the aim of this current issue is to avoid the user fighting to enable hypervisor to avoid this popup at the end of install.

the pre-requisite for hypervisor to start mobylinux vm are
- cpu virtualisation feature, enabled in bios,   
- data execution protection
- SLAT cpu feature (called EPT by intel)
- windows 10 enterprise
..
but in paragraph VIRTUALIZATION MUST BE ENABLED,
SLAT/EPT is not stated in the referred page, and no tool is suggested to check for it...

for instance,  'systeminfo'  command output : 

Hyper-V Requirements:
                           VM Monitor Mode Extensions: Yes
                           Virtualization Enabled In Firmware: Yes
                           Second Level Address Translation: No
                           Data Execution Prevention Available: Yes
`

could easily be checked by installer...
and 'docker toolbox' could be suggested as workaround...
 that's what I did, but not immediately, only after several install attempts and finally discovering that my cpu was not SLAT compliant.

near virtualization,  i was also not able to find  a reference on SLAT/EPT on this page https://docs.docker.com/docker-for-windows/install/#what-to-know-before-you-install

thaJeztah:
Ping @akimd

docbenton:
sorry for silly question, but how to ping/send a msg ? using his email he may has left has a commiter ?

akimd:
Actually by merely mentioning my GitHub ID (which the @ before), thaJeztah pinged me.  And his "ping akim" was the action itself, he was not suggesting "you should ping akim" :)

plemon:
Doc File: [https://docs.docker.com/docker-for-windows/install/](url)

Docs should be improved a little regarding CPU requirements where it says 'CPU SLAT-capable feature'. Users can be helped greatly with a link to check if the target installation CPU meets SLAT capability or not. 

Please consider referencing this MS page so users can do an easy verification:
[...]

PR #6395, opened 2018-04-06
CPU must be SLAT-capable for Hyper-V

docbenton:

good to know before you install that SLAT-Capable cpu is needed for Hyper-V

### Related issues (optional)
#6381

### DIFF
--- a/docker-for-windows/install.md
+++ b/docker-for-windows/install.md
@@ -74,7 +74,7 @@ one typically created during Toolbox install) no longer start. These VMs
 cannot be used side-by-side with Docker for Windows. However, you can still use
 `docker-machine` to manage remote VMs.
 <p />
-* Virtualization must be enabled. Typically, virtualization is enabled by default. This is different from having Hyper-V enabled. For more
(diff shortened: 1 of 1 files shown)


gbarr01:
cc: @akimd

gbarr01:
@akimd ... you sure about the mac similarity? After all, Hyper-V is really _Microsoft Hyper-V_. 

Also, I wonder about this definitive statement about the BIOS and SLAT.  Do all systems need SLAT support, or only some? From what I can tell, only some ...

gbarr01:
I take back my 2nd point per this doc (to which we actually link) :https://docs.microsoft.com/en-us/virtualization/hyper-v-on-windows/reference/hyper-v-requirements

> Hyper-V requires Second Level Address Translation (SLAT) -- present in the current generation of 64-bit processors by Intel and AMD.

gbarr01:
Thanks, @docbenton

akimd:
@gbarr01 Honestly, I have no idea.  A quick search gave nothing, but I expect that the requirements of Hyper-V of MS match those of the Hypervisor framework of Apple.  And after all, these are the same chips.  But given that Apple has full control on the hardware, maybe all the processors of Mac that support virtualization also support SLAT.

Anyway, let's not bother, future will tell.

PR #6397, opened 2018-04-07
Extension fields are only supported from 3.4

Maks3w:
### Proposed changes

Use the minimum supported version in the example

N/A

### Related issues (optional)

N/A


shin-:
See [previous discussion](https://github.com/docker/docker.github.io/pull/6730#issuecomment-391156958) on this topic. It looks like current master uses a mix of both now, meaning #6730 probably missed one. Overall, looks fine to me.

paigehargrave:
Included changes based on technical review.

PR #6398, opened 2018-04-07
Repair overlay standalone tutorial

### LINKED ISSUE #6134: Overlay network is not automatically created on the second host
[network/network-tutorial-overlay.md](https://docs.docker.com/network/network-tutorial-overlay/), CC @gbarr01 
In tutorial "Use an overlay network for standalone containers" Step 4 suggests that an overlay network "test-net" should be automatically created on the second host once user runs a container using that network. However this doesn't happen since no service has been deployed on the swarm using that network.

### LINKED ISSUE #6376: Broken/wrong section
[network/network-tutorial-overlay.md](https://docs.docker.com/network/network-tutorial-overlay/), CC @gbarr01 

Trying to get into docker and swarm I followed the tutorials in the docs. At the section "Use an overlay network for standalone containers" it states in the beginning "creates an alpine service with 3 replicas, connected to the overlay network" but never happens. Therefore the tutorial is incorrect and results in Error: No such container.

Env:
centOS 7
Docker version 18.03.0-ce, build 0520e24

I propose to include a step which explains how to create a service and attach it to the network (docker service create --network test-net [...] ). Afterwards it should be possible to atatch the container (?). As soon as I get something working I might create a pull.

gbarr01:
- Removed part about 3 replicas
- Fixed part about attaching to `alpine2` from host1 -- should be `alpine1`
- Removed command to spawn detached container and then attach -- just don't detach in the first place
- Clarified use of -dit flags

@thaJeztah , the original text referred to this tutorial as demonstrating "service discovery" ... though it seems more like "network discovery" as no services are created. thoughts? 

Fixes #6376 
Fixes #6134

### DIFF
--- a/network/network-tutorial-overlay.md
+++ b/network/network-tutorial-overlay.md
@@ -110,7 +110,7 @@ and will be connected together using an overlay network called `ingress`.
 
     ID                            HOSTNAME            STATUS              AVAILABILITY        MANAGER STATUS
     d68ace5iraw6whp7llvgjpu48 *   ip-172-31-34-146    Ready               Active              Leader
-    nvp5rwavvb8lhdggo8fcf7plg     ip-172-31-35-151    Ready               Active              
+    nvp5rwavvb8lhdggo8fcf7plg     ip-172-31-35-151    Ready               Active
     ouvx2l7qfcxisoyms8mtkgahw     ip-172-31-36-89     Ready               Active
     ```
 
@@ -125,8 +125,8 @@ and will be connected together using an overlay network called `ingress`.
     $ docker node ls --filter role=worker
 
     ID                            HOSTNAME            STATUS              AVAILABILITY        MANAGER STATUS
-    nvp5rwavvb8lhdggo8fcf7plg     ip-172-31-35-151    Ready               Active              
-    ouvx2l7qfcxisoyms8mtkgahw     ip-172-31-36-89     Ready               Active  
+    nvp5rwavvb8lhdggo8fcf7plg     ip-172-31-35-151    Ready               Active
+    ouvx2l7qfcxisoyms8mtkgahw     ip-172-31-36-89     Ready               Active
     ```
 
 5.  List the Docker networks on `manager`, `worker-1`, and `worker-2` and notice
@@ -274,22 +274,21 @@ This tutorial assumes the swarm is already set up and you are on a manager.
 
 ## Use an overlay network for standalone containers
 
-This example does the following:
+This example demonstrates how to communicate between standalone containers on
+different Docker daemons using an overlay network. Steps are:
 
-- initializes a swarm on `host1`
-- joins `host2` to the swarm
-- creates an attachable overlay network
-- creates an `alpine` service with 3 replicas, connected to the overlay network
-- creates a single `alpine` container on `host2`, which is also attached to the
-  overlay network
-- proves that the standalone container can communicate with the service tasks,
-  and vice versa.
+- On `host1`, initialize the node as a swarm (manager)
+- On `host2`, join the node to the swarm (worker)
+- On `host1`, create an attachable overlay network (`test-net`)
+- On `host1`, run an interactive [alpine](https://hub.docker.com/_/alpine/) container (`alpine1`) on `test-net`
+- On `host2`, run an interactive, and detached, [alpine](https://hub.docker.com/_/alpine/) container (`alpine2`) on `test-net`
+- On `host1`, from within a session of `alpine1`, ping `alpine2`
 
 ### Prerequisites
 
-For this test, you need two different Docker hosts, which can communicate with
-each other. Each host needs to be running Docker 17.06 or higher. The following
-ports must be open between the two Docker hosts:
+For this test, you need two different Docker hosts that can communicate with
+each other. Each host must have Docker 17.06 or higher with the following ports
+open between the two Docker hosts:
 
 - TCP port 2377
 - TCP and UDP port 7946
@@ -298,21 +297,20 @@ ports must be open between the two Docker hosts:
 One easy way to set this is up is to have two VMs (either local or on a cloud
 provider like AWS), each with Docker installed and running. If you're using AWS
 or a similar cloud computing platform, the easiest configuration is to use a
-security group which opens all incoming ports between the two hosts and the SSH
+security group that opens all incoming ports between the two hosts and the SSH
 port from your client's IP address.
 
-This example will refer to the hosts as `host1` and `host2`, and the command
-prompts will be labelled accordingly.
-
-The example uses Linux hosts, but the same commands work on Windows.
+This example refers to the hosts as `host1` and `host2` with the command prompts
+labelled accordingly. This example also uses Linux hosts, but the same commands
+work on Windows.
 
 ### Walk-through
 
 1.  Set up the swarm.
 
-    1.  On `host1`, run `docker swarm init`, specifying the IP address for the
-        interface which will communicate with the other host (for instance, the
-        private IP address on AWS).
+    a.  On `host1`, initialize a swarm (and if prompted, use `--advertise-addr`
+        to specify the IP address for the interface that communicates with other
+        hosts in the swarm, for instance, the private IP address on AWS):
 
         ```bash
         (host1) $ docker swarm init --advertise-addr 192.0.2.1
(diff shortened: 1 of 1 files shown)


network/network-tutorial-overlay.md
```diff
@@ -328,66 +326,66 @@ The example uses Linux hosts, but the same commands work on Windows.
...
-    > **Note**: There is nothing to prevent you from using the same container
-    > name on multiple hosts, but automatic service discovery will not work if
-    > you do, and you will need to refer to the containers by IP address.
 
-    Verify that `test-net` was created on `host2`:
+5.  On `host2`, start a detached and interactive container (`alpine2`), without a teriminal, that
+    connects to `test-net`:
+
+    ```bash
+    (host2) $ docker run -di --name alpine2 --network test-net alpine
```
thaJeztah: Missed a `t` here (`-di`  -> `-dit`)

network/network-tutorial-overlay.md
```diff
@@ -401,52 +399,47 @@ The example uses Linux hosts, but the same commands work on Windows.
...
+9.  Clean up your containers and networks:
 
-    Detach from `alpine2` using the `CTRL` + `P` `CTRL` + `Q` sequence.
+    Because the Docker daemons operate independently and these are standalone
+    containers, you must run the commands on the individual hosts.
 
-6.  Stop the containers and remove `test-net` from each host. Because the Docker
-    daemons are operating independently and these are standalone containers, you
-    need to run the commands on the individual hosts.
+    a.  On `host1`, remove `alpine1` and `test-net`:
```
thaJeztah: I think removing `test-net` here will fail, because there's still a container attached to it on the other host

network/network-tutorial-overlay.md
```diff
@@ -328,66 +326,66 @@ The example uses Linux hosts, but the same commands work on Windows.
...
-              --network test-net \
-              alpine
+    (host2) $ docker network ls
     ```
-    The `-dit` flags mean to start the container detached
-    (in the background), interactive (with the ability to type into it), and
-    with a TTY (so you can see the input and output).
-    
-    > **Note**: There is nothing to prevent you from using the same container
-    > name on multiple hosts, but automatic service discovery will not work if
```
thaJeztah: So; I think the confusion here around "service discovery" is basically that it's not a "service", but "dns discovery of containers", is that what you meant?

network/network-tutorial-overlay.md
gbarr01: @thaJeztah , I found that it wasn't necessary. Is it? It works without it. I explain why below.

network/network-tutorial-overlay.md
gbarr01: It is now! That makes sense, @thaJeztah . Thank you.

network/network-tutorial-overlay.md
```diff
@@ -401,52 +399,47 @@ The example uses Linux hosts, but the same commands work on Windows.
...
-    ```
+    b.  On `host2`, stop and remove `alpine2` and  remove `test-net`:
+
+        ```bash
+        (host2) $ docker container stop alpine2
+                $ docker container rm alpine2
+                $ docker network rm test-net
+        ```
 
 ## Communicate between a container and a swarm service
```
joaofnfernandes: Since you're working on this, you might also want to address this issue #6407

network/network-tutorial-overlay.md
```diff
@@ -274,22 +274,21 @@ This tutorial assumes the swarm is already set up and you are on a manager.
...
 
-- initializes a swarm on `host1`
-- joins `host2` to the swarm
-- creates an attachable overlay network
-- creates an `alpine` service with 3 replicas, connected to the overlay network
-- creates a single `alpine` container on `host2`, which is also attached to the
-  overlay network
-- proves that the standalone container can communicate with the service tasks,
-  and vice versa.
+- On `host1`, initialize the node as a swarm (manager)
```
joaofnfernandes: I know this is not done consistently across the docs but we end all bullets with a final stop.

network/network-tutorial-overlay.md
```diff
@@ -328,66 +326,66 @@ The example uses Linux hosts, but the same commands work on Windows.
...
+    b30567dd179f819d11feb516a278b0c560399e5f53db8214ca23432431d5b6c0
+    ```
+
+    The `-d`, `-i`, `-t`
+    [flags](https://docs.docker.com/engine/reference/commandline/exec/#options)
+    start a container that is **detached** (runs in the background),
+    **interactive** (with `stdin` open), and with a **tty** (terminal). We run
+    `alpine1` interactively, and with a terminal, so that we can work within its
+    shell and ping `alpine2`. We run `alpine2` interactively, so that `alpine1`
+    can access its `stdin`/`stdout`, but we don't need a shell so we can make it
```
joaofnfernandes: > We run `alpine2` interactively, so that `alpine1` can access its `stdin`/`stdout`. This is not correct since alpine1 can't access stdin/out of alpine2.

Almost, but not quite.
You deploy alpine2 in interactive so the `/bin/sh` running on alpine2 can have its own stdin/stdout and continue running on the background. Without it, the alpine2 container would stop immediately.

I would prefer not to talk too much about these details since they distract from the main point: creating your network with `--attachable`.
All of the rest are just artifacts of the particular example we're using to illustrate `ping` working across nodes.

network/network-tutorial-overlay.md
```diff
@@ -401,52 +399,47 @@ The example uses Linux hosts, but the same commands work on Windows.
     uw9etrdymism        test-net            overlay             swarm
     ```
 
-5.  Remember that you created `alpine1` from `host1` and `alpine2` from `host2`.
-    Now, attach to `alpine2` from `host1`:
+7.  On `host1`, ping `alpine2` within the `/bin/sh` session of `alpine1`:
```
joaofnfernandes: I think you can just say:  On `host1`, ping `alpine2`
When reading the full sentence I'm left with a couple of questions about `/bin/sh` and session that end up being more distracting than helpful.

network/network-tutorial-overlay.md
gbarr01: With `alpine1`, we  to into `-it` mode and then exit... so the container stops.

network/network-tutorial-overlay.md
gbarr01: If you deploy `alpine1` as detached, then you have to attach it again to run `ping`.  So I'm not going to detach it. 

With `alpine2`, I was not pinging back so there was no need to attach (or in this case not detach). I was demonstrating a more deliberate use of the flags ... which is also good to know.

gbarr01: Ps. I was going to address that one in a separate PR ... we'll see.

network/network-tutorial-overlay.md
gbarr01: I removed most of it.  Simplified.

network/network-tutorial-overlay.md
gbarr01: Simplified. Thanks for the review, Joao. I tightened some things up but I kept it to one ping.

network/network-tutorial-overlay.md
gbarr01: I added it for simplicity.

network/network-tutorial-overlay.md
```diff
@@ -298,154 +298,129 @@ ports must be open between the two Docker hosts:
...
 
+    b.  On `host2`, stop and remove `alpine2` and  remove `test-net`:
+
     ```bash
-    (host2) $ docker container stop alpine2
-            $ docker container rm alpine2
-            $ docker network rm test-net
+    $ docker container stop alpine2
+    $ docker container rm alpine2
+    $ docker network rm test-net
```
fheuer: Removal of test-net not necessary.
Command results in Error: No such network: test-net

gbarr01: @fheuer , please test the new instructions:

1.  Set up a Docker swarm with 1 manager (`host1`) and 1 worker (`host2`).
2.  On `host1`, create an attachable overlay network (`test-net`) and run an interactive alpine container (`alpine1): 

    ```
    $ docker network create --driver=overlay --attachable test-net
    $ docker run -it --name alpine1 --network test-net alpine
    / #
    ```

3.  On `host2`, start a detached and interactive alpine container (`alpine2`) that connects to `test-net`, then list the networks to ensure the overlay network (`test-net`) was created:

    ```
    $ docker run -dit --name alpine2 --network test-net alpine
    $ docker network ls

    ```
4.  On `host1`, ping `alpine2` within the interactive terminal of `alpine1`:

    ```
    / # ping -c 2 alpine2
    ```

fheuer: Yep works! Still the issue stated above though.

gbarr01: Ah ha. I see what you're saying. Thanks for the feedback @fheuer

PR #6416, opened 2018-04-10
changed docker run argument order

anbestephen:
The docker run argument order is `-dit `in most part, bu i found a different order `-idt` here. I believe it is  better to have inconsistent order in the entire documentation. It can be confusing for new-bees if they see these parameters are ordered inconsistently across documentation flow.


joaofnfernandes:
Consistency FTW 🙌

PR #6419, opened 2018-04-10
18.04.0 CE docs release

gbarr01:
- Update product versions in _config.yml

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -13,15 +13,15 @@ safe: false
 lsi: false
 url: https://docs.docker.com
 # This needs to have all the directories you expect to be in the archives (delivered by docs-base in the Dockerfile)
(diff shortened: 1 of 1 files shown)


_config.yml
```diff
@@ -13,15 +13,15 @@ safe: false
...
+keep_files: ["v1.4", "v1.5", "v1.6", "v1.7", "v1.8", "v1.9", "v1.10", "v1.11", "v1.12", "v1.13", "v17.03", "v17.06", "v17.09", "v17.12", "v18.03"]
 exclude: ["_scripts", "apidocs/layouts", "Gemfile", "hooks"]
 
 # Component versions -- address like site.docker_ce_stable_version
 # You can't have - characters in these for non-YAML reasons
 
-docker_ce_stable_version: "18.03"
 latest_stable_docker_engine_api_version: "1.37"
-docker_ce_edge_version: "18.03"
+docker_ce_stable_version: "18.04"
```
joaofnfernandes: Can you confirm that 18.04 is a stable release, and that is not only going to be available on the edge channel?

gbarr01: You're right ... it's an edge.

PR #6428, opened 2018-04-11
Add Windows unset shell variables instructions

AlanI-xx:
Instructions on unsetting docker-machine shell variable settings covered Mac/Linux only. Adding instructions for Windows too.

### Proposed changes

Whilst working through the guide on Windows it took me a few moments to realise what I was supposed to do at this step.  Thought I'd spell it out since the rest of the guide does a good job at breaking down the difference between the Mac/Linux command vs the Windows version.

### Related issues (optional)


joaofnfernandes:
Thanks for the contribution @AlanI-xx.
Let's ship it!

PR #6433, opened 2018-04-12
Take down docker.github.io

### LINKED ISSUE #6101: Take docker.github.io down
We're serving docker.github.io mostly for historical reasons, but we should take it down given that users are still accessing the docs from this property, and this property doesn't do a full docs build.
A bunch of pages are missing, and you can't see previous versions of the docs there.

Users should only have one url - https://docs.docker.com

Update: I thought that this would be a simple setting on GH, but based on [GH docs](https://help.github.com/articles/unpublishing-a-project-pages-site/), it seems that this would require us to rename the repo 😢

allejo:
### Proposed changes

Don't know if you're still interested in taking down the `docker.github.io` domain from #6101, but see my comment https://github.com/docker/docker.github.io/issues/6101#issuecomment-373878079.

This PR does the following:

- Adds a `.nojekyll` so GitHub doesn't build the site as a Jekyll site
- Adds a `index.html` so GitHub has something to fall back on and it'll redirect to `docs.docker.com`
- Adds `index.html` to Jekyll's exclude list so it doesn't interfere with `index.md`
- Abuse GitHub's `404.html` to blindly attempt a redirect to a docs.docker.com URL

Closes #6101 

/cc @joaofnfernandes @thaJeztah


index.html
```diff
@@ -0,0 +1,9 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+  <meta charset="UTF-8">
+  <meta http-equiv="refresh" content="0; url=https://docs.docker.com" />
+  <title>Docker docs</title>
+</head>
+<body></body>
```
thaJeztah: Perhaps this body should also mention that we're redirecting; similar to the 404 page, and provide a link if redirect doesn't kick in (e.g., if you're not being automatically redirected, click here ..)

404.html
```diff
@@ -0,0 +1,14 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+  <meta charset="UTF-8">
+  <title>Docker docs</title>
```
thaJeztah: nit:perhaps `Docker Documentation` instead of "docs"

index.html
```diff
@@ -0,0 +1,9 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+  <meta charset="UTF-8">
+  <meta http-equiv="refresh" content="0; url=https://docs.docker.com" />
+  <title>Docker docs</title>
```
thaJeztah: nit:perhaps `Docker Documentation` instead of "docs"

404.html
```diff
@@ -0,0 +1,14 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+  <meta charset="UTF-8">
```
thaJeztah: Should we have a meta-redirect here as well (perhaps with a 1..2 seconds timeout)?;

```html
<meta http-equiv="refresh" content="1; url=https://docs.docker.com" />
```

404.html
```diff
@@ -0,0 +1,14 @@
...
+  <title>Docker docs</title>
+</head>
+<body>
+    <p>We have moved away from the <strong>docker.github.io</strong> domain. Please visit <a href="https://docs.docker.com">docs.docker.com</a> instead.</p>
+
+    <script>
+        window.location.replace("https://docs.docker.com" + window.location.pathname);
```
thaJeztah: Not sure we want to remove the original URL from history; perhaps use `.assign()` instead?

```javascript
window.location.assign("https://docs.docker.com" + window.location.pathname);
```

404.html
allejo: I put it for 1 second as you suggested; should be enough time for JS to kick in if enabled.

bermudezmt:
@JustinINevill also sounds like a good candidate for discussion during F2F. See #6101 for more context.

bermudezmt:
@allejo I looked into doing a similar redirect for one of the DTR pages but according to http://www.redirect-checker.org/can-i-use-meta-refresh.php and Wikipedia, the meta refresh tag is deprecated and not recommended by W3C and Google. I do find it concerning that `docker.github.io` is **_very_** out of sync with `docs.docker.com`. cc: @thaJeztah @JustinINevill

thaJeztah:
Yes, it's definitely a dirty-hack® / band-aid, but I don't think there's another way currently to remove the docker.github.io domain (see the discussion on https://github.com/docker/docker.github.io/issues/6101)

thaJeztah:
@bermudezmt this LGTY?

allejo:
One idea would be replicating the situation this repo is in:

1. Create a GH org + gh-pages repo (e.g. `sample-org/sample-org.github.io`); see the discussion #6101 to make sure I'm not missing anything
2. Use the `master` branch to generate this Jekyll site
3. Then try out this PR in that repo and ensure the redirect works and it no longer builds as a Jekyll site

bermudezmt:
@allejo thank you for the quick response! Sorry if you're having to repeat information regarding this issue (and PR). I will bring this up again at our next team meeting and let you know how we decide to move forward.

bermudezmt:
Status update 01/14/2019:
Conversed with Github and @peter-docker re this PR. Here's a thoughtful response from Friday Saarelaht at Github which supports the fixes in this PR, with the caveat that `.nojekyll` _might_ affect some site configurations:
```
In the context of GitHub Pages, a "custom domain" refers to any domain other than a GitHub-provided *.github.io subdomain. Looking at the docker/docker.github.io repository, it doesn't appear a custom domain is configured.

If your site is hosted here on GitHub Pages, you can configure docs.docker.com as a custom domain by following our instructions below. If you follow this approach, docker.github.io would then redirect to docs.docker.com automatically.

https://help.github.com/articles/using-a-custom-domain-with-github-pages

However, it doesn't appear docs.docker.com is currently hosted with GitHub Pages. With this in mind, if the goal is to have docker.github.io redirect to an externally-hosted site, then the approach embodied by Pull Request #6433 represent the best available option. It's worth noting that a .nojekyll file disables Jekyll entirely, rendering the _config.yml file changes redundant.

I hope that helps! Let us know if you have any more questions.
```

Next step is for @JustinINevill to check on current site traffic for `docker.github.io`.

allejo:
> It's worth noting that a .nojekyll file disables Jekyll entirely, rendering the _config.yml file changes redundant.

Just as a note, the [`.nojekyll` file is a GH-pages only feature](https://github.com/jekyll/jekyll/issues/6122#issuecomment-308717349); it's set so GH won't compile Jekyll and deploy it anymore. The changes to `_config.yml` are not redundant in *this* case because there's an `index.html` in this PR, which will conflict with the current `index.md` in the repo; `index.html` will take precedence meaning you'd break the **docs.docker.com** homepage if you don't ignore it. So when you build the Jekyll site and host it whereever docs.docker.com is currently being hosted, you want Jekyll to build `index.md` and ignore `index.html` (which is only intended for GH-pages redirects).

JustinINevill:
@allejo thank you very much for your work on this!

@bermudezmt I'm approving the implementation of this based on your testing and the reduced user confusion that will result from removing this now superfluous and not fully working domain.

bermudezmt:
Thank you so much for your patience with this, @allejo . We welcome more web development / Jekyll insights from you!

PR #6435, opened 2018-04-12
Quote port numbers when using short syntax

teohhanhui:
### Proposed changes

Quote port numbers when using short syntax

Comply with our own advice.

> **Note**: When mapping ports in the `HOST:CONTAINER` format, you may experience erroneous results when using a container port lower than 60, because YAML parses numbers in the format `xx:yy` as a base-60 value. For this reason, we recommend always explicitly specifying your port mappings as strings.

https://docs.docker.com/compose/compose-file/#ports

### DIFF
--- a/compose/compose-file/index.md
+++ b/compose/compose-file/index.md
@@ -60,7 +60,7 @@ services:
   vote:
     image: dockersamples/examplevotingapp_vote:before
     ports:
-      - 5000:80
+      - "5000:80"
     networks:
       - frontend
     depends_on:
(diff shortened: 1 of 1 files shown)


gbarr01:
@shin- , agree?

joaofnfernandes:
From what I can see [from the spec](https://github.com/docker/compose/blob/master/compose/config/config_schema_v3.6.json#L207-L226), you can specify ports using:
* A number (integer or floating point)
* A string
* An object with the properties `mode, target, published, protocol`.

So when you use a value like `8000:8000` you're kind of leaving it up to the yaml parser to decide what to do. I can see how some parsers could take that value and match it to the number `8000`, in which case you would not get the behavior you expect.

I don't think it hurts to use quotes and be a bit more verbose, and in the process avoid some headaches.

PR #6444, opened 2018-04-12
Remove references to security scanning for private repos

gbarr01:
Security scanning is only removed for the private repo service. 
It still remains for official images, for publisher’s trying to publish/certify their content, and the EE product.

- Remove image-scan.md and references to it
- Edit repo-general image by removing "and scans"


docker-cloud/builds/image-scan.md
```diff
@@ -1,171 +0,0 @@
----
-description: 'Docker Security Scanning: automatic image scanning'
```
thaJeztah: Should we make this URL redirect to (e.g.) DTR scanning? https://docs.docker.com/datacenter/dtr/2.4/guides/admin/configure/set-up-vulnerability-scans/

joaofnfernandes: Good point. I think it's going to be confusing if I land on a product that's completely unrelated to Docker Cloud.

Maybe during the deprecation period we should leave the article but add a message up in the top linking to the deprecation announcement or something that tells me that security scans are available in our enterprise offer?

gbarr01:
@KickingTheTV , PTAL

KickingTheTV:
@gbarr01 LGTM, although, when I view a preview of it in GitHub, all of the {: target=blank} stuff is showing up next to all of the URLs. I assume all of that will not show up once docs are published?

PR #6448, opened 2018-04-13
Added note to 'ports' section: Port mapping is incompatible with `net…

jgentes:
### Proposed changes

Added note to 'ports' section: Port mapping is incompatible with `network_mode: host`

Took a while to figure this one out.  Source: https://github.com/docker/compose/issues/4799#issuecomment-299605031


joaofnfernandes:
Let's ship it!

PR #6449, opened 2018-04-14
Fix device mapper doc

DanHam:
### Proposed changes

Fixes a few issues with the device-mapper-driver doc. 

Most are fairly minor e.g. typos or simple omissions I've spotted as I've been working through the steps. 

However, the current explanation of how to ensure the logical volume is being monitored is wrong. The `lvs` command is used to *display* info about logical volumes only. To enable/disable monitoring of the LV the `lvchange` command should be used instead.

Happy to squash some of those commits if you think this PR's a bit noisy. I figured a granular approach would be better to start off with in case you wanted to drop some of the changes...


storage/storagedriver/device-mapper-driver.md
```diff
@@ -311,16 +311,28 @@ assumes that the Docker daemon is in the `stopped` state.
...
     ```bash
     $ sudo lvs -o+seg_monitor
 
     LV       VG     Attr       LSize  Pool Origin Data%  Meta%  Move Log Cpy%Sync Convert Monitor
-    thinpool docker twi-a-t--- 95.00g             0.00   0.01                             monitored
+    thinpool docker twi-a-t--- 95.00g             0.00   0.01                             not monitored
     ```
 
+    If the output in the `Monitor` column reports, as above, that the volume is
+    `not monitored` then monitoring needs to be explicitly enabled. Without
```
kolyshkin: a comma is missing before then

kolyshkin:
All LGTM except for a missing comma (disclaimer: I'm not a native speaker).

storage/storagedriver/device-mapper-driver.md
ahh-docker: resolved

ahh-docker:
@L-Hudson PTAL. Ready for merge provided you don't have any edits.

DanHam:
@L-Hudson @ahh-docker @kolyshkin 

A gentle reminder: when you merge, a quick 'Thanks' for the contribution can really go a long way! This makes it much more likely people will bother to contribute again. You can do this really easily with a 'saved reply'.

PR #6450, opened 2018-04-14
ENV section should be before tha RUN in Dockerfile

FelipeRuizGarcia:
Im learning Docker and Im following the get-staterd guide.
Im behind a proxy.

Following the guide such as is document, when I run  `root@felipe# docker build -t happydocker  .  ` I get a error 

Sending build context to Docker daemon  45.75MB
Step 1/14 : FROM python:2.7-slim
 ---> b16fde09c92c
Step 2/14 : WORKDIR /app
 ---> Using cache
 ---> 2842bfbc0a0a
Step 3/14 : ADD . /app
 ---> 23dad6764edc
Step 4/14 : RUN pip install --trusted-host pypi.python.org -r requirements.txt
 ---> Running in e137a0f36fdb
Collecting Flask (from -r requirements.txt (line 1))
  Retrying (Retry(total=4, connect=None, read=None, redirect=None, status=None)) after connection broken by 'ConnectTimeoutError(<pip._vendor.urllib3.connection.VerifiedHTTPSConnection object at 0x7fcf5f021910>, 'Connection to pypi.python.org timed out. (connect timeout=15)')': /simple/flask/

The librarys indicate in requirements.txt file are not installed for error connection : This is a signal that proxy ENV are not passed for the app container yet.
This because in the Dockerfile first execute RUN command and the import the ENV variables.

# Install any needed packages specified in requirements.txt
RUN pip install --trusted-host pypi.python.org -r requirements.txt
# Define environment variable
ENV NAME World
ENV http_proxy host:port

If first indicate ENV variables in the that going to be installed in the container app and then RUN command in the Dockerfile, the image is create succesfully.

### DIFF
--- a/get-started/part2.md
+++ b/get-started/part2.md
@@ -71,14 +71,14 @@ WORKDIR /app
 # Copy the current directory contents into the container at /app
 ADD . /app
 
+# Define environment variable
+ENV NAME World
+
 # Install any needed packages specified in requirements.txt
 RUN pip install --trusted-host pypi.python.org -r requirements.txt
 
 # Make port 80 available to the world outside this container
 EXPOSE 80
-
-# Define environment variable
-ENV NAME World
 
 # Run app.py when the container launches
 CMD ["python", "app.py"]
@@ -179,7 +179,9 @@ friendlyhello         latest              326387cea398
 > Proxy servers can block connections to your web app once it's up and running.
 > If you are behind a proxy server, add the following lines to your
 > Dockerfile, using the `ENV` command to specify the host and port for your
-> proxy servers:
+> proxy servers and make sure that `ENV` section are before that `RUN`, this to
+> set up the variables enviroment before run everything needed.
+> command section :
 >
 > ```conf
 > # Set proxy server, replace host:port with values for your servers
--- a/get-started/part2.md
+++ b/get-started/part2.md
@@ -71,11 +71,11 @@ WORKDIR /app
 # Copy the current directory contents into the container at /app
 ADD . /app
 
-# Define environment variable
-ENV NAME World
-
 # Install any needed packages specified in requirements.txt
 RUN pip install --trusted-host pypi.python.org -r requirements.txt
+
+# Define environment variable
+ENV NAME World
 
 # Make port 80 available to the world outside this container
 EXPOSE 80


get-started/part2.md
```diff
@@ -71,15 +71,15 @@ WORKDIR /app
 # Copy the current directory contents into the container at /app
 ADD . /app
 
+# Define environment variable
+ENV NAME World
```
thaJeztah: I don't think the `$NAME` env-var has to be set before the `RUN`, as this is only used at runtime (when the container is started)

thaJeztah: What we _should_ change though, is the section about setting a proxy through an `ENV` instruction. That's not best practice, because doing so will bake your proxy configuration into the image, which means that everyone that will try to run your image will also be using the proxy you configured (which is likely unique to your environment).

Instead, we should describe:

- using `--build-arg` to set a proxy during build time
- use the `--env` option at runtime
- refer to the `proxy` configuration in the CLI configuration file

ping @gbarr01

Transmigracion: A beginner behind a proxy following the guide its going toget  the error because in these guide does not mention about the option for docker build command behind a  proxy.

I think this should be mentioned in some part of guide.
What you think ?

gbarr01: Hi, @thaJeztah , 

I'm struggling to understand the use of ARG. I tested its use with the `NAME` variable and it's not working as I expect.  The docs say: _An ARG instruction can optionally include a default value... If an ARG instruction has a default value and if there is no value passed at build-time, the builder uses the default._  I'm not seeing this. 

For example, this code does _not_ display "Moby" but "None". 

*Dockerfile*
`ARG NAME="Moby"`

*Build and Run*
```
docker build --build arg NAME  --tag hellomoby:v1 .
docker run --publish 4000:80 hellomoby:v1
```

When I tweak the build-time to `docker build --build arg NAME="Moby" --tag hellomoby:v2 .`, it's still "None".

Only when I run, `docker run --env NAME="Moby" --publish 4000:80 hellomoby:v2`, does it display as "Moby". And when I remove `ARG` from `Dockerfile` altogether, with the same `run` command, it also displays as "Moby." 

I cannot see how `ARG` is having any effect here. Can you see what I'm going wrong? Seems like using `--env` is good enough.

thaJeztah: @gbarr01 let me catch up with you IRL to try and explain (probably faster than writing all down)

ahh-docker:
@thaJeztah can you please let me know if there's any more changes with this? Let me know and I'll get it updated. Thanks!

ahh-docker:
@thaJeztah Can I get a look at this again? Need to know if this is LGTM or a thumbs down. Thanks!

get-started/part2.md
```diff
@@ -74,12 +74,12 @@ ADD . /app
 # Install any needed packages specified in requirements.txt
 RUN pip install --trusted-host pypi.python.org -r requirements.txt
 
-# Make port 80 available to the world outside this container
-EXPOSE 80
-
 # Define environment variable
 ENV NAME World
 
+# Make port 80 available to the world outside this container
```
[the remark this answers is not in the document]
paigehargrave: How about 'Publicizes that the container is listening on port 80'?

paigehargrave:
@thaJeztah - Hi Sebastian,  I've attempted to wordsmith something to cover the use of --build-arg, can you take a look and see what else is needed? Thanks! (Please ignore any formatting SNAFUs - just trying to nail down content.)

Suggested rewrite of Proxy server settings info:

 Proxy servers can block connections to your web app once it's up and running.
 If you are behind a proxy server, you can specify proxy server host and port information in one of the following ways:

1) Use  the `build-args` command in your Dockerfile for build-time argument passing.

docker build --build-arg HTTP_PROXY=http://myyproxy.example.com

The proxy information is only available during the docker build itself, and doesn’t persist in the resulting image (including leaking through docker image history).
This is similar to how docker run -e works. Refer to the docker run documentation f additional details. 
You can use the --build-arg flag without a value, in which case the value from the local environment is propagated into the Docker container being built.
Using this flag will not alter the output you see when the ARG lines from the Dockerfile are echoed during the build process. The ARG instruction lets Dockerfile authors define values that users can set at build-time using the  --build-arg flag. See the Dockerfile reference for more information about ARG.

2)Set defaults in the CLI configuration file. (Should we include this? If so, can we include an example?)

Warning: Using ENV instructions in a Dockerfile to set proxy-server values in the image is  not recommended, because those variable values persist in the intermediate and final images. This is not desirable for the following reasons:
The resulting image can only be run in the environment in which it is built. 
Information about the proxy-server can leak security-sensitive information (name of the proxy-server, or even the name and password if the proxy server requires authentication).

bermudezmt:
@thaJeztah I know this PR has gotten unwieldy. :laughing: Can you please review comment from @paigehargrave (https://github.com/docker/docker.github.io/pull/6450#issuecomment-455350028) which  hopefully addresses your requested changes? Thanks!

L-Hudson:
> @thaJeztah I know this PR has gotten unwieldy. 😆 Can you please review comment from @paigehargrave ([#6450 (comment)](https://github.com/docker/docker.github.io/pull/6450#issuecomment-455350028)) which hopefully addresses your requested changes? Thanks!

Nudged in Slack this morning.

thaJeztah:
> You can use the --build-arg flag without a value

That one might need an example, to illustrate what we mean with "without a value";

```bash
export HTTPS_PROXY=https://proxy.corp.example.com:8080

docker build --build-arg HTTPS_PROXY  ......
```

> 2)Set defaults in the CLI configuration file. (Should we include this? If so, can we include an example?)

I'm thinking if we should include all this information (about proxies) in the "getting started". The information is valuable, but overall, it seems a bit out of place for a "getting started".

There's an existing section on proxy configuration (for the client) here: https://docs.docker.com/network/proxy/, but it may be hard to find ("configure the daemon and client"). Thinking of dedicating a more standalone section about that, to describe daemon configuration, client configuration, as well as setting proxies at runtime and for build.

Other parts in the documentation can link to that section so that we don't have to repeat the information.

paigehargrave:
@thaJeztah Hi Sebastian - I created #8352 to try to capture your suggested changes. Hopefully what I did is a reasonable start. PTAL and many thanks again for your input! 

 It looks like additional information is needed for:

* use the `--env` option at runtime
* refer to the `proxy` configuration in the CLI configuration file

bermudezmt:
Closing in favor of #8352 .

PR #6451, opened 2018-04-14
"apt-get install" -> "apt install"

muxator:
`apt` is available by default since Ubuntu 14.04.

The reference to `apt-cache madison` command is left as-is, since there is no apt command with an equivalent output format.

### Proposed changes

Modify the Ubuntu installation guide to use `apt` (available since 2014) instead of `apt-get`

### DIFF
--- a/install/linux/docker-ce/ubuntu.md
+++ b/install/linux/docker-ce/ubuntu.md
@@ -44,10 +44,10 @@ Older versions of Docker were called `docker` or `docker-engine`. If these are
 installed, uninstall them:
 
 ```bash
-$ sudo apt-get remove docker docker-engine docker.io
+$ sudo apt remove docker docker-engine docker.io
 ```
 
-It's OK if `apt-get` reports that none of these packages are installed.
+It's OK if `apt` reports that none of these packages are installed.
 
 The contents of `/var/lib/docker/`, including images, containers, volumes, and
 networks, are preserved. The Docker CE package is now called `docker-ce`.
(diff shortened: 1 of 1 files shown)


gbarr01:
Seems right to me. @joaofnfernandes , do you agree?

thaJeztah:
We should double check if it's also installed on cloud instances; given that `apt` is meant to target humans, such tools may not always be present on light-weight, and server installs

gbarr01:
Also, Ubuntu 14 and Debian 7 don't support `apt`. Even so, I'm thinking that we can update the command and warn users to use `apt-get` if `apt` is not supported.

muxator:
I have done a quick check with the images on the Docker public registry.

Indeed, for debian 7 the command is **not** there:
```
$ docker run -ti --rm debian:7 bash
Digest: sha256:d949cbb93a58de27eec4c911f27b9f09edeb3d3ce57cf5ce77d4745211c947f6
Status: Downloaded newer image for debian:7
root@7c7031174629:/# apt
bash: apt: command not found
```

While on ubuntu 14.04 `apt` seems to be available:
```
$ docker run -ti --rm ubuntu:14.04 bash
[...]
Digest: sha256:b92dc7814b2656da61a52a50020443223445fdc2caf1ea0c51fa38381d5608ad
Status: Downloaded newer image for ubuntu:14.04
root@0f6347951196:/# apt
apt 1.0.1ubuntu2 for amd64 compiled on Dec  8 2016 16:23:37
```

I have no easy access to cloud instances.

This being a counter-example already, maybe the change should be retired, or rephrased like @gbarr01 was suggesting.

thaJeztah:
I think we should wait for a bit until it's available everywhere; adding extra words to explain users to use `apt-get` instead may overcomplicate the instructions for a minor gain.

muxator:
Agreed, close the pull request.

PR #6463, opened 2018-04-17
Typo

ushuz:
### Proposed changes

Typo

- Correct option name for MTU is `com.docker.network.driver.mtu`
- `ip link del name` -> `ip link del dev`
```
$ sudo ip link del name docker_gwbridge
Not enough information: "dev" argument is required.
```

### DIFF
--- a/network/overlay.md
+++ b/network/overlay.md
@@ -153,7 +153,7 @@ services which publish ports, such as a WordPress service which publishes port
       --ingress \
       --subnet=10.11.0.0/16 \
       --gateway=10.11.0.2 \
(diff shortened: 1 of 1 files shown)


joaofnfernandes:
I can confirm that the driver option matches what's documented in the [network reference docs](https://docs.docker.com/engine/reference/commandline/network_create/#network-internal-mode), and the change to the `ip` command matches what's documented in the [ip command man pages.](http://www.dsm.fordham.edu/cgi-bin/man-cgi.pl?topic=ip-link)

PR #6468, opened 2018-04-17
Update release-notes.md

davidswu:
Add instructions in case dtr-scanning store gets in a restart loop
This is due to a major postgres version upgrade which was required with the upgrade to alpine 3.7

### DIFF
--- a/ee/dtr/release-notes.md
+++ b/ee/dtr/release-notes.md
@@ -108,7 +108,7 @@ specify `--log-protocol`.
   docker volume rm dtr-postgres-$replica
   docker run --rm -it docker/dtr:2.5.0 reconfigure ...
   ```
-  Then resync the vulnerability database as normal
+  Then resync the vulnerability database from the web UI.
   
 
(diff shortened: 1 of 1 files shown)


gbarr01:
cc @joaofnfernandes

ee/dtr/release-notes.md
```diff
@@ -95,6 +95,21 @@ specify `--log-protocol`.
...
+  which is not compatible with this version 10.3.
+  ```
+  To remedy this remove the scanningstore container and the dtr-postgres volume, and perform a reconfigure
+  ```
+  docker stop dtr-scanningstore-$replica
+  docker rm dtr-scanningstore-$replica
+  docker volume rm dtr-postgres-$replica
+  docker run --rm -it docker/dtr:2.5.0 reconfigure ...
+  ```
+  Then resync the vulnerability database as normal
```
joaofnfernandes: s/as normal/from the web UI./
I think that would be clearer.

PR #6472, opened 2018-04-18
updating bash_profile

estrellajm:
### Proposed changes

Consolidating some useful information. The following lines have to be added to the ~/.bash_profile for it to work. I pulled this information from https://docs.docker.com/compose/completion/. 

I thought it would be useful to add it here so that all this info could be in one place.
`
if [ -f $(brew --prefix)/etc/bash_completion ]; then
. $(brew --prefix)/etc/bash_completion
fi
`
If not the above, the other command could be 
`
 [ -f /usr/local/etc/bash_completion ] && . /usr/local/etc/bash_completion
`

which is displayed once you've installed bash-completion


paigehargrave:
@shin - PTAL - thank you!

shin-:
`area/compose` label is misleading I think - this is one for the Desktop team :) That said, LGTM

PR #6483, opened 2018-04-19
Fix broken references to DTR architecture docs

axeal:
### Proposed changes

Fix broken references to DTR architecture docs


joaofnfernandes:
Thanks for fixing this!

PR #6496, opened 2018-04-20
Update dotnetcore.md

Frosthage:
Starting with .Net Core 2.0 restore is executed implicitly by all dotnet commands. Without the --no-restore flag, manually copying and restoring csproj files is redundant. 

https://docs.microsoft.com/en-us/dotnet/core/tools/dotnet-publish?tabs=netcore2x


Frosthage:
Nevermind, obviously the nuget restore is cached.

PR #6506, opened 2018-04-20
Add note about ucp/interlock app traffic

joaofnfernandes:
When upgrading from HRM to Interlock, UCP respects your previous configs. If each application had a dedicated network for routing, that will continue being the case after the upgrade.

However there are a few things users can do to clean up their apps and networks after the upgrade. This PR explains it.

### DIFF
--- a/ee/ucp/interlock/upgrade.md
+++ b/ee/ucp/interlock/upgrade.md
@@ -109,3 +109,18 @@ As part of the upgrade process UCP adds the
 
 You can update your services to remove the old HRM labels, since they won't be
 used anymore.
+
+## Optionally segregate control traffic
+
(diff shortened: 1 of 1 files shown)


mbentley:
Some clarification - the only service you have to care about is the `ucp-interlock` service.  It will automatically control the `ucp-interlock-extension` service for you.  If you move `ucp-interlock` to it's own network, it will automatically update `ucp-interlock-extension`.  Basically you should only ever manage `ucp-interlock` as an actual service.  The other two are managed by the `config.toml` stored in a `docker config`, spawned and updated by `ucp-interlock`.

joaofnfernandes:
Addressed feedback. Shipping it since this is affecting customers in the field and we can always tweak it later.

PR #6507, opened 2018-04-21
Error when try to adding gpg on debian 9

alpmoi:
Error :
gpg: no valid OpenPGP data found.

It works with my solution

### DIFF
--- a/install/linux/docker-ce/debian.md
+++ b/install/linux/docker-ce/debian.md
@@ -137,7 +137,7 @@ from the repository.
 3.  Add Docker's official GPG key:
 
     ```bash
-    $ curl -fsSL {{ download-url-base}}/gpg | sudo apt-key add -
+    $ wget {{ download-url-base}}/gpg | sudo apt-key add gpg
     ```
 
     Verify that you now have the key with the fingerprint


install/linux/docker-ce/debian.md
```diff
@@ -137,7 +137,7 @@ from the repository.
 3.  Add Docker's official GPG key:
 
     ```bash
-    $ curl -fsSL {{ download-url-base}}/gpg | sudo apt-key add -
+    $ wget {{ download-url-base}}/gpg | sudo apt-key add gpg
```
thaJeztah: I think we standardize on using curl, also, the `-fsSL` options make sure that only a TLS connection is used (not sure if `wget` does the same by default);

What do you see if you only attempt to download the gpg file with curl? Does it produce some error? Could it be something in your environment (e.g. a corporate proxy)?

gbarr01: I see no problem with the existing command: 

```
$ curl -fsSL https://download.docker.com/linux/debian/gpg | sudo apt-key add -
OK

$ lsb_release -d
Description:	Debian GNU/Linux 9.4 (stretch)
```

gbarr01: @thaJeztah , I'm guessing that @alpmoi is behind a proxy.

thaJeztah: If the proxy environment is properly setup (`HTTP_PROXY` etc environment variables), those _should be_ automatically selected by `curl`. If there's a MITM proxy, perhaps it's failing because the certificate is invalidated by curl (if things are not properly setup)

ahh-docker:
@thaJeztah can you please let me know if this needs fixing or is ready to go? thanks!

thaJeztah:
It's unclear what the change is fixing, and we're unable to reproduce the problem, so I would suggest closing this PR for now

PR #6509, opened 2018-04-22
Update storage driver docs

ittus:
### Proposed changes

I think using `image size` is easier to understand for reader.


joaofnfernandes:
Sounds good. Let's ship it!

PR #6517, opened 2018-04-23
Fix JSON example on webhooks page

johanbrandhorst:
### Proposed changes
Correct the JSON example on the Webhooks page. The comment_count parameter in the JSON body is a number, not a string.

### DIFF
--- a/docker-hub/webhooks.md
+++ b/docker-hub/webhooks.md
@@ -30,7 +30,7 @@ with the following payload:
     "tag": "latest"
   },
   "repository": {
-    "comment_count": "0",
+    "comment_count": 0,
     "date_created": 1.417494799e+09,
     "description": "",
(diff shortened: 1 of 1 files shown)


joaofnfernandes:
@lilybguo can you confirm?

lilybguo:
@joaofnfernandes I found this bug during one of the testing sessions where we used testkit clusters. I don't know how to load unmerged code onto a cluster and verify. can someone else verify it?

manishtomar:
@lilybguo I believe this about hub webhook documentation. Did you misunderstand to be something else?

lilybguo:
@manishtomar Ha I'm sorry, I was also looking at an orca bug, confused the two.

I don't have any context on this bug. The fix looks good, but I don't think I'm the right person to confirm

joaofnfernandes:
Thanks @manishtomar. Let's ship it!

PR #6535, opened 2018-04-24
Expose the server port changed to 32769

stshow:
Port 30080 is an invalid port in UCP for kubernetes. The valid range is 32768-35535. Replicas updated to match forthcoming screenshots.

### DIFF
--- a/ee/ucp/kubernetes/index.md
+++ b/ee/ucp/kubernetes/index.md
@@ -24,7 +24,7 @@ spec:
   selector:
     matchLabels:
       app: nginx
-  replicas: 2
+  replicas: 3
   template:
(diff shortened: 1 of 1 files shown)


joaofnfernandes:
It seems you've got conflicts. Do you mind rebasing your PR?

joaofnfernandes:
As far as I can tell from the [Kubernetes docs](https://kubernetes.io/docs/concepts/services-networking/service/#type-nodeport), NodePorts need to be in the range 30000-32767.

Given that the doc seems to be correct and that this PR has conflicts, I'm closing this due to inactivity.
Feel free to submit a new PR if the docs are not working for you.

PR #6559, opened 2018-04-26
Fix download urls

joaofnfernandes:


docker-for-mac/install.md
```diff
@@ -4,82 +4,15 @@ keywords: mac, beta, alpha, install, download
...
-explanation and list of prerequisites.
-
-* You can switch between Edge and Stable versions, but you must have only one
-   app installed at a time. Also, you need to save images and export
-   containers you want to keep before uninstalling the current version before
-   installing another. For more about this, see the
-   [FAQs about Stable and Edge channels](faqs.md#stable-and-edge-channels).
+Docker for Mac is the
+[Community Edition (CE)](https://www.docker.com/community-edition)
+of Docker for MacOS. If you already have Docker for Mac installed, skip to
```
gbarr01: If they already have it installed, why would they even be on this page? For comparison, I ripped it out here: https://docs.docker.com/docker-for-windows/install/

joaofnfernandes: Was been conservative because I thought we had added this info recently. I'll kill that part too 🎉

PR #6568, opened 2018-04-27
move mount csv note from services to volumes

trapier:
this callout has more relevance to the general context of volumes than it does to services

note: this one may need to be rebased if #6563 goes in first, or vice versa.


storage/volumes.md
```diff
@@ -81,6 +81,20 @@ If you need to specify volume driver options, you must use `--mount`.
   - The `volume-opt` option, which can be specified more than once, takes a
     key-value pair consisting of the option name and its value.
 
+> **Important:** If your volume driver accepts a comma-separated list as an option,
```
trapier: not sure about indentation here.  let's take a look at the CI build before merging.

PR #6573, opened 2018-04-27
ucp haproxy config: add timeout tunnel

### LINKED ISSUE #6572: ucp haproxy timeout incompatible with DTR
### Problem description
Suspect the [example haproxy config](https://docs.docker.com/ee/ucp/admin/configure/join-nodes/use-a-load-balancer/#configuration-examples) will kill the DTR bootstrapper after 50 seconds of `wait`.  

### Problem location

https://docs.docker.com/ee/ucp/admin/configure/join-nodes/use-a-load-balancer/#configuration-examples

UCP: 3.0.0
DTR: 2.5.0

### Suggestions for a fix
Figure out the knob to set to a longer timeout and patch the doc.

trapier:
also taking the opportunity to use human-readable units

closes #6572

### DIFF
--- a/ee/ucp/admin/configure/join-nodes/use-a-load-balancer.md
+++ b/ee/ucp/admin/configure/join-nodes/use-a-load-balancer.md
@@ -78,9 +78,11 @@ global
 defaults
         mode    tcp
         option  dontlognull
-        timeout connect 5000
-        timeout client 50000
-        timeout server 50000
+        timeout connect     5s
+        timeout client      50s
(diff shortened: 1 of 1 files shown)


trapier:
@mbentley: stole these numbers from https://github.com/mbentley/docker-in-docker/blob/master/scripts/haproxy/entrypoint.sh#L69-L73

can i get a lgtm?

mbentley:
Indeed, LGTM!

PR #6580, opened 2018-04-28
Add exec_die to the list of known container events

### LINKED ISSUE #6579: Missing exec_die event in event docs
Heya!

After some digging around into my `docker events` I found out that my containers are producing a couple of `exec_die` events that aren't listed in the docs. I found a PR that introduces the event here: https://github.com/moby/moby/pull/35744

Am I correctly assuming that this is an event we should add to the list?  

[engine/reference/commandline/events.md](https://docs.docker.com/engine/reference/commandline/events/), CC @gbarr01

nstapelbroek:
Hey there!

This could be an implementation of https://github.com/docker/docker.github.io/issues/6579

### Proposed changes

Since https://github.com/moby/moby/pull/35744 your containers could also dispatch an `exec_die` event. Juding by the pull request, I'm assuming this behavior is intended and should be listed in the "known container events" list.

### Related issues (optional)

Fixes #6579

### DIFF
--- a/_data/engine-cli/docker_events.yaml
+++ b/_data/engine-cli/docker_events.yaml
@@ -19,6 +19,7 @@ long: |-
   - `die`
   - `exec_create`
   - `exec_detach`
+  - `exec_die`
   - `exec_start`
   - `export`
   - `health_status`


gbarr01:
Thanks for filing this, @nstapelbroek. CLI references actually have to be fixed upstream in the cli docs ... from there they they get propagated in the docker-ce docs and finally, docs. I've filed that here: https://github.com/docker/cli/issues/1027

nstapelbroek:
Ah, my bad. I've submitted another PR to the CLI repository. I think we can close this. Thanks @gbarr01

PR #6587, opened 2018-04-30
Add port publish example for particular host IP

### LINKED ISSUE #6478: Can we add document about -p HostIP:HostPort:ContainerPort?
[config/containers/container-networking.md](https://docs.docker.com/config/containers/container-networking/), CC @gbarr01

It is actually working

emanb29:
### Proposed changes
Added an example usage for -p showcasing binding to a particular host IP

### Related issues (optional)
Fixes #6478


joaofnfernandes:
Thanks for the fix! Let's ship it 🚀

PR #6592, opened 2018-04-30

portolkyz:
In Ubuntu memory-swappiness can equal -1 even when swapping is possible. 
See https://askubuntu.com/questions/1028376/docker-containers-memory-swappiness-value-outside-of-documented-range


joaofnfernandes:
It seems that `-1` was used as a default value, but that changed. While the daemon still supports `-1` for backwards compatibility, you should set a value between `0-100`.

You can check https://github.com/moby/moby/pull/33898 for more context.

For that reason, I'm closing this PR but feel free to comment.

portolkyz:
Your reaction is perfectly fine. I suspected the "-1" to be an artifact but could find no information on the subject at all. Thus thanks for the linked issue, too.

PR #6593, opened 2018-04-30
Mac oslog

akimd:
### Proposed changes

- remove unused images (many were named settings-*, but we actually used menu/prefs-*)
- add a small tool to help with it.  Use it with caution: it does not try to check if the image is used elsewhere.
- update the way to get logs on Mac.

Applicable now.


akimd:
@gbarr01 @gtardif PTAL.

README.md
```diff
@@ -278,11 +279,39 @@ Bootstrap JS are loaded.
 
 > **Note**: In general, this is a bad idea.
 
+### Images
+
+Don't forget to remove images that are no longer used.  Try to keep the images
+sorted in the local `images/` directory, with names that naturally group related
+images together in alphabetical order; for instance prefer
+`settings-file-share.png` and `settings-proxies.png` to
+`file-share-settings.png` and `proxies-settings.png`.
```
joaofnfernandes: This is a nice convention! The one I've been using with EE docs is to name images after the article they are supposed to be used and append a number, signifying the order in that article.

Something like `install-ubuntu-1.png`, `install-ubuntu-2.png`.
You could argue that this convention doesn't make sense if you're trying to reuse images, but from experience most images are only used in a single article.

README.md
```diff
@@ -278,11 +279,39 @@ Bootstrap JS are loaded.
...
+
+On Mac, when applicable capture windows rather than capture rectangular regions;
+this eliminates unpleasant background appearing in the documentation, and saves
+the editors the need to crop.  Capture windows without shadows.  To this end,
+once you pressed `Command-Shift-4`, press Option while clicking on the window.
+To disable shadows once for all, run
```
joaofnfernandes: Since you're helping us a lot with docs let me give you a bit of feedback about styleguide so that we are consistent across the board.

* Whenever you think about using `;` just use `.`.
* End the sentence before a snippet with `:`. In this case `To disable ..., run:`

README.md
```diff
@@ -278,11 +279,39 @@ Bootstrap JS are loaded.
...
+
+```bash
+$ defaults write com.apple.screencapture disable-shadow -bool TRUE
+$ killall SystemUIServer  # restart it.
+```
+
+(you can restore shadows later with `-bool FALSE`).
```
joaofnfernandes: More feedback about styleguide. Overall I try to avoid parens:
* If a sentence is important it should not be inside a parens
* If a sentence is going on a tangent or is not that important, just scrap it

I understand why you would use parens, but really good writing is concise and to the point. That means that there's no place for going on tangents or half-important comments.

But if you reeeeeally want to you parens, never start a sentence with them, like you are doing here.

docker-for-mac/Makefile
```diff
@@ -0,0 +1,10 @@
+.PHONY:
```
joaofnfernandes: ❤️

docker-for-mac/troubleshoot.md
```diff
@@ -27,73 +27,66 @@ the log data.
 Choose ![whale menu](images/whale-x.png){: .inline} -->
 **Diagnose & Feedback** from the menu bar.
 
-![Diagnose & Feedback](images/diagnose-feedback.png)
+![Diagnose & Feedback](images/diagnose-feedback.png){:width="600px"}
```
joaofnfernandes: The CSS for the entire site should make images responsive. Is there anything we can improve on the site so that you don't have to do this?

README.md
akimd: Hi Joao.  Thanks a lot for your kind words!

Yes, I'll add that conventions too.  In the current case, I still prefer named suffixes, as there's no idea of sequence at all, but I do get your point.

README.md
akimd: Thanks for these tips, and please, do continue teaching me :)

README.md
akimd: Got it.

docker-for-mac/troubleshoot.md
akimd: Actually I was just doing something I saw Gwen do several times (for instance https://github.com/docker/docker.github.io/commit/fd753e5b275#diff-f0f81f92bf2604019320309498f91445R129 and below), so I thought it would help you guys.  And I was expecting this to be only a visual change, not anything related to bandwidth (or maybe I misunderstood what you meant).

joaofnfernandes: Yes, this is only a visual change. This syntax will generate the following html:
```
<img src="images/diagnose-feedback.png"  alt="Diagnose & Feedback" width="600px">
```
With the CSS we currently use in the docs site, you'll get this behavior:
* The maximum width for the image is 600px
* If the browser window is smaller than that, scale down the image

The images in these articles look great, so my question was about whether there's any change we could make to the CSS to make this the default behavior, so that you wouldn't have to add `{: width="600px"}` for most of the images.

Unfortunatelly I looked a bit more into this and don't think we could have better CSS styles. I've tried a couple of alternatives, but all of them have side-effects that I don't really like.

akimd:
What is needed for this PR to be merged?

joaofnfernandes:
Thanks for the reminder @akimd, as I'd lost this from my radar. Let's ship it!

PR #6602, opened 2018-04-30
Mention trick to check startup order.

alnorris:
Much easier way to check a service has started.

### DIFF
--- a/compose/startup-order.md
+++ b/compose/startup-order.md
@@ -76,7 +76,27 @@ script:
     ```none
     command: ["./wait-for-postgres.sh", "db", "python", "app.py"]
     ```
+- You can also use no tool at all and include the check in the command.
 
+    ```
+        services:
+          practice_docker: 
+            image: dockerhubusername/practice_docker
+            ports: 
+              - 80:3000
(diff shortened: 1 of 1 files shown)


compose/startup-order.md
```diff
@@ -76,7 +76,27 @@ script:
...
     ```
+- You can also use no tool at all and include the check in the command.
 
+    ```
+        services:
+          practice_docker: 
+            image: dockerhubusername/practice_docker
+            ports: 
+              - 80:3000
+            command: bash -c 'while !</dev/tcp/db/5432; do sleep 1; done; python app.py'
```
joaofnfernandes: I like your proposal but I seems that `/dev/tcp` is not available in all distributions. I was thinking maybe

```
while !(nc -v -z server 4321); do echo "sleeping"; sleep 1; done; python app.py
```
But `nc/dns-utils` is also not available in a bunch of distros.

Do you have any other ideas?

paigehargrave:
@shin-  PTAL - thanks!

paigehargrave:
Thank you for your feedback! We have received and reviewed your changes. After a technical review, we did not incorporate your suggested changes. If you have additional concerns, submit a new issue in this repository.

PR #6604, opened 2018-05-01
Change calicoctl version to 3.x for 'Disable IP in IP overlay...'

xinfengliu:
calicoctl 1.6.x does not work for this doc in UCP3.0 environment, suggest to use Calico 3.x.
Note: the ippool yaml spec changed from v1 to v3, I'm not sure if only modifying "ipipMode: Never" is correct, please confirm with folks of network team.


alexmavr:
cc @abhi

joaofnfernandes:
This has been sitting in here for a while. This sounds right so let's ship it.

abhi:
sorry I missed this. What is this used for ? What are we getting out disabling ipip. There should be follow up instructions on how data path connectivity will be achieved right ?

PR #6605, opened 2018-05-01
Add '--pod-cidr' and '--cloud-provider' in ucp install options

xinfengliu:
To match with the output of  "ucp install --help". The "--pod-cidr" is important to let the user choose CIDR when the default pod IP range 192.168.0.0/16 has conflicts to the nodes' network.


joaofnfernandes:
Thanks for fixing this!

PR #6617, opened 2018-05-02
Adding information about the CloudSigma Docker Machine driver

bogdan-ds:
I'm adding an entry to the table which includes the details for the CloudSigma Docker Machine driver.


joaofnfernandes:
Let's ship it!

PR #6619, opened 2018-05-02
Clarify that --existing-config uses latest config

squizzi:
### Proposed changes

`--existing-config` is a flag that uses the latest UCP config found in the environment and does not accept a config as a flag variable.  The current documentation seems to suggest that this flag can accept a given config, when in reality it cannot.  This should help clarify that.


squizzi:
I should add I also plan to open a PR for UCP with these proposed changes to clarify this in the `install --help` text.

joaofnfernandes:
Makes sense, thanks for the contribution!

PR #6622, opened 2018-05-03
Rename dockertrial.com URLs to trial.docker.com

jmorganca:
Update Hosted Trial URL to correct URL


joaofnfernandes:
Good catch! Thanks for fixing it.

PR #6624, opened 2018-05-03
Fix typo in docs

SandraKriemann:
I assume blacklist was meant here. (instead of whitelist)


gbarr01:
Hi, @SandraKriemann , actually, it's a whitelist but the text is confusing here.

PR #6637, opened 2018-05-04
Fix -f flag in docker stack deploy

drndos:
### Proposed changes

Fix flag in example for multiple docker compose files in command docker stack deploy
Because option "-f" does not work.


joaofnfernandes:
@drndos thanks for trying to fix this. The source for these particular pages is managed in the docker/cli repository, so you would have to submit a pull request there.

But it seems that someone already fixed it there - docker/cli#981

I'll close this PR, but I'm sure you'll have tons of other opportunities to fix the docs!

PR #6646, opened 2018-05-06
Fix install link to EE

martencassel:
Fix install link to EE

### Proposed changes

Link to Docker EE and not Docker CE


gbarr01:
Fixes #6676

PR #6665, opened 2018-05-09
Remove 'docker-datacenter' limitation for EE2.0 DCT

xinfengliu:
In EE 2.0 `Run only the images you trust` , you can select team from the organization other than `docker-datacenter`.


joaofnfernandes:
Thanks for fixing this!

PR #6668, opened 2018-05-09

mattiasberge:
corrects information about quota. Its not about guarantee, it's about a hard limit before throttling.

### Proposed changes

The information about the quota states that it is a guarantee for resources, just like the shares, while in fact it's a quota that can be used per time period before hitting the ceiling and being throttled.

Reference:

As seen in https://www.kernel.org/doc/Documentation/scheduler/sched-bwc.txt, it's a burst and not guarantee. 

"2. Limit a group to 2 CPUs worth of runtime on a multi-CPU machine.

	With 500ms period and 1000ms quota, the group can get 2 CPUs worth of
	runtime every 500ms.

	# echo 1000000 > cpu.cfs_quota_us /* quota = 1000ms */
	# echo 500000 > cpu.cfs_period_us /* period = 500ms */

	The larger period here allows for increased burst capacity.
"


gbarr01:
@thaJeztah , do you agree?

joaofnfernandes:
This seems to be consistent with the Linux kernel docs, so let's ship it.
@mattiasberge With this PR I think it's still ambiguous what happens after the process is throttled. By reading that line I wouldn't be able to tell.

From the [kernel docs](https://www.kernel.org/doc/Documentation/scheduler/sched-bwc.txt)

> the tasks belonging to its hierarchy will be throttled and are not allowed to run again until the next period.

So consider submitting a new PR that include this info.

PR #6669, opened 2018-05-09
Update part2.md

davidkpiano:
For starting a new project, the 2nd and 3rd steps in the `Dockerfile` are out of order and produce this error:

```
Could not open requirements file: [Errno 2] No such file or directory: 'requirements.txt'
```

Switching the steps works as expected.

### DIFF
--- a/get-started/part2.md
+++ b/get-started/part2.md
@@ -65,11 +65,11 @@ your new Dockerfile.
 # Use an official Python runtime as a parent image
 FROM python:2.7-slim
 
-# Set the working directory to /app
-WORKDIR /app
-
 # Copy the current directory contents into the container at /app
 ADD . /app
+
+# Set the working directory to /app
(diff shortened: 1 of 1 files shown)


gbarr01:
Hi, @davidkpiano , thanks for the PR. Actually, the order is correct. 

The [WORKDIR](https://docs.docker.com/engine/reference/builder/#workdir) instruction creates and sets `/app` as your working directory, and the [ADD](https://docs.docker.com/engine/reference/builder/#add) instruction adds the contents of your current directory (`.`) to `/app`:

```
# Set the working directory to /app
WORKDIR /app

# Copy the current directory contents into the container at /app
ADD . /app
```

You error message is complaining `requirements.txt` doesn't exist. Is it in your current directory? If so, is it named properly?

gbarr01:
If you have questions, feel free to comment here.

PR #6678, opened 2018-05-10
Update use-a-web-proxy.md

lenestopage:
### Proposed changes

I recently ran into a case where the user was using a proxy that required authentication. In order to get this to work we submitted the URL as such: username:password@domain:port for the --http-proxy and --https-proxy flags. This information was not in the docs.


joaofnfernandes:
Thanks for the contribution. Let's merge it, but next time avoid using `Note:` in your sentences.

PR #6688, opened 2018-05-14
Updating the minimum memory step 4GB->2GB

vin-yu:
SQL Server can now run on a minimum of 2GB now

https://docs.microsoft.com/en-us/sql/linux/sql-server-linux-setup?view=sql-server-linux-2017#system


joaofnfernandes:
Thanks for updating this. Let's ship it.

PR #6694, opened 2018-05-14
Replace 'less' with 'fewer'

arosequist:
### Proposed changes

Sorry for the pedantry, but I think "typo" is a count noun, so it reads better to use "fewer" instead of "less".


joaofnfernandes:
Thanks for making our docs better! Let's ship it 🚀

PR #6698, opened 2018-05-15
Clarify which manager to run the join command on

tedivm:
### Proposed changes

This is a simple change to make it clear which machine the join command will need to be run on.


joaofnfernandes:
Let's ship it 🚀

PR #6704, opened 2018-05-16
Fix typo in proxy.md

pallxk:
### Proposed changes

`--env` environment variable name and value should be combined with `=`.


joaofnfernandes:
Thanks for noticing this. Let's ship it!

PR #6706, opened 2018-05-16
correct a mistake

bochen2014:
you can't leave off `target` as it's a mandatory field;

`docker service create --name my-web --publish published=8080,mode=host --replicas=3  nginx `
invalid argument "published=8080,mode=host" for "-p, --publish" flag: missing mandatory field "target"
See 'docker service create --help'.

### DIFF
--- a/engine/swarm/ingress.md
+++ b/engine/swarm/ingress.md
@@ -171,7 +171,7 @@ in mind.
 - If you expect to run multiple service tasks on each node (such as when you
   have 5 nodes but run 10 replicas), you cannot specify a static target port.
   Either allow Docker to assign a random high-numbered port (by leaving off the
-  `target`), or ensure that only a single instance of the service runs on a
+  `published`), or ensure that only a single instance of the service runs on a
   given node, by using a global service rather than a replicated one, or by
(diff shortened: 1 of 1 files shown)


engine/swarm/ingress.md
```diff
@@ -171,7 +171,7 @@ in mind.
 - If you expect to run multiple service tasks on each node (such as when you
   have 5 nodes but run 10 replicas), you cannot specify a static target port.
   Either allow Docker to assign a random high-numbered port (by leaving off the
-  `target`), or ensure that only a single instance of the service runs on a
+  `published`), or ensure that only a single instance of the service runs on a
```
gbarr01: This, doesn't look right but @thaJeztah can you confirm?

joaofnfernandes:
As the original author points out, with `--publish` you always need to specify the port that you want to open in the container. You use either `--publish <port>` or `--publish target=<port>` for this.

The terminology is a bit confusing but
* `published=<port>` refers to the port on the host
* `target=<port>` refers to the port on the container

So I can confirm that this change is correct. Let's ship it.

PR #6707, opened 2018-05-16
Add node.hostname to Go template placeholder options

omegamormegil:
The following command works on Docker CE 18.03.1-ce so .node.hostname should be included in the doc:

`docker service create --name myservice --hostname="{{.Node.Hostname}}-{{.Service.Name}}" busybox top`

This does not yet work however on the latest EE release, so I'd like some help knowing how best to add a note indicating this on the page.


omegamormegil:
Looks like this was already supposed to be added per https://github.com/docker/cli/pull/534

joaofnfernandes:
docker/cli#534 made its way to the docs with #6288, but it seems no one noticed that `engine/swarm/services.md` also needed updating. Thanks for fixing this!

PR #6708, opened 2018-05-16
Error in the Bypass-Mesh section of the Docs

auxiliary:
One needs to keep the `target` and remove the `published` port in order to let Docker assign ports itself. The documentation says the opposite of this in the **Bypass the routing mesh** section although correctly in other places of this article. In fact, the `target` argument is required and cannot be removed.

### Proposed changes
Substituted the word `target` with `published` in the Bypass-Mesh section in accordance to the rest of the guide.


joaofnfernandes:
It seems that #6706 fixed the same thing, that's why there's a conflict. Since this is fixed I'll close this PR.

@auxiliary File a new PR if this can be improved, and I'm sure there's going to be other opportunities for you to fix the docs!

PR #6717, opened 2018-05-17
Add docs for `local` logging driver.

cpuguy83:
### Proposed changes

Add docs for https://github.com/moby/moby/pull/37092

Depends on the above mentioned PR being merged

### DIFF
--- a/config/containers/logging/configure.md
+++ b/config/containers/logging/configure.md
@@ -150,9 +150,12 @@ see more options.
 | [`etwlogs`](etwlogs.md)       | Writes log messages as Event Tracing for Windows (ETW) events. Only available on Windows platforms.           |
 | [`gcplogs`](gcplogs.md)       | Writes log messages to Google Cloud Platform (GCP) Logging.                                                   |
 | [`logentries`](logentries.md) | Writes log messages to Rapid7 Logentries.                                                                     |
+| [`local`](local.md)           | Logs are stored in a custom format designed for minimal overhead.                                             |
 
 ## Limitations of logging drivers
 
-The `docker logs` command is not available for drivers other than `json-file`
-and `journald`.
+The `docker logs` command is only available on the following drivers:
+- `json-file`
+- `journald`
+- `local`
 
--- a/config/containers/logging/local.md
+++ b/config/containers/logging/local.md
@@ -1,0 +1,71 @@
+---
+description: Describes how to use the local logging driver.
+keywords: local, docker, logging, driver
+redirect_from:
+- /engine/reference/logging/local/
+- /engine/admin/logging/local/
+title: Local File logging driver
+---
+
+The `local` logging driver captures output from container's stdout/stderr and
+writes them to an internal storage that is optimized for performance and disk
+use.
+
+By default the `local` driver preserves 100MB of log messages per container and
+uses automatic compression to reduce the size on disk.
+
+> *Note*: the `local` logging driver currently uses file-based storage. The
+> file-format and storage mechanism are designed to be exclusively accessed by
+> the Docker daemon, and should not be used by external tools as the
+> implementation may change in future releases.
+
+## Usage
+
+To use the `local` driver as the default logging driver, set the `log-driver`
+and `log-opt` keys to appropriate values in the `daemon.json` file, which is
+located in `/etc/docker/` on Linux hosts or
+`C:\ProgramData\docker\config\daemon.json` on Windows Server. For more about
+configuring Docker using `daemon.json`, see
+[daemon.json](/engine/reference/commandline/dockerd.md#daemon-configuration-file).
+
+The following example sets the log driver to `local` and sets the `max-size`
+option.
+
+```json
+{
+  "log-driver": "local",
+  "log-opts": {
+    "max-size": "10m"
+  }
+}
+```
+
+Restart Docker for the changes to take effect for newly created containers. Existing containers do not use the new logging configuration.
+
+You can set the logging driver for a specific container by using the
+`--log-driver` flag to `docker container create` or `docker run`:
+
+```bash
+$ docker run \
+      --log-driver local --log-opt max-size=10m \
+      alpine echo hello world
+```
+
+### Options
+
+The `local` logging driver supports the following logging options:
+
+| Option      | Description                                                                                                                                                                                                   | Example  value                           |
+|:------------|:--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|:-----------------------------------------|
+| `max-size`  | The maximum size of the log before it is rolled. A positive integer plus a modifier representing the unit of measure (`k`, `m`, or `g`). Defaults to 20MB.                                          | `--log-opt max-size=10m`                 |
+| `max-file`  | The maximum number of log files that can be present. If rolling the logs creates excess files, the oldest file is removed. **Only effective when `max-size` is also set.** A positive integer. Defaults to 5. | `--log-opt max-file=3`                   |
+| `compress`  | Toggle compression of rotated log files. Enabled by default. | `--log-opt compress=false` |
+
+### Examples
+
+This example starts an `alpine` container which can have a maximum of 3 log
+files no larger than 10 megabytes each.
+
+```bash
+$ docker run -it --log-opt max-size=10m --log-opt max-file=3 alpine ash
+```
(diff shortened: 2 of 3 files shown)


cpuguy83:
ping @anusha-ragunathan

config/containers/logging/configure.md
```diff
@@ -150,9 +150,12 @@ see more options.
 | [`etwlogs`](etwlogs.md)       | Writes log messages as Event Tracing for Windows (ETW) events. Only available on Windows platforms.           |
 | [`gcplogs`](gcplogs.md)       | Writes log messages to Google Cloud Platform (GCP) Logging.                                                   |
 | [`logentries`](logentries.md) | Writes log messages to Rapid7 Logentries.                                                                     |
+| [`local`](local.md)           | Logs are stored in an custom, efficient logging format designed to cause as little overhead as possible.      |
```
anusha-ragunathan: "an custom" -> "a custom"

Also reword to "designed for minimal overhead"

config/containers/logging/local.md
```diff
@@ -0,0 +1,70 @@
...
+keywords: local, docker, logging, driver
+redirect_from:
+- /engine/reference/logging/local/
+- /engine/admin/logging/local/
+title: Local File logging driver
+---
+
+
+The `local` logging driver captures output from container's stdout/stderr and
+writes them to a local file on disk. The format of these files is unspecified.
```
anusha-ragunathan: WDYM by "format is unspecified" ?

config/containers/logging/local.md
```diff
@@ -0,0 +1,70 @@
...
+By default the `local` driver rotates files once they reach 20MB in size. Each
+rotated file is automatically compressed. The default maximum number of files is
+5, for a total of 100MB (when decompressed) of log data.
+
+The goal of this driver is to be as efficient as possible while still
+remaining consistent. As such log files may not be easily parseable with
```
anusha-ragunathan: "not easily parseable with standard CLI tools". This needs more explanation as to why.

config/containers/logging/local.md
```diff
@@ -0,0 +1,70 @@
...
+standard CLI tools.
+
+## Usage
+
+To use the `local` driver as the default logging driver, set the `log-driver`
+and `log-opt` keys to appropriate values in the `daemon.json` file, which is
+located in `/etc/docker/` on Linux hosts or
+`C:\ProgramData\docker\config\daemon.json` on Windows Server. For more about
++configuring Docker using `daemon.json`, see
```
anusha-ragunathan: remove dubious +

config/containers/logging/local.md
cpuguy83: Trying to convey that we are not going to document the format for public consumption because we don't want people messing with these files.
Also that the format itself may change over time.

config/containers/logging/local.md
```diff
@@ -0,0 +1,70 @@
...
++[daemon.json](/engine/reference/commandline/dockerd.md#daemon-configuration-file).
```
anusha-ragunathan: remove dubious + here as well

config/containers/logging/local.md
anusha-ragunathan: Not sure if that detail is necessary here. If you feel strongly about it, I'm fine.

cpuguy83: I think so because currently people parse the json logs and it makes it impossible to change the internal structure without breaking people.

config/containers/logging/local.md
anusha-ragunathan: Since this is purely to serve as a warning to users, in which case we should add it as a `Note/Warning`. I prefer `Note`, since it is less alarming.

config/containers/logging/local.md
anusha-ragunathan: Thinking a bit more. If we are expecting to change the format and dont provide any tools for parsing, do you think we should make it `experimental`?

cpuguy83: I don't. It's the users choice to use it or not, gating it behind experimental just means no one will use it.

anusha-ragunathan: Let's get @thaJeztah thought's on this

thaJeztah:
> Thinking a bit more. If we are expecting to change the format and dont provide any tools for parsing, do you think we should make it experimental?

@anusha-ragunathan I think I get your confusion;

#### The short answer

We must be explicit that this logging-driver uses an internal storage format that is not for external consumption, and that the format can change at any time. Perhaps something like:

```Markdown
The `local` logging driver captures output from container's stdout/stderr and
writes them to an internal storage that is optimized for performance and disk
use.

By default the `local` driver preserves 100MB of log messages per container and
uses automatic compression to reduce the size on disk.

> *Note*: the `local` logging driver currently uses file-based storage. The
> file-format and storage mechanism are designed to be exclusively accessed by
> the Docker daemon, and should not be used by external tools as the
> implementation may change in future releases.
```

#### The long answer:


What it comes down to, is that everything inside `/var/lib/docker` should be considered an "implementation detail"; this directory is designed to be exclusively accessed by a single daemon, and no external tools should be using those files.

Now comes the tricky bit;

- The API endpoint used for `docker inspect` return the "low level" information of containers. Basically, they directly expose our internal types through the API, which means that _any_ information we keep on a container will be exposed there, and any field that we add becomes visible. This includes information, such as the full path of container's root-mounts, `resolv.conf`, `hosts`, but also the `LogPath` (the file used by the json-file logging driver).
- The JSON-file logging-driver uses, well, JSON. So, even though the format/spec is not documented, it's easy to figure out the schema.
- The file location was exposed through the API (so, intended for public consumption?)

So, as a result;

- Docker didn't support log rotation, so people started writing [log-rotate](https://lime-technology.com/forums/topic/43549-log-rotation-for-docker-or-limit-log-size/). [tools](https://github.com/hortonworks/docker-logrotate). [for docker](https://github.com/blacklabelops/logrotate) (sometimes causing issues where the file was kicked from underneath dockerd).
- There were no logging-drivers (or _plugins_), and the logs API has a weird format that's difficult to use, so to aggregate logs, users started to read the log-files directly (this is what Kubernetes does; it creates symlinks to each-and-every logfile). 


Because of the above: even though the json-file log-format was intended to be an internal format ("implementation detail"), it has now become an unofficial public API. We can't change it without breaking many people.


So, what we need is;

- Be explicit; a published / documented API is a contract, and even though we mention we use an "open schema" for the API (more files may be returned), we haven't been explicit 
- Don't re-use / expose internal types as API types; changing an internal type should never automatically change the API
- If API fields come with descriptions, these should be documented. For example, some information is not part of the API, and their values don't have a defined format (tried doing that for the [`DriverStatus`](https://docs.docker.com/engine/api/v1.37/#operation/SystemInfo), and [`DriverOpts`](https://docs.docker.com/engine/api/v1.37/#operation/VolumeCreate) fields).

config/containers/logging/local.md
```diff
@@ -0,0 +1,70 @@
...
+```
+
+### Options
+
+The `local` logging driver supports the following logging options:
+
+| Option      | Description                                                                                                                                                                                                   | Example  value                           |
+|:------------|:--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|:-----------------------------------------|
+| `max-size`  | The maximum size of the log before it is rolled. A positive integer plus a modifier representing the unit of measure (`k`, `m`, or `g`). Defaults to 20MB.                                          | `--log-opt max-size=10m`                 |
+| `max-file`  | The maximum number of log files that can be present. If rolling the logs creates excess files, the oldest file is removed. **Only effective when `max-size` is also set.** A positive integer. Defaults to 5. | `--log-opt max-file=3`                   |
```
thaJeztah: I should probably comment this on the implementation PR, but isn't "max file" an implementation detail? Can we abstract storage even more, and just allow people to set _"max amount of logs to keep"_ ?

config/containers/logging/local.md
```diff
@@ -0,0 +1,70 @@
...
+| `compress`  | Toggle compression of rotated log files. Enabled by default. | `--log-opt compress=false` |
```
thaJeztah: Same here; "rotated files" is likely implementation detail; we can describe here why someone would want to disable  compression (does it cause a significant overhead? discouraged for high-volume logging? discouraged for "many running containers"?).

Ideally, we'd change this to (e.g.) `disable-compression`. Booleans with `--enablesomething=false` are a bit awkward  😅

config/containers/logging/local.md
```diff
@@ -0,0 +1,70 @@
...
+      alpine echo hello world
+```
+
+### Options
+
+The `local` logging driver supports the following logging options:
+
+| Option      | Description                                                                                                                                                                                                   | Example  value                           |
+|:------------|:--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|:-----------------------------------------|
+| `max-size`  | The maximum size of the log before it is rolled. A positive integer plus a modifier representing the unit of measure (`k`, `m`, or `g`). Defaults to 20MB.                                          | `--log-opt max-size=10m`                 |
```
thaJeztah: (Probably better discussed on the implementation PR, I'll comment there as well); is the storage-format suitable for doing time-based rotation as well?  (see https://github.com/moby/moby/issues/36853)?

config/containers/logging/local.md
thaJeztah: Should we add `labels` here as well? (I think it's a common option, so it may be documented as such, but could be useful to include it in every logging driver)

anusha-ragunathan:
Now that the implementation PR is merged, lets get the doc changes done.

@cpuguy83: Can you address @thaJeztah 's concerns?

cpuguy83:
Updated with explicit wording about the the private nature of the implementation.

tfoxnc:
@anusha-ragunathan Do you approve of the last change? Are you ready for me to merge now?

anusha-ragunathan:
LGTM

PR #6719, opened 2018-05-18
Fix some typos in the compose file docs

xander-marjoram:
### Proposed changes

Fixed some typos in the compose file reference docs 👍 

If I run the command-line snippet as-is in the docs, I get the following error:
> `docker: 'compose' is not a docker command.`


joaofnfernandes:
Thanks for fixing this, let's ship it!

PR #6720, opened 2018-05-18
Update upgrade.md

n00tz:
### Proposed changes

added more specific version numbers to the upgrade engine instruction section for UCP 3.0.1


joaofnfernandes:
Makes sense, let's ship it.

PR #6725, opened 2018-05-19
update compose django example

2tunnels:
### Proposed changes

- Changed `ADD` command to `COPY`, because of [best practices](https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#add-or-copy).
- Updated `django` and `psycopg2` versions.
- Updated django's homepage screenshot.
- Made few tweaks in command invocations (for example you don't need to use `python3` binary, because it's already based on Python 3 image).
- Changed `docker ps` to `docker container ls`. As I understand it's better to move to new commands.


cyroxx:
This is the most comprehensive PR among those trying to fix the broken compose/Django docs due to version incompatibility.

Django 1.x has no support for the current Python 3.7 (see https://code.djangoproject.com/ticket/28814, kudos to @thedarklord1 for the reference).

Related PRs/issues (plus a short comment):
- #7253 bug report that shows the console output 
- #7223 completely removes the version requirement for Django, which is not desirable
- #7191 uses Django>=2.0, but does not specify an upper boundary
- #7206 reports python3.6+django2.1 as an alternative working solution (I got python3.7+django2.1 working, for reference).
- #6990 solves the problem by pinning python to 3.5. _Maybe we should use this idea here and explicitly use python3.7 in order to prevent breakage (cc @2tunnels)?_

IMHO, all these could be closed after accepting this PR.

paigehargrave:
@shin- PTAL - thank you!

PR #6730, opened 2018-05-20
Updated version specification

idontusenumbers:
### Proposed changes

The version specification used in the examples was before the feature was introduced and would thus never parse.

### DIFF
--- a/_includes/content/compose-extfields-sub.md
+++ b/_includes/content/compose-extfields-sub.md
@@ -29,7 +29,7 @@ logging:
 You may write your Compose file as follows:
 
 ```none
-version: '2.1'
+version: '3.4'
 x-logging:
   &default-logging
   options:
@@ -50,7 +50,7 @@ It is also possible to partially override values in extension fields using
 the [YAML merge type](http://yaml.org/type/merge.html). For example:
 
 ```none
-version: '2.1'
+version: '3.4'
 x-volumes:
   &default-volume
   driver: foobar-storage


thaJeztah:
ping @shin- PTAL

idontusenumbers:
Oh! Maybe so.

Maybe I'm misreading the docs.

https://docs.docker.com/compose/compose-file/#extension-fields

image

Should that block be updated instead?

idontusenumbers:
I have a file using extensions. It has 3.4 specified on the top. When I change it to 3.3 I get:
```
ERROR: Version mismatch: file <path> specifies version 3.3 but extension file <path> uses version 3.4
```

thaJeztah:
The confusing bit likely is that the `2.x` and `3.x` schema's are two different "branches" of the compose-file format. `3.x` started as a subset of the `2.x` format, and has options/features designed for the Swarm-mode services. Both the3.x and 2.x format are still actively evolving, which in this case means that the feature was added for the 3.x spec in 3.4, and for the 2.x spec in 2.1

shin-:
[This](https://github.com/docker/docker.github.io/pull/6730#issuecomment-391119868) is correct.

Since this include is used in both the v2 and v3 specs, it's going to be slightly confusing either way I suppose ; that said, if v3 is our recommended version, then it makes sense to optimize the examples for v3.

joaofnfernandes:
Given the feedback, I think it makes sense for us to merge this contribution :D

PR #6733, opened 2018-05-20
Update live-restore.md

mark-church:
### Proposed changes

Changed the wording for live-restore upgrades from "minor" to "patch" as the documentation was incorrect. Only patch daemon upgrades are supported for live-restore.

cc @dmaring @thaJeztah @crosbymichael


joaofnfernandes:
Let's ship it!

PR #6734, opened 2018-05-21
Update upgrade.md

n00tz:
Add language to specifically call out network firewall changes necessary before upgrade.

### DIFF
--- a/ee/ucp/admin/install/upgrade.md
+++ b/ee/ucp/admin/install/upgrade.md
@@ -26,12 +26,9 @@ Ensure that your cluster nodes meet the minimum requirements for memory and
 disk space. In particular, manager nodes must have at least 4GB of memory.
 Learn about [UCP system requirements](system-requirements.md).
 
-> New firewall ports 
->
-> With the introduction of Kubernetes components, additional ports may need
-> to be opened in the firewall. [Ports used](system-requirements.md/#ports-used) are documented
-> in the UCP system requirements. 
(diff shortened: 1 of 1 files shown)


joaofnfernandes:
Thanks for trying to make our users' lives easier!
Even though I understand the intention of this change, I don't agree with it.

We use semantic versioning in UCP and DTR to signal to users that:
* The version is not compatible with older releases
* There might be breaking changes
* There might be different requirements

Given that the system requirements doc already lists all the minimum requirements to a version, we should not duplicate that information anywhere else.
This means the first step of the upgrade docs should be "make sure your cluster complies with the new system requirements (here's the link to check that out)".

There are two reasons for that:
* Keeping information in two places makes things harder to maintain, and information becomes inconsistent over time
* RAM and ports are only a subset of system requirements. There are a ton of other requirements users need to comply with like OS versions, clock synchronization, ...

I know that the current version of the docs already talks about checking RAM and disk space, so it might seem that firewall ports are only an extension to this information.
The information about RAM should not be there to begin with, so let's not continue expanding that section.

@n00tz are you willing to update this PR to clean up the whole section and make clear to users that they should check the system requirements?

n00tz:
@joaofnfernandes I've updated the section to more closely match the existing documentation with specific call out for port openings.

joaofnfernandes:
Thanks for addressing the feedback! Let's ship it.

PR #6754, opened 2018-05-24
Update services.md

lvthillo:
Why should it be difficult? It's indeed assigned to a random node, but it isn't hard to know which one.


joaofnfernandes:
The phrase could be improved but I think you misunderstood it's intention. Of course you can figure out where the service task is running after it is deployed, but there are no assurances that the task will continue running on that node indefinitely.

If for some reason the task is rescheduled and starts running on another node, users accessing the service using the node's IP won't be able to continue using the service anymore.

I'm going to close this PR, but if you're willing, file a new one to clarify that sentence.

PR #6757, opened 2018-05-24
Cannot find reference to --net

jcrben:
### Proposed changes

Is network intended? It looks like this is talking about 
https://docs.docker.com/engine/reference/run/#network-settings
N/A

### Related issues (optional)
N/A


joaofnfernandes:
This seems correct. From what I can recall `--net` was deprecated in favor of `--network`.

PR #6759, opened 2018-05-24
Update instructions to install specific version

### LINKED ISSUE #6758: Version for 'docker-ce' was not found
### Problem description

The specification for installing a specific version of docker-ce at the page "Get Docker CE for Ubuntu" is incorrect. The problem's solution is described here: https://stackoverflow.com/a/50403108/1791115

### Problem location

The problem is at this location: https://docs.docker.com/install/linux/docker-ce/ubuntu/#install-docker-ce-1

Under `2. Install the latest version of Docker CE, or go to the next step to install a specific version:`

The problem occurred only on this page

### Suggestions for a fix

Change the instructions to match the string from the 2nd column of apt-cache

shakegioh:
Issue #6758

### Proposed changes

The instructions for install an older version of docker-ce are inaccurate

Fixes #6758

### DIFF
--- a/install/linux/docker-ce/ubuntu.md
+++ b/install/linux/docker-ce/ubuntu.md
@@ -260,9 +260,9 @@ the repository.
     ```
 
     b. Install a specific version by its fully qualified package name, which is
-       the package name (`docker-ce`) plus the version string (2nd column) up to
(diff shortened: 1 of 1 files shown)


joaofnfernandes:
This has a thumbs up, so let's ship it.

Issue #6749, opened 2018-05-23
Host Networking is Linux Only


nvick:
[network/host.md](https://docs.docker.com/network/host/)

Please update the docs to note that host networking is a Linux only option.

```The host networking driver only works on Linux hosts, and is not supported on Docker for Mac, Docker for Windows, or Docker EE for Windows Server.```

PR #6763, opened 2018-05-25
Add note about host networking in linux

srinivasreddy:
Fixes: #6749


joaofnfernandes:
I've tested this on Docker for Mac and Linux, and this looks correct. Thanks for the fix!

PR #6765, opened 2018-05-25
Fix some broken links

sascha-egerer:
### Proposed changes

Fixed some links to non existing targets

### DIFF
--- a/compose/networking.md
+++ b/compose/networking.md
@@ -4,7 +4,7 @@ keywords: documentation, docs, docker, compose, orchestration, containers, netwo
 title: Networking in Compose
 ---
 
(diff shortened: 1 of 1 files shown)


sascha-egerer:
@gbarr01 Ah yeah they are not broken in the documentation. But broken if you use the markdown rendered version of github (https://github.com/docker/docker.github.io/blob/master/compose/networking.md).
Other links do also have that `.md` suffix and they work both on Github and on the docs page. So looks like the parser on docs.docker.com transforms those links automatically by replacing `.md` by `/`

joaofnfernandes:
Let's ship it!

PR #6766, opened 2018-05-25
Mac: how to run docker-diagnose

akimd:
### Proposed changes

- update instructions about a moved HOME
- show how to run docker-diagnose by hand

Cc @gtardif @jasonbivins

### DIFF
--- a/docker-for-mac/faqs.md
+++ b/docker-for-mac/faqs.md
@@ -75,7 +75,7 @@ Do the following each time:
 
 `Docker.app` is Docker for Mac, a bundle of Docker client, and Docker Engine.
 `Docker.app` uses the macOS Hypervisor.framework to run containers, meaning that
-_**no separate VirtualBox is required**_.
+no separate VirtualBox is required.
 
 ### What are system requirements for Docker for Mac?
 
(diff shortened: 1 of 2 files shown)


docker-for-mac/faqs.md
```diff
@@ -86,10 +86,19 @@ the install guide.
 
 ### Do I need to reinstall Docker for Mac if I change the name of my macOS account?
 
-Yes, if you [change the name of your macOS user account and home
-folder](https://support.apple.com/en-us/HT201548){: target="_blank" class="_"} after Docker for Mac is
-already installed, you need to uninstall and reinstall Docker for Mac. Otherwise, Docker for Mac fails to start. See also,
-[Docker for Mac does not start if Mac user account and home folder are renamed after installing Docker for Mac](troubleshoot.md#docker-for-mac-does-not-start-if-mac-user-account-and-home-folder-are-renamed-after-installing-the-app) in Troubleshooting.
+Starting with Docker for Mac Edge 18.06, this path is relative to the user's
```
gtardif: We can mention just release 18.06 without specifying Edge, since it is going to be also in Stable 18.06.
I would mention this change in a separate PR to merge it when 18.06 is out, separately from the rest

joaofnfernandes:
I think this looks good even if the feedback hasn't been addressed. We can follow-up with a subsequent PR if needed.

PR #6769, opened 2018-05-25
ucp: node-local support dump works on all nodes

trapier:


trapier:
please do not merge yet.  subsequent paragraph also needs a boost.

trapier:
not gonna have time at the moment to finish that second lift.  will do it in a follow-up.

PR #6790, opened 2018-05-31
Changed host port from 80 to 4000 in get-started/part3.md

alexjneves:
### Proposed changes

The host port is set to 80, which is inconsistent with the previous tutorial and doesn't work on Windows. Changing the port to 4000 keeps consistency and works on Windows.

For reference, here is the error when attempting to run the container on port 80:

_C:\Program Files\Docker\Docker\Resources\bin\docker.exe: Error response from daemon: driver failed programming external connectivity on endpoint focused_euler (079893ea98df78bf7043f86072263317be106cbb7f4a3fcde23c2ef19ad39a01): Error starting userland proxy: Bind for 0.0.0.0:80: unexpected error Permission denied.
ERRO[0000] error waiting for container: context canceled_

### DIFF
--- a/get-started/part3.md
+++ b/get-started/part3.md
@@ -26,8 +26,8 @@ Toolbox](/toolbox/overview.md).
 shared image here.
 
 - Be sure your image works as a deployed container. Run this command,
-slotting in your info for `username`, `repo`, and `tag`: `docker run -p 80:80
-username/repo:tag`, then visit `http://localhost/`.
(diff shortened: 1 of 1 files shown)


alexjneves:
Further investigation suggests it's a port collision issue on my machine.

There is still the question of consistency between the tutorials; I've noticed that later tutorials also use port 80

gbarr01:
Thanks, @alexjneves .

PR #6792, opened 2018-05-31
document how to backup volumes, not containers

graingert:
### Proposed changes

document how to backup volumes, not containers

because the heading is under 'volumes'


L-Hudson:
Corrections as stated already in place. Closing the ticket; thanks for your contribution!

PR #6794, opened 2018-05-31
Update network-tutorial-host.md to remove redundant flags/commands

adshmh:
### Proposed changes

- Modify the `docker run` command to remove the unnecessary flags "-i" and "-t". 
- Add a short description for the "--rm" option.
- Remove the "docker container rm" command which seems redundant due to "--rm" option.

### DIFF
--- a/network/network-tutorial-host.md
+++ b/network/network-tutorial-host.md
@@ -28,10 +28,10 @@ host.
 
 ## Procedure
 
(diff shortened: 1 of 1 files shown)


joaofnfernandes:
Let's ship it! Next time consider using the verbose version of the flags like `--detached` instead of `-d`. That way things are a big more obvious to new users.

PR #6798, opened 2018-06-01
Update get-support.md

lenestopage:
### Proposed changes

Created and added a process to capture a support dump, from the CLI, for multiple nodes rather than just a single node.


ee/get-support.md
```diff
@@ -45,6 +45,26 @@ This support dump only contains logs for the node where you're running the
 command. If your UCP is highly available, you should collect support dumps
 from all of the manager nodes.
 
+## Use the CLI to get a support dump (multiple nodes)
+
+To get the support dump from the CLI for multiple nodes, use SSH to log into a UCP manager node
+and run:
```
joaofnfernandes: Can you rewrite this, or move the command below this sentence? After reading this users will expect to see the snippet below.

ee/get-support.md
```diff
@@ -45,6 +45,26 @@ This support dump only contains logs for the node where you're running the
...
+
+To generate support dump details for all hosts in a cluster, from one remote host, please see below.
```
joaofnfernandes: This is a short doc, so you can remove this sentence.

ee/get-support.md
```diff
@@ -45,6 +45,26 @@ This support dump only contains logs for the node where you're running the
...
+
+The below variables need to be set initially and folders created prior to running the command. Please note the variables can be changed to fit your environment:
+
+```$ SUPPORT_DUMP_DIR=<^>/opt<^^>
```
joaofnfernandes: 1. Remove `$` from the beginning of the snippets. It makes it easier for users to copy-paste the snippet.
2. Move the first variable one line below
3. Make the first variable be `SUPPORT_DUMP_DIR=/tmp`

Users will understand that they can customize it, so we don't need to use that weird syntax.

ee/get-support.md
```diff
@@ -45,6 +45,26 @@ This support dump only contains logs for the node where you're running the
...
+$ SUPPORT_DUMP_TEMPDIR=${SUPPORT_DUMP_DIR}/support_dump_temp
+$ DUMP_DATE=$(date +%Y%m%d-%H_%M_%S)
+$ mkdir ${SUPPORT_DUMP_TEMPDIR}
```
joaofnfernandes: Make it `mkdir -p ...` just in case users specify nested directories.

ee/get-support.md
```diff
@@ -45,6 +45,26 @@ This support dump only contains logs for the node where you're running the
...
+$ UCP_VERSION=$(docker service inspect ucp-agent --format '{{range .Spec.TaskTemplate.ContainerSpec.Env}}{{$varval:= split . "="}}{{if eq "IMAGE_VERSION" (index $varval 0)}}{{index $varval 1}}{{end}}{{end}}')
```
joaofnfernandes: This is cool, but can you updated to use the version label instead? Use this:
```
UCP_VERSION=$(docker service inspect ucp-agent --format '{{index .Spec.TaskTemplate.ContainerSpec.Labels "com.docker.ucp.version"}}')
```

ee/get-support.md
```diff
@@ -45,6 +45,26 @@ This support dump only contains logs for the node where you're running the
...
+```
+The below command creates a zip file that contains support details (dsinfo) for each host in your cluster:
+
+```$ for node in $(docker node ls --format '{{if eq .Status "Ready"}}{{.Hostname}}{{end}}'); do echo timing support dump on $node; time docker container run --rm -e constraint:node==$node -v /boot:/boot -v /proc/:/host/proc:ro -v /var/run/docker.sock:/var/run/docker.sock -v /var/run/docker.pid:/var/run/docker.pid:ro -v /var/run/docker:/var/run/docker -v /var/lib/docker:/var/lib/docker -v /var/log:/var/log -v /etc:/etc:ro --privileged --pid=host --network=host --log-driver=json-file docker/ucp-dsinfo:${UCP_VERSION} > ${SUPPORT_DUMP_TEMPDIR}/support_dump_${node}.tgz; mkdir ${SUPPORT_DUMP_TEMPDIR}/${node}; tar -xf ${SUPPORT_DUMP_TEMPDIR}/support_dump_${node}.tgz -C ${SUPPORT_DUMP_TEMPDIR}/${node}; rm -f ${SUPPORT_DUMP_TEMPDIR}/support_dump_${node}.tgz; cd ${SUPPORT_DUMP_TEMPDIR} ; zip -q -r ${SUPPORT_DUMP_DIR}/docker-support-${DUMP_DATE}.zip . *; cd ..; done
```
joaofnfernandes: @wsong I'd like your thumbs up on this. As a user I would be concerned about running this command, give that we're asking users to run a container in privileged mode and attaching to a bunch up sensitive directories.

Is there a better way to do this?

joaofnfernandes:
@lenestopage are you willing to implement the feedback?

lenestopage:
@joaofnfernandes sorry for the late response, definitely will make these changes.

ee/get-support.md
lenestopage: completed.

ee/get-support.md
lenestopage: completed.

ee/get-support.md
lenestopage: completed.

ee/get-support.md
lenestopage: completed.

ee/get-support.md
lenestopage: completed.

ee/get-support.md
trapier: > Is there a better way to do this?

seconded.

trapier: also, for posterity, this approach would only work if the user had an admin UCP client cert bundle loaded.

lenestopage: Thanks @joaofnfernandes going to close this out. I don't think we have a lot of customers beating the doors down for this.

Issue #6793, opened 2018-05-31
I found a broken link


tfoxnc:
1. Go to https://docs.docker.com/ee/ucp/#centralized-cluster-management
2. Click on "Learn about role-based access control."

joaofnfernandes:
Fixes via #6801

Issue #6786, opened 2018-05-30
Change references of "Power PC" to "Power"


yongshin:
There are multiple instances where Power PC is mentioned, we should change it to Power as that is the correct terminology for the architecture we support. 

"IBM Power PC, little endian format" 

[install/linux/docker-ee/rhel.md](https://docs.docker.com/install/linux/docker-ee/rhel/)

joaofnfernandes:
Fixed via #6801

PR #6801, opened 2018-06-01
Fixed broken Link and Power PC to Power

Vimal-Raghubir:
### Proposed changes
1. Fixed broken link in ee/ucp/index.md that wasn't redirecting to the correct site. Changed it to authorization/index.md from access-control/index.md (this file doesn't exist) to fix this issue.

2. Changed references of Power PC to Power in 2 files to reflect the actual architecture used.

### Related issues

Refers to #6793 and #6786


joaofnfernandes:
Let's merge this since it fixes the broken links. From a branding perspective I'm not sure if we should be using "IBM Power" or "IBM Power Systems".

Issue #1013, opened 2017-01-03
Feedback for: docker-cloud/installing-cli.md


Downes:
[docker-cloud/installing-cli.md](https://docs.docker.com/docker-cloud/installing-cli/)

On the previous page in the docs, it says having a running instance of docker on the local machine is optional.

However, on this page it says "First, you should log in using the docker CLI and the docker login command." which makes it look like it is *not* optional.

Maybe the export DOCKERCLOUD_USER command would help here? I don't know.

sanscontext:
Hey @Downes thanks for your feedback. The page you've linked to is a standalone page and isn't part of a tutorial. Originally the tutorials used the `docker-cloud login` command, but since we've centralized our accounts this is now part of the `docker login` command. I'm checking with the team to see what updates I should make to clarify here.

PR #6803, opened 2018-06-02
Update docker-compose run --rm doc

orenyomtov:
### Proposed changes

Add `Overrides the container's restart policy` to `docker-compose run --rm`'s documentation.

### Related issues

It makes the following change explicit in the documentation
https://github.com/funkyfuture/docker-compose/commit/24ab933ebfd3169491beeb18f0041cabebe4d2d2
https://github.com/docker/compose/issues/1013

I had to use Google to and follow the discussion in #1013 in order to know how to properly achieve what I wanted, after not finding the answer in `docker-compose run --help`.

Related to https://github.com/docker/compose/pull/5997

### DIFF
--- a/compose/reference/run.md
+++ b/compose/reference/run.md
@@ -20,6 +20,7 @@ Options:
     -u, --user=""         Run as specified username or uid
     --no-deps             Don't start linked services.
     --rm                  Remove container after run. Ignored in detached mode.
+                          Overrides the container's restart policy.
     -p, --publish=[]      Publish a container's port(s) to the host
     --service-ports       Run command with the service's ports enabled and mapped
                           to the host.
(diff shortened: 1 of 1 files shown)


joaofnfernandes:
@shin- can you confirm?

joaofnfernandes:
@shin- can you give feedback on where you think is the best place to document this? Maybe @orenyomtov is willing to update this PR to make that happen.

orenyomtov:
True, I am willing.

shin-:
Sorry, I wrote the earlier message from a phone in between other tasks. Let me expand on what I meant:

It's a detail that is irrelevant in 99% of use cases. That doesn't mean it shouldn't be documented, but doing so in the CLI output creates confusion by making the information seem more important than it actually is to most users.

Instead, let's add it at the end of the page as an addendum to the description of the inner workings of the `run` command.

Thank you for taking the time to help out!

orenyomtov:
@shin- thanks for the feedback.

Let me know what you think of this update.

joaofnfernandes:
Thanks for following through on this! Let's ship it.

PR #6813, opened 2018-06-04
Cgroup is a namespace.

spinside:
Cgroup is a namespace, it would be confusing to name namespaces and cgroups seperately.

### Proposed changes

Removed control group form the following sentence:
When you start a container with docker run, behind the scenes Docker creates a set of namespaces and control groups for the container.

This is done because Cgroup is according to the Linux Man pages a namespace. It would be confusing to list namespaces and control groups separately. It would imply that cgroup/control group is not a namespace.

Linux man page: 
http://man7.org/linux/man-pages/man7/namespaces.7.html

### DIFF
--- a/engine/security/security.md
+++ b/engine/security/security.md
@@ -21,8 +21,7 @@ There are four major areas to consider when reviewing Docker security:
 
 Docker containers are very similar to LXC containers, and they have
 similar security features. When you start a container with
-`docker run`, behind the scenes Docker creates a set of namespaces and control
-groups for the container.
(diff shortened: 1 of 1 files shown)


gbarr01:
Hi, @spinside , thank you for the submission. I am not going to merge this PR because cgroups and namespaces are not the same thing -- they function differently and need to be distinguished.  

For some guidance please see: 

- https://stackoverflow.com/questions/34820558/difference-between-cgroups-and-namespaces
- https://fr.slideshare.net/jpetazzo/anatomy-of-a-container-namespaces-cgroups-some-filesystem-magic-linuxcon

PR #6821, opened 2018-06-05
typo in ucp ldap config

trapier:


trapier:
since it's not easy to see:

> [...] making a separate LDAP query for each member. as opposed to first [...]

changed to:

> > [...] making a separate LDAP query for each member, as opposed to first [...]

PR #6823, opened 2018-06-05
updated instructions for manual install

TutieFrutie:
Before extracting the archive you must stop the service or the file docker.exe can't be overwritten.
Additionally need to use -Force


joaofnfernandes:
Thanks for the contribution!

PR #6834, opened 2018-06-06
Add note about not using free to determine swap availability inside container.

### LINKED ISSUE #6833: Documentation should mention how swap is reported inside the container
### Problem description

I was tripped up on a swap-related issue with one of my applications, where it was reporting an amount of available swap space different than my configured `--memory-swap` limit set on a `docker run` command.

It would be helpful to mention in the [--memory-swap section of the 'Limit a container's resources' docs](https://docs.docker.com/config/containers/resource_constraints/#--memory-swap-details) how the available swap space can be viewed inside the container (e.g. `cat /sys/fs/cgroup/memory/memory.limit_in_bytes`), and how common tools like `free` won't display the correct value, but rather the host value.

### Problem location

- I couldn't find the information I wanted. I expected to find it near the following URL: https://docs.docker.com/config/containers/resource_constraints/

N/A

### Suggestions for a fix

...working on a PR now.

geerlingguy:
Fixes #6833.

### DIFF
--- a/config/containers/resource_constraints.md
+++ b/config/containers/resource_constraints.md
@@ -118,6 +118,8 @@ Its setting can have complicated effects:
 - If `--memory-swap` is explicitly set to `-1`, the container is allowed to use
   unlimited swap, up to the amount available on the host system.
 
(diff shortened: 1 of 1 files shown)


joaofnfernandes:
Thanks for the contribution!

PR #6842, opened 2018-06-07
Update deploying.md

embray:
### Proposed changes

Since these docs are demonstrating how to run a swarm service with TLS support I assume port 443 was intended here.


joaofnfernandes:
Thanks for the contribution!

PR #6846, opened 2018-06-08
Update vfs-driver.md

fireman777:
edited:
/var/lib/docker/dir/ => /var/lib/docker/vfs/dir/


fireman777:
edited:
/var/lib/docker/dir/ => /var/lib/docker/vfs/dir/

joaofnfernandes:
I've confirmed that this is correct. Let's ship it!

PR #6847, opened 2018-06-08
Update restore.md

gcontad01:


joaofnfernandes:
Thanks for your contribution. Unfortunatelly the page you updated is generated automatically from running `docker run docker/dtr restore --help`. Even if I merged your changes, they would be reverted once a new patch is released.

I'll close this PR, but consider reviewing https://docs.docker.com/ee/dtr/admin/disaster-recovery/restore-from-backup/. If you feel that the article would still benefit from your note, open a new PR to update it.

PR #6848, opened 2018-06-08
Update destroy.md

gcontad01:


joaofnfernandes:
Thanks for your contribution. Unfortunately this article is automatically generated from `docker/dtr destroy --help` so even if I merged your contribution it would be lost once we shipped a new version.

For that reason, I'll close this PR.

PR #6851, opened 2018-06-08
Update install.md

lenestopage:
### Proposed changes

After following the Linux instructions regarding installing compose, I still was unable to run docker-compose --version. Performing the following cleared my issue: ln -s /usr/local/bin/docker-compose /usr/bin/docker-compose

### DIFF
--- a/compose/install.md
+++ b/compose/install.md
@@ -128,6 +128,12 @@ by step instructions are also included below.
     ```bash
     sudo chmod +x /usr/local/bin/docker-compose
     ```
+    
+Note: If after installing compose you still can't use it due to the docker-compose command failing, please do the following:
+
+```bash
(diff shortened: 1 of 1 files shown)


compose/install.md
```diff
@@ -128,6 +128,12 @@ by step instructions are also included below.
     ```bash
     sudo chmod +x /usr/local/bin/docker-compose
     ```
+    
+Note: If after installing compose you still can't use it due to the docker-compose command failing, please do the following:
```
joaofnfernandes: Can you fix this so that it matches our styleguide for notes? It should be

```
> If after installing ...
```

ahh-docker:
@bermudezmt PTAL

PR #6869, opened 2018-06-12
Add warning about package building policy

remram44:
### Proposed changes

Explain to users why Docker packages might be missing. I was told on https://github.com/moby/moby/issues/37247#issuecomment-396416563 that this was an official policy, it should therefore be stated.

### DIFF
--- a/install/linux/docker-ce/centos.md
+++ b/install/linux/docker-ce/centos.md
@@ -149,6 +149,13 @@ from the repository.
     > `yum update` command always installs the highest possible version,
     > which may not be appropriate for your stability needs.
 
+    > **Warning**:
+    >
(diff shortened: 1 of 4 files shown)


joaofnfernandes:
Thanks for the contribution! Unfortunately I don't think we should integrate it 😭
I think the docs make it clear what distribution versions are supported. Using [Ubuntu as an example](https://docs.docker.com/install/linux/docker-ce/ubuntu/):

>To install Docker CE, you need the 64-bit version of one of these Ubuntu versions:
>Bionic 18.04 (Docker CE 18.05 Edge and higher only)
>Artful 17.10
>Xenial 16.04 (LTS)
>Trusty 14.04 (LTS)

I understand what this contribution is trying to achieve, but I don't agree in adding more information in the middle of the article about policies (versions supported and things like that), when that information is already clear at the beginning of the article.

For that reason, I'm closing this PR, but feel free to comment if you disagree.

PR #6870, opened 2018-06-12
fix nginx logs description

drAlberT:
the log file is linked to the device, not viceversa :)


joaofnfernandes:
Looks good, let's ship it!

PR #6871, opened 2018-06-12

fvictorio:
Add missing tag names.

### Proposed changes

Fix minor error in an example in the best practices document.


joaofnfernandes:
Makes sense, let's ship it!

PR #6872, opened 2018-06-12
interlock: add linux to documented constraints

trapier:
lifted this additional constraint from a default UCP 3.0.1 install.


joaofnfernandes:
Thanks for keeping the docs updated!

PR #6873, opened 2018-06-12
Switch places of subtitles DNS and Proxy settings

chibby0ne:
Proxy settings titles was placed in the DNS misconfiguration part, and the DNS settings title was placed in the proxy configuration part. This changes moves the places of the titles so that they both present the section they intend to.


joaofnfernandes:
Thanks for your contribution!

PR #6877, opened 2018-06-13
Update linux-postinstall.md

bitsmaestro:
In section "Manage Docker as a non-root user" added missing step to restart the docker service after making changes in user group membership.

### DIFF
--- a/install/linux/linux-postinstall.md
+++ b/install/linux/linux-postinstall.md
@@ -39,14 +39,20 @@ To create the `docker` group and add your user:
     ```bash
     $ sudo usermod -aG docker $USER
     ```
-
-3.  Log out and log back in so that your group membership is re-evaluated.
+3. Restart the docker service.
+   Since the changes in ownership of the Unix socket are only done when the `docker` daemon starts, we need to restart the service.
+   
+    ```
+    $ systemctl restart docker
+    ```
+    
+4.  Log out and log back in so that your group membership is re-evaluated.
 
     If testing on a virtual machine, it may be necessary to restart the virtual machine for changes to take effect.
 
     On a desktop Linux environment such as X Windows, log out of your session completely and then log back in.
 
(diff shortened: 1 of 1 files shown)


install/linux/linux-postinstall.md
```diff
@@ -39,14 +39,20 @@ To create the `docker` group and add your user:
     ```bash
     $ sudo usermod -aG docker $USER
     ```
-
-3.  Log out and log back in so that your group membership is re-evaluated.
+3. Restart the docker service.
+   Since the changes in ownership of the Unix socket are only done when the `docker` daemon starts, we need to restart the service.
```
joaofnfernandes: Can you confirm this @thaJeztah?

thaJeztah: Hm, don't think that should be needed; these instructions are for adding the current user to the docker group; no changes are made to the permissions on the socket itself?

thaJeztah:
Is that in a situation where the docker group didn't exist yet? 🤔

bitsmaestro:
@thaJeztah  sorry for delay in response 
we are installing docker on **rhel** through  **rhel-7-server-extras-rpms** repo and it does not have docker group. It has dockerroot group available. 

```
# yum info docker
.....
Installed Packages
Name        : docker
Arch        : x86_64
Epoch       : 2
Version     : 1.12.6
Release     : 48.git0fdc778.el7
Size        : 52 M
Repo        : installed
From repo   : rhel-7-server-extras-rpms
```
```
# cat /etc/redhat-release
Red Hat Enterprise Linux Server release 7.4 (Maipo)
```

joaofnfernandes:
I see. This means you're actually not installing using the instructions in Docker docs, so I don't think we should incorporate this change.
I think this is something that should be addressed in Red Hat's docs.

@thaJeztah what do you think?

PR #6885, opened 2018-06-14
Describe that Ubuntu Bionic is supported on edge

thaJeztah:
### Proposed changes

Ubuntu bionic is supported as of Docker 18.05 and up (edge)


thaJeztah:
ping @joaofnfernandes PTAL

joaofnfernandes:
I don't see anything here that would break the build, so let's ship it 😎

PR #6886, opened 2018-06-14
Replacing cloud.docker.com  for hub.docker.com

johandry:


joaofnfernandes:
Makes sense, let's ship it!

PR #6893, opened 2018-06-17
adds port 4000

### LINKED ISSUE #6904: Clarification: curl localhost on the port specified in the docker-compose.yml
[get-started/part3.md](https://docs.docker.com/get-started/part3/)

The instructions mention to `curl -4 http://localhost`, however this will curl to the default port 80. The services are however running on localhost port 4000 as specified in the docker-compose.yml file. I propose the instructions and sample screenshot are changed to specifically mention `curl -4 http://localhost:4000`.

eightlimbed:
In the docker-compose file, the ports sections maps host port 4000 to container port 80, so if we want to see our app in a browser we should check on port 4000.

### Proposed changes
I added port 4000 because that is the port mapped to the container's port 80. So if we want to view our app in the browser or with curl we need to include port 4000 in the url.

### Related issues (optional)
The screenshot image should be updated, too.

Fixes #6904

### DIFF
--- a/get-started/part3.md
+++ b/get-started/part3.md
@@ -158,7 +158,7 @@ is not filtered by service:
 docker container ls -q
 ```
 
-You can run `curl -4 http://localhost` several times in a row, or go to that URL in
+You can run `curl -4 http://localhost:4000` several times in a row, or go to that URL in
 your browser and hit refresh a few times.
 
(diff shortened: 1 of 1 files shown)


joaofnfernandes:
Thanks for the fix. Let's ship it!

dutchminator:
Thanks! :)

PR #6899, opened 2018-06-18
Correctly document "admin" users in UCP "Migrating Kubernetes Roles"

ingshtrom:
### Proposed changes

"admin" users are actually users with the "Full Control" role of the kubernetes namespace that is being edited. 

n/a


ingshtrom:
I tried to fix the spacing on the last line, but even after copying the text from master, it still shows changes. ¯\_(ツ)_/¯

joaofnfernandes:
Thanks for the edit! Given that this is an open repository, I've edited your initial comment to remove mentions to private repositories. Let's ship it.

PR #6911, opened 2018-06-20
Fix formatting of docker compose v3 example

Lappro:
### Proposed changes

The example docker-compose file for compose v3 does not have properly formatted yaml due to incorrect use of markdown.

This PR makes it proper markdown format so nice indenting (and yaml syntax highlighting) is provided.

### DIFF
--- a/compose/compose-file/index.md
+++ b/compose/compose-file/index.md
@@ -24,12 +24,8 @@ how to upgrade, see **[About versions and upgrading](compose-versioning.md)**.
 
 ## Compose file structure and examples
 
-<div class="panel panel-default">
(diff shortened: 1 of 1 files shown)


joaofnfernandes:
I understand what you're trying to do, but I don't agree with merging this change. Here's why:
* If you copy-paste the sample, you'll notice that it is properly indented, unless I'm missing something.
* The sample is purposefully long, so we added a bit of logic to make it collapsible so that users can ignore it without having to scroll a lot.

I think we should keep things as they are. For that reason, I'm closing this PR 😢but feel free to comment.

PR #6925, opened 2018-06-22
Update install-machine.md

ravigehlot:

### DIFF
--- a/machine/install-machine.md
+++ b/machine/install-machine.md
@@ -69,7 +69,7 @@ as:
     active machine
 
 Confirm the version and save scripts to `/etc/bash_completion.d` or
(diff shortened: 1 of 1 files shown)


paigehargrave:
@shin- PTAL - thank you!

shin-:
I don't think "running" the completion script will do anything. We may want to advise users to reload / open a new shell to load the completion script though, same way we do it [here](https://github.com/docker/docker.github.io/blame/master/compose/completion.md#L36-L37) for example.

paigehargrave:
Thank you for your feedback! We have received and reviewed your changes. After a technical review we did not incorporate your suggested changes. If you have additional concerns, submit a new issue in this repository.

PR #6926, opened 2018-06-22
Fix indenting in sample file listing.

rlundy:
### Proposed changes

Add one line break.  This fixes the indenting in the sample file listing.  See rich diff view to get the full effect.

### DIFF
--- a/compose/compose-file/index.md
+++ b/compose/compose-file/index.md
@@ -29,6 +29,7 @@ how to upgrade, see **[About versions and upgrading](compose-versioning.md)**.
     Example Compose file version 3
     <i class="chevron fa fa-fw"></i></div>
     <div class="collapse block" id="collapseSample1">
+
 <pre><code>
 version: "3"
 services:


rlundy:
I realize the contribution guidelines say:

> Don't change whitespace or line wrapping in parts of a file you are not editing for other reasons. Make sure your text editor is not configured to automatically reformat the whole file when saving.

However, I believe this is an exception, because this missing line break completely screws up the formatting of the sample docker-compose file, making it all but unreadable.

Reviewing and merging this pull request should only take a few minutes.

thaJeztah:
I notice that with this change, the closing `</div>` becomes visible (at least on GitHub);

screen shot 2018-07-16 at 17 33 12


On the documentation (the primary use for files in this repository are for reading on docs.docker.com, not on GitHub, as there's sometimes formatting/parsing instructions intended for docs.docker.com), the file seems to render OK (before this change);

screen shot 2018-07-16 at 17 33 45

So, wondering what's causing it to not show up properly on GitHub

thaJeztah:
looks like un-indenting the `</div>` at line 126/127 fixes that; https://github.com/docker/docker.github.io/pull/6926/files#diff-06edd26f33bf2c3f6ee590618f41c62cL126

Mixing HTML and Markdown remains a bit finicky 😓

paigehargrave:

PR #6930, opened 2018-06-22
swarm init for DockerToolbox requires --advertise-addr

camiltoma76:
See more info here:
https://stackoverflow.com/questions/38602903/docker-swarm-init-could-not-choose-an-ip-address-error

### DIFF
--- a/get-started/part3.md
+++ b/get-started/part3.md
@@ -119,6 +119,11 @@ Before we can use the `docker stack deploy` command we first run:
 docker swarm init
 ```
 
+>**Note**: If running via DockerToolbox, you need to run:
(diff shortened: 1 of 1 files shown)


bermudezmt:
@ollypom Thanks for looking into this. I think the note linking to part 4 under [Run your new load-balanced app]([https://docs.docker.com/get-started/part3/#run-your-new-load-balanced-app]) is sufficient and provides a logical flow to the tutorial.

@camiltoma76 Thank you for your feedback! After a careful review, we have decided not to incorporate your suggested changes at this time. If you have additional concerns, submit a new issue in this repository.

PR #6933, opened 2018-06-24
Clear confusion in the restart policies

meshde:
#### Why the changes?
The docs for the restart policy `unless-stopped` were incorrect: 
  > Restart the container unless it is explicitly stopped or Docker itself is stopped or restarted.  

This implies that a container started with this policy will **not be restarted** if the Docker daemon is restarted, which is incorrect. 

#### What was done?
The functioning of the policy `unless-stopped` was clearly explained. For the sake of clarity, the `active` policy was also explained in a bit more detail.

### DIFF
--- a/config/containers/start-containers-automatically.md
+++ b/config/containers/start-containers-automatically.md
@@ -28,8 +28,8 @@ any of the following:
 |:-----------------|:------------------------------------------------------------------------------------------------|
 | `no`             | Do not automatically restart the container. (the default)                                       |
 | `on-failure`     | Restart the container if it exits due to an error, which manifests as a non-zero exit code.     |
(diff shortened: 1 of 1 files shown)


paigehargrave:
@ahh-docker @bermudezmt -  I did some quick research and added a couple words to this in the 'unless-stopped' section. I think this just needs  final look by a technical reviewer (thaJestah) to make sure it's accurate.

paigehargrave:
@thaJeztah - I did some quick research and added a couple words to this in the 'unless-stopped' section. Can you take a final look to make sure it's accurate. @L-Hudson - Laureen - FYI.

JustinINevill:
@paigehargrave to test behaviour before merge.

PR #6944, opened 2018-06-26
Docker for Mac: add page on disk space

djs55:
### Proposed changes

The disk space usage of Docker on Mac has been difficult to manage and has confused lots of people, see

https://github.com/docker/for-mac/issues/371

This patch adds a top-level section to the Docker for Mac docs alongside networking which explains

- where the space has gone
- how to reclaim some of it

In future I'd like to add more about why some systems use `Docker.raw` and some `Docker.qcow2` but we're in a transitional state at the moment. It'll be simpler to explain after the next stable version has been released.

### Related issues (optional)

Related to https://github.com/docker/for-mac/issues/371


gondalez:
Thanks for your work! I think this will help a lot.

Do you think it would be useful to communicate this in the docker mac UI?
Something like this on the disk tab (the learn more link would lead to the docs in this PR):

**optimal** ✅ 
your system supports raw mode which is optimal for docker mac. 
[learn more]

**compatibility mode** ⚠️ 
your version of macos does not support raw mode so docker mac is using qcow. 
upgrade to the latest macos to make docker faster. 
[learn more]

**compatibility mode** ⚠️ 
your image is on a disk not formatted as APFS so docker mac is using qcow.
put your image on an APFS disk or upgrade your hardware to make docker faster.
[learn more]

**compatibility mode** ⚠️ 
your version of macos supports raw mode but you are running qcow from a prior install. 
[remove all data] to make docker faster. 
[learn more]

djs55:
@gondalez I think that's a good idea. What do you think, @gtardif ?

paigehargrave:
@gtardif - PTAL - thank you!

docker-for-mac/space.md
```diff
@@ -0,0 +1,88 @@
...
+### To delete unnecessary containers and images
+
+To check whether you have too many unnecessary containers and images, first list images
+with:
+```bash
+$ docker image ls
+```
+then list containers with:
+```bash
+$ docker container ls -a
```
thaJeztah: Might want to use `docker system df` instead of checking only containers and images; doing so also shows how much the actual use is, and how much is reclaimable https://docs.docker.com/engine/reference/commandline/system_df/

docker-for-mac/space.md
```diff
@@ -0,0 +1,88 @@
...
+and then "Reveal in Finder":
+
+![Disk preferences](images/settings-disk.png)
+
+The preferences window also shows how much disk space the file is currently consuming.
+In this example it is consuming only 2.4 GB out of a maximum of 64 GB.
+
+Note that other tools may display the space usage of the file incorrectly.
+
+## If the file is too big
```
thaJeztah: Should we mention the actual vs virtual size of the file?

djs55: Yes good idea. It's mentioned further down but it would be good to bring it nearer the top since it's bound to be a frequent source of confusion.

paigehargrave: @thaJeztah @djs55 - Thank you both for the input! Please see my latest updates to see if this is what you are looking for and if more is needed.

docker-for-mac/space.md
paigehargrave: Same comment as above. Thank you for the input!

paigehargrave:
@thaJeztah @djs55 - Thank you both for the input! Please see my latest updates to see if this is what you are looking for and if more is needed.

docker-for-mac/space.md
djs55: This looks fine to me!

docker-for-mac/space.md
djs55: This also looks fine to me!

paigehargrave:
Merging this - will open up a separate PR for TOC update. Thank you @djs55 and @thaJeztah !

PR #6950, opened 2018-06-27
fix: configuration reference link

bjsun:
configuration.md not found.


tfoxnc:
@bermudezmt The source files for some pages on docs.docker.com aren't in this repo. They are pulled from other repos managed by engineering. If the link renders on docs.docker.com but not in GitHub, that is probably the case -- I think that is the case with this file.

For this pr, the link to `configuration.md` is correct and shouldn't be changed. Linking to the github.com path will still give you a 404 while browsing on GitHub, and it will not render properly on docs.docker.com.

PR #6953, opened 2018-06-27
Fix misspellings in dockercloud migration doc

hexate:
### Proposed changes

Fix misspellings in the dockercloud migration doc


joaofnfernandes:
Thanks for improving our docs! Let's ship it 🚀

PR #6954, opened 2018-06-27
Fix broken link

alecthegeek:
Changed link

Link to "variable substitution" was borked

NOTE: I cannot test this change

### DIFF
--- a/compose/reference/build.md
+++ b/compose/reference/build.md
@@ -22,7 +22,7 @@ Services are built once and then tagged, by default as `project_service`. For
 example, `composetest_db`. If the Compose file specifies an
 [image](/compose/compose-file/index.md#image) name, the image is
 tagged with that name, substituting any variables beforehand. See [variable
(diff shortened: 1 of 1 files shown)


compose/reference/build.md
```diff
@@ -22,7 +22,7 @@ Services are built once and then tagged, by default as `project_service`. For
 example, `composetest_db`. If the Compose file specifies an
 [image](/compose/compose-file/index.md#image) name, the image is
 tagged with that name, substituting any variables beforehand. See [variable
-substitution](#variable-substitution).
+substitution](compose/compose-file/#variable-substitution).
```
joaofnfernandes: For this to work it you need to make it an absolute link like `/compose/...`. As you have it right now, this will try to find a file at `/compose/reference/compose/compose-file`, which doesn't exist.

tfoxnc:
Thanks for the catch!

PR #6955, opened 2018-06-27
Update supported-platforms.md

carlfischer1:
try to clarify what applies to EE Basic and add Windows Server versions now support w 18.03 engine


carlfischer1:
cc @powersplay

joaofnfernandes:
It seems that this needed a rebase, so I've integrated this changes into #6956.

PR #6980, opened 2018-07-02
add script install section back

tfoxnc:
Per Carl Fischer's request


carlfischer1:
@tfoxnc is there a reason nothing has shown up for netify?

carlfischer1:
@tfoxnc still no netify :(

tfoxnc:
@carlfischer1 We had to re-implement. I will carry this pr into a new one so we can get the Netlify preview and then merge if it looks good.

PR #6989, opened 2018-07-04
Updated gelf.md invalid dash

dintort:
Fixed an ivalid dash in --log-opt gelf-address


dintort:
I have spent a few minutes banging my head against the wall at this problem as I copy-pasted the log-opt string from here:
docker: Error parsing reference: "?-log-opt" is not a valid repository/tag: invalid reference format.

L-Hudson:
Good to go!

PR #6990, opened 2018-07-04
Update django.md

Isakofff:
### Proposed changes

Dockerfile with python - if you write 'FROM python:3' it uses the latest version, today it's 3.7 and it's not working - syntax error. If you change it to concrete version like 'FROM python:3.5' then it's work fine. Thank you.


L-Hudson:
Addressed in a separate PR. Closing this now. Thanks for your patience, @Isakofff

PR #7002, opened 2018-07-09
Update save/restore documentation with a specific section we can link to.

gtardif:
cc @thaJeztah this will be usefull to link from Docker Desktop message dialogs where required.

### DIFF
--- a/docker-for-mac/faqs.md
+++ b/docker-for-mac/faqs.md
@@ -57,19 +57,29 @@ to save images and export the containers you need, then uninstall the current
 version before installing another. The workflow is described in more detail
 below.</font><br>
 
-Do the following each time:
+#### How to save / restore images
 
-1.  Use `docker save` to save any images you want to keep. (See
-    [save](/engine/reference/commandline/save) in the Docker Engine command
-    line reference.)
+The following procedure can be used to save/restore images and container data,
+for example if you want to switch between Edge and Stable, or reset your VM
+disk:
 
-2.  Use `docker export` to export containers you want to keep. (See
-    [export](/engine/reference/commandline/export) in the Docker Engine
-    command line reference.)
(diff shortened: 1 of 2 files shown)


thaJeztah:
Information looks ok, but may not give the full story - images are actually the "easy" part (if you push/pull to a registry; which should be the recommended approach). There's more things people possibly care about; so we should have something describing (I'm not sure if there's anyone currently that has the bandwidth to write that);

- deciding what images to preserve (i.e., not much use of save/load "ubuntu", because it's easier to pull it again)
- how to backup/restore data in _volumes_ (important, as that's where your data should be)
  - reminds me that I'd love to see a separate storage location for that (separate disk on Docker Desktop)
- how to backup/restore _secrets_ and _configs_
- "commit" containers that have manual modifications (not best practice, but definitely something people do)
- (likely less important); container logs

Things you won't be able to preserve;

- containers (you can preserve filesystem changes (`docker commit`), but won't preserve the options and configuration that the container was started with. While you _should_ use (e.g.) a `docker-compose` file for that, not everyone does, and people may not recall what options they used when starting the container.
  - possibly assisting; do a `docker inspect` of those containers and write those to a file, to know what was used to create it.
- services (again, ideally, use a compose file, but otherwise `docker service inspect`)
- networks (same)

gtardif:
@thaJeztah can we get this (small) improvements merge, so that we can at leat use the link, and keep improving this in another PR later on (indeed this is also a bandwidth issue) ?

gtardif:
Just added a link to data volume backup/restore procedure, and changed section title to "How to save / restore data".

gtardif:
@thaJeztah let me know if there is some minor changes that needs be done so that we can merge this, it would be good to have this section so that we can link to it in the 18.06 RC release of Desktop.

PR #7003, opened 2018-07-09
Add doc for awslogs-endpoint log option

haikuoliu:
This is a follow-up PR for this [PR](https://github.com/moby/moby/pull/37374) against Moby.

### Proposed changes

Add explanation for the new `awslogs-endpoint` log option and a note for it.


L-Hudson:
Good to go. Sorry approval took so long; we're working through the backlog. Thank you for contributing, and also for including cute animals, @haikuoliu!

PR #7015, opened 2018-07-11
add back install using script take2

tfoxnc:


tfoxnc:
Didn't work to force Netlify to trigger so closing

Issue #6654, opened 2018-05-07
Swarm routing mesh documents missing step for applying node constraints


ericsmalling:
### Problem description
https://docs.docker.com/ee/ucp/interlock/deploy/production/ fails to mention the step outlined at https://interlock-dev-docs.netlify.com/install/production/#configure-proxy-service which causes the proxy containers to actually get the constraint configurations

The "Note" on the page about deploying a service to cause the proxies to restart does not actually put the new node constraints into them.  Per Slack chat with @ehazlett , this last step (linked above) must be performed:

```
$> docker service update --replicas=2 \
    --constraint-add node.labels.nodetype==loadbalancer \
    --stop-signal SIGQUIT \
    --stop-grace-period=5s \
    $(docker service ls -f 'label=type=com.docker.interlock.core.proxy' -q)
```

### Problem location

https://docs.docker.com/ee/ucp/interlock/deploy/production/ at the "Known issue" box near the bottom of the document.

### Project version(s) affected

UCP 3.0.0+

### Suggestions for a fix
Copy text from https://interlock-dev-docs.netlify.com/install/production/#configure-proxy-service in place of the "Known issue"

ericsmalling:
Note - the host mode page also is missing a similar step as described at https://interlock-dev-docs.netlify.com/usage/host_mode/#configure-proxy-services

Basically, after changing either of these configurations, the proxies service has to be updated to re-schedule it correctly.

martencassel:
This is not clear...

How is interlock-proxy scheduling configurations updated ? 

    - Can i just update ProxyConstraints in the configuration file, 
      by updating the configuration object ? 

      Will interlock service run a docker service update using 
      the new ProxyConstraints if i update its configuration  file ? 

    - Do i have to run a docker service update myself ? 

    - The ProxyConstraints is not documented in the configurations listning among other Proxy* properties:

    https://docs.docker.com/ee/ucp/interlock/deploy/configuration-reference/#core-configurations

L-Hudson:
Closing this ticket due to its age, and the impending refactor. If you think this is in error, feel free to reopen. Thanks!

PR #7016, opened 2018-07-11
Interlock: correct proxy constraint procedure

trapier:
As of UCP 3.0.2 changes to `ProxyConstraints` are not respected on config update.

Fixes https://github.com/docker/docker.github.io/issues/6654

### DIFF
--- a/ee/ucp/interlock/deploy/production.md
+++ b/ee/ucp/interlock/deploy/production.md
@@ -52,10 +52,19 @@ The command should print "loadbalancer".
 
 ## Configure the ucp-interlock service
 
-Now that your nodes are labelled, you need to update the `ucp-interlock`
+Now that your nodes are labelled, you need to update the `ucp-interlock-proxy`
 service configuration to deploy the proxy service with the correct constraints.
 
-Add another constraint to the `ProxyConstraints` array:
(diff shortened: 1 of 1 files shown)


trapier:
fixed shell line ending and force-pushed

trapier:
> why perform a `service ls --filter=`? is this more better than

> `docker service update ... ucp-interlock-proxy`?

If you have more than one service cluster, what is the service name for the non-default clusters? ;-)

trapier:
Had a chat with @mbentley.
 
This PR currently cuts when it needs to add.  Core config determines what happens when `ucp-interlock-proxy` needs to be created.  Like at install or restore from backup time, or in a `docker service rm ucp-interlock-proxy` moment.  So this section of the docs really needs to describe precedence: sometimes you'll want to `docker service update`, but you'll also want to make sure you edit the core config.

I'll fix this up and drop the [DO NOT MERGE].

adamancini:
@trapier makes total sense

bermudezmt:
@adamancini can you please have another look after last commit from @trapier ?

ahh-docker:
@L-Hudson ready for your review then merge provided it's good.

ahh-docker:
@adamancini and @trapier thanks for your time on this!

PR #7018, opened 2018-07-12
Fixed anchor error #depends-on -> #depends_on.

javabrett:
Target https://docs.docker.com/compose/compose-file/#depends_on .  Location https://docs.docker.com/compose/startup-order/ .


javabrett:
Requesting a review for this ... thanks.

PR #7024, opened 2018-07-12
change inspectDockerimage to inspectDockerImage

andrewhsu:
### Proposed changes

Fixed typo `inspectDockerimage`, should be `inspectDockerImage` to match the binary name that needs to be called.


gforghetti:
@andrewhsu Looks good to me.

PR #7045, opened 2018-07-18
interlock: supports only one path per host

trapier:
Current docs show single context root and the option is singular, but we've had a few folks try for multiple contexts on the same hostname.  Added a call out to avoid this issue.

### DIFF
--- a/ee/ucp/interlock/usage/context.md
+++ b/ee/ucp/interlock/usage/context.md
@@ -29,6 +29,8 @@ $> docker service create \
     ehazlett/docker-demo
 ```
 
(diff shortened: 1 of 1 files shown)


peter-docker:
Deploy preview for *modest-edison-79d82e* ready!

Built with commit 6f0252bbf910ad7e0145ac33be4c1134ac35a97b

https://deploy-preview-7045--modest-edison-79d82e.netlify.com

trapier:
Thanks.  Dropped `**Note:**`.  Added `{: .important}`.

PR #7061, opened 2018-07-19
Provide an automatic latest example for Linux

dragon788:
### Proposed changes

Regardless of the version the documentation specifies, this queries the API for the same information that the documentation recommends to find the latest release. It uses `jq` but could also be made to work with just `grep` if no dependencies are desired.

### DIFF
--- a/compose/install.md
+++ b/compose/install.md
@@ -112,10 +112,16 @@ by step instructions are also included below.
 1.  Run this command to download the latest version of Docker Compose:
 
     ```bash
+    # If you have jq installed this will automatically find the latest release binary for your architecture and download it
+    curl --silent "https://api.github.com/repos/docker/compose/releases/latest" | jq --arg PLATFORM_ARCH "$(echo `uname -s`-`uname -m`)" -r '.assets[] | select(.name | endswith($PLATFORM_ARCH)).browser_download_url' | xargs sudo curl -L -o /usr/local/bin/docker-compose --url
+    ```
+    
+    ```bash
+    # If you don't want to install jq or want a specific version of docker-compose specify it in the URL
     sudo curl -L https://github.com/docker/compose/releases/download/{{site.compose_version}}/docker-compose-$(uname -s)-$(uname -m) -o /usr/local/bin/docker-compose
     ```
 
-    > Use the latest Compose release number in the download command.
+    > Use the desired Compose release number in the download command.
     >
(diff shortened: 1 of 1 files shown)


peter-docker:
Deploy preview for *modest-edison-79d82e* ready!

Built with commit 4ab68d6ac252dc647fddc4f99e9fe1796d38ab4d

https://deploy-preview-7061--modest-edison-79d82e.netlify.com

bermudezmt:
Hi @dragon788 - I got this error running the `curl` with `jq` parsing command:
```jq: error: PLATFORM_ARCH/0 is not defined at <top-level>, line 1:
.assets[] | select(.name | endswith($PLATFORM_ARCH)).browser_download_url```
It looks like dashes are not supported in parsing `jq` keys:
```jq: error: PLATFORM_ARCH/0 is not defined at <top-level>, line 1:
.assets[] | select(.name | endswith($PLATFORM_ARCH)).browser_download_url  ```

There's some suggested workarounds but I wanted to hear your thoughts.

thaJeztah:
> Regardless of the version the documentation specifies, this queries the API for the same information that the documentation recommends to find the latest release. It uses jq but could also be made to work with just grep if no dependencies are desired.

I'm a bit "-1" on this change; this change is adding quite some complexity to address situations where the documentation is out of date; for those situations, the right solution is to fix the version that's listed in the documentation.

I _do_ see this section for Windows about installing specific versions of compose;

> Then run the following command to download Docker Compose, replacing `$dockerComposeVersion` with the specific version of Compose you want to use:
>
> ```powershell
> Invoke-WebRequest "https://github.com/docker/compose/releases/download/$dockerComposeVersion/docker-compose-Windows-x86_64.exe" -UseBasicParsing -OutFile $Env:ProgramFiles\docker\docker-compose.exe
> ```

I think that distracts from the main flow in the document; perhaps it would be better to remove that, and have a short section about finding the available versions from the releases page on GitHub https://github.com/docker/compose/releases (to install an older version of compose). But I'm even wondering if we _should_ mention older versions (as the recommendation should be to install the _current_ version.


@shin- wdyt?

ahh-docker:
@shin- PTAL.

shin-:
I agree with @thaJeztah ; I don't think this change is warranted, and I don't think users should be encouraged to download older versions of Compose.

PR #7113, opened 2018-07-26
Fixing SUSE install from repo instructions to correctly locate gpg key

millerdz:
In the **Install using the repository** section, the GPG implied location was identified as `<DOCKER-EE-URL>/sles/12.3/<ARCHITECTURE>/stable-<VERSION>/sles/gpg` which does not exist. The correct location is `<DOCKER-EE-URL>/sles/gpg`. I introduce a second variable here called `DOCKER_EE_BASE_URL` which will be used to distinguish between the root level of the directory and aid in correctly locating the gpg key for the `rpm --import` command.


ahh-docker:
@kolyshkin PTAL. Thanks!

kolyshkin:
This needs to be rebased. Otherwise LGTM

ahh-docker:
@seemethere PTAL. Thanks.

ahh-docker:
@L-Hudson this is ready for you, then it can be merged.

PR #7115, opened 2018-07-26
DTR LB config link

omegamormegil:
Updated article with note about DTR LB requirements. This tripped up a customer on SR# 50917.

### DIFF
--- a/ee/ucp/admin/configure/join-nodes/use-a-load-balancer.md
+++ b/ee/ucp/admin/configure/join-nodes/use-a-load-balancer.md
@@ -34,6 +34,8 @@ or port number.
     * Use the same load balancer with multiple virtual IPs.
 * Configure your load balancer to expose UCP or DTR on a port other than 443.
 
+> Additional requirements
+>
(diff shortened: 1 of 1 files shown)


ee/ucp/admin/configure/join-nodes/use-a-load-balancer.md
```diff
@@ -34,6 +34,8 @@ or port number.
     * Use the same load balancer with multiple virtual IPs.
 * Configure your load balancer to expose UCP or DTR on a port other than 443.
 
+> **Note:** In addition to configuring your load balancer to distinguish between UCP and DTR, configuring a load balancer for DTR has [additional requirements](https://docs.docker.com/ee/dtr/admin/configure/use-a-load-balancer/#load-balance-dtr).
+
```
tfoxnc: 1. Perhaps this should be an important admonition since it can cause it not to work: ee/ucp/admin/configure/join-nodes/use-a-load-balancer.md

2. This links should be relative since it is on the same site.

tfoxnc: @ahh-docker will reformat.

PR #7116, opened 2018-07-26
Linked to UCP and DTR LB requirements

omegamormegil:
Cross linked with UCP LB document since that document covers configuring load balancers for both UCP and DTR.

### DIFF
--- a/ee/dtr/admin/configure/use-a-load-balancer.md
+++ b/ee/dtr/admin/configure/use-a-load-balancer.md
@@ -30,7 +30,11 @@ replicas.
 DTR does not provide a load balancing service. You can use an on-premises
 or cloud-based load balancer to balance requests across multiple DTR replicas.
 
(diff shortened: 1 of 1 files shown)


ee/dtr/admin/configure/use-a-load-balancer.md
```diff
@@ -30,6 +30,8 @@ replicas.
 DTR does not provide a load balancing service. You can use an on-premises
 or cloud-based load balancer to balance requests across multiple DTR replicas.
 
+> **Note:** If you also use UCP there are [additional requirements](https://docs.docker.com/ee/ucp/admin/configure/join-nodes/use-a-load-balancer/#load-balancing-ucp-and-dtr) if you plan to load balance both UCP and DTR using the same load balancer. 
+
```
tfoxnc: Please use newer admonition formatting: https://github.com/docker/docker.github.io/blob/master/test.md#admonitions-notes

PR #7119, opened 2018-07-26
Dockerfile chown not yet supported on Windows

johnstep:
Signed-off-by: John Stephens <johnstep@docker.com>


bermudezmt:
Supported as of https://github.com/docker/docker.github.io/blob/master/engine/release-notes.md#1809 release.

PR #7138, opened 2018-07-30
Update backup.md

gonzalo-contador:


paigehargrave:
@ahh-docker @bermudezmt I think we might be able to close this. Latest version of docs (https://github.com/gonzalo-contador/docker.github.io/blob/patch-5/reference/ucp/3.0/cli/backup.md) has this example (with a couple minor differences).

peter-docker:
This is an unnecessary change and no real reason to merge. Closing issue.

PR #7160, opened 2018-08-02
Update ubuntu.md

amaddio:
Updated the old `docker run` command with the new sub command style `docker container run`.


ahh-docker:
Both commands are still valid. Will confirm if this change is accepted or if we need to make an admonition.

PR #7169, opened 2018-08-04
get-started: remove trusted host option in pip install example

ceh:
### Proposed changes

Remove usage of pip install option `--trusted-host`. I don't think it's sane to endorse usage of this option because it doesn't verify TLS connections, especially if new users will use this as a starting template for Python projects.

See https://pip.pypa.io/en/stable/reference/pip/#general-options for reference about the option.

What do you think?


L-Hudson:

PR #7178, opened 2018-08-06
NFS example for driver_opts

adamancini:
### Proposed changes

People ask me how to do this all the time, and I can't find a good example documented anyhwere else.  I figure this is a better example of the actual usage of the `driver_opts` parameter with a real-world use case.

### DIFF
--- a/compose/compose-file/index.md
+++ b/compose/compose-file/index.md
@@ -1946,9 +1946,12 @@ Specify a list of options as key-value pairs to pass to the driver for this
 volume. Those options are driver-dependent - consult the driver's
 documentation for more information. Optional.
 
-     driver_opts:
-       foo: "bar"
-       baz: 1
+    volumes:
+      example:
+        driver_opts:
+          type: "nfs"
+          o: "addr=10.40.0.199,nolock,soft,rw"
+          device: ":/docker/example"
 
 ### external
 


adamancini:
@trapier @tfox for review

trapier:
Tested the volume spec, and it works for me.  Love that this makes the example actually useful to someone!

Since we're in effect also recommending some NFS mount options here, @squizzi: have any feelings about  `soft,nolock` / anything different you'd prefer?

```
[vagrant@worker-1 ~]$ docker volume inspect nfstest_example  |jq '.[]|.Options'
{
  "device": ":/srv/nfs",
  "o": "addr=192.168.121.1,nolock,soft,rw",
  "type": "nfs"
}
```

output of `mount` in host NS` after deploying this volume to CentOS Linux release 7.4.1708:

```
:/srv/nfs on /var/lib/docker/volumes/nfstest_example/_data type nfs (rw,relatime,vers=3,rsize=1048576,wsize=1048576,namlen=255,soft,nolock,proto=tcp,timeo=600,retrans=2,sec=sys,mountaddr=192.168.121.1,mountvers=3,mountproto=tcp,local_lock=all,addr=192.168.121.1)
```

adamancini:
@trapier what syntax version did you run against?  Want to confirm minimum compose syntax version

adamancini:
I like soft,nolock as it prevents locking when the NFS server is unavailable - should immediately return errors which IMO is preferable than a stuck process.

trapier:
> what syntax version did you run against? Want to confirm minimum compose syntax version

`version: "3.2"`  Also works on 3.1 and 3.0.

> I like soft,nolock as it prevents locking when the NFS server is unavailable - should immediately return errors which IMO is preferable than a stuck process.

Works for me.

@tfoxnc in my opinion this is good for merge.

trapier:
Yep.   Still looks good for merge to me.   Thanks @ahh-docker!

PR #7182, opened 2018-08-07
Add missing cross-references to "Not supported for docker stack deploy"

jeremydouglass:
### Proposed changes

Added three of the option entries which are individually labeled as "ignored when deploying a stack in swarm mode" -- but which do not appear in the list "Not supported" list.

-  `cap_add, cap_drop`
-  `depends_on`

Also for `tmpfs` added a list note that it is not supported in `stack deploy` for `(version 3-3.5)` -- as described on its individual entry


L-Hudson:

PR #7187, opened 2018-08-08
notary: set selinux context while installing

trapier:


ahh-docker:
Merging with @bermudezmt

PR #7191, opened 2018-08-09
Updating Django version in requirements.txt

a6kme:
### Proposed changes

`FROM python:3` in Dockerfile installs Python3.7 by default. And Django 1.X has no support for Python 3.7. 

https://code.djangoproject.com/ticket/28814


a6kme:
No activity from moderators.

PR #7203, opened 2018-08-09
Clarification on removing nodes from Cluster

david-yu:
Adding confirmation to show that node is shut down or left the swarm.


ahh-docker:
@yongshin Thanks. Do we need a screenshot for this? I think that would be a big help. Thanks.

yongshin:
screen shot 2018-09-24 at 9 12 12 pm

yongshin:
image

ahh-docker:
Approving with @bermudezmt

PR #7208, opened 2018-08-12
Add fixed-cidr-v6 field

Golit:
### Proposed changes
Since dockerd does not start without a ipv6 subnet if ipv6 is enabled it is convenient to add this field to the docs.


ahh-docker:
@mark-church PTAL. Thanks.

L-Hudson:

PR #7218, opened 2018-08-14
Correcting command to follow text

jdmills:
### Proposed changes

The proposed change replaced the "docker container rm" command with a "docker stop" command. This is proposed because
  - the text says that you want to stop the container and that it will be automatically removed due to being started with the --rm option.  
  - the command that is given ("docker container rm") will not work even if the container was not started with the --rm option because you must stop the container before removing it.


L-Hudson:

PR #7238, opened 2018-08-20
begin to ucp-ify interlock service clusters

trapier:


trapier:
Todos for this section:
- remove suggestion to deploy a swarm, since ucp already did this for the reader
- check that proxy replicas defaults to 1, and increase it to 2 (otherwise updates would interrupt traffic)

... but those can follow-up.

@adamancini: ptal!

ahh-docker:
@adamancini : can you please review, so I know whether or not to merge this? thanks!

PR #7245, opened 2018-08-21
Update ssh-add-keys-to-agent.md

VokaMut:
### Proposed changes

Changed the keys for the ssh-add command on Windows and Linux

### DIFF
--- a/_includes/content/ssh/ssh-add-keys-to-agent.md
+++ b/_includes/content/ssh/ssh-add-keys-to-agent.md
@@ -55,7 +55,7 @@ passphrases in your keychain.
 2.  Add your SSH private key to the ssh-agent.
 
     ```none
-    $ ssh-add -K ~/.ssh/id_rsa
+    $ ssh-add ~/.ssh/id_rsa
     ```
 
     If you created your key with a different name or have an existing key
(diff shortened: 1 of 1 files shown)


ahh-docker:
@trapier Can you please review this to make sure the change makes sense? I'm looking at the ssh-add man page, and it looks like this stores the key passphrase into memory. If this is a best practice, we can add it. If not, I'll close this. Thanks!

trapier:
The hosted swarm functionality these docs pertain to has been shut down:

> Important Cluster and application management services in Docker Cloud are shutting down on May 21. You must migrate your applications from Docker Cloud to another platform and deregister your Swarms.

Ref: https://docs.docker.com/docker-cloud/migration/

That being said:
- adding keys to `ssh-agent` is standard practice
- the PR addresses an error:
    - (tested) `ssh-add` on Linux has no `-K`
    - `-K` looks like an [osx-ism](https://apple.stackexchange.com/questions/48502/how-can-i-permanently-add-my-ssh-private-key-to-keychain-so-it-is-automatically)
- the pattern suggested by the pr follows [github](https://help.github.com/articles/generating-a-new-ssh-key-and-adding-it-to-the-ssh-agent/), which is heavily ssh-based.

The jekyll with this one was too strong for me, so I built it to make sure it does what it says on the tin.

LGTM.

Thanks for the improvement @VokaMut!

VokaMut:
@ahh-docker What is the state of my PR?

PR #7260, opened 2018-08-23
Remove DockerCon '18 SF advert from site

bermudezmt:
### Proposed changes

This will remove the DockerCon '18 advert on docs.docker.com


bermudezmt:
@peter-docker Where is _site/index.html supposed to come from? I see _site in .gitignore and .dockerignore.

PR #7291, opened 2018-08-30
Update part2.md

cfenwick74:
When running 'docker build -t friendly hello .', I encountered the following error on Step 4/7:
Could not open requirements file: [Errno 2] No such file or directory: 'requirements.txt'

I switched the WORKDIR and ADD steps in the dockerfile and then was able to build successfully.

### Proposed changes

Switch WORKDIR and ADD steps in dockerfile example in Get Started: Part 2

### DIFF
--- a/get-started/part2.md
+++ b/get-started/part2.md
@@ -65,11 +65,11 @@ your new Dockerfile.
 # Use an official Python runtime as a parent image
 FROM python:2.7-slim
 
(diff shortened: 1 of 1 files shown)


adrian-plata:
I ran the instructions as they're listed live and didn't run into an error. I'm going to close this request unless this issue pops up again.

PR #7304, opened 2018-08-31
Added Latest UCP DTR Offline Bundles

ollypom:
### Proposed changes

Updated the Docs to that latest build numbers: UCP 3.0.5 and 2.2.12. DTR 2.5.5

Added Offline Bundles for those Builds as well.


ollypom:
@ry4nz PTAL

ry4nz:
UCP LGTM

PR #7307, opened 2018-08-31
Adding port 12388 for UCP 3.1

yongshin:
### Proposed changes

Adding port 12388 for Kube server API direct access for UCP 3.1


yongshin:
cc @ollypom

PR #7328, opened 2018-09-06
fix swarm service constraint examples

trapier:
1. Be more specific about what happens when constraints aren't satified.
2. There is no `--global`.  Also addressed in network/overlay.md.
3. Label constraints do not match node labels without `node.label` prefix.

AFAIK neither 2 nor 3 have ever been correct

testedon: 17.06.2-ee-16

### DIFF
--- a/network/overlay.md
+++ b/network/overlay.md
@@ -241,9 +241,9 @@ When you connect to a published port on any swarm node (whether it is running a
 given service or not), you are redirected to a worker which is running that
 service, transparently. Effectively, Docker acts as a load balancer for your
 swarm services. Services using the routing mesh are running in _virtual IP (VIP)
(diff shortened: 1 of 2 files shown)


network/overlay.md
```diff
@@ -241,9 +241,9 @@ When you connect to a published port on any swarm node (whether it is running a
 given service or not), you are redirected to a worker which is running that
 service, transparently. Effectively, Docker acts as a load balancer for your
```
trapier: This section is confusing because it conflates the concepts of routing mesh, ingress routing mesh, and VIP, and then goes on to say that they can all be turned off by switching to `dns-rr` (which does not make sense in the ingress context).  Clarification would be a follow-up PR.

PR #7372, opened 2018-09-15
Propose an alternative ssh command

hestellez:
I didn't get it to work with the ip address, it took me a while to find the alternative way.

### DIFF
--- a/engine/examples/running_ssh_service.md
+++ b/engine/examples/running_ssh_service.md
@@ -52,6 +52,8 @@ Docker daemon host:
 
 ```bash
 $ ssh root@192.168.1.2 -p 49154
+# or
+$ ssh root@localhost -p 49154
 # The password is ``screencast``.
 root@f38c87f2a42d:/#
 ```


ahh-docker:
@trapier is this something we want to suggest as best practice? It looks like it might be a network configuration difference on the submitter's part possibly.

trapier:
<meta_criticism>
If this section of documentation is meant to help a user connect to a container's published port for the first time, then it was insufficient before this PR, and remains insufficient after it.  Hopefully those instructions are available elsewhere in the docs, and can be linked from here.
</meta_criticism>


The utility and security of ssh are not a function of the remote host address, so it's not possible to use best practice as a framework for feedback here.  This patch simply makes explicit one of the connection options in the preceding paragraph (emphasis mine) https://github.com/docker/docker.github.io/pull/7372/commits/f23d248ba9a08c8e1dad41970946d3358f14dd3c#diff-5f8208a295d080313a4e95319b359aa2R48:

> And now you can ssh as `root` on the container's IP address (you can find it
 with `docker inspect`) or on port `49154` of the Docker daemon's host IP address
 (`ip address` or `ifconfig` can tell you that) **or `localhost`** if on the
Docker daemon host:

The addition is harmless and would have helped its submitter, so I'm be OK with merging.

LGTM

Thanks @hestellez!

ahh-docker:
Contacted @trapier for some clarification and if this is ready to go.

ahh-docker:
Merge per @trapier

PR #7388, opened 2018-09-19
calicoctl: use docker/ucp-dsinfo and client bundle

trapier:
Use calicoctl from docker/ucp-dsinfo (available as of 3.0.2) via UCP
client bundle.  Removes the need for UCP host shell access with `sudo`,
which some operators do not have.

Also:
- Use `k8s-app=calico-node` when finding calico-node to stop on the rr.
- Place namespace first when using `kubectl`, since that argument
  doesn't change.
- Reword "other `calico-node` pods on other nodes".  At that point in
  the procedure we just deleted `calico-node` pods on the rr, so the
  first "other" isn't necessary.

testedon: ucp 3.0.2

### DIFF
--- a/ee/ucp/admin/configure/deploy-route-reflectors.md
+++ b/ee/ucp/admin/configure/deploy-route-reflectors.md
@@ -127,45 +127,41 @@ kubectl create -f calico-rr.yaml
 ## Configure calicoctl
 
 To reconfigure Calico to use Route Reflectors instead of a node-to-node mesh,
-you'll need to SSH into a UCP node and download the `calicoctl` tool.
-
-Log in to a UCP node using SSH, and run:
-
-```
-sudo curl --location https://github.com/projectcalico/calicoctl/releases/download/v3.1.1/calicoctl \
-  --output /usr/bin/calicoctl
-sudo chmod +x /usr/bin/calicoctl
-```
-
-Now you need to configure `calicoctl` to communicate with the etcd key-value
-store managed by UCP. Create a file named `/etc/calico/calicoctl.cfg` with
-the following content:
+you'll need to tell `calicoctl` where to find the etcd key-value store managed
+by UCP.  From a CLI with a UCP client bundle, create a shell alias to start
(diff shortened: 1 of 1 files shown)


trapier:
expressed more of "what i did" in commit message.

ee/ucp/admin/configure/deploy-route-reflectors.md
```diff
@@ -127,45 +127,39 @@ kubectl create -f calico-rr.yaml
...
+  --pid host \
+  --net host \
+  -e ETCD_ENDPOINTS=127.0.0.1:12378 \
+  -e ETCD_KEY_FILE=/ucp-node-certs/key.pem \
+  -e ETCD_CA_CERT_FILE=/ucp-node-certs/ca.pem \
+  -e ETCD_CERT_FILE=/ucp-node-certs/cert.pem \
+  -v /var/run/calico:/var/run/calico \
+  -v ucp-node-certs:/ucp-node-certs:ro \
+  {{ page.ucp_org }}/ucp-dsinfo:{{ page.ucp_version }} \
+  calicoctl \
```
trapier: @wsong: this is a mouthful of an alias, but I really like the property of not needing to shell into a cluster node and execute something with `sudo` to get the job done.  Makes `calicoctl` feel like `kubectl`.  wdyt?

ollypom: @trapier Yea I absolutely approve of removing a `ssh to your node` type instruction. And i'm actually leaning towards your solution, just because I don't think many operators would want to expose the etcd cluster port externally, or risk leaving admin etcd certificates lying on a workstation. 

The only down sides being, you might have to run constraints like `-e constraint:ostype==linux`.

Also we have a third set of calicoctl instructions here. https://docs.docker.com/ee/ucp/kubernetes/install-cni-plugin/#disable-ip-in-ip-overlay-tunneling

trapier:
Added `-e constraint:ostype==linux` `UCP_VERSION` calculation.  Ready for review!

trapier:
simplified calculation of ucp version.

PR #7410, opened 2018-09-23
Update completion.md

popovicsandras:
### Proposed changes

Following the official Docker compose documentation, while reading how to apply the bash command-line completion, I realised copy & pasting the code snippet didn't work. See my one liner PR to understand the reasons. (My bash_completion.d was not under /etc, like it is neither for many brew users).


bermudezmt:
Hey @popovicsandras , this has been addressed by #8239 . Depending on your macOS version, `brew` _could_ be installed on a different path by default. Please open an issue if there are still problems with the updated page.

PR #7416, opened 2018-09-25
Clarifying some confusing explanations for limiting cpus resource.

artdgn:
### Proposed changes

In the current version of the docs, there are two places in which the explanation given for limiting cpus resources in docker-compose format is confusing. It is possible to understand them as a ratio of total CPU power of the machine instead of a ratio of relative to one CPU core. 
The detailed explanations in the docker-compose format documentation are not confusing, but someone reading just the tutorials may be confused and not understand why their service is underpowered relative to their specifications. 
I've tried to make it less confusing while not adding too much text.


L-Hudson:
Good to go

PR #7423, opened 2018-09-26
Update compose install command to always pick the latest version.

nphmuller:
This changes the Compose download command for Linux systems to always point to the latest version. 
This makes the command easier to use in installation scripts.


usha-mandya:
Thanks for the review @ulyssessouza.

@nphmuller Thanks for the suggestion. Unfortunately, I'll have to close the PR without merging your changes due to the reasons that Ulysses highlighted above. Feel free to continue the conversation after I close.

PR #7431, opened 2018-09-27
soften statement about storage drivers low I/O

BretFisher:
### Proposed changes

While performance is affected when comparing storage driver I/O to native filesystem performance, to summarize it as "low" is a bit harsh, so I softened it up to prevent people from overreacting from the info. More details are given later in the document anyway.


paigehargrave:
@thaJeztah @BretFisher PTAL and see if what I added will suffice. Thanks!

BretFisher:
LGTM Thanks!

paigehargrave:
Added to current version of the file so am closing this. @thaJeztah @BretFisher Thank you both!

PR #7433, opened 2018-09-27
remove aufs "default storage driver" statements

BretFisher:
### Proposed changes

In many cases, `aufs` is no longer the default storage driver, so I just changed the examples to not mention the default. If you'd like it to actually state overlay2 is the default in most/some cases I can change that, but felt it was best left to the "Selecting a storage driver" page.

### DIFF
--- a/storage/storagedriver/index.md
+++ b/storage/storagedriver/index.md
@@ -292,7 +292,7 @@ layer. This means that the writable layer is as small as possible.
 
 When an existing file in a container is modified, the storage driver performs a
 copy-on-write operation. The specifics steps involved depend on the specific
-storage driver. For the `aufs`, `overlay` and `overlay2` drivers, the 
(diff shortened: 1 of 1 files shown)


storage/storagedriver/index.md
```diff
@@ -292,8 +292,8 @@ layer. This means that the writable layer is as small as possible.
 
 When an existing file in a container is modified, the storage driver performs a
 copy-on-write operation. The specifics steps involved depend on the specific
-storage driver. For the default `aufs` driver and the `overlay` and `overlay2`
-drivers, the copy-on-write operation follows this rough sequence:
+storage driver. For the `aufs`, `overlay` and `overlay2` drivers, the
```
thaJeztah: nit: the documentation uses Oxford comma, so there should be a comma before `and overlay` 😅 

(just a nit, so I'm ok with merging as-is, if it's not updated)

PR #7436, opened 2018-09-27
Clean up Data backup command

clemenko:
The restore procedure was incorrect. Simplifying the command to work with the restore.

### Proposed changes

The restore procedure was incorrect. Simplifying the command to work with the restore.


ahh-docker:
Approving with @bermudezmt

PR #7444, opened 2018-09-28
linked ucp user bundle and clarified --ucp-node

cdunkelb:
### Proposed changes

Linked the UCP user bundle to instructions on how to download and use a UCP user bundle. Added a note to clarify that --ucp-node is the node you are trying to install DTR on. See a lot of people setting this to one of their UCP manager nodes by accident or ignorance.

### Related issues (optional)

Overall, the docs do not advocate the use of the user/client bundle enough. Once UCP is installed, everything should be run through the user/client bundle. It should be a rare exception to run commands directly against the engine.


ahh-docker:
@bermudezmt and I reviewed. GTG.

PR #7453, opened 2018-10-01
docker-for-mac: make command line more intuitive

waldyrious:
For a getting-started guide, it makes sense to use the expanded option names that make it explicit what each option does. Later on users will have the chance to learn the shortcuts.

The equal sign convention is also used for the options that take arguments, to make it clearer what each part of that command line refers to.


waldyrious:
Direct link to the relevant section in the preview, for convenience: https://deploy-preview-7453--docsdocker.netlify.com/docker-for-mac/#explore-the-application

PR #7455, opened 2018-10-02
get-started/part2: clarify usage of -t/--tag

waldyrious:
As this is an introductory document, using short option and relying on default behaviors without explanation may cause some confusion.

This change it explicit how the --tag option is being used.

It also uses an equal sign to visually group the option and its value, so that the . at the end of the command doesn't go unnoticed.


ahh-docker:
I'll also include `-t` as well as `tag` for clarification, so users can do either one.

waldyrious:
@ahh-docker I can make that change in this PR if you wish. Let me know what you prefer.

ahh-docker:
Sure feel free :) Let me know when you're done.

waldyrious:
@ahh-docker is 48f1f64 what you meant?

ahh-docker:
@waldyrious I cleaned up the wording a bit. If this looks good to you, let me know and I'll merge the change. Thanks!

waldyrious:
Sure, that change makes sense :) feel free to merge whenever convenient.

PR #7456, opened 2018-10-02
Update override-a-vulnerability.md

bermudezmt:
- Update note on hiding vulnerability as per Slack discussion

### DIFF
--- a/ee/dtr/user/manage-images/override-a-vulnerability.md
+++ b/ee/dtr/user/manage-images/override-a-vulnerability.md
@@ -22,8 +22,8 @@ vulnerability, and click **hide**.
 
 ![Vulnerability list](../../images/override-vulnerability-2.png){: .with-border}
 
-The vulnerability is hidden for the particular image and component. If this
-vulnerability shows up in other images, it is still reported.
+The vulnerability is hidden system-wide and will no longer be reported as a vulnerability
+on other affected images.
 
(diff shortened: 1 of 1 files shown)


ee/dtr/user/manage-images/override-a-vulnerability.md
```diff
@@ -22,8 +22,8 @@ vulnerability, and click **hide**.
 
 ![Vulnerability list](../../images/override-vulnerability-2.png){: .with-border}
 
-The vulnerability is hidden for the particular image and component. If this
-vulnerability shows up in other images, it is still reported.
+The vulnerability is hidden system-wide and will no longer be reported as a vulnerability
+on other affected images.
```
ericsmalling: Suggest rewordling line 26 to be: `on other affected images with the same layer SHAs. `

Might also add text to illustrate an example of two images sharing a common base where the vulnerability exists vs two images with same CVE vulnerability but differing layers.    

Not sure what the clearest verbiage would be for that though.

bermudezmt: Any thoughts on the above, @davidswu ? I can reword line 26 and merge so this update gets out on the next site build. 

The additional text might require more thought - we can track it via #7457 .

caervs: Yeah, I agree we need to clarify that or else people might think that the override is at the component level.

PR #7461, opened 2018-10-04
Fixes #7396

### LINKED ISSUE #7396: I found a broken link
<Describe how you got there>
https://docker.github.io/reference/ > Engine CLI > 404

moreal:
### Proposed changes

In Issue #7396, 404 Error was found. So I fixed the reason from docker/docker.github.io/reference.md

### Related issues (optional)

Fixes #7396

### DIFF
--- a/reference.md
+++ b/reference.md
@@ -19,7 +19,7 @@ various APIs, CLIs, and file formats.
 
 | CLI                                                           | Description                                                                                                     |
 |:--------------------------------------------------------------|:----------------------------------------------------------------------------------------------------------------|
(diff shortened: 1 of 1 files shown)


bermudezmt:
@moreal The original link works in `docs.docker.com`: https://docs.docker.com/engine/reference/commandline/cli/

moreal:
@bermudezmt oh, I see. but in docker.github.io, it doesn't work. I think their hosts are different with each other. I am hard to understard why this structure was designed like this

bermudezmt:
@moreal the site just grew over time and you are correct, some of the files/images are hosted in different locations. I honestly can't tell you where yet but the priority now is to make sure all links and images work on docs.docker.com.

PR #7495, opened 2018-10-11
NFS volume with service create

omegamormegil:
Added example. There aren't many good examples using NFS with service create in the docs and this is a common customer question.

### DIFF
--- a/storage/volumes.md
+++ b/storage/volumes.md
@@ -462,7 +462,7 @@ $ docker service create -d \
 #### NFSv4
 ```bash
 docker service create -d \
-    --name nfs-service3 \
+    --name nfs-service \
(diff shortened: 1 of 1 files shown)


bermudezmt:
@adamancini can you please have a look? Thanks! :)

thaJeztah:
The syntax is CSV, so probably this should work (untested);

```
docker service create -d \
    --name nfs-service \
    --mount type=volume,source=nfsvolume,target=/app,volume-driver=local,volume-opt=type=nfs,volume-opt=device=:/,"volume-opt=o=10.0.0.10,rw,nfsvers=4,async" \
    nginx:latest

bermudezmt:
@thaJeztah This is what ended up working for me:
`docker service create -d \
    --name nfs-service3 \
    --mount 'type=volume,source=nfsvolume,target=/app,volume-driver=local,volume-opt=type=nfs,volume-opt=device=:/,"volume-opt=o=10.0.0.10,rw,nfsvers=4,async"' \
    nginx:latest`

@caervs Thank you for looking into this. And thanks, @omegamormegil, for the PR.

PR #7496, opened 2018-10-12
Don't spam browser history when scrolling thru page

### LINKED ISSUE #6299: Prevent page scrolling from spamming browser history
### Problem description

Scrolling past a section header on a typical docs page appears to automatically add it to the browser history.  This leads to a *lot* of irritating noise in the browser history.

For example, I visit https://docs.docker.com/engine/reference/commandline/build, and then scroll to the bottom.  This is what I see in my browser history:

screen shot 2018-03-25 at 13 45 08

This occurs in (at least) Chrome and Safari.

### Suggestions for a fix

`window.location.hash` is being manipulated in a lot of places in the JS.  I'm not sure which of these causes the issue, but I suspect it's the root of the problem.

allejo:
Been a while since I've looked through this repository, I see all new faces on the docs team.

Hi there!

### Proposed changes

As you scroll through the site, your browser history will get horrendously spammed with every single header; even if you just scroll past it.

This change consists of the following:

- Rewrite the `highlightRightNav()` function to be far simpler
- Do not modify browser history at all. Now, your browser history will only change when you actually **click** on a heading on the TOC (default browser behavior).
- Change the auto-TOC script to no longer include the page path (just the IDs are needed)

### Related issues (optional)

Fixes #6299


pmckeetx:
This looks good. Thanks, @allejo, for cleaning this up.

peter-docker:
Looks good. Thanks, @allejo, for cleaning this up.

peter-docker:
@ahh-docker feel free to merge whenever you're ready.

PR #7500, opened 2018-10-12
Renamed "target" example directory to "artifacts"

alkukampela:
### Proposed changes

Renamed example directory to "target" so it's different than keyword "target".


paigehargrave:

PR #7525, opened 2018-10-16
clear up confusion between compose file versions

BretFisher:
Really this type of info could be placed elsewhere and referenced in this note, so maybe suggest where it would live better. 

See comment https://github.com/moby/moby/issues/31101#issuecomment-428472115 for how we could better communicate the "best fit" file versions depending on someones use case. This is a top 3 issue of anyone I meet trying to use compose files with or without swarm. They are all confused thinking 2.x branch is not supported or maintained now that 3.x exists.

### DIFF
--- a/compose/extends.md
+++ b/compose/extends.md
@@ -159,8 +159,8 @@ backup, include the `docker-compose.admin.yml` as well.
 
 ## Extending services
 
-> **Note**: The `extends` keyword is supported in earlier Compose file formats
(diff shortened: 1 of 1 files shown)


bermudezmt:
@shin- sorry we're bombarding you with Compose PRs. Can you please have a look? Thanks!

bermudezmt:
Thanks, @shin- ! 
@garethr , can you please have a look? :)

PR #7530, opened 2018-10-17
remove docker-ce as thing to remove

andrewhsu:
### Proposed changes

Update instructions for installing docker-ee on centos and rhel by taking out the part that says to remove the `docker-ce` package because the `docker-ee` package will do this automatically.


bermudezmt:
Reviewed with @ahh-docker

PR #7549, opened 2018-10-21
Replace fd:// by unix::// in service override

rija:
### Proposed changes

On: https://docs.docker.com/install/linux/linux-postinstall/#configuring-remote-access-with-systemd-unit-file

The line below part of instructions to create a docker.service override for configuring connections:
    ```
    ExecStart=/usr/bin/dockerd -H fd:// -H tcp://127.0.0.1:2375
    ```
creates the error below preventing dockerd to start:

**"Failed to load listeners: no sockets found via socket activation: make sure the service was started"**

Replacing in the ExecStart line above ``fd://`` by ``unix://`` (the later being the syntax used in the docker.service default systemd unit) fixes the problem.

I thought it would help readers who copy/paste the code snippet to have it working out-of-the-box for this common use case and for the syntax to be consistent with that in ``/lib/systemd/system/docker.service``.

Environment: Centos 7 on AWS EC2 using Docker 18.06.1 from repository. No hosts configured in daemon.json.

### DIFF
--- a/install/linux/linux-postinstall.md
+++ b/install/linux/linux-postinstall.md
@@ -156,7 +156,7 @@ Configuring Docker to accept remote connections can be done with the `docker.ser
     ```none
     [Service]
     ExecStart=
-    ExecStart=/usr/bin/dockerd -H fd:// -H tcp://127.0.0.1:2375
+    ExecStart=/usr/bin/dockerd -H unix:// -H tcp://127.0.0.1:2375
     ```
 
 3. Save the file.


ahh-docker:
@seemethere PTAL. Thanks!

bermudezmt:
@rija Thank you for your feedback! We have received and reviewed your changes. Since this is a duplicate of #7982, I'm closing this and will continue the discussion and review in that PR.

PR #7552, opened 2018-10-22
ucp roles: fix punctuation

trapier:
```
git grep -l  'images, They' |xargs -n1 sed -i 's:images, They:images. They:'
```


bermudezmt:
@trapier thanks for making the changes for all versions.

PR #7583, opened 2018-10-29
Update services.md

orlvn:
Tasks, not services are scheduled on nodes.


bermudezmt:
@mark-church PTAL. Thanks!

PR #7585, opened 2018-10-29
Friendly script for distros based on Ubuntu

diegoleme:
### Proposed changes

Use `-u` to show upstream LSB data.

I use Elementary OS, it's a distro based on Ubuntu and with the command `lsb_release -cs` the output is `juno`.

If I use `lsb_release -u -cs` the output is `bionic`


bermudezmt:
@thaJeztah can you please have a look? Thanks!

paigehargrave:

PR #7600, opened 2018-10-31
update networking.md

tgogos:
### Proposed changes

The use of "expose" here is misleading and should be replaced with "publish".


ahh-docker:
@L-Hudson PTAL. thanks!

PR #7635, opened 2018-11-08
Manager min update to 8GB

darwin-traver:


ahh-docker:
LGTM with @bermudezmt

PR #7645, opened 2018-11-09
Update ce-ee-node-activate.md

sujaypillai:
Updated `docker version` info before and after activation. The engine server has the label "Community" [while using CE] and "Enterprise" after activation.

### Proposed changes

With the new stable release of CE there is no label -ce in the version info i.e. 18.09.0-ce has been renamed to 18.09.0. So updated the docker version output in relative sections according to the new release after my test on our server.


ahh-docker:
@bermudezmt also reviewed. Merging.

PR #7662, opened 2018-11-12
Fix grammer of markdown in swarm-mode.md

unasuke:
### Proposed changes
Fix (maybe) wrong markdown syntax in swarm-mode guide. The sentences are not any header contents. See also commonmark spec https://spec.commonmark.org/0.28/#setext-heading

https://docs.docker.com/engine/swarm/swarm-mode/

| before | after |
| --- | --- |
| screenshot from 2018-11-12 21-47-55 | screenshot from 2018-11-12 21-48-11 |

I don't remove `====` because of the reason of introduced that is unclear in commit 8f465b4a012d. But I think that it is good to remove `====`.


unasuke:
Resolved at #7776 . Thanks. :blush:

PR #7667, opened 2018-11-12
removed --emergency-repair flag

bkdocker:
reconfigure-db command needs to be run without the --emergency-repair flag,  because you should only use that when a table has lost a majority of replicas.
Please refer to https://github.com/docker/escalation/issues/943


bermudezmt:
@bkdocker @jlhawn I think it would still be important to note the `--emergency-repair` flag use case instead of just removing it. To get this PR merged faster, will we need to apply the update to all versions of UCP? I see only 2.2 has been updated.

jlhawn:
Yes, I believe the documentation should note that the `--emergency-repair` flag *is required* when there is loss of quorum in any of the RethinkDB tables. It's *probably* okay to always use that flag but it does have a slight risk that table config could get out of sync if all of the database servers aren't connected at the time you run the command which I why I recommended to only use it if you have to.

bermudezmt:
@jlhawn does b174e34cb027ab98517ffd35b4074fac34a98966 work? I'm going to add the same note to versions `3.0` and `3.1` of this page if so.

Issue #7666, opened 2018-11-12
Put a link to all commands with option --format


SimonHeimberg:
### Problem description

Commands having a option `--format` tell to use a `Go template`, but do not give additional info how to do this. A link would be useful.

### Problem location

https://docs.docker.com/engine/reference/commandline/inspect/ has many examples, this is kind of ok. But a link to the general explication would be nice here too.

Most commands like listed give no additional info:
* https://docs.docker.com/engine/reference/commandline/container_ls/
* https://docs.docker.com/engine/reference/commandline/container_inspect/
* https://docs.docker.com/engine/reference/commandline/image_ls/
* ...

### Suggestions for a fix

Put a link to https://docs.docker.com/config/formatting/ on every explication of command line option `--format`.

SimonHeimberg:
A search for `--format` (https://github.com/docker/docker.github.io/search?q=--format&type=Code) shows many places.

Some commands like [stats](https://github.com/docker/docker.github.io/blob/27b2befe44a9cafe23a5223385c7d3141c9f6425/_data/engine-cli/docker_stats.yaml#L62), [stack ls](https://github.com/docker/docker.github.io/blob/27b2befe44a9cafe23a5223385c7d3141c9f6425/_data/engine-cli/docker_stack_ls.yaml#L48) and [history](https://github.com/docker/docker.github.io/blob/27b2befe44a9cafe23a5223385c7d3141c9f6425/_data/engine-cli/docker_history.yaml#L73) have a list of valid names. Bot others like `container ls` have nothing.

craig-osterhout:
Thanks for sharing this issue. This particular content is maintained in an upstream repo. Since this issue is dated, I'm not sure if it is still applicable. If the issue is still applicable, please help create an issue or PR in the source repo at https://github.com/docker/cli/tree/master/docs/reference/commandline
Closing this as unable to fix in this repo.

PR #7670, opened 2018-11-13
Show hint how to find out what data can be printed

SimonHeimberg:
This hint is unnecessary for inspect commands since this is the default. But for other commands, I regularly find out again this trick.

### Proposed changes

Show hint how to find out what data can be printed.

### Related issues

Will be more helpful after #7666 is done.


SimonHeimberg:
Setting the title as H1 is probably wrong style. Improvements and suggestions welcome.

ahh-docker:
LGTM with @bermudezmt

PR #7687, opened 2018-11-14
Update Kube version for kubectl download

darwin-traver:
Did not modify windows kubectl version instructions, as the path for 1.11.2 did not work.


ahh-docker:
@david-yu PTAL to make sure version is accurate for specific versions, or do we want to use generic versions.

david-yu:
@ahh-docker I believe we are on 1.11.5 now due to the CVE vulnerability fix we recently pushed: https://docs.docker.com/ee/ucp/release-notes/

ahh-docker:
@david-yu if this looks good, let us know and we'll go ahead and merge.

PR #7689, opened 2018-11-15
UCP Upgrade Offline Bundles

ollypom:
### Proposed changes
The UCP Offline Bundles for Upgrade was still on UCP 3.0, i've updated this to UCP 3.1 bundles.


L-Hudson:
good to go

PR #7699, opened 2018-11-15
Added `demo_` prefix to interlock label

stshow:
Customer feedback received for the demo stack example: 

```
It leads users to expect that if they define a network in their compose file that they can use that network name in the Interlock network label which is not the case.  The example "works" because the service is only attached to a single network, but in actuality the Interlock network label would need to be prefixed with the stack name in order for the example to work correctly.
```
The concern is valid; so I have added `demo_` as a prefix to the `com.docker.lb.network` label.

### DIFF
--- a/ee/ucp/interlock/usage/index.md
+++ b/ee/ucp/interlock/usage/index.md
@@ -10,7 +10,7 @@ start using it in your swarm services.
 In this example we'll deploy a simple service which:
 
 * Has a JSON endpoint that returns the ID of the task serving the request.
-* Has a web UI that shows how many tasks the service is running.
+* Has a web interface that shows how many tasks the service is running.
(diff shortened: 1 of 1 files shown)


ee/ucp/interlock/usage/index.md
```diff
@@ -46,7 +46,7 @@ the host header, that request is forwarded to the demo service.
 should attach to in order to be able to communicate with the demo service.
 To use layer 7 routing, your services need to be attached to at least one network.
 If your service is only attached to a single network, you don't need to add
-a label to specify which network to use for routing.
+a label to specify which network to use for routing. To attach to more than one network, prefix this label with the service name as shown in the above example.
```
bermudezmt: Does this look right, @stshow ?

PR #7710, opened 2018-11-16
Update docker sdk go example

alejandrox1:
Signed-off-by: Jorge Alarcon Ochoa <alarcj137@gmail.com>

### Proposed changes

The example doesn't compile due to the use of `github.com/docker/docker/pkg/stdcopy` instead of `io.Copy`.
Also, replaced `newEnvClient`, which is deprecated, with `NewClientWithOpts`.

### DIFF
--- a/develop/sdk/index.md
+++ b/develop/sdk/index.md
@@ -177,23 +177,22 @@ Docker API directly, or using the Python or Go SDK.
 package main
 
 import (
-    "io"
+    "context"
     "os"
 
     "github.com/docker/docker/client"
     "github.com/docker/docker/api/types"
     "github.com/docker/docker/api/types/container"
     "github.com/docker/docker/pkg/stdcopy"
-
-    "golang.org/x/net/context"
 )
 
 func main() {
     ctx := context.Background()
-    cli, err := client.NewEnvClient()
+    cli, err := client.NewClientWithOpts(client.FromEnv)
     if err != nil {
         panic(err)
     }
+    cli.NegotiateAPIVersion(ctx)
 
     _, err = cli.ImagePull(ctx, "docker.io/library/alpine", types.ImagePullOptions{})
     if err != nil {


bermudezmt:
@tiborvass can you please have a look?  :)

develop/sdk/index.md
```diff
@@ -177,20 +177,18 @@ Docker API directly, or using the Python or Go SDK.
...
     "github.com/docker/docker/api/types/container"
     "github.com/docker/docker/pkg/stdcopy"
-
-    "golang.org/x/net/context"
 )
 
 func main() {
     ctx := context.Background()
-    cli, err := client.NewEnvClient()
+    cli, err := client.NewClientWithOpts(client.FromEnv, client.WithVersion("1.38"))
```
tiborvass: @alejandrox1 A reason you would want to specify the API version is for better stability, in the unlikely case we break older APIs (which would be a bug). There's a risk people would blindly copy and paste version 1.38 and that we are not going to keep this up-to-date with every API version bump. For that reason, I think it would be better to drop the `client.WithVersion("1.38")` part. cc @thaJeztah WDYT ?

thaJeztah: There's two sides to this;

Yes, I agree that making that the default example is tricky (people _do_ copy blindly, so that may be an issue)

There are situations where it may be desirable to pin to a specific API version (as the API may change, so if other parts of your code depend on the behavior of a specific version). So, we could _mention_ this as an option (but in a separate example / part of these docs).

The docker CLI takes API changes into account, and

- performs API-version negotiation (picking the latest API version that's supported by the daemon)
- disables/enables certain features, depending on the version of the API that's used


I think we should also show/tell about the use of `NegotiateAPIVersion` https://github.com/moby/moby/blob/bb1914b19572524b9f7d2b3415f146c545c1bb8b/client/client.go#L342-L370

tiborvass: @thaJeztah how about we keep it simple for this one?

bermudezmt: @tiborvass is the updated code sample basically correct but we want to drop `client.WithVersion`? I'm all for keeping it simple at this point.

thaJeztah: `NewEnvClient()` is deprecated, so perhaps we should change to `client.NewClientWithOpts(client.FromEnv)`

And add `NegotiateAPIVersion()`;

```go
func main() {
	ctx := context.Background()
	cli, err := client.NewClientWithOpts(client.FromEnv)
	if err != nil {
		panic(err)
	}
	cli.NegotiateAPIVersion(ctx)

	_, err = cli.ImagePull(ctx, "docker.io/library/alpine", types.ImagePullOptions{})
	if err != nil {
		panic(err)
	}

	resp, err := cli.ContainerCreate(ctx, &container.Config{
		Image: "alpine",
		Cmd:   []string{"echo", "hello world"},
	}, nil, nil, "")
	if err != nil {
		panic(err)
	}

	if err := cli.ContainerStart(ctx, resp.ID, types.ContainerStartOptions{}); err != nil {
		panic(err)
	}

	statusCh, errCh := cli.ContainerWait(ctx, resp.ID, container.WaitConditionNotRunning)
	select {
	case err := <-errCh:
		if err != nil {
			panic(err)
		}
	case <-statusCh:
	}

	out, err := cli.ContainerLogs(ctx, resp.ID, types.ContainerLogsOptions{ShowStdout: true})
	if err != nil {
		panic(err)
	}

	if _, err := stdcopy.StdCopy(os.Stdout, os.Stderr, out); err != nil {
		panic(err)
	}
}
```

> And I can imagine various versions for sticking to a specific Docker version (i.e., kubernetes/kubeadm only supporting 18.06).

Not sure that's the API version that's the issue here; I think at some point (e.g.) Docker 17.03 was listed as supported, because the validation suite had not been run against more recent versions. With the code above, using the client against an older version will automatically negotiate the supported API version

alejandrox1:
Updated the code snippet following @thaJeztah 's comments.

PR #7728, opened 2018-11-20
Update example for deprecated name

siyu6974:
secret.external.name is deprecated in favor of secret.name

### Proposed changes

update doc to use the new format

### DIFF
--- a/compose/compose-file/index.md
+++ b/compose/compose-file/index.md
@@ -2352,6 +2352,11 @@ secrets:
   my_second_secret:
     external: true
     name: redis_secret
+    
+  // Deprecated syntax for compose file version 3.5-
+  my_second_secret:
+    external:
+     name: redis_secret
 ```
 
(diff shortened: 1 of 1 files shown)


shin-:
This change will make the example non-functional for people using a version of the Compose file [under 3.5](https://docs.docker.com/compose/compose-file/compose-versioning/#version-35) when the new syntax was introduced. We definitely should recommend using it, but it would still be good to at least mention the deprecated form IMO.

siyu6974:
@shin- I added back old example and mark it as deprecated.

compose/compose-file/index.md
```diff
@@ -2345,10 +2345,18 @@ is different from the name that exists within the service. The following
 example modifies the previous one to use the external secret called
 `redis_secret`.
 
+### Compose File v3.5 and above
```
bermudezmt: @shin- can you confirm this looks good? Thanks for contributing, @siyu6974 !

PR #7733, opened 2018-11-21
Add notes that log-opts in  daemon.json must be a string

thaJeztah:
relates to https://github.com/moby/moby/issues/38242

log-opts are passed to logging-drivers as-is, so the daemon is not
aware what value-type each option takes.

For this reason, all options must be provided as a string, even if
they are used as numeric values by the logging driver.

For example, to pass the "max-file" option to the default (json-file)
logging driver, this value has to be passed as a string;

```json
{
  "log-driver": "json-file",
  "log-opts": {
    "max-size": "10m",
    "max-file": "3"
  }
}
```

When passed as a _number_ (`"max-file": 3`), the daemon will invalidate
the configuration file, and fail to start;

    unable to configure the Docker daemon with file /etc/docker/daemon.json: json: cannot unmarshal number into Go value of type string

This patch adds an example to the daemon.json to show these  values
have to be passed as strings.

Also included some small changes in additional commits; description for those changes can be found in their respective commit messages


thaJeztah:
ping @ahh-docker ptal

L-Hudson:
LGTM

PR #7735, opened 2018-11-21
change dash to bash.

crgaurav:
in this file, I have seen dash instead of bash.

### DIFF
--- a/develop/develop-images/dockerfile_best-practices.md
+++ b/develop/develop-images/dockerfile_best-practices.md
@@ -463,7 +463,7 @@ RUN set -o pipefail && wget -O - https://some.site | wc -l > /number
 ```
 > Not all shells support the `-o pipefail` option.
 >
-> In such cases (such as the `b ash` shell, which is the default shell on
+> In such cases (such as the `bash` shell, which is the default shell on
 > Debian-based images), consider using the _exec_ form of `RUN` to explicitly
 > choose a shell that does support the `pipefail` option. For example:
 >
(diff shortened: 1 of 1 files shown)


develop/develop-images/dockerfile_best-practices.md
```diff
@@ -463,7 +463,7 @@ RUN set -o pipefail && wget -O - https://some.site | wc -l > /number
 ```
 > Not all shells support the `-o pipefail` option.
 >
-> In such cases (such as the `dash` shell, which is the default shell on
+> In such cases (such as the `bash` shell, which is the default shell on
```
thaJeztah: `dash` is actually correct; https://lwn.net/Articles/343924/

bermudezmt: Oops, will revert.

thaJeztah: Not sure it actually became the default though (so perhaps that part should be removed).

And, yes, the `dash` name is confusing (and not very commonly known)

thaJeztah: The default (at least for the docker images) is "bash", but `dash` is definitely in there;

```bash
docker inspect --format '{{.Config.Cmd}}' debian:stretch
[bash]

docker run --rm debian:stretch sh -c 'which bash; which dash'
```

bermudezmt: Can you please review https://github.com/docker/docker.github.io/pull/7770? Thanks!

PR #7739, opened 2018-11-22
fix json

bnfinet:
The example json response does not parse correctly

Improper escaping of the `repository.dockerfile` "[/var/cache/apt-cacher-ng]" substring causes the parser to fail.


L-Hudson:
LGTM

PR #7741, opened 2018-11-22
Fix typo in ubuntu.md

rasa:


L-Hudson:
@ahh-docker, would you take a look and verify this for me? Thanks.

install/linux/docker-ce/ubuntu.md
```diff
@@ -204,7 +204,7 @@ the repository.
 
     ```bash
     $ sudo add-apt-repository \
-       "deb [arch=ppc64el] {{ download-url-base }} \
+       "deb [arch=ppc64le] {{ download-url-base }} \
```
thaJeztah: Afaik, Debian and Ubuntu use [`ppc64el`](https://wiki.ubuntu.com/ppc64el) for historic reasons. More details about this can be found on the Debian mailing-list; https://lists.debian.org/debian-powerpc/2014/08/msg00042.html

> On 08/28/2014 02:40 PM, Lior Kaplan wrote:
> > [...] why they chose to swap
> > the le (little endian) to el.[...]
> 
> It's an existing tradition, AFAICT:
> 
> Looking at the ppc64el patch for dpkg, the maintainer's response is [1]:
> 
> """
> I see this being supported already by config.{guess, sub}, good :), and
> *the arch name matches existing conventions*, so I'll merge the patch
> [...]
> """
> 
> So, looking at dpkg's cputable [2] 'Debian name' column, the 'el/eb'
> suffix for endianness is indeed, clearly, the convention.
> 
> Interestingly, on the 'GNU name' column, it looks like that is the
> convention too, being 'le' different than the others ('el/eb').
> (although I'm not familiar w/ most other existing GNU names/aliases,
>  nor the reasons/how they're chosen)

The repositories are therefore named `ppc64el` instead of `ppc64le`; https://download.docker.com/linux/ubuntu/dists/xenial/pool/stable/ppc64el/

PR #7749, opened 2018-11-26
Add network-routing clarification to toolbox/for-mac comparison

TCAllen07:
### Proposed changes

Adds a single line mentioning what I think is a very important distinction to make between docker-toolbox and docker-for-mac, namely that the latter doesn't support routing traffic to containers. As a result, running containers are not exposed how one might expect. Copied part of the line from [docker-for-mac's own docs](https://docs.docker.com/docker-for-mac/networking/#i-cannot-ping-my-containers).


L-Hudson:
LGTM

PR #7763, opened 2018-11-28
corrected spacing for the --unmanaged-cni option

nicolaka:


L-Hudson:
looks good -- approved

PR #7768, opened 2018-11-29
Extra `.0` should be removed for Centos and RHEL `sudo yum-config-manager --enable docker-ee-stable-18.09`

ekitagawa:
false: sudo yum-config-manager --enable docker-ee-stable-18.09.0
correct: sudo yum-config-manager --enable docker-ee-stable-18.09
 
I bumped into this while setting up a test env on Centos.  Did a quick test on RHEL, and I could see the same issue.


L-Hudson:
reviewed -- looks good!

PR #7783, opened 2018-12-03
Remove .md suffix from URLs

link2xt:
Without this change links on https://docs.docker.com/install/ don't work


link2xt:
Checked https://deploy-preview-7783--docsdocker.netlify.com/install/, it looks correct now

PR #7799, opened 2018-12-06
Update index.md to correctly reflect aarch64 OS

JeffUnderhill:
Per the script @ https://get.docker.com/ aarch64 is supported on Ubuntu, Debian, fedora & CentOS but the table on this page didn't correctly represent aarch64 fedora support. This change adds that check mark to the fedora row in the aarch64 column.


L-Hudson:
editorial review. GTG.

PR #7801, opened 2018-12-06
Update https.cnf - Clarify extfile directions

jonpspri:
Directions around `extfile.cnf` for client certificates was somewhat confusing.  Edited the document to clarify that a separate file was to be created, not append a line to the file that had just been created for the server.


L-Hudson:
LGTM

PR #7806, opened 2018-12-06
Added UCP 3.1.1 and 3.0.7 Offline Bundles

ollypom:
### Proposed changes

Added UCP 3.1.1 and 3.0.7 Offline Bundles


L-Hudson:
good to go

PR #7824, opened 2018-12-08
Fix XFS 'ftype' option typo

praseodym:
The required XFS option is called ftype, not fstype:

```
xfs_info / | grep ftype
naming   =version 2              bsize=4096   ascii-ci=0 ftype=1
```


L-Hudson:
GTG

PR #7847, opened 2018-12-12
Edit Wordpress quick start docs

Djiit:
### Proposed changes

Using tips from https://github.com/docker-library/mysql/issues/361#issuecomment-356744484, I edit the docs so one can copy/paste the example and *just* run it.

### Related issues (optional)

- https://github.com/docker-library/mysql/issues/361#issuecomment-356744484

### DIFF
--- a/compose/wordpress.md
+++ b/compose/wordpress.md
@@ -39,6 +39,7 @@ Compose to set up and run WordPress. Before starting, make sure you have
     services:
        db:
          image: mysql:5.7
+         command: --disable-partition-engine-check
          volumes:
            - db_data:/var/lib/mysql
          restart: always


compose/wordpress.md
```diff
@@ -39,6 +39,7 @@ Compose to set up and run WordPress. Before starting, make sure you have
     services:
        db:
          image: mysql:5.7
+         command: --disable-partition-engine-check
```
thaJeztah: Trying to grasp the issue that you linked to (and the documentation; https://dev.mysql.com/doc/mysql-partitioning-excerpt/5.7/en/partitioning.html

When starting with a fresh installation, wouldn't the database be initialized correctly? Or does the problem only occur if you have an existing database, and try to run it on a version of mysql > 5.7.20 ?

Looking at the example here, there's a few things that stand out;

- We do specify the version of MySQL to use (`mysql:5.7`, which means `5.7.x`)
- We _don't_ specify the version of WordPress to use (instead, using `wordpress:latest`)

Using `:latest` is definitely _not_ best practice (today it's `wordpress:5`, but tomorrow that could be `wordpress:6`).

I suggest we should;

- specify the version of wordpress to use
- use the `-alpine` version of WordPress (it's a lot smaller 😄)
- update the mysql version to a more current version (if possible); looks like WordPress lists `5.6+` as requirement https://wordpress.org/about/requirements/. If the current relese (MySQL 8) is supported, perhaps we can switch to that.


### Bonus points:

To show best practice;

- run the `mysql` image as non-root (https://hub.docker.com/_/mysql#running-as-an-arbitrary-user)
- looks like the wordpress image already defaults to run as non-root (https://hub.docker.com/_/wordpress#running-as-an-arbitrary-user), so perhaps we're all set there

thaJeztah:
ping @ltangvald  @tianon could you have a look at my comments, and help out on this one?

compose/wordpress.md
tianon: > `command: --disable-partition-engine-check`

I'm definitely out of my depth on the implications of what this option does for MySQL, but in general if it only applies to Docker for Windows users, I wonder if a comment would be more appropriate? (So that we use the MySQL-provided default settings elsewhere.)

> specify the version of `wordpress` to use

That's fair, although older versions of the Docker image aren't officially supported and thus won't be updated and the WordPress compatibility track record is pretty impeccable (not to mention their security update process -- they handle security very, very aggressively these days :sweat_smile:).

> use the `-alpine` version of WordPress (it's a lot smaller :smile:)

Don't let the shiny size distract from the fact that switching to Alpine means switching to FPM and thus requiring NGINX (or another reverse proxy that can do FastCGI) in front with non-trivial configuration.  If you're looking for a simple example, the Apache tags are definitely your best bet (since they'll pretty much just work right out of the box).

> update the mysql version to a more current version (if possible); looks like WordPress lists 5.6+ as requirement https://wordpress.org/about/requirements/. If the current relese (MySQL 8) is supported, perhaps we can switch to that.

There has been some issue with PHP + MySQL 8 due to authentication changes in MySQL 8, although I _think_ all of those are resolved.  IMO, MySQL 5.7 is still a great choice for a simple WordPress example (although @ltangvald might disagree, which I'd be very interested in :smile::heart:).

> run the `mysql` image as non-root

It only runs as root in order to be able to `chown` on startup -- it immediately then steps down from that, so I think if you're going for security, you'd get the same benefits without all the caveats by using `--security-opt no-new-privileges` to ensure that once it steps down from root, it can't get back (no matter what crafty setuid binaries exist).

> looks like the `wordpress` image already defaults to run as non-root (https://hub.docker.com/_/wordpress#running-as-an-arbitrary-user), so perhaps we're all set there

By default it works similar to what I've described for the `mysql` image above, so should be pretty sane out of the box.

usha-mandya:
Closing the PR based on the comments above

PR #7849, opened 2018-12-13
HIt to visit kitematic github page for kitematic

jackfruhecolab:
### Proposed changes

Added a sentence letting people know that they can still install kitematic via the GitHub repo.
This is important as of 12-2018, because the kitematic.com home page references this URL for installation of Kitematic. I will also try to find a way to update that page directly, I am trying to cover all bases.


L-Hudson:
LGTM. Thanks for the addition. Merging and closing.

PR #7850, opened 2018-12-13
`lsb_release -cs` might give 'n/a' which is not what we want

qudongfang:
### Proposed changes

`lsb_release -cs` might give 'n/a' which is not what we want.
It is a [bug](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=845651) in debian.

### DIFF
--- a/install/linux/docker-ce/debian.md
+++ b/install/linux/docker-ce/debian.md
@@ -159,7 +159,7 @@ from the repository.
     **test** repository, add the word `edge` or `test` (or both) after the
     word `stable` in the commands below.
 
-    > **Note**: The `lsb_release -cs` sub-command below returns the name of your
+    > **Note**: The `cut -f1 -d'/' /etc/debian_version` sub-command below returns the name of your
     > Debian distribution, such as `jessie`.
 
     To also add the **edge** repository, add `edge` after `stable` on the last
@@ -176,7 +176,7 @@ from the repository.
     ```bash
     $ sudo add-apt-repository \
        "deb [arch=amd64] {{ download-url-base }} \
-       $(lsb_release -cs) \
+       $(cut -f1 -d'/' /etc/debian_version) \
        stable"
     ```
 
@@ -185,7 +185,7 @@ from the repository.
 
     ```bash
     $ echo "deb [arch=armhf] {{ download-url-base }} \
-         $(lsb_release -cs) stable" | \
+         $(cut -f1 -d'/' /etc/debian_version) stable" | \
         sudo tee /etc/apt/sources.list.d/docker.list
     ```
 
@@ -194,7 +194,7 @@ from the repository.
 
     ```bash
     $ echo "deb [arch=arm64] {{ download-url-base }} \
-         $(lsb_release -cs) stable" | \
+         $(cut -f1 -d'/' /etc/debian_version) stable" | \
         sudo tee /etc/apt/sources.list.d/docker.list
     ```
 


L-Hudson:
@mataneja can you take a look at this please?

install/linux/docker-ce/debian.md
```diff
@@ -159,7 +159,7 @@ from the repository.
     **test** repository, add the word `edge` or `test` (or both) after the
     word `stable` in the commands below.
 
-    > **Note**: The `lsb_release -cs` sub-command below returns the name of your
+    > **Note**: The `cut -f1 -d'/' /etc/debian_version` sub-command below returns the name of your
```
thaJeztah: I'm a bit on the fence on this one; reading the [Debian ticket](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=845651);

> Debian Bug report logs - #845651
> lsb-release: `--codename returns n/a` on stretch without apt sources configured

If I understand correctly, the ticket is closed as a "won't fix", because it's the intended behavior if no apt sources are configured, or if no network-connection is available;

> I have thought about this issue for some time, and I think that the result is 
> actually correct. Let me explain:
> 
> When (as currently), stretch is the testing release, /etc/debian_version 
> contains "stretch/sid", as shipped by base-files. It is therefore impossible 
> to rely on that file to differentiate between a host running testing or 
> unstable without asking apt what is actually preferred when installing 
> packages (through parsing `apt-cache policy`). That's how `lsb-release --
> codename` returns "sid" _xor_ "stretch".
> 
> stretch's base-files is currently in version 9.7, and ships with "stretch/sid" 
> in /etc/debian_version. But if you look at base-files in the current stable 
> (8+deb8u6), its /etc/debian_version currently contains "8.6", and with that, 
> `lsb-release --codename` returns "jessie" consistently, and without relying on 
> apt.

Based on the above, using `/etc/debian_version` could therefore give the _wrong_ version (if I interpret correctly).

Looking at the issue at hand; the `n/a` would be returned if no apt sources are configured and/or no networking connection is available;
the installation instructions for [installing using the repository](https://docs.docker.com/install/linux/docker-ce/debian/#set-up-the-repository) however, start with;


> #### Set up the repository
> 
> 1.  Update the `apt` package index:
> 
> ```bash
> sudo apt-get update
> ```

Does the `lsb_release -cs` still fail at that point?


Looking at the content of `/etc/debian_version` - that file may not contain the codename, but the numeric _version_. For example;

```bash
docker run --rm debian:stretch sh -c 'cat /etc/debian_version'
9.6

docker run --rm debian:jessie sh -c 'cat /etc/debian_version'
8.11
```

So, trying the instructions (I'm using the instructions for `arm64` here, as `lsb_release` is not available in the container);

```bash
docker run --rm debian:stretch sh -c 'echo "deb [arch=amd64] https://download.docker.com/linux/debian $(cut -f1 -d"/" /etc/debian_version) stable" | tee /etc/apt/sources.list.d/docker.list'
deb [arch=amd64] https://download.docker.com/linux/debian 9.6 stable
```

Running the instructions will add this repository;

```
deb [arch=amd64] https://download.docker.com/linux/debian 9.6 stable
```

Which will fail;

```
docker run --rm debian:stretch sh -c 'echo "deb [arch=amd64] https://download.docker.com/linux/debian $(cut -f1 -d"/" /etc/debian_version) stable" | tee /etc/apt/sources.list.d/docker.list && apt-get update'
deb [arch=amd64] https://download.docker.com/linux/debian 9.6 stable
Ign:1 http://cdn-fastly.deb.debian.org/debian stretch InRelease
Get:2 http://cdn-fastly.deb.debian.org/debian stretch-updates InRelease [91.0 kB]
Get:3 http://cdn-fastly.deb.debian.org/debian stretch Release [118 kB]
Get:4 http://cdn-fastly.deb.debian.org/debian stretch Release.gpg [2434 B]
Get:5 http://security-cdn.debian.org/debian-security stretch/updates InRelease [94.3 kB]
Get:6 http://cdn-fastly.deb.debian.org/debian stretch-updates/main amd64 Packages [5152 B]
Get:7 http://cdn-fastly.deb.debian.org/debian stretch/main amd64 Packages [7089 kB]
Get:8 http://security-cdn.debian.org/debian-security stretch/updates/main amd64 Packages [464 kB]
Fetched 7864 kB in 1s (4951 kB/s)
Reading package lists...
E: The method driver /usr/lib/apt/methods/https could not be found.
E: Failed to fetch https://download.docker.com/linux/debian/dists/9.6/InRelease  
E: Some index files failed to download. They have been ignored, or old ones used instead.
```


@qudongfang can you confirm if `lsb_release -cs` works after `apt-get update` has been run?

bermudezmt:
@qudongfang can you please review comments from @thaJeztah ?

install/linux/docker-ce/debian.md
qudongfang: Thanks for the detailed review.

It seems that `lsb_release -cs` is still the most reliable way to get codename.
Let's stick to it.

I'll close this PR soon.

qudongfang:
> @qudongfang can you please review comments from @thaJeztah ?

Thanks for the detailed review.

It seems that `lsb_release -cs` is still the most reliable way to get codename.
Let's stick to it.

I'll close this PR soon.

thaJeztah:
Thanks @qudongfang !

PR #7789, opened 2018-12-04
Update port mapping 4000 to 80 for consistency

raunakkathuria:
### Proposed changes

As of now, docker-compose.yml file `ports` section is in https://docs.docker.com/get-started/part3/, is not consistent with terminal recording and also with what is specified in https://docs.docker.com/get-started/part5/

```
diff --git a/get-started/part3.md b/get-started/part3.md
index 9a1436f571..95c5132bbd 100644
--- a/get-started/part3.md
+++ b/get-started/part3.md
@@ -84,7 +84,7 @@ services:
       restart_policy:
         condition: on-failure
     ports:
-      - "4000:80"
+      - "80:80"
     networks:
       - webnet
 networks:
```
> Required changes

PR #7858, opened 2018-12-14
Update get-started/part3.md port number and image

uronly14me:
### Proposed changes
I found that the port in the code example and image are wrong. Just make it 4000 and use app-in-browser.png.
In previous PRs, they announced about only port number in example code. So the image and other words were not changed. I just make them consistent.

### Related issues
 #7789, #6790

### DIFF
--- a/get-started/part3.md
+++ b/get-started/part3.md
@@ -84,7 +84,7 @@ services:
       restart_policy:
         condition: on-failure
     ports:
-      - "80:80"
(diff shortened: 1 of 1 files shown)


thaJeztah:
Whoops; looks like this is a duplicate of https://github.com/docker/docker.github.io/pull/7862 (or the other way around 😊 )

PR #7862, opened 2018-12-14
Changed port to 80 instead of 4000

QuentinAdt:
### Proposed changes

The docker-compose.yml specifies that it runs on port 80

### DIFF
--- a/get-started/part3.md
+++ b/get-started/part3.md
@@ -158,7 +158,7 @@ is not filtered by service:
 docker container ls -q
 ```
 
-You can run `curl -4 http://localhost:4000` several times in a row, or go to that URL in
+You can run `curl -4 http://localhost` several times in a row, or go to that URL in
 your browser and hit refresh a few times.
 
 ![Hello World in browser](images/app80-in-browser.png)


get-started/part3.md
```diff
@@ -158,7 +158,7 @@ is not filtered by service:
 docker container ls -q
 ```
 
-You can run `curl -4 http://localhost:4000` several times in a row, or go to that URL in
+You can run `curl -4 http://localhost` several times in a row, or go to that URL in
```
thaJeztah: Hm.. I see this was changed in https://github.com/docker/docker.github.io/pull/7789

@L-Hudson I think we should revert the changes from https://github.com/docker/docker.github.io/pull/7789. Using port 4000 was done on purpose, because;

1. Port 80 has a higher risk of being in use already (either by Docker, or another service running on the host)
2. Using a different port to _publish_ the container's port than the port that the container exposes;
  - Makes it apparent that port-mapping can re-map container ports
  - Illustrates which port is the "host" port and which port the "container" port (the shorthand syntax can confuse users that haven't used docker before; when using `80:80` it's not apparent which `80` is which).
  - Finally (but less important), port `80` is in the privileged port range and may be blocked  by company policy

thaJeztah:
Alternative PR: https://github.com/docker/docker.github.io/pull/7865

PR #7864, opened 2018-12-14
Numbers were confused

gobold1979:
In the tutorial, you connect to alpine1 (ONE!), do some stuff, can't ping alpine2 and I guess the author was confused at this point because immediately after not being able to ping alpine2 comes detaching from the one you're connected to and that's alpine1.


L-Hudson:
Great catch, thank you!

PR #7867, opened 2018-12-17
use port 4000 instead of 80 on host

ryannhg:
### Proposed changes

I changed the `docker-compose.yml` example to map port `4000` to the host, as described later in the tutorial:

> "Map port 4000 on the host to web’s port 80."

Thanks for the great tutorial!
Ryan


L-Hudson:
LGTM

Issue #3903, opened 2017-07-17
Docker Restart Options


MarkDucommun:
[compose/reference/restart.md](https://docs.docker.com/compose/reference/restart/), CC @londoncalling 

What are the options?  It feels like this is the place where they would be enumerated.

londoncalling:
Hi @MarkDucommun , There is only one flag with a short and a long form (`-t` or `--timeout`), as documented in [compose/reference/restart.md](https://docs.docker.com/compose/reference/restart/)

However, in looking at the current Help on the command, I'm wondering whether `--timeout` has changed to `--time`? 

```
victoriabialas at snapair in ~/repos/docker.github.io on master
$ docker restart --help

Usage:	docker restart [OPTIONS] CONTAINER [CONTAINER...]

Restart one or more containers

Options:
      --help       Print usage
  -t, --time int   Seconds to wait for stop before killing the container (default 10)
```

Can you verify, @shin- and/or @dnephin ? The `--help` on the command ^^ doesn't quite seem to match the docs. (I'm running `17.06.0-ce` and `docker-compose version 1.14.0, build c7bdf9e`.)

cc: @mstanleyjones

dnephin:
@londoncalling you ran `docker restart` not `docker-compose restart`. The docs are correct.

londoncalling:
@dnephin , right you are!

```
$ docker-compose restart --help
Restart running containers.

Usage: restart [options] [SERVICE...]

Options:
  -t, --timeout TIMEOUT      Specify a shutdown timeout in seconds.
                             (default: 10)
```
@MarkDucommun  closing this issue, as the docs already include the options you were looking for.

amelvill:
Just wanted to throw this out there,

I notice that the [docker-compose restart](https://docs.docker.com/compose/reference/restart/) command docs page has a rather low score at the time I'm writing this, (+36/-207) and I can't leave comments on the page directly, so I figured I'd just leave an idea here,

It's not really Docker's fault on this, but when you google something like "docker-compose restart", the first page that comes up is this: [docker-compose restart](https://docs.docker.com/compose/reference/restart/), when I searched for that I was looking for the version 2 docker-compose.yml config file field [restart](https://docs.docker.com/compose/compose-file/compose-file-v2/#restart). If I would've searched for "docker-compose.yml restart" it would've came up with the right page, but I didn't know docker-compose (the CLI utility) had a restart command at the time.

[...]

PR #7875, opened 2018-12-17
Related pages links for docs per suggestion in PR#3903

londoncalling:
Signed-off-by: Victoria Bialas <vicky@thoughtspot.com>

### Proposed changes

* Added some cross-references to related Compose docs per most recent customer comments from @amelvill in PR #3903 on `restart` options (thought this might help)

### Related issues (optional)

* PR #3903

cc: @joaofnfernandes @shin-

### DIFF
--- a/compose/reference/restart.md
+++ b/compose/reference/restart.md
@@ -17,3 +17,8 @@ Restarts all stopped and running services.
 If you make changes to your `docker-compose.yml` configuration these changes are not reflected after running this command.
 
 For example, changes to environment variables (which are added after a container is built, but before the container's command is executed) are not updated after restarting.
+
+See also, [restart](/compose/compose-file/index.md#restart) in Compose file v3 and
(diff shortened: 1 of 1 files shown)


shin-:
@londoncalling 👋

I'd suggest introducing this paragraph with something like "if you're looking to configure a service's restart policy, please refer to [...]" instead, so people know what they're being referred to before clicking ; especially because the command and the policy are separate concepts that don't interact with one another.

londoncalling:
Hi Joffrey ,
That makes sense. I’ll make that change. I also think there are a few more
pages where I could add a similar x-ref. I’ll find those too.

Vicky

londoncalling:
@shin- sorry this took so long .. let me know what you think. Happy Holidays, hope you are well 🎁🎄

bermudezmt:
Thanks, @londoncalling and @shin- . Happy new year to both of you!

PR #7877, opened 2018-12-17
Added Azure VNet Peering configuration

millerdz:
### Proposed changes

Using Azure VNet Peering allows connectivity between Azure Subscriptions and Azure Subnets when deploying UCP manager nodes to one Azure Subscription and Azure Subnet, and Windows-only worker nodes to another Azure Subscription and Azure Subnet. Known limitations also added to the documentation.


ollypom:
@millerdz Thanks a lot for submitting more Azure Docs :) 

However right now there is an internal investigation going to understand and test what `UCP manager nodes and [DTR](/ee/dtr) nodes can be deployed in a separate Azure Subscription on separate Azure Subnet than the worker nodes` means within Docker EE. My 2 cents would be to wait for that investigation to conclude before reviewing this PR. 

@mark-church WDYT?

millerdz:
@mklamb said this should move over to kbase. Will work with him on that.

millerdz:
Cancelling PR

PR #7878, opened 2018-12-18
match instructions using port 4000

clongfield:
The tutorial uses the external network 4000 but the docker compose maps to port 80. The compose file does not work for the given tutorial with the port mapping of 80:80


L-Hudson:
Change has already been made through other tickets; thanks for the catch though!

PR #7886, opened 2018-12-19
Update part3.md

hrzn:
### Proposed changes

Made the docker compose file consistent with the explanation lower in the text: 
"Map port 4000 on the host to `web`'s port 80."


L-Hudson:
corrected with a different PR; thanks!

PR #7891, opened 2018-12-19
Update Offline Zip for Windows 18.09

ollypom:
### Proposed changes

I noticed the ee_engine_version was incorrectly changed https://github.com/docker/docker.github.io/commit/18b804bbd6c47aa737880b0813ce599cb4745817 to a platform release number not an engine release number. (i.e. 17.06 to 2.1, not 17.06 to 18.09)

This seems to have broken a few outputs and offline installation instructions. I have changed this value to the latest engine release number which should fixed things. I did a quick look through and I couldn't find any variable that was dependent on this value being a platform release number.  

*Examples of issue:*

https://docs.docker.com/install/windows/docker-ee/#use-a-script-to-install-docker-ee

```
# On an online machine, download the zip file.
invoke-webrequest -UseBasicparsing -Outfile docker-17.06.2-ee-17.zip https://download.docker.com/components/engine/windows-server/2.1/docker-17.06.2-ee-17.zip
``` 

Should be: 

```
# On an online machine, download the zip file.
invoke-webrequest -UseBasicparsing -Outfile docker-17.06.2-ee-17.zip https://download.docker.com/components/engine/windows-server/17.06/docker-17.06.2-ee-17.zip
```

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -22,7 +22,7 @@ exclude: ["_scripts", "apidocs/layouts", "Gemfile", "hooks"]
 latest_stable_docker_engine_api_version: "1.39"
 docker_ce_stable_version: "18.09"
 docker_ce_edge_version: "18.09"
-docker_ee_version: "2.1"
+docker_ee_version: "18.09"
 compose_version: "1.23.1"
 machine_version: "0.16.0"
 distribution_version: "2.6"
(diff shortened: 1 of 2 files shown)


_config.yml
```diff
@@ -95,7 +95,7 @@ defaults:
   - scope:
       path: "install"
     values:
-      win_latest_build: "docker-17.06.2-ee-17"
+      win_latest_build: "docker-18.09.0"
```
bermudezmt: Looks like we've released `18.09.1` according to https://download.docker.com/components/engine/windows-server/index.json. Should we update to that version?

ollypom: The only 18.09.1 package is a release candidate "18.09.1-rc1", there is not a stable GA release yet of 18.09.1 :)

PR #7894, opened 2018-12-19
removed trusty from docs to match docker ce build

michaelnugent:
### Proposed changes

Removed Ubuntu Trusty (14.04 LTS) from docs since the build was removed from the docker ce code as well:
https://github.com/docker/docker-ce/commit/d2178be92a56cb45b25ebe9275340ef9cd54c8a6#diff-9be7b725da4b31d018cc9310c9d57ec9

This removes the need for the tab as well so I removed that structure.  I also bumped the example that mentions Trusty forward to Xenial (16.04 LTS)

This applies builds of docker-ce after 18.09.0.

### Related issues (optional)
Link above.

### DIFF
--- a/install/linux/docker-ce/ubuntu.md
+++ b/install/linux/docker-ce/ubuntu.md
@@ -37,8 +37,6 @@ versions:
 
 Docker CE is supported on Ubuntu on `x86_64`, `armhf`, `s390x` (IBM Z), and `ppc64le` (IBM Power) architectures.
 
-> **`ppc64le` and `s390x` limitations**: Packages for IBM Z and Power architectures are only available on Ubuntu Xenial and above.
-
 ### Uninstall old versions
 
 Older versions of Docker were called `docker` or `docker-engine`. If these are
--- a/install/linux/docker-ce/ubuntu.md
+++ b/install/linux/docker-ce/ubuntu.md
@@ -34,7 +34,6 @@ versions:
 
 - Bionic 18.04 (LTS)
 - Xenial 16.04 (LTS)
-- Trusty 14.04 (LTS)
 
 Docker CE is supported on Ubuntu on `x86_64`, `armhf`, `s390x` (IBM Z), and `ppc64le` (IBM Power) architectures.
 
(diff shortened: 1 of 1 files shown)


install/linux/docker-ce/ubuntu.md
```diff
@@ -170,7 +144,7 @@ the repository.
     > Ubuntu distribution, such as `xenial`. Sometimes, in a distribution
     > like Linux Mint, you might need to change `$(lsb_release -cs)`
     > to your parent Ubuntu distribution. For example, if you are using
-    >  `Linux Mint Rafaela`, you could use `trusty`.
+    >  `Linux Mint Sarah`, you could use `xenial`.
```
michaelnugent: Sarah -> Xenial:
https://linuxmint.com/download_all.php

install/linux/docker-ce/ubuntu.md
```diff
@@ -34,7 +34,6 @@ versions:
 
 - Bionic 18.04 (LTS)
 - Xenial 16.04 (LTS)
-- Trusty 14.04 (LTS)
 
 Docker CE is supported on Ubuntu on `x86_64`, `armhf`, `s390x` (IBM Z), and `ppc64le` (IBM Power) architectures.
```
thaJeztah: Looks like the line below (line 41) can also be removed;

> **`ppc64le` and `s390x` limitations**: Packages for IBM Z and Power architectures are only available on Ubuntu Xenial and above.

michaelnugent:
The -ee version of this may need to be changed as well and, if anyone wants, I can include that, but I don't know if -ee has the same support structure.

install/linux/docker-ce/ubuntu.md
```diff
@@ -68,36 +67,11 @@ outlined below.
```
thaJeztah: In the section above, we can probably change the bullet list (because all supported Ubuntu versions are now on kernel 4 or above;

```Markdown
 - For new installations on version 4 and higher of the Linux kernel, `overlay2`
   is supported and preferred over `aufs`.
 - For version 3 of the Linux kernel, `aufs` is supported because `overlay` or
   `overlay2` drivers are not supported by that kernel version.
```

Perhaps combine the first bullet with the first sentence; something like;

```Markdown
Docker CE on Ubuntu supports `overlay2` and `aufs` storage drivers. For new
installations, `overlay2` is the default, and preferred over `aufs`. If you
need to use `aufs`, you need to do additional preparation as outlined below.
```

thaJeztah:
@ahh-docker PTAL

thaJeztah:
> The -ee version of this may need to be changed as well and, if anyone wants, I can include that, but I don't know if -ee has the same support structure.

For EE, we probably need to keep Trusty; Docker EE provides LTS versions of the Docker Engine, where versions 17.06, 18.03, and 18.09 are currently supported (and 17.06 and 18.03 still support being installed on Ubuntu 14.04 - see https://success.docker.com/article/compatibility-matrix)

install/linux/docker-ce/ubuntu.md
ahh-docker: Done.

ahh-docker:
@thaJeztah PTAL. thanks!

thaJeztah:
I see I did this already in 
https://github.com/docker/docker.github.io/commit/5047fdee863cb65c5da45a49897868154ec07403, so looks like we no longer need this one

ahh-docker:

PR #7904, opened 2018-12-20
Update release-notes.md

adamancini:
### Proposed changes

added callout for rethinkcli version tag - if you don't specify the tag you'll get `:latest` which is over 2 years old.


adamancini:
@bermudezmt the `rethinkcli` image does not match up 1:1 with our rethinkdb image we use in DTR.  https://hub.docker.com/r/dockerhubenterprise/rethinkcli only has a couple supported tags, `2.3.0` is the latest one. 

I think it's safe to hardcode that image version.

bermudezmt:
@caervs @adamancini thank you both! Always learning something new from both of you. 

Ryan, I will see if that tip is already lurking somewhere and if not, I'll get that documented.

PR #7912, opened 2018-12-21
Add missing "t"

muffl0n:
com.docker.lb.defaul_backend -> com.docker.lb.default_backend


L-Hudson:
great catch, thank you!

PR #7922, opened 2018-12-23
Remove reference to pre-1.12 standalone swarm

thaJeztah:
Docker 1.12 and under has reached EOL a long time ago, and should not be used (as they have unpatched vulnerabilities).

Standalone swarm is now in maintenance mode, so we can probably remove this reference.

ping @ahh-docker @L-Hudson PTAL


ahh-docker:
LGTM

PR #7950, opened 2018-12-31
Docker Machine - Installation steps broken

spartha80:
https://github.com/docker/machine/downloads/ is no longer available. The path mentioned in the documentation is to be replaced with https://github.com/docker/machine/releases/tags

### Proposed changes
The path to the DockerMachine installation files on Github has been updated. The old path is broken and no longer working.

### DIFF
--- a/machine/install-machine.md
+++ b/machine/install-machine.md
@@ -19,12 +19,9 @@ on GitHub.
 
 1.  Install [Docker](/engine/installation/index.md){: target="_blank" class="_" }.
 
-2.  
+2.  Download the Docker Machine binary and extract it to your PATH.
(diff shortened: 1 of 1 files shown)


machine/install-machine.md
```diff
@@ -19,28 +19,31 @@ on GitHub.
 
 1.  Install [Docker](/engine/installation/index.md){: target="_blank" class="_" }.
 
-2.  Download the Docker Machine binary and extract it to your PATH.
+2.  
+
+
+the Docker Machine binary and extract it to your PATH.
```
shin-: This part needs to be restored

PR #7958, opened 2019-01-01
Distinguish command from placeholder

vicklau:
To avoid ambiguity where the command (the plain word) is the same as the placeholder (now in <>) such as `image` and `tag` and no actual examples are given to clarify.

### DIFF
--- a/get-started/part2.md
+++ b/get-started/part2.md
@@ -321,7 +321,7 @@ username, repository, and tag names so that the image uploads to your
 desired destination. The syntax of the command is:
 
 ```shell
-docker tag image username/repository:tag
+docker tag <image id> <username>/<repository id>:<tag name>
 ```
 
 For example:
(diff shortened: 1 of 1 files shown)


get-started/part2.md
```diff
@@ -321,7 +321,7 @@ username, repository, and tag names so that the image uploads to your
 desired destination. The syntax of the command is:
 
 ```shell
-docker tag image username/repository:tag
+docker tag <image id> <username>/<repository id>:<tag name>
```
thaJeztah: I'm a bit unsure about these; `image id` may actually be confusing, because it's not just the `IMAGE ID` that can be used; it can be any reference to an existing image (which could be the image's ID, or `image:tag`, or `image@sha256:digest`). Same for `<repository id>` (as it's not an _id_).

I realise these things are difficult to get right; even more so because Docker Hub uses `username/repository`, but "username" is only by convention on Docker Hub to use "username" as the first part (it's not part of the distribution spec; e.g. `foo/bar/baz:tag` would be acceptable on some registries.

Perhaps using some more obvious names would help; e.g.

```
docker tag mylocalimage username/someimage:sometag
```

But not sure if that makes it more clear; also, some actual example is already shown below, so... still on the fence if the proposed changes make this more clear

L-Hudson:
Going to go ahead and close this for ambiguity. It's more a language call than an error per se.

PR #7965, opened 2019-01-02
Update rails.md - incorporate feedback from 6254

paigehargrave:
### Related issues (optional)

Relates to #6254

### DIFF
--- a/compose/rails.md
+++ b/compose/rails.md
@@ -9,7 +9,7 @@ a Rails/PostgreSQL app. Before starting, [install Compose](install.md).
 
 ### Define the project
 
-Start by setting up the four files needed to build the app. App will run inside a Docker container containing its dependencies. Defining dependencies is done using a file called `Dockerfile`. To begin with, the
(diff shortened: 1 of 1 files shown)


compose/rails.md
```diff
@@ -31,7 +31,7 @@ Next, create a bootstrap `Gemfile` which just loads Rails. It'll be overwritten
 in a moment by `rails new`.
 
     source 'https://rubygems.org'
-    gem 'rails', '5.2.0'
+    gem 'rails', '~>5'
```
bermudezmt: @shin- will this work or should it be `5.0.x`? I forget whether Ruby is strict with versioning.

paigehargrave: @shin- Incorporated your feedback from #6254 - PTAL. Thanks!

shin-: Not a ruby expert, but:

> The specifier `~>` has a special meaning, best shown by example. `~> 2.0.3` is identical to `>= 2.0.3` and `< 2.1`. `~> 2.1` is identical to `>= 2.1` and `< 3.0`. 

[[source]](https://bundler.io/v1.5/gemfile.html)

So `~>5` means `>= 5.0.0, < 6` which matches what we want.

PR #7975, opened 2019-01-05
Remove bad, breaky advice

Drugoy:
`yum remove docker-selinux` and `yum remove docker-engine-selinux` result into a prompt to remove `container-selinux` package, which, in turn, is a dependency for `docker-ce`, which results into a prompt to remove it as well.

### DIFF
--- a/install/linux/docker-ce/centos.md
+++ b/install/linux/docker-ce/centos.md
@@ -49,8 +49,6 @@ $ sudo yum remove docker \
                   docker-latest \
                   docker-latest-logrotate \
                   docker-logrotate \
-                  docker-selinux \
-                  docker-engine-selinux \
                   docker-engine
 ```
 


thaJeztah:
Interesting, wondering why it would suggest removing that as well.

IIRC, the reason for removing these, was especially because they may interfere with `container-selinux`

@seemethere any ideas?

seemethere:
I believe `container-selinux` conflicts/obsoletes `docker-?(engine)-selinux`. Not entirely sure why it would prompt you to remove `container-selinux` though when you do not have the `docker-selinux` package installed.

@Drugoy can you let us know how to reproduce your error? i.e. what was your package setup.

Drugoy:
@thaJeztah 
because `docker-selinux` is mentioned in <a href="https://github.com/containers/container-selinux/blob/master/contrib/container-selinux.spec#L65">Obsoletes section for container-selinux</a>

@seemethere 
easy: `sudo yum install -y docker-ce && sudo yum remove -y docker-selinux && yum list installed | grep docker-ce` and observe `docker-ce` being removed.

seemethere:
Could reproduce, PR LGTM

PR #7988, opened 2019-01-08
Add redirects for apidocs pages

bermudezmt:
Removes apidocs references and redirect the pages to /docker-hub.


JustinINevill:
Cleaning outdated pages, all current api docs remain.

PR #7989, opened 2019-01-08
Removed outdated API docs and stub pages w/o functional content.

ahh-docker:
No pretty pics, but at least the API docs are gone.


JustinINevill:
Only removing outdated pages, all current api docs remain.

PR #7992, opened 2019-01-09
Fix compose file defaults

ushuz:
### Proposed changes

Correct and complement default values in compose file documentation to reflect the actual defaults in code.

- `restart_policy.delay` defaults to `5s`

https://github.com/docker/swarmkit/blob/87d4106b01c7764aab51847f445ac9369fae161c/api/defaults/service.go#L23-L26

- `update_ / rollback_config`'s `monitor` defaults to `5s` and `parallelism` defaults to `1`. 

https://github.com/docker/swarmkit/blob/87d4106b01c7764aab51847f445ac9369fae161c/api/defaults/service.go#L29-L40

### DIFF
--- a/compose/compose-file/index.md
+++ b/compose/compose-file/index.md
@@ -736,7 +736,7 @@ Configures if and how to restart containers when they exit. Replaces
 
 - `condition`: One of `none`, `on-failure` or `any` (default: `any`).
 - `delay`: How long to wait between restart attempts, specified as a
-  [duration](#specifying-durations) (default: 0).
+  [duration](#specifying-durations) (default: 5s).
 - `max_attempts`: How many times to attempt to restart a container before giving
   up (default: never give up). If the restart does not succeed within the configured
   `window`, this attempt doesn't count toward the configured `max_attempts` value.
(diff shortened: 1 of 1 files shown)


bermudezmt:
@shin- significant content updates. Can you please have a look?

thaJeztah:
We may have to be slightly careful with mentioning these as the _compose file_ defaults, given that the default on the client is (e.g. `0`), but if the client does not define a value, the _server_ sets a default. Overall these defaults have been "stable" (haven't changed, or changed much), but they _may_ change at some point and/or some of them may be configurable on the daemon.

ushuz:
@thaJeztah 

I'm not sure about docker client side defaults. But from what I observed, my services are actually using the defaults that I found in swarmkit source code, so current defaults in documentation are incorrect and misleading.

I deployed those services using `docker stack` command on both 17.12.1 and 18.06.1. This is the `deploy` section from my compose file:

```yaml
      deploy:
        ...
        restart_policy:
          max_attempts: 3
```

and this is what I found with `docker service inspect`:

```json
...
                "RestartPolicy": {
                    "Condition": "any",
                    "Delay": 5000000000,
                    "MaxAttempts": 3
                },
...
            "UpdateConfig": {
                "Parallelism": 1,
                "FailureAction": "pause",
                "Monitor": 5000000000,
                "MaxFailureRatio": 0,
                "Order": "stop-first"
            },
            "RollbackConfig": {
                "Parallelism": 1,
                "FailureAction": "pause",
                "Monitor": 5000000000,
                "MaxFailureRatio": 0,
                "Order": "stop-first"
            },
...
```

thaJeztah:
Right; those defaults are not set in the service definition though; they're set by the swarmkit manager.


The reason you see those values is that the services API for has a `?insertDefaults=true` option that augments the active defaults, and allows you to see the active configuration, with defaults added in. The service itself actually doesn't have those values set.

Here's an example (without the defaults visible);

```
curl --unix-socket /var/run/docker.sock http://localhost/v1.39/services/test
```

Output formatted and collapsed for readability;


<details>

```json
{
  "ID": "vvtqh7js4klwx7o2r66ihr0qq",
  "Version": {
    "Index": 11
  },
  "CreatedAt": "2019-01-13T21:01:55.663610489Z",
  "UpdatedAt": "2019-01-13T21:01:55.663610489Z",
  "Spec": {
    "Name": "test",
    "Labels": {},
    "TaskTemplate": {
      "ContainerSpec": {
        "Image": "nginx:alpine@sha256:385fbcf0f04621981df6c6f1abd896101eb61a439746ee2921b26abc78f45571",
        "Init": false,
        "DNSConfig": {},
        "Isolation": "default"
      },
      "Resources": {
        "Limits": {},
        "Reservations": {}
      },
      "Placement": {
        "Platforms": [
          {
            "Architecture": "amd64",
            "OS": "linux"
          },
          {
            "OS": "linux"
          },
          {
            "Architecture": "arm64",
            "OS": "linux"
          },
          {
            "Architecture": "386",
            "OS": "linux"
          },
          {
            "Architecture": "ppc64le",
            "OS": "linux"
          },
          {
            "Architecture": "s390x",
            "OS": "linux"
          }
        ]
      },
      "ForceUpdate": 0,
      "Runtime": "container"
    },
    "Mode": {
      "Replicated": {
        "Replicas": 1
      }
    },
    "EndpointSpec": {
      "Mode": "vip"
    }
  },
  "Endpoint": {
    "Spec": {}
  }
}
```
</details>

<p><br /></p>
And the same service, but this time with the `?insertDefaults=true` option, you see the defaults that are used (but determined by the manager);

```bash
curl --unix-socket /var/run/docker.sock 'http://localhost/v1.39/services/test?insertDefaults=true' 
```

<details>

```json
{
  "ID": "vvtqh7js4klwx7o2r66ihr0qq",
  "Version": {
    "Index": 11
  },
  "CreatedAt": "2019-01-13T21:01:55.663610489Z",
  "UpdatedAt": "2019-01-13T21:01:55.663610489Z",
  "Spec": {
    "Name": "test",
    "Labels": {},
    "TaskTemplate": {
      "ContainerSpec": {
        "Image": "nginx:alpine@sha256:385fbcf0f04621981df6c6f1abd896101eb61a439746ee2921b26abc78f45571",
        "Init": false,
        "StopGracePeriod": 10000000000,
        "DNSConfig": {},
        "Isolation": "default"
      },
      "Resources": {
        "Limits": {},
        "Reservations": {}
      },
      "RestartPolicy": {
        "Condition": "any",
        "Delay": 5000000000,
        "MaxAttempts": 0
      },
      "Placement": {
        "Platforms": [
          {
            "Architecture": "amd64",
            "OS": "linux"
          },
          {
            "OS": "linux"
          },
          {
            "Architecture": "arm64",
            "OS": "linux"
          },
          {
            "Architecture": "386",
            "OS": "linux"
          },
          {
            "Architecture": "ppc64le",
            "OS": "linux"
          },
          {
            "Architecture": "s390x",
            "OS": "linux"
          }
        ]
      },
      "ForceUpdate": 0,
      "Runtime": "container"
    },
    "Mode": {
      "Replicated": {
        "Replicas": 1
      }
    },
    "UpdateConfig": {
      "Parallelism": 1,
      "FailureAction": "pause",
      "Monitor": 5000000000,
      "MaxFailureRatio": 0,
      "Order": "stop-first"
    },
    "RollbackConfig": {
      "Parallelism": 1,
      "FailureAction": "pause",
      "Monitor": 5000000000,
      "MaxFailureRatio": 0,
      "Order": "stop-first"
    },
    "EndpointSpec": {
      "Mode": "vip"
    }
  },
  "Endpoint": {
    "Spec": {}
  }
}
```

</details>

<p><br /></p>

Given that the compose file (nor the `docker cli`) set those defaults, we should be careful _how_ we document them, to prevent the assumption that they're set by the client (and part of the service's specification); also because its _possible_ that some of those options are _configurable_ on the daemon (in which case the default may be different). I think for some other locations we attempted to describe this, something like;

"If no value is specified, the default is defined by the daemon."


But perhaps we can find a better approach.

usha-mandya:
@ushuz - Thank you for taking the time to review and update our docs. We really appreciate your effort.

I've discussed this with @thaJeztah. The long-term plan is to update the CLI to avoid printing default values that are determined by the daemon, and only display the standard defaults.
Meanwhile, we'll update docs to clarify and list options that derive their default values from the deamon. We'll create an internal ticket to track this work.

Thanks again for your contribution and hope to see more PRs from you.

PR #7999, opened 2019-01-10
UCP Azure Installation patch

ollypom:
### Proposed changes

Updated the documentation to call out that the Azure Object Name and Compute name need to match. And that the azure.json file needs 0644 permissions.


L-Hudson:
Good to go

PR #8003, opened 2019-01-11
Update Mac OS dependency from El Capitan to Sierra

wtriplett:
The link referenced at the top takes the user to a download page claiming Mac OS Sierra 10.12 is required.

### Proposed changes

Updated the Mac OS system requirements to Sierra


L-Hudson:
Good to go

PR #8005, opened 2019-01-11
Installation docs: remove distros that are no longer supported

thaJeztah:
Removes debian jessie, debian wheezy, ubuntu trusty, and Fedora 26, 27

PR #8006, opened 2019-01-11
Remove edge documentation

thaJeztah:


thaJeztah:
ping @ahh-docker @seemethere @L-Hudson PTAL; I think this one should be ready for review; I based this on top of https://github.com/docker/docker.github.io/pull/8005, so I'll rebase once that's merged.

**I kept commits logically grouped; please use _merge_ instead of _squash and merge_** to preserve the information from these commits.

_config_authoring.yml
```diff
@@ -2,36 +2,38 @@ name: Docker Documentation
...
 # jekyll serve --incremental --config _config_authoring.yml
 
-docker_ce_stable_version: "17.09"
-latest_stable_docker_engine_api_version: "1.32"
-docker_ce_edge_version: "17.09"
-docker_ee_version: "17.06"
-compose_version: "1.16.1"
-machine_version: "0.12.2"
+latest_engine_api_version: "1.39"
+docker_ce_version: "18.09"
```
thaJeztah: This is gonna conflict with https://github.com/docker/docker.github.io/pull/8010, but shouldn't be hard to fix once that's merged

_config.yml
```diff
@@ -16,12 +16,11 @@ url: https://docs.docker.com
...
-# Component versions -- address like site.docker_ce_stable_version
+# Component versions -- address like site.docker_ce_version
 # You can't have - characters in these for non-YAML reasons
 
-latest_stable_docker_engine_api_version: "1.39"
-docker_ce_stable_version: "18.09"
-docker_ce_edge_version: "18.09"
+latest_engine_api_version: "1.39"
+docker_ce_version: "18.09"
 docker_ee_version: "18.09"
```
thaJeztah: FWIW; was thinking of combining the two variables above to just `engine_version` (given that these only hold the major version, and those are in sync now), but left that for a possible follow-up

ahh-docker:
@L-Hudson PTAL

_config_authoring.yml
```diff
@@ -129,29 +147,48 @@ defaults:
...
+      ucp_version: "3.1.1"
+  - scope: # This is a bit of a hack for the get-support.md topic.
+      path: "ee"
+    values:
+      ucp_org: "docker"
+      ucp_repo: "ucp"
+      dtr_repo: "dtr"
+      ucp_version: "3.1.1"
+      dtr_version: "2.6.0"
+      dtr_latest_image: "docker/dtr:2.6.0"
```
bermudezmt: @thaJeztah should be `2.6.1` for latest version, yeah? FWIW, I don't think anyone's been using `_config_authoring.yml` for the last release, but it will be nice to keep this up-to-date and usable for the next one. It was my understanding that this was an alternate config file for incremental changes and previewing those changes on your local build. Or did I miss the point entirely?

_config_authoring.yml
bermudezmt: #8010 has been merged.

_config.yml
bermudezmt: I like the idea of having one variable for Engine version.

_config_authoring.yml
thaJeztah: Ah; yes, it's updated now that https://github.com/docker/docker.github.io/pull/8010 is merged.

> I don't think anyone's been using _config_authoring.yml for the last release

Yeah, it's likely not often used; we should either remove it (or make sure to keep it in sync)

I'll do a quick rebase and sync it again

thaJeztah: updated :+1:

_config.yml
thaJeztah: I'll do it in a follow-up; if we reserve this variable for the _major_ (`YY.MM`) version only, that should work; we may need to keep a separate variable if we want to keep track of patch releases possibly, but I don't think we currently do.

bermudezmt: We have page variables for DTR and UCP that represent _major.minor_ versioning, but I haven't seen an instance for Engine. We could double-check though, but I don't think it will be an issue.

thaJeztah:
Just to be sure (I see the default is "squash and merge"); 

> **I kept commits logically grouped; please use _merge_ instead of _squash and merge_** to preserve the information from these commits. 🙏 🤗

L-Hudson:
Not sure if the continuous-integration/jenkins/pr-merge check is necessary. But this is good to merge.

thaJeztah:
Asked about that yesterday; and looks like the `continuous-integration/jenkins/pr-merge` is not ready yet, so safe to ignore

PR #8018, opened 2019-01-14
Update index.md

Shofol:
### Proposed changes

The first image looks smaller and unbalanced. I have set the width to 100% which looks better.


L-Hudson:
Good to go

PR #8019, opened 2019-01-14
Fix link in security.md

hedenface:
Format (Markdown) for "Content trust in Docker" link was incorrect. Also updated the link itself to be inline with the rest of them (at the bottom of the page).


hedenface:
Verified that it works with the deploy preview, FWIW

PR #8024, opened 2019-01-15
fix example of docker-network

neotusca:
### Proposed changes

I fixed example string from 'bridge alpine4' to 'alpine-net alpine3'.
It's just work right.

### DIFF
--- a/network/network-tutorial-standalone.md
+++ b/network/network-tutorial-standalone.md
@@ -306,7 +306,7 @@ connected to both networks.
 
     $ docker run -dit --name alpine4 --network alpine-net alpine ash
 
-    $ docker network connect bridge alpine4
+    $ docker network connect alpine-net alpine3
     ```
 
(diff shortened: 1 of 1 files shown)


L-Hudson:
@thaJeztah , please take a look?

network/network-tutorial-standalone.md
```diff
@@ -306,7 +306,7 @@ connected to both networks.
 
     $ docker run -dit --name alpine4 --network alpine-net alpine ash
 
-    $ docker network connect bridge alpine4
+    $ docker network connect alpine-net alpine3
```
thaJeztah: Looking at the description above, the original text was correct;

> you need to use `docker network connect` afterward to connect `alpine4` to the `bridge` network as well.

The example describes that container `alpine4` is connected to network `alpine-net`, and after that is connected to network `bridge` (so that it is now connected to both networks)

PR #8027, opened 2019-01-16
Fix default directory path for Windows

heaths:
As previously stated, the entire `%ProgramData%` directory would be ignored. This also uses a PowerShell syntax for which most people may not be familiar, but most often PowerShell users know how to translate from supported environment variables. The main point is to limit the directory to ignore to just docker. Many, many other applications write to `%ProgramData%`.

### DIFF
--- a/engine/security/antivirus.md
+++ b/engine/security/antivirus.md
@@ -8,7 +8,7 @@ When antivirus software scans files used by Docker, these files may be locked
 in a way that causes Docker commands to hang.
 
 One way to reduce these problems is to add the Docker data directory
-(`/var/lib/docker` on Linux, `$Env:ProgramData` on Windows Server, or `$HOME/Library/Containers/com.docker.docker/` on Mac) to the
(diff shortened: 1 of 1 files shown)


heaths:
As this is a doc change I imagine the build issue is unrelated; however, I don't have access to this public PR build to see what failed.

heaths:
@thaJeztah could you take a look at the pr-merge build? This is a simple content change and I can't imagine what might go wrong, but I don't have permissions to see the build results.

ahh-docker:
@jasonbivins PTAL. Thanks!

jasonbivins:
I agree LGTM - it's much better to suggest we exclude only our subdirectory `%ProgramData%\docker`  vs everything in the directory `$Env:ProgramData`

ahh-docker:
@L-Hudson PTAL. If it's good, please go ahead and merge. Thanks!

L-Hudson:
Good to go

PR #8032, opened 2019-01-16
Fix incorrect capitalization of --env argument

lindhe:
I found, in version `18.03.0-ce, build 0520e24` of docker, that the --env argument *cannot* be capital letters.

**I have only tested this for http_proxy and https_proxy, not ftp_proxy and no_proxy, so I just made the assumption that it should be the same.**

Perhaps the CLI should be more lenient?

### Proposed changes

Change the documentation of the argument given to the `--env` flag to be lowercase letters, since uppercase does not work.

### DIFF
--- a/network/proxy.md
+++ b/network/proxy.md
@@ -60,8 +60,8 @@ instead.
 
 | Variable      | Dockerfile example                                | `docker run` Example                                |
 |:--------------|:--------------------------------------------------|:----------------------------------------------------|
-| `HTTP_PROXY`  | `ENV HTTP_PROXY "http://127.0.0.1:3001"`          | `--env HTTP_PROXY="http://127.0.0.1:3001"`          |
-| `HTTPS_PROXY` | `ENV HTTPS_PROXY "https://127.0.0.1:3001"`        | `--env HTTPS_PROXY="https://127.0.0.1:3001"`        |
-| `FTP_PROXY`   | `ENV FTP_PROXY "ftp://127.0.0.1:3001"`            | `--env FTP_PROXY="ftp://127.0.0.1:3001"`            |
-| `NO_PROXY`    | `ENV NO_PROXY "*.test.example.com,.example2.com"` | `--env NO_PROXY="*.test.example.com,.example2.com"` |
+| `HTTP_PROXY`  | `ENV HTTP_PROXY "http://127.0.0.1:3001"`          | `--env http_proxy="http://127.0.0.1:3001"`          |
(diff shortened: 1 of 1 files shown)


network/proxy.md
```diff
@@ -60,8 +60,8 @@ instead.
 
 | Variable      | Dockerfile example                                | `docker run` Example                                |
 |:--------------|:--------------------------------------------------|:----------------------------------------------------|
-| `HTTP_PROXY`  | `ENV HTTP_PROXY "http://127.0.0.1:3001"`          | `--env HTTP_PROXY="http://127.0.0.1:3001"`          |
-| `HTTPS_PROXY` | `ENV HTTPS_PROXY "https://127.0.0.1:3001"`        | `--env HTTPS_PROXY="https://127.0.0.1:3001"`        |
-| `FTP_PROXY`   | `ENV FTP_PROXY "ftp://127.0.0.1:3001"`            | `--env FTP_PROXY="ftp://127.0.0.1:3001"`            |
-| `NO_PROXY`    | `ENV NO_PROXY "*.test.example.com,.example2.com"` | `--env NO_PROXY="*.test.example.com,.example2.com"` |
+| `HTTP_PROXY`  | `ENV HTTP_PROXY "http://127.0.0.1:3001"`          | `--env http_proxy="http://127.0.0.1:3001"`          |
```
thaJeztah: Can you describe the problem you're having with Docker 18.03? Environment variables on Linux are case-sensitive, so both uppercase and lowercase can be used. FWIW, there is no official standard on the "right" capitalisation for these proxy env-vars (some tools on linux prefer lowercase, other uppercase, but mostly fall back to the other one if it's not set).

Using uppercase should work fine; in this example, I start a container having both the upper and lowercase version, and print the active environment variables inside the container (using the `env` linux command); both are set;

```bash
docker run --rm --env https_proxy="something" --env HTTPS_PROXY="something else" busybox env                                                                                                          PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
HOSTNAME=112c56ef9f44
https_proxy=something
HTTPS_PROXY=something else
HOME=/root
```

lindhe: Interesting. My experience was that it did not set the environment variables at all when using uppercase letters. But I will have to double check this.

L-Hudson: @lindhe please let me know how you'd like to resolve this ticket. Thanks!

lindhe: Sorry, I forgot about it. Let's close it, and if I do get back to it eventually I'll reopen it.

lindhe:
If I get time, I'll check out this problem in the future and see if I can reproduce. Until then: let's assume I was in the wrong.

PR #8033, opened 2019-01-16
ucp: support dump running version

trapier:
Use a bash shell variable to run support dump at the same revision as
the cluster is running.  This is a copy-paste convenience for users in offline environments that might not be running the latest version.

/cc @ekitagawa as a heads up that we might want to do the same thing
with powershell a few lines lower.

### DIFF
--- a/ee/get-support.md
+++ b/ee/get-support.md
@@ -35,11 +35,12 @@ To get the support dump from the Web UI:
 To get the support dump from the CLI, use SSH to log into a node and run:
 
 ```bash
+UCP_VERSION=$((docker container inspect ucp-proxy --format {% raw %}'{{index .Config.Labels "com.docker.ucp.version"}}'{% endraw %} 2>/dev/null || echo -n {{ page.ucp_version  }})|tr -d [[:space:]])
 docker container run --rm \
   --name ucp \
   -v /var/run/docker.sock:/var/run/docker.sock \
   --log-driver none \
-  {{ page.ucp_org }}/{{ page.ucp_repo }}:{{ page.ucp_version }} \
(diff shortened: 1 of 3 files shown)


trapier:
yep.  needed raw endraw.  waiting on next render.

trapier:
raw endraw added blank lines to the code block.  moving to in-line.

ahh-docker:
@trapier lemme know when this is ready to go and I'll merge it.

ee/get-support.md
```diff
@@ -35,11 +35,12 @@ To get the support dump from the Web UI:
 To get the support dump from the CLI, use SSH to log into a node and run:
 
 ```bash
+UCP_VERSION=$((docker container inspect ucp-proxy --format {% raw %}'{{index .Config.Labels "com.docker.ucp.version"}}'{% endraw %} 2>/dev/null || echo -n {{ page.ucp_version  }})|tr -d [[:space:]])
```
trapier: @bkdocker: it works, but this got a little crazy.  do you still think this is a good idea?

bkdocker: looks great

trapier:
testing.  found syntax error in the 2.2 copy.

trapier:
Tested by copy-paste from a local build of the docs for 2.2, 3.0, 3.1 with and without ucp-proxy container present.   Should be good to go once it passes CI.  Thanks!

ahh-docker:
@trapier LGTM. Anything @L-Hudson needs to do for editing?

L-Hudson:
Good to go

ahh-docker:
Thanks @trapier! Your help is always appreciated :)

PR #8041, opened 2019-01-17
Fix error (small change, only 1 line)

1c7:
### Proposed changes
https://docs.docker.com/registry/deploying/
image

### DIFF
--- a/registry/deploying.md
+++ b/registry/deploying.md
@@ -64,7 +64,6 @@ as `my-ubuntu`, then pushes it to the local registry. Finally, the
 
     ```bash
     $ docker image remove ubuntu:16.04
-    $ docker image remove localhost:5000/my-ubuntu
     ```
 
 5.  Pull the `localhost:5000/my-ubuntu` image from your local registry.


registry/deploying.md
```diff
@@ -64,7 +64,6 @@ as `my-ubuntu`, then pushes it to the local registry. Finally, the
 
     ```bash
     $ docker image remove ubuntu:16.04
-    $ docker image remove localhost:5000/my-ubuntu
```
thaJeztah: Whoops, just too late, but this change looks not correct to me;

This documentation describes how to;

- pull an image (ubuntu)
- tag it under a different name (`localhost:5000/my-ubuntu`)
- push it to your local registry (`docker push localhost:5000/my-ubuntu`)

After that, this step describes to _remove_ the image from your local cache (`docker image remove localhost:5000/my-ubuntu`)

To verify that you can pull it again from your local registry in the next step.

The "remove" step is so that you can see that it was really pulled from the registry (and it not being the image you already had locally).

I think this change should be reverted

1c7: I am confused.
base on the text, the tutorial is trying to 
1. pull a imae
2. push that image to local registry
3. remove that image from local machine
4. pull that image back from local registry

but `docker image remove ubuntu:16.04` and `docker image remove localhost:5000/my-ubuntu`
seem like remove both local & registry.
that doesn't seem right to me.

1c7: I am a docker newbie, at this point I only spend about 10 day learning Docker from different source

1c7: @thaJeztah

thaJeztah: > seem like remove both local & registry.

No, `docker image remove` will only ever remove your local images (remove them from your local cache), never from a registry.

1c7: got it, Thanks for clear things out

PR #7468, opened 2018-10-05
Update end-to-end-install.md

ajeetraina:
Infrastructure:

```
Client: Docker Enterprise Edition (EE) 2.0
 Version:      17.06.2-ee-16
 API version:  1.30
 Go version:   go1.8.7
 Git commit:   9ef4f0a
 Built:        Thu Jul 26 16:41:28 2018
 OS/Arch:      linux/amd64
Server: Docker Enterprise Edition (EE) 2.0
 Engine:
  Version:      17.06.2-ee-16
  API version:  1.30 (minimum version 1.12)
  Go version:   go1.8.7
  Git commit:   9ef4f0a
  Built:        Thu Jul 26 16:40:18 2018
  OS/Arch:      linux/amd64
  Experimental: false
```

```
openusm@node-e1:~$ docker info
Containers: 30
 Running: 27
 Paused: 0
 Stopped: 3
Images: 24
Server Version: 17.06.2-ee-16
Storage Driver: overlay2
 Backing Filesystem: extfs
 Supports d_type: true
 Native Overlay Diff: true
Logging Driver: json-file
Cgroup Driver: cgroupfs
Plugins: 
 Volume: local
 Network: bridge host macvlan null overlay
 Log: awslogs fluentd gcplogs gelf journald json-file logentries splunk syslog
Swarm: active
 NodeID: a0cy301twurh8na6xngp6obmt
 Is Manager: true
 ClusterID: b92spaikb3jc6u6lt3yfwxtpk
 Managers: 1
 Nodes: 1
 Orchestration:
  Task History Retention Limit: 5
 Raft:
  Snapshot Interval: 10000
  Number of Old Snapshots to Retain: 0
  Heartbeat Tick: 1
  Election Tick: 10
 Dispatcher:
  Heartbeat Period: 5 seconds
 CA Configuration:
  Expiry Duration: 3 months
  Force Rotate: 0
  External CAs:
    cfssl: https://10.140.0.4:12381/api/v1/cfssl/sign
 Root Rotation In Progress: false
 Node Address: 10.140.0.4
 Manager Addresses:
  10.140.0.4:2377
Runtimes: runc
Default Runtime: runc
Init Binary: docker-init
containerd version: 6e23458c129b551d5c9871e5174f6b1b7f6d1170
runc version: 462c82662200a17ee39e74692f536067a3576a50
init version: 949e6fa
Security Options:
 apparmor
 seccomp
  Profile: default
Kernel Version: 4.15.0-1021-gcp
[...]

PR #8046, opened 2019-01-18
Updated Docker Enterprise End to End Install Docs

ollypom:
### Proposed changes
I've updated the End to End Install Docs, removing the duplicated System Requisites information. This information was starting to become stale, so keeping it one place will be easier to maintain. I've also updated the Images to EE 2.1.

Included this System Requirement change from https://github.com/docker/docker.github.io/pull/7468


L-Hudson:
Good to go.

PR #8050, opened 2019-01-18
typo in "Get Started, Part 2": extraneous 'a'

stuartcrobinson:
### Proposed changes

in the **Build the app** section of https://docs.docker.com/get-started/part2/, replace:

`if you want to use the a shorter option` with
`if you want to use the shorter option`

In context:

> Now run the build command. This creates a Docker image, which we’re going to name using the --tag option. Use -t if you want to use the a shorter option.


L-Hudson:
Good catch, thanks!

L-Hudson:
Good catch, thanks!

PR #8057, opened 2019-01-21
Update deploying.md

Mueslint:
Got a problem while copy/pasting the full command, throws "repository name must be lowercased". 

### Proposed changes

I still don't know if it's due to mac OS X or `zsh` but the line wasn't working and I had to replace
```-v `(pwd)`/certs:/certs``` with `-v "$(pwd)"/certs:/certs` 

same for the part with basic auth 

Found the fix here
https://stackoverflow.com/questions/48522615/docker-error-invalid-reference-format-repository-name-must-be-lowercase

It might be too specific to be accepted...

### DIFF
--- a/registry/deploying.md
+++ b/registry/deploying.md
@@ -216,7 +216,7 @@ If you have been issued an _intermediate_ certificate instead, see
     $ docker run -d \
       --restart=always \
       --name registry \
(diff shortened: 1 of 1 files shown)


bermudezmt:
Thank you so much for this PR, @Mueslint !

PR #8059, opened 2019-01-21
Document registry s3 transfer acceleration option.

monder:
### Proposed changes

This change adds documentation for yet unmerged PR here as requested:
https://github.com/docker/distribution/pull/2166


L-Hudson:
good to go

monder:
@L-Hudson Hey, sorry if I somehow misphrased it, but the original PR is still not merged and that flag is still not available in `docker/distribution` =(

PR #8061, opened 2019-01-22
Adding proxy wildcard documentation

ChaosGramer:
Wildcard documentation was missing in the official docs

### DIFF
--- a/config/daemon/systemd.md
+++ b/config/daemon/systemd.md
@@ -96,24 +96,34 @@ you need to add this configuration in the Docker systemd service file.
     ```
 
 3.  If you have internal Docker registries that you need to contact without
-    proxying you can specify them via the `NO_PROXY` environment variable:
+    proxying you can specify them via the `NO_PROXY` environment variable.
 
+    The `NO_PROXY` variable specifies a string that contains comma-separated values for hosts that should be excluded from proxying. 
+    These are the options you have to specify the excluded hosts: 
+    * IP address prefix (`1.2.3.4`) or in CIDR notation (`1.2.3.4/8`)    
+    * Domain name, or a special DNS label (`*`)
+      * A domain name matches that name and all subdomains. A domain name with
+           a leading "." matches subdomains only. Example:
+           ```
+           Given the domains foo.example.com and example.com.
+           "foo.com" matches "foo.com" and "bar.foo.com"
+           ".foo.com" mataches only "bar.foo.com"
+           ```
+    * A single asterisk (*) indicates that no proxying should be done
+    * Literal port numbers are accepted by IP address prefixes (`1.2.3.4:80`) and domain names (`foo.example.com:80`)
+    
+    Config examples:
+    
     ```conf
     [Service]    
-    Environment="HTTP_PROXY=http://proxy.example.com:80/" "NO_PROXY=localhost,127.0.0.1,docker-registry.somecorporation.com"
+    Environment="HTTP_PROXY=http://proxy.example.com:80/" "NO_PROXY=localhost,127.0.0.1,docker-registry.example.com,.corp"
     ```
 
     Or, if you are behind an HTTPS proxy server:
 
     ```conf
     [Service]    
-    Environment="HTTPS_PROXY=https://proxy.example.com:443/" "NO_PROXY=localhost,127.0.0.1,docker-registry.somecorporation.com"
-    ```
-    
-    The NO_PROXY variable accepts wildcards. You just need to have a value starting with a dot:
-    ```conf
-    [Service]    
-    Environment="HTTPS_PROXY=https://proxy.example.com:443/" "NO_PROXY=localhost,127.0.0.1,.somecorporation.com"
+    Environment="HTTPS_PROXY=https://proxy.example.com:443/" "NO_PROXY=localhost,127.0.0.1,docker-registry.example.com,.corp"
     ```
 
 4.  Flush changes:
--- a/config/daemon/systemd.md
+++ b/config/daemon/systemd.md
@@ -121,7 +121,7 @@ you need to add this configuration in the Docker systemd service file.
            "foo.com" matches "foo.com" and "bar.foo.com"
            ".foo.com" mataches only "bar.foo.com"
            ```
-    * A single asterisk (*) indicates that no proxying should be done
+    * A single asterisk (`*`) indicates that no proxying should be done
     * Literal port numbers are accepted by IP address prefixes (`1.2.3.4:80`) and domain names (`foo.example.com:80`)
     
     Config examples:
--- a/config/daemon/systemd.md
+++ b/config/daemon/systemd.md
@@ -111,7 +111,7 @@ you need to add this configuration in the Docker systemd service file.
     * A single asterisk (`*`) indicates that no proxying should be done
     * Literal port numbers are accepted by IP address prefixes (`1.2.3.4:80`)
       and domain names (`foo.example.com:80`)
-    These are the options you have to specify the excluded hosts: 
+    These are the options you can specify to exclude hosts:
     * IP address prefix (`1.2.3.4`) or in CIDR notation (`1.2.3.4/8`)    
     * Domain name, or a special DNS label (`*`)
       * A domain name matches that name and all subdomains. A domain name with
--- a/config/daemon/systemd.md
+++ b/config/daemon/systemd.md
@@ -109,6 +109,12 @@ you need to add this configuration in the Docker systemd service file.
     [Service]    
     Environment="HTTPS_PROXY=https://proxy.example.com:443/" "NO_PROXY=localhost,127.0.0.1,docker-registry.somecorporation.com"
     ```
+    
+    The NO_PROXY variable accepts wildcards. You just need to have a value starting with a dot:
+    ```conf
+    [Service]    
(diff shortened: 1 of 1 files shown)


L-Hudson:
@thaJeztah can you take a look? Thanks!

config/daemon/systemd.md
```diff
@@ -109,6 +109,12 @@ you need to add this configuration in the Docker systemd service file.
     [Service]    
     Environment="HTTPS_PROXY=https://proxy.example.com:443/" "NO_PROXY=localhost,127.0.0.1,docker-registry.somecorporation.com"
     ```
+    
+    The NO_PROXY variable accepts wildcards. You just need to have a value starting with a dot:
```
thaJeztah: Wildcards may not be the correct term to use here (as it doesn't accept `*`).

Given that there's no canonical standard for the formats supported in `NO_PROXY` (e.g. curl accepts other values than golang does, see the discussion in https://github.com/golang/go/issues/16704), I suggest to use an adoption of the description from the Golang source-code (as those are the formats that are accepted by the docker daemon); https://github.com/golang/go/blob/4da84adc0c55d92f20b761d24a6ec0508e55dc7c/src/vendor/golang_org/x/net/http/httpproxy/proxy.go#L39-L51

thaJeztah: oh, actually; it does accept `*` but (I think) only as single value (`*`), not as _part of_ a domain or IP-address 😅

ChaosGramer: What about adding this text as detailed explanation? 
(Taken from the go source code you linked) 
Notice for the NO_PROXY enviroment  variable.
It specifies a string that contains comma-separated values specifying hosts that should be excluded from proxying. Each value is represented by an IP address prefix (1.2.3.4), an IP address prefix in CIDR notation (1.2.3.4/8), a domain name, or a special DNS label (*).
An IP address prefix and domain name can also include a literal port number (1.2.3.4:80).

A domain name matches that name and all subdomains. A domain name with
a leading "." matches subdomains only. For example "foo.com" matches
"foo.com" and "bar.foo.com"; ".y.com" matches "x.y.com" but not "y.com".
A single asterisk (*) indicates that no proxying should be done.

A best effort is made to parse the string and errors are ignored.

L-Hudson:
Followed up with @thaJeztah on slack

config/daemon/systemd.md
thaJeztah: Sorry for the delay; that suggestion looks pretty good; some small nits;

- Don't use "Notice for"; start with "The `NO_PROXY` environment variable....." (and combine with the next sentence)
- Put values in backticks so that they show in a fixed width font (I assume you omitted that in your comment because it was just to discuss, but just in case 😇🤗)
- For bonus points; use `foo.example.com` and/or `foo.example.org` for the domain example; those domains exist especially for documentation purposes (there's many places where we didn't use them, but since were updating this paragraph, we might as well fix it along the way :tada:)

ChaosGramer: Thanks for the feedback. Going to prepare a new commit.

config/daemon/systemd.md
```diff
@@ -96,18 +96,34 @@ you need to add this configuration in the Docker systemd service file.
...
+    * IP address prefix (`1.2.3.4`) or in CIDR notation (`1.2.3.4/8`)    
+    * Domain name, or a special DNS label (`*`)
+      * A domain name matches that name and all subdomains. A domain name with
+           a leading "." matches subdomains only. Example:
+           ```
+           Given the domains foo.example.com and example.com.
+           "foo.com" matches "foo.com" and "bar.foo.com"
+           ".foo.com" mataches only "bar.foo.com"
+           ```
+    * A single asterisk (*) indicates that no proxying should be done
```
thaJeztah:
```suggestion
    * A single asterisk (`*`) indicates that no proxying should be done
```

config/daemon/systemd.md
```diff
@@ -96,18 +96,34 @@ you need to add this configuration in the Docker systemd service file.
     ```
 
 3.  If you have internal Docker registries that you need to contact without
-    proxying you can specify them via the `NO_PROXY` environment variable:
-
+    proxying you can specify them via the `NO_PROXY` environment variable.
+
+    The `NO_PROXY` variable specifies a string that contains comma-separated values for hosts that should be excluded from proxying. 
+    These are the options you have to specify the excluded hosts:
```
thaJeztah: Perhaps change "have to" to "can";

```suggestion
    These are the options you can specify to exclude hosts:
```

config/daemon/systemd.md
```diff
@@ -96,18 +96,34 @@ you need to add this configuration in the Docker systemd service file.
...
 3.  If you have internal Docker registries that you need to contact without
-    proxying you can specify them via the `NO_PROXY` environment variable:
-
+    proxying you can specify them via the `NO_PROXY` environment variable.
+
+    The `NO_PROXY` variable specifies a string that contains comma-separated values for hosts that should be excluded from proxying. 
+    These are the options you have to specify the excluded hosts: 
+    * IP address prefix (`1.2.3.4`) or in CIDR notation (`1.2.3.4/8`)    
+    * Domain name, or a special DNS label (`*`)
+      * A domain name matches that name and all subdomains. A domain name with
```
thaJeztah: I think we don't need this bullet to be one level deeper; and we may be able to get rid of the code _block_ below. I gave this a quick go, and posted a suggestion in a comment on a few lines up 🤗

config/daemon/systemd.md
```diff
@@ -96,18 +96,34 @@ you need to add this configuration in the Docker systemd service file.
     ```
 
 3.  If you have internal Docker registries that you need to contact without
-    proxying you can specify them via the `NO_PROXY` environment variable:
-
+    proxying you can specify them via the `NO_PROXY` environment variable.
+
+    The `NO_PROXY` variable specifies a string that contains comma-separated values for hosts that should be excluded from proxying.
```
thaJeztah: I think we don't need this bullet to be one level deeper; and we may be able to get rid of the code _block_ below.

I gave this a quick go, and think something like this might work:

```suggestion
    The `NO_PROXY` variable specifies a string that contains comma-separated
    values for hosts that should be excluded from proxying. These are the
    options you can specify to exclude hosts: 
    * IP address prefix (`1.2.3.4`) or in CIDR notation (`1.2.3.4/8`)    
    * Domain name, or a special DNS label (`*`)
    * A domain name matches that name and all subdomains. A domain name with
      a leading "." matches subdomains only. For example, given the domains
      `foo.example.com` and `example.com`;
      * `foo.com` matches `foo.com` and `bar.foo.com`, and
      * `.foo.com` matches only `bar.foo.com`
    * A single asterisk (`*`) indicates that no proxying should be done
    * Literal port numbers are accepted by IP address prefixes (`1.2.3.4:80`)
      and domain names (`foo.example.com:80`)
```

Which will look something like this;

The `NO_PROXY` variable specifies a string that contains comma-separated
values for hosts that should be excluded from proxying. These are the
options you can specify to exclude hosts: 
* IP address prefix (`1.2.3.4`) or in CIDR notation (`1.2.3.4/8`)    
* Domain name, or a special DNS label (`*`)
* A domain name matches that name and all subdomains. A domain name with
  a leading "." matches subdomains only. For example, given the domains
  `foo.example.com` and `example.com`;
  * `foo.com` matches `foo.com` and `bar.foo.com`, and
  * `.foo.com` matches only `bar.foo.com`
* A single asterisk (`*`) indicates that no proxying should be done
* Literal port numbers are accepted by IP address prefixes (`1.2.3.4:80`)
  and domain names (`foo.example.com:80`)

thaJeztah: Oh! I see there's a mis-match between the domains; updating my comment

thaJeztah: Done 👍

ChaosGramer:
All your changes accepted thanks fo the feedback 👍

thaJeztah:
oh! looks like GitHub applied them after each other, leading to duplicate parts 😞 their "suggestion" feature is a bit limited currently. Erm. think you'd have to do that on the command-line instead of on GitHub then 😓 

Here's the whole block that I suggested;

```markdown
    The `NO_PROXY` variable specifies a string that contains comma-separated
    values for hosts that should be excluded from proxying. These are the
    options you can specify to exclude hosts: 
    * IP address prefix (`1.2.3.4`) or in CIDR notation (`1.2.3.4/8`)    
    * Domain name, or a special DNS label (`*`)
    * A domain name matches that name and all subdomains. A domain name with
      a leading "." matches subdomains only. For example, given the domains
      `foo.example.com` and `example.com`;
      * `example.com` matches `example.com` and `foo.example.com`, and
      * `.example.com` matches only `foo.example.com`
    * A single asterisk (`*`) indicates that no proxying should be done
    * Literal port numbers are accepted by IP address prefixes (`1.2.3.4:80`)
      and domain names (`foo.example.com:80`)
```

L-Hudson:
@ChaosGramer can you please address? I'll happily merge once you're clear.

ChaosGramer:
@L-Hudson What is missing? I addressed all the comments from @thaJeztah

L-Hudson:
Weird, one of the comments wasn't clearing this morning, but it's cleared now. Thank you so much for working through this, @ChaosGramer and @thaJeztah! Much appreciated.

PR #8062, opened 2019-01-22
Fix UCP App Package ToC

ollypom:
### Proposed changes

Fix the ToC for UCP App Package. Looks good in the preview :) 

From:

image

To

image


L-Hudson:
Good to go!

PR #8063, opened 2019-01-22
ucp config: custom api headers syntax

trapier:
Example for this table needs double brackets because it's an array.
Tested on UCP 3.0.8.


L-Hudson:
Good to go

PR #8064, opened 2019-01-23
add new Jenkinsfile

allysmith424:
- this file sets up builds for CSE in ci.docker.com
- master publishes to stage, published to production


L-Hudson:
Sent @peter-docker a nudge on Slack. :: nudge nudge ::

PR #8066, opened 2019-01-23
Update build_enhancements.md with troubleshooting hints

lbndev:
Add troubleshooting tips from https://github.com/moby/moby/issues/38303

### Proposed changes

Add troubleshooting tips from the issue below.
This is from my personal experience.
Finding this information in the documentation would have saved me hours of reasearch, so I think it would be beneficial to others to find the explanation and solution directly in the docs.

### Related issues (optional)

https://github.com/moby/moby/issues/38303

### DIFF
--- a/develop/develop-images/build_enhancements.md
+++ b/develop/develop-images/build_enhancements.md
@@ -193,7 +193,7 @@ $ docker build --ssh default .
 ## Troubleshooting : issues with private registries
 
 #### x509: certificate signed by unknown authority
-If you are fetching images from insecure registry (with self-signed certificates) and/or using such aa registry as a mirror, you are facing a known issue in Docker 18.09 :
+If you are fetching images from insecure registry (with self-signed certificates) and/or using such a registry as a mirror, you are facing a known issue in Docker 18.09 :
 ```
 [+] Building 0.4s (3/3) FINISHED
  => [internal] load build definition from Dockerfile
@@ -211,7 +211,7 @@ Solution : secure your registry properly. You can get SSL certificates from Let'
 
 #### image not found when the private registry is running on Sonatype Nexus version < 3.15
 
-If you are running a privatee registry using Sonatype Nexus version < 3.15, and receive an error similar to the following :
+If you are running a private registry using Sonatype Nexus version < 3.15, and receive an error similar to the following :
 ```
 ------
  > [internal] load metadata for docker.io/library/maven:3.5.3-alpine:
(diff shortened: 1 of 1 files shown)


develop/develop-images/build_enhancements.md
```diff
@@ -189,3 +189,39 @@ Once the `Dockerfile` is created, use the `--ssh` option for connectivity with t
 ```bash
 $ docker build --ssh default .
 ```
+
+## Troubleshooting : issues with private registries
+
+#### x509: certificate signed by unknown authority
+If you are fetching images from insecure registry (with self-signed certificates) and/or using such aa registry as a mirror, you are facing a known issue in Docker 18.09 :
```
caervs: `s/aa/a`

Also alternatively the user can set [insecure-registries](https://docs.docker.com/registry/insecure/)

develop/develop-images/build_enhancements.md
```diff
@@ -189,3 +189,39 @@ Once the `Dockerfile` is created, use the `--ssh` option for connectivity with t
...
+ > resolve image config for docker.io/docker/dockerfile:experimental:
+------
+failed to do request: Head https://repo.mycompany.com/v2/docker/dockerfile/manifests/experimental: x509: certificate signed by unknown authority
+```
+Solution : secure your registry properly. You can get SSL certificates from Let's Encrypt for free. See https://docs.docker.com/registry/deploying/
+
+
+#### image not found when the private registry is running on Sonatype Nexus version < 3.15
+
+If you are running a privatee registry using Sonatype Nexus version < 3.15, and receive an error similar to the following :
```
caervs: `s/ee/e`

develop/develop-images/build_enhancements.md
```diff
@@ -189,3 +189,39 @@ Once the `Dockerfile` is created, use the `--ssh` option for connectivity with t
...
+ => ERROR resolve image config for docker.io/docker/dockerfile:experimental
+------
+ > resolve image config for docker.io/docker/dockerfile:experimental:
+------
+failed to do request: Head https://repo.mycompany.com/v2/docker/dockerfile/manifests/experimental: x509: certificate signed by unknown authority
+```
+Solution : secure your registry properly. You can get SSL certificates from Let's Encrypt for free. See https://docs.docker.com/registry/deploying/
+
+
+#### image not found when the private registry is running on Sonatype Nexus version < 3.15
```
manishtomar: I am not sure what is our policy on documenting problems when talking to non-docker registries as adding info like this might give users expectation of having more info about other registries.

Issue #7954, opened 2018-12-31
Remote Syslog configuration not available in UCP 3.1


bootswithdefer:
[ee/ucp/admin/configure/store-logs-in-an-external-system.md](https://docs.docker.com/ee/ucp/admin/configure/store-logs-in-an-external-system/)

Docs should be updated to say that UCP 3.1 does not support Remote Syslog configuration from the UI, instead the Docker Engine must be configured in daemon.json

bermudezmt:
Closed by #8071 . Thank you, @ollypom !

Issue #7940, opened 2018-12-28
Remote syslog setup has been removed in 3.1.x < , misleading screenshot of UI


kavikat:
[ee/ucp/admin/configure/store-logs-in-an-external-system.md](https://docs.docker.com/ee/ucp/admin/configure/store-logs-in-an-external-system/)

kavikat:
The screenshot of the remote syslog setup in the UI is misleading for UCP 3.1.x users as this feature has ben removed.

bermudezmt:
Closed by #8071 .

Issue #7682, opened 2018-11-14
Updating doc for Store logs in Remote Server


prashshr:
### Problem description

Firstly, in UCP `3.1.0` under Logs there is no option for "Configuring Remote Syslog Server" anymore. The current document page though still shows an example of the same. 

* UCP `3.0.1` Admin Settings
![image](https://user-images.githubusercontent.com/20244490/48493221-bb650e00-e82b-11e8-95c4-96547c0ed162.png)

* Document Page
![image](https://user-images.githubusercontent.com/20244490/48493304-dd5e9080-e82b-11e8-8926-2d820f9c0293.png)

Secondly,
The ELK containers don't support latest tags. The example doesn't seem to take that into consideration.

### Problem location

Document Page: [store-logs-in-an-external-system](https://docs.docker.com/ee/ucp/admin/configure/store-logs-in-an-external-system/)

### Project version(s) affected

UCP: 3.0.1

### Suggestions for a fix

Possibly just need to remove the example.

prashshr:
Relates to https://github.com/docker/orca/issues/14915

bermudezmt:
Hi @prashshr . This has been deprecated as of UCP 3.1. @ollypom took care of this via #8071 . Thanks for bringing it up!

PR #8071, opened 2019-01-24
Removing Syslog UI from UCP Docs

ollypom:
### Proposed changes

I have actually removed the "Configure Logging in UCP" page from the docs. I know we don't like removing pages from Docs, however this Syslog UI feature was EOL'd in UCP 3.1. At the same time the ELK stack documentation was stale as we have moved this work across to Solution Briefs. https://success.docker.com/solutions. 

Mentioned in:

https://github.com/docker/docker.github.io/issues/7940 
https://github.com/docker/docker.github.io/issues/7954
https://github.com/docker/docker.github.io/issues/7682


L-Hudson:
Good to go

ahh-docker:
thanks @ollypom! :)

bermudezmt:
Discussed with @ollypom offline - we may need to add a redirect within success.docker.com for people looking for the old Syslog doc. The closest article we have currently is https://success.docker.com/article/elasticsearch-logstash-kibana-logging. cc: @JustinINevill

PR #8077, opened 2019-01-25
Added User Permission note to DTR Promotions

ollypom:
### Proposed changes
After a bit of feedback from customers, I have added a User Permission note to DTR Promotions page.


L-Hudson:
Good to go. Thanks for the addition!

PR #8079, opened 2019-01-25
Fix missing .0 in engine version release notes

thaJeztah:
The version is 18.09.0, but I wasn't sure if we have links to this anchor, so decided to push it as a separate PR

ping @ahh-docker


L-Hudson:
good to go

PR #8081, opened 2019-01-25
Update Kubernetes and Calico version #s

darwin-traver:
### Proposed changes
As per UCP 3.1.x, Calico was updated to 3.2.3
As per UCP 3.1.1, Kubernetes version was updated to 1.11.5.

### Related issues (optional)
https://github.com/docker/orca/pull/15472
https://docs.docker.com/ee/ucp/release-notes/#311


L-Hudson:
Good to go -- thanks for the catch!

PR #8084, opened 2019-01-28
Update collect-cluster-metrics.md

gabrielmcg:
Port forwarding is used to expose the UI, not the API?

### Proposed changes
Port forwarding is used to expose the UI, not the API?


L-Hudson:
Thanks for the catch!

PR #8049, opened 2019-01-18
Updating Docker Content Trust Documentation

ollypom:
### Proposed changes

Here I am updating the Content Trust and the Automating Content Trust pages to take into account the latest `docker trust` commands, and the latest runtime enforcement in the Enterprise engine. 

I have also removed some of the content around Client side enforcement.

This PR also address https://github.com/docker/docker.github.io/issues/7709

PR #8086, opened 2019-01-28
DCT Plugin is configured with Enforced not Enabled

ollypom:
### Proposed changes

Mistake on my part in https://github.com/docker/docker.github.io/pull/8049, the DCT plugin is configured with an "enforced" variable not "enabled".


L-Hudson:
Oops! Thanks for the catch!

PR #8089, opened 2019-01-28
update install instructions of docker engine to include containerd

andrewhsu:
Update instructions for installing and updating the 18.09 engine to include `containerd.io` package so the underlying container runtime package is latest and greatest.


_includes/ee-linux-install-reuse.md
```diff
@@ -145,7 +145,7 @@ You only need to set up the repository once, after which you can install Docker
 1.  Install the latest patch release, or go to the next step to install a specific version:
 
     ```bash
-    $ sudo yum -y install docker-ee
+    $ sudo yum -y install docker-ee containerd.io
```
seemethere: It's not technically necessary to install `containerd.io` when initially installing `docker-ee` since `docker-ee` will automatically install `containerd.io` as a dependency (just like it does with `docker-ee-cli`.)

thaJeztah: I think it's better to keep it symmetrical; mentioning it here makes people more aware that there's multiple packages involved.

I think we should actually include `docker-ee-cli` here as well

ahh-docker: I agree with @thaJeztah on this. Keep it symmetrical. Right now, the only install I have is docker-ee. If they need to install all three packages explicitly (`docker-ee`,containerd.io`, and `docker-ee-cli`, we need to state it.

thaJeztah: During _install_ it's not strictly needed (which is what @seemethere tried to explain), but mentioning it here will make it less of a surprise if we later on tell them that for _updating_ they need to update them all.

So for that reason I think mentioning all packages here helps. (but, my €0.02)

ahh-docker: @thaJeztah Maybe we need a note that you might need to upgrade packages separately, and that the install consists of three packages. For the refactor, I have that you need to check to see the Docker service and containerd are running, but I don't break it down into the three packages. Is that necessary for the install now?

seemethere: Yeah I think it should be an all or nothing, either we mention the `docker-ee-cli`, `docker-ee`, and `containerd.io` packages altogether, or we just mention `docker-ee`.

I'm for mentioning all of `docker-ee-cli`, `docker-ee`, and `containerd.io`.

andrewhsu: i'll update the PR to include all 3 packages

andrewhsu:
@seemethere @thaJeztah @ahh-docker I've force-pushed an update to this PR. Ready for review.

ahh-docker:
LGTM. Good to merge provided @L-Hudson says it's good.

PR #8090, opened 2019-01-28
ucp: add "heartbeat failure" node state

trapier:
Also:
- Align table columns
- Alpha sort table by symptom

(text content of other rows in the table not modified)

PLMK if the `Also`s should be split into separate commits to make this easier to review.


trapier:
This PR for @gonzo919, on the request line.

L-Hudson:
> Also:
> 
> * Align table columns
> * Alpha sort table by symptom
> 
> (text content of other rows in the table not modified)
> 
> PLMK if the `Also`s should be split into separate commits to make this easier to review.

totally fine to have had those in the same commits.

PR #8091, opened 2019-01-29
Adding wordpress environment variable.

td-gonzales:
Wordpress needs to know what db to it's user has access to. Without this mysql throws access denied errors.


L-Hudson:
thanks for the add!

PR #8092, opened 2019-01-29
Added docker-compose for oh-my-zsh autocomplete

TheRijn:
### Proposed changes

Added instruction for docker-compose autocompletion in oh-my-zsh

### DIFF
--- a/compose/completion.md
+++ b/compose/completion.md
@@ -54,11 +54,11 @@ completion.
 
 #### With oh-my-zsh
 
-Add `docker` to the plugins list in `~/.zshrc`:
+Add `docker` and/or `docker-compose` to the plugins list in `~/.zshrc`:
 
 ```shell
 plugins=(
-  <existing-plugins> docker
+  <existing-plugins> docker docker-compose
 )
 ```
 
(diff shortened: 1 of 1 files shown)


bermudezmt:
@paigehargrave Can you please review? I tested this for Ubuntu but not on a Mac.

paigehargrave:
@bermudezmt - We tried this on my machine, but I think there might be prerequisite steps that need to be performed. Let me see if I can find someone in support to check this on a Mac.

compose/completion.md
```diff
@@ -13,77 +13,84 @@ for the bash and zsh shell.
...
+```
 
-Place the completion script in `/etc/bash_completion.d/`
-(or `/usr/local/etc/bash_completion.d/` on a Mac):
+### Mac
+
+##### Install via Homebrew
+
+1. Install with `brew install bash-completion`.
+2. Place the completion script in `/usr/local/etc/bash_completion.d/`.
```
jose-bigio: I think step `2` is describing the `curl` command below. 

There seems to be an inconsistency with what step 2 describes and what the command does. 
Step `2` says to place the completion script in `/usr/local/etc/bash_completion.d/` but the `curl` command has 

`-o /etc/bash_completition.d/docker-compose`. 

So it isn't clear what is intended.

usha-mandya: Hi @jose-bigio, we have resolved the inconsistencies you have pointed out by testing the procedure on Mac.

PR #8105, opened 2019-01-30
Remove leading whitespace in code blocks from Django docs

danielcompton:
### Proposed changes

Some of the code blocks in the Django example documentation had extra leading whitespace. This PR removes it to make it easier to copy/paste.


bermudezmt:
Thank you, @danielcompton !

PR #8112, opened 2019-01-30
Updated UCP Audit Logging

ollypom:
### Proposed changes

From UCP 3.1.2 onwards we support UCP audit logging with in the UCP configuration file and the UCP UI. In this PR I have updated the documentation accordingly, including some examples on how retrieve the ucp audit logs.

Finally I have updated the release notes for UCP 3.1.2


ollypom:
@david-yu PTAL. 

Also I have removed the current the documentation which tells customers to use the `/api/ucp/config/logging` endpoint. In favour of using the UCP UI and the UCP Configuration file. Is that correct as the `/api/ucp/config/logging` endpoint is still available??

david-yu:
This looks great. Yes the endpoint is still availabler @ollypom there just now a UI on top of it. So you can use either the CLI or UI to enable it.

david-yu:
@ollypom Could you put the turning on of the Audit Logs via API back into the docs?

bermudezmt:
Thank you, @david-yu ! Can you please do your edit magic, @L-Hudson ? :)

ee/ucp/admin/configure/create-audit-logs.md
```diff
@@ -4,64 +4,115 @@ description: Learn how to create audit logs of all activity in UCP
...
+3) Select **Audit Logs**
 
-1. Download the UCP Client bundle [Download client bundle from the command line](https://success.docker.com/article/download-client-bundle-from-the-cli).
+4) In the **Configure Audit Log Level** section, select the relevant logging
+level. 
 
-2.  Retrieve JSON for current audit log configuration.
+![Enabling Audit Logging in UCP](../../images/auditlogging.png){: .with-border}
 
-    ```
```
david-yu: We should still leave this in actually as an alternative to doing this over the CLI

ee/ucp/admin/configure/create-audit-logs.md
```diff
@@ -4,64 +4,115 @@ description: Learn how to create audit logs of all activity in UCP
...
-    }
-    ```
+Enabling UCP audit logging via the UCP Configuration file can be done before 
+or after a UCP installation. Following the UCP Configuration file documentation 
+[here](../ucp-configuration-file/).
 
-4. Send the JSON request for the auditlog config with the same API path but with the `PUT` method.
+The section of the UCP configuration file that controls UCP auditing logging is:
 
-    ```
```
david-yu: We should still leave this in actually as an alternative to doing this over the CLI

david-yu:
@bermudezmt I requested changes on this PR so changing from approved to requesting changes.

ollypom:
@david-yu Re-added API configuration in the latest commit.

PR #8115, opened 2019-01-30
Update Ubuntu installation instructions with latest key fingerprint output

danielcompton:
### Proposed changes

Using Ubuntu 18.04, the key fingerprint output format has changed. This updates the output to match what users are likely to see when checking the fingerprint. All the same information is there, but this will be easier to visually parse.

### DIFF
--- a/install/linux/docker-ce/ubuntu.md
+++ b/install/linux/docker-ce/ubuntu.md
@@ -120,6 +120,7 @@ from the repository.
 
     ```bash
     $ sudo apt-key fingerprint 0EBFCD88
+    
     pub   rsa4096 2017-02-22 [SCEA]
           9DC8 5822 9FC7 DD38 854A  E2D8 8D81 803C 0EBF CD88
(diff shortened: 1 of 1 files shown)


ahh-docker:
@thaJeztah PTAL. I don't see any issues with this, and if you're good with it, I'll go ahead and merge.

install/linux/docker-ce/ubuntu.md
```diff
@@ -120,11 +120,10 @@ from the repository.
 
     ```bash
     $ sudo apt-key fingerprint 0EBFCD88
-
```
thaJeztah: I think the empty line (although not actually there when you run the command) was added in the documentation to better separate the command to run, and the output it produces.

Would be good to add it back, but definitely not a show-stopper 🤗

ahh-docker: Agreed. I'll add it back and merge.

PR #8116, opened 2019-01-31
Note extending individual services only works up to version 2.1

danielcompton:
### Proposed changes

Adds a note that the `extends` field only works up to version 2.1 at the header of the page. I think it would be better to tell people this up-front than to have them click on the link, only for it to be unavailable for them.


L-Hudson:
Good catch, thanks!

PR #8122, opened 2019-01-31
Removing a link to the UCP WS1809 offline bundle

ollypom:
### Proposed changes

I've removed a link to the UCP WS1809 offline bundle as I don't believe we have a separate bundle for this. 

```
$ docker manifest inspect docker/ucp-agent-win:3.1.3 | jq -r '.manifests |.[].platform."os.version"' | sort
10.0.14393.2551 # WS2016 LTSC / WS1607 SAC
10.0.16299.904  # WS1709 SAC
10.0.17134.523  # WS1803 SAC
10.0.17763.253  # WS2019 LTSC / WS1809 SAC
```

The 10.0.17763.253 images will work fine for both Win Server 2019 LTSC and Win Server 1809 SAC.

### DIFF
--- a/_data/ddc_offline_files_2.yaml
+++ b/_data/ddc_offline_files_2.yaml
@@ -6,10 +6,6 @@
 - product: "ucp"
   version: "3.1"
   tar-files: 
-    - description: "3.1.3 Windows Server 2019"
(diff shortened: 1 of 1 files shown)


ahh-docker:
@ry4nz  PTAL. Thanks.

PR #8131, opened 2019-02-02
Updated footer year

pvsmounish:
Updated footer year from 2018 to 2019.


L-Hudson:
great catch, thank you!

PR #8137, opened 2019-02-04
Update part2.md

arhaasdev:
### Proposed changes

Eliminates the "the" from "The second item is the Google's DNS [...]".

I did this because I think that speaking of "the Google" in documentation is overly colloquial at best.

### Related issues (optional)


L-Hudson:
great catch, thank you!

PR #8149, opened 2019-02-05
Remove non-sense angle brackets.

wi1dcard:
### Proposed changes

Remove non-sense angle brackets.


L-Hudson:
thanks for the catch!

PR #8155, opened 2019-02-06
interlock: advise host mode on ip hashing

trapier:
Noticed references to *client* IP address in ip hashing docs, which called to mind https://github.com/moby/moby/issues/25526. Verified [nginx docs](https://nginx.org/en/docs/http/ngx_http_upstream_module.html#ip_hash) agree with respect to hashing based on client ip.  Host mode networking (already documented) addresses the concern.


L-Hudson:
thank you!

PR #8163, opened 2019-02-06
Interlock vip backend mode

trapier:
Introduce Interlock VIP backend mode.


bermudezmt:
This has been reviewed and merged in an internal repo.

PR #8165, opened 2019-02-06
Fix info about the location of the search box in the Console app

onmishkin:
### Proposed changes

Fix info about the location of the search box in the Console app


paigehargrave:
@bermudezmt - LGTM

PR #8173, opened 2019-02-07
Update description of private collection

bermudezmt:
Closes issue opened within private repo.


bermudezmt:
@david-yu I used your note word-for-word as a starting point because I wasn't sure what triggers the creation of a private collection. Is it a system-initiated event triggered by a newly created user logging in for the first time?

PR #8174, opened 2019-02-07
Specify that mac_address is a legacy option

bermudezmt:
Adds a note on `mac_address` being a legacy option.


bermudezmt:
Thank you, @seemethere.

PR #8176, opened 2019-02-07
Update part2.md

Ekwinder:
Specified where to create an empty directory while adding the Dockerfile.


L-Hudson:
Thanks for the change!

PR #8177, opened 2019-02-07
Update gettingstarted.md

zeexxu:
### Proposed changes
suggest to remove `ADD . /code` after adding the volume as it becomes irrelevant.

### DIFF
--- a/compose/gettingstarted.md
+++ b/compose/gettingstarted.md
@@ -225,6 +225,7 @@ Edit `docker-compose.yml` in your project directory to add a [bind mount](/engin
 The new `volumes` key mounts the project directory (current directory) on the
 host to `/code` inside the container, allowing you to modify the code on the
 fly, without having to rebuild the image.
+(At this point you may want to remove `ADD . /code` from the created Dockerfile)
 
 ## Step 6: Re-build and run the app with Compose
 


L-Hudson:
@thaJeztah PTAL, thanks!

compose/gettingstarted.md
```diff
@@ -225,6 +225,7 @@ Edit `docker-compose.yml` in your project directory to add a [bind mount](/engin
 The new `volumes` key mounts the project directory (current directory) on the
 host to `/code` inside the container, allowing you to modify the code on the
 fly, without having to rebuild the image.
+(At this point you may want to remove `ADD . /code` from the created Dockerfile)
```
thaJeztah: Hm, no, we should not recommend doing this; the code that's copied into the actual image that is built is what's used when deploying the image/container to production.

Bind-mounting source-code is something that can help during development of your project; instead of having to rebuild your container image for each code-change, you can make edits on your host, and (for example, if you develop in a scripting language that doesn't require re-compiling), you can refresh your browser, and see the changes.

However, bind-mounting the source code for this purpose should really only be used for development. Using it in "production" is roughly the equivalent to "logging into the production server, and making ad-hoc code changes".

Looking at this page however, I think we can add more explanation to that; perhaps even show the use of multiple compose files to separate the `development` and `production` situation; in the production situation there would be no bind-mount for the source code, and for development, an additional "override" compose-file could be used to bind-mount the source-code during development.


Also, there's some improvements to be made in the example;

1. prefer the use of `COPY` over `ADD` if you don't need the "magic features" of `ADD` (automatic extraction of compressed archives,  or adding files from remote sources)

2. Be specific what to copy to optimize use of the build-cache

In the example, `ADD` copies the entire code, but initially only needs the `requirements.txt` to install dependencies. However, because _all_ code is copied, any change to the code will trigger a full rebuild of the container (and re-installing the dependencies), even if no changes were made to the dependencies.

We can also swap the order of the `WORKDIR` to simplify this step; with those changes, the Dockerfile may be better written as;

```Dockerfile
FROM python:3.4-alpine
WORKDIR /code
COPY ./requirements.txt .
RUN pip install -r requirements.txt
CMD ["python", "app.py"]
COPY . .
```

In the above, the first `COPY` only copies the `requirements.txt`, so `pip install ...` is only run if a requirement changed. Because of that, all steps will be cached, except for the `COPY . . ` (which copies the source-code into the image).

L-Hudson: Thanks @thaJeztah. Moving this ticket back over into backlog.

usha-mandya:
Closing the PR based on the comment above. We can look into improving the Compose getting started as a new work item.

PR #8121, opened 2019-01-31
Rename resin to balena and fix 404 link

JackuB:
### Proposed changes

Found a 404 link because [resin.io](https://resin.io) changed name to [balena.io](https://balena.io) - also updated the code example to a newer version.

PR #8179, opened 2019-02-07
Update balena link

mikesimos:
### Proposed changes
Update balena link

### DIFF
--- a/docker-for-mac/multi-arch.md
+++ b/docker-for-mac/multi-arch.md
@@ -15,7 +15,7 @@ This does not require any special configuration in the container itself as it us
 <a href="http://wiki.qemu.org/" target="_blank">qemu-static</a> from the Docker for
 Mac VM.
 
-You can run an ARM container, like the <a href="https://www.balena.io/blog/how-resin-io-works/" target="_blank">
+You can run an ARM container, like the <a href="https://www.balena.io/what-is-balena/" target="_blank">
 balena</a> arm builds:
 
 ```
--- a/docker-for-mac/multi-arch.md
+++ b/docker-for-mac/multi-arch.md
@@ -6,29 +6,51 @@ redirect_from:
 title: Leverage multi-CPU architecture support
 notoc: true
 ---
+Docker images can support multiple architectures, which means that a single
+image may contain variants for different architectures, and sometimes for different
+operating systems, such as Windows.
 
-Docker Desktop for Mac provides `binfmt_misc` multi architecture support, so you can run
-containers for different Linux architectures, such as `arm`, `mips`, `ppc64le`,
-and even `s390x`.
+When running an image with multi-architecture support, `docker` will
+automatically select an image variant which matches your OS and architecture.
+
+Most of the official images on Docker Hub provide a [variety of architectures](https://github.com/docker-library/official-images#architectures-other-than-amd64).
+For example, the `busybox` image supports `amd64`, `arm32v5`, `arm32v6`,
+`arm32v7`, `arm64v8`, `i386`, `ppc64le`, and `s390x`. When running this image
+on an `x86_64` / `amd64` machine, the `x86_64` variant will be pulled and run,
+which can be seen from the output of the `uname -a` command that's run inside
+the container:
+
+```bash
+$ docker run busybox uname -a
+
+Linux 82ef1a0c07a2 4.9.125-linuxkit #1 SMP Fri Sep 7 08:20:28 UTC 2018 x86_64 GNU/Linux
+```
+
+**Docker Desktop for Mac** provides `binfmt_misc` multi-architecture support,
+which means you can run containers for different Linux architectures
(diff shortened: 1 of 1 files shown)


docker-for-mac/multi-arch.md
```diff
@@ -15,7 +15,7 @@ This does not require any special configuration in the container itself as it us
 <a href="http://wiki.qemu.org/" target="_blank">qemu-static</a> from the Docker for
 Mac VM.
 
-You can run an ARM container, like the <a href="https://www.balena.io/blog/how-resin-io-works/" target="_blank">
+You can run an ARM container, like the <a href="https://www.balena.io/what-is-balena/" target="_blank">
```
thaJeztah: Given that these are just to illustrate that you can use a different architecture, I'm wondering if we should use a different image to illustrate this (i.e., I don't think there should be a need to explain what Balena is to show this functionality. Perhaps use [`arm32v7/hello-world`](https://hub.docker.com/r/arm32v7/hello-world) and [`ppc64le/hello-world`](https://hub.docker.com/r/ppc64le/hello-world) instead.

Or (if we want to show the `uname -a`) we can use the busybox images;


```
docker run --rm arm32v7/busybox uname -a  
Linux 9e3873123d09 4.9.125-linuxkit #1 SMP Fri Sep 7 08:20:28 UTC 2018 armv7l GNU/Linux

docker run --rm ppc64le/busybox uname -a
Linux 57a073cc4f10 4.9.125-linuxkit #1 SMP Fri Sep 7 08:20:28 UTC 2018 ppc64le GNU/Linux
```

bermudezmt: @thaJeztah we thought we addressed the Balena link issue via https://github.com/docker/docker.github.io/pull/8121. Is the point to avoid referring to it at all and use a different image in the example `docker run` commands?

thaJeztah: > Is the point to avoid referring to it at all and use a different image in the example `docker run` commands?

Yes; I think linking to the Balena project only distracts from what we're trying to document here (it'd be similar to showing that you can run an NGINX container, and linking to the "what is NGINX" documentation).

So, instead, we could use something that's more focused on what we're trying to illustrate here (running an image that's using an architecture that doesn't match the native architecture of your machine). Busybox is fairly common (and already used in our documentation), so likely doesn't require any explanation.

We _could_ explain the output of the `uname -a` Linux command that is used in the example (point to the output showing `armv7l` or `ppc64le`).

Perhaps it should start with explaining that many of the official images are "multi arch", and that Docker will run the variant that matches the host's architecture; something like:


    Docker images can support multiple architectures, which means that a single
    image contains variants for different architectures (and sometimes
    different operating systems, such as Windows).

    When running an image that provides multiple architectures, docker will
    automatically select the variant of the image that matches your operating
    system and architecture.

    Most of the official images on Docker Hub provide a [variety of architectures](https://github.com/docker-library/official-images#architectures-other-than-amd64).
    For example, the `busybox` image supports `amd64`, `arm32v5`, `arm32v6`,
    `arm32v7`, `arm64v8`, `i386`, `ppc64le`, and `s390x`. When running this image
    on an `x86_64` / `amd64` machine, the `x86_64` variant will be pulled and run,
    which can be seen from the output of the `uname -a` command that's run inside
    the container;

    ```bash
    $ docker run busybox uname -a

    Linux 82ef1a0c07a2 4.9.125-linuxkit #1 SMP Fri Sep 7 08:20:28 UTC 2018 x86_64 GNU/Linux
    ```

    Docker Desktop for Mac provides `binfmt_misc` multi architecture support,
    which means so you can run containers for different Linux architectures,
    such as `arm`, `mips`, `ppc64le`, and even `s390x`.

    This does not require any special configuration in the container itself as it uses
    <a href="http://wiki.qemu.org/" target="_blank">qemu-static</a> from the Docker for
    Mac VM.

    Because of this, you can run an ARM container, like the `arm32v7` or `ppc64le`
    variants of the busybox image:

    ```bash
    $ docker run arm32v7/busybox uname -a  

    Linux 9e3873123d09 4.9.125-linuxkit #1 SMP Fri Sep 7 08:20:28 UTC 2018 armv7l GNU/Linux

    $ docker run ppc64le/busybox uname -a

    Linux 57a073cc4f10 4.9.125-linuxkit #1 SMP Fri Sep 7 08:20:28 UTC 2018 ppc64le GNU/Linux
    ```

    Notice that this time, the `uname -a` output respectively shows `armv7l` and
    `ppc64le`.

bermudezmt: Very educational yet again, @thaJeztah . I've incorporated the changes to the page.

bermudezmt:
@paigehargrave hi Paige, can you give this a final look? :)

thaJeztah:
Thanks @bermudezmt @mikesimos - LGTM

PR #8181, opened 2019-02-07
interlock does not ssl_only

trapier:
`com.docker.lb.ssl_only` is not recognized by Interlock


L-Hudson:
Good to go. Thanks, @trapier

PR #8182, opened 2019-02-07
Updated as certificate  as expire in  documented version

jondkent:
### Proposed changes

The current  documentation points to use a version of notary which has an expired certificate. 

Updated the version of notary_autobuilds to 0.5.1 from 0.4.2.

Change the shared volume to notarycerts:/var/lib/notary/fixtures from notarycerts:/go/src/github.com/docker/notary/fixtures  as location of certificate has changed.

### Related issues (optional)

Fixes issue I raised in https://github.com/theupdateframework/notary/issues/1417


L-Hudson:
Good to go; thanks for the contribution!

Issue #8180, opened 2019-02-07
UCP 3.0 documents how to use HRM


trapier:
### Problem description

https://docs.docker.com/datacenter/ucp/3.0/guides/user/services/use-domain-names-to-access-services/

^ documents HRM, which is not available in UCP 3.0.  UCP 3.0 shipped with interlock which uses different network names and service labels and web ui and bits.

### Suggested solution

Backport interlock docs to UCP 3.0, removing notes on any features not available in UCP 3.0 Interlock.  I'm not aware of any, but it's worth noting that it's a backport.

trapier:
As suspected on #8192, something is amiss here:

```
docker.github.io]$ git grep -ic kubernetes datacenter/ucp/3.0/
datacenter/ucp/3.0/guides/admin/install/install-on-azure.md:15
datacenter/ucp/3.0/reference/api/main.js:1
datacenter/ucp/3.0/reference/cli/install.md:3
datacenter/ucp/3.0/reference/cli/upgrade.md:3
docker.github.io]$ 
```

bermudezmt:
@trapier we are looking into this.

JustinINevill:
@bermudezmt to:
1. Remove HRM from 3.0
2. Backport interlocks docs to 3.0

PR #8192, opened 2019-02-08
ucp port requirements: add calico health

trapier:


trapier:
Would like to have copied this to the UCP 3.0 install requirements [1], but they seem to be missing all traces of kubernetes, which makes me wonder if https://github.com/docker/docker.github.io/issues/8180 is systemic, i.e. if current UCP 3.0 docs are instead a copy of UCP 2.2 docs.

[1] https://docs.docker.com/datacenter/ucp/3.0/guides/admin/install/system-requirements/#ports-used

L-Hudson:
Good to go from my perspective. @trapier do you want this merged now?

PR #8206, opened 2019-02-09
Fix typo: docker_compose.yml

sergey-bon:


sergey-bon:
@thaJeztah Oh boy, you were quicker than it took to clone the repo.

PR #8207, opened 2019-02-09
Add an example json-file log line to the documentation.

toph:
### Proposed changes

Was very difficult to find a reference for what JSON keys were in the json-file logging driver output. Added a simple example to the doc.

I copied the example line from another site and tweaked it slightly, so would be good for reviewer to confirm format is correct.

### DIFF
--- a/config/containers/logging/json-file.md
+++ b/config/containers/logging/json-file.md
@@ -13,6 +13,10 @@ and writes them in files using the JSON format. The JSON format annotates each l
 origin (`stdout` or `stderr`) and its timestamp. Each log file contains information about
 only one container.
 
+```json
+{"log":"Log line is here\n","stream":"stdout","time":"2019-01-01T11:11:11.111111111Z"}
+```
+
 ## Usage
 
 To use the `json-file` driver as the default logging driver, set the `log-driver`


L-Hudson:
@zaggash PTAL, or let me know who should be reviewing this. Thanks!

zaggash:
Looks good.
Should we add the location of the files somewhere ?

/var/lib/docker/containers/<containerID>/<containerID>-json.log

L-Hudson:
good to go

thaJeztah:
oh, I saw this come by my notifications, but forgot to comment. I'm a bit on the fence for this change, as the files created by the `json-file` logging driver are the "internal storage mechanism" of that driver, and used to facilitate running `docker logs` (which reads from those files). Those files are not designed for consumption by other tools (and accessing them with external tools could interfere with docker, e.g. if docker is rotating those files)

Being an internal storage mechanism, their format can therefore change at any time (which is why the format was not documented in the documentation, as the file-format is not a spec/public api).

@toph Was there a specific reason for you to access those files?

toph:
@thaJeztah I was receiving the logs with structure preserved by FluentD.

thaJeztah:
@tobh is that with the build-int fluentd logging-driver? (https://docs.docker.com/config/containers/logging/fluentd/), or (e.g.) using `logspout`? (because the built-int fluentd logging driver does not use the files that are written by the json-file driver).

toph:
Looks like our ops team has fluentd configured to forward disk logs from
/var/log/containers/.  The built in driver looks nicer.  Thank you for the
pointer!

As for the patch--totally up to you guys.  The current docs actually
discuss some details JSON format and some of the expected values (stdout,
stderr), but are currently silent on the key names.

Toph

toph:
Specifically we are using:

https://github.com/fluent/fluentd-kubernetes-daemonset

Issue #8191, opened 2019-02-08
NO_PROXY does not support CIDR notation.


ryujisnote:
[config/daemon/systemd.md](https://docs.docker.com/config/daemon/systemd/)

NO_PROXY does not support CIDR notation.
Please modify the document.
`IP address prefix (1.2.3.4) or in CIDR notation (1.2.3.4/8)` -> `IP address prefix (1.2.3.4)` (deleted)

- CentOS 7.6
- Docker EE 18.09.1 from yum install.
- 3 initialized manager node.

write NO_PROXY on /etc/systemd/system/docker.service.d/http-proxy.conf
```
[Service]    
Environment="HTTP_PROXY=http://proxy.example.com:80/" "NO_PROXY=localhost,127.0.0.1,192.168.0.0/24"
```

I done try `docker swarm join` from **manager node**.
ex. 192.168.0.0/24 -> doesn't work. (blocked error message from proxy content filter)
ex. 192.168.0.2 -> works properly.

refer: https://unix.stackexchange.com/questions/23452/set-a-network-range-in-the-no-proxy-environment-variable

thaJeztah:
Ah, good catch. The description was updated in https://github.com/docker/docker.github.io/pull/8061, where it was taken from the Go description itself.

However, CIDR notation was added in Go 1.11 (support was added through https://github.com/golang/net/commit/c21de06aaf072cea07f3a65d6970e5c7d8b6cd6d, and https://github.com/golang/go/commit/4da84adc0c55d92f20b761d24a6ec0508e55dc7c), but the current docker release uses Go 1.10, which uses the old implementation, and does not yet support it; https://github.com/golang/go/blob/go1.10.6/src/net/http/transport.go#L1261-L1313

thaJeztah:
The next Docker Engine release (19.03) will be using Go 1.11 (or above), so will support it; perhaps we should open a tracking issue for that to amend the docs after the next release @L-Hudson  ?

ryujisnote:
I think, you should specify that older version software does not guarantee to CIDR.
If possible, I think that we should consider all of the past versions.

ryujisnote:
Fixed in [moby/moby/issue#36951](https://github.com/moby/moby/issues/36951#issuecomment-477825346)
There is no need to write the manager node's IP to NO_PROXY.

PR #8217, opened 2019-02-11
NO_PROXY does not support CIDR notation.

ryujisnote:
NO_PROXY does not support CIDR notation. #8191

### Proposed changes

Delete CIDR notation.

### Related issues (optional)

NO_PROXY does not support CIDR notation. #8191


paigehargrave:
@thaJeztah Thank you!

PR #8221, opened 2019-02-12
Update Windows EE Engine Install Docs

ollypom:
### Proposed changes

I have removed a few dead links, tried to provide a bit of clarity to the pre-requisites, but mostly just reformatted chunks of the text to follow the 80 character limit we use else where in the docs. 

Finally following the 18.09.2 engine release yesterday (11th Feb), i've updated the Win variable in the config file.

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -94,7 +94,7 @@ defaults:
   - scope:
       path: "install"
     values:
-      win_latest_build: "docker-18.09.1"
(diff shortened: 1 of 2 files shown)


bermudezmt:
Thank you, @ollypom . FYI, it looks like the link to our Windows Server 2016 labs is currently broken. I've removed the link for now.

PR #8223, opened 2019-02-12
Add --skip-cloud-provider option for 3.0 and 3.1

paigehargrave:
Hi @david-yu and @wsong - PTAL and make sure I've added the correct information in the correct locations - thank you!!


paigehargrave:
@stevenfollis - PTAL and verify that I've added everything from https://github.com/docker/docs-private/pull/956/ - thank you!

stevenfollis:
Hi @paigehargrave thanks for putting this together!. 

Looks like a formatting issue at `/reference/ucp/3.1/cli/install/` with `--unmanaged-cli` not wrapping like the others?

image

Also [I think](https://github.com/docker/docs-private/issues/814#issuecomment-462541389) that the 3.0 docs should also include the additional instruction?

Probably a separate issue but should these be alphabetized? Wasn't sure of the ordering, as it appears to begin alphabetized and then stops.

paigehargrave:
@stevenfollis - Thank you for catching the spacing error - it looked fine locally for me. I have alphabetized the table, so PTAL. Also, I thought I had all the new commands where, so please let me know what you think is missing. Many thanks again for your time here!

paigehargrave:
@stevenfollis - Forgot to mention - there appears to be a duplicate entry at the bottom (the last entry). Which one is correct, or should I merge them?

stevenfollis:
For the duplicative `--unmanaged-cli` that's a good question. The [CLI uses](https://github.com/docker/orca/blob/71965012cb93f3e1e0af8408eb5daccd3848adfe/bootstrap/main.go#L116-L119) the shorter version (would need capitalization), but the more verbose version sounds nice. Not sure what standard we use - verbatim CLI or case-by-case.

For 3.0 should we match the alphabetization? 
https://deploy-preview-8223--docsdocker.netlify.com/datacenter/ucp/3.0/reference/cli/install/

david-yu:
@mark-church Can you help here in terms of which description we want in terms of `--unmanaged-cli`? It looks like it shows up twice on the ucp install command

stevenfollis:
(Copying/pasting the two descriptions for ease of discussion) -

image

mark-church:
The second phrasing is a little clearer. I'll just reword it ...

>Defaults to `false` and indicates that the CNI plugin for Kubernetes networking is managed by UCP. The default managed CNI plugin is Calico. When set to `true` UCP will not deploy or manage the lifecycle of a default CNI plugin. The CNI plugin would be deployed and managed independently of UCP. Note that when `unmanaged-cni=true` networking in the cluster will not function for Kubernetes until a CNI plugin is deployed.

paigehargrave:
@mark-church - Thank you! I've added your wording (with a few tweaks).

PR #8225, opened 2019-02-12
Fixed list indentations

millerdz:
### Proposed changes

List indentations were incorrect yielding the list in one line instead of multiple lines. Fixed the indentation so the lists render correctly.


L-Hudson:
Great catch, thanks!

PR #8239, opened 2019-02-13
Updated UCP rethinkDB default cache size to 1GB

darwin-traver:
### Proposed changes

With newer releases of UCP, the rethinkDB default cache size was upped to 1GB  as per:
https://github.com/docker/orca/commit/06baa587159072e7b1aeb6601095ff636aa2806e


bermudezmt:
Thank you, @darwin-traver ! Really helps to have a cross-reference like the commit you've linked to.

darwin-traver:
@bermudezmt I always want to link back to the change PR (so to have someone else review what I am seeing), and so we can know when the version cut-off for the change took place. Much thanks for the quick merge!

PR #8254, opened 2019-02-14
Add a reference to the `local` log-driver

russellballestrini:
### Proposed changes

This log driver isn't documented, so I'm giving it a first pass.


L-Hudson:
thanks for the add!

PR #8288, opened 2019-02-17
Extra regions

SychevIgor:
### Proposed changes
Microsoft added more regions, but 4 regions not mentioned in this guide.
Please, add 4 azure regions to a list

You can test it via azure cli command
az account list-locations --query "[].name"

### DIFF
--- a/docker-for-azure/faqs.md
+++ b/docker-for-azure/faqs.md
@@ -90,6 +90,8 @@ All regions can be found here: [Microsoft Azure Regions](https://azure.microsoft
 An excerpt of the above regions to use when you create your service principal are:
 
 ```none
+australiacentral
+australiacentral2
 australiaeast
 australiasoutheast
 brazilsouth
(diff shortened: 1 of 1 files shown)


docker-for-azure/faqs.md
```diff
@@ -118,4 +118,8 @@ westeurope
 westindia
 westus
 westus2
+australiaeast
+australiasoutheast
```
thaJeztah: Looks like the two australia regions are already listed at the top

docker-for-azure/faqs.md
```diff
@@ -118,4 +118,8 @@ westeurope
...
+francecentral
+francesouth
```
thaJeztah: Could you keep the list sorted alphabetically?

SychevIgor: It's super complicated task, but I did it. :)

P.S. I don't have access to Azure in China or to US Gov/DoD, so I can't add all regions

PR #8290, opened 2019-02-17
update Backup.md

dockerpac:
- Update DTR backups and restore to point to actual backup and restore page, and not root "disaster recovery" page.
- Update Docker Engine --> Docker swarm in the Restore section to be same labels as the Backup section


L-Hudson:
Good to go!

Issue #7828, opened 2018-12-10
Node RBAC instructions don't seem to be updated to EE 2.1/UCP 3.1


nicolaka:
The instructions listed are for EE 2.0 not latest 2.1 version. Current docs don't work in UCP 3.1.

[ee/ucp/authorization/isolate-nodes.md](https://docs.docker.com/ee/ucp/authorization/isolate-nodes/)

bermudezmt:
@nicolaka what isn't covered in the 3.1 version of the page? Can you please give us some more details?

nicolaka:
@bermudezmt the second part of that page is basically not up-to-date with the new introduction of the Kubernetes-native RBAC in 3.1. Along with the images associated with it ( it's not just the images btw, it's foundational not application for anyone using version 3.1 or newer). 

The correct method to associating a namespace to nodes is in the namespace definition itself. The info is provided here : https://kubernetes.io/docs/reference/access-authn-authz/admission-controllers/#podnodeselector.

Happy to help if you have more questions.

paigehargrave:
@nicolaka  Thank you for the information! We are working on the UCP backlog, and will put this on our 'list'. Is this causing customer problems? If so, we can prioritize it and get to it sooner.

paigehargrave:
@nicolaka - I've attempted some updates in https://github.com/docker/docker.github.io/pull/8309. Please bear with me as I try to learn this stuff! (And yes, you're allowed to laugh at my attempt. :)) PTAL and see what needs to be fixed/added/deleted. Thank you!

nicolaka:
I provided my feedback @paigehargrave

traci-morrison:
Updates made in https://github.com/docker/docker.github.io/pull/8309.

PR #8309, opened 2019-02-19
RBAC instruction update for isolate-nodes.md

paigehargrave:
Attempted to incorporate info from https://kubernetes.io/docs/reference/access-authn-authz/admission-controllers/#podnodeselector. 

#7828

### DIFF
--- a/ee/ucp/authorization/isolate-nodes.md
+++ b/ee/ucp/authorization/isolate-nodes.md
@@ -208,11 +208,11 @@ Create a grant to the `namespace-name` namespace:
 
 1. On the **Create Grant** page, select **Namespaces**.
 
-![](../images/isolate-nodes-5.png){: .with-border}
+    ![](../images/isolate-nodes-5.png){: .with-border}
 
-Select the **namespace-name** namespace, and create a `Full Control` grant.
+2. Select the **namespace-name** namespace, and create a `Full Control` grant.
 
-![](../images/isolate-nodes-6.png){: .with-border}
+    ![](../images/isolate-nodes-6.png){: .with-border}
 
 ### Associate nodes with the namespace
 
--- a/ee/ucp/authorization/isolate-nodes.md
+++ b/ee/ucp/authorization/isolate-nodes.md
@@ -226,20 +226,42 @@ Select the **ops-nodes** namespace, and create a `Full Control` grant for the
 
 ### Link the namespace to a node collection
 
-The last step is to link the Kubernetes namespace the `/Prod` collection.
+The last step is to link the Kubernetes namespace to the `/Prod` collection. Namespaces are associated with a node collection using the namespace definition information provided in a configuration and an annotation file:
+    
+#### Configuration file
+The configuration file specifies backend behavior in json or yaml using the following format:
+    ```
+    podNodeSelectorPluginConfig:
+     clusterDefaultNodeSelector: name-of-node-selector
+     namespace1: ops-nodes
+    ```
+The `PodNodeSelector` configuration file is specified in the file provided to the API server with the command line flag '--admission-control-config-file` using the following format:
+    ```
+    kind: AdmissionConfiguration
+    apiVersion: apiserver.k8s.io/v1alpha1
+    plugins:
+    - name: PodNodeSelector
+      path: podnodeselector.yaml
+    ```
 
-1.  Navigate to the **Namespaces** page, and find the **ops-nodes** namespace
-    in the list.
-2.  Click the **More options** icon and select **Link nodes in collection**.
+#### Annotation file
+The `scheduler.alpha.kubernetes.io/node-selector` annotation key assigns node selectors to namespaces:
+    ```
+    apiVersion: v1
+    kind: Namespace
+    metadata:
+      annotations:
+        scheduler.alpha.kubernetes.io/node-selector: name-of-node-selector
+      name: ops-nodes
+    ```
 
-    ![](../images/isolate-nodes-7.png){: .with-border}
+`PodNodeSelector` actions
+    - If the namespace contains an annotation with `scheduler.alpha.kubernetes.io/node-selector` key, use that value as the node selector.
+    - If the namespace lacks a `scheduler.alpha.kubernetes.io/node-selector` key annotation, use the `clusterDefaultNodeSelector` defined in the PodNodeSelector plugin configuration file as the node selector.
+    - The pod’s node selector is evaluated against the namespace node selector. Conflicts result in rejection.
+    - The pod’s node selector is evaluated against the namespace-specific whitelist defined in the plugin configuration file. Conflicts result in rejection.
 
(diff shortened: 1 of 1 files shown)


ee/ucp/authorization/isolate-nodes.md
```diff
@@ -226,20 +226,42 @@ Select the **ops-nodes** namespace, and create a `Full Control` grant for the
 
 ### Link the namespace to a node collection
```
nicolaka: With UCP 3.1, linking the  namespace to a node collection is not available. It's obsolete.

nicolaka: Also, I don't think the title of the section should be `Deploy a Kubernetes application` It's misleading because that's not really what we're trying to do here. The goal should be around `Isolating Nodes to Kubernetes Namespaces` then you can start with the first step which is creating a namespace.

ee/ucp/authorization/isolate-nodes.md
```diff
@@ -226,20 +226,42 @@ Select the **ops-nodes** namespace, and create a `Full Control` grant for the
...
 
-The last step is to link the Kubernetes namespace the `/Prod` collection.
-
-1.  Navigate to the **Namespaces** page, and find the **ops-nodes** namespace
-    in the list.
-2.  Click the **More options** icon and select **Link nodes in collection**.
+The last step is to link the Kubernetes namespace to the `/Prod` collection. Namespaces are associated with a node collection using the namespace definition information provided in a configuration and an annotation file:
+    
+#### Configuration file
+The configuration file specifies backend behavior in json or yaml using the following format:
```
nicolaka: So from my understanding based on https://kubernetes.io/docs/reference/access-authn-authz/admission-controllers/#configuration-file-format-1, there are two methods to accomplish isolating nodes to a namespace. Format # 1 outlines how to do it all at once from a single configuration file or Format #1 specify it within every Namespace definition. TBH I'm recommending the second format as it's more straight forward and can be done when the namespace is created. But you can link this kubernetes document in our doc and say here are the two formats.

ee/ucp/authorization/isolate-nodes.md
```diff
@@ -226,20 +226,42 @@ Select the **ops-nodes** namespace, and create a `Full Control` grant for the
...
+    ```
+    kind: AdmissionConfiguration
+    apiVersion: apiserver.k8s.io/v1alpha1
+    plugins:
+    - name: PodNodeSelector
+      path: podnodeselector.yaml
+    ```
 
-    ![](../images/isolate-nodes-7.png){: .with-border}
+#### Annotation file
```
nicolaka: I  think we should just use this method and try to explain that as long as we define the `scheduler.alpha.kubernetes.io/node-selector: name-of-node-selector` when we create the namespace, all applications deployed in that namespace will be pinned to the nodes with the node selector specified.

ee/ucp/authorization/isolate-nodes.md
```diff
@@ -226,20 +226,42 @@ Select the **ops-nodes** namespace, and create a `Full Control` grant for the
...
+`PodNodeSelector` actions
+    - If the namespace contains an annotation with `scheduler.alpha.kubernetes.io/node-selector` key, use that value as the node selector.
+    - If the namespace lacks a `scheduler.alpha.kubernetes.io/node-selector` key annotation, use the `clusterDefaultNodeSelector` defined in the PodNodeSelector plugin configuration file as the node selector.
+    - The pod’s node selector is evaluated against the namespace node selector. Conflicts result in rejection.
+    - The pod’s node selector is evaluated against the namespace-specific whitelist defined in the plugin configuration file. Conflicts result in rejection.
 
-    ![](../images/isolate-nodes-8.png){: .with-border}
+> Note: `PodNodeSelector` allows pods to be forced to run on specifically labeled nodes. Refer the PodTolerationRestriction admission plugin, which allows preventing pods from running on specifically tainted nodes.
 
 ### Deploy a Kubernetes workload to the node collection
```
nicolaka: This is obsolete.

 In a nutshell, Admins wishing to pin all applications to deployed in the red namespace to nodes in the west region need to do two things.
1. label the nodes ( e.g zone=west)
2. add an scheduler  node selector annotation as part of the namespace definition:
```
apiVersion: v1
    kind: Namespace
    metadata:
      annotations:
        scheduler.alpha.kubernetes.io/node-selector: zone=west
      name: red
```
We don't need to overcomplicate it like it is in the kube docs:)

ee/ucp/authorization/isolate-nodes.md
ollypom: I think I agree with "I'm not too sure if it's good to have a single file describing Node RBAC for Swarm and Kubernetes". As the Kubernetes node isolation solution here doesn't integrated with Collections, which is primarily what this page describes.

ee/ucp/authorization/isolate-nodes.md
paigehargrave: @nicolaka Added an additional statement in the doc. Thank you!

ee/ucp/authorization/isolate-nodes.md
paigehargrave: Thank you - deleting obsolete info.

paigehargrave:
@nicolaka @ollypom Thank you very much for the input! I've added some updates - PTAL, and will split the file to separate out Swarm info vs Kubernetes. I just need to ask one more favor, and that's if either of you can point out any information in the top part of the file that is Swarm or Kubernetes specific. Thank you again for your time here!

paigehargrave:
@ollypom - Sorry for the inbox spam - if you have a chance, can you give this a quick look for Swarm-specific info and then I'll try to wrap it up - thanks!

traci-morrison:
@nicolaka PTAL. Are the changes in this PR still applicable? Thanks.

nicolaka:
@traci-morrison yes.

traci-morrison:
@KoryKessel-Docker PTAL. I did not make any additional edits.

ee/ucp/authorization/isolate-nodes.md
```diff
@@ -206,79 +198,47 @@ for Kubernetes workloads.
...
-**Collections**.
+Create a grant to the `namespace-name` namespace:
+
+1. On the **Create Grant** page, select **Namespaces**.
 
 ![](../images/isolate-nodes-5.png){: .with-border}
 
-Select the **ops-nodes** namespace, and create a `Full Control` grant for the
-`Ops` team.
+Select the **namespace-name** namespace, and create a `Full Control` grant.
```
KoryKessel-Docker: This step should be numbered (#2).

KoryKessel-Docker: A suggestion: Indent the two screenshots so that they align with the start of text following the step numbers.

traci-morrison:
@KoryKessel-Docker Thanks for the review. Changes made. PTAL.

PR #8311, opened 2019-02-19
Update part4.md

mmoser18:
Some commands need to be run as Administrator on Windows or else they will fail or produce empty output.

### Proposed changes

Added two notes stating that these commands need to be run as Administrator in Windows or else they won't work or create no output.


L-Hudson:
@jasonbivins PTAL, thanks!

jasonbivins:
Looks good to me, ok to merge

L-Hudson:
Good to go. Thanks @mmoser18  and @jasonbivins!

PR #8314, opened 2019-02-19
engine/security: update apparmor docs to not unload all profiles

### LINKED ISSUE #8289: AppArmor documentation recommends to unload *all* AppArmor profiles which can lead to security issues
### Problem description

The AppArmor-related documentation (engine/security/apparmor.md) contains:

```
To unload a profile from AppArmor:

# stop apparmor
$ /etc/init.d/apparmor stop
# unload the profile
$ apparmor_parser -R /path/to/profile
# start apparmor
$ /etc/init.d/apparmor start
```

This is a terrible idea - "/etc/init.d/apparmor stop" will unload *all* profiles and remove protection from all running processes (not only docker). When starting AppArmor again ("/etc/init.d/apparmor start"), the profiles get loaded, but *can't be applied to running processes*, so these processes *stay unconfined*. Needless to say that this can lead to (being able to exploit) security issues which would have been prevented by the AppArmor profile.

### Problem location

- I saw a problem on the following URL: https://docs.docker.com/engine/security/apparmor/

### Suggestions for a fix

You don't need to (and shouldn't) stop AppArmor just because you want to unload a profile.

Please remove the /etc/init.d/apparmor calls from the section quoted above, and shorten it to

```
To unload a profile from AppArmor:

# unload the profile
$ apparmor_parser -R /path/to/profile
```

ericchiang:
### Proposed changes

Restarting AppArmor isn't necessary to unload a profile. In fact, it has the negative impact of causing all running processes to become unrestricted. See #8289.

Update docs to only unload the profile.

### Related issues (optional)

Fixes #8289

Signed-off-by: Eric Chiang <ericchiang@google.com>

### DIFF
--- a/engine/security/apparmor.md
+++ b/engine/security/apparmor.md
@@ -55,12 +55,8 @@ $ docker run --rm -it --security-opt apparmor=your_profile hello-world
 To unload a profile from AppArmor:
 
 ```bash
-# stop apparmor
-$ /etc/init.d/apparmor stop
 # unload the profile
 $ apparmor_parser -R /path/to/profile
-# start apparmor
-$ /etc/init.d/apparmor start
 ```
 
 ### Resources for writing profiles


L-Hudson:
Thanks for the contribution, @ericchiang!

PR #8326, opened 2019-02-20
Update D4Azure CE channels

carlfischer1:
remove edge channel which is no longer supported in Docker for Azure/AWS


L-Hudson:
good to go

PR #8327, opened 2019-02-20
Update D4AWS CE channels

carlfischer1:
remove edge channel which is no longer supported in Docker for Azure/AWS


carlfischer1:
Not sure why netify failed on my second commit

L-Hudson:
Good to go

PR #8329, opened 2019-02-20
docker compose autocompletion content refresh

usha-mandya:


bermudezmt:
This incorporates changes from #8092

Issue #2626, opened 2017-04-04
carry Content Trust guide from ucp 2.0 to latest version


riyazdf:
### Problem description

We should have the UCP 2.0.x [Content Trust docs](https://docs.docker.com/datacenter/ucp/2.0/guides/content-trust) that detail signing policy carried forward to the latest docs location.

### Problem location

https://docs.docker.com/datacenter/ucp/2.0/guides/content-trust should also be on the "User Guides" section of latest UCP (ex: https://docs.docker.com/datacenter/ucp/2.1/guides/user/secrets/)

### Project version(s) affected

UCP Latest (2.1.x)

### Suggestions for a fix

Should be able to copy the guide verbatim into the new user guides directory; the flow and instructions are the same.

joaofnfernandes:
For the latest releases of UCP and DTR we revamped these articles, so while the content exists it is now spread across multiple places.

* [Run only images you trust](https://docs.docker.com/datacenter/ucp/2.1/guides/admin/configure/run-only-the-images-you-trust/)
* [Configure your Docker engine](https://docs.docker.com/datacenter/dtr/2.2/guides/user/access-dtr/)
* [Configure your Notary client](https://docs.docker.com/datacenter/dtr/2.2/guides/user/access-dtr/configure-your-notary-client/)
* [Sign an image](https://docs.docker.com/datacenter/dtr/2.2/guides/user/manage-images/sign-images/)
* [Delegate image signing](https://docs.docker.com/datacenter/dtr/2.2/guides/user/manage-images/sign-images/delegate-image-signing/)
* [Manage trusted repositories](https://docs.docker.com/datacenter/dtr/2.2/guides/user/manage-images/sign-images/manage-trusted-repositories/#where-to-go-next)

Since the content exists, I'm closing this issue. Feel free to open a new one if you notice that we're missing content or that the content we have is difficult to find.

joaofnfernandes:
It looks like #2656 had already tried to address this, so I'll take a look and check if we should revert it.

PR #8233, opened 2019-02-13
Backport Swarm and Kubernetes deployment docs for UCP 3.0

bermudezmt:
Signed-off-by: Maria Bermudez <bermudez.mt@gmail.com>
### Proposed changes

This addresses #8180 and backports the following content:
- UCP architecture (removes silly redirect to 3.1 version of content)
- Adds Swarm (including interlock) and Kubernetes sections 
- Updates image links based on new subdirectories

### Unreleased project version (optional)

UCP 3.0

### Related issues (optional)

Closes #8180

PR #8333, opened 2019-02-20
Found an old 'DCT with CI' page in the latest UCP Docs

ollypom:
### Proposed changes

So it appears we have been carrying an legacy DCT with UCP page for few releases now, but it hasn't been referenced in the ToC since 2017. 

The page is here: https://github.com/docker/docker.github.io/blob/master/ee/ucp/admin/configure/use-trusted-images-for-ci.md

In my opinion the page needed a complete re-write, and also starts to overlap with some of the information here. https://github.com/docker/docker.github.io/blob/master/engine/security/trust/trust_automation.md as noted in https://github.com/docker/docker.github.io/issues/2626

So I have added the page back into the ToC for legacy pages (UCP 2.1 and 2.2). But removed the page for UCP 3.0 and 3.1.

It appears it was removed from the TOC in this PR: 
https://github.com/docker/docker.github.io/pull/2722


ollypom:
@bermudezmt I'm wondering if this needs to wait on https://github.com/docker/docker.github.io/pull/8233 to stop the Toc getting more confusing :D?

PR #8340, opened 2019-02-21
Relnotes 02 2019 updates

paigehargrave:
Added updates to DTR, UCP, and Engine release notes.

Also attempted to carry known issues forward (for each major release) to address feedback indicating that not carrying this information forward is confusing, and might convey that previous known issues were fixed.

### DIFF
--- a/datacenter/dtr/2.3/guides/release-notes.md
+++ b/datacenter/dtr/2.3/guides/release-notes.md
@@ -11,6 +11,14 @@ known issues for each DTR version.
 You can then use [the upgrade instructions](admin/upgrade.md),
 to upgrade your installation to the latest release.
 
+## Version 2.4.10
+
+(20 February 2019)
+
+### Bug Fixes
+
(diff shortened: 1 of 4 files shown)


david-yu:
@paigehargrave It looks like you might have the wrong folks assigned as approvers here, I would recommend adding @jose-bigio and @caervs

paigehargrave:
@david-yu Thank you! I'm still trying to keep 'who's who' straight. :)

ry4nz:
@paigehargrave There are lots of `known issues` added to UCP release notes.  Is it intentional ?

paigehargrave:
@ry4nz Hi Ryan - Sorry for the extra hassle here. We got feedback that the whole Known Issues info is confusing. Known issues get mentioned in one release and then they 'disappear', which leads people to believe that they got fixed, which is usually not the case. So what I did was try to consolidate the known issues for each release. Where I had issue numbers, I tried to remove the ones that got fixed. However, in the majority of cases, there is no issue number to check, SO if you can scan the known issues to see what still applies, that would be great.

Issue #8151, opened 2019-02-05
Input/output error after building image while trying to start


panagiotismavridis:
[get-started/part2.md](https://docs.docker.com/get-started/part2/)
On Windows 10, while in the powershell while finishing most of the STEPS, just before running one can get something similar to:

Error response from daemon: driver failed programming external connectivity on endpoint compassionate_shockley (4354ce12687e6a9558c53934037f55f3720f136586b5b3ba1241dece190d21c6): Error starting userland proxy: mkdir /port/tcp:0.0.0.0:4000:tcp:172.17.0.2:80: input/output error.

**FIX: If you restart Docker client it runs it with no issues.**
I would suggest in the docs they include a note, that restart is needed. It is a getting started guide, while in part 2 is could be frustrating to start getting these kind of errors 👍

paigehargrave:
@panagiotismavridis - I've added a blurb for restarting - PTAL and make sure that it's in the right location. Thank you for bringing this to our attention!

bermudezmt:
Closing as per #8358 . Please submit another issue if you're still experiencing problems with this guide.

PR #8358, opened 2019-02-26
Added restart info before running

paigehargrave:
#8151

### DIFF
--- a/get-started/part2.md
+++ b/get-started/part2.md
@@ -216,7 +216,7 @@ be something like `--tag=friendlyhello:v0.0.1`.
 
 ## Run the app
 
-Run the app, mapping your machine's port 4000 to the container's published port
+Restart the Docker client. Then run the app, mapping your machine's port 4000 to the container's published port
 80 using `-p`:
 
 ```shell


get-started/part2.md
```diff
@@ -216,7 +216,7 @@ be something like `--tag=friendlyhello:v0.0.1`.
 
 ## Run the app
 
-Run the app, mapping your machine's port 4000 to the container's published port
+Restart the Docker client. Then run the app, mapping your machine's port 4000 to the container's published port
```
thaJeztah: Not entirely sure what is meant with "restarting the docker client" (the `docker` cli is not a long-living process, and is started again with each command you run).

@panagiotismavridis could you explain what you meant with "restarting the client"?

thaJeztah: /cc @ddebroy @StefanScherer

StefanScherer: Hm, maybe Docker Desktop, but then I only could think of that @paigehargrave already had another container running with machine port 4000 mounted.

Otherwise I don't think this is needed. This tutorial looks good as it is.

bermudezmt: @panagiotismavridis can you confirm that you had the scenario described by @StefanScherer above? If so, the change may not be necessary. Please let us know what your environment looks like.

bermudezmt:
@panagiotismavridis Thank you for your feedback! We have received and reviewed your changes. Based on technical reviews, we did not incorporate your suggested changes. If you have additional concerns, please submit a new issue in this repository.

Issue #8378, opened 2019-02-28
update ucp storage requirements for /var/lib/kubelet ephemeral storage


trapier:
### Problem description

UCP system requirements currently indicate minimum requirements:

> - 4GB of free disk space for the `/var` partition for manager nodes
> - 500MB of free disk space for the `/var` partition for worker nodes

I see a couple issues with this:
1. Units.  Should be base-2 `GiB` and `MiB`.
1. Experience suggests when a manager reaches 4GiB-1B or a worker reaches 500MiB-1B, pods get evicted.  So the currently documented minimum requirement is actually a dysfunctional state.
3. Precision of filesystem path.  The operative path is actually `/var/lib/kubelet`, which is quite often on a different filesystem than `/var`.

### Problem location

https://docs.docker.com/ee/ucp/admin/install/system-requirements/#minimum-requirements

### Project version(s) affected

- UCP 3.1.x

### Suggestions for a fix

> **Note**: IANA UCP development engineer.  Please get a technical :+1: from one on the following (would be required for \#4).

1. Change units to `GiB` and `MiB` or otherwise indicate that units are base-2 not base-10.
2. Change minimum requirement to `current_minimum + 1 GiB` for managers and workers.
3. Express requirement as applying to specific path `/var/lib/kubelet`.
4. (For @ericsmalling): identify and document a *recommendation* as well as a minimum 

/cc https://success.docker.com/article/kublet-restarting-after-upgrade-to-universal-control-plane-31

paigehargrave:
@trapier I created #8385 as a start - PTAL. Thank you for bringing this forward!

paigehargrave:
@trapier @quadespresso @shantiallamudi Hi all - I've merged #8385 - please let me know if additional changes are needed or if we can close this issue - thank you!

quadespresso:
LGTM

paigehargrave:
@quadespresso Thank you! I'm going to close this....

PR #8385, opened 2019-03-01
Kubelet storage requirements for UCP 3.1

paigehargrave:
#8378


paigehargrave:
@david-yu @trapier Thank you for reviewing!

PR #8389, opened 2019-03-01
Update info for deploy with compose.md

paigehargrave:
Addresses https://github.com/docker/docs-private/issues/791


ee/ucp/kubernetes/deploy-with-compose.md
```diff
@@ -30,28 +30,19 @@ version: '3.3'
...
-        limits:
-          memory: 16M
-        reservations:
-          memory: 16M
 
   db:
     build: db
-    image: dockerdemos/lab-db
-```
+    image: dockersamples/k8s-wordsmith-db
```
david-yu: Looks like you forgot the ``` here but otherwise looks good

paigehargrave:
@stevenfollis @sixeyed - Hi - Can you both PTAL and see if anything else is needed? Thank you!

stevenfollis:
Looks like `words` needs one more space in front of it, otherwise is invalid YAML :(

ee/ucp/kubernetes/deploy-with-compose.md
```diff
@@ -30,27 +30,19 @@ version: '3.3'
...
-    image: dockerdemos/lab-web
-    volumes:
-     - "./web/static:/static"
+    image: dockersamples/k8s-wordsmith-web
     ports:
-     - "80:80"
+     - "8080:80"
 
-  words:
+ words:
```
stevenfollis: add an extra space here to properly validate the YAML

paigehargrave: OK - I think I got it - thanks!

paigehargrave: @bermudezmt Maria - would you be able to help out with a couple screen shots per @stevenfollis in https://github.com/docker/docs-private/issues/791?

bermudezmt:
@david-yu for the sake of getting this done faster, I'll drop the errors I've encountered here:

1) https://github.com/dockersamples/k8s-wordsmith-demo/blob/master/kube-deployment.yml - results in an "Invalid YAML" error
2) https://github.com/dockersamples/k8s-wordsmith-demo/blob/master/docker-compose-simple.yml - results in a "Stack: Service unavailable" error. 

What kind of role binding do I need to create for my namespace in order to deploy a stack?

stevenfollis:
FWIW, when using the Raw view on GH, https://raw.githubusercontent.com/dockersamples/k8s-wordsmith-demo/master/kube-deployment.yml works:

```yaml
apiVersion: v1
kind: Service
metadata:
  name: db
  labels:
    app: words-db
spec:
  ports:
    - port: 5432
      targetPort: 5432
      name: db
  selector:
    app: words-db
  clusterIP: None
---
apiVersion: apps/v1beta1
kind: Deployment
metadata:
  name: db
  labels:
    app: words-db
spec:
  template:
    metadata:
      labels:
        app: words-db
    spec:
      containers:
      - name: db
        image: dockersamples/k8s-wordsmith-db
        ports:
        - containerPort: 5432
          name: db
---
apiVersion: v1
kind: Service
metadata:
  name: words
  labels:
    app: words-api
spec:
  ports:
    - port: 8080
      targetPort: 8080
      name: api
  selector:
    app: words-api
  clusterIP: None
---
apiVersion: apps/v1beta1
kind: Deployment
metadata:
  name: words
  labels:
    app: words-api
spec:
  replicas: 5
  template:
    metadata:
      labels:
        app: words-api
    spec:
      containers:
      - name: words
        image: dockersamples/k8s-wordsmith-api
        ports:
        - containerPort: 8080
          name: api
---
apiVersion: v1
kind: Service
metadata:
  name: web
  labels:
    app: words-web
spec:
  ports:
    - port: 8081
      targetPort: 80
      name: web
  selector:
    app: words-web
  type: LoadBalancer
---
apiVersion: apps/v1beta1
kind: Deployment
metadata:
  name: web
  labels:
    app: words-web
spec:
  template:
    metadata:
      labels:
        app: words-web
    spec:
      containers:
      - name: web
        image: dockersamples/k8s-wordsmith-web
        ports:
        - containerPort: 80
          name: words-web
```

image


Shows properly in browser when using the `NodePort`:

image

This said, it deploys a `LoadBalancer` service by default. This assumes the user is deploying onto a cloud infrastructure or similar. Probably not a deal breaker if we presume or instruct them to use the NodePort IP address. The LoadBalancer IP for me did not work, but may be a misconfiguration in my environment.

bermudezmt:
@stevenfollis I managed to reboot my testkit instance running `v3.1.4`, and was able to deploy a `LoadBalancer` service like you've suggested. How do I access the app from the browser like what is shown in your screenshot? By NodePort IP, do you mean using the Cluster IP and Node Port info for the load balancer? 

@david-yu I'm a bit concerned that this is already deviating too much from the original sample application. I also tested a different service from QA called `dockercoins` which worked with both v3.1.4 and v3.2.0-alpha2 but that's also a completely different setup from the original demo. Let me know your thoughts on how we can proceed with updating this doc. Maybe there's a different Kubernetes demo that we can use.

For reference, I've filed a related JIRA issue to track the "Invalid yml" error: https://docker.atlassian.net/browse/ENGORC-2094

david-yu:
This should work with the Compose deployment on Kube. I would also use what Stephen had described. Usually I copy and paste the contents in the editor instead of using the files themselves, however both approaches should work.

```
version: '3.3'

services:

  db:
    image: dockersamples/k8s-wordsmith-db

  words:
    image: dockersamples/k8s-wordsmith-api
    deploy:
      replicas: 5

  web:
    image: dockersamples/k8s-wordsmith-web
    ports:
     - "32768:80"
```

We should use this sample application because is the sample app that we used at DockerCon EU and the others are fairly out of date.

david-yu:
> @stevenfollis I managed to reboot my testkit instance running `v3.1.4`, and was able to deploy a `LoadBalancer` service like you've suggested. How do I access the app from the browser like what is shown in your screenshot? By NodePort IP address, do you mean using the Cluster IP and Node Port info of the app load balancer? I can't seem to load the app using those details.
> 
> @david-yu I'm a bit concerned that this is already deviating too much from the original sample application. I also tested a different service from QA called `dockercoins` which worked with both v3.1.4 and v3.2.0-alpha2 but that's also a completely different setup from the original demo. Let me know your thoughts on how we can proceed with updating this doc. Maybe there's a different Kubernetes demo that we can use.
> 
> For reference, I've filed a related JIRA issue to track the "Invalid yml" error: https://docker.atlassian.net/browse/ENGORC-2094

As far as accessing the app, you need to make sure that AWS has that specific port opened up. Also you need to go the public ip address of the cluster and hit the specific port on the public ip. I believe the cluster ip shown might be the private ip of the cluster.

PR #8391, opened 2019-03-01
Update running_ssh_service.md

### LINKED ISSUE #8390: SSH Service docs example creates insecure sshd server configuration
[engine/examples/running_ssh_service.md](https://docs.docker.com/engine/examples/running_ssh_service/)

This creates an insecure-by-default configuration. Specifically, the Dockerfile [here](https://docs.docker.com/engine/examples/running_ssh_service/#build-an-eg_sshd-image) specifies a static root password, `screencast`. Hard-coding this password into the container creates a security issue; if the container is bridged to the internet, either via NAT or a badly-configured hosting service, it becomes an easy target for attackers.

plygrnd:
### Proposed changes

The Dockerfile [here](https://docs.docker.com/engine/examples/running_ssh_service/#build-an-eg_sshd-image) specifies a static root password, `screencast`. Hard-coding this password into the container creates a security issue; if the container is bridged to the internet, either via NAT or a badly-configured hosting service, it becomes an easy target for attackers.

This change goes some way towards fixing this by specifying that readers should generate a random password before building this image; however, this is a temporary solution. Long term, we need to determine how to wrap this Dockerfile in something that generates a random password at build time.

### Related issues (optional)

Fixes #8390

### DIFF
--- a/engine/examples/running_ssh_service.md
+++ b/engine/examples/running_ssh_service.md
@@ -6,16 +6,24 @@ title: Dockerize an SSH service
 
 ## Build an `eg_sshd` image
 
+### Generate a secure root password for your image
+
+Using a static password for root access is dangerous. Create a random password before proceeding.
+
+### Build the image
+
 The following `Dockerfile` sets up an SSHd service in a container that you
 can use to connect to and inspect other container's volumes, or to get
-quick access to a test container.
+quick access to a test container. 
+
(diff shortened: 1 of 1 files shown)


L-Hudson:
@quadespresso  PTAL thanks!

L-Hudson:
thank you @quadespresso and @plygrnd!

Issue #8251, opened 2019-02-14
Missing information


kavikat:
[ee/ucp/admin/install/upgrade.md](https://docs.docker.com/ee/ucp/admin/install/upgrade/)

There is no information regarding upgrade paths, this is historically inconsistent with UCP upgrade docs and is important information to share with customers.  Please refer to :
https://docs.docker.com/v17.09/datacenter/ucp/2.2/guides/admin/install/upgrade/#recommended-upgrade-paths

paigehargrave:
@ry4nz - Hi Ryan, perhaps there is some history on this that I'm unaware of. The previous upgrade path info is:
If you're running a UCP version that's lower than 2.1, first upgrade to the latest 2.1 version, then upgrade to 2.2. Use these rules for your upgrade path to UCP 2.2:

From UCP 1.1: UCP 1.1 -> UCP 2.1 -> UCP 2.2
From UCP 2.0: UCP 2.0 -> UCP 2.1 -> UCP 2.2
From UCP 2.1: UCP 2.1 -> UCP 2.2

Does this still apply to UCP 3.1, and if not, what is the upgrade path? Thank you in advance!

ry4nz:
For UCP 3.1, upgradable from 3.0
For UCP 3.0, upgradable from 2.2
For UCP 2.2: upgradable from 2.1

bermudezmt:
Closed via #8400

PR #8400, opened 2019-03-04
Update UCP upgrade path

paigehargrave:
#8251


paigehargrave:
@ry4nz Can you PTAL and make sure this is OK? Thanks!

PR #8397, opened 2019-03-03
Download latest version of docker-compose from github directly.

Maannniii:

PR #8411, opened 2019-03-06
Clarify that compose installation is for stable release

thaJeztah:
relates to https://github.com/docker/docker.github.io/pull/8397#discussion_r262560113

This patch updates the compose installation instruction to
be more clear about the default instructions installing
the current stable release.

Inline nodes about installing different versions has been
reduced, and the "Master builds" section has been renamed
to "Install pre-release builds", to capture installing either
"master" builds, or pre-releases (release-candidates) from
GitHub.


thaJeztah:
ping @chris-crone @L-Hudson ptal

PR #8413, opened 2019-03-06
Dockerfile: optimize for caching

thaJeztah:
## (See commit messages of individual commits for details)

The implementation of the Dockerfile first copied the current
version of the documentation, to be followed by fetching upstream
resources, and adding the archived documentation.

As a result, the build-cache was "busted" on every change that was
made locally, causing a full rebuild of both "fetching upstream resources",
building the reference documentation, and generating the archived
versions of the documentation.

With the changes made to the "fetch-upstream-resources.sh" script,
it's now possible to build those docs in advance (in a separate build-
stage), and append the pre-built docs _after_ the current version
of the documentation was built.

This patch updates the Dockerfile to build reference-docs and
archived docs in separate build-stages, optimizing the build for
caching (wich the assumption that no changes are made in upstream
resources between builds). Not only caused this builds to take longer
than needed, but each rebuild also caused new "dangling" images to
be created (in case of the "classic" builder), or new build-caches
to be created (when using BuildKit).

Note that when _deploying_ the documentation, no caching should be
used (to enforce a full rebuild of the documentation, and guarantee
that reference-docs are "fresh").

Before this change, rebuilding the documentation (after a previous
build, so with cache present) took ~185 seconds;

```
docker build -t docs .

[+] Building 184.7s (26/26) FINISHED

=> [builder 2/5] COPY . md_source 2.8s
=> [builder 3/5] RUN bash ./md_source/_scripts/fetch-upstream-resources.sh md_source 52.7s
=> [builder 4/5] RUN jekyll build -s md_source -d /usr/share/nginx/html --config md_source/_config.yml 113.0s
=> [builder 5/5] RUN find /usr/share/nginx/html -type f -name '*.html' | grep -vE "v[0-9]+\." | while read i; do sed -i 's#href="https://docs.docker.com/#href="/#g' "$i"; done 4.4s
=> CACHED [stage-1 2/8] COPY --from=docs/docker.github.io:nginx-onbuild /etc/nginx/conf.d/default.conf /etc/nginx/conf.d/default.conf 0.0s
=> CACHED [stage-1 3/8] COPY --from=docs/docker.github.io:v17.03 /usr/share/nginx/html /usr/share/nginx/html 0.0s
=> CACHED [stage-1 4/8] COPY --from=docs/docker.github.io:v17.06 /usr/share/nginx/html /usr/share/nginx/html 0.0s
[...]

PR #8414, opened 2019-03-06
[WIP] Dockerfile: split stages for "dev" and "deploy"

thaJeztah:
~This is a follow-up to https://github.com/docker/docker.github.io/pull/8413 - only the last commit is new~ (rebased)

The current Dockerfile defaults to include all archived documentation. While this
is the right thing to do when _deploying_ the documentation, it is far from optimal
when working on changes in the documentation, and building the documentation locally
(in which case, the archived documentation is likely not interesting).

This patch changes the Dockerfile to only build the current version of the documentation
by default.

> **Note** when building with the "classic" builder (so not using BuildKit), the archived
> documentation will still be built, because the classic builder is not able to skip
> unused build-stages.

This change might require changes in how the documentation is deployed, because in that
case, the documentation _including_ archives should be built.

Alternatively, the default could be changed to include the archives, but people wanting
to build docs without archives could set `--target=dev` when running `docker build`.

Before this change a full rebuild of the docs, including archives, took ~247 seconds
(already optimised if BuildKit is used, which fetches the archived docs in parallel):

```
docker build -t docs --target=deploy .

[+] Building 247.1s (30/30) FINISHED
=> FROM docker.io/docs/docker.github.io:v18.03 151.4s
=> FROM docker.io/docs/docker.github.io:v17.03 75.0s
=> FROM docker.io/docs/docker.github.io:v17.09 96.5s
=> FROM docker.io/docs/docker.github.io:v17.06 134.9s
=> FROM docker.io/docs/docker.github.io:v17.12 105.2s
...
=> [archives 1/5] COPY --from=docs/docker.github.io:v17.03 /usr/share/nginx/html /usr/share/nginx/html 1.7s
=> [archives 2/5] COPY --from=docs/docker.github.io:v17.06 /usr/share/nginx/html /usr/share/nginx/html 2.8s
=> [archives 3/5] COPY --from=docs/docker.github.io:v17.09 /usr/share/nginx/html /usr/share/nginx/html 3.8s
=> [archives 4/5] COPY --from=docs/docker.github.io:v17.12 /usr/share/nginx/html /usr/share/nginx/html 4.6s
=> [archives 5/5] COPY --from=docs/docker.github.io:v18.03 /usr/share/nginx/html /usr/share/nginx/html 2.7s
```

The size of the built image is ~1.6 GB

```
REPOSITORY TAG IMAGE ID CREATED SIZE
docs latest 8fd1a9d31da7 4 minutes ago 1.61GB
```

With this change, the archived docs are not fetched, and a full rebuild of the docs is reduced to ~200 seconds

```
[+] Building 203.1s (20/20) FINISHED
```

Given that the archives are not included by default, the size of the image also reduced considerably,
and is now about 340MB

```
REPOSITORY TAG IMAGE ID CREATED SIZE
docs latest bb7b1b9af649 7 seconds ago 339MB
```


thaJeztah:
Let me put this one as "WIP" for a moment, to verify if (and what) changes are needed for deploying the docs; who's best to ask?

thaJeztah:
(I'll rebase this one after https://github.com/docker/docker.github.io/pull/8413 is merged, to get rid of the extra commits)

bermudezmt:
@peter-docker can you PTAL at this one? I have merged #8413 but need an extra pair of eyes on this change.

bermudezmt:
@thaJeztah looks like this change breaks the archived versions of our site. Let me know if I missed anything on my end with testing.
Screenshot from 2019-03-08 15-29-38

thaJeztah:
Ah, yes, that's expected if it's using the default `--target`; if for deployment the `deploy` target could be used, it would include the archives

```
docker build -t docs --target=deploy .
```

Not sure if that's easy to do though; so we could swap the defaults, and require people that want to build _without_ archives to use (e.g.) `docker build --target=dev .`

FWIW, the _ideal_ situation would be to deploy each archive as their own container (this is what I was originally was working on in https://github.com/docker/docker.github.io/pull/941)

bermudezmt:
@thaJeztah sorry I was running around in SF so I had missed your comments on how to test this change. Specifying `--target=deploy` does build all the archives and it's much snappier. I have no problem with merging this PR, unless you wanted to swap the defaults on this PR. I would update the README to include that info if we went that way. 

I guess the only additional change I could think of would be to have the dropdown links of the archived versions link to localhost instead of `docs.docker.com`. I suspect that will be a trivial change, but wanted to note that here before I forget.

bermudezmt:
@shampabh Let's also have a look at this one during our team meeting.

thaJeztah:
This was superseded by https://github.com/docker/docker.github.io/pull/10331 and https://github.com/docker/docker.github.io/pull/10342

PR #8420, opened 2019-03-07
Dockerfile follow-up fixes and improvements

thaJeztah:
Whoops; noticed that my previous change in https://github.com/docker/docker.github.io/pull/8413 introduced a small bug; fixing it in this PR; the fix is in the first commit; other changes are follow-ups to make some small improvements (to be used in https://github.com/docker/docker.github.io/pull/8414)

### DIFF
--- a/_scripts/fetch-library-samples.sh
+++ b/_scripts/fetch-library-samples.sh
@@ -1,0 +1,49 @@
+#!/usr/bin/env bash
+
+# Get the Library docs
+svn co https://github.com/docker-library/docs/trunk ./_samples/library || (echo "Failed library download" && exit 1)
+# Remove symlinks to maintainer.md because they break jekyll and we don't use em
+find ./_samples/library -maxdepth 9  -type l -delete
+# Loop through the README.md files, turn them into rich index.md files
+FILES=$(find ./_samples/library -type f -name 'README.md')
+for f in ${FILES}
+do
+  curdir=$(dirname "${f}")
+  justcurdir="${curdir##*/}"
+  if [ -e ${curdir}/README-short.txt ]
+  then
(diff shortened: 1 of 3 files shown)


thaJeztah:
ping @L-Hudson @bermudezmt ptal

_scripts/fetch-library-samples.sh
```diff
@@ -0,0 +1,49 @@
+#!/usr/bin/env bash
+
+# Get the Library docs
+svn co https://github.com/docker-library/docs/trunk ./_samples/library || (echo "Failed library download" && exit 1)
```
bermudezmt: @thaJeztah sorry this hasn't been communicated clearly in a previous PR. We have updated the library logic to have the sample library links point to Hub pages instead. I don't think fetching them is still necessary except for archived versions. Let me double check on this though.

thaJeztah: Yeah, I never liked the samples being there, but it was still in the dockerfile, so I kept it

thaJeztah: oh, looks like the "samples home" https://docs.docker.com/samples/ links to the docker hub pages, but the navigation still has the "fetched" versions as well

screenshot 2019-03-07 at 23 26 35

And those pages are still active; https://docs.docker.com/samples/library/alpine/

I can work on turning those into redirects, and remove them from the actual docs (that may need a bit of thinking, so I'd prefer doing that in a follow-up to not uphold the fixes in this PR)

bermudezmt: Both of those comments are accurate - just haven't had the time to come back and clean those up. :( Thank you - will merge this one then!

PR #8428, opened 2019-03-08
remove references to obsolete versions, and rewrite stdin section

thaJeztah:
This removes references to obsolete versions of Docker, and rewrites
the section on building from a Dockerfile piped through `stdin`.

Also included (second commit) are some touch-ups of the Markdown to allow better-highlighting (still depends on rouge 2.x to be used to actually highlight Dockerfiles; https://github.com/docker/docker.github.io/pull/1021)


L-Hudson:
good to go!

PR #8431, opened 2019-03-08
Added extra tip for "connectivity trouble"

kryptonite303:
### Proposed changes

Added an extra tip to the "Having connectivity trouble?" section for those who are coming directly from part 3 and haven't changed the port forwarding configuration in their docker-compose.yml file. 
Part 3 has port 4000 forwarding to 80, instead of 80 to 80. This is to save people any headache from trying to troubleshoot why their results aren't lining up with the screenshot example above.


L-Hudson:
Thanks for the clarification, @kryptonite303

PR #8441, opened 2019-03-11
DTR from failure has to be destroyed.

gonzalo-contador:
I have added the following.  If you are restoring DTR after a failure, please maker sure you have destroyed the old DTR fully. 
 

### Proposed changes

Many customer do not understand that in order to reinstall DTR from back up the old DTR MUST BE DESTROYED.


L-Hudson:
@bermudezmt PTAL, thanks!

PR #8443, opened 2019-03-12
replacing TODO with proper link

BretFisher:
Doc had a bad link with TODO in it, maybe waiting for the eventual blog post URL. I replaced that text with the proper blog link :)


L-Hudson:
You're the bomb. Thank you so much for the catch/fix!

PR #8453, opened 2019-03-13
Update network-tutorial-macvlan.md

mayank23:
### Proposed changes

Line number 54 has the following sentence: 

> The`-dit` flags start the container in the background but allow you to attach

However, when the command is invoked it uses `--itd`. 

> `docker run --rm -itd `

Matching up description with the actual command

### DIFF
--- a/network/network-tutorial-macvlan.md
+++ b/network/network-tutorial-macvlan.md
@@ -55,7 +55,7 @@ on your network, your container appears to be physically attached to the network
     to it. The `--rm` flag means the container is removed when it is stopped.
 
     ```bash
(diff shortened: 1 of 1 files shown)


paigehargrave:
@bermudezmt PTAL - Does the order of the options matter? e.g `-idt` versus `-dit`? I'm not sure this is an issue.

mayank23:
@paigehargrave It's not an issue, I just made the PR for consistency between the description and command used 😄

adrian-plata:
Nice catch, thanks for helping us be more consistent!

PR #8458, opened 2019-03-14
Change "Linux" limitation wording

mipearson:
### Proposed changes

I read the previous wording as "Only works when run from a Linux host", where what it means is "Only works when building a Linux container".

In other words, build enhancements work just fine if building Linux containers via Docker for Mac/Windows.

I've changed the wording to make this clearer.


bermudezmt:
Thank  you, @tonistiigi and @mipearson .

PR #8462, opened 2019-03-14
Update running_riak_service.md

nuttwerx:
fix error when apt-get is unable to locate package curl in cache

### Proposed changes

When building the riak image and installing curl with apt-get the command fails with the error "E: Unable to locate package curl", executing an 'apt-get update' on the same line fixes the issue


craig-osterhout:

PR #8476, opened 2019-03-17
Shorten all (-a) and quite (-q) in recap cheatsheet

renjithspace:
### Proposed changes

Shorten all, `-a` and quite, `-q` params  to `--aq`

### DIFF
--- a/get-started/part2.md
+++ b/get-started/part2.md
@@ -425,10 +425,10 @@ docker container ls -a             # List all containers, even those not running
 docker container stop <hash>           # Gracefully stop the specified container
 docker container kill <hash>         # Force shutdown of the specified container
 docker container rm <hash>        # Remove specified container from this machine
-docker container rm $(docker container ls -a -q)         # Remove all containers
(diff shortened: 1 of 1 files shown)


renjithspace:
Screenshot from 2019-03-19 21-30-01
@thaJeztah  Both work the same way. I suggesting it's better to have in shorten way.

paigehargrave:
https://github.com/docker/docker.github.io/pull/8878. @thaJeztah @renjithspace PTAL - thank you!

thaJeztah:
Without an explanation in this part of the docs, I don't think we should make this change (also see my comment above); we should either keep it as-is, or decide if we want to explain that shorthand flags can be concatenated (which could be too much information at this point, and possibly distract from the general flow in this getting started)

adrian-plata:
@renjithspace thanks for contributing! As was stated above, we'll be keeping `-a` and `-q` separate given the nature of the page. I'll be closing this PR as a result.

Issue #8445, opened 2019-03-12
Show error message that appears if you haven't added a DTR to UCP


programmerq:
[ee/dtr/user/manage-images/sign-images/trust-with-remote-ucp.md](https://docs.docker.com/ee/dtr/user/manage-images/sign-images/trust-with-remote-ucp/)

Another example error to consider adding:

```
$ docker run --rm -it dtr.example.com/admin/trustdemo:1 echo hello world
docker: Error response from daemon: dtr.example.com must be a registered trusted registry.
See 'docker run --help'.
```

This error message can happen until you register DTR with the currently running UCP, as outlined in the registering-dtr-with-a-remote-universal-control-plane section of this document.

This error message didn't reveal a whole lot when googled, and there should be a clear pointer from this error message to that section.

bermudezmt:
Closed via #8480 .

PR #8480, opened 2019-03-19
Additional error for troubleshooting section

bermudezmt:
Addresses #8445 .

### DIFF
--- a/ee/dtr/user/manage-images/sign-images/trust-with-remote-ucp.md
+++ b/ee/dtr/user/manage-images/sign-images/trust-with-remote-ucp.md
@@ -249,7 +249,7 @@ signed the image with the wrong user keys.
 #### DTR URL must be a registered trusted registry
 
 ```
-Error response from daemon: dtr.example.com must be a registered trusted registry. See 'docker run --help'.
+Error response from daemon: dtr.example.com must be a registered trusted registry.
 ```
 
 This means you have not registered DTR to work with a remote UCP instance yet, as outlined in [Registering DTR with a Remote Universal Control Plane](#registering-dtr-with-a-remote-universal-control-plane).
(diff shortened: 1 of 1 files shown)


ee/dtr/user/manage-images/sign-images/trust-with-remote-ucp.md
```diff
@@ -243,6 +246,14 @@ This means that the image was signed correctly, however the user who signed the
 image does not meet the signing policy in cluster 2. This could be because you
 signed the image with the wrong user keys.
 
+#### DTR URL must be a registered trusted registry
+
+```
+Error response from daemon: dtr.example.com must be a registered trusted registry. See 'docker run --help'.
```
ollypom: I would remove the 'See docker run --help' bit. That is a second line in the error message, and is relevant to how the container was started not related to the trust error message. Took me a double take to work out where it came from :D

bermudezmt: Yikes, so picky. But will do!

programmerq: Hmmm, that second bit _is_ there and isn't on a separate line.

The `docker` cli templatizes the error, and puts "Error response from daemon: " at the front, and "See 'docker run --help'" at the end. In the middle is the actual text from the daemon, in this case UCP is acting like the docker daemon and generates that part of the error.

Since the "See 'docker run --help'" portion of the error message is 100% inline and isn't separated by a newline, I'm inclined to keep it in the docs since that's the actual text of the error message that will be seen by the user.

ollypom: @programmerq Ahh ok, I didn't realise that. Happy to have the `docker run --help` put back in, to help customers find an exact match for the error message. I'll do a quick PR. 

In reality that "docker run --help" isn't going to help them very much in troubleshooting that particular error. But that's a whole different can of worms.... :D

PR #8484, opened 2019-03-19
[WIP] Move the Kubernetes pod cidr note to the prerequisites

dockerpac:
### Proposed changes
I moved the part "- The Kubernetes `pod-cidr` must match the Azure Vnet of the hosts. " from an additionnal note at the end of the documentation to the Prerequisites section of the documentation

PR #8485, opened 2019-03-19
Patch UCP Azure instructions 

ollypom:
### Proposed changes

I have moved the warning to ensure a pod-cidr is within an Azure Subnet more obvious.

### DIFF
--- a/ee/ucp/admin/install/install-on-azure.md
+++ b/ee/ucp/admin/install/install-on-azure.md
@@ -207,6 +207,9 @@ The `--pod-cidr` option maps to the IP address range that you configured for
 the subnets in the previous sections, and the `--host-address` maps to the
 IP address of the master node.
 
+> Note: The `pod-cidr` range must be within an Azure subnet attached to the
+> host.
+
 ```bash
 docker container run --rm -it \
   --name ucp \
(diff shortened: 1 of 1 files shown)


ollypom:
@dockerpac WDYT?

ollypom:
Ahhh I've just seen you made https://github.com/docker/docker.github.io/pull/8484. 

I kept the note within the installation instructions, however now made it bold in a markdown note. The reason I've kept it in the `Install UCP` section, is because the Pod-Cidr is not defined until down at the top of the page, only within the install section :)

dockerpac:
Ok that makes sense.
Still I think it should be also noted in the "Prerequisites paragraph" because basically, if i'm not aware of this restriction, i could want to install my nodes in a small vnet, and use the default 192.168.0.0/16 overlay network for pods cidr as in all other docker ee installations.
So for my point of view, with this information in mind, you would know at the design phase (not at the installation phase) that you need to provision a vnet large enough to have your nodes but also all of your pods.
This is explained in details in the paragraph "Considerations for IPAM Configuration", but I think it should be also explicitly put in the "Prerequisites" paragraph

ollypom:
@dockerpac Ok that makes sense, I've added a section specifically calling out the vnet/subnet sizes in the Pre-reqs, and referencing the "Considerations for IPAM Configuration".

WDYT?

dockerpac:
👍

PR #8506, opened 2019-03-21
[WIP] Split UCP RBAC into Kubernetes and Swarm Sections

ollypom:
This is WIP do not merge :) 

### Proposed changes

In UCP 3.1 and going forward the Swarm and Kubernetes RBAC models are very different. In UCP 3.0 we didn't support Kubernetes RBAC objects, therefore Kubernetes RBAC objects were translated to UCP / Swarm Objects and most of the RBAC documentation still applied. Prior to UCP 3.0, we only supported Swarm, so all docs were fine :) 

Following on from this issue https://github.com/docker/docker.github.io/issues/7828 and relevant PR https://github.com/docker/docker.github.io/pull/8309, we are starting to see the pain of having Swarm and Kubernetes tutorials on the same Docs pages. The Swarm and Kube RBAC models within UCP are now very different so trying to describe both models on the same page gets very confusing.... :(

This initial commit is purely for feedback on the ToC design. The content hasn't been split yet, and there will be a lot of dead links related to my changes to the ToC. Following https://github.com/docker/docs-private/issues/795 I have gone for:

Today
```
-> Authorization
---> The Basics
---> Tutorials and Use Cases
```

Proposed Design
```
-> Authorization
--> External Auth
--> Kubernetes
---> Concepts
---> Tutorials and Use Cases
--> Swarm
---> Concepts
---> Tutorials and Use Cases
```

I went through all existing RBAC documentation and tried to put it into those Buckets. I also found some Kubernetes RBAC documentation and External Auth documentation under UCP -> Admin -> Configure, and brought that under RBAC. Putting all RBAC documentation in one place. 

One of the things to consider is a lot of the Kubernetes RBAC documentation on kubernetes.io is solid, so we don't need to replicate many of the basic concepts definitions here. However I think the content focus on the kubernetes side would be more on tutorials / use cases, like the use case suggested in https://github.com/docker/docker.github.io/pull/8309.

@bermudezmt @paigehargrave WDYT of the Design? Its going to take me a bit of time to split out the content but wanted consensus that 1) this is a good idea to split out Kube and Swarm, 2) this Toc design is ok before I dive head first into it :)


bermudezmt:
@ollypom I like the direction of the TOC. Thank you - and let me or @paigehargrave know how we can help.

PR #8511, opened 2019-03-21
Update index.md [Sample applications table]

zeniri:
Providing entry to dockerize .Net Core application from the left navigation panel into the "Sample applications" table

### Proposed changes
.Net Core application entry needs to be available in Sample application table which differs from the below entry of ASP.NET Core entry as it uses a database whereas the .Net Core application doesn't thus, making its entry inside the table. Moreover the same entry is visible in the left navigation panel.


zeniri:
My Pleasure @bermudezmt :)

PR #8518, opened 2019-03-22
env added

dbaklikov:
### Proposed changes

Fixes "OSError: [Errno 8] Exec format error" issue.
https://stackoverflow.com/questions/55271912/flask-cli-throws-oserror-errno-8-exec-format-error-when-run-through-docker

### DIFF
--- a/compose/gettingstarted.md
+++ b/compose/gettingstarted.md
@@ -26,6 +26,7 @@ Define the application dependencies.
 
 2.  Create a file called `app.py` in your project directory and paste this in:
 
+        #!/usr/bin/env python
         import time
 
         import redis


davidism:
I'd prefer #8609 as it is a more comprehensive update to the tutorial. I'm one of the maintainers of Flask.

davidism:
This should be closed, it was addressed by the PR I linked above.

compose/gettingstarted.md
```diff
@@ -26,6 +26,7 @@ Define the application dependencies.
 
 2.  Create a file called `app.py` in your project directory and paste this in:
 
+        #!/usr/bin/env python
```
thaJeztah, line 29: Was there a reason for adding this? This file will not be used as an executable script, but is called in the dockerfile with;

```dockerfile
CMD ["python", "app.py"]
```

(so it's run by python, and not in a shell)

davidism: Yes, at one point Werkzeug/Flask had an issue with its reloader on Windows because volumes mount all files as 777 since Windows doesn't have an executable bit. This is no longer needed because it was fixed in Werkzeug. (That `CMD` line may have been changed to fix it too, can't remember.)

thaJeztah: Oh! For some reason, github opened on the changes tab, and I didn't see your earlier comment about the change that was already merged

Alright; let's close this one; thanks!

PR #8412, opened 2019-03-06
Compose: put Windows Server in its own tab for installation

thaJeztah:
follow-up to https://github.com/docker/docker.github.io/pull/8411, but kept it in a separate PR in case this needed discussion 👍 

With this change:

screenshot 2019-03-06 at 14 02 11

PR #8519, opened 2019-03-22
Fixing download location

ekitagawa:
Download command below points to a dir that doesn't exist
```
Invoke-WebRequest "https://github.com/docker/compose/releases/download/1.23.2/docker-compose-Windows-x86_64.exe" -UseBasicParsing -OutFile $Env:ProgramFiles\Docker\Docker\resources\bin\docker-compose.exe
```
On Windows Server 2019 (and probably on 2016), other exe files are under "$Env:ProgramFiles\Docker\" so that should be a good place to put the compose's executable.  This directory is also registered in "PATH" of the system, so users can call out `docker-compose --version` command on the next step without additional configuration.


bermudezmt:
Thank you @ekitagawa ! Adding your note here for reference: "The command pointed to the directory that doesn't exist, so I filed https://github.com/docker/docker.github.io/pull/8519". #8412

PR #7353, opened 2018-09-11
Added Azure Config File

ollypom:
### Proposed changes
Added some Azure Infrastructure prerequisites

Added the instructions to configure the Azure Config File.

PR #8522, opened 2019-03-22
UCP Install on Azure Patch

ollypom:
### Proposed changes

I updated the grammar on the 2nd pre-req from

```
The Azure Vnet and Subnet must be appropriately sized for your environment, and addresses from this pool are consumed by Kubernetes Pods.
```

to
```
The Azure Vnet and Subnet must be appropriately sized for your environment, as addresses from this pool will be consumed by Kubernetes Pods.
```

I also did some raw markdown formatting on the rest of the pre-reqs :)

### DIFF
--- a/ee/ucp/admin/install/install-on-azure.md
+++ b/ee/ucp/admin/install/install-on-azure.md
@@ -211,8 +211,7 @@ The `--pod-cidr` option maps to the IP address range that you configured for
 the subnets in the previous sections, and the `--host-address` maps to the
 IP address of the master node.
 
-> Note: The `pod-cidr` range must be within an Azure subnet attached to the
-> host.
+> Note: The `pod-cidr` range must match the Azure Vnet Subnet attached the hosts
 
 ```bash
 docker container run --rm -it \
--- a/ee/ucp/admin/install/install-on-azure.md
+++ b/ee/ucp/admin/install/install-on-azure.md
@@ -43,12 +43,6 @@ to successfully deploy Docker UCP on Azure:
   Configuration](#considerations-for-ipam-configuration).
 - All UCP worker and manager nodes need to be attached to the same Azure
   Subnet.
-- All UCP workers and managers need to have the `Orchestrator` tag in Azure with
-  the Kubernetes version as the value, following this format:
-  `Orchestrator=Kubernetes:x.y.z`. This value may change in each UCP release. To
-  find the relevant version, see the UCP [Release
-  Notes](/ee/ucp/release-notes). For example, the tag for UCP `3.1.0` would be
-  `Orchestrator=Kubernetes:1.11.2`. 
 - The Azure Virtual Machine Object Name needs to match the Azure Virtual Machine
   Computer Name and the Node Operating System's Hostname which is the FQDN of
   the host, including domain names. Note that this requires all characters to be in lowercase.
--- a/ee/ucp/admin/install/install-on-azure.md
+++ b/ee/ucp/admin/install/install-on-azure.md
@@ -66,8 +66,8 @@ objects are being deployed.
 
 For Docker UCP to integrate with Microsoft Azure,each UCP node in your cluster
 needs an Azure configuration file, `azure.json`. Place the file within
-`/etc/kubernetes`, and as the `azure.json` file is mounted into the Kubelet
-container it needs at least 0644 permissions.
+`/etc/kubernetes`. Since the config file is owned by `root`, set its permissions 
+to `0644` to ensure the container user has read access.
 
 The following is an example template for `azure.json`. Replace `***` with real values, and leave the other
 parameters as is.
--- a/ee/ucp/admin/install/install-on-azure.md
+++ b/ee/ucp/admin/install/install-on-azure.md
@@ -70,9 +70,10 @@ objects are being deployed.
 
 ### Azure Configuration File
 
-For Docker UCP to integrate with Microsoft Azure,each UCP node in your cluster needs an Azure 
-configuration file, `azure.json`. Place the file within `/etc/kubernetes`, and give it 0644 
-permissions.
+For Docker UCP to integrate with Microsoft Azure,each UCP node in your cluster
+needs an Azure configuration file, `azure.json`. Place the file within
+`/etc/kubernetes`, and as the `azure.json` file is mounted into the Kubelet
+container it needs at least 0644 permissions.
 
 The following is an example template for `azure.json`. Replace `***` with real values, and leave the other
 parameters as is.
@@ -276,7 +277,9 @@ This Swarm Config will need to be named `com.docker.ucp.config`.
     igca3q30jz9u3e6ecq1ckyofz   com.docker.ucp.config                                     1 days ago          1 days ago
     ```
 
-5. You are now ready to [install UCP](#install-ucp). 
+5. You are now ready to [install UCP](#install-ucp). As you have already staged
+   a UCP configuration file, you will need to add `--existing-config` to the
+   install command below.  
 
 If you need to adjust this value post-installation, see [instructions](../configure/ucp-configuration-file/)
 on how to download the UCP configuration file, change the value, and update the configuration via the API. 
@@ -288,7 +291,9 @@ reconciled, and you will have to manually edit the IP count in Azure.
 Run the following command to install UCP on a manager node. The `--pod-cidr`
 option maps to the IP address range that you have configured for the Azure
 subnet, and the `--host-address` maps to the private IP address of the master
-node.
+node. Finally if you have set the [Ip Count
+Value](#adjusting-the-ip-count-value) you will need to add `--existing-config`
+to the install command below.
 
 > Note: The `pod-cidr` range must match the Azure Virtual Network's Subnet
 > attached the hosts. For example, if the Azure Virtual Network had the range
--- a/ee/ucp/admin/install/install-on-azure.md
+++ b/ee/ucp/admin/install/install-on-azure.md
@@ -13,26 +13,21 @@ Calico / Azure integration.
 
 ## Docker UCP Networking
 
-Docker UCP configures the Azure IPAM module for Kubernetes to allocate
-IP addresses to Kubernetes pods.  The Azure IPAM module requires each Azure
-virtual machine that's part of the Kubernetes cluster to be configured with a pool of
-IP addresses.
+Docker UCP configures the Azure IPAM module for Kubernetes to allocate IP
+addresses to Kubernetes pods.  The Azure IPAM module requires each Azure virtual
+machine that's part of the Kubernetes cluster to be configured with a pool of IP
+addresses.
 
 There are two options for provisoning IPs for the Kubernetes cluster on Azure
-- Docker UCP provides an automated mechanism to configure and maintain IP pools 
-  for standalone Azure virtual machines. This service runs within the calico-node daemonset 
-  and by default will provision 128 IP address for each node. This value can be 
-  configured through the `azure_ip_count`in the UCP 
-  [configuration file](../configure/ucp-configuration-file) before or after the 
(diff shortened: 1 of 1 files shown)


bermudezmt:
@xinfengliu @stevenfollis can you PTAL at this PR which should hopefully address your concerns on the recent UCP-Azure docs update?

ee/ucp/admin/install/install-on-azure.md
```diff
@@ -211,8 +209,7 @@ The `--pod-cidr` option maps to the IP address range that you configured for
 the subnets in the previous sections, and the `--host-address` maps to the
 IP address of the master node.
 
-> Note: The `pod-cidr` range must be within an Azure subnet attached to the
-> host.
+> Note: The `pod-cidr` range must match the Azure Vnet Subnet attached the hosts
```
stevenfollis: Pedantic but if we're adjust should it read `Azure Virtual Network Subnet`?

stevenfollis: would apply to more than just subnets actually, as `vnets` is in use throughout doc

ee/ucp/admin/install/install-on-azure.md
```diff
@@ -36,31 +36,29 @@ There are two options for provisoning IPs for the Kubernetes cluster on Azure
...
-Group hosting the UCP Nodes. Note, if using a separate networking Resource 
-Group the same Service Principal will need `Network Contributor` access to this 
-Resource Group.
+- All UCP worker and manager nodes need to be attached to the same Azure
+  Subnet.
+- All UCP workers and managers need to have the `Orchestrator` tag in Azure with the Kubernetes version as the value, following this format: `Orchestrator=Kubernetes:x.y.z`. This value may change in
+  each UCP release. To find the relevant version please see the UCP [Release
+  Notes](../../release-notes). For example for UCP `3.1.0` the tag would be
+  `Orchestrator=Kubernetes:1.11.2`. 
+- The Azure Virtual Machine Object Name needs to match the Azure Virtual Machine
```
xinfengliu: I would like to add one more requirement for Azure VM name and OS hostname: _the name must be all lowercased_. In k8s world, the node name will be converted to lowercases, causing mismatch with Azure VM name which is not lowercased. We had customer cases for this issue.

Just curious, what's the difference between `Azure Virtual Machine Object Name` and `Azure Virtual Machine Computer Name`?

stevejr:
Can we add one more update in stating that the Azure VNet does not need to equal the subnet but that the pod cidr must equal the subnet.

ee/ucp/admin/install/install-on-azure.md
ollypom: I've gone through an expanded `vnets` to `Virtual Networks` everywhere I can :)

ee/ucp/admin/install/install-on-azure.md
ollypom: I've added the lowercase comment to the pre-reqs. 

I may have named these wrong but Object Name (found in resource ID) and the Virtual Machine Computer name can be different. I've often seen these different as part of ARM template deployments. @stevenfollis does this sound right?

The Kubernetes code is first checking the ID, I.e 

```
"id": "/subscriptions/<sub-i>/resourceGroups/<resource-group>/providers/Microsoft.Compute/virtualMachines/manager-01"
```

I think it gets this value from the VM hostname. The Azure Computer name by default is the OS Hostname.

In the Portal - Red Line = ID, Blue Line = Computer Name

image

ollypom:
@stevejr @stevenfollis @xinfengliu Do you mind taking a look at the latest commit. Including the new section on configuring the IP Count variable. I will also add a bit more meat to the IPAM considerations section shortly.

ee/ucp/admin/install/install-on-azure.md
```diff
@@ -13,54 +13,48 @@ Calico / Azure integration.
...
 
-- All UCP Nodes (Managers and Workers) need to be deployed into the same 
-Azure Resource Group. The Azure Networking (Vnets, Subnets, Security Groups) 
-components could be deployed in a second Azure Resource Group.
-- The Azure Vnet and Subnet must be appropriately sized for your
-  environment, and addresses from this pool are consumed by Kubernetes Pods. For more information, see
-  [Considerations for IPAM
+- All UCP Nodes (Managers and Workers) need to be deployed into the same Azure
+  Resource Group. The Azure Networking components (Virtual Network, Subnets,
+  Security Groups)  could be deployed in a second Azure Resource Group.
```
stevenfollis: In a second Azure Resource Group but in the same Azure Subscription, correct? Unaware if we support cross-subscription RGs but suspect now (have had customers with 101 Azure Subs and cross-sub anything is painful)

ee/ucp/admin/install/install-on-azure.md
```diff
@@ -206,13 +200,59 @@ for each virtual machine in the virtual machine scale set.
...
+  azure_ip_count = "20" # Note this value may be different for your environment
+```
+
+To pre-load this configuration file prior to a UCP installation, connect to the Virtual Machine where you would you like to install UCP and initiate a Swarm. 
+
+```
+$ docker swarm init
+```
+
+Then preload the UCP configuration file as a Swarm Config.
```
stevenfollis: The section felt choppy to me. We show a bunch of config examples but should take a moment to explain the context of what a config object does. If we need to create one, let's go ahead and instruct them to do a `touch azure-config.toml` or similar and be specific on if they need to take an entire existing config and edit it, or if it's sufficient to just use a 2-liner such as the above's:
```
[cluster_config]
  azure_ip_count = "20"
```

ee/ucp/admin/install/install-on-azure.md
ollypom: Yes, the second RG is in the same subscription. I don't think there is a field to support a second subscription in the azure.json.

ee/ucp/admin/install/install-on-azure.md
ollypom: @stevenfollis Sorry to be pain but do you mind taking a second look? I've expanded this section out a bit :)

ee/ucp/admin/install/install-on-azure.md
```diff
@@ -13,54 +13,52 @@ Calico / Azure integration.
...
+  the Kubernetes version as the value, following this format:
+  `Orchestrator=Kubernetes:x.y.z`. This value may change in each UCP release. To
+  find the relevant version please see the UCP [Release
+  Notes](../../release-notes). For example for UCP `3.1.0` the tag would be
+  `Orchestrator=Kubernetes:1.11.2`. 
+- The Azure Virtual Machine Object Name needs to match the Azure Virtual Machine
+  Computer Name and the Node Operating System's Hostname (This is the FQDN of
+  the host, including domain names). Note this value is case sensitive, and all
+  values should be in lowercase.
+- An Azure Service Principal with `Contributor` access to the Azure Resource
```
ollypom: @henrytk Does this clarification answer your question from https://github.com/docker/docker.github.io/issues/8551?

stevejr:
@ollypom - Was you going to state the minimum value the azure_ip_count should be?

ollypom:
@stevejr the minimum value of `azure_ip_count` can actually be 0, which is mentioned in this PR if you have manually allocated IP addresses.

I do plan to add to the "Considerations for IPAM.." section with examples around small subnets (which I think is what your asking :D). However I think i'm going to do that in a second PR as this ones already bitten off more than it planned too :)

stevenfollis:
The improved `Adjusting the IP Count Value` is looking really good, thanks @ollypom. Clear to follow, good examples of when `0` makes sense vs. other numbers for `auzre_ip_count`.

mark-church:
@ddebroy can you take a look at this for a review as well?

ee/ucp/admin/install/install-on-azure.md
```diff
@@ -4,81 +4,78 @@ description: Learn how to install Docker Universal Control Plane in a Microsoft
...
+- `aadClientId` - The Azure Service Principal ID.
+- `aadClientSecret` - The Azure Service Principal Secret Key.
 
 ### Azure Configuration File
 
-For Docker UCP to integrate into Microsoft Azure, you need to place an Azure 
-configuration file within each UCP node in your cluster, at 
-`/etc/kubernetes/azure.json`. The `azure.json` file needs 0644 permissions.
+For Docker UCP to integrate with Microsoft Azure,each UCP node in your cluster needs an Azure 
+configuration file, `azure.json`. Place the file within `/etc/kubernetes`, and give it 0644
```
bermudezmt: @ollypom can we clarify why the configuration file needs 0644 permissions?

ollypom: @bermudezmt No problem. Provided a little bit more context in the latest commit. Is that enough?

ollypom:
@dockerpac Great point. Thanks for pointing that out. I have added that in 2 places in the latest commit.

ee/ucp/admin/install/install-on-azure.md
bermudezmt: Thank you, @ollypom . Note to myself to revisit your Slack message and add a bit more detail around `0644` file permissions.

bermudezmt:
@ddebroy I noticed a lot of discussion and interest around this docs updates both here and on Slack. Can we please get a comprehensive technical review when you get a chance?

ee/ucp/admin/install/install-on-azure.md
```diff
@@ -4,81 +4,79 @@ description: Learn how to install Docker Universal Control Plane in a Microsoft
...
-- The Azure Virtual Machine Object Name needs to match the Azure Virtual Machine 
-Computer Name and the Node Operating System's Hostname. Note this applies to the 
-FQDN of the host including domain names. 
-- An Azure Service Principal with `Contributor` access to the Azure Resource 
-Group hosting the UCP Nodes. Note, if using a separate networking Resource 
-Group the same Service Principal will need `Network Contributor` access to this 
-Resource Group.
+- All UCP worker and manager nodes need to be attached to the same Azure
+  Subnet.
+- All UCP workers and managers need to have the `Orchestrator` tag in Azure with
```
ddebroy: I don't recall who exactly consumes the `orchestrator` label from Azure - core kubernetes, UCP controllers or Calico? If a node were to be switched from UCP to Swarm or vice versa in UCP UI, does the label need to be added/removed out-of-band as well? This needs to be clarified by someone from UCP. @wsong any thoughts?

wsong: I'm actually not familiar with this Orchestrator tag; UCP and Swarm certainly never consume it. Is this for the Kube cloud provider code, or does this tag somehow get translated to a Swarmkit node label?

ddebroy: Thanks Wayne. @ollypom PR: https://github.com/docker/docker.github.io/pull/7353 introduced the Orchestrator label as a pre-req for Azure nodes. Do you happen to recall the context around this - which component is the consumer of these labels?

ollypom: @ddebroy The original pre-req came from @jtestard's notes here https://github.com/docker/orca/issues/13772 for Azure Disk. I know that @joeabbey looked at it more detail and didn't think it was required https://github.com/docker/orca/issues/16019. As that issue remained opened though, I've never removed it from the Public docs as precaution.

bermudezmt:
@ddebroy Is this good to go?

ee/ucp/admin/install/install-on-azure.md
ddebroy: Thanks for the context, @ollypom . Can we remove this please?

ollypom: @ddebroy Thanks, I've removed this prereq.

ddebroy:
lgtm.

PR #8532, opened 2019-03-26
use access tokens for CLI access to DTR

omegamormegil:
Added note advising users to use access tokens with DTR after configuring SAML for UI authentication. This is necessary per slack conversation here: https://docker.slack.com/archives/C1329M30T/p1547584120183300


omegamormegil:
Please check the formatting of my note. The link might not be the correct format.

PR #8538, opened 2019-03-28
Documenting compose-inside-docker gotcha

reinout:
### Proposed changes

This is a documentation fix for https://github.com/docker/compose/issues/6506

When you use the dockerized docker-compose, a `~` in your `docker-compose.yml` doesn't expand to what you expect (`/home/myuser`), but to `/root` as that is the home dir inside the docker it is running in :-)

### Related issues

https://github.com/docker/compose/issues/6506

### DIFF
--- a/compose/install.md
+++ b/compose/install.md
@@ -200,6 +200,10 @@ $ sudo curl -L --fail https://github.com/docker/compose/releases/download/{{site
 $ sudo chmod +x /usr/local/bin/docker-compose
 ```
 
+Warning: as docker-compose runs inside a container, variable expansion in your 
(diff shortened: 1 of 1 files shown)


gqqnbig:
a345ccc reads better for me. "as" states a fact that docker-compose runs inside a container. "when" implies a possibility that sometimes docker-compose runs outside a container.

GitHunter0:
Good initiative, that caught me off guard too, I spent hours wrestling with this very issue...

usha-mandya:
Closing this PR as this use case isn't supported in Compose V2.

PR #8553, opened 2019-03-29
for-mac: removed references to daemon-proxy config propagating to containers

javabrett:
### Proposed changes

Updated the Docker for Mac Proxy settings description, to match current behaviour, removing references to previous behaviour where DD for M would auto-magically-propagate the proxy settings for the Docker daemon (for image pulls) to container build-time and run-time environment variables.

I believe/recall that this used to be the behaviour, and it was a bit magic and differed from Docker on Linux, where the Docker daemon's proxy settings are separate and do not propagate to container builds/runs (you need to set `--build-arg http_proxy=...` or `-e http_proxy=...`.  I don't believe source is public, and can't find any release note for this change, so can't verify when this change occurred.

If you do manually set some proxy settings (for the daemon) and restart, or have some system proxy settings, running the `docker run -it alpine env` in the existing documentation should prove that these settings no-longer propagate as claimed in the current documentation.

Fixed the incorrectly worded:
> Docker uses this proxy when pulling containers

Docker pulls images not containers.

DD for M Version 2.0.0.3 (31259).

### Related issues

Fixes https://github.com/docker/for-mac/issues/2715 .

### DIFF
--- a/docker-for-mac/index.md
+++ b/docker-for-mac/index.md
@@ -170,31 +170,22 @@ For more information, see:
 
 #### Proxies
 
-Docker Desktop detects HTTP/HTTPS Proxy Settings from macOS and automatically
-propagates these to Docker and to your containers. For example, if you set your
-proxy settings to `http://proxy.example.com`, Docker uses this proxy when
-pulling containers.
+HTTP/HTTPS Proxy Settings are used by the Docker daemon when pulling images
+from remote registries. Docker can inherit the macOS system proxy settings
+(which could be a direct-connection, no-proxy), or you can manually configure
+specific HTTP and HTTPS proxy servers.
 
-![Proxies settings](images/menu/prefs-proxies.png)
+Previous versions of Docker Desktop for Mac also propagated the Docker daemon
+proxy setting to container environments. This is no longer the case, and
+Docker Desktop for Mac now behaves the same as the Docker daemon on Linux. If
+your container build or run requires proxy settings, you must pass them
+as `--build-arg http_proxy=http://...` or `-e http_proxy=http://...`
+respectively.
 
-When you start a container, your proxy settings propagate into the containers.
-For example:
-
-```
-$ docker run -it alpine env
-PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
-HOSTNAME=b7edf988b2b5
-TERM=xterm
-HOME=/root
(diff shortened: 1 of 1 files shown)


windoverwater:
Any chance that this can be pushed a long?  On a Mac and stuck behind a proxy and cannot set the proxy as documented - it looks like this could be the issue.  Thanks!

javabrett:
Rebased.  Bump for review.

sruffell:
I have a question about this, since it appears that I'm seeing the proxy settings propagating to the containers.

When I look at the proxy setting screens I see "Configure the proxies used by the Docker Engine to pull images". I therefore thought that those settings were *only* used by docker engine when pulling images.

However when running curl both outside and inside the container (with my https_proxy environment variable set to the site proxy, I see the one outside coming from curl and the one inside coming from an ocaml-cohttp user agent when docker desktop is configured to use the proxy.

When curling inside the container:
```
Hypertext Transfer Protocol
    CONNECT 64.233.177.104:443 HTTP/1.1\r\n
        [Expert Info (Chat/Sequence): CONNECT 64.233.177.104:443 HTTP/1.1\r\n]
        Request Method: CONNECT
        Request URI: 64.233.177.104:443
        Request Version: HTTP/1.1
    host: 64.233.177.104:443\r\n
    user-agent: ocaml-cohttp/1.1.0\r\n
    \r\n
    [Full request URI: 64.233.177.104:443]
    [HTTP request 1/1]
    [Response in frame: 25]
```

When curling outside the container:
```
Hypertext Transfer Protocol
    CONNECT www.google.com:443 HTTP/1.1\r\n
        [Expert Info (Chat/Sequence): CONNECT www.google.com:443 HTTP/1.1\r\n]
        Request Method: CONNECT
        Request URI: www.google.com:443
        Request Version: HTTP/1.1
    Host: www.google.com:443\r\n
    User-Agent: curl/7.64.0\r\n
    Proxy-Connection: Keep-Alive\r\n
    \r\n
    [Full request URI: www.google.com:443]
    [HTTP request 1/1]
    [Response in frame: 46]
```

Is this a bug, or should there be a setting to opt into using the proxy settings for the containers globally or not?

usha-mandya:
Hey @javabrett Thanks for the update. The Proxies section has since been updated to indicate that proxy settings do not propagate to the containers - https://docs.docker.com/docker-for-mac/#proxies. I'll have to close this PR without merging the changes due to this reason. Feel free to continue the conversation after I close.

PR #8562, opened 2019-04-01
port change from 80 to 4000

Django987:
usually I think it is 4000, so I thought here it should be 4000, as well to keep it consistent


Django987:
since it is explained which ports to use and where they are defined, it is actually obvious and maybe sharpens the understanding of the role that docker-compose.yml plays

though it is little bit irritating, that the port in the .yml changes

PR #8567, opened 2019-04-01
Include Ubuntu version in Dockerfile

vsocrates:
### Proposed changes

Added a version to the example Dockerfile since newer versions of Ubuntu throw errors, such as `E: gnupg, gnupg2 and gnupg1 do not seem to be installed, but one of them is required for this operation
`


bermudezmt:
Thank you, @thaJeztah and @vsocrates .

PR #8577, opened 2019-04-02
Updated the UCP 3.1.4 release notes to include Centos 7.6 support

ollypom:
### Proposed changes

Updated the 3.1.4 release notes to include Centos 7.6 support

PTAL @andrewhsu @ry4nz

### DIFF
--- a/ee/ucp/release-notes.md
+++ b/ee/ucp/release-notes.md
@@ -69,7 +69,7 @@ upgrade your installation to the latest release.
 
 ### New platforms
 * Added support for SLES 15.
-* Added support for Oracle 7.6.
+* Added support for CentOS 7.6.
 
 ### Kubernetes
 * Kubernetes has been updated to version 1.11.7. (docker/orca#16157)


bermudezmt:
@ry4nz @andrewhsu can you please confirm that this is the case?

ee/ucp/release-notes.md
```diff
@@ -69,7 +69,7 @@ upgrade your installation to the latest release.
 
 ### New platforms
 * Added support for SLES 15.
-* Added support for Oracle 7.6.
```
bermudezmt: @ollypom as per @david-yu and https://success.docker.com/article/compatibility-matrix, Oracle 7.6 is still supported. Going to check on CentOS 7.6 with @andrewhsu...

ollypom: @bermudezmt Yep, but Oracle 7.6 was a "New Platform" in 3.1.3, mentioned further down in the release notes. Hence why I think adding it a New Platform in 3.1.4 was a typo, and instead it should of read CentOS :)

bermudezmt: Ah, great point. Still waiting on confirmation for CentOS 7.6 support. cc: @david-yu

bermudezmt:
Thank you so much, @ry4nz and @andrewhsu . 

As always, thanks for all the docs contributions, @ollypom .

PR #8580, opened 2019-04-03
Storage backend data migration updates

bermudezmt:
Addresses https://github.com/docker/docs-private/issues/1001.

- Adds details on new `docker/dtr reconfigure` flag
- Adds a new page for switching storage backends
- Nudges people to upgrade to 2.6.4 while retaining all warnings and flashing signs for previous wiped tags issue


ee/dtr/admin/configure/external-storage/nfs.md
```diff
@@ -53,12 +53,18 @@ To support **NFS v4**, more NFS options have been added to the CLI. See [New Fea
...
+In DTR 2.6.4, a new flag, `--storage-migrated`, [has been added to `docker/dtr reconfigure`](/reference/dtr/2.6/cli/reconfigure/) which lets you indicate the migration status of your storage data during a reconfigure. [Upgrade to 2.6.4](/reference/dtr/2.6/cli/upgrade/) and follow [Best practice for data migration in 2.6.4](/ee/dtr/admin/configure/external-storage/storage-backend-migration/#best-practice-for-data-migration) when switching storage backends.
+
 ```bash
 docker run --rm -it \
   docker/dtr:{{ page.dtr_version}} reconfigure \
   --ucp-url <ucp_url> \
   --ucp-username <ucp_username> \
-  --dtr-storage-volume <dtr-registry-nf>
+  --nfs-storage-url <dtr-registry-nf>
+  --async-nfs
```
caervs: Fyi `--async-nfs` is only for `NFSv4`. Not sure if that's relevant here.

ee/dtr/admin/configure/external-storage/s3.md
```diff
@@ -133,15 +133,26 @@ DTR supports the following S3 regions:
 
 ## Update your S3 settings on the web interface
 
-There is currently an issue with [changing your S3 settings on the web interface](/ee/dtr/release-notes#version-26) which leads to erased metadata. Make sure to [back up your DTR metadata](/ee/dtr/admin/disaster-recovery/create-a-backup/#back-up-dtr-metadata) before you proceed. 
+When running 2.5.x (with experimental garbage collection) or 2.6.0-2.6.3, there is an issue with [changing your S3 settings on the web interface](/ee/dtr/release-notes#version-26) which leads to erased metadata. Make sure to [back up your DTR metadata](/ee/dtr/admin/disaster-recovery/create-a-backup/#back-up-dtr-metadata) before you proceed.
```
caervs: The `S3` issue still exists in `2.6.4`; so far we have a fix for `NFS`.

ee/dtr/admin/configure/external-storage/storage-backend-migration.md
```diff
@@ -0,0 +1,78 @@
...
+
+Docker recommends the following steps for your storage backend and metadata migration:
+
+1. Disable garbage collection by selecting "Never" under **System > Garbage Collection**, so blobs referenced in the backup that you create continue to exist. See [Garbage collection](/ee/dtr/admin/configure/garbage-collection/) for more details.
+
+    ![](/ee/dtr/images/garbage-collection-0.png){: .img-fluid .with-border}
+
+2. [Back up your existing metadata](/ee/dtr/admin/disaster-recovery/create-a-backup/#back-up-dtr-metadata). See [docker/dtr backup](/reference/dtr/2.6/cli/backup/) for CLI command description and options. 
+
+3. Migrate the contents of your current storage backend to the new one you are switching to. For example, upload your NFS contents to an S3 bucket if you're switching from NFS to S3.
```
caervs: We don't have a fix for the `NFS -> S3` case yet, so I would recommend using an example like `moving your contents to a new NFS server`

ee/dtr/admin/configure/external-storage/storage-backend-migration.md
```diff
@@ -0,0 +1,78 @@
...
+- If you have a long maintenance window, you can skip some steps from above and do the following:
+
+    1. Put DTR in "read-only" mode. To do so, send the following API request:
+   
+        ```bash
+        curl -u <username>:$TOKEN -X POST "https://<dtr-url>/api/v0/meta/settings" -H "accept: application/json" -H "content-type: application/json" -d "{ \"readOnlyRegistry\": true }"
+        ``` 
+        On success, you should get a `202 Accepted` response.
+
+    2. Migrate the contents of your current storage backend to the new one you are switching to. For example, upload your NFS contents to an S3 bucket if you're switching from NFS to S3.
```
caervs: Same as above

ee/dtr/admin/configure/external-storage/storage-backend-migration.md
```diff
@@ -0,0 +1,78 @@
...
+
+    ![](/ee/dtr/images/garbage-collection-0.png){: .img-fluid .with-border}
+
+2. [Back up your existing metadata](/ee/dtr/admin/disaster-recovery/create-a-backup/#back-up-dtr-metadata). See [docker/dtr backup](/reference/dtr/2.6/cli/backup/) for CLI command description and options. 
+
+3. Migrate the contents of your current storage backend to the new one you are switching to. For example, upload your NFS contents to an S3 bucket if you're switching from NFS to S3.
+
+4. [Restore DTR from your backup](/ee/dtr/admin/disaster-recovery/restore-from-backup/) and specify your new storage backend. See [docker/dtr destroy](/reference/dtr/2.6/cli/destroy/) and [docker/dtr restore](/reference/dtr/2.6/cli/backup/) for CLI command descriptions and options.
+
+5. Run garbage collection in "blob" mode. You can either SSH into a DTR node, or [use a UCP client bundle](/ee/ucp/user-access/cli/) to run the following command:
```
cquon: Are we not doing the API method for this part?

ee/dtr/admin/configure/external-storage/storage-backend-migration.md
```diff
@@ -0,0 +1,78 @@
...
+
+    ```bash
+    docker exec -it dtr-jobrunner-<replica_id> sh
+    ```
+    > See [Find your replica ID](/ee/dtr/admin/disaster-recovery/create-a-backup/#find-your-replica-id) for tips on determining your replica ID.
+
+    Within the running container, type `/bin/job_executor` to start the `job_executor` binary.
+    
+    ```bash
+    /bin/job_executor
```
cquon: Is it better to just give them the command of `/bin/job_executor onlinegc blobs` ?

ee/dtr/admin/configure/external-storage/storage-backend-migration.md
bermudezmt: Do you remember what happened to dogfooding when we tried the API call? :fire:

ee/dtr/admin/configure/external-storage/storage-backend-migration.md
bermudezmt: I think so, great feedback. I'll get to yours and @caervs 's comments before EOD.

ee/dtr/admin/configure/external-storage/storage-backend-migration.md
```diff
@@ -0,0 +1,78 @@
...
+
+## DTR 2.6.4 and above
+
+In DTR 2.6.4, a new flag, `--storage-migrated`, [has been added to `docker/dtr reconfigure`](/reference/dtr/2.6/cli/reconfigure/) which lets you indicate the migration status of your storage data during a reconfigure.  
+
+### Best practice for data migration
+
+Docker recommends the following steps for your storage backend and metadata migration:
+
+1. Disable garbage collection by selecting "Never" under **System > Garbage Collection**, so blobs referenced in the backup that you create continue to exist. See [Garbage collection](/ee/dtr/admin/configure/garbage-collection/) for more details.
```
jose-bigio: Is the best practice to disable garbage collection for the duration of the migration?
I'm guessing that is the intention, but it could be made more clear.

ee/dtr/admin/configure/external-storage/storage-backend-migration.md
```diff
@@ -0,0 +1,78 @@
...
+    Within the running container, type `/bin/job_executor` to start the `job_executor` binary.
+    
+    ```bash
+    /bin/job_executor
+    
+    / # onlinegc blobs
+    ```
+    > Note that the first line results in a display of the binary name, its usage, and available commands including `onlinegc`. To learn more about a command, run `<command_name>` --help.
+
+    Running garbage collection in blob mode destroys any new blobs which are not referenced in your previously created backup.
```
jose-bigio: Can we add more details about what this means?
Is the purpose of running garbage collection in blob mode to destroy orphaned blobs that may have existed in the old storage backend?

Is there a reason why enabling Online Garbage Collection wouldn't be able to clean these blobs up?
Similar question that I asked before, should Online Garbage Collection not be turned on anymore, or once the user has confirmed that their migration is complete should they turn online GC back on?

bermudezmt: @caervs can we get your input? I wasn't sure whether we need to share as much detail as possible around blob mode but if I remember Ryan's comment from stand-up, taking care of orphaned blobs should just be part of garbage collection in the future. Right now, when you run `/bin/job_executor`, onlinegc gives you the following options:
```
all       Garbage collect all types
     metadata  Garbage collect image metadata
     joblogs   Garbage collect job logs
     events    Garbage collect events
     blobs     Garbage collect dangling blobs
```

ee/dtr/admin/configure/external-storage/storage-backend-migration.md
```diff
@@ -0,0 +1,78 @@
...
+        ```bash
+        curl -u <username>:$TOKEN -X POST "https://<dtr-url>/api/v0/meta/settings" -H "accept: application/json" -H "content-type: application/json" -d "{ \"readOnlyRegistry\": true }"
+        ``` 
+        On success, you should get a `202 Accepted` response.
+
+    2. Migrate the contents of your current storage backend to the new one you are switching to. For example, upload your NFS contents to an S3 bucket if you're switching from NFS to S3.
+
+   3. [Reconfigure DTR](/reference/dtr/2.6/cli/reconfigure) while specifying the `--storage-migrated` flag to preserve your existing tags. 
+
+- If you are not worried about inconsistencies in data, skip steps 1 and 2 and perform a reconfigure.
```
jose-bigio: I don't know if I am understanding correctly. 

Is this basically saying, if you don't care about your images then you can run a reconfigure?

Data inconsistencies makes it sound like data will be retained, but it could be wrong. 

Is the below comment indicating that there are data inconsistencies? 

> Starting in DTR 2.6, switching storage backends initializes a new metadata store and erases your existing tags

Or is this comment indicating that your images will get deleted?

ee/dtr/admin/configure/external-storage/storage-backend-migration.md
caervs: Indeed. We should have better user exposure  `blob` mode in a future release.

We need the admin to do this now because Imagine the following user scenario:

```
1. Admin takes back-up
2. Admin disables online gc
3. Admin starts migrating content to new server (this could take hours or even days)
4. Users push new tags and delete old tags
5. Storage migration gets to a blob that is now orphaned but still referenced in the back-up. Thankfully this blob still exists because GC is disabled
6. However storage migration also copies over the new blobs that have been pushed since back-up because it's dumb and just enumerates over all the blobs in the backend
7. So when restore is complete, the admin runs blob mode gc to get rid of the new orphaned blobs
```

PR #8584, opened 2019-04-03
Dynamic host port -> specific container port

adamrp:
Adds an example of how to publish a specific container port to a dynamically allocated host port that's *only* available to the host (and not public). This took me a while to google and find on stack overflow.

### Proposed changes

1-line documentation change to highlight a previously difficult-to-find feature that is probably of interest to the general public.


usha-mandya:
@adamrp Thanks for the PR. The `master` branch has been renamed to `main`. Some of the existing PRs have been automatically closed as a result of this change. Could you create a new PR against our latest docs if the suggested changes are still relevant?  Thank you.

PR #8591, opened 2019-04-04
Add HSTS warning for specifying --dtr-external-url

bermudezmt:
Addresses https://docker.atlassian.net/browse/ENGDTR-256.

@caervs do you think this is sufficient to address the HSTS issue?


bermudezmt:
Thank you, @caervs !

bermudezmt:
Future work: Add example error message:
Chrome:
```
You cannot visit <dtr-url> right now because the website uses HSTS. Network errors and attacks are usually temporary, so this page will probably work later.
```

PR #8609, opened 2019-04-07
Update Python and Flask usage in Compose tutorial

davidism:
* Uses the modern Python 3.7 image, as 3.4 is EOL.
* Use `COPY` instead of `ADD`, since behavior of `ADD` is not used, as recommended in the `Dockerfile` reference.
* Separates copying and installing requirements from copying project, to make rebuilds more efficient if no requirements change.
* Uses the recommended `flask run` command. This is especially needed on Windows, where `app.py` incorrectly looks like an executable file when copying into Docker. https://github.com/pallets/werkzeug/issues/1482
* Uses the `FLASK_ENV` env var to control development mode (debugger, reloader). This is only set when demonstrating the `volumes` option for live updating files, as it should not be used in production. Note that the development server (`flask run`, `app.run()`) should not be used in production either, but I think that's outside the scope of this change.
* Install gcc and dev headers so simple C extensions such as MarkupSafe and SQLAlchemy speedups can compile on Alpine (wheels on pypi are only built for glibc).

### DIFF
--- a/compose/gettingstarted.md
+++ b/compose/gettingstarted.md
@@ -86,6 +86,7 @@ following:
     WORKDIR /code
     ENV FLASK_APP app.py
     ENV FLASK_RUN_HOST 0.0.0.0
+    RUN apk add --no-cache gcc musl-dev linux-headers
     COPY requirements.txt requirements.txt
     RUN pip install -r requirements.txt
     COPY . .
@@ -96,6 +97,7 @@ This tells Docker to:
 * Build an image starting with the Python 3.7 image.
 * Set the working directory to `/code`.
 * Set environment variables used by the `flask` command.
+* Install gcc so Python packages such as MarkupSafe and SQLAlchemy can compile speedups.
 * Copy `requirements.txt` and install the Python dependencies.
 * Copy the current directory `.` in the project to the workdir `.` in the image.
 * Set the default command for the container to `flask run`.
--- a/compose/gettingstarted.md
+++ b/compose/gettingstarted.md
@@ -86,19 +86,23 @@ itself.
 In your project directory, create a file named `Dockerfile` and paste the
 following:
 
-    FROM python:3.4-alpine
-    ADD . /code
+    FROM python:3.7-alpine
     WORKDIR /code
+    ENV FLASK_APP app.py
+    ENV FLASK_RUN_HOST 0.0.0.0
+    COPY requirements.txt requirements.txt
     RUN pip install -r requirements.txt
-    CMD ["python", "app.py"]
+    COPY . .
+    CMD ["flask", "run"]
 
 This tells Docker to:
 
-* Build an image starting with the Python 3.4 image.
-* Add the current directory `.` into the path `/code` in the image.
+* Build an image starting with the Python 3.7 image.
 * Set the working directory to `/code`.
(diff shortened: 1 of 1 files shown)


compose/gettingstarted.md
```diff
@@ -86,19 +82,23 @@ itself.
...
 
 This tells Docker to:
 
-* Build an image starting with the Python 3.4 image.
-* Add the current directory `.` into the path `/code` in the image.
+* Build an image starting with the Python 3.7 image.
 * Set the working directory to `/code`.
-* Install the Python dependencies.
-* Set the default command for the container to `python app.py`.
+* Set environment varialbes used by the `flask` command.
```
chris-crone: `s/varialbes/variables/`

compose/gettingstarted.md
```diff
@@ -158,13 +158,13 @@ Hub registry.
...
-2.  Enter `http://0.0.0.0:5000/` in a browser to see the application running.
+2.  Enter http://localhost:5000/ in a browser to see the application running.
 
     If you're using Docker natively on Linux, Docker Desktop for Mac, or Docker Desktop for
     Windows, then the web app should now be listening on port 5000 on your
-    Docker daemon host. Point your web browser to `http://localhost:5000` to
+    Docker daemon host. Point your web browser to http://localhost:5000 to
     find the `Hello World` message. If this doesn't resolve, you can also try
-    `http://0.0.0.0:5000`.
+    http://127.0.0.1:5000 or http://0.0.0.0:5000.
```
chris-crone: I don't think we need `127.0.0.1` here as it's the same as localhost.

davidism: I've noticed that sometimes Docker will only bind to `localhost` or `127.0.0.1`, not both for some reason. However, `0.0.0.0` would never be correct, so I'll remove that.

compose/gettingstarted.md
```diff
@@ -86,19 +82,25 @@ itself.
...
 This tells Docker to:
 
-* Build an image starting with the Python 3.4 image.
-* Add the current directory `.` into the path `/code` in the image.
+* Build an image starting with the Python 3.7 image.
 * Set the working directory to `/code`.
-* Install the Python dependencies.
-* Set the default command for the container to `python app.py`.
+* Set environment variables used by the `flask` command.
+* Install gcc so Python packages such as MarkupSafe and SQLAlchemy can compile speedups.
```
chris-crone: nit: Since this is just a Compose get started guide, I'm not sure we need to include Python optimisations.

davidism: Since this is one of two official Docker documents introducing devs to using Flask with Docker, I'd rather they get a reasonable experience without having to discover it later, especially since it's non-obvious that speedups are not compiled. Overall I think the focus is still on Compose.

davidism: Do you agree, or would you like me to change something?

chris-crone: My concern is that we're conflating getting started with Compose and optimising a Python app. That might overload new users with too much information.

I did LGTM the PR though as it's net better than what we have currently.

@usha-mandya is there anything else we need?

usha-mandya: Can't think of anything else, @chris-crone. Thanks.

PR #8623, opened 2019-04-10
Add the 'Install on Azure' page back to the TOC for UCP 3.0

ollypom:
### Proposed changes

@bermudezmt I found it! 

So it appears the doc was still here, just got removed from the ToC. In this PR I add it back to the TOC and have just reworded the install command at the bottom to keep it in line with the UCP 3.1 Azure PR.


bermudezmt:
@ollypom Thank you!!

PR #8630, opened 2019-04-11
Use site parameter to use latest compose file versions in examples

thaJeztah:
### Proposed changes

Make sure that examples use the latest version of the compose file format, to encourage using the latest version, and to prevent users from running into "not supported by this version" problems when copy/pasting, and combining examples that use different versions.

Also add a note about `version: x` not being equivalent to `version: x.latest`, address some v3/v2 issues, and YAML syntax error

> **Note that there are still some examples using fixed versions** in the UCP sections; we need to evaluate those to make sure the right (and supported) versions are used for UCP (which may be different than "latest").


thaJeztah:
ping @ulyssessouza @chris-crone 


Also; I could use some help on the UCP examples, as I'm not sure what is supported 😅 (grep for `version: ` to find remaining "fixed" versions).

thaJeztah:
@bermudezmt @paigehargrave ptal

compose/compose-file/compose-versioning.md
```diff
@@ -121,9 +121,9 @@ discoverable at a hostname that's the same as the service name. This means
 [links](compose-file-v2.md#links) are largely unnecessary. For more details, see
 [Networking in Compose](compose-file-v2.md#networking.md).
 
-> **Note**: when specifying the compose file version to use, make sure to
-> specify both the "major" and "minor" version. If no minor version is given,
-> it defaults to `0`, not the "latest" minor version, and features added in
+> **Note**: When specifying the Compose file version to use, make sure to
+> specify both the _major_ and _minor_ numbers. If no minor version is given,
+> `0` is used by default and not the latest minor version. As a result, features added in
```
thaJeztah: Thanks! Looks better 👍 

Was later thinking; perhaps this note should be put somewhere at the top as well (or instead)

compose/compose-file/compose-versioning.md
```diff
@@ -121,9 +121,24 @@ discoverable at a hostname that's the same as the service name. This means
 [links](compose-file-v2.md#links) are largely unnecessary. For more details, see
 [Networking in Compose](compose-file-v2.md#networking.md).
 
+> **Note**: When specifying the Compose file version to use, make sure to
```
bermudezmt: This is a much-needed update, @thaJeztah . Thank you for taking it on. Before I merge, please have a look at my minor stylistic and text edits.

compose/compose-file/compose-versioning.md
bermudezmt: Aww, you beat me to it. :)

compose/compose-file/compose-versioning.md
thaJeztah: Yes, I kept seeing people being confused (or making this mistake due to copy/pasting old examples), so thought to call it out explicitly 🤗 

Saves me some time as well in the long run explaining it 😂

compose/compose-file/compose-versioning.md
thaJeztah: Also; while you're at it; could you update the other instance below as well?

compose/compose-file/compose-versioning.md
bermudezmt: Done. Have a look at the final change. Ok to squash and merge this one, @thaJeztah ?

PR #8639, opened 2019-04-11
Removing internal JIRA links

darwin-traver:
### Proposed changes
Removed the internal JIRA links (leaving just the ENGORC-## as the reference).


paigehargrave:
@darwin-traver Thank you!

PR #8654, opened 2019-04-16
Document stage inheritance

pieterdd:
### Proposed changes

I recently discovered a hidden feature for multi-stage builds in a [Medium article](https://medium.com/@tonistiigi/advanced-multi-stage-build-patterns-6f741b852fae#b3d3) that is really useful to us, but as far as I know wasn't officially documented yet. I added a paragraph or two about how it works and what its use case is.

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -191,9 +191,6 @@ fetch-remote:
     default_branch: "master"
     ref: "master"
     paths:
-      - dest: "build/customize/bake"
-        src:
-          - "docs/manuals/bake/**"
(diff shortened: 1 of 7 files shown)


jedevc:
Using previous stages in the `FROM` directives is already documented here: https://github.com/docker/docker.github.io/blob/master/develop/develop-images/multistage-build.md#use-a-previous-stage-as-a-new-stage.

I think explicitly calling out how ENV variables are inherited is a good idea though, though it should be under that section rather that right after the named stages.

PR #8665, opened 2019-04-18
ubuntu.md: remove old docker-ce-cli

kolyshkin:
I hit the following error when "upgrading" docker-ce 18.09 to docker-ee 17.06:

> dpkg: error processing archive /var/cache/apt/archives/docker-ee_3%3a17.06.2~ee~19~3-0~ubuntu_amd64.deb (--unpack):
 trying to overwrite '/usr/share/fish/vendor_completions.d/docker.fish', which is also in package docker-ce-cli 5:18.09.4~2.1.rc1-0~ubuntu-xenial

This commit adds `docker-ce-cli` to the list in "uninstall old packages" to fix this.

### DIFF
--- a/install/linux/docker-ee/ubuntu.md
+++ b/install/linux/docker-ee/ubuntu.md
@@ -47,7 +47,7 @@ Older versions of Docker were called `docker` or `docker-engine`. In addition,
 if you are upgrading from Docker CE to Docker EE, remove the Docker CE package.
 
 ```bash
-$ sudo apt-get remove docker docker-engine docker-ce docker.io
+$ sudo apt-get remove docker docker-engine docker-ce docker-ce-cli docker.io
 ```
 
 It's OK if `apt-get` reports that none of these packages are installed.


thaJeztah:
Actually; @kolyshkin, could you also update the other distros (at least debian, and the docker-ce variants?)

kolyshkin:
> could you also update the other distros (at least debian, and the docker-ce variants?)

Sorry it took me so long to get to it.

Apparently, docker-ce variants does not need it (to my best understanding) as those repos do _provide_ docker-ce-cli and this will be a smooth upgrade.

docker-ee for rhel and centos already has `yum remove ... docker-client` which I guess is what's needed.

In general, in order to figure this out one would need to make a matrix out of ce and ee releases and just give it a try. I might take a look at CE and EE packages once I figure out how to access -ee repos without asking for a key each time.

So, for now, this PR is as good as it gets.

bermudezmt:
Thank you, @kolyshkin and @thaJeztah

PR #8668, opened 2019-04-18
Remove deadlink in the Interlock ToC

ollypom:
### Proposed changes

Following on from this commit https://github.com/docker/docker.github.io/commit/a1707d5b1abe6ea1b80978faeaacef827d1c367b#diff-638c09f3fbe609cac9fc081b266f2f93 the default-service.md page was removed with the content moving into here https://docs.docker.com/ee/ucp/interlock/usage/interlock-vip-mode/#usage. However the ToC was missed from that commit, so just picking that up here along with a redirect.


ee/ucp/interlock/usage/interlock-vip-mode.md
```diff
@@ -4,6 +4,7 @@ description: Learn about task and VIP backend routing modes for Layer 7 routing
 keywords: routing, proxy, interlock
 redirect_from:
   - https://interlock-dev-docs.netlify.com/usage/default_backend/
```
bermudezmt: @ollypom redirecting from an external site actually doesn't work. I know you didn't add this - but I'll also remove. Thanks for the fix!

bermudezmt: External site redirect PR reverted via #8670.

PR #8681, opened 2019-04-19
Modified command for OSX compatibility

omegamormegil:
OSX doesn't support `grep -m1 -Po '(?<=DTR_VERSION=)\d.\d.\d')` so modified the command to use `grep -m1 DTR_VERSION | sed 's/.*DTR_VERSION=\([0-9].[0-9].[0-9]\)",/\1/')` which works on Linux and OSX.

### DIFF
--- a/reference/dtr/2.6/cli/backup.md
+++ b/reference/dtr/2.6/cli/backup.md
@@ -27,7 +27,8 @@ docker run -i --rm --log-driver none docker/dtr:{{ page.dtr_version }} \
 {% raw %}
 ```bash
 DTR_VERSION=$(docker container inspect $(docker container ps -f \
-  name=dtr-registry -q) | grep -m1 -Po '(?<=DTR_VERSION=)\d.\d.\d'); \
+  name=dtr-registry -q) | grep -m1 DTR_VERSION | \
+  sed 's/.*DTR_VERSION=\([0-9].[0-9].[0-9]\)",/\1/'); \
 REPLICA_ID=$(docker ps --filter name=dtr-rethinkdb \
   --format "{{ .Names }}" | head -1 | sed 's|.*/||' | sed 's/dtr-rethinkdb-//'); \
 read -p 'ucp-url (The UCP URL including domain and port): ' UCP_URL; \


reference/dtr/2.6/cli/backup.md
```diff
@@ -27,7 +27,8 @@ docker run -i --rm --log-driver none docker/dtr:{{ page.dtr_version }} \
 {% raw %}
 ```bash
 DTR_VERSION=$(docker container inspect $(docker container ps -f \
-  name=dtr-registry -q) | grep -m1 -Po '(?<=DTR_VERSION=)\d.\d.\d'); \
+  name=dtr-registry -q) | grep -m1 DTR_VERSION | \
```
bermudezmt: @jose-bigio can I please have you test this command on macOS? Not doubting you @omegamormegil - just getting a second pair of eyes. :)

thaJeztah: Could we use a `--format=....` to get the value? (I assume this version is kept in an env-var on the container/image?)

bermudezmt: @thaJeztah I was able to get the SHA with the following extraneous command:
`DTR_VERSION=$(docker inspect -f '{{.Image}}' $(docker ps -q -f name=dtr-garant) | cut -f 3 -d '-') && echo $DTR_VERSION`

The version definitely exists in the JSON response to:
```
docker ps --format='{{json .}}' | grep "dtr" | jq . | less
```
Snippet:
```
"Command": "\"/bin/api\"",
  "CreatedAt": "2019-04-02 22:31:44 +0000 UTC",
  "ID": "18d086929ada",
  "Image": "docker/dtr-api:2.6.4"
```

It's my first time playing around with Go template formatting, so need more education. :) cc: @caervs :wave:

thaJeztah: Ah, hm.. looks like that would still be adding some complication (and requires `jq`, which (although really handy) is not a standard-tool.

Looks like the image does not have a label or env-var set to have the version. Perhaps that's something to consider (but obviously out of scope for this documentation change). e.g.;

```bash
docker build -t lets-pretend-i-am-dtr -<<EOF
FROM busybox
# not needed, but could be convenient in the container itself
ENV DTR_VERSION=2.6.4
LABEL com.docker.product=dtr
LABEL com.docker.product-version=2.6.4
EOF
```

(I'm using reverse-dns notation for the labels; not required, but it prevents them from conflicting with labels that users set, and the `com.docker.` is documented as "reserved for internal use by docker")

With a label in place, we can filter on that label (instead of on the name of the container), so the container could have any name you want:

```bash
docker container create --name any-name-you-like lets-pretend-i-am-dtr:latest
```

And then we can filter containers, and extract the version from the label value;

```bash
docker ps -a \
  --filter label=com.docker.product=dtr \
  --format '{{ .Label "com.docker.product-version" }}'

# prints: 2.6.4
```

bermudezmt: Going to come back to this later, @thaJeztah , but current thoughts are:
1) Update this command to include extracting the DTR URL
2) Extract the DTR version via this curl request from @cquon :
  `curl -k -X GET --url https://<dtr_url>/basic_info -H "Accept:application/json" -u <admin_username>:<admin_password>`

@cquon @RobbKistler Is this something you can help us tighten up for "advanced usage" of the backup procedure for DTR?

jose-bigio: @bermudezmt  do you still want me to double check this command?

Is it: 
```
DTR_VERSION=$(docker container inspect $(docker container ps -f \
  name=dtr-registry -q) | grep -m1 DTR_VERSION | \
  sed 's/.*DTR_VERSION=\([0-9].[0-9].[0-9]\)",/\1/'); \
REPLICA_ID=$(docker ps --filter name=dtr-rethinkdb \
  --format "{{ .Names }}" | head -1 | sed 's|.*/||' | sed 's/dtr-rethinkdb-//'); \
read -p 'ucp-url (The UCP URL including domain and port): ' UCP_URL; \
read -p 'ucp-username (The UCP administrator username): ' UCP_ADMIN; \
read -sp 'ucp password: ' UCP_PASSWORD; \
docker run --log-driver none -i --rm \
  --env UCP_PASSWORD=$UCP_PASSWORD \
  docker/dtr:$DTR_VERSION backup \
  --ucp-username $UCP_ADMIN \
  --ucp-url $UCP_URL \
  --ucp-ca "$(curl https://${UCP_URL}/ca)" \
  --existing-replica-id $REPLICA_ID > \
  dtr-metadata-${DTR_VERSION}-backup-$(date +%Y%m%d-%H_%M_%S).tar
```

That's a long command.

bermudezmt: @jose-bigio the command works for Linux but isn't compatible with macOS. @thaJeztah and I are looking for a cleaner way to extract the DTR version.

jose-bigio: Honestly we should just backport https://github.com/docker/dhe-deploy/pull/10144

reference/dtr/2.6/cli/backup.md
```diff
@@ -27,7 +27,8 @@ docker run -i --rm --log-driver none docker/dtr:{{ page.dtr_version }} \
 {% raw %}
 ```bash
 DTR_VERSION=$(docker container inspect $(docker container ps -f \
```
jose-bigio: In the case of using this with OSX I think it may be helpful to specify that your client bundle should be sourced first. 

As a side note I tried to run this on my mac with a client bundle, but it didn't work.

Here is the output for reference: 
```
DTR_VERSION=$(docker container inspect $(docker container ps -f \
  name=dtr-registry -q) | grep -m1 DTR_VERSION | \
  sed 's/.*DTR_VERSION=\([0-9].[0-9].[0-9]\)",/\1/'); \
REPLICA_ID=$(docker ps --filter name=dtr-rethinkdb \
  --format "{{ .Names }}" | head -1 | sed 's|.*/||' | sed 's/dtr-rethinkdb-//'); \
read -p 'ucp-url (The UCP URL including domain and port): ' UCP_URL; \
read -p 'ucp-username (The UCP administrator username): ' UCP_ADMIN; \
read -sp 'ucp password: ' UCP_PASSWORD; \
docker run --log-driver none -i --rm \
  --env UCP_PASSWORD=$UCP_PASSWORD \
  docker/dtr:$DTR_VERSION backup \
  --ucp-username $UCP_ADMIN \
  --ucp-url $UCP_URL \
  --ucp-ca "$(curl https://${UCP_URL}/ca)" \
  --existing-replica-id $REPLICA_ID > \
  dtr-metadata-${DTR_VERSION}-backup-$(date +%Y%m%d-%H_%M_%S).tar
read: -p: no coprocess
read: -p: no coprocess
read: -p: no coprocess
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0curl: (6) Could not resolve host: ca
docker: invalid reference format.
See 'docker run --help'.
```

bermudezmt:
@omegamormegil I have created and merged #8708 for interim fixes. This is where we are with this PR:
 - Example chained command does not work on macOS for 2.7.x
 - There is a cleaner way to set the replica ID, which I've already committed to `master`
 - If we continue to refine the chained commands, we will need a cleaner way to extract the DTR version that the operator (i.e., UCP admin) is working with. The cleanest path we've found so far is to hit an API endpoint `basic_info`, which includes `CurrentVersion` as one of the API response fields.

> I feel like something like a replica id, and the version are something that the operator might know, but maybe I am wrong. Even if these are things that an operator might not know splitting up the commands and making it something like a bash script might make it more readable and more maintainable.

To add to @jose-bigio 's comment, it may help us to get a better understanding of the comfort level of UCP administrators in general when it comes to a) working with the command line, and b) performing data backups in general.

usha-mandya:

PR #8687, opened 2019-04-21
Removed Reference to Interlock Preview Image

ollypom:
### Proposed changes

I found some references to the internal Interlock preview image, at the same time I have updated all references to interlock images with the UCP org and tag following the normal schema.


paigehargrave:
@ollypom Thank you!!!!

PR #8706, opened 2019-04-23
Update secure-swarm-tls.md

paulcosma:
### Proposed changes
Fix broken link


adrian-plata:
This link is currently working as is, does it still fail for you @paulcosma?

paulcosma:
I've checked the "Docker security" link from the bottom of this page
https://github.com/docker/docker.github.io/blob/master/swarm/secure-swarm-tls.md
This is returning 404 for me.

adrian-plata:
Oh I see. Yes, you're correct. Thanks for discovering that miss!

PR #8724, opened 2019-04-25
Make changes to device-mapper.md

sagarvelankar:
#### Added steps to fix issue where docker daemon does not start with below error

unable to configure the Docker daemon with file /etc/docker/daemon.json: the following directives are specified both as a flag and in the configuration file: storage-driver: (from flag: overlay2, from file: devicemapper)

#### Made the required changes for below

Only the running container snapshot is mounted on /var/lib/docker/devicemapper/mnt
No empty directories exist for Images and Stopped Containers
Also, Snapshot Devices exist only for Running Containers Writable Layer
No devices exist for Images and Stopped Containers and they are just in the thin pool.
Verified in CentOS and Ubuntu.

### DIFF
--- a/storage/storagedriver/device-mapper-driver.md
+++ b/storage/storagedriver/device-mapper-driver.md
@@ -80,6 +80,34 @@ For production systems, see
 
     Docker does not start if the `daemon.json` file contains badly-formed JSON.
 
+    Perform the steps following the below error 
+    otherwise Docker fails to start with below error.
+    
+    ```bash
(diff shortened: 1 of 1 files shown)


usha-mandya:
@sagarvelankar Thanks for the PR. The `master` branch has been renamed to `main`. Some of the existing PRs have been automatically closed as a result of this change. Could you create a new PR against our latest docs if the suggested changes are still relevant? Thank you.

PR #8732, opened 2019-04-26
Fixed syntax error

omegamormegil:
Last edit to the REPLICA_ID command introduced a syntax error by adding an extra ')'. Removed it.

PR #8733, opened 2019-04-26
Fixed syntax error

omegamormegil:
Last edit to the REPLICA_ID command added an extra ')' which causes the command to fail. I fixed it.


thaJeztah:
looks like this is a duplicate of https://github.com/docker/docker.github.io/pull/8732 - let me close this one

PR #8741, opened 2019-04-29
Update from microsoft/* to mcr.microsoft.com/*.

olegburov:
### Proposed changes

Switch IIS Nanoserver Images from legacy location at Docker Hub Registry (`microsoft/*`) to a new Microsoft Container Registry (`mcr.microsoft.com/*`).

### DIFF
--- a/engine/swarm/secrets.md
+++ b/engine/swarm/secrets.md
@@ -308,7 +308,7 @@ This example assumes that you have PowerShell installed.
         --name my-iis
         --publish published=8000,target=8000
         --secret src=homepage,target="\inetpub\wwwroot\index.html"
-        microsoft/iis:nanoserver  
+        mcr.microsoft.com/windows/servercore/iis:nanoserver  
     ```
 
     > **Note**: There is technically no reason to use secrets for this


engine/swarm/secrets.md
```diff
@@ -308,7 +308,7 @@ This example assumes that you have PowerShell installed.
         --name my-iis
         --publish published=8000,target=8000
         --secret src=homepage,target="\inetpub\wwwroot\index.html"
-        microsoft/iis:nanoserver  
+        mcr.microsoft.com/windows/servercore/iis:nanoserver
```
thaJeztah: Looks like this is not a multi-arch image, and the nano server variant is only available for Windows Server 2016? Does that mean we need separate examples for each Windows version? 

https://mcrflowprodcentralus.cdn.mscr.io/mcrprod/windows/servercore/iis?P1=1556504889&P2=1&P3=1&P4=AHd3YG3PHOSpFDBjMbQTd7OW6tqY7NNDFIrFh9H91zY%3D&se=2019-04-29T02%3A28%3A09Z&sig=ZBHQxZlVqFX%2FRzGdeJa7NgwLd5iTHyQHSs3kwWkjcL4%3D&sp=r&sr=b&sv=2015-02-21

olegburov: @thaJeztah That's because IIS on Nano Server no longer supported. I beleive 2016 was the last one. If we need IIS for this example, then it's better use `mcr.microsoft.com/windows/servercore/iis:windowsservercore`, otherwise just `mcr.microsoft.com/windows/nanoserver:1809` w/o IIS.

thaJeztah: And nanoserver on WS 2016 is no longer supported, so I guess we need the Windowsservercore variant, but given that they deprecated `:latest` for all their images (I think), and don't use multi-arch, we'd need separate examples for each version of Windows?

what a mess 😔

PR #8759, opened 2019-05-04
[WIP] Add docker registry plugin reference

thaJeztah:
@bermudezmt @cquon ptal

still WIP, as I think we want more information to be included (enterprise-only etc), but mainly testing if this works and would be a good start


bermudezmt:
Closing since this is now a dupe of a PR in a different repo.

PR #8761, opened 2019-05-04
[WIP] Add Docker App plugin reference documentation

thaJeztah:
related: https://github.com/docker/app/pull/533

marking this "WIP" as the plugin is shipping with Docker 19.03, which has not been released yet. We also need to review the information in the YAML, as we would likely want to indicate that this is a plugin.

In addition, looks like formatting of the examples (obtained from the extended description in code) doesn't work well;

Screenshot 2019-05-04 at 14 02 14

We should either use the same approach as in the docker/cli, and move those to markdown files in the upstream repository (which will get included in the generated YAML), or find a way to add Markdown formatting to the code (and check if it gets rendered properly both on the CLI and in the docs)

### DIFF
--- a/_data/docker-app/docker_app.yaml
+++ b/_data/docker-app/docker_app.yaml
@@ -1,0 +1,45 @@
+command: docker app
+short: Docker Application
+long: A tool to build and manage Docker Applications.
+pname: docker
+plink: docker.yaml
+cname:
+- docker app bundle
+- docker app completion
+- docker app init
+- docker app inspect
+- docker app install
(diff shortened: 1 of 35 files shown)


thaJeztah:
@bermudezmt @chris-crone @simonferquel

thaJeztah:
This was done in https://github.com/docker/docker.github.io/commit/9186d90fc7c436b9b0673ef8f54bb7efa9f9c35d (part of https://github.com/docker/docker.github.io/pull/9098)

PR #8806, opened 2019-05-17
Enable the stable repository after adding it

microfortnight:
I tried installing current docker from Fedora repositories.
It seems the manual is missing a step.
The stable repository is added but also has to be activated, otherwise the package cannot be installed.


usha-mandya:
@microfortnight Thanks for the PR. The `master` branch has been renamed to `main`. Some of the existing PRs have been automatically closed as a result of this change. Could you create a new PR against our latest docs if the suggested changes are still relevant? Thank you.

PR #8809, opened 2019-05-17
Removes reference to docker.app in Windows

PedroLamas:
### Proposed changes

I've removed a reference to authorizing a "docker.app" that doesn't actually exist in Windows


PedroLamas:
Obsolete as this has been fixed by another commit.

PR #8810, opened 2019-05-17
Added UCP Kubernetes Secure RBAC Defaults

ollypom:
### Proposed changes

I removed this page https://github.com/docker/docker.github.io/commit/a49b24f892c2ebdd32fc878ab70dd3bde82979f2#diff-ee052237e3d155e4f28346f36847a480 as it was on related to UCP 3.0 and not required in UCP 3.1. However on the bottom of that page was details on our security defaults for Kubernetes, that bit we need to keep as its important for UCP 3.1 and UCP 3.2.

I was unsure the best to put them, so left them on `index.md` however i'm open to alternatives. 

Note we've actually being carrying them as a `Known Issue` in the release notes, which I personally think is incorrect as its not a Known Issue, its a security feature @paigehargrave  :)

https://docs.docker.com/ee/ucp/release-notes/#317.

```
To deploy Pods with containers using Restricted Parameters, the user must be an admin and a service account must explicitly have a ClusterRoleBinding with cluster-admin as the ClusterRole. Restricted Parameters on Containers include:
Host Bind Mounts
Privileged Mode
Extra Capabilities
Host Networking
Host IPC
Host PID
```

### DIFF
--- a/ee/ucp/authorization/index.md
+++ b/ee/ucp/authorization/index.md
@@ -23,7 +23,7 @@ For an example, see [Deploy stateless app with RBAC](deploy-stateless-app.md).
 
 ## Subjects
 
-A subject represents a user, team, organization, or service account. A subject
+A subject represents a user, team, organization, or a service account. A subject
 can be granted a role that defines permitted operations against one or more
 resource sets.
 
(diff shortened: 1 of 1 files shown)


bermudezmt:
@paigehargrave can you PTAL? Thanks!

bermudezmt:
Thank you, @ollypom and @paigehargrave

PR #8815, opened 2019-05-20
all 3 links redirecting to the same page

SychevIgor:


SychevIgor:
ok, after used provided his email, portal redirected to real article. 
Canceled this PR

PR #8827, opened 2019-05-21
Propose 3 as the number of manager nodes

david-yu:
Propose 3 managers as the default number of manager nodes.


paigehargrave:
@bermudezmt - PTAL and merge if everything looks OK - thank you!

PR #8850, opened 2019-05-26
Fixed typo and added port number

Yegor-V:
Lost couple minutes figuring out on which port is app running on vm-s. So I think it would be nice to add port number in description or change picture a little.


adrian-plata:
LGTM thanks for the help!

PR #8854, opened 2019-05-28
del: remove duplicated 'depends_on' section

uburuntu:
### Proposed changes

Delete duplicated section `depends_on`, leaving lexicographic correct one before `deploy`.

First: https://docs.docker.com/compose/compose-file/#depends_on
Second: https://docs.docker.com/compose/compose-file/#depends_on-1 (deleted in this PR)


uburuntu:
Little bump 🙂

PR #8879, opened 2019-06-01
Remove pay thru Docker section

paigehargrave:


bermudezmt:
@gnomestar can you PTAL? Thanks!

PR #8893, opened 2019-06-04
Updated Windows Docker EE JSON index Link

ollypom:
I noticed there were inconsistencies on this page. In line 151 we used the `download.docker.com` JSON index, on 225 we used `dockermsft.blob.core.windows.net` JSON index. This commit makes the page consistent.

### DIFF
--- a/install/windows/docker-ee.md
+++ b/install/windows/docker-ee.md
@@ -148,7 +148,7 @@ manually, via a script, or on air-gapped systems.
     ```
 
     If you need to download a specific Docker EE Engine release, all URLs can be
-    found on this [JSON index](https://download.docker.com/components/engine/windows-server/index.json)
+    found on this [JSON index](https://dockermsft.blob.core.windows.net/dockercontainer/DockerMsftIndex.json)
 
 2.  Copy the zip file to the machine where you want to install Docker. In a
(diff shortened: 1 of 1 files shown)


thaJeztah:
@stefanscherer @ddebroy ptal

install/windows/docker-ee.md
```diff
@@ -222,8 +222,8 @@ To update Docker Engine - Enterprise to the most recent release, specify the
 ```powershell
 Install-Package -Name docker -ProviderName DockerMsftProvider -RequiredVersion {{ site.docker_ee_version }} -Update -Force
 ```
-The required version number must match a versions available on the [JSON
-index](https://dockermsft.blob.core.windows.net/dockercontainer/DockerMsftIndex.json)
+The required version number must match a version available on the [JSON
+index](https://download.docker.com/components/engine/windows-server/index.json)
```
StefanScherer: Thanks, nice catch. 
We should keep the msft link to the JSON file as this is used by the DockerMsftProvider. So, the other link should be updated to be consistent. 

The docker link can also be used to access preview and release candidates which not what we want here in the docs.

ollypom: @StefanScherer No problem, just forced a commit that switched it around, making the msft link consistent in both places. Thanks :)

PR #8907, opened 2019-06-07
Reduce user failure in part 4 port mapping.

riftrid3r:
### Proposed changes
As a new user and because of the default ls output, the text at https://docs.docker.com/get-started/part4/ still maintains port 2376 as default port for setting the swarm leader. A new user, as a default procedure, will specify port 2376 when initializing a swarm manager... Despite the short quip about default management ports, the tutorial instructions can be confusing and frustrating to new users. I suggest the example command be this:

`docker-machine ssh myvm1 "docker swarm init --advertise-addr <myvm1 ip>:2377"`

Thus the note following about daemon port 2376 and setting management to port 2377 will be intuitive as will the correct procedure for setting the swarm leader.

### DIFF
--- a/get-started/part4.md
+++ b/get-started/part4.md
@@ -170,7 +170,7 @@ to become a swarm manager with `docker swarm init` and look for output like
 this:
 
 ```shell
-$ docker-machine ssh myvm1 "docker swarm init --advertise-addr <myvm1 ip>"
+$ docker-machine ssh myvm1 "docker swarm init --advertise-addr <myvm1 ip>:2377"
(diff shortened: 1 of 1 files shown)


DawnWood-Docker:
HI @riftrid3r Thank you for taking the time to update the [Get Started](https://docs.docker.com/get-started/) guide. We just published a new Get Started guide that uses best practices for the latest release of Docker, including a section on using Docker Desktop and another on Kubernetes. 
I'm closing this pr because the content is all new. 
Would love to hear your feedback after looking at the new guide.

PR #8911, opened 2019-06-09
Add missing documentation from https://github.com/docker/machine/pull/3665

wwsean08:
### Proposed changes
When looking for how to use azure with docker-machine I was surprised to not see it documented how you can do it in an automated fashion.  While researching this I found that when the functionality was added in https://github.com/docker/machine/pull/3665 there was documentation.  This recreates that documentation which appears to have been lost.

### DIFF
--- a/machine/drivers/azure.md
+++ b/machine/drivers/azure.md
@@ -15,7 +15,7 @@ You need an Azure Subscription to use this Docker Machine driver.
 [azure]: http://azure.microsoft.com/
 [trial]: https://azure.microsoft.com/free/
 
-## Authentication
+## Authentication (browser-based)
 
 The first time you try to create a machine, Azure driver asks you to
 authenticate:
@@ -26,6 +26,8 @@ authenticate:
     Enter the code [...] to authenticate.
 
 After authenticating, the driver remembers your credentials up to two weeks.
+Once your credentials expire, you will be prompted to log in again.
+This is practical for daily use; but not recommended for headless applications.
 
 > **KNOWN ISSUE:** There is a known issue with Azure Active Directory causing stored
 > credentials to expire within hours rather than 14 days when the user logs in with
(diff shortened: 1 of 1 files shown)


wwsean08:
Not quite sure what the build error is, any help on that would be appreciated

wwsean08:
anything I can do to help get this moving forward thru the merge request process?

machine/drivers/azure.md
```diff
@@ -35,6 +37,29 @@ After authenticating, the driver remembers your credentials up to two weeks.
 
 [aad-docs]: https://azure.microsoft.com/documentation/articles/virtual-machines-windows-create-aad-work-id/
 
+## Authentication (service principal account)
+1. Create an Azure Service Principal account ([described here][sp-1] or [here][sp-2]) to get a
```
StefanScherer: The link to the Microsoft docs could be added directly in the markdown link. 
The link to the Packer docs is not needed.

machine/drivers/azure.md
```diff
@@ -35,6 +37,29 @@ After authenticating, the driver remembers your credentials up to two weeks.
...
+   `client_id` and `client_secret`
+2. Use `azure role assignment create` command to give your service account
```
StefanScherer: Maybe we should use `az` instead of the old `azure` command here.

machine/drivers/azure.md
```diff
@@ -35,6 +37,29 @@ After authenticating, the driver remembers your credentials up to two weeks.
...
+```
+$ docker-machine create --driver azure --azure-subscription-id <subs-id> \
+    --azure-client-id <client-id> --azure-client-secret <client-secret> \
+    <machine-name>
+```
+
+Azure Service Principal Account credentials are recommended for headless
+applications (such as CI/CD systems).
+
+The Service Principal Account password created through the `azure ad app create`
```
StefanScherer: Same here, use `az` instead of `azure`.

cadavre:
For 4 years there is Service Principal Account auth flow available and it's still not documented? Are you serious?! xD Via https://github.com/docker/machine/pull/3665 .

craig-osterhout:
Thank you for contributing. Closing as Docker Machine is now deprecated.

PR #8919, opened 2019-06-11
Adding that DRAIN should not be used  on managers.

gonzalo-contador:
In addition this feature may disable DTR or other node capabilities when used on Manager nodes. Using Drain on managers is highly discouraged.


usha-mandya:

PR #8929, opened 2019-06-12
Update docker-ee.md

### LINKED ISSUE #8927: Changes to Microsoft Windows Server Installation instructions due to PSCX installation
### Problem description

I came across and issue where on of the Windows Server 2016 machines had PSCX installed and there looks to be a bug in that is you have PSCX installed it will use that version of the Extract-Archive.  If you run it you will get the error:

Install-Package : A parameter cannot be found that matches parameter name 'DestinationPath'. At line:1 char:1 + Install-Package -Name docker -ProviderName DockerMsftProvider -Verbos ... + ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ + CategoryInfo : InvalidOperation: (Microsoft.Power....InstallPackage:InstallPackage) [Install-Package], Exception + FullyQualifiedErrorId : FailedToDownload,Install-Package,Microsoft.PowerShell.PackageManagement.Cmdlets.InstallPackage

The solution that I used was to use the instructions "Use a script to install Docker EE"

### Problem location

On this [page](https://docs.docker.com/install/windows/docker-ee/) under "Use a script to install Docker EE" I had to change the command on the "# Extract the archive" to use 

Expand-Archive docker-18.09.6.zip -DestinationPath $Env:ProgramFiles -Force

TO:

Microsoft.PowerShell.Archive\Expand-Archive docker-18.09.6.zip -DestinationPath $Env:ProgramFiles -Force

All Versions

### Suggestions for a fix

Expand-Archive docker-18.09.6.zip -DestinationPath $Env:ProgramFiles -Force

TO:

Microsoft.PowerShell.Archive\Expand-Archive docker-18.09.6.zip -DestinationPath $Env:ProgramFiles -Force

stonesbg:
Make change to allow for the ability to install even if you have the Pscx Modules installed

### Proposed changes

Change the command that is used for Expand-Archive.  There is an issue if you have the Pscx (Powershell Community Extensions) installed on a server then it will use that modules version of the Expand-Archive.  The change is just to isolate the Expand-Archive to use the Microsoft.PowerShell.Archive version of Expand-Archive

### Related issues (optional)

Fixes #8927

### DIFF
--- a/install/windows/docker-ee.md
+++ b/install/windows/docker-ee.md
@@ -159,7 +159,7 @@ manually, via a script, or on air-gapped systems.
     Stop-Service docker
     
     # Extract the archive.
-    Expand-Archive {{ filename }} -DestinationPath $Env:ProgramFiles -Force
+    Microsoft.PowerShell.Archive\Expand-Archive {{ filename }} -DestinationPath $Env:ProgramFiles -Force
 
     # Clean up the zip file.
     Remove-Item -Force {{ filename }}


usha-mandya:
@stevenfollis Could you PTAL? Thanks.

stevenfollis:
Thanks Brian for your contribution!

After reviewing, I would strongly prefer to not move away from `Expand-Archive` as it is the default, 1st-party command from MSFT that works out of the box with Windows Server 2016. 

I reproduced the issue on a fresh box just now and can confirm the conflict, but it looks like this collision [has been identified some time ago](https://github.com/Pscx/Pscx/issues/8#issue-162625325) within the Pscx project. To even install Pscx required forcing a collision via the `-AllowClobber` flag.

It would be preferable either for the Pscx project to remove the collision, or for users to use the `Import-Module -Prefix` workaround [mentioned in the thread](https://github.com/Pscx/Pscx/issues/8#issuecomment-229091898) to avoid collisions. 

I did confirm that the `-Prefix` workaround works, and if the issue comes up regularly in the future perhaps we pipe that into a KB article. However at this time I'd recommend that we keep the instructions as-is to align most closely with the default Windows behavior.

stonesbg:
@stefanfoulis Thanks for the follow up.  I am fine with closing out this issue.  Thanks for taking a look

usha-mandya:
Thank you both. Closing this PR.

PR #8935, opened 2019-06-13
Update part2.md

bjellesma:
It might be useful for beginners to put the positional argument before the optional arguments on `docker build` just because I ended up typing it myself and missing the period at first. So the command would be:

```shell
docker build . --tag=friendlyhello 
```

instead of 

```shell
docker build --tag=friendlyhello .
```

### Proposed changes

I am proposing

```shell
docker build . --tag=friendlyhello 
```

instead of 

```shell
docker build --tag=friendlyhello .
```

for part2.md

### DIFF
--- a/get-started/part2.md
+++ b/get-started/part2.md
@@ -159,7 +159,7 @@ Now run the build command. This creates a Docker image, which we're going to
 name using the `--tag` option. Use `-t` if you want to use the shorter option.
 
 ```shell
-docker build --tag=friendlyhello .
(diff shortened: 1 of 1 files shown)


DawnWood-Docker:
Hey @bjellesma Thanks for the pr! We love it when you take time to make the docs better. We just merged a pr that adds a remark to this page. It says:
_The trailing .(dot) in the docker build command is to represent that the docker file is in the current directory._

PR #8939, opened 2019-06-13
docker-compose: change overview from /compose/overview to /compose/

thaJeztah:
the /compose/ (index page) didn't contain any useful information, and wasn't included in the navigation, so we might as well use "overview" as the landing page, which makes the URL also a bit more logical.


thaJeztah:
@chris-crone

chris-crone:
SGTM, thanks @thaJeztah

thaJeztah:
@usha-mandya ptal

usha-mandya:
That makes sense. Thanks for looking into this @thaJeztah.

PR #8941, opened 2019-06-13
Standardize “digitalocean” keyword

sonicdoe:
Most DigitalOcean-related pages use the [`digitalocean`](https://docs.docker.com/glossary/?term=digitalocean) keyword, however, [one page used `digital ocean`](https://docs.docker.com/machine/examples/ocean/) and [another `Digital Ocean`](https://docs.docker.com/machine/drivers/digital-ocean/). This pull request standardizes all of them to `digitalocean`.

See also #8938.


adrian-plata:
Thanks for helping with this!

PR #8945, opened 2019-06-13
Added logging driver section

stevenfollis:
### Proposed changes

Added a short blurb pointing readers to the Docker Logging Driver documentation. If a user does not setup log rotation at installation time they run the chance of maxing out the local disk with log files, bricking a cluster. Adding a pointer in the install docs, similar to the storage driver information, should help flag that need earlier in the installation process.

(Had a customer have this happen to them)


ahromis:
LGTM

stevenfollis:
@bermudezmt That's high praise from your excellent self! Thanks! :)

PR #8957, opened 2019-06-18
Update aws.md with information on SG

### LINKED ISSUE #8956: Documentation on AWS Security Groups is not complete
### Problem description

Documentation states:

> If you specify a security group yourself using the --amazonec2-security-group flag, the above ports are checked and opened and the security group is modified.

If your security group has opened a range of ports, on which it has the required ports within, docker machine will not detect this and create new rules for this for any interface (`0.0.0.0` by default)

This can be dangerous, as it can inadvertently open port 22 to the world.

### Problem location

- I saw a problem on the following URL: https://docs.docker.com/machine/drivers/aws
- Other: This is not specified in the documentation.

Current version

### Suggestions for a fix

Will create a PR to update documentation.

ahoulgrave:
### Proposed changes

Improved documentation regarding AWS security groups with information about how ports are opened.

### Related issues (optional)
Closes #8956


ahoulgrave:
What's the status on this?

PR #8972, opened 2019-06-21
Clarify which admin is required to use privileged options

quadespresso:
Signed-off-by: Jim Carroll <jim.carroll@docker.com>

### Proposed changes

Corrected to be consistent with the [3.1.7 release notes](https://docs.docker.com/ee/ucp/release-notes/#317), per internal FIELD-1770.


paigehargrave:
LGTM

PR #8974, opened 2019-06-21
Corrected command to list running containers

wasabi991011:
### Proposed changes
I corrected the command so that now it will list the container IDs, which is what is needed to stop a container. Previously, the command would only list image IDs, which is useless in this context and confusing for new users.


wasabi991011:
Misread the document, please ignore

PR #8975, opened 2019-06-23
Added build command explanation

mohamedanees6:
### Proposed changes

Added clear definition following the stackoverflow question which has more than 1 lakh views, which means many people get confused in running the command correctly.

https://stackoverflow.com/questions/28996907/docker-build-requires-1-argument-see-docker-build-help


DawnWood-Docker:
Thank you, @mohamedanees6! Approved.

PR #8979, opened 2019-06-24
Updated the UCP Configuration Timeout Settings

ollypom:
### Proposed changes

I noticed this page has not been updated since the default timeouts were changed last year. It is now the same as https://docs.docker.com/ee/ucp/admin/configure/set-session-timeout/


ollypom:
@DawnWood-Docker PTAL : )

PR #8987, opened 2019-06-25
Update Dockerfile for 18.04

pabloab:
Update Dockerfile for 18.04
Installed necessary dependencies: gnupg2 ca-certificates
Updated key according to official documentation
Used redirection to avoid warning about parsing stdout.


adrian-plata:
@thaJeztah what do you think? This page is no longer reference directly in the docs and only exists in the repo now. The reference was deleted in 2017 (https://github.com/docker/docker.github.io/commit/2ab413d8c8f7b59f56a805906a050a7b251789e5#diff-65dce3586b94a23fe9ffdd566a4f2eff)

thaJeztah:
Hm, yes, if we no longer reference this Dockerfile, it might be better to remove it instead

adrian-plata:
Closing this issue as we'll be removing this file.

PR #9009, opened 2019-06-28
Broken link for ucp-config

kavikat:
Broken link, update link or remove reference to link.


ollypom:
@kavikat Thanks for spotting the broken link. However I think we should still keep a link in there but with the correct URL.

Do you mind swapping `../ucp-configuration-file/` to `./ucp-configuration-file/`? Thanks :)

usha-mandya:
This is fixed in https://github.com/docker/docker.github.io/pull/9056.

PR #9018, opened 2019-07-01
The notary config file section had a comma missing

nirajrules:
Docker SA from South East


ollypom:
LGTM. @adrian-plata PTAL

PR #9027, opened 2019-07-03
recommend overlay2

rscohn2:
### Proposed changes

changed overlay to overlay2 in the description about controlling storage. I blindly followed the recommendation to use overlay and tar stopped working inside the container. 

### Related issues (optional)

I saw an issue similar to this: https://github.com/coreos/bugs/issues/1095


usha-mandya:
Thanks for the update.

PR #9041, opened 2019-07-09
typo last command interlock-proxy name

ivansharamok:
### Proposed changes

last command has service name `interlock=proxy` but should be `interlock-proxy`


usha-mandya:
@ivansharamok Thank you for fixing this.

PR #9045, opened 2019-07-10
Add DNS names used by DTR

ivansharamok:
### Proposed changes

added list of DNS names that DTR calls to page `ee/dtr/admin/configure/use-a-web-proxy`. These DNS names should be whitelisted if proxy only works with whitelisted domains.

PR #9050, opened 2019-07-11
add DNS names used by DTR

ivansharamok:
### Proposed changes

added list of DNS names that DTR calls to page `ee/dtr/admin/configure/use-a-web-proxy`. These DNS names should be whitelisted if proxy only works with whitelisted domains.


ivansharamok:
This PR is replacement for closed one #9045

Issue #8851, opened 2019-05-27
python 2.7 no longer maintained and requirements.txt error


IamKraZ:
[get-started/part2.md](https://docs.docker.com/get-started/part2/)

When i go to create the new image in the tutorial Python throws the error:

**_DEPRECATION: Python 2.7 will reach the end of its life on January 1st, 2020. Please upgrade your Python as Python 2.7 won't be maintained after that date. A future version of pip will drop support for Python 2.7.
ERROR: Could not open requirements file: [Errno 2] No such file or directory: 'requirements.txt'
The command '/bin/sh -c pip install --trusted-host pypi.python.org -r requirements.txt' returned a non-zero code: 1_**

and my test image is not created. Is there a work around? or am I doing it wrong, lol?

**Edit**
Also what is the  "none" file?

**_REPOSITORY          TAG                 IMAGE ID            CREATED             SIZE
"none"              "none"              d4e018f2c61b        12 seconds ago      120MB
python              2.7-slim            eb40dcfcbc42        2 weeks ago         120MB_**

I  attempted to create the Dockerfile 
**docker build -t friendlyhello .**
 
and got this:

**_Sending build context to Docker daemon   5.12kB
Step 1/7 : FROM python:2.7-slim
2.7-slim: Pulling from library/python
743f2d6c1f65: Already exists
9bfbedfce8de: Already exists
7f4da2474cef: Already exists
ffc893575679: Already exists
Digest: sha256:686bc67cbebf4c4fb0d96d55650d8704d00ccb6b9c6bdd0bee5fad48b827a6cb
Status: Downloaded newer image for python:2.7-slim
 ---> eb40dcfcbc42
Step 2/7 : WORKDIR /app
 ---> Running in 21b1d8621926
Removing intermediate container 21b1d8621926
 ---> c23711eb7762
Step 3/7 : COPY . /app
 ---> d4e018f2c61b
Step 4/7 : RUN pip install --trusted-host pypi.python.org -r requirements.txt
 ---> Running in 8a077703e81b
DEPRECATION: Python 2.7 will reach the end of its life on January 1st, 2020. Please upgrade your Python as Python 2.7 won't be maintained after that date. A future version of pip will drop support for Python 2.7.
ERROR: Could not open requirements file: [Errno 2] No such file or directory: 'requirements.txt'
The command '/bin/sh -c pip install --trusted-host pypi.python.org -r requirements.txt' returned a non-zero code: 1_**

[...]

PR #9058, opened 2019-07-13
Changes in part2.md for pip to work with requirements.txt

jromerof:
As you're copying all the files to /app requirements.txt has to be referenced from the /app folder to work, otherwise it returns ERROR: Could not open requirements file: [Errno 2] No such file or directory: 'app/requirements.txt' (there is a issue on that https://github.com/docker/docker.github.io/issues/8851)

Also first PR ever so sorry in advance if I make a mistake.

### DIFF
--- a/get-started/part2.md
+++ b/get-started/part2.md
@@ -70,7 +70,7 @@ WORKDIR /app
 COPY . /app
 
 # Install any needed packages specified in requirements.txt
-RUN pip install --trusted-host pypi.python.org -r requirements.txt
+RUN pip install --trusted-host pypi.python.org -r /app/requirements.txt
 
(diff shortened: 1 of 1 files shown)


ollypom:
Hey @jromerof, I'm afraid I'm not able to reproduce your error. 

In the Dockerfile we are setting the working directory with `WORKDIR /app`, so all RUN commands from this point are already ran from inside `/app`. 

To verify this, if I run my newly built container, you can see i'm in the `/app` directory already, with my files already in there.

```
$ docker run --rm -it friendlyhello:latest sh
# pwd
/app
# ls
Dockerfile  app.py  requirements.txt
``

PR #9063, opened 2019-07-15
Found Troubleshoot Redirect in UCP v1 Archive

ollypom:
### Proposed changes
Found a few old redirects in the UCP v1 Archive. I found them through

https://www.docker.com/ddc-2 > https://success.docker.com/article/ucp-says-i-have-unhealthy-controllers-managers-what-should-i-do -> https://docs.docker.com/datacenter/ucp/1.1/monitor/troubleshoot-ucp/.

So this just brings the redirects up to latest images.


DawnWood-Docker:
@ollypom LGTM

PR #9064, opened 2019-07-15
better documentation on network organization

ingshtrom:
### Proposed changes
Currently in UCP if your administrators use Swarm Collections heavily and do not grant everyone else full administrative privileges, then it can become confusing and cumbersome as users have networks and other stack resources existing in separate collections. For example, let's say we have a new user named `billy` and he has access to several collections:
* `/Swarm/Shared/Private/billy` => View Only access. This is the default set by UCP on the first login of `billy` and we have the default Grant to be `View Only`
* `/Swarm/service1` => Full Control access. This is where Billy and his team deploy one of their services
* `/Swarm/service2` => Full Control access. This is where Billy is creating a new service for his team

The user `billy` has his Default Collection set to the default value of `/Swarm/Shared/Private/billy` and doesn't change it because he manages multiple collections, so a default collection does not mean much to him. 

After he is ready to deploy `service2`, he runs `docker stack deploy -c service1.yaml service2` with the following compose file:
```
version: '3.6'

services:
  web:
    image: nginx:latest
    ports:
      - mode: host
        protocol: tcp
        target: 80
    deploy:
      replicas: 1
      update_config:
        parallelism: 1
        failure_action: rollback
        delay: 30s
      labels:
        com.docker.ucp.access.label: "/service2"
      resources:
        limits:
            memory: 128M
        reservations:
            memory: 128M
```

He then receives an error like this:
```
Creating network service2_default
failed to create network service2_default: Error response from daemon: access denied:
no access to Network Create, on collection swarm/shared/private/billy
```

He is very confused because this is what the documentation says to do [here](https://docs.docker.com/ee/ucp/swarm/deploy-to-collection/). 

...

This scenario (and those like it) is what this documentation PR is trying to solve. I'm not 100% sure the wording is correct, but I want to make it clear that to correctly put networks, and really all resources in a stack, into the expected collection, then the user MUST set the label accordingly. The confusing part currently is that the network is automatically created by the `docker stack deploy...` command, so it can easily go unnoticed by the user until their permissions are locked down further.

### DIFF
--- a/ee/ucp/swarm/deploy-to-collection.md
+++ b/ee/ucp/swarm/deploy-to-collection.md
@@ -111,7 +111,7 @@ To confirm that the service deployed to the `/Shared/wordpress` collection:
 It is important to note that by default Docker Stacks will create a default `overlay`
 network for your stack. It will be
 attached to each container that is deployed. This works if you have full control over
-your Default Collection or are an adminitrator. If your administrators have locked
+your Default Collection or are an administrator. If your administrators have locked
 down UCP to only allow you access to specific collections and you manage multiple
 collections, then it can get very difficult to manage the networks as well and you
 might run into permissions errors. To fix this, you must define a custom network
--- a/ee/ucp/swarm/deploy-to-collection.md
+++ b/ee/ucp/swarm/deploy-to-collection.md
@@ -109,7 +109,7 @@ To confirm that the service deployed to the `/Shared/wordpress` collection:
 ### Notes
 
 It is important to note that by default Docker Stacks will create a default `overlay`
-network for your stack if you expose any ports in any of the services. It will be
+network for your stack. It will be
 attached to each container that is deployed. This works if you have full control over
 your Default Collection or are an adminitrator. If your administrators have locked
 down UCP to only allow you access to specific collections and you manage multiple
--- a/ee/ucp/swarm/deploy-to-collection.md
+++ b/ee/ucp/swarm/deploy-to-collection.md
@@ -109,7 +109,7 @@ To confirm that the service deployed to the `/Shared/wordpress` collection:
 ### Notes
 
 It is important to note that by default Docker Stacks will create a default `overlay`
-network for your stack if you expose any ports in any of the service. It will be
+network for your stack if you expose any ports in any of the services. It will be
 attached to each container that is deployed. This works if you have full control over
 your Default Collection or are an adminitrator. If your administrators have locked
 down UCP to only allow you access to specific collections and you manage multiple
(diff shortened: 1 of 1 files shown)


ee/ucp/swarm/deploy-to-collection.md
```diff
@@ -57,20 +57,30 @@ services:
...
       MYSQL_ROOT_PASSWORD: example
-    deploy:  
+    deploy:
       labels:
         com.docker.ucp.access.label: /Shared/wordpress
+
+networks:
+  wp:
+    driver: overlay
+    lables:
```
thaJeztah: s/lables/labels/

ee/ucp/swarm/deploy-to-collection.md
```diff
@@ -57,20 +57,30 @@ services:
...
+    networks:
+      - wp
     environment:
       MYSQL_ROOT_PASSWORD: example
-    deploy:  
+    deploy:
       labels:
         com.docker.ucp.access.label: /Shared/wordpress
+
+networks:
```
thaJeztah: I wonder though if the original example intended the network to be created from this stack, or if this was meant to be an external network (created up-front); if so, this page probably needs a description on how to create the network first, and the network in this stack should have `external: true`

ee/ucp/swarm/deploy-to-collection.md
```diff
@@ -57,20 +57,30 @@ services:
...
+  wp:
```
thaJeztah: Alternatively, you can name this one `default:`, in which case it configures the default network for the stack (in that case it's not needed to reference the `wp` network in all services

ingshtrom: That's good to know! TIL.  I think I'll leave it for now as I think being more explicit in the documentation is better.

ee/ucp/swarm/deploy-to-collection.md
ingshtrom: Agreed, either way, it's ambiguous and not helpful to the reader. I'll let someone from docs comment on which direction they would prefer this to go.

ee/ucp/swarm/deploy-to-collection.md
ingshtrom: oops. Copied and then modified my test incorrectly. Fixing

usha-mandya:
@DawnWood-Docker  - Could you PTAL the updates when you get a chance?

ingshtrom:
So will someone else merge this for me as I don't have the ability to do it myself? I know that the CI failed, but I'm not sure exactly what that CI is for or how to fix it. I looked at the failure and it looks like a failure to source a UCP bundle maybe? 

Any help appreciated!

usha-mandya:
@ingshtrom - I can merge the PR.

The CI failure isn't related to this PR. The infra team is aware of this issue and are planning to fix this. I'll follow up on this.

PR #9067, opened 2019-07-16
Update part2.md

oschihin:
so other's dont stumble and fall ...


usha-mandya:
@oschihin - thanks for your contribution.

PR #9079, opened 2019-07-17
Fix docker logging's options indentation that would cause ScannerError when copied

hong823:
### Proposed changes

Fix docker logging's options incorrect indentation that would cause yaml.scanner.ScannerError when copied directly to `docker-compose.yml` and it might confuse developers.

Link: https://docs.docker.com/config/containers/logging/awslogs/


usha-mandya:
Fixes #9079

PR #9081, opened 2019-07-18
updating relnotes for july patch

DawnWood-Docker:
### Proposed changes

updated engine and dtr release notes for July patch


DawnWood-Docker:
need these to go live today and no one else is around.

PR #9083, opened 2019-07-18
Release Notes formatting

ollypom:
### Proposed changes

I noticed the Runtime in 18.03.1-ee-10 was not set as H3. 

Also noticed a UCP Rel Note had found itself in the Engine Rel notes, it is correctly documented in https://github.com/docker/docker.github.io/blob/master/ee/ucp/release-notes.md#318, so I removed it from Engine Rel Notes.


usha-mandya:
LGTM

PR #9085, opened 2019-07-18
put emphasis on current online gc method

programmerq:
changed the order of the information to highlight the current behavior, while preserving the past behavior as an afterthought. Calling out the past behavior in the current version of the docs is relevant because many may not know that the behavior changed.

### Proposed changes

clarify information about a behavior change for both new and existing users of DTR

### DIFF
--- a/ee/dtr/admin/configure/garbage-collection.md
+++ b/ee/dtr/admin/configure/garbage-collection.md
@@ -16,11 +16,11 @@ the scheduled time, DTR:
 1. Identifies and marks unused image layers.
 2. Deletes the marked image layers.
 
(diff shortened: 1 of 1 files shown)


adrian-plata:
@programmerq I have a suggested text update below. If it looks good to you, please update the PR and I'll merge these changes.

As of v2.6, DTR uses online garbage collection. This allows DTR to run garbage
collection without setting DTR to read-only/offline mode. In previous versions,
garbage collection would set DTR to read-only/offline mode so DTR would reject
pushes. Online garbage collection was an experimental feature in v2.5.

PR #9086, opened 2019-07-18
3.0 release configs

allysmith424:
Configs updated for the archiving of 18.09 and release of 3.0. I'd especially appreciate eyes on `_config.yml` to ensure we have all version variables correct.


allysmith424:
> This looks great! Thanks Ally.
> 
> One Nit in the Dockerfile looks like Distribution version is now 2.7

Nice spot @ollypom -- thanks

_config.yml
```diff
@@ -106,7 +120,14 @@ defaults:
...
       dtr_repo: "dtr"
-      dtr_version: "2.6.8"
+      dtr_version: "2.7.0"
+  - scope:
+      path: "datacenter/dtr/2.6"
+    values:
+      hide_from_sitemap: true
+      dtr_org: "docker"
+      dtr_repo: "dtr"
+      dtr_version: "2.6.6"
```
ollypom: Should be 2.6.8

_config.yml
```diff
@@ -156,8 +177,15 @@ defaults:
...
-      dtr_version: "2.6.8"
+      ucp_version: "3.2.0"
+      dtr_version: "2.7.0"
+  - scope:
+      path: "datacenter/ucp/3.1"
+    values:
+      hide_from_sitemap: true
+      ucp_org: "docker"
+      ucp_repo: "ucp"
+      ucp_version: "3.1.7"
```
ollypom: Should be 3.1.9

PR #9095, opened 2019-07-21
Corrected typo

bukkasamudram:
Before:
Use docker inspect devtest to verify that the volume was created and mounted correctly. Look for the Mounts section:
After:
Use docker inspect myvol2 to verify that the volume was created and mounted correctly. Look for the Mounts section:


bukkasamudram:
HI, Corrected the type. Please merge to master.

PR #9107, opened 2019-07-23
Fix confusing 'log-opt' key in json part of docs

w4tsn:
### Proposed changes

I changed the `log-opt` key to `log-opts` in the documentation of logging drivers in the part where `json` configs are used, because this causes confusion which of the keys is the correct one.


usha-mandya:
@w4tsn Thank you for your contribution.

PR #9109, opened 2019-07-23
Sync published with master 

usha-mandya:
Sync published with master for DTR and Compose updates


usha-mandya:
@ollypom could you PTAL?

PR #9112, opened 2019-07-23
fix common use cases broken link

jk2K:


adrian-plata:
LGTM, thanks for contributing!

PR #9145, opened 2019-07-26
Sync published with master

usha-mandya:


usha-mandya:
That's it. Thanks @DawnWood-Docker.

DawnWood-Docker:
@Usha LGTM Olly's commit is here as well

DawnWood-Docker:
nice catch

PR #9192, opened 2019-08-01
suse.md: info about sles and docker-ee versions

kolyshkin:
1. `12.3` is not the only SLES version supported.

2. Latest Docker EE version is 19.03

3. Provide a complete example.


kolyshkin:
@ollypom @usha-mandya PTAL

PR #9201, opened 2019-08-04
Setup Ubuntu as default distro

helge000:
### Proposed changes
If we have more then one WSL distro installed, Ubuntu has do be the default distro in order to be reconnoitred by WSL2 preview prerequisites check
Capture


helge000:
Jenkins error seems unrelated?

usha-mandya:
@helge000 - Thanks for your feedback. This issue was fixed in #9248.

PR #9203, opened 2019-08-05
Fix a link in release-notes.md

martynasb:


adrian-plata:
Nice catch, thanks for contributing!

PR #9211, opened 2019-08-07
fix: replace non printable characters

alexisvisco:
The command example sould not contains non printable character. 

If we copied the command, the docker client gave us this:

```
unknown shorthand flag: 'â' in -–log-driver
See 'docker run --help'.
```


usha-mandya:
@alexisvisco- Thank you for your feedback. We were not able to reproduce the error you've described. Closing this PR due to this reason. If I have misunderstood, please reopen it with more information.

PR #9226, opened 2019-08-08
Add notes about DTR install 

traci-morrison:
Signed-off-by: Traci Morrison <traci.morrison@docker.com>

### Proposed changes

Adding information to let users know to upgrade UCP before installing DTR.

### Related issues (optional)
https://docker.atlassian.net/browse/ENGDOCS-60

### DIFF
--- a/ee/dtr/admin/install/index.md
+++ b/ee/dtr/admin/install/index.md
@@ -15,11 +15,12 @@ infrastructure meets the [system requirements](./system-requirements) that DTR n
 
 ## Step 2. Install UCP
 
->**Note:** Before installing DTR:
-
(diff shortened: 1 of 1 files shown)


cemckenzie:
Was this meaning "before installing DTR" ?
_**Note:** Before installing UCP:_

On this line:
_* UCP should be upgraded to the most recent version._
is OK if DTR is not yet installed, but when doing an upgrade we want the customer to upgrade UCP before DTR for each major version.

For instance I have a customer that is jumping up 4 major versions, jumping UCP 4 versions and then starting to upgrade DTR would have gotten them in trouble.  I think if the official advice is to step one major version at a time, first UCP, then DTR, repeat, we will stay clear of issues.

traci-morrison:
@cemckenzie I made those changes. PTAL.

PR #9234, opened 2019-08-10
fix COPY failed

linchiwei123:
since README.md listed in https://github.com/docker-library/hello-world/blob/master/.dockerignore , you will get COPY failed without this change.


usha-mandya:
@linchiwei123 - Thank you for your contribution.

PR #9237, opened 2019-08-11
DOCS - correct units of measure for memory restrictions

discdiver:
Docs say *m* means megabytes, but it actually changes mebibytes. Same for other units of measure.


usha-mandya:
@thaJeztah - Could you PTAL and confirm if this is correct?

PR #9252, opened 2019-08-14
Updated Docker Engine for clarity.

Maannniii:
Since Docker Engine is renamed as Docker. Specifying **Docker Engine**  creates confusion for persons who are not aware of Docker and Docker Engine are same. Either both the terms should be mentioned or the updated one should be mentioned. Hence **updated Docker Engine Community as Docker Community**.


usha-mandya:
@Maannniii - Thank you for your feedback. Unfortunately, this change is against our current branding. We'll have to close this PR and discuss this internally.

PR #9257, opened 2019-08-14
Add install Helm on Docker EE section

traci-morrison:
Signed-off-by: Traci Morrison <traci.morrison@docker.com>

### Proposed changes

Add install Helm on Docker EE section
### Related issues (optional)
https://docker.atlassian.net/browse/ENGDOCS-33


traci-morrison:
@ollypom PTAL.

DawnWood-Docker:
Hey @traci-morrison let's also loop in @david-yu because he opened the issue. 
A couple of minor things that I noticed:
1. Should we call it Docker Enterprise instead of Docker EE? EE is overloaded. The way I understand it we talk about Docker Enterprise unless we're specifically calling out Docker Enterprise Engine (which is also a part of Docker Enterprise). In the past Docker EE stood for Docker Enterprise Edition but we don't use the "Edition" nomenclature anymore. I noticed some other instances of EE so I'll follow up. 
2. What are your plans for [Enable helm tiller](https://docs.docker.com/ee/ucp/admin/configure/enable-helm-tiller/)? Does it make sense to redirect this to https://docs.docker.com/ee/ucp/user-access/kubectl/? The content looks duplicated to me but I may be missing something.

traci-morrison:
Thanks @DawnWood-Docker. @david-yu Could you also review to let us know if more changes are required? Then I will work with Dawn to update this PR and get a second technical review. Thanks.

david-yu:
We should call it Docker Enterprise and not EE. I am ok with mentioning Helm as part of the Kube CLI docs page, but we should at least move the content over to that page as opposed to assume that kubectl covers the specific ask, just wanted to be clear.

PR #9260, opened 2019-08-15
Simplified EE variables in _config

ollypom:
### Proposed changes

Before we had 2 UCP configuration variables. 1 for pages under "ee" and
1 for pages under "ee/ucp". We had the same pattern for DTR. THis commit
removes the variable at the ee/ucp level, in the hope that all files
will take it from the /ee variable.

Will check with the preview to make sure nothing has broken.

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -116,12 +116,17 @@ defaults:
       path: "datacenter"
     values:
       enterprise: true
+  # Latest Enterprise Platform Release
   - scope:
-      path: "ee/dtr"
+      path: "ee"
     values:
+      ucp_org: "docker"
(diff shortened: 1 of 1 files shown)


usha-mandya:
Thanks for simplifying this, @ollypom. I can see the correct version number within the install and upgrade pages for UCP and DTR. Is there anything else we should check as part of this implementation?

ollypom:
@DawnWood-Docker @usha-mandya Awesome thanks. Nothing specifically in terms of what to test. Just a big component to change so was curious if you found any errors when clicking around. 

Will merge now, and make sure https://docs-stage.docker.com/ looks good too :)

PR #9270, opened 2019-08-18
[sec] nginx/compose: Drop aforementioned loophole

das7pad:
### Proposed changes

Let me quote from the receipt on nginx as an authentication handler:  

> Note: Docker does not recommend binding your registry to localhost:5000 without authentication. This creates a potential loophole in your Docker Registry security. As a result, anyone who can log on to the server where your Docker Registry is running can push images without authentication

This loophole was added a few lines below. Let's drop it.

### DIFF
--- a/registry/recipes/nginx.md
+++ b/registry/recipes/nginx.md
@@ -181,8 +181,6 @@ Review the [requirements](/registry/recipes/index.md#requirements), then follow 
 
     registry:
       image: registry:2
-      ports:
-        - 127.0.0.1:5000:5000
       volumes:
         - ./data:/var/lib/registry
     ```


das7pad:
https://deploy-preview-9270--docsdocker.netlify.com/registry/recipes/nginx/

registry/recipes/nginx.md
```diff
@@ -181,8 +181,6 @@ Review the [requirements](/registry/recipes/index.md#requirements), then follow
 
     registry:
       image: registry:2
-      ports:
-        - 127.0.0.1:5000:5000
```
thaJeztah: Without any ports published, the registry won't me accessible, which makes it ... not very useful?

das7pad: One would access the registry via the nginx container.

authorized User -> nginx -> registry
unauthorized User -> nginx -> return 401

thaJeztah: Ah, you're right, sorry, read too fast

PR #9271, opened 2019-08-19
Removed extraneous sample azure.json entries

stevenfollis:
### Proposed changes

Currently we specify quite a few extra lines in the `azure.json` file that are unneeded due to the default values. 

I removed `cloudProviderBackoff` and its extraneous related values, and `useManagedIdentityExtension` since the default value is the same. 

Removing these entries tidies up and simplifies the sample `azure.json` for end users.

I also standardized on 3 `***` placeholder characters, as currently there is an inconsistent mixture of sometimes 3, and sometimes 4.

### DIFF
--- a/ee/ucp/admin/install/cloudproviders/install-on-azure.md
+++ b/ee/ucp/admin/install/cloudproviders/install-on-azure.md
@@ -86,19 +86,10 @@ parameters as is.
     "aadClientId": "***",
     "aadClientSecret": "***",
     "resourceGroup": "***",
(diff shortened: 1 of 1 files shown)


stevenfollis:
There was also an extra `/` characters in `"subnetName": "/****",` that I removed. 

All of this should be in accordance with the [docs](https://github.com/kubernetes/cloud-provider-azure/blob/master/docs/cloud-provider-config.md#cluster-config)

PR #9286, opened 2019-08-22
Update ubuntu.md

jescobar-docker:
The path to the .deb files was incorrect. It was updated. A clarification was added to the following paragraph.


traci-morrison:
Closing this PR because the Ruby version is outdated, which causes the CI check to fail. The work will be done via https://github.com/docker/docker.github.io/pull/10116.

PR #9293, opened 2019-08-22
Updating bind-mounts.md

danielartine:
Adding new info to clarify --mount vs -v. Showing that type=volume creates the endpoint for the user.


lena-larionova:
Thanks for your feedback. Unfortunately, since this page talks specifically about bind mounts, we're actually referring to the specific differences between the --v flag and the bind-type mount (--mount type=bind). We will work on clarifying this to prevent any more confusion.

PR #9329, opened 2019-08-29
Fix invocation of `dockerdesktop-admin`

mat007:
### Proposed changes

Fix invocation of `dockerdesktop-admin` for the first two instances to look like the last one.


usha-mandya:
Thanks for fixing this @mat007.

PR #9340, opened 2019-08-31
Fix the effect of `--memory-swap` is unset.

tao12345666333:
Signed-off-by: Jintao Zhang <zhangjintao9020@gmail.com>

### Proposed changes

**Fix the effect of `--memory-swap` is unset.**

If `--memory-swap` is unset, and `--memory` is set, the container can use as much swap as the `--memory` setting, if the host container has swap memory configured.
  For instance, if `--memory="300m"` and `--memory-swap` is not set, the
  container can use 300m of memory and 300m of swap.

The code is here https://github.com/moby/moby/blob/master/daemon/daemon_unix.go#L343-L346
### Related issues (optional)

Fixed https://github.com/docker/docker.github.io/pull/2854

cc @crosbymichael @mistyhacks

### DIFF
--- a/config/containers/resource_constraints.md
+++ b/config/containers/resource_constraints.md
@@ -111,9 +111,9 @@ Its setting can have complicated effects:
   [Prevent a container from using swap](#prevent-a-container-from-using-swap).
 
 - If `--memory-swap` is unset, and `--memory` is set, the container can use
-  twice as much swap as the `--memory` setting, if the host container has swap
+  as much swap as the `--memory` setting, if the host container has swap
   memory configured. For instance, if `--memory="300m"` and `--memory-swap` is
-  not set, the container can use 300m of memory and 600m of swap.
+  not set, the container can use 600m in total of memory and swap.
 
 - If `--memory-swap` is explicitly set to `-1`, the container is allowed to use
   unlimited swap, up to the amount available on the host system.


crosbymichael:
The code shows 2*memory.  So the current docs saying 600m are correct.  Why do you think it is 300m?

tao12345666333:
The code is here.

```
	if hostConfig.Memory > 0 && hostConfig.MemorySwap == 0 {
		// By default, MemorySwap is set to twice the size of Memory.
		hostConfig.MemorySwap = hostConfig.Memory * 2
	}
```

The code means set `hostConfig.MemorySwap` (it is memory + swap).

And I have do some tests.

*   `--memory="300m"` and `--memory-swap` is not set

```
(MoeLove) ➜  ~ docker run --rm -d --memory="300m" redis
2642f20501822ee37d1580a9c6e6babf35a718d8210dbcb8b2f7faabb5f107eb
(MoeLove) ➜  ~ # check its cgroups 
(MoeLove) ➜  ~ cat /sys/fs/cgroup/memory/system.slice/docker-$(docker inspect --format '{{ .Id}}' $(docker ps -ql)).scope/memory.limit_in_bytes
314572800
(MoeLove) ➜  ~ cat /sys/fs/cgroup/memory/system.slice/docker-$(docker inspect --format '{{ .Id}}' $(docker ps -ql)).scope/memory.memsw.limit_in_bytes
629145600
(MoeLove) ➜  ~ docker stop $(docker ps -ql)
2642f2050182
```

* `--memory="300m"` and `--memory-swap="600m"` : 

```
(MoeLove) ➜  ~ docker run --rm -d --memory="300m" --memory-swap="600m" redis
62e68ba6df5f8e66cc02eb075e4d2663315bf024fd406ae735241c5b341f5814
(MoeLove) ➜  ~ cat /sys/fs/cgroup/memory/system.slice/docker-$(docker inspect --format '{{ .Id}}' $(docker ps -ql)).scope/memory.limit_in_bytes
314572800
(MoeLove) ➜  ~ cat /sys/fs/cgroup/memory/system.slice/docker-$(docker inspect --format '{{ .Id}}' $(docker ps -ql)).scope/memory.memsw.limit_in_bytes
629145600
(MoeLove) ➜  ~ docker stop $(docker ps -ql)
62e68ba6df5f
```

The results of the two tests are consistent.

The cgroups `memory.limit_in_bytes` means for `memory` and `memory.memsw.limit_in_bytes` means for `memory + swap`. So this means we can only use `629145600 - 314572800`(300m) of swap.

tao12345666333:
@thaJeztah PTAL. Thanks.

config/containers/resource_constraints.md
```diff
@@ -111,9 +111,9 @@ Its setting can have complicated effects:
   [Prevent a container from using swap](#prevent-a-container-from-using-swap).
 
 - If `--memory-swap` is unset, and `--memory` is set, the container can use
-  twice as much swap as the `--memory` setting, if the host container has swap
+  as much swap as the `--memory` setting, if the host container has swap
   memory configured. For instance, if `--memory="300m"` and `--memory-swap` is
-  not set, the container can use 300m of memory and 600m of swap.
+  not set, the container can use 300m of memory and 300m of swap.
```
AkihiroSuda: "container can use 600m in total of memory and swap" ?

https://www.kernel.org/doc/Documentation/cgroup-v1/memory.txt

tao12345666333: yes.  memory + swap = 600m

AkihiroSuda: so swap can be more than 300m, right? The current text seems confusing

tao12345666333: Usually we don't directly say how many swaps are available; and the limitations of cgroups also limit memory usage and memory + swap usage.

tao12345666333: >so swap can be more than 300m,  right?

In my understanding, it should be a maximum of 300m.

tao12345666333: My initial change here was to correct it;
"container can use 600m in total of memory and swap" is more suitable, but compared to other examples, it may not be straightforward.

AkihiroSuda: Not straightforward, but it's kernel's limitation

https://www.kernel.org/doc/Documentation/cgroup-v1/memory.txt
```
* why 'memory+swap' rather than swap.
The global LRU(kswapd) can swap out arbitrary pages. Swap-out means
to move account from memory to swap...there is no change in usage of
memory+swap. In other words, when we want to limit the usage of swap without
affecting global LRU, memory+swap limit is better than just limiting swap from
an OS point of view.
```

tao12345666333: Although I think this is explained from a **design perspective**. 

I think your suggestion is reasonable :+1: and this can avoid some confusion.
Let me update this text.

tao12345666333:
ping @AkihiroSuda @thaJeztah  PTAL. Thanks.

thaJeztah:
Trying to visualise the behavior for myself, roughly what it comes down to is;

- if a system supports swapping, and no physical memory is available, memory is swapped out to disk. So, e.g., having 500MB of memory available, also needs 500MB of swap space, so that that memory can be swapped to disk if the system is under pressure
- to be able to use more memory than the amount of physical memory (RAM) available, swap space must be available _in addition_ to the amount above, so to have (e.g.) 250MB of swap, a total of 750MB of swap must be set (first 500MB to allow swapping out memory, and the additional 250MB to "overflow" and allow using more memory than physical RAM)
    - "physical RAM" here meaning: the amount of RAM that is set as limit


And from this PR (and the discussion above), my understanding is that the kernel actually _reserves_ the first 500MB to only use it if RAM has to be swapped to disk, and not for "overflowing" memory (thus using more memory than physical RAM),
so even if (example above) the swap limit is 750MB, only 250MB of that will be used by default for swapping, and the other 500MB will remain unused (until physical RAM runs out on the host).

Still wondering where we originally picked the default "swap is twice the amount of memory" from, but I guess that's not something we can change

We should think of adding a graphic to this to explain this better (as it's quite complicated / confusing)

cpuguy83:
> Still wondering where we originally picked the default "swap is twice the amount of memory" from, but I guess that's not something we can change

An old rule of thumb for swap.

thaJeztah:
Whoops, forgot to LGTM/merge. Changes look good to me with the info above. Thanks!

PR #9358, opened 2019-09-04
Sync published with master

usha-mandya:
Publish the latest updates on docs.docker.com


usha-mandya:
@DawnWood-Docker - Could you PTAL?

PR #9362, opened 2019-09-04
Minor formatting

darwin-traver:
### Proposed changes
Minor table formatting update for new switch (--file)


darwin-traver:
@ollypom Minor formatting issue with the other switch added in UCP 3.2.x (--file)

PR #9365, opened 2019-09-04
minor formatting updates

DawnWood-Docker:


DawnWood-Docker:
lol @lena-larionova we even talked about that and I still forgot. I'll fix it next time.

PR #9380, opened 2019-09-05
fixing pull location

adrian-plata:
Signed-off-by: Adrian Plata <adrian.plata@docker.com>

### DIFF
--- a/_scripts/fetch-upstream-resources.sh
+++ b/_scripts/fetch-upstream-resources.sh
@@ -54,8 +54,8 @@ ENGINE_SVN_BRANCH="branches/19.03"
 ENGINE_BRANCH="19.03"
 
 # Distribution
-DISTRIBUTION_SVN_BRANCH="branches/release/2.6"
-DISTRIBUTION_BRANCH="release/2.6"
+DISTRIBUTION_SVN_BRANCH="branches/release/2.7"
+DISTRIBUTION_BRANCH="release/2.7"
 
 # Directories to get via SVN. We use this because you can't use git to clone just a portion of a repository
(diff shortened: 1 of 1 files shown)


adrian-plata:
@ollypom @thaJeztah PTAL. I can also change the other pull distribution location for Distribution. I want to make sure this doesn't break any pages.

_scripts/fetch-upstream-resources.sh
```diff
@@ -54,7 +54,7 @@ ENGINE_SVN_BRANCH="branches/19.03"
 ENGINE_BRANCH="19.03"
 
 # Distribution
-DISTRIBUTION_SVN_BRANCH="branches/release/2.6"
+DISTRIBUTION_SVN_BRANCH="branches/release/2.7"
 DISTRIBUTION_BRANCH="release/2.6"
```
ollypom: I've just built this locally to make sure you can pull in `registry/spec/deprecated-schema-v1/` and you can with the change on line 57 :) 

Do we also need update line 58 too?

adrian-plata: that's what I was wondering. There are a few line differences between the 2.6 and 2.7 versions of configurations.md. Also, 3 years vs 8 months in terms of last updates. @thaJeztah any thoughts on this?

thaJeztah: Yes, we should change line 58 as well

I think I had a branch somewhere to get rid of these env-vars (or unify them), but they should indeed be in sync always

adrian-plata: Okay, I'll make the changes then. Thanks to you both!

PR #9406, opened 2019-09-09
Remove web interface upgrade section

millerdz:
### Proposed changes

Upgrading through the UCP UI is no longer an option as of UCP 3.2.0, so I've removed the section of instructions of that upgrade technique.


ollypom:
@millerdz Thanks Derrick :) Just pushed a second commit to remove the images and a few other references that mentioned "UI Upgrade".

Side note the formatting on this page isn't very good. My IDE removed all end of line whitespaces, but the page could do with the 80 character treatment in another PR :)

PR #9448, opened 2019-09-15
Swapped Docker Cluster Versions to Variables

ollypom:
### Proposed changes

- Made the quickstart yaml formatting consistent across AWS and Azure. 
- Swapped Docker Cluster Versions to Variables

### DIFF
--- a/cluster/aws.md
+++ b/cluster/aws.md
@@ -17,69 +17,72 @@ When you create a docker cluster in AWS, the created cluster has:
 
 Create a `cluster.yml` file with the following information:
 ```yaml
-    variable:
-      domain: "YOUR DOMAIN, e.g. docker.com"
-      subdomain: "A SUBDOMAIN, e.g. cluster"
-      region: "THE AWS REGION TO DEPLOY, e.g. us-east-1"
-      email: "YOUR.EMAIL@COMPANY.COM"
-      ucp_password:
-        type: prompt
+variable:
+  domain: "YOUR DOMAIN, e.g. docker.com"
+  subdomain: "A SUBDOMAIN, e.g. cluster"
+  region: "THE AWS REGION TO DEPLOY, e.g. us-east-1"
+  email: "YOUR.EMAIL@COMPANY.COM"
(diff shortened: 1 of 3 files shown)


cluster/aws.md
```diff
@@ -165,16 +168,16 @@ provider:
...
     storage_volume: /dev/xvdb
-    version: ee-stable-18.09.5
+    version: ee-stable-{{ site.docker_ee_version }}
   registry:
     url: https://index.docker.io/v1/
     username: user
   ucp:
     username: admin
-    version: docker/ucp:3.1.6
+    version: {{ page.ucp_org }}/{{ page.ucp_repo }}:{{ page.ucp_version }}
```
thaJeztah: Given that docker cluster should (I think) always be tied to the _current_ version, should we use `site.ucp_version` and `site.dtr_version` for these?

I _do_ notice that `site.dtr_version` is currently `2.7` (so not versioned at the _patch_ level (`2.7.2`). Not sure if that's intentional

ollypom: Hey @thaJeztah , we use the `site.ucp_version` (`3.2`) variable only for navigation I think.

Most of the enterprise docs use the page variable `page.ucp_version` (`3.2.3`) variable,  as we don't actually publish images without a patch number (i.e `docker/ucp:3.2` isn't a thing). 

To be honest I'm not sure if we want the patch number as part of the navigation, so it makes sense having 2 variables. I guess we could make 2 site variables `site.ucp_version` and `site.ucp_version_patch` or something, but that would break the archiving process so I would want to avoid that if possible. Today when an enterprise doc goes from `/ee/ucp` to `datacenter/ucp/3.1` we don't have to adjust the content, as it picks up the new `page.ucp_version`.

usha-mandya:
Thanks for your feedback. Docker Enterprise docs are now part of Mirantis. You can take a look at the Enterprise documentation at https://docs.mirantis.com/docker-enterprise/.

Let me close this PR here for that reason. Feel free to continue the conversation after I close.

PR #9453, opened 2019-09-16
Update broken links

traci-morrison:
### Proposed changes
Update broken links in https://docs.docker.com/ee/ucp/admin/install/cloudproviders/install-on-azure/ 

### Related issues (optional)
https://docker.atlassian.net/browse/ENGDOCS-122


traci-morrison:
@lena-larionova @DawnWood-Docker PTAL.

PR #9457, opened 2019-09-16
Update to move Google Tag Manager lower on the page.

erikhedin:
### Proposed changes

Hi Team,

We are seeing an issue with docs.docker.com's javascript re-writing the window.location value as the page is loading.  This change is to move GTM to the bottom of the page to allow the window.location function to run before GTM is loaded.

Please let me know the staging url once the changes are built so we can test.

Thanks for the help,

Erik

### DIFF
--- a/_layouts/docs.html
+++ b/_layouts/docs.html
@@ -53,13 +53,6 @@
 			display: block;
 		}
 	</style>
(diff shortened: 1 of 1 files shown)


erikhedin:
We were able to fix this with a helper function to patch the window.location issue.

PR #9474, opened 2019-09-18
Correct the compression default value

Gui13:
# Proposed changes

Modified the documentation for the 'compress' logger option.

Using `"compress": "disabled"` doesn't work and makes the docker service crash, `"compress": "false"` does the job. The documentation is not clear if we have to put disabled or false, so I put 'false' as default.


ChrisChinchilla:
Thanks @Gui13 but it looks like a different PR made the same changes, but thanks again for the contribution :)

PR #9478, opened 2019-09-18
Update index.md

jescobar-docker:
Changed the order for the actions to be in chronological order. Currently, it seems you have to find the replica ID before running the command. This is emphasized by the "Then run". This is not the case. You use the replica ID after you run the command.


ollypom:
@lena-larionova Following your changes in https://github.com/docker/docker.github.io/pull/9485/. PTAL :)

Issue #8892, opened 2019-06-04
404 at /engine/reference/#kernel-memory-constraints & /engine/reference/#user-memory-constraints


ve-Leon:
### Problem description
404 page appears on the above URLS.  

### Suggestions for a fix

Pages still available through /engine/reference/run/* instead of /engine/reference/*

thaJeztah:
Where did you find the link to those pages?

leonBeis:
https://docs.docker.com/engine/reference/run/
The right side bar, seems all links/anchors there are broken in the same way as the 2 mentioned in the original post.

thaJeztah:
Ah, yes, I recall looking at that in another issue; looks like a `<base href..>` is inserted in the page's HTML, which cases that;

<img width="348" alt="Screenshot 2019-06-04 at 17 47 06" src="https://user-images.githubusercontent.com/1804568/58893715-f1e37700-86f0-11e9-9ccd-f407579ae5bc.png">

thaJeztah:
Thanks again for reporting! I think this issue should now be fixed; I just tried on that page (right-click -> copy link), and they look to be good now; for example,  https://docs.docker.com/engine/reference/run/#network-settings

let me close this ticket, but feel free to continue the conversation 👍

PR #9480, opened 2019-09-18
Stop referring to deprecated 'links' feature

psy-q:
#8892 # Proposed changes

The `links` option in compose files is deprecated, `depends_on` can be used in its place in some cases (if even necessary).


thaJeztah:
@ulyssessouza @chris-crone ptal

PR #9481, opened 2019-09-18
Updated for EE 3.0

omegamormegil:
Instructions hadn't been updated for EE 3.0

### DIFF
--- a/install/linux/docker-ee/ubuntu.md
+++ b/install/linux/docker-ee/ubuntu.md
@@ -137,13 +137,15 @@ from the repository.
 
 4. Temporarily add a `$DOCKER_EE_VERSION` variable into your environment.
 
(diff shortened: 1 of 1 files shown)


ollypom:
Hey @omegamormegil Thanks for the commit. Just a few nits :) 

- I switched the EE Version to use a variable rather than being hard set on the page. 
- I removed mention of Ubuntu 14.04 and AUFS as we no longer support 14.04 on Enterprise Engine 19.03 as per https://success.docker.com/article/compatibility-matrix

Final nit - In future do you mind creating branches on a fork and PRing from there. Rather than creating branches on the upstream repo? Thanks :)

omegamormegil:
@ollypom Sorry if I'm doing it wrong, but I'm just using the built-in "Edit this page button" on the doc. Is this button not leading to the desired outcome?

edit_page

PR #9483, opened 2019-09-18
simply the experimental box

adrian-plata:
Sample original page: https://docs.docker.com/engine/reference/commandline/assemble/
Sample new page: https://deploy-preview-9483--docsdocker.netlify.com/engine/reference/commandline/assemble/

Signed-off-by: Adrian Plata <adrian.plata@docker.com>

### Proposed changes

Simplified the experimental box to reduce clutter on pages.

### DIFF
--- a/_includes/cli.md
+++ b/_includes/cli.md
@@ -47,13 +47,13 @@ your client and daemon API versions.
 
 > This command is experimental.
 >
-> This  command is experimental on the Docker client. It should not be used in
+> This command is experimental on the Docker client. It should not be used in
 > production environments.
 > To enable experimental features in the Docker CLI, edit the
 > [config.json](/engine/reference/commandline/cli.md#configuration-files)
-> and set `experimental` to `enabled`.
->
-> {% include experimental.md %}
+> and set `experimental` to `enabled`. You can go [here](https://github.com/docker/docker.github.io/blob/master/_includes/experimental.md)
+> for more information.
+{: .important }
 
 {% endif %}
 


_includes/cli.md
```diff
@@ -47,13 +47,13 @@ your client and daemon API versions.
...
 >
-> This  command is experimental on the Docker client. It should not be used in
+> This command is experimental on the Docker client. It should not be used in
 > production environments.
 > To enable experimental features in the Docker CLI, edit the
 > [config.json](/engine/reference/commandline/cli.md#configuration-files)
-> and set `experimental` to `enabled`.
->
-> {% include experimental.md %}
+> and set `experimental` to `enabled`. You can go [here](https://github.com/docker/docker.github.io/blob/master/_includes/experimental.md)
```
thaJeztah: I'm a bit confused why we're linking to GitHub for this one, instead of to a page at docs.docker.com

adrian-plata: I'm linking it there because the page at docs.docker.com lives on the CLI master repo so it hasn't been published yet. Once the changes go live, I'll link it to there.

thaJeztah: Ah, gotcha; but I thought we had the other include that has the same content https://github.com/docker/docker.github.io/blob/master/_includes/experimental-feature.md

adrian-plata: yeah I'm not sure There's not much of a difference between them. I think the one you linked was made specifically for desktop. I'm not sure though, maybe @usha-mandya might have a better idea.

thaJeztah: Looks like on the CLI reference, we use an include that looks like this; https://docs.docker.com/engine/reference/commandline/deploy/

Screenshot 2019-10-01 at 21 11 47


And there we link to the knowledgeable for a description of Experimental; https://success.docker.com/article/experimental-features

adrian-plata: Ah, dang, looks like some pages are labeled "_experimental_" and others are labeled "_experimentalcli_".

I can make _experimental_ and _experimentalcli_ the same or go into those pages and assign them _experimentalcli_, in case we want _experimental_ to be for another use case.

https://docs.docker.com/engine/reference/commandline/checkpoint/ is another with the same window.

thaJeztah: experimental is for those features that are enabled on the daemon side, whereas experimentalcli is for those that are enabled on the cli

adrian-plata: oh I see. Besides that, are there specific differences between them? How they're enabled or anything else?

usha-mandya: @adrian-plata `experimental-features.md` was the first version of the experimental features blurb. I see it only used in the Mac FAQs topic. I can update this topic to use `experimental.md` so we don't have to maintain two files. 

Agree with @thaJeztah on GitHub linking. We should point users to our docs instead. There is a new node for Experimental features in the TOC update we are working on. Wondering if we can point users to an overview page here instead of linking to GitHub when the changes are live. Thoughts?

adrian-plata: Adding a pointer to the cli experimental information section. https://github.com/docker/docker.github.io/pull/9569 

If there are no differences, it makes sense to update the pages so they're the same. Eventually, we'll probably want to use one if that's the case.

PR #9502, opened 2019-09-23
Case correction for Linux-like systems in docs

lyubick:
If using variables in uppercase this guide doesn't work, precising that at least in Linux-like (Ubuntu etc.) variables must be in lowercase.

Tested on a Ubuntu 19.04 while creating Docker image FROM 18.04 via Dockerfile.

### DIFF
--- a/network/proxy.md
+++ b/network/proxy.md
@@ -66,3 +66,12 @@ instead.
 | `FTP_PROXY`   | `ENV FTP_PROXY "ftp://127.0.0.1:3001"`            | `--env FTP_PROXY="ftp://127.0.0.1:3001"`            |
 | `NO_PROXY`    | `ENV NO_PROXY "*.test.example.com,.example2.com"` | `--env NO_PROXY="*.test.example.com,.example2.com"` |
 
+Please note if you are setting environment variables in Linux-like systems (Ubuntu etc.)
+variables should be in lowercase.
+
+| Variable      | Dockerfile example                                | `docker run` Example                                |
+|:--------------|:--------------------------------------------------|:----------------------------------------------------|
+| `HTTP_PROXY`  | `ENV http_proxy "http://127.0.0.1:3001"`          | `--env http_proxy="http://127.0.0.1:3001"`          |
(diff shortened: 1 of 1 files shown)


network/proxy.md
```diff
@@ -66,3 +66,12 @@ instead.
 | `FTP_PROXY`   | `ENV FTP_PROXY "ftp://127.0.0.1:3001"`            | `--env FTP_PROXY="ftp://127.0.0.1:3001"`            |
 | `NO_PROXY`    | `ENV NO_PROXY "*.test.example.com,.example2.com"` | `--env NO_PROXY="*.test.example.com,.example2.com"` |
 
+Please note if you are setting environment variables in Linux-like systems (Ubuntu etc.)
+variables should be in lowercase.
```
thaJeztah: I don't think this is correct; also see https://github.com/docker/docker.github.io/pull/8032#discussion_r248736355

the TL;DR is that there is no official standard for these, and some software uses upper, other uses lowercase (and most look for one, then fallback to the other); see https://unix.stackexchange.com/a/212972/78656 and https://wiki.archlinux.org/index.php/Proxy_server#Environment_variables


the best option for these is to configure the docker client to set the proxies, as it will configure _both_ upper and lowercase environment variables automatically; https://github.com/docker/docker.github.io/blob/6adba1c7c2af987df0d3fdef769f87ef98c4bed4/network/proxy.md#configure-the-docker-client

lyubick: Sorry for late response.
In overall I can agree with you, but from personal experience (I am newbie to Docker) when I was creating Dockerfile and put variables in UPPER, nothing worked to me, and I continued to google it, and found solution on Stackoverflow. IMHO it is very strange that documentation does not contains known issues.
Ok, maybe it is better just make a note mentioning that case may vary, without lower case example?

thaJeztah: Yes, a note that mentions that some software expects it to be lowercase might be more appropriate.

That said, I think we should remove the mention of using `ENV` in the Dockerfile, because it really isn't recommended doing so (because you'd only be able to run the image in an environment that has the specified proxy/proxies.

Let me have a look at rewriting this section

PR #9517, opened 2019-09-25
Add commands to export scripts

traci-morrison:
### Proposed changes
Changes to https://docs.docker.com/cluster/

Adding section on exporting scripts

### Related issues (optional)

https://docker.atlassian.net/browse/ENGDOCS-39


traci-morrison:
@david-yu @DawnWood-Docker PTAL.

PR #9519, opened 2019-09-26
Add notice about port binding and overriding of UFW to compose file reference

Kovah:
### Proposed changes

I already work with Docker for a long time, but recently found out that binding ports will override / ignore UFW firewall rules. This leads to unwanted vulnerabilities if you restrict access to a specific port via UFW, add the port to a Docker compose setup and then find out that the whole world could access the app via the port because Docker manages his own iptables rules.
As this may lead to severe vulnerabilities, I added a notice to the `ports` config section of the compose file references. Nothing too heavy, just a short notice that bound port may be accessible even if blocked via UFW.
Currently, the information is kinda buried in the /network/iptables page.

### DIFF
--- a/compose/compose-file/compose-file-v2.md
+++ b/compose/compose-file/compose-file-v2.md
@@ -1040,10 +1040,10 @@ on which platform the service's build will be performed.
 
 Expose ports. Either specify both ports (`HOST:CONTAINER`), or just the container
 port (an ephemeral host port is chosen).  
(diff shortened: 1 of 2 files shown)


Kovah:
Any news on this?

Kovah:
Just noticed that this is still open. Will this ever be merged?

Kovah:
I'll close this now. Seems you are not interested in contributions.

PR #9525, opened 2019-09-27
Missing mandatory UCP URL in the example

zaggash:


ollypom:
Hey @zaggash thank you for the PR. However I don't think it is mandatory to specify the URL in interactive mode, the DTR bootstrapper will prompt you for the URL. The `dtr join` reference doc is currently consistent with the install command.

Therefore I'll close this PR for now. However if you think it is worth specifying the UCP URL consistently across the DTR CLI reference documentation, happy to accept the PR :) Also is there a reason why you opened this on DTR 2.6 not 2.7?

PR #9529, opened 2019-09-27
daily pr from master to published

DawnWood-Docker:

### DIFF
--- a/compose/extends.md
+++ b/compose/extends.md
@@ -62,7 +62,7 @@ services.
 
     web:
       image: example/my_web_app:latest
-      links:
+      depends_on:
         - db
         - cache
 
@@ -136,7 +136,7 @@ Start with a **docker-compose.yml**.
 
     web:
       image: example/my_web_app:latest
-      links:
+      depends_on:
         - db
 
     db:
@@ -147,7 +147,7 @@ export or backup.
 
     dbadmin:
       build: database_admin/
-      links:
+      depends_on:
         - db
 
 To start a normal environment run `docker-compose up -d`. To run a database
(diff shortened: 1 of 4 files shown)


app-template/working-with-template.md
```diff
@@ -10,30 +10,47 @@ keywords: Docker, application template, Application Designer,
...
+templates. To create a custom template based on your requirements, you must
+complete the following steps:
 
 1. Create a service container image
 2. Create the service template definition
 3. Add the service template to the library
 4. Share the service template
 
-### Create a service container image
+### A service container image
```
lena-larionova: Not sure why this was changed, the previous titled seems to make more sense, and is more consistent with the other sections.

app-template/working-with-template.md
```diff
@@ -10,30 +10,47 @@ keywords: Docker, application template, Application Designer,
...
 
-A service template provides the description required by Docker Template to scaffold a project. A service template runs inside a container with two bind mounts:
+A service template provides the description required by Docker Template to
+scaffold a project. A service template runs inside a container with two bind
+mounts:
 
-1. `/run/configuration`, a JSON file which contains all settings such as parameters, image name, etc. For example:
+1. `/run/configuration`, a JSON file which contains all settings such as
```
lena-larionova: If you look at this in the preview, there are now two "step 1"s in this section.

DawnWood-Docker:
closing so that we can fix the issues that Lena raises. We'll republish once those are fixed.

app-template/working-with-template.md
usha-mandya: Thanks for spotting this, @lena-larionova. This wasn't displayed as expected due to the code between the numbered list. This is fixed in https://github.com/docker/docker.github.io/pull/9537.

app-template/working-with-template.md
usha-mandya: Yup, makes sense. The title has been updated in https://github.com/docker/docker.github.io/pull/9537.

PR #9543, opened 2019-10-01
Enhanced Export Artifacts script

stevenfollis:
### Proposed changes

- Fixed an errant line break
- Switched to more verbose `docker container X` syntax rather than shorthand aliases
- Swapped shorthand flags to `--` variants for clarity


usha-mandya:
Thanks for updating this @stevenfollis.

PR #9549, opened 2019-10-01
ucp: move client bundle verification steps

trapier:
* Move client bundle verification steps to the correct section
* Add a sentence indicating expected output of verification commands.


trapier:
doh.  i keep opening branches against the main repo.

PR #9560, opened 2019-10-02
Add note on swarm mode restore process

traci-morrison:
### Proposed changes

Adding a note to https://docs.docker.com/engine/swarm/admin_guide/ about swarm mode and CA certificates.

### Related issues (optional)

https://docker.atlassian.net/browse/ENGDOCS-167


traci-morrison:
@miao0miao @DawnWood-Docker PTAL when the preview is available. Thanks.

PR #9586, opened 2019-10-07
Update system-requirements.md

RomainBelorgey:
Specify that user namespaces should not be configured on nodes before installing UCP.

Internal issue tracking this feature: https://github.com/docker/orca/issues/3450


traci-morrison:
Work completed in https://github.com/docker/docker.github.io/pull/9948.

Issue #9587, opened 2019-10-07
Link to "Cloudstor for Azure" broken


sujaypillai:
[cluster/cluster-file.md](https://docs.docker.com/cluster/cluster-file/)

The actual link to - "Cloudstor for Azure"should be 
https://docs.docker.com/docker-for-azure/persistent-data-volumes/

instead of -
https://docs.docker.com/cluster/[/docker-for-azure/persistent-data-volumes/

sujaypillai:
Thanks for accepting my changes

PR #9588, opened 2019-10-07
Update cluster-file.md

sujaypillai:
### Proposed changes

The actual link to - "Cloudstor for Azure"should be
https://docs.docker.com/docker-for-azure/persistent-data-volumes/

instead of -
https://docs.docker.com/cluster/[/docker-for-azure/persistent-data-volumes/

### Related issues (optional)

Resolves - https://github.com/docker/docker.github.io/issues/9587


lena-larionova:
LGTM

PR #9597, opened 2019-10-08
UCP Sys Req Version Compatibility

ollypom:
### Proposed changes

This section is currently inaccurate, however instead of patching this, I felt it was a duplicate of the Compatibility Matrix linked in line 151.


traci-morrison:
@ollypom Before merging this PR, can you PTAL at this Jira issue: https://docker.atlassian.net/browse/ENGDOCS-114. My plan is to remove the duplicate section from [here](https://docs.docker.com/ee/ucp/admin/install/plan-installation/). This means the firewall info will remain in the sys requirements doc. Thoughts? Thanks.

PR #9615, opened 2019-10-10
Remove 3.2.2 section

traci-morrison:
### Proposed changes

Remove 3.2.2 section from https://docs.docker.com/ee/ucp/release-notes/

### Related issues (optional)
https://docker.atlassian.net/browse/ENGORC-2773


ry4nz:
Instead delete them, shall we still list 3.2.2 but stating that we discovered a bug and pulled out the image, for transparency?

PR #9634, opened 2019-10-14
Add small explanation of issues with traditional init systems in containers

SgtCoDFish:
### Proposed changes

Adds minor paragraph with some details on why "traditional" init systems aren't ideal for using inside docker containers.

Very open to adding more detail if required/suggested!


SgtCoDFish:
I think this has been inflight for long enough :joy:  Gonna close it.

PR #9644, opened 2019-10-15
Add brew services instructions

hansenc:
### Proposed changes

Added `brew services` instructions. I think Homebrew is common enough among developers on macOS to include here, and `brew services` makes starting local machines on startup very easy, especially for people not familiar with how `launchd` works.

None

### Related issues (optional)
None


craig-osterhout:
Thank you for contributing. Closing as Docker Machine is now deprecated.

PR #9650, opened 2019-10-15
Add link to Azure roles document

traci-morrison:
### Proposed changes

- Updating https://docs.docker.com/cluster/azure/ 
- Adding info on creating Azure roles with explicit permissions on Docker Cluster

### Related issues (optional)
https://docker.atlassian.net/browse/ENGDOCS-205


traci-morrison:
@DawnWood-Docker PTAL. I believe the correct spelling for the product name is Docker Cluster, correct? If so, I need to make those changes. I just noticed the spelling in the preview. Thanks.

PR #9623, opened 2019-10-11
adding redirect and fixing typo in part 5

DawnWood-Docker:

PR #9651, opened 2019-10-16
Update references to the get-started pages

RRoggia:
### Proposed changes

Update the develop section references to the get-started section, since the get-started was updated.

### Related issues (optional)
I didn't create an issue, but here is the problem I found.

Some changes to the get-started section (#9608, #9609, #9623) led the develop section to be outdated.

> Scale your app as a swarm service

is referring to the `get-started/part3.md` which after the changes reflect the kubernetes deployment instead of the swarm. Scaling the swarm service is now described on the `get-started/part4.md`.

Also,
> Define your app stack using a compose file

which before the changes was the `get-started/part5.md` now is not part of the scope of the get-started anymore.


ithinco:
Hi @RRoggia, do you know why stack is not in the scope of get-started? thanks

PR #9660, opened 2019-10-17
Updated 'Use a pre-existing network' section

Prashansa-K:
The already given criteria wasn't working. After experimentation, it was found that the way I proposed is the actual way we can use an already existing network in Docker Compose.

## In 'Use a pre-existing network' section
networks:
   my-pre-existing-network:
         external: true

##Earlier, the mentioned steps were: 
networks:
  default:
    external:
      name: my-pre-existing-network

## This doesn't work. So, I tried and found out that actual way should be as such:
networks:
   my-pre-existing-network:
         external: true


Prashansa-K:
@lena-larionova @DawnWood-Docker @GordonTheTurtle @adrian-plata 
Please Review

PR #9664, opened 2019-10-17
fix Desktop Edge download URLs

usha-mandya:
Fixing the Edge download URLs for Mac and Windows.

Fixes https://github.com/docker/for-win/issues/4946


traci-morrison:
LGTM

PR #9686, opened 2019-10-19
Correcting Ubuntu SSL path

josejayesh:
Correcting Ubuntu SSL  path

### DIFF
--- a/registry/insecure.md
+++ b/registry/insecure.md
@@ -129,7 +129,8 @@ certificate at the OS level.
 #### Ubuntu
 
 ```bash
-$ cp certs/domain.crt /usr/local/share/ca-certificates/myregistrydomain.com.crt
+$ mkdir -p /etc/docker/certs.d/myregistrydomain.com
+$ cp certs/domain.crt /etc/docker/certs.d/myregistrydomain.com
 update-ca-certificates
 ```
 


registry/insecure.md
```diff
@@ -129,7 +129,8 @@ certificate at the OS level.
 #### Ubuntu
 
 ```bash
-$ cp certs/domain.crt /usr/local/share/ca-certificates/myregistrydomain.com.crt
+$ mkdir -p /etc/docker/certs.d/myregistrydomain.com
+$ cp certs/domain.crt /etc/docker/certs.d/myregistrydomain.com
```
cquon: I don't think this directory is specific to the Docker daemon and not for the OS level.

http://manpages.ubuntu.com/manpages/precise/man8/update-ca-certificates.8.html

RobbKistler: There are two ways to get an engine to trust a registry.  The first is adding a root cert to the system trust store, which is what is currently documented in registry/insecure.md. 

The second way is to add it to /etc/docker/certs.d/myregistrydomain.com.  There is no need to do an "update-ca-certs" type of step.  And it might not even be necessary to restart the docker engine (I can't remember).  The location is different on windows/mac. 

The down side of the second way is that it only affects the engine.  You'd also have to tell the CLI about the cert for when it contacts the registry directly (docker trust, docker manifest).

lena-larionova:
Thanks for your contribution @josejayesh, but it looks like the current doc has the path we want to keep as the official content.

PR #9692, opened 2019-10-21
clarify pre-reqs for Windows DDE

usha-mandya:
Fixes https://docker.atlassian.net/browse/ENGDOCS-218.

### DIFF
--- a/ee/desktop/admin/install/windows.md
+++ b/ee/desktop/admin/install/windows.md
@@ -12,7 +12,15 @@ This page contains information about the system requirements and specific instru
 
 - Windows 10 Pro or Enterprise version 15063 or later.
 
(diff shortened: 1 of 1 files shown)


usha-mandya:
@jasonbivins could you PTAL?

Believe this is also true for Community Windows installation - https://docs.docker.com/docker-for-windows/install/#system-requirements.

Could you confirm so that I can update the Community version as well? Thanks.

jasonbivins:
LGTM - warning is clear and commands to enable the requirements are included.

the Community edition installer checks for and then automatically enables Containers and Hyper-V

the EE installer doesn't since it's expected that EE offices have an onsite administrator with their own infrastructure install processes.

PR #9717, opened 2019-10-22
FIx Type for HitlessServiceUpdate

arkodg:


arkodg:
PTAL @euanh

arkodg:
cc @joeabbey

PR #9721, opened 2019-10-23
Update install.md

biofool:
What should the output of sysctl be?

### Proposed changes

Document what the output should be if docker can be run
sysctl kern.hv_support


usha-mandya:
@biofool Thanks for the contribution. I've made some updates to your suggestion to fix markdown formatting and to add a description of the output.

PR #9741, opened 2019-10-26
Remove deprecated `links` feature

lekterable:
### Proposed changes

Removed deprecated `links` feature and replaced it with `depends_on` to keep the order.
I've also removed quotes around the ports, just for the consistency as it was the only string which had them even tho they are optional.


laurazard (approved):
LGTM! Thank you :)

dockertopia (approved):
Thanks @lekterable.

PR #9751, opened 2019-10-28
Clarify use of --pod-cidr and --service-cluster-ip-range

omegamormegil:
Clarified where flags are used. Customer reported confusion in case# 101388 because commands were not specified.


omegamormegil:
Kindly check the links from my edits - they are supposed to link to page https://docs.docker.com/reference/ucp/3.2/cli/install/

PR #9797, opened 2019-11-04
Add table for IP ranges

traci-morrison:
### Proposed changes

- https://docs.docker.com/ee/ucp/admin/install/plan-installation/#avoid-ip-range-conflicts
- Add table in `Avoid IP range conflicts` section
- Expand definitions for subnets

### Related issues (optional)
https://docker.atlassian.net/browse/ENGDOCS-230


traci-morrison:
@arueth @trapier PTAL.

ee/ucp/admin/install/plan-installation.md
```diff
@@ -42,11 +42,19 @@ this.
...
-Kubernetes `pod-cidr` - CIDR range for Kubernetes pods, default `192.168.0.0/16`.  
-Kubernetes `service-cluster-ip-range` - CIDR range for Kubernetes services, default `10.96.0.0/16`.
+The following table indicates which subnet configurations can safely overlap explicitly **between** clusters and which can overlap **within** a cluster.
+
+
+| Subnet                                   | Overlap between clusters             | Overlap within a cluster             |
+|------------------------------------------|------------------------------------------|------------------------------------------|
+| `default-address-pools`                  | Only accessible on the local node.  Can be the same between clusters, even on the same infra subnet. Can be the same on all nodes in a cluster. |                                          |
+| `fixed-cidr` and `bip`                   | docker0 subset of `default-address-pools` and for the purposes of avoiding subnet overlaps, potentially redundant to `default-address-pools`.  Not a required configuration for subnet overlap avoidance. Can be the same on all nodes in a cluster. |                                          |
```
trapier: Educational value of the middle column on this row is diminished its original objective, which was to express disagreement.

There's some nuance on this point that needs to be fleshed out and Aaron and I need to discuss.  As I see it:

* `docker0` address can be configured with `default-address-pools` (proof below)
* For some use case that I haven't yet imagined a user might want to have docker0 somewhere specific _within_ `default-address-pools`, or even outside `default-address-pools`, in which case the user would still have recourse to  `fixed-cidr` and `bip`.

Since the purpose of this table is to summarize what can overlap, I motion to remove "what it does" and this kind of use case information from this center column for discussion elsewhere in the doc.  If that makes sense to @arueth I'm happy to take a shot at making sure the info is present elsewhere.


```bash
# revert to default

[vagrant@worker-1 ~]$ sudo systemctl stop docker
[vagrant@worker-1 ~]$ sudo sh -c 'rm -rf /var/lib/docker/*'
[vagrant@worker-1 ~]$ sudo ip l del docker0
[vagrant@worker-1 ~]$ ls /etc/docker/daemon.json
ls: cannot access /etc/docker/daemon.json: No such file or directory
[vagrant@worker-1 ~]$ sudo systemctl start docker
[vagrant@worker-1 ~]$ ip -o -4 a sh docker0
163: docker0    inet 172.17.0.1/16 brd 172.17.255.255 scope global docker0\       valid_lft forever preferred_lft forever

# set default-address-pools

[vagrant@worker-1 ~]$ sudo systemctl stop docker
[vagrant@worker-1 ~]$ sudo sh -c 'rm -rf /var/lib/docker/*'
[vagrant@worker-1 ~]$ sudo ip l del docker0
[vagrant@worker-1 ~]$ sudo mv /tmp/daemon.json /etc/docker/daemon.json
[vagrant@worker-1 ~]$ cat /etc/docker/daemon.json
{
    "default-address-pools": [ {"base":"10.0.0.0/16","size":24}]
}
[vagrant@worker-1 ~]$ sudo systemctl start docker
[vagrant@worker-1 ~]$ ip -o -4 a sh docker0
164: docker0    inet 10.0.0.1/24 brd 10.0.0.255 scope global docker0\       valid_lft forever preferred_lft forever
```

trapier: Making that suggestion more explicit:
1. Keep the current bulleted list.  At a minimum, it documents defaults which the table does not.
2. Introduce the table below the bulleted list
3. Remove "what it does" and "use case" information from the table, making sure they are present elsewhere in the doc.

Setting "changes requested" on my review, because at a minimum the current patch removes documentation of the defaults, which we need to keep.

arueth:
I’m out this week but can discuss further and review it early next week.

traci-morrison:
@trapier I incorporated the suggested changes to the table and list. Let me know what other changes you and @arueth have after the review. Thanks.

traci-morrison:
@traci-morrison Change "IP-IP" to "IPIP"

traci-morrison:
@trapier @DawnWood-Docker PTAL.

PR #9828, opened 2019-11-08
[docs] Update json-file.md

mator:
Add "compress" option to json-log driver options and increase default file size and file count.

### Proposed changes

add "compress" option to daemon.json example


usha-mandya:
Thanks for the PR. As this is just an example, the increased file size and count aren't super critical here. Closing this PR for this reason.

PR #9830, opened 2019-11-08
update gettingstarted - docker-compose.yml use existing redis image

michalbcz:
original "redis/alpine" does not exist
docker-compose up ends up with "ERROR: pull access denied for redis/alpine, repository does not exist or may require 'docker login': denied: requested access to the resource is denied" (there is nothing in https://hub.docker.com/search?q=redis%2Falpine&type=image)

### DIFF
--- a/compose/gettingstarted.md
+++ b/compose/gettingstarted.md
@@ -119,7 +119,7 @@ the following:
         ports:
           - "5000:5000"
       redis:
-        image: "redis:alpine"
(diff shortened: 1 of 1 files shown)


lena-larionova:
It looks like redis:alpine does exist, take a look at the third tag down: https://hub.docker.com/_/redis?tab=tags

I tested it out with a `docker pull redis:alpine` and it worked. Are you sure you used the correct command? You have `redis/alpine` in your example.

PR #9837, opened 2019-11-09
Refer to the raw content from github

jose-bigio:
### Proposed changes

Refer to the raw content instead of to the github page.

Signed-off-by: Jose Bigio <jose.bigio@docker.com>


jose-bigio:
@traci-morrison @arkodg

PR #9855, opened 2019-11-12
Add notes on using the DockerCli to switch engines

espoelstra:
### Proposed changes

I added a brief node with examples on changing the engine from a CLI. I had to search all over and it wasn't easy to find any mention of this in the official documentation


usha-mandya:
@espoelstra Thanks for the PR. Not sure why a user would prefer a CLI way of switching between Windows and Linux containers when there's an easy-to-use option in Docker Desktop. As this topic is dedicated to Docker Desktop,  it would make sense to keep instructions related to Docker Desktop only in this topic. I'll have to close this PR without accepting your change for this reason.

PR #9880, opened 2019-11-14
Fix broken link

rupakg:
### Proposed changes

Fix broken link


usha-mandya:
@rupakg @lena-larionova - I believe the link should point to the Support article - https://success.docker.com/article/how-to-resolve-kubernetes-errors-after-upgrading-ucp instead of the github page on kbase. WDYT?

rupakg:
@usha-mandya I was not even aware that the page existed in success. Yes, you are right - it should point there.

PR #9881, opened 2019-11-14
logging/gelf: recommend to disable compression

kolyshkin:
It was discovered that compression in gelf logging driver
is highly inefficient in terms of CPU. IOW, lots of CPU
is used when compression is enabled.

The natural solution would be to change the default
to `none`, i.e. no compression, but that would break
some setups (in particular, using Logstash < 7.4).

The only option left is to warn the user about the
consequences of leaving the compression enabled.

For more background details, please see
https://github.com/moby/moby/pull/40101

PS it makes sense to backport this to older engine version docs, too.


kolyshkin:
@adrian-plata @lena-larionova PTAL

kolyshkin:
Looks ok but maybe there's a way to somehow emphasize this warning?
https://deploy-preview-9881--docsdocker.netlify.com//config/containers/logging/gelf/

adrian-plata:
We could bold it or possibly put it in a note carrot. That might get funky. I’ll try some alternatives tomorrow.

adrian-plata:
Unfortunately, tables tend to break most formatting so we can't use the caret. To draw more attention, I'll make it bold in another PR.

adrian-plata:
PR in  question: https://github.com/docker/docker.github.io/pull/9885

PR #9889, opened 2019-11-15
Change Golang version number

traci-morrison:
### Proposed changes

- https://docs.docker.com/ee/ucp/release-notes/
- Change Golang version number for UCP 3.2.4


traci-morrison:
@DawnWood-Docker PTAL.

PR #9897, opened 2019-11-18
updated typo in the command scale missing hiphen

msonowal:
### Proposed changes

fixed typo in the command `scale` for missing hyphen 

in

https://docs.docker.com/compose/reference/scale/


lena-larionova:
This is deliberate; in this context, `docker-compose scale web=2 worker=3` is correct, it's just deprecated. You're probably thinking of `docker-compose up --scale`, which is the current method: https://docs.docker.com/compose/reference/up/ . 

I'll update the topic to make that note about deprecation pop a bit more.

PR #9902, opened 2019-11-18
Add SAML user admin information

traci-morrison:
### Proposed changes

- https://docs.docker.com/ee/ucp/admin/configure/integrate-saml/
- Add optional `Attribute` with the name `is-admin` to identify if the user is an administrator

### Related issues (optional)

- https://docker.atlassian.net/browse/ENGDOCS-252


traci-morrison:
@ry4nz PTAL.

PR #9915, opened 2019-11-20
Add info on Kube secrets

traci-morrison:
### Proposed changes
- https://docs.docker.com/ee/ucp/ucp-architecture/
- Add info for how we handle secrets in Kubernetes
- Formatting changes

### Related issues (optional)
- https://docker.atlassian.net/browse/ENGDOCS-244

### DIFF
--- a/ee/ucp/ucp-architecture.md
+++ b/ee/ucp/ucp-architecture.md
@@ -22,8 +22,7 @@ the Docker CLI client and Docker Compose.
 
 ## Under the hood
 
-UCP leverages the clustering and orchestration functionality provided
-by Docker.
+UCP leverages the clustering and orchestration functionality provided by Docker.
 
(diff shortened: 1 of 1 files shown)


traci-morrison:
@david-yu @jlhawn @usha-mandya PTAL.

david-yu:
Hi @traci-morrison , do you know if there is also place to call out Secrets in a first class location in docs? Perhaps somewhere in a security section under Kube?

traci-morrison:
> Hi @traci-morrison , do you know if there is also place to call out Secrets in a first class location in docs? Perhaps somewhere in a security section under Kube?

Hi @david-yu We talk about Kube network encryption, but according to @jlhawn, what we have now would probably best belong in the architecture docs as it’s not configurable.

PR #9984, opened 2019-12-06
Remove "docker engine" commands

thaJeztah:
relates to https://github.com/docker/cli/issues/2213

> These subcommands were created to allow upgrading a Docker Community engine to Docker Enterprise, but never really took off. With the migration of the docker "enterprise" products to Mirantis, we should remove these subcommands as releases may no longer be done in sync in future.


justincormack:
LGTM

thaJeztah:
rebased, and moved out of draft: the commands have not yet been removed from the `docker` CLI, but we won't be supporting this workflow, so we should remove the documentation.

Issue #7630, opened 2018-11-08
Quickstart: Compose and WordPress - not working


jiajiewu1988:
[compose/wordpress.md](https://docs.docker.com/compose/wordpress/)

Hi,

I followed this tutorial and trying to setup wordpress and mysql by docker on my AWS Ubuntu vm, but after finished with all the steps, wordpress container does not start running, it keeps restarting.

Any one know if there are any issues?

bh-adrienne:
I am also having this issue, the quickstart example does not work as stated.

td-gonzales:
[This](https://github.com/docker/docker.github.io/pull/8091) pull request fixes this problem.

td-gonzales:
That pull has been merged so this issue can be closed. #8091

PR #9990, opened 2019-12-08
Add unlinked Wordpress example to Samples homepage

dannaf:
### Proposed changes

The [Wordpress example](https://docs.docker.com/compose/wordpress/) is linked from [samples-for-compose](https://docs.docker.com/compose/samples-for-compose/), along with the [Django](https://docs.docker.com/compose/django/) and [Rails](https://docs.docker.com/compose/rails/) examples, but it is not listed in this proposed-edited [Samples](https://docs.docker.com/samples/) home page, unlike those other two examples which are listed.   This is inconsistent, and can make that example harder to find.

Also the Wordpress example lacks the left sidebar navigation, which this change hopefully fixes as well.

### Related issues (optional)

Not very related, but looks like this Wordpress example could use a review for content---or simply closing of a dangling open issue about its content: https://github.com/docker/docker.github.io/issues/7630

### DIFF
--- a/samples/index.md
+++ b/samples/index.md
@@ -37,6 +37,7 @@ Run popular software using Docker.
 | [PostgreSQL](/engine/examples/postgresql_service/) | Run a Dockerized PostgreSQL instance. |
 | [Rails + PostgreSQL](/compose/rails/) | Run a Dockerized Rails + PostgreSQL environment. |
 | [Riak](/engine/examples/running_riak_service/) | Run a Dockerized Riak instance. |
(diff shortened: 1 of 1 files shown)


usha-mandya:
@dannaf Thanks again for your contribution. The Wordpress example link appears broken in the preview - https://deploy-preview-9990--docsdocker.netlify.com/samples/.

Could you fix this? Thanks.

craig-osterhout:
Thanks for sharing the update. Closing as the content has diverged too much.

PR #10006, opened 2019-12-11
Update configure-rbac-kube.md

scjane:


traci-morrison:
Thank you for taking the time to review and update our documentation. We consider the use of gerunds and participles to be acceptable in our documentation. Therefore, I'm closing this PR without merging it.

PR #10008, opened 2019-12-11
Commandline reference: add note to "enterprise only" subcommands

thaJeztah:
fixes https://github.com/docker/cli/issues/2217

The `assemble`, `cluster`, `registry`, and `template` subcommands are provided by plugins that only ship with the enterprise engine / cli.

This patch annotates those commands to print a note that they only ship with the enterprise version of docker;

Screenshot 2019-12-11 at 10 20 09

### DIFF
--- a/_includes/cli.md
+++ b/_includes/cli.md
@@ -29,6 +29,15 @@ your client and daemon API versions.
 
 {% endif %}
 
+{% if page.enterprise_only == true %}
+
+> This command is only available on Docker Enterprise Edition.
+>
+> Learn more about [Docker Enterprise products](/ee/supported-platforms/){: target="_blank" class="_"}.
+{: .important }
+
+{% endif %}
+
 {% if site.data[include.datafolder][include.datafile].experimental %}
 
 > This command is experimental.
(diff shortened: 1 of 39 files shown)


thaJeztah:
@usha-mandya @nebuk89 @chris-crone PTAL

_includes/cli.md
```diff
@@ -29,6 +29,15 @@ your client and daemon API versions.
 
 {% endif %}
 
+{% if page.enterprise_only == true %}
+
+> This command is only available on Docker Enterprise Edition.
+>
+> Learn more about [Docker Enterprise products](/ee/supported-platforms/){: target="_blank" class="_"}.
```
usha-mandya: @thaJeztah  Should we point users to the Enterprise Overview page ( /ee/ ) instead of /ee/supported-platforms/?

thaJeztah: I was a bit in doubt, because the `/ee/supported-platforms/` is the one that's underneath the "engine" section in the TOC (and also linked from the homepage)

Do you think the other one is better to link to?

usha-mandya: Yea, there is an overlap between `/ee/supported platforms` and `/ee/`. For Enterprise CLI plugins, I think it makes sense to point to the Overview page underneath Engine. 

I'll approve the PR now. Thanks.

PR #10016, opened 2019-12-11
Update ASP.NET Core samples to use version 3.1

mthalman:
Updates the documentation for ASP.NET Core samples so that they are based on .NET Core 3.1.  At its most basic, this involves just referencing the 3.1 image tags.  But since there were breaking changes from 2.x to 3.x, the instructions needed to be updated to account for those changes.

### DIFF
--- a/compose/aspnet-mssql-compose.md
+++ b/compose/aspnet-mssql-compose.md
@@ -14,7 +14,7 @@ and [Docker Compose](/compose/install/) installed on your
 platform of choice: Linux, Mac or Windows.
 
 For this sample, we create a sample .NET Core Web Application using the
-`microsoft/dotnet:2.1-sdk` Docker image. After that, we create a `Dockerfile`,
+`mcr.microsoft.com/dotnet/core/sdk:3.1` Docker image. After that, we create a `Dockerfile`,
 configure this app to use our SQL Server database, and then create a
 `docker-compose.yml` that defines the behavior of all of these components.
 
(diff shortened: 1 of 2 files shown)


mthalman:
@MichaelSimons, @richlander - Please review

compose/aspnet-mssql-compose.md
```diff
@@ -29,25 +29,28 @@ configure this app to use our SQL Server database, and then create a
...
 
-    ```conf
-    FROM microsoft/dotnet:2.1-sdk
+    ```dockerfile
+    FROM mcr.microsoft.com/dotnet/core/sdk:3.1
+    RUN ["dotnet", "tool", "install", "--global", "dotnet-ef"]
+    ENV PATH "~/.dotnet/tools/:$PATH"
     COPY . /app
     WORKDIR /app
     RUN ["dotnet", "restore"]
```
MichaelSimons: The pattern used here is very different than in the [asp.net sample](engine/examples/dotnetcore.md).  I realize this PR is about upgrading to 3.1 but it is something that jumped out at me.

mthalman: That's basically because the EF migrations are being used when the container is run.  It's very much an "interactive development" kind of sample.  Since this is just about getting the existing samples in a working state on 3.1, I'm not going to concern myself with this aspect of the sample at this time.

craig-osterhout:
Thanks for sharing the update. Closing this PR as this content will be removed from the docs repo.
Most of the samples have moved outside of the docs to https://github.com/docker/awesome-compose
Within the docs, a new .net-based guide was added at https://docs.docker.com/language/dotnet/
If you have any feedback on the new content, please let us know.

PR #10030, opened 2019-12-13
Update the metadata and keywords for SEO

usha-mandya:
Updated the robots.txt and keywords in Docker Desktop docs for SEO.

### DIFF
--- a/docker-for-mac/install.md
+++ b/docker-for-mac/install.md
@@ -1,6 +1,6 @@
 ---
 description: How to install Docker Desktop on Mac
-keywords: mac, beta, alpha, install, download
+keywords: mac, install, download, run, docker, local, machine
 title: Install Docker Desktop on Mac
 ---
 
--- a/robots.txt
+++ b/robots.txt
@@ -33,6 +33,4 @@ Disallow: /datacenter/ucp/1.1
 
 Disallow: /apidocs/v*
 
-Disallow: /machine/
-
 Sitemap: https://docs.docker.com/sitemap.xml
--- a/robots.txt
+++ b/robots.txt
@@ -33,4 +33,6 @@ Disallow: /datacenter/ucp/1.1
 
 Disallow: /apidocs/v*
 
+Disallow: /machine/
+
 Sitemap: https://docs.docker.com/sitemap.xml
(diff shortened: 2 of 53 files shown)


robots.txt
```diff
@@ -33,4 +33,6 @@ Disallow: /datacenter/ucp/1.1
 
 Disallow: /apidocs/v*
 
+Disallow: /machine/
```
thaJeztah: Wasn't there an option to set this in the metadata of the files that should not be indexed? (that way we wouldn't have to update the robots.txt)

docker-for-mac/install.md
```diff
@@ -1,6 +1,6 @@
 ---
 description: How to install Docker Desktop on Mac
-keywords: mac, beta, alpha, install, download
+keywords: mac, install, download, run, docker, local, machine
```
nebuk89: do we have a different page that will win download instead?

robots.txt
usha-mandya: Not aware of the option to set the metadata of the file. Could you provide more details @thaJeztah?

docker-for-mac/install.md
usha-mandya: Do you mean a different page where we provide Desktop downloads?

robots.txt
thaJeztah: looks if we set `hide_from_sitemap` in the metadata; https://github.com/docker/docker.github.io/blob/24718c5e47eb85b7b219d34f13a4460df3616984/_layouts/docs.html#L105

docker-for-mac/install.md
nebuk89: As discussed other pages that are better suited have the install tag

robots.txt
usha-mandya: I guess it has to be set on a per-page level and not at the directory level? Any idea if affects both  docs.docker.com and Google search results? Thanks.

usha-mandya: Removed the update from robots.txt after discussing with @thaJeztah.

usha-mandya:
Sanity checked Docker Machine pages in the Netlify preview to ensure the pages render as expected. They look good.
https://deploy-preview-10030--docsdocker.netlify.com/machine/overview/

thaJeztah:
Thanks!

PR #10084, opened 2019-12-24
In line 657,it should be can't instead of can

ShivangGoswami:
### Proposed changes
In line 657,it should be can't instead of can.
"Can" in place of "can't" is making the meaning of the paragraph confusing.In additional,it is in contrast with the example provided.


usha-mandya:
@ShivangGoswami Thank you for updating the documentation.

ShivangGoswami:
@usha-mandya Thank you for approving my PR.

PR #10122, opened 2020-01-09
Should be 'back-end can reach it' judging from the context

johnsonexosite:
### Proposed changes

Judging from the context, it means to reach the database. So, the word to be used should be "back-end".


usha-mandya:
Thanks for the PR. The `master` branch has been renamed to `main`. Some of the existing PRs have been automatically closed as a result of this change. Could you create a new PR against our latest docs if the suggested changes are still relevant? Thank you.

PR #10137, opened 2020-01-11
Update minimal MacOS Version requirement

Weyx:
The new minimal requirement for running full version of Docker is 10.12.0 Sierra, not Yosemite.

### Proposed changes

Changed the docker minimal MacOS version requirement in Docker Toolbox wiki.


usha-mandya:
@Weyx Thank you for taking the time to update docs. The minimum version of macOS is now 10.13. See https://docs.docker.com/docker-for-mac/install/#system-requirements.

Could you update the file to reflect the updated min macOS version? Thanks.

Weyx:
I have updated the minimum version to MacOS 10.13 High Sierra

PR #10144, opened 2020-01-14
Add formatting for OL

traci-morrison:
### Proposed changes
- https://docs.docker.com/ee/ucp/swarm/
- Created an ordered list (OL) for the steps


traci-morrison:
@Dawn-Wood @KoryKessel-Docker PTAL

PR #10167, opened 2020-01-20
Update content_trust.md

mnimm:
To make it clear that DCT does not protect against image layer tampering.

### Proposed changes

The documentation suggests that DCT protects against changes to images on the local disk. However, it does not. DCT only protects against compromised repositories, but not against attacks on the local filesystem. So far, there was a warning saying that DCT does not protect RUNNING containers. However, it does not even protect any containers or images locally. This should be clear so people do not make wrong assumptions leading to security issues.

-

### Related issues (optional)

-


mnimm:
Replaced by https://github.com/docker/docker.github.io/pull/10168

PR #10168, opened 2020-01-20
Update content_trust.md

mnimm:
Make it clear that DCT does not protect against layer tampering on the docker host.

### Proposed changes

The documentation suggests that DCT protects against changes to images on the local disk. However, it does not. DCT only protects against compromised repositories, but not against attacks on the local filesystem. So far, there was a warning that DCT does not protect RUNNING containers. However, it does not even protect any containers or images locally on a docker host. This should be clear so people do not make wrong assumptions leading to security issues.

-

### Related issues (optional)

-

### DIFF
--- a/engine/security/trust/content_trust.md
+++ b/engine/security/trust/content_trust.md
@@ -20,7 +20,7 @@ client-side or runtime verification of the integrity and publisher of specific
 image tags.
 
 Through DCT, image publishers can sign their images and image consumers can
-ensure that the images they use are signed. Publishers could be individuals
+ensure that the images they pull are signed. Publishers could be individuals
 or organizations manually signing their content or automated software supply
 chains signing content as part of their release process.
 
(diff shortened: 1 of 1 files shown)


mcandre:
I vote have `docker run` emit a warning, terminate prior to the image entrypoint, exit non-zero, and offer a CLI flag to forcibly override in the event of breaking digital signatures.

mnimm:
@mcandre: afaik the problem is that `docker run` cannot verify image signatures of locally cached images, because images get unpacked once they are pulled. So this is works-as-designed according to the Docker security team.
Also see my blogpost here: https://www.impidio.com/blog/manipulating-signed-docker-images

mcandre:
Ah, okay.

In that case, Docker could maintain a mapping between those.

usha-mandya:
@mnimm Thanks for the update.

PR #10180, opened 2020-01-23
Remove $ from Ubuntu install code examples

ryan-blunden:
Currently, you can't just copy an entire code snippet and paste because the $ prevents the command from executing successfully. This requires precise cursor control to only select the text after `$`.

Removing the `$` from code snippets fixes this.

Starting with changes to a single file to get feedback, but I would love to see this should be done for all docs.

### DIFF
--- a/install/linux/docker-ce/ubuntu.md
+++ b/install/linux/docker-ce/ubuntu.md
@@ -46,7 +46,7 @@ Older versions of Docker were called `docker`, `docker.io`, or `docker-engine`.
 If these are installed, uninstall them:
 
 ```bash
-$ sudo apt-get remove docker docker-engine docker.io containerd runc
+sudo apt-get remove docker docker-engine docker.io containerd runc
 ```
 
 It's OK if `apt-get` reports that none of these packages are installed.
@@ -95,13 +95,13 @@ from the repository.
 1.  Update the `apt` package index:
 
     ```bash
-    $ sudo apt-get update
+    sudo apt-get update
     ```
 
 2.  Install packages to allow `apt` to use a repository over HTTPS:
 
     ```bash
-    $ sudo apt-get install \
+    sudo apt-get install \
         apt-transport-https \
         ca-certificates \
         curl \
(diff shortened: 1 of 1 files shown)


ryan-blunden:
Thanks for describing the problem in a larger context and other related UX challenges I had not considered.

> The tricky bit is that for some of these, removing the prompt "works". If an example contains only actual instructions that are copy/paste-able, that's ok.

This would indicate we can immediately improve the UX of these docs by removing the $ for these specific examples.

> some examples need specific information filled in (e.g., see the <VERSION_STRING> in one of the examples in this PR, which cannot be hard-coded (or easily be "filled in" in the docs))

I understand the challenge here, but how does keeping the $ help?

> some examples contain both actual commands to execute, and example output of the commands. Having a prompt before the commands to execute helps distinguishing them from the output

I agree this is a tougher problem and one we could tackle as en entirely separate issue, researching other open source projects to see how they have addressed this.

Removing the from `$` from single command examples would improve the docs UX today [in thousands of code snippets](https://sourcegraph.com/search?q=repo%3A%5Egithub%5C.com%2Fdocker%2Fdocker%5C.github%5C.io%24+%60%60%60bash%5Cn%5Cs*%5C%24+count%3A1000&patternType=regexp) and I'd ask you to reconsider this PR in terms of focussing on progress, not a perfect solution.

ryan-blunden:
It's worth noting that there are [many examples of the `$` being omitted](https://sourcegraph.com/search?q=repo:%5Egithub%5C.com/docker/docker%5C.github%5C.io%24+%60%60%60bash%5Cn%5Cs*%5Cw%2B+count:1000&patternType=regexp).

ryan-blunden:
What do you think @thaJeztah?

usha-mandya:
@ryan-blunden Thanks for the PR. The `master` branch has been renamed to `main`. Some of the existing PRs have been automatically closed as a result of this change. Could you create a new PR against our latest docs if the suggested changes are still relevant? Thank you.

PR #10184, opened 2020-01-23
Add info on ssh agent usage

usha-mandya:
### Proposed changes

Added info on how users can access the host's SSH agent inside containers.

### Related issues (optional)

https://docker.atlassian.net/browse/DESKTOP-1928

### DIFF
--- a/docker-for-mac/faqs.md
+++ b/docker-for-mac/faqs.md
@@ -78,24 +78,7 @@ For more information and examples, see
 
 ### Can I use an SSH agent inside a container?
 
-Yes, you can use the host’s SSH agent inside a container. To do this:
-
-1. Bind mount the SSH agent socket by adding the following parameter to your `docker run` command:
-
-    `-v /run/host-services/ssh-auth.sock:/run/host-services/ssh-auth.sock`
-
-1. Add the `SSH_AUTH_SOCK` environment variable in your container:
-
-    `-e SSH_AUTH_SOCK="/run/host-services/ssh-auth.sock"`
-
-To enable the SSH agent in Docker Compose, add the following flags to your service:
-
- ```
-    volumes:
-      - /run/host-services/ssh-auth.sock:/run/host-services/ssh-auth.sock
-    environment:
-      - SSH_AUTH_SOCK=/run/host-services/ssh-auth.sock
- ```
+Yes, you can use the host’s SSH agent inside a container. For more information, see [SSH agent forwarding](/docker-for-mac/networking/#ssh-agent-forwarding).
 
 ### How do I add custom CA certificates?
 
--- a/docker-for-mac/faqs.md
+++ b/docker-for-mac/faqs.md
@@ -76,10 +76,32 @@ We recommend that you publish a port, or connect from another container. You can
 For more information and examples, see
 [I want to connect to a container from the Mac](networking.md#i-want-to-connect-to-a-container-from-the-mac) in the [Networking](/docker-for-mac/networking/) topic.
 
+### Can I use an SSH agent inside a container?
+
+Yes, you can use the host’s SSH agent inside a container. To do this:
+
+1. Bind mount the SSH agent socket by adding the following parameter to your `docker run` command:
+
+    `-v /run/host-services/ssh-auth.sock:/run/host-services/ssh-auth.sock`
+
+1. Add the `SSH_AUTH_SOCK` environment variable in your container:
+
+    `-e SSH_AUTH_SOCK="/run/host-services/ssh-auth.sock"`
+
+To enable the SSH agent in Docker Compose, add the following flags to your service:
+
+ ```
+    volumes:
+      - /run/host-services/ssh-auth.sock:/run/host-services/ssh-auth.sock
+    environment:
+      - SSH_AUTH_SOCK=/run/host-services/ssh-auth.sock
+ ```
+
 ### How do I add custom CA certificates?
 
 Docker Desktop supports all trusted certificate authorities (CAs) (root or intermediate). For more information on adding server and client side certs, see
 [Add TLS certificates](/docker-for-mac/index/#adding-tls-certificates) in the Getting Started topic.
+
 
 ### How do I add client certificates?
 
--- a/docker-for-mac/faqs.md
+++ b/docker-for-mac/faqs.md
@@ -78,7 +78,7 @@ For more information and examples, see
 
 ### Can I use an SSH agent inside a container?
 
-Yes, you can use the host’s SSH agent inside a container. For more information, see [SSH agent forwarding](/docker-for-mac/networking/#ssh-agent-forwarding).
+Yes, you can use the host’s SSH agent inside a container. For more information, see [SSH agent forwarding](/docker-for-mac/osxfs/#ssh-agent-forwarding).
 
 ### How do I add custom CA certificates?
 
--- a/docker-for-mac/faqs.md
+++ b/docker-for-mac/faqs.md
@@ -85,7 +85,6 @@ Yes, you can use the host’s SSH agent inside a container. For more information
 Docker Desktop supports all trusted certificate authorities (CAs) (root or intermediate). For more information on adding server and client side certs, see
 [Add TLS certificates](/docker-for-mac/index/#adding-tls-certificates) in the Getting Started topic.
 
-
 ### How do I add client certificates?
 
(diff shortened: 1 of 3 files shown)


docker-for-mac/faqs.md
```diff
@@ -76,11 +76,33 @@ We recommend that you publish a port, or connect from another container. You can
...
+    `-e SSH_AUTH_SOCK="/run/host-services/ssh-auth.sock"`
+
+To enable the SSH agent in Docker Compose, add the following flags to your service:
+
+ ```
+    volumes:
+      - /run/host-services/ssh-auth.sock:/run/host-services/ssh-auth.sock
+    environment:
+      - SSH_AUTH_SOCK=/run/host-services/ssh-auth.sock
+ ```
```
AkihiroSuda: docker build --ssh should be mentioned as well?

guillaumerose: I couldn't make it work. Any idea how we can do it ?
This sock is only accessible by the daemon in the VM. 

```
docker build --ssh default=/run/host-services/ssh-auth.sock -t foo .
could not parse ssh: [default=/run/host-services/ssh-auth.sock]: stat /run/host-services/ssh-auth.sock: no such file or directory
```

AkihiroSuda: build --ssh is for the sock on the client

usha-mandya:
Moved the details to the 'Networking' topic to avoid cluttering the FAQ page with instructions.

docker-for-mac/networking.md
```diff
@@ -39,9 +39,30 @@ $ docker run -p 8000:80 -d nginx
 Now, connections to `localhost:8000` are sent to port 80 in the container. The
 syntax for `-p` is `HOST_PORT:CLIENT_PORT`.
 
+### SSH agent forwarding
```
thaJeztah: Should this topic be in either "file system sharing" (https://docs.docker.com/docker-for-mac/osxfs/) or a separate page? (it's not directly related to `networking`)

docker-for-mac/networking.md
```diff
@@ -39,9 +39,30 @@ $ docker run -p 8000:80 -d nginx
...
+
+Docker Desktop for {{Arch}} allows you to use the host’s SSH agent inside a container. To do this:
+
+1. Bind mount the SSH agent socket by adding the following parameter to your `docker run` command:
+
+    `-v /run/host-services/ssh-auth.sock:/run/host-services/ssh-auth.sock`
```
thaJeztah: Although this notation is "easier", the shorthand `-v <host-path>:<container-path>` has known problems with race-conditions, where `<host-path>` does not (yet) exist, it would create a _directory_ with the given name.

Not sure if it applies here, but the long form syntax does not have this problem (it would produce an _error_ instead of silently creating the path).

Should we promote using the long-hand syntax? (see https://docs.docker.com/storage/bind-mounts/#choose-the--v-or---mount-flag)

```bash
--mount type=bind,src=/run/host-services/ssh-auth.sock,target=/run/host-services/ssh-auth.sock
```

docker-for-mac/networking.md
```diff
@@ -39,9 +39,30 @@ $ docker run -p 8000:80 -d nginx
...
+
+1. Add the `SSH_AUTH_SOCK` environment variable in your container:
+
+    `-e SSH_AUTH_SOCK="/run/host-services/ssh-auth.sock"`
+
+To enable the SSH agent in Docker Compose, add the following flags to your service:
+
+ ```
+    volumes:
```
thaJeztah: Given that `volumes:` can either be at the top-level (to _define_ a volume) or under a service (to _use_ a volume), perhaps we should show a more complete example (with `services:`)?

    ```yaml
    services:
      web:
        image: nginx:alpine
        volumes:
          - type: bind
            source: /run/host-services/ssh-auth.sock
            target: /run/host-services/ssh-auth.sock
        environment:
          - SSH_AUTH_SOCK=/run/host-services/ssh-auth.sock
    ```

docker-for-mac/faqs.md
thaJeztah: Yes, where possible, we should also point people to the `--ssh` option for `docker build` (when using buildkit) to promote safe sharing of credentials. I think we still may have to add proper documentation for that though? https://github.com/docker/cli/pull/1419)

guillaumerose: Thanks for the link @thaJeztah, it works for me now :)
Maybe we can add a note that this is only required for old build system ?

docker-for-mac/networking.md
guillaumerose: I didn't know that! Thanks. 

The `/run/host-services/ssh-auth.sock` path is always present. It is created at VM startup time before dockerd boot. I will add a ticket to double-check if a race condition can happen.

docker-for-mac/networking.md
guillaumerose: Thanks, it's a better example!

docker-for-mac/networking.md
guillaumerose: Yes!

docker-for-mac/networking.md
thaJeztah: The 'auto creating host paths' was a "feature" (but causing more troubles than benefits); we tried deprecating/removing it, but some people rely on it. That's why we picked the new (`--mount`) syntax to "break" the feature (but continue having the feature for the "old" `-v foo:bar` shorthand).

It's mostly problematic when bind-mounting _files_ (bind-mounting files is troublesome in all cases, because they are based on inodes, not the _path_, so editing the file could mean that a _new_ file is created at the location, but the container still mounting the old file).

In this case, one possible approach could be to mount the _directory_ containing the socket, so

```bash
-v /run/host-services/:/run/host-services/
```

if `/run/host-services` is a regular directory inside the VM (so not a mount on its own), then worst case, the directory is created when the container is started (but the socket not being there yet)

docker-for-mac/faqs.md
usha-mandya: Looks like we need to double check the `build --ssh` option before documenting it as supported. Please let me know when this is complete and I'll update the docs.

docker-for-mac/networking.md
usha-mandya: I've moved the content to the file system sharing topic as suggested.

docker-for-mac/networking.md
usha-mandya: Updated the example

docker-for-mac/networking.md
usha-mandya: Updated the example to use --mount notation

usha-mandya:
@thaJeztah @guillaumerose @mat007 Thanks for your review. I've updated the topic based on the suggestions noted above. PTAL.

docker-for-mac/faqs.md
```diff
@@ -76,11 +76,16 @@ We recommend that you publish a port, or connect from another container. You can
...
+### Can I use an SSH agent inside a container?
+
+Yes, you can use the host’s SSH agent inside a container. For more information, see [SSH agent forwarding](/docker-for-mac/osxfs/#ssh-agent-forwarding).
+
 ### How do I add custom CA certificates?
 
 Docker Desktop supports all trusted certificate authorities (CAs) (root or intermediate). For more information on adding server and client side certs, see
 [Add TLS certificates](/docker-for-mac/index/#adding-tls-certificates) in the Getting Started topic.
 
+
```
mat007: nit: extra line?

usha-mandya: Fixed :)

docker-for-mac/faqs.md
```diff
@@ -76,6 +76,10 @@ We recommend that you publish a port, or connect from another container. You can
 For more information and examples, see
 [I want to connect to a container from the Mac](networking.md#i-want-to-connect-to-a-container-from-the-mac) in the [Networking](/docker-for-mac/networking/) topic.
 
+### Can I use an SSH agent inside a container?
```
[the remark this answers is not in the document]
usha-mandya: Created https://docker.atlassian.net/browse/ENGDOCS-326 to track this and the question raised above to separate the sockets.

docker-for-mac/faqs.md
usha-mandya: Created https://docker.atlassian.net/browse/ENGDOCS-326 to track this work item.

PR #10197, opened 2020-01-26
Adding --delete-untagged parameter description

villevaltonen:
### Proposed changes

Based on the pull request https://github.com/docker/distribution/pull/2302 there is a parameter for cleaning up untagged manifests available. In the original pull request it is called "-m" / "--delete-manifests", but in the current version of registry and it's garbage-collector, it's called "--delete-untagged" (checked with "registry garbage-collect -h" inside the latest registry-image based container). Anyway, it was missing from the documentation so I thought it would be useful to add there.

Changes added on the page "https://deploy-preview-10197--docsdocker.netlify.com/registry/garbage-collection/".

### Related issues (optional)

https://github.com/docker/distribution/issues/2301
https://github.com/docker/distribution/pull/2302

### DIFF
--- a/registry/garbage-collection.md
+++ b/registry/garbage-collection.md
@@ -90,10 +90,10 @@ This type of garbage collection is known as stop-the-world garbage collection.
 
 Garbage collection can be run as follows
 
-`bin/registry garbage-collect [--dry-run] /path/to/config.yml`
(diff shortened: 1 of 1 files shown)


villevaltonen:
Closing after waiting some time.

PR #10214, opened 2020-01-29
Fix heading in UCP release notes

traci-morrison:
### Proposed changes

- https://docs.docker.com/ee/ucp/release-notes/
- Fix the Kubernetes heading in the 3.2.5 section


traci-morrison:
@KoryKessel-Mirantis PTAL.

PR #10221, opened 2020-01-30
Update links on where to download Docker

thorerik:
Since the hub Onboarding overlay links back to the docs, it just becomes a logical loop.

I've updated the links to link to each product's edition on the hub, rather than the onboarding overlay


stephen-turner:
I haven't understood the problem with the current URL. I was able to download Desktop from there with no loop.

thorerik:
It's been fixed in the onboarding overlay since I made the PR ¯\\\_(ツ)_/¯

stephen-turner:
Ah, that explains it! Thanks.

PR #10225, opened 2020-01-31
Fix Liquid syntax error

traci-morrison:
### Proposed changes
- https://docs.docker.com/ee/ucp/interlock/usage/service-clusters/
- Fixes the Liquid syntax error when staging docs

### Related issues (optional)
- https://docker.atlassian.net/browse/ENGDOCS-328


traci-morrison:
@BillMills @Dawn-Wood PTAL

PR #10226, opened 2020-02-01
Fix missing comma

TribalNightOwl:
### Proposed changes

Added a missing comma.


dHannasch:
Your added comma is definitely an improvement, but might it be better to write "Any boolean values (true, false, yes, no) need to be enclosed in quotes"?

usha-mandya:
Thank you @TribalNightOwl and @dHannasch. :)

PR #10228, opened 2020-02-01
Update overlay.md

### LINKED ISSUE #8038: Re-frame the "secure out of the box" phrase regarding swarm overlay networking
[engine/userguide/networking/overlay-security-model.md](https://docs.docker.com/engine/userguide/networking/overlay-security-model/)

"Docker swarm mode overlay network security model: **Overlay networking for Docker Engine swarm mode comes secure out of the box** ... // You can also encrypt data exchanged between containers on different nodes on the overlay network."

I would like to see this often-repeated (and summarised) phrase "secure out of the box" re-framed so that it is completely clear that it only applies to the activity of swarm nodes exchanging control plane information (use of the secure Gossip protocol) rather than to the general swarm overlay network's data plane, for which encryption must be enabled explicitly.

For example, it is more correct and informative to say "The control plane coordinating overlay networking for Docker Engine swarm mode comes secure out of the box."

It would be better to not have any bold statements from the official documentation that could be read/spoken in isolation (i.e. without wider context) that unintentionally mislead those who read/hear them, especially for security-related matters.

terryburton:
Closes #8038.

### DIFF
--- a/network/overlay.md
+++ b/network/overlay.md
@@ -11,9 +11,9 @@ redirect_from:
 The `overlay` network driver creates a distributed network among multiple
 Docker daemon hosts. This network sits on top of (overlays) the host-specific
 networks, allowing containers connected to it (including swarm service
-containers) to communicate securely. Docker transparently handles routing of
-each packet to and from the correct Docker daemon host and the correct
-destination container.
+containers) to communicate securely when encryption is enabled. Docker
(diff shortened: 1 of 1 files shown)


usha-mandya:
@arkodg Could you PTAL?

usha-mandya:
Thank you @terryburton and @arkodg!

PR #10232, opened 2020-02-04
Docker Machine is no longer installed by default

ivorcummings:
Docker Machine is no longer installed by default along with other Docker products

### Proposed changes

Update machine/install-machine.md to reflect changes to Docker products.


usha-mandya:
@ivorcummings Thanks for updating the documentation.

PR #8135, opened 2019-02-04
Update old apt-get command

RukkiesMan:
### Proposed changes

This tutorial uses the old `apt-get` command. This commit updates that command to the new command `apt`.

PR #10258, opened 2020-02-07
Replace apt-get with apt in install instructions

rangeli:
### Proposed changes

Replaced `apt-get` with `apt` in Ubuntu install instructions.


usha-mandya:
@rangeli Thank you for taking the time to update our docs. Unfortunately, we won't be able to merge your changes due to the reasons mentioned in #8135. Thanks again for your feedback. We hope to see more PRs from you. :)

PR #10262, opened 2020-02-10
refactor build

jdrouet:
### Proposed changes

In order to keep having previews and simplify the build process, I implemented the documentation with github-actions.
When a PR is opened, the result of the build is sent to netlify and a link is published on the PR in order to have a privew.
When the PR is merged to master, the result of the build is sent to github pages (in branch `/gh-pages`)

### What's done

- [x] build on gh actions
- [x] deploy to netlify when opening PR
- [x] clean the netlify site when the PR is merged/closed
- [ ] notify when PR is deployed on netlify (can only be tested on upstream)

### DIFF
--- a/.github/actions/netlify-clean/Dockerfile
+++ b/.github/actions/netlify-clean/Dockerfile
@@ -1,0 +1,8 @@
+FROM node:lts-alpine
+
+RUN apk add --no-cache jq
+RUN npm i -g netlify-cli
+
+COPY entrypoint.sh /entrypoint.sh
+
+CMD ["/entrypoint.sh"]
--- a/.github/actions/netlify-deploy/action.yml
+++ b/.github/actions/netlify-deploy/action.yml
@@ -22,8 +22,8 @@ outputs:
 runs:
   using: 'docker'
   image: 'Dockerfile'
-  env:
-    NETLIFY_ACCOUNT_SLUG: ${{ inputs.NETLIFY_ACCOUNT_SLUG }}
-    NETLIFY_AUTH_TOKEN: ${{ inputs.NETLIFY_AUTH_TOKEN }}
-    NETLIFY_DIRECTORY: ${{ inputs.NETLIFY_DIRECTORY }}
-    NETLIFY_SITE_NAME: ${{ inputs.NETLIFY_SITE_NAME }}
+  args:
+    - ${{ inputs.NETLIFY_ACCOUNT_SLUG }}
+    - ${{ inputs.NETLIFY_AUTH_TOKEN }}
+    - ${{ inputs.NETLIFY_DIRECTORY }}
+    - ${{ inputs.NETLIFY_SITE_NAME }}
--- a/.github/actions/netlify-deploy/action.yml
+++ b/.github/actions/netlify-deploy/action.yml
@@ -1,0 +1,29 @@
+# action.yml
+name: 'Deploy to netlify'
+description: 'Creates a site and deploy to it'
+inputs:
+  NETLIFY_ACCOUNT_SLUG:
+    description: 'Account slug to link the site to'
+    required: true
+  NETLIFY_AUTH_TOKEN:
+    description: 'Authentication token to netlify'
+    required: true
+  NETLIFY_DIRECTORY:
+    description: 'Directory to deploy to site'
+    required: true
+  NETLIFY_SITE_NAME:
+    description: 'Name to give to the site that will be created'
+    required: true
+outputs:
+  site_id:
+    description: 'id of the site'
+  site_url:
+    description: 'url of the site'
+runs:
+  using: 'docker'
+  image: 'Dockerfile'
+  env:
+    NETLIFY_ACCOUNT_SLUG: ${{ inputs.NETLIFY_ACCOUNT_SLUG }}
+    NETLIFY_AUTH_TOKEN: ${{ inputs.NETLIFY_AUTH_TOKEN }}
+    NETLIFY_DIRECTORY: ${{ inputs.NETLIFY_DIRECTORY }}
+    NETLIFY_SITE_NAME: ${{ inputs.NETLIFY_SITE_NAME }}
--- a/.github/workflows/deploy-master.yml
+++ b/.github/workflows/deploy-master.yml
@@ -1,0 +1,15 @@
+on:
+  push:
+    branches:
+      - master
+
+jobs:
+  build-static-page:
+    name: build
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: build image
+        run: docker build -t documentation:latest .
+      - name: copy static files locally
+        run: docker run -v ${PWD}:/output documentation:latest cp -r /usr/share/nginx/html /output/_site
--- a/.github/workflows/deploy-master.yml
+++ b/.github/workflows/deploy-master.yml
@@ -7,6 +7,8 @@ jobs:
   build-static-page:
     name: build
     runs-on: ubuntu-latest
+    env:
+      DOCKER_BUILDKIT: '1'
     steps:
       - name: print docker info
         run: docker version && docker info
--- a/.github/workflows/deploy-master.yml
+++ b/.github/workflows/deploy-master.yml
@@ -8,6 +8,8 @@ jobs:
     name: build
     runs-on: ubuntu-latest
     steps:
+      - name: print docker info
+        run: docker version && docker info
       - uses: actions/checkout@v2
       - name: build image
         run: docker build --build-arg=ENABLE_ARCHIVES=true --target=current -t documentation:latest .
--- a/.github/workflows/deploy-master.yml
+++ b/.github/workflows/deploy-master.yml
@@ -13,3 +13,5 @@ jobs:
         run: docker build -t documentation:latest .
       - name: copy static files locally
         run: docker run -v ${PWD}:/output documentation:latest cp -r /usr/share/nginx/html /output/_site
+      - name: upload files to S3 bucket
+        run: aws s3 sync _site s3://docs.docker.com-test-us-east-1/ --delete
--- a/.github/workflows/deploy-master.yml
+++ b/.github/workflows/deploy-master.yml
@@ -10,7 +10,7 @@ jobs:
     steps:
       - uses: actions/checkout@v2
       - name: build image
-        run: docker build -t documentation:latest .
+        run: docker build --build-arg=ENABLE_ARCHIVES=true --target=current -t documentation:latest .
       - name: copy static files locally
         run: docker run -v ${PWD}:/output documentation:latest cp -r /usr/share/nginx/html /output/_site
       - name: upload files to S3 bucket
--- a/.github/workflows/deploy-pr.yml
+++ b/.github/workflows/deploy-pr.yml
@@ -1,0 +1,22 @@
+name: deploy website to netlify
+
+on: [pull_request]
+
+jobs:
+  build-static-page:
+    name: build
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: build image
+        run: docker build -t documentation:latest .
+      - name: copy static files locally
+        run: docker run -v ${PWD}:/output documentation:latest cp -r /usr/share/nginx/html /output/_site
+      - name: deploy to netlify
+        uses: ./.github/actions/netlify-deploy
+        id: netlify_deploy
+        with:
+          NETLIFY_ACCOUNT_SLUG: ${{ secrets.NETLIFY_ACCOUNT_SLUG }}
+          NETLIFY_AUTH_TOKEN: ${{ secrets.NETLIFY_AUTH_TOKEN }}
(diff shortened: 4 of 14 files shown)


anusha-ragunathan:
Looks like this PR was meant to be reviewed by someone else. potentially @usha-mandya ?

jdrouet:
@anusha-ragunathan sorry, I typed `usha` and the magic of the autocomplete happened... That's exactly what I meant.

thaJeztah:
Some questions;

- Should this be using a separate Dockerfile, so that there's still a Dockerfile to build it locally?
- The new version doesn't appear to be including the archives, or fetch the upstream resources (files from other repositories); correct?
- There's a script to cleanup/fix URL's to prevent URLs pointing to `docs.docker.com` (instead making them relative); is that still preserved?

usha-mandya:
@thaJeztah - Do we need the archive functionality after Enterprise docs are removed from our repo?

jdrouet:
@thaJeztah good point! I went too fast, my bad. Fixed that ;)

.github/workflows/deploy-pr.yml
```diff
@@ -0,0 +1,29 @@
...
+jobs:
+  build-static-page:
+    name: build
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v1
+      - name: build image
+        run: |
+          docker build -t documentation:latest .
+          docker run -v ${PWD}:/output documentation:latest cp -r /usr/share/nginx/html /output/_site
```
thaJeztah: Does the docker version running have Buildkit support? If so, we could use `DOCKER_BUILDKIT=1 docker build --output ...`

StefanScherer: I think so, it actually is the moby 3.0.6(?) compiled by MSFT

thaJeztah: We'll probably need to modify the Dockerfile to have a specific target for that (buildstage "from scratch" with only the content)

usha-mandya:
@jdrouet Can we check if it's feasible to add the markdown lint check as part of GitHub actions? https://medium.com/@christinavhastenrath/github-actions-markdown-lint-setup-86d7f6ec934d

.github/workflows/deploy-master.yml
```diff
@@ -0,0 +1,15 @@
...
+jobs:
+  build-static-page:
+    name: build
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v1
+      - name: build image
+        run: docker build -t documentation:latest .
+      - name: copy static files locally
+        run: docker run -v ${PWD}:/output documentation:latest cp -r /usr/share/nginx/html /output/_site
```
thaJeztah: I opened https://github.com/docker/docker.github.io/pull/10331 to allow using the `--output` option, and to have the option of building without archives, or to build the archives separately (which could be useful if we want to build those "once", and only update the "current" docs for updates.

.github/workflows/deploy-pr.yml
```diff
@@ -0,0 +1,29 @@
...
+          NETLIFY_AUTH_TOKEN: ${{ secrets.NETLIFY_AUTH_TOKEN }}
+          NETLIFY_DIRECTORY: _site
+          NETLIFY_SITE_NAME: "${{ github.repository }}/${{ github.head_ref }}"
+      - name: notify website is live
+        uses: ./.github/actions/create-status-check
+        with:
+          SITE_URL: ${{ steps.netlify_deploy.outputs.site_url }}
+          GITHUB_REPOSITORY: ${{ github.repository }}
+          GITHUB_SHA: ${{ github.sha }}
+          GITHUB_TOKEN: ${{ github.token }}
```
thaJeztah: nit: missing newline at end of file 😅 (can sometimes be problematic when using git)

thaJeztah: I see some other files are also having this btw

usha-mandya:
Thanks Jeremie. Can we also create a job to clear the cache when we merge PRs from master to published? It currently takes 30-40 minutes for the changes to appear live on the docs.docker.com. It would be great if we can add this to the GH actions pipeline.

CC @StefanScherer

.github/workflows/deploy-pr.yml
jdrouet: fixed!

.github/workflows/deploy-master.yml
StefanScherer: As a next step we could upload the `_site` to the test S3 bucket.
It maybe be simple like this:

```yaml
      - name: upload files to S3 bucket
        run: aws s3 sync _site s3://docs.docker.com-test-us-east-1/ --delete
```

.github/workflows/deploy-master.yml
```diff
@@ -0,0 +1,17 @@
...
+      - master
+
+jobs:
+  build-static-page:
+    name: build
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: build image
+        run: docker build -t documentation:latest .
```
thaJeztah: This can now be;

```suggestion
        run: DOCKER_BUILDKIT=1 docker build --build-arg=ENABLE_ARCHIVES=true --target=deploy-source --output=/output/_site .
```

.github/workflows/deploy-master.yml
```diff
@@ -0,0 +1,17 @@
...
+      - name: copy static files locally
```
thaJeztah: This step is no longer needed if the above is used

.github/workflows/deploy-pr.yml
```diff
@@ -0,0 +1,29 @@
...
+on: pull_request
+
+jobs:
+  build-static-page:
+    name: build
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: build image
+        run: docker build -t documentation:latest .
```
thaJeztah: Assuming we're not interested in building the archives for pull requests, this can now be:

```suggestion
        run: DOCKER_BUILDKIT=1 docker build --build-arg=ENABLE_ARCHIVES=false --target=deploy-source --output=/output/_site .
```

.github/workflows/deploy-pr.yml
```diff
@@ -0,0 +1,29 @@
...
+      - name: copy static files locally
```
thaJeztah: This step is no longer needed with the above change

StefanScherer:
I'm currently watching the PR build, the step "deploy to netlify" takes > 24 minutes, maybe we could skip the old docs versions (folders `v17.06`, ... `v1*`) ?

thaJeztah:
@StefanScherer I just left a comment with (assuming buildkit is supported) steps to build without archives; https://github.com/docker/docker.github.io/pull/10262#discussion_r383359542 😅

thaJeztah:
Hmm. some failure 🤔 

```
#16 exporting to client
215
#16 copying files 6.62kB done
216
#16 ERROR: failed to send stat compose/compose-file/compose-versioning: EOF
217
------
218
 > exporting to client:
219
------
220
failed to send stat compose/compose-file/compose-versioning: EOF
221
##[error]Process completed with exit code 1.
```

thaJeztah:
Perhaps it would be good to have a `docker version` and `docker info` step, so that we know what it's running on? (have a bit more info in case of bugs/failures)

.github/workflows/deploy-master.yml
jdrouet: doesn't seem to work with builkit https://github.com/docker/docker.github.io/runs/466748285?check_suite_focus=true 🤔

StefanScherer:
A `docker version` shows

```
Run docker version
Client:
 Version:           3.0.10+azure
 API version:       1.40
 Go version:        go1.12.14
 Git commit:        99c5edceb48d64c1aa5d09b8c9c499d431d98bb9
 Built:             Tue Nov  5 00:55:15 2019
 OS/Arch:           linux/amd64
 Experimental:      false

Server:
 Engine:
  Version:          3.0.10+azure
  API version:      1.40 (minimum version 1.12)
  Go version:       go1.12.14
  Git commit:       ea84732a77
  Built:            Fri Jan 24 20:08:11 2020
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          v1.2.11
  GitCommit:        f772c10a585ced6be8f86e8c58c2b998412dd963
 runc:
  Version:          1.0.0-rc10
  GitCommit:        dc9208a3303feef5b3839f4323d9beb36df0a9dd
 docker-init:
  Version:          0.18.0
  GitCommit:        fec3683
```

This is a build by Azure team for GH actions and Azure DevOps runners based on moby/moby. 🤷‍♂

thaJeztah:
`API version:      1.40` - looks like that should be corresponding with Docker 19.03, but no idea what patch version 🤔 

@jdrouet could you;

- keep a `docker version` and `docker info` step? I think it's good to keep such a step, to get information about the environment we're running in
- try enabling buildkit for all steps again? (also for the `deploy to netlify` step)

Wondering if the `--output` failure was either a memory issue on the machines, or a deadlock in the daemon (there was a bug in buildkit, possibly related)

.github/actions/netlify-deploy/action.yml
```diff
@@ -22,8 +22,8 @@ outputs:
...
   using: 'docker'
   image: 'Dockerfile'
-  env:
-    NETLIFY_ACCOUNT_SLUG: ${{ inputs.NETLIFY_ACCOUNT_SLUG }}
-    NETLIFY_AUTH_TOKEN: ${{ inputs.NETLIFY_AUTH_TOKEN }}
-    NETLIFY_DIRECTORY: ${{ inputs.NETLIFY_DIRECTORY }}
-    NETLIFY_SITE_NAME: ${{ inputs.NETLIFY_SITE_NAME }}
+  args:
+    - "${{ inputs.NETLIFY_ACCOUNT_SLUG }}"
+    - "${{ inputs.NETLIFY_AUTH_TOKEN }}"
```
thaJeztah: These will be passed as arguments for `docker run`; are these sanitized/removed from the output? e.g. I see 

```
/usr/bin/docker run --name e87b52a6ddbf692d905bf172f9691ee9985c2b_2fbd7a --label e87b52 --workdir /github/workspace --rm -e DOCKER_BUILDKIT -e INPUT_NETLIFY_ACCOUNT_SLUG -e INPUT_NETLIFY_AUTH_TOKEN -e INPUT_NETLIFY_DIRECTORY -e INPUT_NETLIFY_SITE_NAME -e HOME -e GITHUB_REF -e GITHUB_SHA -e GITHUB_REPOSITORY -e GITHUB_RUN_ID -e GITHUB_RUN_NUMBER -e GITHUB_ACTOR -e GITHUB_WORKFLOW -e GITHUB_HEAD_REF -e GITHUB_BASE_REF -e GITHUB_EVENT_NAME -e GITHUB_WORKSPACE -e GITHUB_ACTION -e GITHUB_EVENT_PATH -e RUNNER_OS -e RUNNER_TOOL_CACHE -e RUNNER_TEMP -e RUNNER_WORKSPACE -e ACTIONS_RUNTIME_URL -e ACTIONS_RUNTIME_TOKEN -e GITHUB_ACTIONS=true -v "/var/run/docker.sock":"/var/run/docker.sock" -v "/home/runner/work/_temp/_github_home":"/github/home" -v "/home/runner/work/_temp/_github_workflow":"/github/workflow" -v "/home/runner/work/docker.github.io/docker.github.io":"/github/workspace" e87b52:a6ddbf692d905bf172f9691ee9985c2b  "" "" "_site" "docker/docker.github.io/refactor-build"
```

.github/actions/netlify-clean/Dockerfile
```diff
@@ -0,0 +1,8 @@
+FROM node:lts-alpine
```
thaJeztah: Should we use a current version? I notice this produces some warnings:

```
npm WARN deprecated core-js@2.6.11: core-js@<3 is no longer maintained and not recommended for usage due to the number of issues. Please, upgrade your dependencies to the actual version of core-js@3.
29
```

.github/actions/netlify-deploy/action.yml
jdrouet: `"" ""` are the `NETLIFY_ACCOUNT_SLUG` and `NETLIFY_AUTH_TOKEN`. Github actions does hide them by default but here it seems that they are not transferred properly...

.github/actions/netlify-clean/Dockerfile
jdrouet: We only install `netlify-cli` (latest) and we don't have a way to force the installation of `core-js@>=3`. We could make a PR to `netlify-cli` to fix that but it might be a deeper dependency causing that.

.github/actions/netlify-deploy/action.yml
thaJeztah: Passing as an option is still potentially dangerous; they could be logged, etc., so if we can avoid it, we should

jdrouet: By default, github actions inject the secret *only* for PR made by collaborators, so the risk is kind of handled this way.

PR #10263, opened 2020-02-10
improve description of iptables design, add notes about securing Docker hosts

Svedrin:
Hi,

I've added a few paragraphs to [the "Docker and iptables"](https://docs.docker.com/network/iptables/) page that go into the details of how `FORWARD`, `DOCKER-USER` and `DOCKER` relate and how people can secure Docker hosts effectively, without being surprised by their rules being ineffective.

If you have any comments, I'd be happy to address them :)

Cheers,
Michael.

### DIFF
--- a/network/iptables.md
+++ b/network/iptables.md
@@ -68,6 +68,17 @@ the source and destination. For instance, if the Docker daemon listens on both
 topic. See the [Netfilter.org HOWTO](https://www.netfilter.org/documentation/HOWTO/NAT-HOWTO.html)
 for a lot more information.
 
+## Docker on a router
+
+Docker also sets the policy for the `FORWARD` chain to `DROP`. If your Docker
+host also acts as a router, this will result in that router not forwarding
+any traffic anymore. If you want your system to continue functioning as a
+router, you can add explicit `ACCEPT` rules to the `DOCKER-USER` chain to
+allow it:
+
+```bash
+$ iptables -I DOCKER-USER -i src_if -o dst_if -j ACCEPT
+```
 
 ## Prevent Docker from manipulating iptables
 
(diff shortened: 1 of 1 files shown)


Svedrin:
sorry for pushing more and more commits after creating the PR :sweat_smile: I think I'm done now.

Svedrin:
turns out I wasn't... but, I hope I am now. :stuck_out_tongue:

network/iptables.md
```diff
@@ -43,20 +61,49 @@ $ iptables -I DOCKER-USER -m iprange -i ext_if ! --src-range 192.168.1.1-192.168
...
+
+By default, the Docker daemon will expose ports on the `0.0.0.0` address, i.e.
+any address on the host. If you want to change that behavior to e.g. only
+expose ports on an internal IP address, you can use the `--ip` option to
+specify a different IP address. This method heightens security (connections
+are not allowed on IPs the service is not bound to), without the necessity
+for disabling the iptables mechanisms altogether, because they do honor this
+setting and do not expose services on IPs they shall not be exposed on.
+
+However, setting `--ip` only changes the _default_,
```
arkodg: I don't think this paragraph adds much value , especially the other container management systems section. 

LGTM on the rest

Svedrin: @arkodg thanks for the review! I shortened that section and removed the part in question. I'd like to keep the part that explains that `--ip` only sets a default though, because I tried to use that option to prevent my containers from going public, and it surprised me that this didn't work. What do you think? Is there a section that would fit better, where we could move it to?

Svedrin:
@arkodg do I need to do anything else to have the PR merged? sorry if this is documented somewhere, it's my first PR here, I might have missed something. :)

arkodg:
@Svedrin, just needs another pair of 👀 
PTAL @usha-mandya @thaJeztah

usha-mandya:
LGTM. Thank you @Svedrin and @arkodg.

Svedrin:
thanks @arkodg @usha-mandya !

PR #10277, opened 2020-02-12
Mention how to change restart policy on running containers

hn3000:
### Proposed changes

Mention that `docker update` can change the restart policy for an already running container.

Otherwise, the reader may be led to believe the policy can only be set when the container is started (which used to be true before version 1.11?).


hn3000:
Is there any chance of getting this reviewed?

PR #10285, opened 2020-02-13
ubuntu: add instructions for updating package repositories

thaJeztah:
Add some warnings/recommendations for non-LTS Ubuntu releases, and provide instructions for updating the package repository after upgrading.

### DIFF
--- a/install/linux/docker-ce/ubuntu.md
+++ b/install/linux/docker-ce/ubuntu.md
@@ -40,6 +40,25 @@ versions:
 Docker Engine - Community is supported on `x86_64` (or `amd64`), `armhf`, `arm64`, `s390x`
 (IBM Z), and `ppc64le` (IBM Power) architectures.
 
+### Intermediate (non-LTS) Ubuntu versions
+
+Ubuntu intermediate (non-LTS) releases are [supported for 9 months](https://ubuntu.com/about/release-cycle){: target="_blank" class="_" },
+after which they reach end-of-life (EOL). The package repositories at download.docker.com
+may contain packages for older Ubuntu releases that reached EOL. These packages
+may be outdated, and do not receive updates (including security updates).
+
+For production use, we recommend installing the latest Ubuntu LTS version. If you
+are running a non-LTS version of Ubuntu, make sure to upgrade to the latest supported
+version. Be aware that Docker's release cycle may not align with Ubuntu releases,
+which means that packages for the latest Ubuntu release may not be available at
+the day of release. Before installing or upgrading, verify that Docker packages
+are available for your version of Ubuntu.
+
+When upgrading your Ubuntu version on a machine that already has Docker installed,
+the `apt` repository may need to be updated to match the upgraded Ubuntu version.
+Refer to the [update the repository after upgrading Ubuntu versions](#update-the-repository-after-upgrading-ubuntu-versions)
+section below for details.
+
 ### Uninstall old versions
 
 Older versions of Docker were called `docker`, `docker.io`, or `docker-engine`.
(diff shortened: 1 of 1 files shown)


install/linux/docker-ce/ubuntu.md
```diff
@@ -304,6 +323,40 @@ Continue to [Post-installation steps for Linux](/install/linux/linux-postinstall
...
+     origin download.docker.com
+```
+
+If the Ubuntu version in the output does not match the version you have installed,
+the repository URL needs to be updated. To update the package repository to match
+your current version, either removing all lines containing `download.docker.com`
+in `/etc/apt/sources.list`, or use the `add-apt-repository --remove` command:
+
+```bash
+sudo add-apt-repository --remove "deb {{ download-url-base }} disco stable"
```
thaJeztah: @seemethere @tianon wondering if there's a way to remove _all_ `download.docker.com` (using a wildcard?) references, and if we should add such a step to the `get.docker.com` installation script.

thaJeztah:
@usha-mandya @seemethere @StefanScherer ptal 🤗

install/linux/docker-ce/ubuntu.md
tianon: IMO we should just make sure `get.docker.com` always uses the same filename, which solves the issue in a simpler way for folks who use the script.

We could do something clever here with `find`+`sed` to remove all references, but I think just an example `grep` to help them find the offending entries would be cleaner (and puts the onus of accidentally screwing up `sources.list` on the user instead 😅😇).

thaJeztah: > IMO we should just make sure get.docker.com always uses the same filename, which solves the issue in a simpler way for folks who use the script.

You mean; instead of having separate directories/repositories for each version of the distro (`disco`, `xenial`, ..), have all packages in the same location (not exactly sure how the indexes work internally for `.deb` repositories 😅)? Or do you mean; make sure we always write the  repository to a file inside `/etc/apt/sources.list.d` instead of append it to `/etc/apt/sources.list/` ?

> We could do something clever here with find+sed to remove all references, but I think just an example grep to help them find the offending entries would be cleaner (and puts the onus of accidentally screwing up sources.list on the user instead

Yes, so that was my main train of thought with using the `add-apt-repository --remove` utility to just remove it (assuming that's relatively safe, and wouldn't do harm to other repositories), followed by adding the new / right repository.

I could have a look at providing a `grep` example, or just keep it at "do whatever you usually do, but you likely want to be looking in this file" (as I have currently tried to write it); wdyt?

seemethere:
iirc get.docker.com always puts the apt repos into
/etc/apt/sources.list.d/docker.list

thaJeztah:
Yes, I was surprised `add-apt-repository` doesn't do that (I found some posts indicating that that's what it would do as well)

seemethere:
Yeah the only file that needs to be changed is the `/etc/apt/sources.list.d/docker.list`

We should probably just change the installation instructions to actually just point users to do the same echo command we use in get.docker.com

https://github.com/docker/docker-install/blob/3674c8a1294892f11798c857f46ed20baed37396/install.sh#L348-L358

install/linux/docker-ce/ubuntu.md
tianon: > Or do you mean; make sure we always write the repository to a file inside `/etc/apt/sources.list.d` instead of append it to `/etc/apt/sources.list/` ?

Yeah, I just mean `/etc/apt/sources.list.d/docker.list` consistently like @seemethere mentioned. :+1:

> I could have a look at providing a `grep` example, or just keep it at "do whatever you usually do, but you likely want to be looking in this file" (as I have currently tried to write it); wdyt?

Yeah, my thought was something as simple as `grep -rnF 'docker.com' /etc/apt/sources.list*`, which should tell them where they've got an entry for it already:

```console
$ grep -rnF 'docker.com' /etc/apt/sources.list*
/etc/apt/sources.list.d/docker.list:1:deb [ arch=amd64 ] https://download.docker.com/linux/debian buster stable
```

I'm definitely a fan of avoiding `add-apt-repository` because IMO it's extraneous (and isn't installed in minimal systems).

ArthurFlag:
@thaJeztah I think we can close this PR considering its age. Wdyt?

PR #10287, opened 2020-02-13
Re-enable "anchorlink" script

thaJeztah:
This allows for easier sharing of links to specific sections on a page.

With this change, hovering over H1, H2, or H3 headings will reveal a small "anchor" icon/link, which can be used to copy/paste to reference that section (e.g. https://docs.docker.com/install/linux/docker-ce/ubuntu/#install-from-a-package):

Screenshot 2020-02-13 at 13 46 45

### DIFF
--- a/_layouts/docs.html
+++ b/_layouts/docs.html
@@ -295,7 +295,7 @@
 		hljs.initHighlightingOnLoad();
 	</script>
 	{% endif %}
-	<!-- <script src="/js/anchorlinks.js"></script> -->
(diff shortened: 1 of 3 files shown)


thaJeztah:
FWIW; we had this enabled in the past, but it was disabled in favour of the right-hand side navigation (which also works, but I personally find less intuitive to find a link to a specific section) 😅 😊

thaJeztah:
@usha-mandya ptal 🤗

PR #10288, opened 2020-02-13
Remove highlightjs in favour of rouge highlighting

thaJeztah:
The current version of Rouge we use (note: we should update to a more current version even), now supports highlighting Dockerfiles, so I don't think we need highlightjs anymore.

This patch removes highlightjs (it was already disabled by default), as it didn't seem to be used anywhere, and cleans up / unifies some CSS files.

CSS might still need some tweaking, so opening as draft for now


develop/develop-images/dockerfile_best-practices.md
```diff
@@ -418,7 +418,7 @@ The following examples show the different acceptable formats. Explanatory commen
 > Strings with spaces must be quoted **or** the spaces must be escaped. Inner
 > quote characters (`"`), must also be escaped.
 
-```Dockerfile
```
thaJeztah: looks like rouge is case-sensitive; we should look if newer versions are more flexible (GitHub itself is not case-sensitive, so perhaps that's the case)

PR #10304, opened 2020-02-17
fix yaml formatting

usha-mandya:
Attempt to fix yaml formatting


usha-mandya:
@StefanScherer Could you PTAL and let me know if there are additional updates required? Thanks.

StefanScherer:
@usha-mandya Yes, that LGTM.

before:
before

after:
after

I opened VSCode and I also have markdownlint extension installed. I'm surprised it does not warn about the two spaces for multi paragraph enumerations. I looked into the markdownlint docs and found https://github.com/markdownlint/markdownlint/blob/master/docs/RULES.md#md030---spaces-after-list-markers, from what I can read there, yes two spaces should be used if the enumeration has multiple paragraphs. But to see the warning in VSCode I had to create a `.markdownlint.json` file with

```json
{
    "default": true,
    "MD013": { "line_length": -1 },
    "MD030": { "ol_multi": 2 }
}
```

to see warnings like I saw in Atom:

vscode

Beyond that Atom + remark-lint has warned about the potential wrong indentation of the yaml block when there is only one space after the enumeration.

atom

So the solution is either this PR to add one space at each enumeration or remove one space for all the backtick code blocks. I think this PR is simpler to achieve. Maybe we should also look through the docs for similar places.

thaJeztah:
Nice find! Interesting that it prepends the space, based on the numbered list 🤔

PR #10307, opened 2020-02-17
Update index.md

savijatv1:
The title for part2.md in the welcome section changed to 'Build and run your Docker image'. It goes without saying, but it is still better to specify Docker in the title.
2. This sentence is a repetition - Containers are not new, but their use for easily
deploying applications is. Suggest deleting.
3. Here, in this orientation topic, you've used the second person 'you', but in the next topic, you are changing it to first person 'we' (for example, the first sentence in the next topic goes like 'now that we've installed..'. Suggest using the second person throughout.

### DIFF
--- a/get-started/index.md
+++ b/get-started/index.md
@@ -61,7 +61,7 @@ Welcome! We are excited that you want to learn Docker. The Docker Quickstart tra
 
 1. Set up your Docker environment (on this page)
 
-2. [Build and run your image](part2.md)
+2. [Build and run your Docker image](part2.md)
 
 3. [Share images on Docker Hub](part3.md)
 


savijatv1:
One more comment. The CLI references at the end of the page are not shown as hyperlinks because of the code format. On the next page (part 2) it shows.

usha-mandya:
@savijatv1 Thank you for the suggestions. As you've noted earlier, it is implied that we are referring to Docker images. Also, as we try and keep the title, toc, and the hyperlink consistent, adding `Docker` in all these instances would be a repetition. Therefore, I'm closing this PR without merging it.

Appreciate your feedback on the difference in hyperlinks between the Orientation and part-2 topics. I'll create a separate PR to fix these.

PR #10314, opened 2020-02-18
machine: update default image for Rackspace driver

cardoe:
Updated the documented default image for the Rackspace driver because the prior image is deprecated and can no longer be used. This goes hand in hand with the change in https://github.com/docker/machine/pull/4786


craig-osterhout:
Thank you for contributing. Closing as no longer applicable.

Issue #8973, opened 2019-06-21
Documentation implies --icc=false by default


tammersaleh:
### Problem description

The first bullet in [Differences between user-defined bridges and the default bridge](https://docs.docker.com/network/bridge/#differences-between-user-defined-bridges-and-the-default-bridge) ("User-defined bridges provide better isolation and interoperability between containerized applications") is incorrect or misleading.

It says that two containers on the default bridge network cannot talk to each other unless their ports are explicitly exposed via `-p` or `-P`.  However, [the inter-container communication](https://docs.docker.com/v17.09/engine/userguide/networking/default_network/container-communication/#communication-between-containers) document says that `--icc=true` is set by default, allowing such communication.  My testing confirms this:

##### The Dockerfile:

``` dockerfile
FROM python:3.7.3
RUN pip3 install flask==1.0.2
WORKDIR /app
COPY app.py .
EXPOSE 8000
ENV PORT=8000
CMD ["./app.py"]
```

#### The commands:

``` console
$ docker network ls --filter=driver=bridge
NETWORK ID    NAME   DRIVER   SCOPE
3761c22399bc  bridge bridge   local

$ docker build -qt app .
sha256:e12465b379ec...

$ docker run -d --name app app
ec3cd0fc4cd62d0d774...
```

(note the lack of `-p` or `-P` above)

``` console
$ docker inspect app | grep IPAddress
"IPAddress": "172.17.0.2",

$ docker run superorbital/bash curl -s 172.17.0.2:8000
{"hello":"world"}
```

Maximization:
This issue can be closed now as the bullet point has been removed by https://github.com/docker/docker.github.io/pull/10320. Not sure why Github didn't auto-close on merge

usha-mandya:
Closing this issue. Thank you @Maximization!

PR #10320, opened 2020-02-19
Remove incorrect difference between user-defined and default bridge

### LINKED ISSUE #8437: Difference between docker0 and user-defined network not exists?
[network/bridge.md](https://docs.docker.com/network/bridge/)

In this part, the article told me that if I use the default bridge network, I need to use -p flag to open container's port to another container, but if I use user-defined bridge network I do not need do that. Because the containers which connect to a user-defined bridge network open ports to each other by default. However, I had created 2 containers connected to docker0 which is the default bridge network, The command was:
`docker container run -it -d centos1 centos`
`docker container run -it -d centos2 centos`
 I found that this two containers open ports to each too.  After installed the openssh-server and openssh-clients,  I can use ssh connect to another container between them. 

Docker version was docker-ce-18.09 and OS was centos 7

Maximization:
### Proposed changes

This PR removes the first bullet point from the list differences between user-defined bridge networks and the default bridge network:

[User-defined bridges provide better isolation and interoperability between containerized applications](https://docs.docker.com/network/bridge/#differences-between-user-defined-bridges-and-the-default-bridge)

The default bridge network, similar to user-defined bridges, opens all ports to other containers and allows for inter-container communication by default. This is elaborated and demonstrated in #8973. Other relevant issue: #8437.

### Related issues (optional)

Closes #8437 and #8973

### DIFF
--- a/network/bridge.md
+++ b/network/bridge.md
@@ -34,25 +34,6 @@ network.**
 
 ## Differences between user-defined bridges and the default bridge
 
-- **User-defined bridges provide better isolation and interoperability between containerized applications**.
-
-  Containers connected to the same user-defined bridge network automatically
-  expose **all ports** to each other, and **no ports** to the outside world. This allows
-  containerized applications to communicate with each other easily, without
-  accidentally opening access to the outside world.
-
-  Imagine an application with a web front-end and a database back-end. The
-  outside world needs access to the web front-end (perhaps on port 80), but only
-  the back-end itself needs access to the database host and port. Using a
-  user-defined bridge, only the web port needs to be opened, and the database
-  application doesn't need any ports open, since the web front-end can reach it
-  over the user-defined bridge.
-
-  If you run the same application stack on the default bridge network, you need
-  to open both the web port and the database port, using the `-p` or `--publish`
-  flag for each. This means the Docker host needs to block access to the
-  database port by other means.
-
 - **User-defined bridges provide automatic DNS resolution between containers**.
 
   Containers on the default bridge network can only access each other by IP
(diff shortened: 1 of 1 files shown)


usha-mandya:
@arkodg Could you PTAL? Thanks.

thaJeztah:
One difference would be that; all containers without a `--network` specified default to be on the `bridge` network, which can be a risk as unrelated stacks/services/containers are then able to communicate.

Using a user-defined network provides a scoped network in which only containers attached to that network are able to communicate

network/bridge.md
```diff
@@ -34,36 +34,16 @@ network.**
...
-  accidentally opening access to the outside world.
-
-  Imagine an application with a web front-end and a database back-end. The
-  outside world needs access to the web front-end (perhaps on port 80), but only
-  the back-end itself needs access to the database host and port. Using a
-  user-defined bridge, only the web port needs to be opened, and the database
-  application doesn't need any ports open, since the web front-end can reach it
-  over the user-defined bridge.
-
-  If you run the same application stack on the default bridge network, you need
```
arkodg: this is the only paragraph that needs to be removed, the one above seems accurate IMHO

Maximization: This is part of a list of differences between user-defined bridges and the default bridge. While the above is true for user-defined bridges, it's _not_ a difference with the default bridge network.

Therefore, I think it's misleading to present it as a difference.

arkodg:
sg, can we also incorporate @thaJeztah's comment

Maximization:
@arkodg Done

PR #10331, opened 2020-02-21
Dockerfile: allow building without archives, and exporting HTML source

thaJeztah:
This PR adds two features to the Dockerfile:

### Dockerfile: allow building docs without archives

This adds a `ENABLE_ARCHIVES` build-arg, which allows building
the documentation without archives.

To build the docs without argives, add `--build-arg ENABLE_ARCHIVES=false` when building the Dockerfile. Preferably, BuildKit should be enabled, as otherwise unused stages would still be _built_ (but not used in the final image)

> Note that currently, the archives drop-down is still added unconditionally (so also included if archives are disabled).

### Dockerfile: add target to output the generated HTML

This patch adds a stage that only contains the generated files. It can be used to export the generated HTML for hosting the documentation on a non-containerised service (e.g. to deploy to an s3 bucket).

When using BuildKit, use the `--output` option to build the files and to copy them to your local filesystem.

For example, to build current docs, including archives:

    DOCKER_BUILDKIT=1 docker build --target=deploy-source --output=./_site .

And to build without archives:

    DOCKER_BUILDKIT=1 docker build --target=deploy-source --build-arg ENABLE_ARCHIVES=false --output=./_site .

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -18,6 +18,8 @@ ARG ENGINE_BRANCH="19.03.x"
 # Distribution
 ARG DISTRIBUTION_BRANCH="release/2.7"
 
+# Set to "false" to build the documentation without archives
+ARG ENABLE_ARCHIVES=true
 
 ###
 # Set up base stages for building and deploying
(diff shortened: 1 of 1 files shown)


thaJeztah:
Relates to https://github.com/docker/docker.github.io/pull/10262 (https://github.com/docker/docker.github.io/pull/10262#discussion_r378482992)

@jdrouet @usha-mandya PTAL

usha-mandya:
Checked out the branch and tried building docs locally; it works as expected. I don't see any archives being built as part of the new process.

@jdrouet Could you take a look?

thaJeztah:
oh; when merging this one, please don't use the `squash and merge` option; it's useful to keep the commits separate as they explain each step/change

PR #10332, opened 2020-02-21
Fix Engine API toc

usha-mandya:
- Fixed the Engine API table of contents
- Removed a duplicate entry for SDK as both **Overview** and **SDKs** entries were pointing to the same page which resulted in both entries being highlighted in the toc.


usha-mandya:
@thaJeztah PTAL

_data/toc.yaml
```diff
@@ -953,14 +953,12 @@ reference:
     path: /engine/reference/commandline/dockerd/
 - sectiontitle: Application Programming Interfaces (APIs)
   section:
-  - sectiontitle: Docker Engine API
+  - sectiontitle: Docker Engine SDK and API
```
thaJeztah: Bit on the fence on this one; API is more likely something people would look for, so _if_ we want to include "sdk", I'd put it last, but perhaps even "keep it as-is" (see below)

```suggestion
  - sectiontitle: Docker Engine API and SDK
```

_data/toc.yaml
```diff
@@ -953,14 +953,12 @@ reference:
...
     section:
-    - path: /engine/api/
+    - path: /develop/sdk/
```
thaJeztah: I prefer `/engine/api/`, and think we should consider splitting the topics into two pages.

Reason I think `/engine/api/` is better, is that the URL navigation is more logical

- https://docs.docker.com/engine/api/ -> General overview of the API, and list of API versions
- https://docs.docker.com/engine/api/v1.40/ -> documentation about a specific API version

With the above, I'd opt for removing the "sdk" sections from the page, and adding a separate page  on which we describe the SDKs. I see there's already a `/engine/api/sdks` below; we can reuse that location for this.

There's also some "enterprise API mismatch" parts mentioned that are either outdated or should be removed (I can help with that)

_data/toc.yaml
usha-mandya: Added SDK first to be consistent with the topic title:
- https://docs.docker.com/develop/sdk/
- https://docs.docker.com/develop/sdk/examples/

_data/toc.yaml
usha-mandya: `/engine/api` doesn't have an overview topic atm. It is currently redirected to `/develop/sdk`. I can create a new index file at  `/engine/api`  with changes to Engine API. 

Also, `/engine/api/sdks` doesn't exist; this is also being redirected to `/develop/sdk`. We can move topics from `/develop/sdks` to `/engine/api/sdk`/.

usha-mandya:
Changed the following based on the discussion with Sebastiaan:

- Moved the topics from `/develop/sdk` to /enging/api/`.
- Added the redirects from `/develop/sdk/`
- Updated the toc
- Updated the docker version, API version, etc in the content

thaJeztah:
@usha-mandya I pushed some commits to this branch; have a look at those changes if those LGTY (easiest to step through commits to see each change in context)

usha-mandya:
Thank you @thaJeztah. The changes look good to me.

PR #10341, opened 2020-02-24
update registry s3 storage driver docs to add skipverify flag

eug48:
### Proposed changes

Document useful flag that was missing from the docs

### DIFF
--- a/registry/storage-drivers/s3.md
+++ b/registry/storage-drivers/s3.md
@@ -109,6 +109,17 @@ Amazon S3 or S3 compatible services for object storage.
   </tr>
   <tr>
     <td>
+      <code>skipverify</code>
+    </td>
+    <td>
+      no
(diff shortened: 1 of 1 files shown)


thaJeztah:
Doing a quick check if this flag is in the current (2.7.1) release; looks like this option was added in https://github.com/docker/distribution/pull/2632, and I see that change is in the v2.7.1 tag at least; https://github.com/docker/distribution/blob/v2.7.1/registry/storage/driver/s3-aws/s3.go#L95 (https://github.com/docker/distribution/blame/2461543d988979529609e8cb6fca9ca190dc48da/registry/storage/driver/s3-aws/s3.go#L95)

usha-mandya:
Thank you @eug48 for updating the documentation :)

Thanks @thaJeztah for verifying the update!

PR #10343, opened 2020-02-24
Engine API: use template for API reference pages

thaJeztah:
This adds a template for the Engine API docs; each version of the engine API now uses a Markdown file (instead of HTML), which is converted to HTML by Jekyll.

With this change, we can more make changes across all API versions if needed


thaJeztah:
@usha-mandya PTAL - to verify this change, the engine API docs should still render as usual

thaJeztah:
Thanks!

PR #10356, opened 2020-02-26
Update django.md

simbirsky:

### DIFF
--- a/compose/django.md
+++ b/compose/django.md
@@ -146,6 +146,7 @@ In this section, you set up the database connection for Django.
                 'ENGINE': 'django.db.backends.postgresql',
                 'NAME': 'postgres',
                 'USER': 'postgres',
+                'PASSWORD': 'postgres',
                 'HOST': 'db',
                 'PORT': 5432,
             }


simbirsky:
> Please also change the section "9" of https://github.com/docker/docker.github.io/blob/master/compose/django.md#define-the-project-components to also mention the password in the `docker-compose.yml`.
> Like:
> 
> ```
> version: '3'
> 
> services:
>   db:
>     image: postgres
>     restart: always
>     environment:
>       POSTGRES_PASSWORD: postgres
>   web:
>     build: .
>     command: python manage.py runserver 0.0.0.0:8000
>     volumes:
>       - .:/code
>     ports:
>       - "8000:8000"
>     depends_on:
>       - db
> ```

What for? It will work without environment block, but password in the settings.py is important parameter without which it will not work.

ulyssessouza:
To make it explicit. That says: 
"We set the password in the container, then we use the same password in the `settings.py`"

yesnik:
@simbirsky Hi Mihail, how do you think can we close this PR? In master branch of this repo password `postgres` is used: https://github.com/docker/docker.github.io/blob/master/compose/django.md

PR #10360, opened 2020-02-26
Update tables from html to markdown format

usha-mandya:
Converted the tables in https://docs.docker.com/registry/storage-drivers/ from HTML to markdown format.


usha-mandya:
@thaJeztah Could you PTAL?

registry/storage-drivers/swift.md
```diff
@@ -10,231 +10,31 @@ storage.
...
-      - POST
-      - DELETE
-    </code>
-    </td>
-</tr>
-</table>
+|  Optional parameter | Description |
+|:--------------|:---------|
+| `tempurlcontainerkey`  |  Specify whether to use container secret key to generate temporary URL when set to true, or the account secret key otherwise. |
+| `tempurlmethods`  |  Array of HTTP methods that are supported by the TempURL middleware of the Swift server. For example: - `tempurlmethods:` `- GET` `- PUT` `- HEAD` `- POST` `- DELETE` |
```
thaJeztah: Hm.. this might've been a reason they picked HTML for this; this example was a multi-line example.

We can "work around" that problem by using the alternative (JSON) array notation for YAML, and just show the _value_ only;

```suggestion
| `tempurlmethods`  |  Array of HTTP methods that are supported by the TempURL middleware of the Swift server. For example: `["GET", "PUT", "HEAD", "POST", "DELETE"]` |
```

registry/storage-drivers/s3.md
```diff
@@ -9,186 +9,29 @@ Amazon S3 or S3 compatible services for object storage.
...
+| `storageclass`  |  no  | The S3 storage class applied to each registry file. The default is `STANDARD`.|
+| `s3accelerate`  |  no  | Specifies whether the registry should use S3 Transfer Acceleration. You must enable acceleration endpoint on a bucket before using this option. A boolean value. The default is `false`.|
 
 > **Note** You can provide empty strings for your access and secret keys to run the driver
 > on an ec2 instance and handles authentication with the instance's credentials. If you
 > use [IAM roles](http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/iam-roles-for-amazon-ec2.html),
 > omit these keys to fetch temporary credentials from IAM.
 
-`region`: The name of the aws region in which you would like to store objects (for example `us-east-1`). For a list of regions, you can look at http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/using-regions-availability-zones.html
+`region`: The name of the aws region in which you would like to store objects (for example `us-east-1`). For a list of regions, see [Regions, Availability Zones, and Local Zones](http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/using-regions-availability-zones.html).
```
thaJeztah: Looks like these descriptions overlap with the ones in the table above.

Don't think it's worth spending time on right now (as it would require checking the difference between both), but something I just noticed

registry/storage-drivers/swift.md
```diff
@@ -10,231 +10,31 @@ storage.
...
-      The endpoint type used when connecting to swift. Possible values are <code>public</code>, <code>internal</code>, and <code>admin</code>. Default is <code>public</code>.
-    </td>
-  </tr>
-</table>
+| Parameter     | Required | Description                                                                                                                                                                                                                                                         |
+|:--------------|:---------|:--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| `authurl`  |  yes  | URL for obtaining an auth token. https://storage.myprovider.com/v2.0 or https://storage.myprovider.com/v3/auth |
+| `username`  |  yes  |	Your Openstack user name. |
+| `password`  |  yes | Your Openstack password. |
+| `region`  | no   |The Openstack region in which your container exists. |
```
thaJeztah: Does this require a space? (haven't tried building this PR as I was waiting for some other builds to complete 😅)

```suggestion
| `region`  | no   | The Openstack region in which your container exists. |
```

(some more like that below)

From a readability perspective, it'd be great if we could align the columns in the Markdown source, but that can be a pain to do (unless your editor allows automatically formatting it)

registry/storage-drivers/s3.md
usha-mandya: Sure, thanks.

registry/storage-drivers/swift.md
usha-mandya: Yes, that needs a space. Thanks for catching that. I'll take a look for other occurrences.

usha-mandya: I> Does this require a space? (haven't tried building this PR as I was waiting for some other builds to complete 😅)

You can check the preview here - https://5e567880b5e20800072d1e32--angry-visvesvaraya-c62c8e.netlify.com/registry/storage-drivers/s3/

PR #10361, opened 2020-02-26
Remove obsolete file sharing information

mat007:
### Proposed changes

I removed obsolete information related to file sharing on Windows from the Docker Desktop docs.
Since we released Docker Desktop Stable 2.2.0.x and Edge 2.2.1.0 the new file sharing implementation does not need credentials or firewall configuration anymore.

### DIFF
--- a/docker-for-windows/index.md
+++ b/docker-for-windows/index.md
@@ -232,12 +232,7 @@ Choose the local directories you'd like to share with your Linux containers. Fil
  not for Windows containers. For Linux containers, you need to share the drive
  where the Dockerfile and volume are located. Otherwise, you get `file not found` or `cannot start service` errors at runtime. See [Volume mounting requires shared drives for Linux containers](troubleshoot.md#volume-mounting-requires-shared-drives-for-linux-containers).
 
-File sharing permissions are tied to the credentials you provide here. If
-you run `docker` commands under a different username than the one configured
-here, your containers cannot access the mounted volumes.
-
-When you select a local drive to share with your containers, Docker Desktop prompts you for your Windows system (domain)
-username and password. Enter your credentials and then click **Apply & Restart**.
+**Apply & Restart** makes the directory available to containers using Docker's bind mount (`-v`) feature.
 
 > Tips on shared drives, permissions, and volume mounts
 >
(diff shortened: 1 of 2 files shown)


docker-for-windows/index.md
```diff
@@ -248,30 +243,11 @@ username and password. Enter your credentials and then click **Apply & Restart**
...
-To share the drive, allow connections between the Windows host machine and the
-virtual machine in Windows Firewall or your third party firewall software. You
-do not need to open port 445 on any other network.
-
-By default, allow connections to `10.0.75.1` on port 445 (the Windows host) from
-`10.0.75.2` (the virtual machine). If your firewall rules seem correct, you may
-need to toggle or
-[reinstall the File and Print sharing service on the Hyper-V virtual network card](http://stackoverflow.com/questions/42203488/settings-to-windows-firewall-to-allow-docker-for-windows-to-share-drive/43904051#43904051)
-
 #### Shared drives on demand
```
thaJeztah: The changes don't affect this? (as in: this feature is still "current"?)

docker-for-windows/index.md
```diff
@@ -248,30 +243,11 @@ username and password. Enter your credentials and then click **Apply & Restart**
 >
  * Docker Desktop sets permissions to read/write/execute for users, groups and others [0777 or a+rwx](http://permissions-calculator.org/decode/0777/).
    This is not configurable. See [Permissions errors on data directories for shared volumes](troubleshoot.md#permissions-errors-on-data-directories-for-shared-volumes).
->
- * Ensure the domain user has access to shared drives, as described in [Verify domain user has permissions for shared drives](troubleshoot.md#verify-domain-user-has-permissions-for-shared-drives-volumes).
```
thaJeztah: Do we know what share / number of users are still on older versions? Thinking if we need to move the content to a separate page, and add a mention "if you're using version XX or older", which either instructs them to upgrade to the latest & greatest, or point to the troubleshooting page for older versions.


/cc @usha-mandya @stephen-turner @nebuk89  for thoughts

docker-for-windows/index.md
mat007: It’s a separate feature, still «current».

docker-for-windows/index.md
usha-mandya: I believe we support our latest and greatest on Docker Desktop. Therefore, not sure if moving obsolete features to another page and duplicating the entire topic is a good option.  Users can browse through the release notes to identify features added/removed from specific versions.

stephen-turner: I agree with Usha. We can't keep documenting all old versions when we change something.

thaJeztah: Okay, thanks for checking. I've seen this error/situation popping up in many situations (some users linger a bit behind "latest & greatest"), so wanted to double check 👍

usha-mandya:
Thank you @thaJeztah.

PR #10363, opened 2020-02-27
fix update docker command

logankimmel:
If you supply the `--interactive` flag to `docker run` the prompt will never come back. That said, if somebody were to run this as a cron job, it would fail as the `ucp-phase2` container name will already exist.


usha-mandya:
@traci-morrison Could you PTAL? Thanks.

traci-morrison:
> @traci-morrison Could you PTAL? Thanks.

Sure, I approved the changes. Thanks.

PR #10370, opened 2020-02-28
Lint markdown

jdrouet:
### Proposed changes

Add a linter step to keep the markdown codebase clean.
We should discuss about if we should change the linter rules or change the actual codebase.
This PR depends on https://github.com/docker/docker.github.io/pull/10369

### DIFF
--- a/.github/workflows/build-pr.yml
+++ b/.github/workflows/build-pr.yml
@@ -14,3 +14,14 @@ jobs:
       - uses: actions/checkout@v2
       - name: build image
         run: docker build --build-arg=ENABLE_ARCHIVES=false --target=current -t documentation:latest .
+
+  lint-markdown:
+    name: linter
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: markdownlint-cli
+        uses: nosborn/github-action-markdown-cli@v1.1.1
+        with:
+          files: .
+          config_file: ".markdownlint.json"
--- a/.github/workflows/build-pr.yml
+++ b/.github/workflows/build-pr.yml
@@ -1,0 +1,16 @@
+name: build docker image when PR is opened
+
+on: pull_request
+
+jobs:
+  build-static-page:
+    name: build
+    runs-on: ubuntu-latest
+    env:
+      DOCKER_BUILDKIT: '1'
+    steps:
+      - name: print docker info
+        run: docker version && docker info
+      - uses: actions/checkout@v2
+      - name: build image
+        run: docker build --build-arg=ENABLE_ARCHIVES=false --target=current -t documentation:latest .


thaJeztah:
Thanks!!

We should change some of these at least, tweak others, and add "TODOs" for those we _want_ but that need some fixing in the existing files.

I didn't see a failure about "line length"; perhaps I looked over it, but we should try to enforce 80-characters (possibly a "margin" that is allowed).


I had a first look over the results, and leaving suggestions/comments below:


```
_includes/ee-linux-install-reuse.md: 76: MD046/code-block-style Code block style [Expected: indented; Actual: fenced]
```

Given that we use (mostly) GitHub-flavored markdown, which uses fences instead of indentation,
we should allow both styles.


```
_includes/ee-linux-install-reuse.md: 228: MD004/ul-style Unordered list style [Expected: dash; Actual: asterisk]
```

This can be useful to use a consistent style (pick either `*` or `-` for bullet-lists), but we'd have to do a "once-over" to update existing pages.

```
_includes/enterprise_label_shortform.md: 4: MD047/single-trailing-newline Files should end with a single newline character
```

We should keep this one and fix those files; "missing newlines at end of files" can be hairy when rebasing/merging/etc things with Git.

```
_includes/github-pr.md: 12: MD034/no-bare-urls Bare URL used [Context: "https://github.com""]
```

Also useful: we need to fix these


```
_samples/library/adminer.md: 16: MD012/no-multiple-blanks Multiple consecutive blank lines [Expected: 1; Actual: 2]
```

Sounds useful to get a consistent style; may need tweaking to our needs, because sometimes multiple lines are useful to more clearly separate sections.

```
app-template/api-reference.md: 123: MD024/no-duplicate-heading/no-duplicate-header Multiple headings with the same content [Context: "### spec"]
437
```

Looks useful: multiple headings with the same content could be confusing for anchor-links

```
app-template/working-with-template.md: 111: MD031/blanks-around-fences Fenced code blocks should be surrounded by blank lines [Context: "```"]
app-template/working-with-template.md: 240: MD010/no-hard-tabs Hard tabs [Column: 1]
app-template/working-with-template.md: 317: MD030/list-marker-space Spaces after list markers [Expected: 2; Actual: 1]
```

These look also useful, and could catch problematic formatting

```
app/working-with-app.md: 13: MD009/no-trailing-spaces Trailing spaces [Expected: 0 or 2; Actual: 1]
```

Also useful for potentially formatting issues

```
app/working-with-app.md: 273: MD031/blanks-around-fences Fenced code blocks should be surrounded by blank lines [Context: "```"]
503
app/working-with-app.md: 273: MD040/fenced-code-language Fenced code blocks should have a language specified [Context: "```"]
```

These are useful: we should always have a language specified, and blank lines before/after these blocks

```
app/working-with-app.md: 294: MD014/commands-show-output Dollar signs used before commands without showing output [Context: "$ docker app render --output d..."]
```

This one is interesting! Needs discussion, but these are the situations where we could probably use the "copy/paste" button we were talking about

```
compose/django.md: 195: MD027/no-multiple-space-blockquote Multiple spaces after blockquote symbol [Context: ">       edit `ALLOWED_HOSTS` i..."]
```

Looks potentially useful

```
datacenter/dtr/2.0/user-management/create-and-manage-teams.md: 44: MD045/no-alt-text Images should have alternate text (alt text)
```

Interesting: would require some work to update these though (so possibly: "disabled", but add "TODO" to re-enable the check)

```
datacenter/dtr/2.1/guides/release-notes.md: 20: MD036/no-emphasis-as-heading/no-emphasis-as-header Emphasis used instead of a heading [Context: "Bugs fixed"]
```

Also potentially useful if we want a consistent style (but to be discussed)

```
datacenter/dtr/2.1/guides/repos-and-images/push-an-image.md: 25: MD006/ul-start-left Consider starting bulleted lists at the beginning of the line [Expected: 0; Actual: 2]
```

Useful: we should be consistent on these


```
_includes/content/ssh/ssh-find-keys.md: 81: MD033/no-inline-html Inline HTML [Element: hr]
```

While we should _avoid_ extensive use of HTML, I think it's ok to allow it, as there
are situations where HTML is "cleaner" than trying to do it with Markdown.

These are ones to discuss / look at:

This might be a false positive; we need to check (source code _looks_ ok at a glance). Could be useful though if numbered-lists are accidentally broken up.

```
_includes/ee-linux-install-reuse.md: 89: MD029/ol-prefix Ordered list item prefix [Expected: 4; Actual: 5; Style: 1/2/3]
```


```
_includes/content/ssh/ssh-find-keys.md: 12: MD041/first-line-heading/first-line-h1 First line in file should be a top level heading [Context: "1.  Open a command-line termin..."]
```

At some point, H1 headings on pages were removed from Markdown, in favor of the `title:` metadata field, which would then generate a H1 header based on that.

I _think_ the main reason for that was to prevent situations where the first H1 was updated, but the `title:` metadata was not updated accordingly. I agree that _could_ sometimes be an issue. On the other hand, sometimes you want the HTML `<title>` to be slightly different than the first H1?

PR #10400, opened 2020-03-04
ci: copy static files to s3 on master build

jdrouet:
### Proposed changes

Deploy generated static files to s3

### DIFF
--- a/.github/workflows/build-master.yml
+++ b/.github/workflows/build-master.yml
@@ -1,0 +1,21 @@
+on:
+  push:
+    branches:
+      - master
+
+jobs:
+  build-static-page:
+    name: build
+    runs-on: ubuntu-latest
+    env:
+      DOCKER_BUILDKIT: '1'
+    steps:
+      - name: print docker info
+        run: docker version && docker info
+      - uses: actions/checkout@v2
+      - name: build image
+        run: docker build --build-arg=ENABLE_ARCHIVES=true --target=current -t documentation:latest .
+      - name: copy static files locally
+        run: docker run -v ${PWD}:/output documentation:latest cp -r /usr/share/nginx/html /output/_site
+      - name: upload files to S3 bucket
+        run: aws s3 sync _site s3://docs.docker.com-test-us-east-1/ --delete
--- a/.github/workflows/build-master.yml
+++ b/.github/workflows/build-master.yml
@@ -9,13 +9,15 @@ jobs:
     runs-on: ubuntu-latest
     env:
       DOCKER_BUILDKIT: '1'
+      AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
+      AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
     steps:
       - name: print docker info
         run: docker version && docker info
       - uses: actions/checkout@v2
-      - name: build image
-        run: docker build --build-arg=ENABLE_ARCHIVES=true --target=current -t documentation:latest .
-      - name: copy static files locally
-        run: docker run -v ${PWD}:/output documentation:latest cp -r /usr/share/nginx/html /output/_site
+      - name: build current docs
+        run: docker build --build-arg=ENABLE_ARCHIVES=true --target=deploy-source --output=./_site .
       - name: upload files to S3 bucket
-        run: aws s3 sync _site s3://docs.docker.com-test-us-east-1/ --delete
+        run: aws s3 sync --acl public-read _site s3://docs.docker.com-test-us-east-1/ --delete
+      - name: invalidate docs website cache
+        run: aws --region us-east-1 lambda invoke --function-name arn:aws:lambda:us-east-1:710015040892:function:docs-test-cache-invalidator response.json


.github/workflows/build-master.yml
```diff
@@ -0,0 +1,21 @@
...
+    steps:
+      - name: print docker info
+        run: docker version && docker info
+      - uses: actions/checkout@v2
+      - name: build image
+        run: docker build --build-arg=ENABLE_ARCHIVES=true --target=current -t documentation:latest .
+      - name: copy static files locally
+        run: docker run -v ${PWD}:/output documentation:latest cp -r /usr/share/nginx/html /output/_site
+      - name: upload files to S3 bucket
+        run: aws s3 sync _site s3://docs.docker.com-test-us-east-1/ --delete
```
StefanScherer: For the `aws` call we need two environment variables from the secrets
```suggestion
        run: aws s3 sync _site s3://docs.docker.com-test-us-east-1/ --delete
        env:
          AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
          AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
```

.github/workflows/build-master.yml
```diff
@@ -0,0 +1,21 @@
...
+    name: build
+    runs-on: ubuntu-latest
+    env:
+      DOCKER_BUILDKIT: '1'
+    steps:
+      - name: print docker info
+        run: docker version && docker info
+      - uses: actions/checkout@v2
+      - name: build image
+        run: docker build --build-arg=ENABLE_ARCHIVES=true --target=current -t documentation:latest .
```
thaJeztah: This should probably be `--target=deploy-source` ?

.github/workflows/build-master.yml
```diff
@@ -0,0 +1,21 @@
...
+      - name: copy static files locally
+        run: docker run -v ${PWD}:/output documentation:latest cp -r /usr/share/nginx/html /output/_site
```
thaJeztah: If we're not able to use the `docker build --output` option, possibly;

```
docker create --name tmp documentation:latest
docker cp tmp:/ /output/_site
```

We could change the `deploy-src` target to put everything in a `/_site` subdirectory if that's more convenient; https://github.com/docker/docker.github.io/blob/20267e58b96b1d41689e01eacd61172112017aa9/Dockerfile#L89

StefanScherer: Oh I just saw that we have all these stages in the Dockerfile. Maybe then it's easier to create a target stage with the AWS CLI installed (`FROM dockereng/awscli:1.16.156`), then we do not have to copy the files multiple times. Maybe saves some valuable GH actions build time.

thaJeztah: That image itself is private, so perhaps just repeat the steps in a stage, as all it does is an install of the cli;

```
RUN apk -Uuv add groff less python py-pip \
 && pip install awscli==${VERSION} \
 && apk --purge -v del py-pip
```

Tricky bit is that we'd have to pass the credentials to `docker build` as a build-arg, which is potentially dangerous, so if the original idea to use `docker build --output` works, that would be a cleaner option

thaJeztah: oh, actually, the image itself isn't private, but still likely that we could just install it inline in the dockerfile 🤔  (if the `--output` variant doesn't work out)

.github/workflows/build-master.yml
```diff
@@ -0,0 +1,23 @@
+on:
+  push:
+    branches:
+      - master
+      # gh-mster-3 must be removed before merging this PR. It was added to test the pipeline.
```
StefanScherer: I have added a temporary commit to run the pipeline directly in this PR branch to have a better feedback loop.

.github/workflows/build-master.yml
StefanScherer: Good idea. I've updated the `docker build` as shown in the `Dockerfile` .

.github/workflows/build-master.yml
StefanScherer: Good idea. I've updated the `docker build` as shown in the `Dockerfile` and use the `--output` parameter.

StefanScherer:
OK, we have a green build https://github.com/docker/docker.github.io/pull/10400/checks?check_run_id=500109088
It's ready for review.
Docs updated at https://docs-test.docker.com/
I'll remove my temporary commit 35bfe9465834961d068906bc72c9df3c0b5a4ca5 before we merge this.

usha-mandya:
@StefanScherer The https://docs-test.docker.com/ serves only the landing page. Clicking on any other tabs/pages throws `Access Denied` error. Is this expected?

.github/workflows/build-master.yml
```diff
@@ -0,0 +1,27 @@
...
+    env:
+      DOCKER_BUILDKIT: '1'
+      AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
+      AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+    steps:
+      - name: print docker info
+        run: docker version && docker info
+      - uses: actions/checkout@v2
+      - name: build current docs
+        run: docker build --build-arg=ENABLE_ARCHIVES=true --target=deploy-source --output=./_site .
```
thaJeztah: the `--output` option is working now? ❤️ 🎉

.github/workflows/build-master.yml
```diff
@@ -0,0 +1,27 @@
...
+      - name: list _site
+        run: ls -l _site
+      - name: upload files to S3 bucket
+        run: aws s3 sync --acl public-read _site s3://docs.docker.com-test-us-east-1/ --delete
```
thaJeztah: I'm wondering how well it picks up that the "archives" never change between builds; is that something we need to optimise? (e.g. resetting date/time if it compares that metadata)

StefanScherer: I ran this command twice. For the first run the S3 bucket was completely empty, and afterwards no index.html or any other file at top level were uploaded. In sub directories I could find index.html files. 🤔 

I ran it again with the ls -l _site commit to "debug" the GH action pipeline, now all files seems to be uploaded.
I wonder how predictable the aws s3 sync command is. @tiborvass any insights? :-)

StefanScherer: Alright, I found my problem why index.html was missing. I did more runs here and the upload/sync looks good now. Pages can be viewed after we fixed a configuration in AWS.
The version selection also works to see older docs.

.github/workflows/build-master.yml
StefanScherer: Removed the commit for final review.

.github/workflows/build-master.yml
StefanScherer: It could be that all files get synced due to new timestamps from the newly built container image for each pipeline run.
From a costs perspective it's inbound traffic from the internet and that's free.

StefanScherer:
The latest build of this pipeline is green https://github.com/docker/docker.github.io/runs/503458050?check_suite_focus=true, after that I've removed the workaround to run this pipeline directly in this PR. It now only will be triggered from master branch.

The current red status is because netlify responds with 502, that's unrelated to this PR.

@usha-mandya @thaJeztah PTAL

StefanScherer:
Thanks. Let‘s get this in to evaluate more when it‘s active for master branch.

thaJeztah:
@StefanScherer can you change your "changes requested" to "approved" ?

Screenshot 2020-03-12 at 21 01 33

StefanScherer:
Thanks @thaJeztah, didn‘t see that I was blocking it. 😅

PR #10401, opened 2020-03-04
[WIP] create check when deployed on netlify

jdrouet:
## Proposed changes

Create a check with a direct link to the netlify site


usha-mandya:
@jdrouet I guess we don't need this PR anymore. Could you PTAL and close it if it's not required? Thank you.

PR #10418, opened 2020-03-07
Git instructions fix. Remove '-app' from path

dougmcintyre:
### Proposed changes

Path changed for the cd command that follows the git clone step.


dougmcintyre:
Disregard. 
I didn't notice the second directory.

PR #10429, opened 2020-03-10
Fix wrong device path in NFSv4 volume example

jureaky:
In section Create a service which creates an NFS volume,
NFSv3 command states right device path(:/var/docker-nfs),
but NFSv4 command is not configured to refer that path.
So updated the path.


usha-mandya:
@arkodg Could you PTAL?

PR #10433, opened 2020-03-10
docker-compose documentation version

hochan222:
hochan@hochan-node:~/Desktop/pfc/dockercompose$ sudo docker-compose up --build
ERROR: Version in "./docker-compose.yml" is unsupported. You might be seeing this error because you're using the wrong Compose file version. Either specify a version of "2" (or "2.0") and place your service definitions under the `services` key, or omit the `version` key and place your service definitions at the root of the file to use version 1.

### Proposed changes

Due to the evolution of version, '3' no longer matches version in docker-compose.

### DIFF
--- a/compose/index.md
+++ b/compose/index.md
@@ -30,7 +30,7 @@ so they can be run together in an isolated environment.
 
 A `docker-compose.yml` looks like this:
 
-    version: '3'
+    version: '2.0'
     services:
       web:
         build: .


thaJeztah:
@ulyssessouza I don't understand this change; I tried this example with version 3, and that worked without issues; was this problem occurring on an outdated version of docker compose?

hochan222:
> @ulyssessouza I don't understand this change; I tried this example with version 3, and that worked without issues; was this problem occurring on an outdated version of docker compose?

My version is,
docker-compose version 1.8.0, build unknown

According to the documentation in the current version, you need to specify the version as x.x.
https://docs.docker.com/compose/compose-file/compose-versioning/

**Versioning**
There are currently three versions of the Compose file format:

Version 1, the legacy format. This is specified by omitting a version key at the root of the YAML.

Version 2.x. This is specified with a version: '2' or version: '2.1', etc., entry at the root of the YAML.

Version 3.x, the latest and recommended version, designed to be cross-compatible between Compose and the Docker Engine’s swarm mode. This is specified with a version: '3' or version: '3.1', etc., entry at the root of the YAML.

**Note**: When specifying the Compose file version to use, make sure to specify both the major and minor numbers. If no minor version is given, 0 is used by default and not the latest minor version. As a result, features added in later versions will not be supported. For example:

version: "2"
is equivalent to:

version: "2.0"


So I fixed it because I thought it was correct to add .0 to the back.

thaJeztah:
> My version is,
> docker-compose version 1.8.0, build unknown

Wow, compose 1.8.0 is _really_ old; that version was released four years ago; https://github.com/docker/compose/releases/tag/1.8.0. I'd definitely recommend installing a more current version: https://docs.docker.com/compose/install/

> Note: When specifying the Compose file version to use, make sure to specify both the major and > minor numbers. If no minor version is given, 0 is used by default and not the latest minor version. 

That's correct (although there's a proposal to change the behaviour: https://github.com/docker/compose/issues/7201)

While there are some features that are not supported by the v3.x format, most features are supported, and I'd prefer the examples to use the v3 format where possible, because it's currently the most portable format (supported both by `docker-compose` (standalone), and `docker stack deploy` (to deploy to kubernetes or swarm services).

I think we could change these to use;

```
version: "{{ site.compose_file_v3 }}"
```

So that the latest supported v3 compose file format is filled in automatically.


@ulyssessouza @usha-mandya wdyt?

ulyssessouza:
Actually the `link` made me accept thinking it was just a V2 feature. 

But checking the docs, looks like it's supported in V3 too.
If that's correct, I think we should revert that change.

PR #10442, opened 2020-03-13
ci: Upload docs only in upstream repo

StefanScherer:
Improve the GH actions pipeline to only run the upload and invalidate cache steps in upstream repo.
Otherwise contributors get an error mail from GitHub when they update their `master` branch.

Example:
Screen Shot 2020-03-13 at 10 54 51 AM

I first thought what's wrong with the pipeline in upstream, but then I saw that the full workflow ran in my fork I updated my master branch:

```
git fetch upstream
git rebase upstream/master
git push
```

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -2,7 +2,10 @@
 .git
 .gitignore
 tests
-Dockerfile
-docker-compose.yml
 _site
+CONTRIBUTING.md
+Dockerfile
+Dockerfile.archive
+docker-compose.yml
 Gemfile.lock
+Jenkinsfile
--- a/.github/workflows/build-master.yml
+++ b/.github/workflows/build-master.yml
@@ -7,10 +7,9 @@ jobs:
   build-static-page:
     name: build
     runs-on: ubuntu-latest
+    timeout-minutes: 10
     env:
       DOCKER_BUILDKIT: '1'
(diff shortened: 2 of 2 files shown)


StefanScherer:
For reference, I got the idea from the GitHub Actions Advent Calendar:  https://www.edwardthomson.com/blog/github_actions_13_conditionals.html

StefanScherer:
Thanks, I‘ll update and add a timeout 😅💰

StefanScherer:
I've added a timeout.
And I also found out that we deploy a few files from the repo in the final website. I guess we don't want a downloadable https://docs.docker.com/Jenkinsfile 😅

.dockerignore
```diff
@@ -2,7 +2,12 @@
...
+_site
+CONTRIBUTING.md
 Dockerfile
+Dockerfile.archive
 docker-compose.yml
-_site
 Gemfile.lock
+Jenkinsfile
+LICENSE
+README.md
```
thaJeztah: These may actually be needed, as they contain the copyright / licensing info (which should generally be distributed with each copy)

StefanScherer: Good point. I also thought about that. That's what code reviews are good for.
I've updated the commit and keep these two files in the `_site` folder.

PR #10449, opened 2020-03-14
docker-compose documentation version

hochan222:
hochan@hochan-node:~/Desktop/pfc/dockercompose$ sudo docker-compose up --build
ERROR: Version in "./docker-compose.yml" is unsupported. You might be seeing this error because you're using the wrong Compose file version. Either specify a version of "2" (or "2.0") and place your service definitions under the services key, or omit the version key and place your service definitions at the root of the file to use version 1.

### Proposed changes

Due to the evolution of version, '3' no longer matches version in docker-compose.

### Related issues (optional)
docker-compose documentation version #10451
docker-compose documentation version #10450
docker-compose documentation version #10449

docker-compose documentation version #10433 [Merged]

### DIFF
--- a/compose/gettingstarted.md
+++ b/compose/gettingstarted.md
@@ -112,7 +112,7 @@ and the [Dockerfile reference](/engine/reference/builder.md).
 Create a file called `docker-compose.yml` in your project directory and paste
 the following:
 
-    version: '3'
+    version: '2.0'
     services:
       web:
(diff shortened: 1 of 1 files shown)


thaJeztah:
What version of Docker Compose (`docker-compose --version`) are you running? I tried the example, and don't get an error

PR #10455, opened 2020-03-15
Improve Desktop Windows filesharing docs

djs55:
### Proposed changes

In Docker Desktop 2.2.0.0 stable the filesharing implementation was changed from Samba to a FUSE-based system much more similar to what is used on the Mac. This means that inotify now works.

In Docker Desktop 2.2.2.0 edge the filesharing implementation Windows was changed from per-drive to per-folder.

This was released in Docker Desktop stable 2.3.0.x too.

This PR proposes to update the docs to cover both these recent changes. Let me know if you'd like the changes broken up into stable changes and edge changes.


docker-for-mac/index.md
```diff
@@ -147,7 +147,11 @@ You can also move the disk image to a different location. If you attempt to move
 
 #### File sharing
 
-Choose the local directories you'd like to share with your containers. File sharing is required for volume mounting if the project lives outside of the `/Users` directory. In that case, share the drive where the Dockerfile and volume are located. Otherwise, you get `file not found` or `cannot start service` errors at runtime`.
+Use File sharing to allow local directories on the Mac to be shared with Linux containers.
+This is especially useful for
+editing source code in an IDE on the host while running and testing the code in a container.
+By default the `/Users` directory is shared. If your project is outside this directory then it must be added
```
mat007: I think we share a few more by defaults, like the folders which have temp directories?

docker-for-mac/index.md
```diff
@@ -156,10 +160,16 @@ File share settings are:
 - **Apply & Restart** makes the directory available to containers using Docker's
   bind mount (`-v`) feature.
 
-  There are some limitations on the directories that can be shared:
+The following paths will always be shared from Linux, rather than the Mac host:
```
mat007: I don’t understand what this means.

docker-for-windows/troubleshoot.md
```diff
@@ -122,36 +122,21 @@ deployment
...
 containers). Click ![whale menu](images/whale-x.png){: .inline}
- and then **Settings** > **Shared Drives** and share the drive that contains the
+ and then **Settings** > **Shared Folders** and share the folder that contains the
 Dockerfile and volume.
 
 #### Understand symlinks limitations
 
-Symlinks work within and across containers. However, symlinks created outside of
-containers (for example, on the host) do not work. To learn more, see [Are
+Symlinks work within and across containers. To learn more, see [Are
```
mat007: So, they are no limitations now, are they? Maybe we should change the section title?

docker-for-mac/index.md
djs55: I'm in two minds about whether to try to include (and explain) this. I'm trying to say that `docker run -v /dev:/dev` will share the `/dev` in the helper VM, not on the host. Some things like `/dev/zero` will work but obviously `/dev/sda` might not, depending on what your expectations were. Maybe I'll remove it.

thaJeztah:
ping @djs55 looks like this needs a rebase

docker-for-mac/index.md
djs55: This PR no longer includes the Mac stuff.

PR #10458, opened 2020-03-15
Dockerfile: update to github-pages 198, and some other fixes/improvements

thaJeztah:


Dockerfile
```diff
@@ -24,9 +24,7 @@ ARG ENABLE_ARCHIVES=true
 ###
 # Set up base stages for building and deploying
 ###
-
-# Get basic configs and Jekyll env
-FROM docs/docker.github.io:docs-builder AS builderbase
+FROM starefossen/github-pages:198 AS builderbase
```
thaJeztah: This (at least for now) cuts out our custom image, which was based on `starefossen/github-pages`, with some other changes that we currently don't use)

thaJeztah:

Dockerfile
StefanScherer: Good idea to remove complexity!

thaJeztah: The idea of the custom image was that we could update or do fixes in that image, and rebuild the archived docs, without having to update each of those branches.

If we stop publishing those archives that may no longer be relevant

usha-mandya:
Thank you @thaJeztah.

PR #10485, opened 2020-03-18
Deploy docs-stage to S3 instead of swarm

StefanScherer:
This PR changes the pipeline how to deploy master branch to docs-stage.docker.com
It removes the stage in Jenkins pipeline and activates the S3 upload in the GitHub action workflow.

PR #10492, opened 2020-03-20
Upload published website to S3

StefanScherer:
Similar to #10485, enable the S3 website GH Actions workflow for the `published` branch.

Information for reviewers:
- DNS at this point still is configured to the nginx service, but we already can test the upload and cache invalidation.
- We should merge this PR before switching DNS to have the same content both in the nginx service and in S3.
- We should merge this to `published` branch before switching DNS to have the correct content in the S3 bucket.

### DIFF
--- a/.github/workflows/build-published.yml
+++ b/.github/workflows/build-published.yml
@@ -1,0 +1,30 @@
+on:
+  push:
+    branches:
+      - published
+
+jobs:
+  build-static-page:
+    name: build
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+    env:
+      DOCKER_BUILDKIT: '1'
+    steps:
+      - name: print docker info
+        run: docker version && docker info
+      - uses: actions/checkout@v2
+      - name: build current docs
+        run: docker build --build-arg=ENABLE_ARCHIVES=true --target=deploy-source --output=./_site .
+      - name: upload files to S3 bucket
+        run: aws s3 sync --acl public-read _site s3://docs.docker.com-us-east-1/ --delete
+        env:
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -15,36 +15,6 @@ pipeline {
...
-                export DOCKER_HOST=$DOCKER_HOST_STRING
-                docker service update --detach=false --force --image docs/docker.github.io:prod-${env.BUILD_NUMBER} docs-docker-com_docs --with-registry-auth
-                curl -X POST -H 'Content-type: application/json' --data '{"text":"Successfully published docs. https://docs.docker.com/"}' $SLACK
-              """
-            }
-          }
-        }
-      }
-    }
     stage( 'docs-private' ) {
```
StefanScherer: The remaining Jenkinsfile after this PR only contains stages for `docs-private`. 
Should I keep it or just completely remove the Jenkinsfile in this PR?

.github/workflows/build-published.yml
```diff
@@ -0,0 +1,32 @@
+on:
+  push:
+    branches:
+      - published
+      # TODO remove the following line before merging this PR
```
thaJeztah: Looks like this needs to be removed before merging?

StefanScherer: Correct, this was to run the workflow once in this PR to test it. I've updated the PR.

StefanScherer:
Removed the TODO, I had to wait for Netlify to finish to avoid even more commit running in parallel.

StefanScherer:
netlify-deploy red again, but it's unrelated.

PR #10494, opened 2020-03-20
Moved the DDE TOC entry

usha-mandya:
Moved Docker Desktop Enterprise to the bottom of the TOC to reduce visibility.


usha-mandya:
@thaJeztah Could you PTAL?

PR #10502, opened 2020-03-21
Update stack-deploy.md

frankdean:
Testing the local registry with the `curl` command hung.

The `curl` test is successful after adding `mode=host` for the registry publish command.

### DIFF
--- a/engine/swarm/stack-deploy.md
+++ b/engine/swarm/stack-deploy.md
@@ -40,7 +40,7 @@ a throwaway registry, which you can discard afterward.
 1.  Start the registry as a service on your swarm:
 
     ```bash
-    $ docker service create --name registry --publish published=5000,target=5000 registry:2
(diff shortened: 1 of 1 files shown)


thaJeztah:
Could you provide more info about the environment in which this failed? I don't think `mode=host` should be needed for this to work

frankdean:
The environment was Debian 10 (Buster).  I've done some more digging and `mode=hosts` isn't the right solution, as you say.

The issue seems to be IPv6 related.  Adding the `--ipv4` parameter to the `curl` command gives the correct response.  Alternatively, using `127.0.0.1` instead of `localhost` works too.  I notice elsewhere in the tutorial, `127.0.0.1` is used and indeed I couldn't push the images to the registry using `localhost` but could with `127.0.0.1`.

Perhaps the better solution is to replace all uses of `localhost` with `127.0.0.1` within the tutorial.

PR #10508, opened 2020-03-23
Use original short forms of Docker commands

justincormack:
The long forms of Docker commands have not caught on. I don't think we
should continue to encourage their use especially in the get started
guide. There is no evidence from online searches that anyone is really
using them. They are overly pedantic and verbose.

Signed-off-by: Justin Cormack <justin.cormack@docker.com>

### DIFF
--- a/get-started/part2.md
+++ b/get-started/part2.md
@@ -141,7 +141,7 @@ Now that you have some source code and a Dockerfile, it's time to build your fir
 Make sure you're in the directory `node-bulletin-board/bulletin-board-app` in a terminal or PowerShell using the `cd` command. Let's build your bulletin board image:
 
 ```script
-docker image build -t bulletinboard:1.0 .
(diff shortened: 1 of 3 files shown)


get-started/part2.md
```diff
@@ -141,7 +141,7 @@ Now that you have some source code and a Dockerfile, it's time to build your fir
 Make sure you're in the directory `node-bulletin-board/bulletin-board-app` in a terminal or PowerShell using the `cd` command. Let's build your bulletin board image:
 
 ```script
-docker image build -t bulletinboard:1.0 .
+docker build -t bulletinboard:1.0 .
```
chris-crone: Do we want to make this `--tag` instead of `-t`?

justincormack: yeah probably a good idea, we mostly use the long form flags as they are less confusing

Issue #6207, opened 2018-03-12
To doc: Add support for templated secrets and configs


gbarr01:
References: 
https://github.com/docker/docker-ce/blob/v18.03.0-ce-rc3/CHANGELOG.md
https://github.com/docker/cli/pull/896
https://github.com/moby/moby/pull/33702
https://github.com/moby/moby/pull/36366
https://github.com/docker/swarmkit/pull/2133

Possible doc locations: 
https://docs.docker.com/engine/swarm/secrets/

thaJeztah:
Wrote down a basic example here; https://github.com/moby/moby/issues/37377#issuecomment-401870956

thaJeztah:
@ahh-docker PTAL

L-Hudson:
Closing this ticket due to its age, and the impending refactor. If you think this is in error, feel free to reopen. Thanks!

jean-airoldie:
This [blog](https://blog.sunekeller.dk/2018/04/docker-18-03-config-and-secret-templating/) offers decent documentation in the meantime.

matthanley:
> Closing this ticket due to its age, and the impending refactor. If you think this is in error, feel free to reopen. Thanks!

Does that mean usage of this feature will be changing? Any details of the referenced _refactor_?

ev3rl0ng:
Hi, long time listener, first time caller. Sorry to revive an old thread, but I am trying to find more details around this very topic.

I was hoping to find out what the expected scope of the template would be in the case of a service or a stack deployment.

For example, could I:

- reference all replicas in a service with an array in a range style template from the golang Templates library?
- reference properties of another service in the same stack?

I understand that other platforms (terraform/ansible/puppet) could most likely provide what I'm looking for here, but there's a unique position that the config level templating provides here that those platforms would end up using or accomplishing via other means.

If I can get some initial pointers on where to start (new to this code base), I'd be more than happy to dig into it and generate the documentation around it if it is indeed possible.

Thanks for your time!

dmitriy-shleht:
Hi, view templates `{{secret "foo"}}` work perfectly in normal configuration files, but if the configuration in the JSON format, it becomes an unhappy due to extra quotes. What to do in this case where to make a new issues?
If I scraps shield, then my secret does not work (`{{secret \"user_setup_pass\"}}`)
```json
{
  "ConnectionStrings": {
    "Oracle": "uid=CA;pwd={{secret "user_setup_pass"}};direct=true;"
  }
}
```

[...]

PR #10512, opened 2020-03-24
Add example for templated config

pszczekutowicz:
### Proposed changes

There is no example how to use templated configs so I added one.

### DIFF
--- a/engine/swarm/configs.md
+++ b/engine/swarm/configs.md
@@ -278,6 +278,56 @@ This example assumes that you have PowerShell installed.
     docker config rm homepage
     ```
 
+### Example: Use a templated config
+
+To create a configuration in which the content will be generated using a
+template engine, use the `--template-driver` parameter and specify the engine
+name as its argument. The template will be rendered when container is created.
+
+1.  Save the following into a new file `index.html.tmpl`.
+
+    ```html
+    <html>
+      <head><title>Hello Docker</title></head>
+      <body>
+        <p>Hello {% raw %}{{ env "HELLO" }}{% endraw %}! I'm service {% raw %}{{ .Service.Name }}{% endraw %}.</p>
+      </body>
+    </html>
+    ```
+
+2.  Save the `index.html.tmpl` file as a swarm config named `homepage`. Provide
+    parameter `--template-driver` and specify `golang` as template engine.
+
+    ```bash
+    $ docker config create --template-driver golang homepage index.html.tmpl
+    ```
+
+3.  Create a service that runs Nginx and has access to the environment variable
+    HELLO and to the config.
+
+    ```bash
+    $ docker service create \
+         --name hello-template \
+         --env HELLO="Docker" \
+         --config source=homepage,target=/usr/share/nginx/html/index.html \
+         --publish published=3000,target=80 \
+         nginx:alpine
+    ```
+
+4.  Verify that the service is operational: you can reach the Nginx server, and
+    that the correct output is being served.
+
+    ```bash
+    $ curl http://0.0.0.0:3000
+
+    <html>
+      <head><title>Hello Docker</title></head>
+      <body>
+        <p>Hello Docker! I'm service hello-template.</p>
+      </body>
+    </html>
+    ```
+
 ### Advanced example: Use configs with a Nginx service
 
 This example is divided into two parts.


usha-mandya:
@thaJeztah Could you PTAL? Thanks.

engine/swarm/configs.md
```diff
@@ -612,3 +612,53 @@ configuration file.
...
+template engine, use the `--template-driver` parameter and specify the engine
+name as its argument. The template will be rendered when container is created.
+
+1.  Save the following into a new file `index.html`.
+
+    ```html
+    <html>
+      <head><title>Hello Docker</title></head>
+      <body>
+        <p>{% raw %}{{ env "HELLO" }}{% endraw %}! I'm service {% raw %}{{ .Service.Name }}{% endraw %}.</p>
```
thaJeztah, line 628: nit: perhaps we should hard-code "Hello"

```suggestion
        <p>Hello {% raw %}{{ env "HELLO" }}{% endraw %}! I'm service {% raw %}{{ .Service.Name }}{% endraw %}.</p>
```

engine/swarm/configs.md
```diff
@@ -612,3 +612,53 @@ configuration file.
...
+
+To create a configuration in which the content will be generated using a
+template engine, use the `--template-driver` parameter and specify the engine
+name as its argument. The template will be rendered when container is created.
+
+1.  Save the following into a new file `index.html`.
+
+    ```html
+    <html>
+      <head><title>Hello Docker</title></head>
```
thaJeztah, line 626: Do we want to template this as well?

engine/swarm/configs.md
```diff
@@ -612,3 +612,53 @@ configuration file.
...
+    ```bash
+    $ docker config create --template-driver golang homepage index.html
+    ```
+
+3.  Create a service that runs Nginx and has access to the environment variable
+    HELLO and to the config.
+
+    ```bash
+    $ docker service create \
+         --name nginx \
```
thaJeztah, line 645: Perhaps change this to `hellodocker` or (e.g.) `hello-template` to  make it more recognisable and to prevent it conflicting with other examples

```suggestion
         --name hello-template \
```

engine/swarm/configs.md
```diff
@@ -612,3 +612,53 @@ configuration file.
...
+         --env HELLO="Hello Docker" \
+         --config source=homepage,target=/usr/share/nginx/html/index.html \
+         --publish published=3000,target=80 \
+         nginx:latest
```
thaJeztah, line 649: I prefer using the `nginx:alpine` variant in examples, as it's more lightweight, and most (if not all) examples we use don't need the debian-based variant.

```suggestion
         nginx:alpine
```

engine/swarm/configs.md
```diff
@@ -612,3 +612,53 @@ configuration file.
 
 You have now updated your `nginx` service's configuration without the need to
 rebuild its image.
+
+### Example: Use templated config in Nginx service
```
thaJeztah, line 616: Perhaps we should;

- keep it short, and remove "in Nginx service" from the title (there's already some Nginx examples on the page, which could make it harder to distinguish them)
- I think we should move this example before the advanced example ("Advanced example: Use configs with a Nginx service")

```suggestion
### Example: Use a templated config
```

engine/swarm/configs.md
```diff
@@ -612,3 +612,53 @@ configuration file.
...
+
+To create a configuration in which the content will be generated using a
+template engine, use the `--template-driver` parameter and specify the engine
+name as its argument. The template will be rendered when container is created.
```
thaJeztah, line 620: Something for a follow-up; we should add a list of placeholders that can be used in the reference docs. The `service create` page has a list: https://docs.docker.com/engine/reference/commandline/service_create/#create-services-using-templates, but templated configs can also reference secrets (see https://github.com/moby/moby/pull/33702), and I don't think that's currently documented

engine/swarm/configs.md
```diff
@@ -612,3 +612,53 @@ configuration file.
...
+
+1.  Save the following into a new file `index.html`.
```
thaJeztah, line 622: Thinking out loud; would it be good to name this file `index.html.tmpl` so that examples below can illustrate that the filename;

- doesn't have to match the name of the config (already illustrated: the config is named `homepage`)
- doesn't have to match the name used inside the container/service

engine/swarm/configs.md
```diff
@@ -612,3 +612,53 @@ configuration file.
...
+         --config source=homepage,target=/usr/share/nginx/html/index.html \
+         --publish published=3000,target=80 \
+         nginx:latest
+    ```
+
+4.  Verify that the service is operational: you can reach the Nginx server, and
+    that the correct output is being served.
+
+    ```bash
+    $ curl https://0.0.0.0:3000
```
thaJeztah, line 656: This should be `http://` in this example (as we don't setup TLS)

engine/swarm/configs.md
```diff
@@ -612,3 +612,53 @@ configuration file.
...
+
+    <html>
+      <head><title>Hello Docker</title></head>
+      <body>
+        <p>Hello Docker! I'm service nginx.</p>
```
thaJeztah, line 661: If we change the name of the service; `s/nginx/hello-template/`

thaJeztah:
Partially addresses https://github.com/docker/docker.github.io/issues/6207

engine/swarm/configs.md
pszczekutowicz: Should it be done as a separate environmental variable `TITLE` or reuse `HELLO`?

engine/swarm/configs.md
pszczekutowicz: Yes, it's not documented

engine/swarm/configs.md
thaJeztah: I think we can reuse `HELLO` to keep it simple, but either way is fine; wdyt?

pszczekutowicz:
@thaJeztah I think I implemented all your suggestions and I posted one question.

PR #10513, opened 2020-03-24
Remove `restart` and `stop` from command list.

Zayah117:
UCP commands `restart` and `stop` have been deprecated and are no longer available in UCP 3.2.

https://docs.docker.com/ee/ucp/release-notes/#deprecations


thaJeztah:
@Dawn-Wood PTAL

PR #10525, opened 2020-03-27
Update insecure.md

olimungo:
Couldn't make it work with the port in the path. I guess it's a typo.

### DIFF
--- a/registry/insecure.md
+++ b/registry/insecure.md
@@ -77,7 +77,7 @@ This is more secure than the insecure registry solution.
     depends on your OS.
 
     - **Linux**: Copy the `domain.crt` file to
-      `/etc/docker/certs.d/myregistrydomain.com:5000/ca.crt` on every Docker
+      `/etc/docker/certs.d/myregistrydomain.com/ca.crt` on every Docker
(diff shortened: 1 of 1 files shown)


registry/insecure.md
```diff
@@ -77,7 +77,7 @@ This is more secure than the insecure registry solution.
     depends on your OS.
 
     - **Linux**: Copy the `domain.crt` file to
-      `/etc/docker/certs.d/myregistrydomain.com:5000/ca.crt` on every Docker
+      `/etc/docker/certs.d/myregistrydomain.com/ca.crt` on every Docker
```
thaJeztah, line 80: @dmcgowan could you have a look? IIRC the `:port` was needed in the path (as different registries could run on different ports), but not sure

dmcgowan: The port is needed if the registry is not on 80 or 443

usha-mandya:
@olimungo Thanks for the PR. The `master` branch has been renamed to `main`. Some of the existing PRs have been automatically closed as a result of this change. Could you create a new PR against our latest docs if the suggested changes are still relevant? Thank you.

PR #10547, opened 2020-03-31
Changed all 'apt-get" to "apt"

Arkasian:
### Proposed changes

As of, you state in there, this is for: Debian 10 or Raspbian 9, it is recommended for those to use "apt" not "apt-get".
apt is 100% compatible with apt-get even in case of repos and other such as: add-apt-key, apt-cache etc; I've tested that installing Docker Client and Docker Engine on my Linux Mint 19.3 host. - replacing apt-get to apt works very well.

### DIFF
--- a/install/linux/docker-ce/debian.md
+++ b/install/linux/docker-ce/debian.md
@@ -25,6 +25,8 @@ Raspbian versions:
 - Buster 10 (stable)
 - Stretch 9 / Raspbian Stretch
 
+> **Note**: If you want to install Docker Client and Docker Engine on older versions of deb-based Linux distros, you can use "apt-get" instead of "apt".
+
 Docker Engine - Community is supported on `x86_64` (or `amd64`), `armhf`, and `arm64` architectures.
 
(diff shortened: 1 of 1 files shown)


thaJeztah:
We use `apt-get` instead of `apt`, because many people use these instructions as a guide to automate installs or updates. While `apt` provides nice output, it does not have a stable interface, and as such is not recommended for that purpose; here's what's being printed when using `apt` instead of `apt-get`;

```bash
apt install -y curl > install_logs.txt

WARNING: apt does not have a stable CLI interface. Use with caution in scripts.
```

I'll close this PR because of the above, but feel free to continue the conversation

Arkasian:
So, would just mention to use "apt" or just mention it is exactly the same to use that one or that one manually from CLI? Because you recommend this for users on Debian 10, but commands are for apt-get which was preinstalled till Debian 7 Wheezy. As I mentioned in my PR, it is important to acknowledge people that apt is recommended when manually installing something.

Issue #10546, opened 2020-03-31
broken doc link from start-the-docker-daemon to configure-docker-to-start-on-boot


GailMarieDB:
### Problem description
uncharted waters for link:
https://docs.docker.com/install/linux/linux-postinstall//#configure-docker-to-start-on-boot

### Problem location
link is on page: https://docs.docker.com/config/daemon/systemd/#start-the-docker-daemon

- I saw a problem on the following URL: <URL>
https://docs.docker.com/install/linux/linux-postinstall//#configure-docker-to-start-on-boot

- I couldn't find the information I wanted. I expected to find it near the following URL: <URL>

- Other: <DETAILS>

PR #10548, opened 2020-03-31
Fix various links that were generating URLs with `.md`

thaJeztah:
There's a bug in the "jekyll-relative-links" plugin that causes wrapped links to not work. Also replacing some links to point to the .md file, so that IDE's can check if the anchors are valid.
Finally, replaced some links to point to their new locations, so that users don't get redirected..

Found wrapped markdown links by grepping for `\[[^\]]*\n.*\]\([^\n]+.md`

fixes https://github.com/docker/docker.github.io/issues/10546
closes https://github.com/docker/docker.github.io/pull/10551


thaJeztah:
ping @usha-mandya PTAL

compose/compose-file/compose-file-v2.md
```diff
@@ -1510,8 +1510,8 @@ volume_driver: mydriver
 > entry in the [top-level `volumes` option](#volume-configuration-reference).
 
 
-See [Docker Volumes](/engine/userguide/dockervolumes.md) and
-[Volume Plugins](/engine/extend/plugins_volume.md) for more information.
+See [Docker Volumes](/storage/volumes.md) and
+[Volume Plugins](/engine/extend/plugins_volume/) for more information.
```
usha-mandya, line 1514: @thaJeztah this entry has .md removed from the link. Is this intentional?

usha-mandya: Also, couldn't find the `/extend/` directory within `/engine/`. Do we autogenerate these files?

thaJeztah: Ah, yes, this is one of the files we fetch during the build.

I could change it back if you want, but it's also ok to use the "final" URL (without .md), as there's nothing to link to while working in this repository

compose/compose-file/compose-file-v2.md
```diff
@@ -271,7 +271,7 @@ An entry with the ip address and hostname is created in `/etc/hosts` inside cont
 Specify a build’s container isolation technology. On Linux, the only supported value
 is `default`. On Windows, acceptable values are `default`, `process` and
 `hyperv`. Refer to the
-[Docker Engine docs](/engine/reference/commandline/run.md#specify-isolation-technology-for-container---isolation)
+[Docker Engine docs](/engine/reference/commandline/run.md##specify-isolation-technology-for-container---isolation)
```
thaJeztah, line 274: oops; double `##` here; let me fix that

compose/compose-file/compose-file-v2.md
usha-mandya: Thanks. Yes, not necessary to add the .md extension if the file aren't present.

ee/ucp/kubernetes/storage/use-azure-files.md
```diff
@@ -287,7 +285,5 @@ subjects:
 
 ## Where to go next
 
-- [Deploy an Ingress Controller on
-  Kubernetes](/ee/ucp/kubernetes/layer-7-routing/)
-- [Discover Network Encryption on
-  Kubernetes](/ee/ucp/kubernetes/kubernetes-network-encryption/)
+- [Deploy an Ingress Controller on Kubernetes](/ee/ucp/kubernetes/layer-7-routing/)
+- [Discover Network Encryption on Kubernetes](../kubernetes-network-encryption.md)
```
usha-mandya, line 289: Wondering why we have a relative path and an absolute path for topics in the same directory

engine/api/sdk-index-new.md
```diff
@@ -0,0 +1,189 @@
...
+
+[Read the full Docker Engine Go SDK reference](https://godoc.org/github.com/docker/docker/client).
+
+### Python SDK
+
+-	**Recommended**: Run `pip install docker`.
+
+-	**If you can't use `pip`**:
+
+	1.	[Download the package directly](https://pypi.python.org/pypi/docker/).
```
usha-mandya, line 44: Is this a formatting issue?

ee/ucp/kubernetes/storage/use-azure-files.md
thaJeztah: Looked at some of those (probably worth a discussion after ee-topics have migrated). So:

Advantage of relative links:

- they work when browsing the markdown on GitHub (I think absolute links don't work there, but would have to double-check)
- moving a whole directory to a different location doesn't break links to pages in the same directory
- ^^ this could help with, e.g. moving whole of `ee/ucp` topic to an archived location (`datacenter/ucp/3.x`), and still linking to related topics for the _same_ version (not the "current" version)

Disadvantage of relative links:

- they can be harder to read. it's "ok" when linking to a child directory (`child-dir/page.md`), but becomes hard to follow when linking to parent directories (`../../../somedir/page.md`)
- it's hard to find/replace links when things move (can't search for `/some/old/dir/page.md` to find which links needs updating)


I think the disadvantages are real; mostly when linking to "other" topics (linking to, e.g., a reference-docs page from a product manual). I'm ok with using relative links to link within the same topic though.

engine/api/sdk-index-new.md
thaJeztah: oh! boo! this was a scratch file I had on my machine, shouldn't have committed it in this PR; let me remove it 😅. Thanks for catching!

thaJeztah: Fixed 👍

ee/ucp/kubernetes/storage/use-azure-files.md
usha-mandya: Sure, thanks for that. Agree, relative links can be very confusing when linking to parent directories. We can discuss this in one of our weekly syncs.

PR #10550, opened 2020-03-31
Update docker-compose install commands

nikitabuyevich:
Update the install documentation for the docker-compose install commands to always download the latest release instead of hard coding the release version in the install command.

Linking to releases: https://help.github.com/en/github/administering-a-repository/linking-to-releases

### DIFF
--- a/compose/install.md
+++ b/compose/install.md
@@ -98,7 +98,7 @@ and want to install Docker Compose.
     Compose (v{{site.compose_version}}):
 
     ```powershell
-    Invoke-WebRequest "https://github.com/docker/compose/releases/download/{{site.compose_version}}/docker-compose-Windows-x86_64.exe" -UseBasicParsing -OutFile $Env:ProgramFiles\Docker\docker-compose.exe
+    Invoke-WebRequest "https://github.com/docker/compose/releases/latest/download/docker-compose-Windows-x86_64.exe" -UseBasicParsing -OutFile $Env:ProgramFiles\Docker\docker-compose.exe
     ```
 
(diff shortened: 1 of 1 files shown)


thaJeztah:
Thanks!

I Was wondering if I could find a reason for using the fixed version numbers (tracked it all the way back to when this page still lived in the docker/compose repository; https://github.com/docker/compose/pull/25)

Some reasons I could find/come up with;

- archived versions of the docs to reflect the state of versions at the time the version was archived
- I saw one or two occasions where the version in the documentation was rolled back (didn't look for reasons, but it could be if there was a faulty release and it had to temporarily be rolled back, waiting for a fixed version).
- providing a more "stable" URL

Perhaps there were concerns about `latest` pointing to a pre-release, but I don't think that's the case (it currently redirects to the "stable" release, even though there's an `-rc` available)


Note that there's more references to the `compose_version`, some of which may have to be updated (there's one explaining how to download a different version); that one may require more explanation if we change to `latest`, because the URL format for versioned URLs is different than for latest (`/latest/download/...` versus `download/VERSION/...` - the position of the "version" is reversed)

So, while this change prevents stale links in the documentation (and possibly users bookmarking older releases), updating the version is part of our release checklist (so generally should be up to date), and _does_ remove some control over what version is published in the docs (which should usually not be needed, but it could be that we want to coordinate "when" a release is published in the documentation).

@usha-mandya @ulyssessouza any thoughts?

nikitabuyevich:
That's true. I think it won't hurt to list both as an option then. 

I proposed this change since for most package installations, we just always install the latest version.

PR #10553, opened 2020-04-01
remove Jenkinsfile from dockerignore

thaJeztah:
There's no longer a Jenkinsfile in this repo, so we can remove it


StefanScherer:
Good catch. 😀

PR #10556, opened 2020-04-01
Add Catalina zsh support statement

dominict:

### DIFF
--- a/install/linux/linux-postinstall.md
+++ b/install/linux/linux-postinstall.md
@@ -36,7 +36,7 @@ creates a Unix socket accessible by members of the `docker` group.
 
 To create the `docker` group and add your user:
 
-1.  Create the `docker` group.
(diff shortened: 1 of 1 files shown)


install/linux/linux-postinstall.md
```diff
@@ -36,7 +36,7 @@ creates a Unix socket accessible by members of the `docker` group.
 
 To create the `docker` group and add your user:
 
-1.  Create the `docker` group.
+1.  Create the `docker` group. On a current MacOS, you may find the following commands absent in the Terminal. In this case, try creating the group in the Users and Groups System Preference Pane. Then, add your user account there too.
```
thaJeztah, line 39: This page is specifically for Linux; on Docker Desktop, there's no need to create a `docker` group, as the user already has access to the engine API.

This also isn't related to Catalina or zsh; macOS never had the `groupadd` / `useradd` command line tools (they're specific to Linux)

thaJeztah:
Closing this PR because of the above, but thanks again for contributing

PR #10562, opened 2020-04-02
Change naming pattern for Dockerfiles to <purpose>.Dockerfile

### LINKED ISSUE #9245: Use <purpose>.dockerfile naming convention
### Problem

In multi-stage builds, we tend to require multiple `Dockerfile` files. Some organisations opt for separating these into their own folders such as:
```
- /root
-- /docker
--- /cli
---- Dockerfile
--- /db
---- Dockerfile
--- /nginx
---- Dockerfile
```

This is OK, but can be difficult to find "the right file" you want to edit in an IDE as you have to rely on being able to see the path name.

Another solution is to add the context to the Dockerfile name, such as for the extension:
```
- /root
-- /docker
--- Dockerfile.cli
--- Dockerfile.db
--- Dockerfile.nginx
```

But this creates another issue: we no longer have syntax highlighting out-of-the-box.

What if we standardise and use `<purpose>.dockerfile` as the preferred naming convention? Such as:
```
- /root
-- /docker
--- cli.dockerfile
--- db.dockerfile
--- nginx.dockerfile
```

The `.dockerfile` extension already works for syntax highlighting in `VS Code` using the [glob pattern](https://github.com/microsoft/vscode-docker/blob/b39024077508500cdcfbba469dfb7c9d413d51e6/src/constants.ts#L33)

Also, using a lowercase `.dockerfile` would match with general naming convention for other extensions. I think we should avoid capitalising `.Dockerfile` but keeping `Dockerfile` for extensionless naming (e.g. in the case of only ever requiring the use of one `Dockerfile` for your project).

### Required changes

Adapt documentation to promote the use of `<purpose>.dockerfile` naming convention, ensure any `docker-compose.yml` files also contain the updated `Dockerfile` filenames.

dan2k3k4:
Fixes #9245


usha-mandya:
@thaJeztah Could you PTAL?

thaJeztah:
> But this creates another issue: we no longer have syntax highlighting out-of-the-box.

What IDE are you using? In my IDE, it's configured to Highlight `Dockerfile`, `*.Dockerfile` and `Dockerfile.*`.

Not a big fan of changing this, because the Dockerfiles now no longer group together

dan2k3k4:
@thaJeztah Create a file like `Dockerfile.php` in `PHPStorm` or Microsoft VS `code`, it will think it's a "PHP" file because it's `*.php`.

thaJeztah:
I understand, but that's not at hand here; the `Dockerfile.archive` Dockerfile is only used by the docs team when creating an archive branch, so possibly we could even move it elsewhere if that's the biggest problem

dan2k3k4:
Ah for `Dockerfile.archive` - I can revert that change.

Perhaps we could mention that people could setup their `Dockerfile`s like this:
```
/project
  - /docker
    - /php
      - Dockerfile
    - /nginx
      - Dockerfile
    - /mariadb
      - Dockerfile
```

So that the path defines the grouping for the `Dockerfile` images?

I guess I just want to see some documentation to get people to avoid using a name like `Dockerfile.php` which could be parsed by PHP either in IDE or elsewhere 😄

dan2k3k4:
@thaJeztah, I removed the changes to the `Dockerfile.archive` file

ThomasLohmann:
@dan2k3k4  @thaJeztah any updates on this topic? 

Also what is currently more preferable without less issues in IDEs, Kubernetes etc.: Dockerfile.<purpose> vs <purpose>.Dockerfile?

Regards,
Thomas

dan2k3k4:
@ThomasLohmann we've been using `<service-name>. dockerfile` and VSCode (and I believe Jetbrains, e.g. PHPStorm) pick it up just fine for syntax highlighting.

ThomasLohmann:
@dan2k3k4 great! And there is no issue to deploy images in this way on Kubernetes?

dan2k3k4:
@ThomasLohmann nope, [https://github.com/uselagoon/lagoon-images/blob/main/images/php-fpm/8.0.Dockerfile](uselagoon/lagoon-images) repository is done this way too and that hosts several thousand different websites including the Australian government.

ThomasLohmann:
@dan2k3k4 perfect! 

Let's hope that docker will support purpose.dockerfile instead of purpose.Dockerfile soon. It would then be more like a real file extension.

dan2k3k4:
The capitalisation doesn't matter. We use lowercase d in our images but the Lagoon repo is using uppercase, I think just for clarity.

PR #10549, opened 2020-03-31
Template: remove "base href" and fix javascript link-fixing

thaJeztah:
The base href caused browsers to produce incorrect URLs for anchor links on various pages, for example, pointing to:

    https://docs-test.docker.com/engine/reference/#foreground

Instead of

    https://docs-test.docker.com/engine/reference/run/#foreground

While there might be links generated incorrectly (links including `.md`), we should fix those when generating the HTML instead of trying to fix them up using JavaScript.

Unfortunately, Jekyll doesn't process links to markdown files in includes, so for includes in this repository, these were changed to "absolute" "html" links (not linking to Markdown). After that, the only place we need the JavaScript link fix is within the reference docs.

I rewrite the script to be more efficient, documented each step (sorry, quite verbose on the comments), and limited its scope to only run on the reference docs, and on those, only on links within the "content".

fixes https://github.com/docker/docker.github.io/issues/2655
fixes https://github.com/docker/docker.github.io/issues/10647

relates to https://github.com/docker/docker.github.io/issues/247
relates to https://github.com/docker/docker.github.io/issues/8070
relates to https://github.com/docker/docker.github.io/issues/6772
relates to https://github.com/docker/docker.github.io/issues/5711
relates to https://github.com/docker/docker.github.io/issues/5388
relates to https://github.com/docker/docker.github.io/issues/3952
relates to https://github.com/docker/docker.github.io/issues/2655

PR #10573, opened 2020-04-05
Optimize keywords-list logic in docs layout

ashmaroli:
### Proposed changes

The Liquid logic to render a list of keywords for a document consists of using a recursive
`{% capture %}` of words wrapped in an `<a/>` tag. This *recursive capturing* is inefficient both speed-wise and memory-usage-wise.

This pull request refactors the logic by:
- wrapping the core logic in a conditional that checks if `page.keywords | split:","` results in a non-empty array right away.
- using Liquid's whitespace-control feature to conform to whitespace quirks of an HTML `<span />` tag yet nest and indent Liquid markup for readability.

### DIFF
--- a/_layouts/docs.html
+++ b/_layouts/docs.html
@@ -177,11 +177,16 @@
 							{{ content }}
 							<!-- tags -->
 							{% unless page.notags == true %}
-							{% assign keywords = page.keywords | split:"," %}
-							{% for keyword in keywords %}{% assign strippedKeyword = keyword | strip %}
-							{% capture keywordlist %}{{ keywordlist }}<a href="https://docs.docker.com/search/?q={{strippedKeyword}}">{{strippedKeyword}}</a>{% unless forloop.last %}, {% endunless %}{% endcapture %}
-							{% endfor %}
-							{% if keywordlist.size > 0 %}<span class="glyphicon glyphicon-tags" style="padding-right: 10px"></span><span style="vertical-align: 2px">{{ keywordlist }}</span>{% endif %}
+							{% assign keywords = page.keywords | split:"," -%}
+							{% if keywords.size > 0 -%}
+							  <span class="glyphicon glyphicon-tags" style="padding-right: 10px"></span><span style="vertical-align: 2px">
+							  {%- for keyword in keywords -%}
+							    {%- assign strippedKeyword = keyword | strip -%}
+							    <a href="https://docs.docker.com/search/?q={{ strippedKeyword }}">{{ strippedKeyword }}</a>
+							    {%- unless forloop.last %}, {% endunless -%}
+							  {% endfor -%}
+							  </span>
+							{% endif -%}
 							{% endunless %}
 							<!-- link corrections -->
               <script language="JavaScript">


_layouts/docs.html
```diff
@@ -177,11 +177,16 @@ <h2>Next steps</h2>
...
 							<!-- tags -->
 							{% unless page.notags == true %}
-							{% assign keywords = page.keywords | split:"," %}
-							{% for keyword in keywords %}{% assign strippedKeyword = keyword | strip %}
-							{% capture keywordlist %}{{ keywordlist }}<a href="https://docs.docker.com/search/?q={{strippedKeyword}}">{{strippedKeyword}}</a>{% unless forloop.last %}, {% endunless %}{% endcapture %}
-							{% endfor %}
-							{% if keywordlist.size > 0 %}<span class="glyphicon glyphicon-tags" style="padding-right: 10px"></span><span style="vertical-align: 2px">{{ keywordlist }}</span>{% endif %}
+							{% assign keywords = page.keywords | split:"," -%}
+							{% if keywords.size > 0 -%}
+							  <span class="glyphicon glyphicon-tags" style="padding-right: 10px"></span><span style="vertical-align: 2px">
```
thaJeztah, line 182: Just noticing this here (not needed to change as part of this PR); we should get rid of these inline styles, and either add a CSS class or add a style for `glyphicon` /  `glyphicon-tags` to our stylesheet to do the padding.

Also need to check why we're adding the extra `<span>` here, and if instead we could use the `<a>` itself to do this (haven't tried yet)

_layouts/docs.html
```diff
@@ -177,11 +177,16 @@ <h2>Next steps</h2>
...
+							  {%- for keyword in keywords -%}
+							    {%- assign strippedKeyword = keyword | strip -%}
+							    <a href="https://docs.docker.com/search/?q={{ strippedKeyword }}">{{ strippedKeyword }}</a>
```
thaJeztah, line 185: Also ok for a follow-up; we should remove the hard-coded `https://docs.docker.com` domain here. We're actually removing it again after generating the docs, so might as well just not add it in the first place 😂

thaJeztah:
@usha-mandya ptal; I built this locally, and tag-links at the bottom of pages still rendered correctly with this patch, so I think it's good to go

ashmaroli:
@thaJeztah Thank you for reviewing this.
Were you able to notice any improvement in the total build time with this branch?

ashmaroli:
On the whole, total build times can be improved by reducing the work done by Liquid.
(Specifically, the logic for basehref (which is being removed in a different PR), the logic for liquid-only-toc and the logic for computing reading time of a page.

thaJeztah:
Oh, I didn't time the build TBH (started a build and continued working on another project, 😅). We should definitely try to get rid of some of the weird "hacks" that are currently in place.

https://github.com/docker/docker.github.io/pull/10549 removes some of it; we know that PR is a bit risky, as there may still be broken links to markdown pages that moved, which is why we didn't merge it before the weekend 😂; even then: I much prefer actually _fixing_ broken links at the source over "trying to fix them up during build" (or even with JavaScript 🙀).

I should note that with Mirantis having acquired the "enterprise" part of our products, soon the enterprise section will be fully migrated to the Mirantis documentation website, which will significantly reduce the number of pages in the docs (the enterprise section also has various archived versions included in the documentation)

usha-mandya:
Thank you So much, @ashmaroli for cleaning this up. Thank you @thaJeztah for your insights.

ashmaroli:
Thanks for the review and merge @usha-mandya :smiley:

PR #10579, opened 2020-04-06
Update Docs header and footer

usha-mandya:
### Proposed changes

- Updated the footer to remove redundant links and ensured the entries and layout match the footer displayed on the WWW site.
- Removed Sign in and Create Docker ID option from the header
- Removed Get Support from the right-nav

### DIFF
--- a/_includes/footer.html
+++ b/_includes/footer.html
@@ -20,49 +20,41 @@
             <div class="row">
                 <div class="col-xs-12 col-sm-3 col-md-3">
                     <ul class="footer_links">
-                        <li><a href="https://www.docker.com/what-docker">What is Docker</a></li>
+                        <li><b><a href="https://www.docker.com/why-docker">Why Docker</a></b></li>
                         <li><a href="https://www.docker.com/what-container">What is a Container</a></li>
+                    </ul>
+                </div>
+                <div class="col-xs-12 col-sm-3 col-md-3">
+                    <ul class="footer_links">
+                        <li><b><a href="https://www.docker.com/products/overview">Products</a></b></li>
+                        <li><a href="https://www.docker.com/products/docker-desktop">Docker Desktop</a></li>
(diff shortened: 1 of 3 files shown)


usha-mandya:
Deploy preview - https://dockerdocs-preview.netlify.com/

usha-mandya:
@thaJeztah PTAL

_includes/footer.html
```diff
@@ -89,7 +81,6 @@
                     <li class="fa fa-github"><a href="https://github.com/docker">GitHub</a></li>
                     <li class="fa fa-linkedin"><a href="https://www.linkedin.com/company/docker">Linkedin</a></li>
                     <li class="fa fa-facebook"><a href="https://www.facebook.com/docker.run">Facebook</a></li>
-                    <li class="fa fa-reddit"><a href="http://www.reddit.com/r/docker">Reddit</a></li>
```
thaJeztah, line 92: We don't actively use Reddit? (note: if we bring it back, we should use `https://` instead of `http://`

_includes/footer.html
```diff
@@ -80,7 +72,7 @@
         <div class="bottom_footer">
             <div class="footer-copyright col-xs-12 col-md-8">
                 <p class="copyright">
-                    Copyright &copy; 2019 Docker Inc. All rights reserved. </p>
+                    Copyright &copy; 2020 Docker Inc. All rights reserved. </p>
```
thaJeztah, line 75: If we _do_ publish years, this should probably be 2013-2020

Not sure how the `All rights reserved` portion relates with the license (which is Apache 2 for the documentation source)

usha-mandya: Haven't changed the text here. The format is also inline with the copyrights format on docker.com. 

@lisaberry Could you provide some guidance here? Thanks.

lisaberry: 2013-2020.  Leave the other text in

usha-mandya: Updated the years. Thank you both!

PR #10593, opened 2020-04-08
Re-order Compose TOC

charleswhchan:
### Proposed changes

Re-order Compose TOC:
  - Environment variables in Compose
  - Environment variables in .env file

This makes more sense since environment variables and .env file is first mentioned in the section labelled "Environment variables in Compose"

### DIFF
--- a/_data/toc.yaml
+++ b/_data/toc.yaml
@@ -1100,10 +1100,10 @@ manuals:
     title: Getting started
   - path: /compose/swarm/
     title: Using Compose with Swarm
-  - path: /compose/env-file/
-    title: Environment file
   - path: /compose/environment-variables/
     title: Environment variables in Compose
+  - path: /compose/env-file/
+    title: Environment file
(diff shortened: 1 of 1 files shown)


_data/toc.yaml
```diff
@@ -1100,10 +1100,10 @@ manuals:
     title: Getting started
   - path: /compose/swarm/
     title: Using Compose with Swarm
-  - path: /compose/env-file/
-    title: Environment file
   - path: /compose/environment-variables/
     title: Environment variables in Compose
+  - path: /compose/env-file/
+    title: Environment variables in .env file
```
usha-mandya, line 1106: @charleswhchan The re-ordering makes sense. Thank you. Could you retain the original title `Environment file` for this entry?

charleswhchan: Fixed. Please check commit b36210d again. Thanks.

PR #10596, opened 2020-04-08
Update Swarm deploy doc

charleswhchan:
### Proposed changes
Improve the Swarm deploy doc to make it more user friendly


craig-osterhout:
@charleswhchan, thank you for the contribution. As Docker Desktop can run on Windows, scrolling up is the best universal way to get the info and grep would not be applicable in that case. As for the file name, it has been updated.

PR #10598, opened 2020-04-08
Consolidate Desktop TOC entries

usha-mandya:
Moved Docker Desktop for Mac and Docker Desktop for Windows topics to a high-level 'Docker Desktop' entry.

### DIFF
--- a/_data/toc.yaml
+++ b/_data/toc.yaml
@@ -1116,64 +1116,66 @@ manuals:
     title: Sample apps with Compose
   - path: /release-notes/docker-compose/
     title: Docker Compose release notes
-- sectiontitle: Docker Desktop for Mac
+- sectiontitle: Docker Desktop
   section:
-  - path: /docker-for-mac/
-    title: Getting started
-  - path: /docker-for-mac/install/
-    title: Install Docker Desktop for Mac
-  - path: /docker-for-mac/dashboard/
-    title: Dashboard
-  - path: /docker-for-mac/kubernetes/
-    title: Deploy on Kubernetes
-  - path: /docker-for-mac/docker-toolbox/
-    title: Docker Desktop for Mac vs. Docker Toolbox
-  - path: /docker-for-mac/multi-arch/
-    title: Leveraging Multi-CPU architecture support
-  - path: /docker-for-mac/networking/
-    title: Networking
-  - path: /docker-for-mac/osxfs/
-    title: File system sharing
-  - path: /docker-for-mac/osxfs-caching/
-    title: Performance tuning for volume mounts (shared filesystems)
-  - path: /docker-for-mac/space/
-    title: Disk utilization
-  - path: /docker-for-mac/troubleshoot/
-    title: Logs and troubleshooting
-  - path: /docker-for-mac/faqs/
-    title: FAQs
-  - path: /docker-for-mac/opensource/
-    title: Open source licensing
-  - path: /docker-for-mac/release-notes/
-    title: Stable release notes
-  - path: /docker-for-mac/edge-release-notes/
-    title: Edge release notes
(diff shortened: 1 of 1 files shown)


_data/toc.yaml
```diff
@@ -1116,64 +1116,66 @@ manuals:
...
-    title: FAQs
-  - path: /docker-for-windows/opensource/
-    title: Open source licensing
-  - path: /docker-for-windows/release-notes/
-    title: Stable release notes
-  - path: /docker-for-windows/edge-release-notes/
-    title: Edge release notes
-  - path: /docker-for-windows/wsl-tech-preview/
-    title: Docker Desktop WSL 2 backend
+  - sectiontitle: Mac
```
thaJeztah, line 1121: I was slightly on the fence if keeping `Docker Desktop for` would be better, but just built this branch, and it's probably good as-is.

usha-mandya: Yea, I tried with `Docker Desktop for` in the first iteration, but decided to go with `Mac` and `Windows' later to avoid repetition.

image

usha-mandya:
Having a consolidated "overview" / "index" for Docker Desktop: describe what Docker Desktop is
not specific to "Windows" or "Mac"
it could provide a support matrix for Windows/MacOS versions, similar to https://docs.docker.com/engine/install/
Perhaps we should consider moving the "dashboard" topics to the "common" section as well
Consider updating the URLs (/desktop/ (common topics), /desktop/mac/ (mac specific), /desktop/windows/ (windows specific)

Thanks @thaJeztah. Great points. 

- Yes, looking to introduce a generic overview (similar to Engine overview). On that note, we should have a list of mandatory topics in each product manual, for example,  an overview, release note, faq, troubleshooting.
- Re. Common topics: There are a couple of other topics in addition to Dashboard, for example, Networking, OS, etc. that we can look to move to the common section. I believe we can either go down the route we took in DDE (https://docs.docker.com/ee/desktop/) and list topics and then document specific instructions for Mac and Windows, or retain the existing structure and let users choose a section depending on what OS they are on, and provide information related to their OS. I'm slightly leaning towards the latter as Mac users don't have to sift through the pages to find info applicable to Mac, and vice versa. Happy to discuss this further in a separate thread.

usha-mandya:
> Following our slack conversation - happy to go ahead with this as long as we monitor the result :)

Thanks @nebuk89. I'll make a note to compare page views after the changes go live.

thaJeztah:
>  believe we can either go down the route we took in DDE (https://docs.docker.com/ee/desktop/) and list topics and then document specific instructions for Mac and Windows, or retain the existing structure and let users choose a section depending on what OS they are on, and provide information related to their OS. I'm slightly leaning towards the latter as Mac users don't have to sift through the pages to find info applicable to Mac, and vice versa.

The DDE looks a bit too-deeply-nested (e.g. Admin -> Install -> Install DDE on Mac / Install DDE on Windows).

I would consolidate pages that are common, but only differ in minor parts (e.g. screenshots, possibly some options that slightly differ). To "personalise" those pages to the platform in use, we could use tabs (similar to the [`linux instal pages`](https://docs.docker.com/engine/install/debian/)); I would love to see a way for _all_ tabs to toggle at the same time, and possibly store preferences in a cookie (toggle mac/windows, and have all tabs on a page point to the correct platform; store that preference and when I navigate to another page that also has mac/windows have the tabs default to my last picked platform)

usha-mandya:
>I would love to see a way for all tabs to toggle at the same time, and possibly store preferences in a cookie (toggle mac/windows, and have all tabs on a page point to the correct platform; store that preference and when I navigate to another page that also has mac/windows have the tabs default to my last picked platform)

It would be great if we can provide a way to automatically toggle between Mac and Windows.

PR #10604, opened 2020-04-08
Improvements to Docker Swarm service doc

charleswhchan:
### Proposed changes
Clarify replicate mode and update the example for global service

### DIFF
--- a/engine/swarm/how-swarm-mode-works/services.md
+++ b/engine/swarm/how-swarm-mode-works/services.md
@@ -106,14 +106,14 @@ For a replicated service, you specify the number of identical tasks you want to
 run. For example, you decide to deploy an HTTP service with three replicas, each
 serving the same content.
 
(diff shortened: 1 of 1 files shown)


engine/swarm/how-swarm-mode-works/services.md
```diff
@@ -106,14 +106,14 @@ For a replicated service, you specify the number of identical tasks you want to
 run. For example, you decide to deploy an HTTP service with three replicas, each
 serving the same content.
 
-A global service is a service that runs one task on every node. There is no
+A global service is a service that runs one copy of the task on every node. There is no
```
thaJeztah, line 109: Don't think "one copy of the task" is correct here; as `task` in itself is already meaning "an instance of the service"

charleswhchan: Thanks for the clarification. I will decline the PR.

PR #10606, opened 2020-04-08
Fixed Issue #8552

### LINKED ISSUE #8552: Correction to SSH Service Page
### Problem location

https://docs.docker.com/engine/examples/running_ssh_service/

There is a fault or a missing char in dockerfile so sshd server don't work when tried to login, access denied error is given, even the password is correct.

### Suggestions for a fix

Problem is that, sshd_config's default file starts with '#' sign but below command don't have it.

RUN sed -i 's/PermitRootLogin prohibit-password/PermitRootLogin yes/' /etc/ssh/sshd_config

So, it is being fixed with adding '#' as below

RUN sed -i 's/#PermitRootLogin prohibit-password/PermitRootLogin yes/' /etc/ssh/sshd_config

clarkimusmax:
Fixed #8552.  Made the SSH configuration sed line on the page match the line in the Dockerfile.

https://github.com/docker/docker.github.io/issues/8552

### DIFF
--- a/engine/examples/running_ssh_service.md
+++ b/engine/examples/running_ssh_service.md
@@ -25,7 +25,7 @@ FROM ubuntu:16.04
 RUN apt-get update && apt-get install -y openssh-server
 RUN mkdir /var/run/sshd
 RUN echo 'root:THEPASSWORDYOUCREATED' | chpasswd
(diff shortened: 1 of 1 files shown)


craig-osterhout:

PR #10612, opened 2020-04-09
ci: Add GitHub action to upload S3 website configuration

StefanScherer:
Add a GitHub action to upload the S3 website configuration from a JSON file in this repo.
With that PR it's possible to maintain the redirections directly from this repo by making pull requests to
- `_website-config-docs-stage.json` for https://docs-stage.docker.com
- `_website-config-docs.json` for https://docs.docker.com

An example for the docs-stage.docker.com site with a redirect of all `v17.03/` to `/`
Important keys are
- `KeyPrefixEquals` is set to `v17.03/`
- `ReplaceKeyPrefixWith` is set to empty string `""` which means `/`
- `HostName` and `Protocol` must be set for each redirect rule, otherwise AWS redirects to the internal name. That's why we need two different JSON files.

```json
  {
   "Condition": {
    "HttpErrorCodeReturnedEquals": null,
    "KeyPrefixEquals": "v17.03/"
   },
   "Redirect": {
    "HostName": "docs-stage.docker.com",
    "HttpRedirectCode": null,
    "Protocol": "https",
    "ReplaceKeyPrefixWith": "",
    "ReplaceKeyWith": null
   }
  },
```


StefanScherer:
A temporary build that also ran in this PR branch was green: https://github.com/docker/docker.github.io/pull/10612/checks?check_run_id=573354449 and it successfully uploaded the website config for docs-stage.

StefanScherer:
Ping @thaJeztah this is the replacement for the nginx redirect rules we had in the past. Now a similar config for S3 website is back in the repo to maintain it through PRs.

thaJeztah:
Thanks! I just recalled we also need to take `redirect_from` entries in the `/ee` and `/datacenter` into account before we remove the content; here's a quick 'n dirty way to get which pages have redirects from outside of `/(ee|datacenter|compliance)`;

```bash
echo "{" > extra-redirects.json
grep -E ': "/(compliance|datacenter|ee)/' <(curl -fsSL https://docs.docker.com/redirects.json | jq .) >> extra-redirects.json
echo '  "/dummy/": "/"' >> extra-redirects.json
echo "}" >> extra-redirects.json
```


[extra-redirects.json.txt](https://github.com/docker/docker.github.io/files/4455732/extra-redirects.json.txt)

thaJeztah:
For those pages, I think we'd need dummy pages in `/ee`, `/datacenter` and `/compliance`

StefanScherer:
This PR only adds automation to  the current status. I've uploaded the website configs manually with a similar Go tool, there's no aws cli command for that.

For the three EE related folders I plan to do this in a new PR after the merge. We can simply remove the three folders and add three redirection rules to the new website*.json files. We will test this PR in docs-stage to see if everything gets a new URL prefix with the target domain name.

thaJeztah:
> We can simply remove the three folders and add three redirection rules to the new

Right, so that's the tricky bit; there's redirects that are currently in place that redirect _to_ the `/ee` or `/datacenter` directory, but due to how Jekyll constructs redirects, those redirects are configured in the _target_ page.

For example, https://docs.docker.com/reference/ucp/3.0/cli/support/ redirects to https://docs.docker.com/datacenter/ucp/3.0/reference/cli/support/

That redirect is defined in https://github.com/docker/docker.github.io/blame/663ea14a3ca321d8f4a257d3b5ca8648ba5aec2d/datacenter/ucp/3.0/reference/cli/support.md#L6

Which means that once that page is removed, the redirect is no longer in place. So for pages that define a redirect, we should either replace them with a "dummy" page that only has the `redirect_from` page set (so that Jekyll generates the redirect page), or commit the generated redirect HTML's to git

thaJeztah:
Updated, with redirect "from" `/ee/` and `/datacenter/` removed

[extra-redirects.json.txt](https://github.com/docker/docker.github.io/files/4493136/extra-redirects.json.txt)

StefanScherer:
I don't know how much effort we want to put into the forwarding to Mirantis. Do we have an "SLA" of how many of the old links should work/redirect to them? 😅

PR #10616, opened 2020-04-10
"updating" is spelled "apdating" on line 29

kavikat:


ashmaroli:
@kavikat You must realize that its not **apdating** but **adapting** as in *adapt-ing*.. and that your correction is still a typo in every sense...

thaJeztah:
yes, this PR doesn't look correct; `adapting` is the correct word here

PR #10631, opened 2020-04-13
Update Go SDK examples

flaviuvadan:
### Proposed changes
This PR proposes several Go SDK example changes. The latest version of the Golang Docker client (v1.13.1) includes several changes that are not reflected in the Docker docs:
1. `client.NewClientWithOpts` is no longer available;
2. `client.ContainerWait` no longer returns `(<-chan container.ContainerWaitOKBody, <-chan error)`. Rather, it returns `(int64, error)`.

This motivate the following changes:
1. Update the **Run a container** section to use `client.NewEnvClient()` and the new `ContainerWait` return values;
2. Update **Run a container in the background** to use `client.NewEnvClient()`;
3. Update **List and manage containers** to use `client.NewEnvClient()`;
4. Update **Stop all running containers** to use `client.NewEnvClient()`;
5. Update **Print the logs of a specific container** to use `client.NewEnvClient()`;
6. Update **List all images** to use `client.NewEnvClient()`;
7. Update **Pull an image** to use `client.NewEnvClient()`;
8. Update **Pull an image with authentication** to use `client.NewEnvClient()`;
9. Update **Commit a container** to use `client.NewEnvClient()` and the new `ContainerWait` return values.

These changes are consistent with the latest versions of the client package, which can be found [here](https://pkg.go.dev/github.com/docker/docker/client?tab=doc).

Changes can be seen [on Netlify](https://deploy-preview-10631--docsdocker.netlify.com/engine/api/sdk/examples/#run-a-container).

### Related issues (optional)

I have used the Docker docs "Edit this page" option - no issue to link.

### DIFF
--- a/engine/api/sdk/examples.md
+++ b/engine/api/sdk/examples.md
@@ -52,7 +52,7 @@ import (
 
 func main() {
 	ctx := context.Background()
-	cli, err := client.NewClientWithOpts(client.FromEnv, client.WithAPIVersionNegotiation())
+	cli, err := client.NewEnvClient()
 	if err != nil {
 		panic(err)
 	}
@@ -76,13 +76,9 @@ func main() {
 		panic(err)
 	}
 
-	statusCh, errCh := cli.ContainerWait(ctx, resp.ID, container.WaitConditionNotRunning)
-	select {
-	case err := <-errCh:
-		if err != nil {
-			panic(err)
-		}
-	case <-statusCh:
+	_, err := client.ContainerWait(ctx, resp.ID)
+	if err != nil {
+		panic(err)
 	}
 
 	out, err := cli.ContainerLogs(ctx, resp.ID, types.ContainerLogsOptions{ShowStdout: true})
@@ -157,7 +153,7 @@ import (
 
 func main() {
 	ctx := context.Background()
-	cli, err := client.NewClientWithOpts(client.FromEnv, client.WithAPIVersionNegotiation())
+	cli, err := client.NewEnvClient()
 	if err != nil {
 		panic(err)
 	}
@@ -240,7 +236,7 @@ import (
 
 func main() {
 	ctx := context.Background()
-	cli, err := client.NewClientWithOpts(client.FromEnv, client.WithAPIVersionNegotiation())
+	cli, err := client.NewEnvClient()
 	if err != nil {
 		panic(err)
 	}
@@ -316,7 +312,7 @@ import (
 
 func main() {
 	ctx := context.Background()
-	cli, err := client.NewClientWithOpts(client.FromEnv, client.WithAPIVersionNegotiation())
+	cli, err := client.NewEnvClient()
 	if err != nil {
 		panic(err)
 	}
@@ -397,7 +393,7 @@ import (
 
 func main() {
 	ctx := context.Background()
-	cli, err := client.NewClientWithOpts(client.FromEnv, client.WithAPIVersionNegotiation())
+	cli, err := client.NewEnvClient()
 	if err != nil {
 		panic(err)
 	}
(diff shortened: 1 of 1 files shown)


usha-mandya:
@flaviuvadan Thanks for update.

@thaJeztah Could you please review the suggested changes?

thaJeztah:
> The latest version of the Golang Docker client (v1.13.1)

Ah, yes, I suspect you got bit by go modules 😞. Unfortunately, the introduction of go modules caused `go get` (and equivalents) to break when using the docker client as a dependency.

Where previously `go get` would give you the latest code from master, Go modules now give you the latest tagged release that "looks like SemVer". In this case that's `v1.13.1` (which is over three years old); we recently integrated the tags from the docker/docker-ce repository so probably now you'll get `v17.12.1-rc1`, which Go modules also considers a SemVer tag, and which is the branch-point from master.

The best approach to get a version of the docker api client that matches a docker release, is to specify the tag of that release in your `go.mod` (e.g., `v19.03.8` to get the current release).

Unfortunately things get more complicated (or at least confusing) from there, because go modules does not work with release branches (which is what's often used for projects that actively maintain multiple releases (including golang itself));



docker run -it --rm -v (pwd):/project -w /project golang:1.14.2


Create your `main.go`

```go
package main

import (
	"context"
	"fmt"

	"github.com/docker/docker/api/types"
	"github.com/docker/docker/client"
)

func main() {
	ctx := context.Background()
	cli, err := client.NewClientWithOpts(client.FromEnv, client.WithAPIVersionNegotiation())
	if err != nil {
		panic(err)
	}

	containers, err := cli.ContainerList(context.Background(), types.ContainerListOptions{})
	if err != nil {
		panic(err)
	}

	for _, container := range containers {
		fmt.Println(container.ID)
	}
}
```

Initialize go modules and add `v19.03.8` of the docker dependency;

```bash
export GO111MODULE=on
go mod init myproject
GO111MODULE=on go get github.com/docker/docker@v19.03.8

# go: downloading github.com/docker/docker v17.12.0-ce-rc1.0.20200309214505-aa6a9891b09c+incompatible
# go: github.com/docker/docker v19.03.8 => v17.12.0-ce-rc1.0.20200309214505-aa6a9891b09c+incompatible
```

This is where things get confusing, because due to go modules not supporting release branches, it will replace the version you picked with some (seemingly) random version. That version is based on the "latest" tag it finds on the master branch that looks like SemVer. In our case, `v17.12.0-ce-rc1`; the version that will be picked also depends on which version of Go is used (they broke this in the last couple of releases).

Your `go.mod` will now look something like this (notice that even though the version is shown as `v17.12`, the actual commit (https://github.com/moby/moby/commit/aa6a9891b09c) matches [`v19.03.8`](https://github.com/moby/moby/releases/tag/v19.03.8);

```
module myproject

go 1.14

require github.com/docker/docker v17.12.0-ce-rc1.0.20200309214505-aa6a9891b09c+incompatible // indirect
```

To prevent confusion, you may want to add a comment to your `go.mod` to explain what the actual version is. Also be aware that if your project is not a "final" project, but meant to be used as a module in other projects, that things might still break (go modules will likely pick the actual `v17.12.0-ce-rc1`, which is two years old).

If your project is intended to be used as a module for other projects, you may want to add a `replace` rule to your `go.mod` to forcibly break dependent projects (preventing them to get an incorrect version); similar to https://github.com/moby/buildkit/blob/c62983bbdee18c2bab62ccd8ee1727b86243de76/go.mod#L84 (I think that repository needs to be updated to use anything > 17.12 now :disappointed:)

We're still looking at options to make this easier to work with, but it's quite complicated. The repository uses tags for the _binary_ releases, and has never used SemVer. With Go modules, Golang now assumes that any tag on a repository is created for Go modules (and implicitly assumes that anything that _looks like_ SemVer _must be_ SemVer); even though we don't try to break compatibility between releases, our API client (and other packages) are _not_ SemVer (they match the docker releases, and may introduce changes in the Go "api").

flaviuvadan:
@thaJeztah, I now see the problem. Yes, the project that I am working on definitely brought in v1.13.1. Thank you for the detailed explanation! This was a great lesson/reminder. I guess there are no changes that are required for now, so I will close the PR. You mentioned "I think that repository needs to be updated to use anything > 17.12 now". Could I pick that up?

PR #10655, opened 2020-04-17
Add Execution Policy advice

rolandsaven:
### Proposed changes
If PowerShell's execution policy is set to **Restricted**, the installation fails. This is not necessarily a problem, however, there is no clear indication or error message about the cause. This is not a fault on your side, however, some not so experienced in devops it could be an issue.

The error message you get is `Unable to find package providers (DockerMsftProvider).`, which does not really help to solve the issue.

The execution policy should be set to at least **RemoteSigned**, so that the installation can complete.

### Related issues
https://github.com/OneGet/MicrosoftDockerProvider/issues/14#issuecomment-541286847  - please see only the linked comment (the main issue is not related)


usha-mandya:
@Dawn-Wood PTAL.

usha-mandya:

Let me close this PR here for that reason. Feel free to continue the conversation after I close.

PR #10665, opened 2020-04-20
Update Docker Desktop admin settings documentation

mat007:
### Proposed changes

Updated the Docker Desktop admin settings documentation.

@usha-mandya @ebriney


usha-mandya:
Will merge to master when we are ready to go live.

desktop/enterprise/admin/configure/mac-admin.md
```diff
@@ -118,6 +114,12 @@ The following `admin-settings.json` code and table provide the required syntax a
         "locked" : true
       }
     },
+
+  "filesharingDirectories": {
+    "locked":false,
+    "value":["/Users", "..."]
+  },
```
thaJeztah, line 121: The trailing comma should be removed here, otherwise the JSON is invalid 😅 

```suggestion
  }
```

I think the formatting is also a bit off (indentation is different here compared to lines higher up); perhaps you can remove the blank lines, and reformat the whole JSON in this example?

desktop/enterprise/admin/configure/windows-admin.md
```diff
@@ -154,11 +154,11 @@ The following `admin-settings.json` code and table provide the required syntax a
       }
     },
 
-  "sharedDrives": {
-    "locked": true,
-    "value": [ ]
+  "filesharingDirectories": {
+    "locked":false,
+    "value":["%USERPROFILE%", "..."]
   },
```
thaJeztah, line 160: Same here; should be no trailing comma

```suggestion
  }
```

desktop/enterprise/admin/configure/mac-admin.md
mat007: I had removed the trailing commas initially but then I realized they were there before.
Arguably it’s not really a JSON but an example of part of a JSON…
It’s not valid anyway as for instance the `"..."` here is just a placeholder.

Anyway I don’t really have a strong opinion, I think an alternative could be to add
```
  ...
}
```
at the end to emphasize it’s merely a sample?

thaJeztah: ah, yes, that'd work

Just noticed the trailing comma, and thought it was a complete example

Would be nice to have the formatting/identation a bit consistent though

mat007: I added a commit to format the JSON samples.

Issue #10632, opened 2020-04-13
"docker run -dit" vs "docker run -d"


kzhui125:
from https://github.com/docker/docker.github.io/blob/master/config/containers/start-containers-automatically.md

`$ docker run -dit --restart unless-stopped redis`

What's the difference between "docker run -dit" and "docker run -d"?

Thanks,

thaJeztah:
`-dit` is a short-hand notation for `-d -i -t`.

Usually `-i -t` (or `-it`) is used to run a container interactively, but it can also be used to run a container "detached" (`-d`) (in the background) for situations where the container's process expects a TTY (`-t`) to be attached to keep running.

In this example, a `redis` container is started, which doesn't require `-i` or `-t` to be set to keep running, so I think we can remove it to prevent confusion.

I'll open a PR to remove it

thaJeztah:
opened https://github.com/docker/docker.github.io/pull/10698 to remove it

PR #10698, opened 2020-04-24
remove -it from example, as it's not needed for redis

thaJeztah:
Adding `-it` on a "detached" (`-d`) container is needed
if the container's main process expects a TTY attached to keep
running.

It's not needed for a redis container, so we can remove it here

fixes https://github.com/docker/docker.github.io/issues/10632


thaJeztah:
@usha-mandya PTAL

PR #10711, opened 2020-04-27
fixed max_replicas_per_node example

ticalcster:
### Proposed changes
The example for max_replicas_per_node was not inside the placement dict.


usha-mandya:
@ulyssessouza Could you PTAL?

PR #10717, opened 2020-04-28
compose: add WordPress example to TOC

thaJeztah:
I noticed the WordPress sample was linked in various places, but was not in the TOC.

This PR also removes links to individual examples, instead linking to the "compose samples" page. Perhaps we should consider including a link to the awesome compose repository, which also has the WordPress example (but slightly updated); https://github.com/docker/awesome-compose/tree/master/wordpress-mysql


compose/link-env-deprecated.md
```diff
@@ -39,5 +39,7 @@ Fully qualified container name, such as `DB_1_NAME=/myapp_web_1/myapp_db_1`
```
usha-mandya, line 39: This page isn't linked in the TOC - https://deploy-preview-10717--docsdocker.netlify.app/compose/link-env-deprecated/

compose/faq.md
```diff
@@ -111,9 +111,9 @@ github](https://github.com/search?q=in%3Apath+docker-compose.yml+extension%3Ayml
```
usha-mandya, line 111: This may be out of scope of this PR, but wondering if it makes sense to move the FAQ topic within Compose docs (Product Manual). It currently appears in the Compose Reference section.

compose/link-env-deprecated.md
thaJeztah: Good catch; should we remove the file and redirect it somewhere?

usha-mandya: I think that's fair. How about redirect to the Environment variables topic - https://deploy-preview-10717--docsdocker.netlify.app/compose/environment-variables/?

PR #10725, opened 2020-04-29
Update release note URLs

usha-mandya:
Updated the URLs to point to release note docs

### DIFF
--- a/release-notes/index.md
+++ b/release-notes/index.md
@@ -10,8 +10,9 @@ find release notes for each component in the product manuals section. We suggest
 that you regularly visit the release notes to learn about updates.
 
 - [Docker Engine](/engine/release-notes/)
-- [Docker Desktop for Mac](/docker-for-mac/) 
-- [Docker Desktop for Windows](/docker-for-windows/) 
-- [Docker Desktop Enterprise](/desktop/enterprise/)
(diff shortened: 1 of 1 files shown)


usha-mandya:
@thaJeztah PTAL

thaJeztah:
Ah, yes, my thinking with linking to the "index" page for these products (except for "engine", which didn't have an index at the time), was to keep the list simple (and if each product manual had the same "layout", it would be easy to find what you're looking for).

usha-mandya:
> Ah, yes, my thinking with linking to the "index" page for these products (except for "engine", which didn't have an index at the time), was to keep the list simple (and if each product manual had the same "layout", it would be easy to find what you're looking for).

One of the items I'll be working on in Q2 is to make our 'What's New' in Docker more visible. We could try a better approach for that. Agree that we don't have a uniform layout today. As this topic is about release notes, thought it makes sense to redirect users to the specific release notes page to find info about new features, updates, etc.

thaJeztah:
Yes, I think this is fine (it's only adding two new entries to the list for the "edge" versions)

Still not sure if we want to keep this page in the "TOC" altogether, but given that not all projects have the same structure, it's probably good to keep for now to help discoverability

PR #10728, opened 2020-04-30
Few corrections and suggestions in doc

ajay143444:
Removed is from line 80 as this is not required

### DIFF
--- a/engine/security/rootless.md
+++ b/engine/security/rootless.md
@@ -183,7 +183,7 @@ $ dockerd-rootless.sh --experimental --storage-driver vfs
 As Rootless mode is experimental, you need to run
 `dockerd-rootless.sh` with `--experimental`.
 You also need `--storage-driver vfs` unless using Ubuntu or Debian 10 kernel.
-You don't need to care these flags if you manage the daemon using systemd, as
+You don't need to care about these flags if you manage the daemon using systemd, as
 these flags are automatically added to the systemd unit file.
 
 Remarks about directory paths:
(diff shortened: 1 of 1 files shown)


engine/security/rootless.md
```diff
@@ -126,7 +126,7 @@ testuser::231072:65536
 
 ## Install
 
-The installation script is available at https://get.docker.com/rootless .
+The installation script is available at [here](https://get.docker.com/rootless).
```
thaJeztah, line 129: This would now read as "The installation script is available at here". Perhaps (given that this would likely be used on the command line, so I think we should not hide the actual URL). If we want the link to be clickable, I think we should also add a "target=blank", so that the script opens in a new tab/window instead of navigating the current page away from the documentation;

```suggestion
The installation script is available at [https://get.docker.com/rootless](https://get.docker.com/rootless){: target="_blank" class="_" }.
```

engine/security/rootless.md
```diff
@@ -155,7 +155,7 @@ export DOCKER_HOST=unix:///run/user/1001/docker.sock
 
 To install the binaries manually without using the installer, extract
 `docker-rootless-extras-<version>.tar.gz` along with `docker-<version>.tar.gz`:
-https://download.docker.com/linux/static/stable/x86_64/
+from [here](https://download.docker.com/linux/static/stable/x86_64/)
```
thaJeztah, line 158: Same here

```suggestion
from [https://download.docker.com/linux/static/stable/x86_64/](https://download.docker.com/linux/static/stable/x86_64/){: target="_blank" class="_" }
```

ajay143444:
Thank you for your suggestions. Implemented the same.

PR #10759, opened 2020-05-05
landing page: change style

jdrouet:
### Proposed changes

Update the landing page of the documentation according to the design

https://5eb16ad8f08678ca4b0cfed3--wip-docs.netlify.app/


_includes/body-landing.html
```diff
@@ -0,0 +1,863 @@
...
+              </h6>
+              <p>
+                Take a walkthrough that covers writing your first app, data
+                storage, networking, and swarms, and ends with your app running
+                on production servers in the cloud.
+              </p>
+              <p>
+                <a href="https://docs.docker.com/get-started/">Learn Docker</a>
+              </p>
+              <h6>FAQ's</h6>
```
usha-mandya, line 347: Oops! Looks like we missed it in the review. Could you update this to 'FAQs'?

_includes/body-landing.html
```diff
@@ -0,0 +1,863 @@
...
+                </a>
+              </p>
+              <p class="only-win">
+                <a href="https://docs.docker.com/docker-for-windows/faqs/#what-is-the-difference-between-the-stable-and-edge-versions-of-docker-desktop">
+                  What is the difference between the Stable and Edge versions of Docker Desktop?
+                </a>
+              </p>
+              <p class="only-win">
+                <a href="https://docs.docker.com/docker-for-windows/faqs/#how-do-i-connect-from-a-container-to-a-service-on-the-host">
+                  How do I connect from a container to a service
```
usha-mandya, line 365: Missing '?'

_includes/body-landing.html
```diff
@@ -0,0 +1,863 @@
...
+          <h5 class="title">What's new?</h5>
+          <p>
+            Learn about the cool new features, updates, and bug fixes.
+          </p>
+        </a>
+      </div>
+      <div class="col-xs-12 col-sm-6 col-lg-4 card-holder">
+        <a class="card" href="https://docs.docker.com/engine/">
+          <img src="/images/manuals.svg" alt="products manuals" />
+          <h5 class="title">Products manuals</h5>
```
usha-mandya, line 119: Could you update this to 'Product manuals'?

_includes/body-landing.html
```diff
@@ -0,0 +1,844 @@
...
+            src="../images/docker-docs-logo.svg"
+            alt="Docker Docs"
+            title="Docker Docs"
+          />
+        </a>
+      </div>
+      <div class="collapse navbar-collapse" id="header-content">
+        <ul class="nav navbar-nav">
+          <li class="active">
+            <a href="https://docs.docker.com/">Home</a>
```
thaJeztah, line 22: Can we change the links to not include the domain name?

```suggestion
            <a href="/">Home</a>
```

That way it's possible to test the links locally (or on preview domains)

_includes/body-landing.html
```diff
@@ -0,0 +1,844 @@
...
+      <div class="navbar-header">
+        <button type="button" class="navbar-toggle collapsed" data-toggle="collapse" data-target="#header-content" aria-expanded="false">
+          <span class="sr-only">Toggle navigation</span>
+          <span class="icon-bar"></span>
+          <span class="icon-bar"></span>
+          <span class="icon-bar"></span>
+        </button>
+        <a class="navbar-brand" href="/">
+          <img
+            src="../images/docker-docs-logo.svg"
```
thaJeztah, line 13: Might be safer to use an absolute path for this. Given that this isn an "include" it could potentially be included in different paths, in which case the link would break

(either that, or set the logo through CSS, but that's more of a "nice to have")

_includes/body-landing.html
```diff
@@ -0,0 +1,844 @@
...
+          spellcheck="false"
+          dir="auto"
+        />
+      </form>
+    </div>
+  </section>
+
+  <section class="container">
+    <div class="row">
+      <div class="col-xs-12 col-sm-6 col-lg-4 card-holder">
```
thaJeztah, line 75: After this is merged, we could have a look at separate includes for these tiles, then it'd be easier to "swap out" things in case we want to hightlight something else.

_includes/body-landing.html
```diff
@@ -0,0 +1,844 @@
...
+                Check out the following topics to learn how to build, run, and
+                deploy your applications using Docker.
+              </p>
+            </div>
+          </div>
+          <div class="row">
+            <div class="col-xs-12 col-md-6">
+              <a href="https://docs.docker.com/engine/reference/builder/"
+                >Write a Dockerfile</a
+              >
```
thaJeztah, line 154: these newlines look a bit awkward (some more of these below); perhaps unwrap these lines (they will be shorter after we remove `https://docs.docker.com` from the URL's)

_includes/body-landing.html
```diff
@@ -0,0 +1,844 @@
...
+            <div class="col-xs-12 col-md-6">
+              <h2 class="title">Stack Overflow</h2>
+              <p>
+                Browse through some of the frequent questions asked about Docker
+                on Stack Overflow.
+              </p>
+              <p>
+                <a href="https://stackoverflow.com/questions/24319662/from-inside-of-a-docker-container-how-do-i-connect-to-the-localhost-of-the-mach">
+                  From inside of a Docker container, how do I connect to the localhost of the machine?
+                </a>
```
thaJeztah, line 695: We need to make sure that links to external sites (as a default) have a `target=_blank`, because we don't want readers to navigate away from our documentation.

This could be hard-coded (add a `target=_blank`), or we could implement a small script to handle this, a-la https://stackoverflow.com/questions/12041935/how-to-automatically-add-target-blank-to-external-links-only

(instead of that script adding a `target=blank` attribute, we could (also/instead) have it add a css-class so that  we can decorate external links

_includes/body-landing.html
```diff
@@ -0,0 +1,844 @@
...
+    </div>
+    <div class="row">
+      <div class="col-xs-12 col-sm-6 col-md-4 card-holder">
+        <a href="https://events.docker.com/chapters/" class="card">
+          <h5 class="title">Find a local meetup</h5>
+          <p>Attend one of the 200+ Docker Meetups around the globe.</p>
+        </a>
+      </div>
+      <div class="col-xs-12 col-sm-6 col-md-4 card-holder">
+        <a href="https://www.docker.com/events" class="card">
```
thaJeztah, line 738: Same for these links; even though they're on "our" domain; they're navigating away from the docs, so should probably open in a new tab/window

_includes/head.html
```diff
@@ -0,0 +1,74 @@
...
+          display: block;
+      }
+  </style>
+  <!-- Google Tag Manager -->
+  <script>(function(w,d,s,l,i){w[l]=w[l]||[];w[l].push({'gtm.start':
+  new Date().getTime(),event:'gtm.js'});var f=d.getElementsByTagName(s)[0],
+  j=d.createElement(s),dl=l!='dataLayer'?'&l='+l:'';j.async=true;j.src=
+  'https://www.googletagmanager.com/gtm.js?id='+i+dl;f.parentNode.insertBefore(j,f);
+  })(window,document,'script','dataLayer','GTM-WL2QLG5');</script>
+  <!-- End Google Tag Manager -->
```
thaJeztah, line 25: Remind me that we need to make including these scripts "conditional"; we should set an `env-var` (or option) when building docs for publication, and not include these scripts when building the docs (e.g.) on Netlify and/or for local viewing.

_layouts/docs.html
```diff
@@ -14,269 +14,10 @@
...
-    <script>
-    jQuery(document).ready(function(){hookupTOCEvents();});
-    </script>
-</body>
+{% include head.html %}
+{% if page.landing == true %}
+    {% include body-landing.html %}
+{%else%}
+    {% include body.html %}
+{% endif %}
```
thaJeztah, line 22: Niceee 👍 


(was considering using a different `_layout` for the landing page instead of re-using the same for everything, but this looks clean)

_includes/head.html
```diff
@@ -0,0 +1,74 @@
...
+
+  <meta name="viewport" content="width=device-width, initial-scale=1.0">
+  <meta name="keywords" content="{% if page.keywords %}{{ page.keywords }}{% else %}docker, docker open source, docker platform, distributed applications, microservices, containers, docker containers, docker software, docker virtualization{% endif %}">
+  <link rel="stylesheet" href="/css/font-awesome.min.css">
+  <link rel="stylesheet" href="/css/bootstrap.min.css">
+  <link id="pygments" rel="stylesheet" href="/css/pygments/perldoc.css">
+  <link id="pagestyle" rel="stylesheet" href="/css/style.css">
+
+  <!-- Go get "Open Sans" font from Google -->
+  <link href="https://fonts.googleapis.com/css?family=Open+Sans" rel="stylesheet">
```
thaJeztah, line 55: Something to look into; we're fetching these styles and fonts directly from fonts.googleapis.com. Not sure if things improved, but my experience from way-back-when was that this was _slow_ (and I think they disabled caching so that metrics could be collected). Perhaps no longer the case, but if it is, we should look at hosting those fonts and styles locally (like we do for the other font)

_includes/body-landing.html
jdrouet: I was thinking about it while doing it, but some browsers consider `target=_blank` as spam.
But I'd prefer use hard-coded `target=_blank` than a script considering the actual size of the website.

_includes/body-landing.html
jdrouet: Yeah, my bad, my linter is used to jsx and behaves weirdly on html... I'll fix that, thank you for noticing ;)

_includes/body-landing.html
jdrouet: lgtm

_includes/body-landing.html
jdrouet: About the logo through CSS, considering we are using bootstrap, I'm following the examples as much as possible to avoid fighting the framework 🤷

thaJeztah: Yeah, the site / theme has grown to become a bit of a mish-mash; using CSS is definitely not an urgent thing (bigger mess to clean up than that)

_includes/body-landing.html
thaJeztah: The script would run on the current page only, but agreed, hard coding works

usha-mandya:
Thanks for taking a look @thaJeztah.

_includes/body-landing.html
```diff
@@ -0,0 +1,880 @@
...
+      </div>
+      <div class="col-xs-12 col-sm-6 col-lg-4 card-holder">
+        <a class="card" href="/get-docker/">
+          <img
+            src="/images/download-docker.svg"
+            alt="download and install docker"
+          />
+          <h5 class="title">Download and install</h5>
+          <p>
+            Download and install Docker on your machine in a few easy steps.
```
thaJeztah, line 99: @usha-mandya wondering if we considered adding direct links to the install pages for "DD mac", "DD windows", and "Linux" here, to provide a more fluent workflow.

(Perhaps we didn't add them to keep the cards "cleaner", just wondering)

_includes/body-landing.html
```diff
@@ -0,0 +1,880 @@
...
+            <a href="/develop/">Guides</a>
+          </li>
+          <li>
+            <a href="/engine/">Product Manual</a>
+          </li>
+          <li>
+            <a href="/reference/">Reference</a>
+          </li>
+          <li>
+            <a href="https://www.docker.com/blog/">Articles</a>
```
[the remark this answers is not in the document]
usha-mandya: Yup, anything that links to pages within docs shouldn't be opening in a new tab. Can we only have _blank for URLs that go to external pages?

_includes/body-landing.html
```diff
@@ -0,0 +1,880 @@
...
+        <div class="cardlet container-fluid">
+          <div class="row">
+            <div class="col-xs-12">
+              <h5 class="title">Most popular articles</h5>
+              <p>Check out the most popular articles on docs.docker.com.</p>
+            </div>
+          </div>
+          <div class="row">
+            <div class="col-xs-12 col-md-6 col-lg-12">
+              <a href="/engine/install/ubuntu/" target="_blank">
```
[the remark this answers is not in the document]
usha-mandya: Agree. Only target=_blank for external pages.

_includes/body-landing.html
usha-mandya: Yes, we considered the option to add specific links in each card in the initial mocks, but these were later replaced by the single cards we see now. We can revisit this in phase-2.

_includes/body-landing.html
```diff
@@ -0,0 +1,880 @@
...
+      <div class="collapse navbar-collapse" id="header-content">
+        <ul class="nav navbar-nav">
+          <li class="active">
+            <a href="/">Home</a>
+          </li>
+          <li>
+            <a href="/develop/">Guides</a>
+          </li>
+          <li>
+            <a href="/engine/">Product Manual</a>
```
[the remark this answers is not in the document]
usha-mandya: Merged 10798.

_includes/body-landing.html
```diff
@@ -0,0 +1,880 @@
...
+            <div class="col-xs-12 col-md-6">
+              <a href="/get-started/part2/" target="_blank">Build and run an image</a>
+            </div>
+            <div class="col-xs-12 col-md-6">
+              <a href="/config/labels-custom-metadata/" target="_blank">
+                Manage Docker objects
+              </a>
+            </div>
+            <div class="col-xs-12 col-md-6">
+              <a href="/develop/develop-images/multistage-build/" target="_blank">
```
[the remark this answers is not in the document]
usha-mandya: @jdrouet Not sure I missed a discussion on this. Only external links must open in a new tab.

_includes/body-landing.html
thaJeztah: :+1: SGTM

_includes/body-landing.html
```diff
@@ -0,0 +1,880 @@
...
+  <nav class="navbar header">
+    <div class="container">
+      <div class="navbar-header">
+        <button
+          type="button"
+          class="navbar-toggle collapsed"
+          data-toggle="collapse"
+          data-target="#header-content"
+          aria-expanded="false"
+        >
```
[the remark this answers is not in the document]
jdrouet: Is it? It looks more readable to me. In my opinion, considering the number of parameters, better put them on a new line... Maybe we should put a linter and not worry about that.

PR #10762, opened 2020-05-06
Replace links with relative paths

mecm1993:
### Proposed changes

Replaced broken links with relative paths to the `md` file. There is a shell script that will replace the relative path with the actual URL path accordingly. Therefore, the links should point to the `md` files. 

These changes were only applied in `/ee` folder. The same changes could be applied to the other folders. I would like your opinion or suggestions about it, before doing the following changes in the other folders.

### Related issues (optional)

 - #10748 
 - #10365 
 - There could be more


thaJeztah:
Docker Enterprise docs have been moved to Mirantis. If there's still an issue, could you report this over there? https://docs.mirantis.com

Issue #10770, opened 2020-05-07
Auto-restart of Docker rootless on boot requires an additional command


flavienbwk:
[engine/security/rootless.md](https://docs.docker.com/engine/security/rootless/)

Hi,

In `Usage > Daemon` : Docker rootless, in order to execute the `~/.config/systemd/user/docker.service` file need to be enabled by systemctl.

So on top of the `sudo loginctl enable-linger $(whoami)` command, the user has to execute :

```bash
systemctl --user enable docker
```

Then it works perfectly fine.

PR #10771, opened 2020-05-07
Auto-restart of Docker rootless on boot requires an additional command

flavienbwk:
### Proposed changes

In `Usage > Daemon` : Docker rootless, in order to execute the `~/.config/systemd/user/docker.service` file need to be enabled by systemctl.

So on top of the `sudo loginctl enable-linger $(whoami)` command, the user has to execute :

```bash
systemctl --user enable docker
```

Then it works perfectly fine.

### Related issues (optional)

Related to issue [#10770](https://github.com/docker/docker.github.io/issues/10770)

### DIFF
--- a/engine/security/rootless.md
+++ b/engine/security/rootless.md
@@ -167,9 +167,10 @@ Use `systemctl --user` to manage the lifecycle of the daemon:
 $ systemctl --user start docker
 ```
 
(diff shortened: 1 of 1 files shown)


AkihiroSuda:
cc @tonistiigi @tiborvass

Issue #10785, opened 2020-05-11
Add fish completion to doc, since it is available in docker


dmetzler1988:
[docker-for-mac/index.md](https://docs.docker.com/docker-for-mac/)

Hi,

can you please add fish-shell completion also to the bash-completion documentation since docker.app contains fish completions also?

```bash
mkdir ~/.config/fish/completions
set etc /Applications/Docker.app/Contents/Resources/etc
ln -s $etc/docker.fish-completion ~/.config/fish/completions/docker.fish
ln -s $etc/docker-compose.fish-completion ~/.config/fish/completions/docker-compose.fish
```

Thanks and best regards

PR #10786, opened 2020-05-11
Added fish-shell auto completion to docker for mac installation guide.

dmetzler1988:
### Proposed changes

I added fish-shell auto completion installation information for mac.

### Related issues (optional)

#10785

### DIFF
--- a/docker-for-mac/index.md
+++ b/docker-for-mac/index.md
@@ -470,18 +470,18 @@ class="_"}. To activate completion for Docker commands,
 these files need to be copied or symlinked to your Fish-shell `completions/`
 directory.
 
-Create at first the `completions` directory if not exist:
+Create the `completions` directory:
 
 ```bash
-mkdir ~/.config/fish/completions
+mkdir -p ~/.config/fish/completions
 ```
 
 Now add fish completions from docker.
 
 ```bash
 set etc /Applications/Docker.app/Contents/Resources/etc
-ln -s $etc/docker.fish-completion ~/.config/fish/completions/docker.fish
-ln -s $etc/docker-compose.fish-completion ~/.config/fish/completions/docker-compose.fish
+ln -shi /Applications/Docker.app/Contents/Resources/etc/docker.fish-completion ~/.config/fish/completions/docker.fish
+ln -shi /Applications/Docker.app/Contents/Resources/etc/docker-compose.fish-completion ~/.config/fish/completions/docker-compose.fish
 ```
 
 ## Give feedback and get help
--- a/docker-for-mac/index.md
+++ b/docker-for-mac/index.md
@@ -462,6 +462,28 @@ ln -s $etc/docker.zsh-completion /usr/local/share/zsh/site-functions/_docker
 ln -s $etc/docker-compose.zsh-completion /usr/local/share/zsh/site-functions/_docker-compose
 ```
 
+### Fish-Shell
+
+Fish-shell also supports tab completion [completion
+system](https://fishshell.com/docs/current/#tab-completion){:target="_blank"
+class="_"}. To activate completion for Docker commands,
+these files need to be copied or symlinked to your Fish-shell `completions/`
+directory.
+
+Create at first the `completions` directory if not exist:
+
+```bash
+mkdir ~/.config/fish/completions
+```
+
+Now add fish completions from docker.
+
+```bash
+set etc /Applications/Docker.app/Contents/Resources/etc
+ln -s $etc/docker.fish-completion ~/.config/fish/completions/docker.fish
+ln -s $etc/docker-compose.fish-completion ~/.config/fish/completions/docker-compose.fish
+```
+
 ## Give feedback and get help
 
 To get help from the community, review current user topics, join or start a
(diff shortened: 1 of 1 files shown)


usha-mandya:
@mat007 Could you PTAL?

mat007:
Hmm I don’t know much about fish, @thaJeztah maybe?

thaJeztah:
I seem to recall that the correct path for these is `/usr/local/share/fish/vendor_completions.d/` (see https://github.com/moby/moby/pull/17144). I'm wondering if that's something Docker Desktop could setup automatically (so that no manual steps are needed)?

dmetzler1988:
@thaJeztah as you can see in linked docu from fish-shell (is added to the md) and [here the direct link](https://fishshell.com/docs/current/#where-to-put-completions) you can see the paths what they recommend for it.   
If you think that it should be more globally, i can change the path :)

docker-for-mac/index.md
```diff
@@ -462,6 +462,28 @@ ln -s $etc/docker.zsh-completion /usr/local/share/zsh/site-functions/_docker
...
+Fish-shell also supports tab completion [completion
+system](https://fishshell.com/docs/current/#tab-completion){:target="_blank"
+class="_"}. To activate completion for Docker commands,
+these files need to be copied or symlinked to your Fish-shell `completions/`
+directory.
+
+Create at first the `completions` directory if not exist:
+
+```bash
+mkdir ~/.config/fish/completions
```
thaJeztah, line 476: Can you add `-p`, so that it will create the directory if it doesn't exist, but doesn't produce an error if it already did?

```suggestion
mkdir -p ~/.config/fish/completions
```

docker-for-mac/index.md
```diff
@@ -462,6 +462,28 @@ ln -s $etc/docker.zsh-completion /usr/local/share/zsh/site-functions/_docker
...
+```
+
+Now add fish completions from docker.
+
+```bash
+set etc /Applications/Docker.app/Contents/Resources/etc
+ln -s $etc/docker.fish-completion ~/.config/fish/completions/docker.fish
+ln -s $etc/docker-compose.fish-completion ~/.config/fish/completions/docker-compose.fish
```
thaJeztah, line 484: Bit on the fence if we should use the intermediate variable (as it will only be used once), or just inline the path in the commands;

```suggestion
ln -s /Applications/Docker.app/Contents/Resources/etc/docker.fish-completion ~/.config/fish/completions/docker.fish
ln -s /Applications/Docker.app/Contents/Resources/etc/docker-compose.fish-completion ~/.config/fish/completions/docker-compose.fish
```

Also thinking if we should add the `h` (don't follow symlink if the target is a symlink) and `i` (prompt for overwriting target if it exists) options; WDYT?

```suggestion
ln -shi /Applications/Docker.app/Contents/Resources/etc/docker.fish-completion ~/.config/fish/completions/docker.fish
ln -shi /Applications/Docker.app/Contents/Resources/etc/docker-compose.fish-completion ~/.config/fish/completions/docker-compose.fish
```

docker-for-mac/index.md
```diff
@@ -462,6 +462,28 @@ ln -s $etc/docker.zsh-completion /usr/local/share/zsh/site-functions/_docker
...
 
+### Fish-Shell
+
+Fish-shell also supports tab completion [completion
+system](https://fishshell.com/docs/current/#tab-completion){:target="_blank"
+class="_"}. To activate completion for Docker commands,
+these files need to be copied or symlinked to your Fish-shell `completions/`
+directory.
+
+Create at first the `completions` directory if not exist:
```
thaJeztah, line 473: "at first" looks a bit redundant here. With my suggestion below (to add the `-p` option), I think we can simplify the sentence, and remove the "if not exist";

```suggestion
Create the `completions` directory:
```

Alternatively, start with "first";

```suggestion
First, create the `completions` directory:
```

thaJeztah:
I briefly discussed with the Desktop team to check if it's possible to instal the symlinks by default when using Docker Desktop. They informed me that that should be possible, so perhaps in a future release we can remove these instructions.

Looking at the order in which Fish checks paths, I'm ok with using this path for now; effectively this path (`~/.config/....`) takes precedence over the "system wide" path, which means that users that manually created this symlink will continue using it even if Docker for Mac is updated to create a symlink out of the box.

dmetzler1988:
@thaJeztah thanks for checking and the review. The changes are done :)

docker-for-mac/index.md
```diff
@@ -462,6 +462,28 @@ ln -s $etc/docker.zsh-completion /usr/local/share/zsh/site-functions/_docker
...
+Create the `completions` directory:
+
+```bash
+mkdir -p ~/.config/fish/completions
+```
+
+Now add fish completions from docker.
+
+```bash
+set etc /Applications/Docker.app/Contents/Resources/etc
```
thaJeztah, line 482: This line can be removed now

```suggestion
```

dmetzler1988: Oh sorry, yes.

PR #10798, opened 2020-05-13
Remove "product manuals landingpage" placeholder

thaJeztah:
This removes the placeholder in favor of a redirect to the first product manual in the section, so that `/manuals/` redirects users to a sensible location.


thaJeztah:
@usha-mandya PTAL

PR #10801, opened 2020-05-13
WSL 2: add note about inotify

djs55:
### Proposed changes

Added a note about inotify to the WSL 2 best-practices guide..

It's important for both
- high performance; and
- inotify
to store files in the WSL Linux distro and bind mount them from there rather than remote them from the Windows host.

This applies to Docker Desktop with WSL 2


docker-for-windows/wsl.md
```diff
@@ -84,7 +84,7 @@ The following section describes how to start developing your applications using
 
 ## Best practices
 
-- To get the best out of the file system performance, we recommend that you avoid mounting from the Windows file system (even on a WSL distro). For example, avoid `docker run -v /mnt/c/users:/users`. Instead, we recommend that you mount files from your distro's local file system, for example, `docker run -v ~/my-project:/sources <my-image>`.
+- We recommend storing source code and other data that will be bind-mounted into Linux containers (e.g. with `docker run -v`) in the Linux filesystem, rather than the Windows filesystem. Linux containers will only receive file change events ("inotify events") if the original files are stored in the Linux filesystem. Performance will also be much higher if the files are bind-mounted from Linux, rather than remoted from the Windows host. Therefore avoid `docker run -v /mnt/c/users:/users` (where `/mnt/c` is mounted from Windows). Instead, from a Linux shell use a command like `docker run -v ~/my-project:/sources <my-image>` where `~` is expanded by the Linux shell to `$HOME`.
```
thaJeztah, line 87: Contents looks good to me; have some nits;

- If possible, wrap to ~80 chars
- Avoid royal "We" (perhaps take the original approach? "To get the best .."
- Avoid future tense in tech docs
- I think we standardise on using a comma after `e.g.`

Wondering if we should break it up into sub-bullets, something along the line of;

> - To get the best out of the file system performance when bind-mounting files:
>     - Store source code and other data that is bind-mounted into Linux containers
>       (i.e., with `docker run -v <host-path>:<container-path>`) in the Linux
>       filesystem, rather than the Windows filesystem.
>     - Linux containers only receive file change events ("inotify events") if the
>       original files are stored in the Linux filesystem.
>     - Performance is much higher when files are bind-mounted from the Linux
>       filesystem, rather than remoted from the Windows host. Therefore avoid
>       `docker run -v /mnt/c/users:/users` (where `/mnt/c` is mounted from Windows).
>     - Instead, from a Linux shell use a command like `docker run -v ~/my-project:/sources <my-image>`
>       where `~` is expanded by the Linux shell to `$HOME`.


```suggestion
- To get the best out of the file system performance when bind-mounting files:
    - Store source code and other data that is bind-mounted into Linux containers
      (i.e., with `docker run -v <host-path>:<container-path>`) in the Linux
      filesystem, rather than the Windows filesystem.
    - Linux containers only receive file change events ("inotify events") if the
      original files are stored in the Linux filesystem.
    - Performance is much higher when files are bind-mounted from the Linux
      filesystem, rather than remoted from the Windows host. Therefore avoid
      `docker run -v /mnt/c/users:/users` (where `/mnt/c` is mounted from Windows).
    - Instead, from a Linux shell use a command like `docker run -v ~/my-project:/sources <my-image>`
      where `~` is expanded by the Linux shell to `$HOME`.
```

thaJeztah:
Thanks!

PR #10805, opened 2020-05-13
Removed quotes from add-apt-repository

m4t5:
The quotes in the add-apt-repository stopped it working on Linux Mint 19.2 (based on Ubuntu Bionic).

I removed the quotes for amd-64; other architectures may also need them removing.

### DIFF
--- a/engine/install/ubuntu.md
+++ b/engine/install/ubuntu.md
@@ -136,9 +136,9 @@ from the repository.
 
     ```bash
     $ sudo add-apt-repository \
-       "deb [arch=amd64] {{ download-url-base }} \
+       deb [arch=amd64] {{ download-url-base }} \
        $(lsb_release -cs) \
-       stable"
+       stable
     ```
 
     </div>


tianon:
In this case, the quotes are correct -- according to https://manpages.ubuntu.com/manpages/focal/en/man1/add-apt-repository.1.html, the `REPOSITORY` argument is either `ppa:<user>/<ppa-name>` or a raw line to be added to `sources.list`, which is what we've got here (since this isn't a launchpad.net-hosted PPA, but rather a full repo).

IMO (I am Not A Maintainer in this repo, so take this for what it's worth), `add-apt-repository` itself isn't adding any value here and we'd be better off doing something simpler like:

```bash
echo "deb ... stable" | sudo tee /etc/apt/sources.list.d/docker.list
```

tianon:
To illustrate further:

```console
$ docker pull ubuntu:bionic
bionic: Pulling from library/ubuntu
Digest: sha256:3235326357dfb65f1781dbc4df3b834546d8bf914e82cce58e6e6b676e23ce8f
Status: Image is up to date for ubuntu:bionic
docker.io/library/ubuntu:bionic

$ docker run -it --rm ubuntu:bionic
root@874b03285550:/# apt-get update -qq 
root@874b03285550:/# apt-get install -yqq software-properties-common
...
root@874b03285550:/# add-apt-repository 'deb [arch=amd64] https://download.docker.com/linux/ubuntu bionic stable'
...
root@874b03285550:/# grep docker /etc/apt/sources.list
deb [arch=amd64] https://download.docker.com/linux/ubuntu bionic stable
# deb-src [arch=amd64] https://download.docker.com/linux/ubuntu bionic stable
```

m4t5:
Hi Tianon,

Thanks for taking the time to review this - and I appreciate (and learned from) the extra illustration. 

I think (and I appreciate this is probably out of scope) that there may be a problem with add-apt-repository on Mint 19.

I tried it with linuxmintd/mint19-amd64

$ docker pull linuxmintd/mint19-amd64
...
$ docker run -it --rm  linuxmintd/mint19-amd64
root@0850392cff0d:/# apt-get update -qq 
root@0850392cff0d:/# apt-get install -yqq software-properties-common
debconf: delaying package configuration, since apt-utils is not installed
root@04e14c6190ef:/# apt install apt-utils
...
root@be24ede5fe0a:/# add-apt-repository 'deb [arch=amd64] https://download.docker.com/linux/ubuntu bionic stable'
Traceback (most recent call last):
  File "/usr/lib/linuxmint/mintSources/mintSources.py", line 25, in <module>
    import gi
ImportError: No module named gi

In conclusion, I think directly editing /etc/apt/sources.list is much more reliable and I'll be doing that in future.

Thanks again.

Mat

PR #10815, opened 2020-05-15
add home button to every page

jdrouet:
### Proposed changes

Add missing home button to non home pages

Screenshot 2020-05-15 at 09 55 03

### DIFF
--- a/_data/toc.yaml
+++ b/_data/toc.yaml
@@ -1,5 +1,8 @@
 
 horizontalnav:
+- title: Home
+  path: /
+  node: home
 - title: Guides
   path: /develop/
   node: guides
@@ -12,6 +15,8 @@ horizontalnav:
 - title: Samples
   path: /samples/
   node: samples
+
+home: {}
 
 guides:
 - title: Get Docker


usha-mandya:
Thanks Jeremie. I can see the Home nav on other tabs now.

The search box doesn't load when I click on other tabs. It works fine in prod. Sometimes the browser needs to be refreshed for the search box appear. Not sure if it's just me?

jdrouet:
@usha-mandya I have the same weird behavior about the search bar. I'll try to figure out why. Do you have any scenario in mind where it should not appear?

usha-mandya:
@jdrouet Just noticed that this issue doesn't occur in https://github.com/docker/docker.github.io/pull/10815.

thaJeztah:
Also make sure to check on mobile to see if the top-navigation doesn't wrap and cover the left-hand nav (I recall we ran into issues if there were too many items in the top-nav)

jdrouet:
@thaJeztah it seems ok on a simulated iphone. On my android, it works fine also.

Screen Shot 2020-05-15 at 13 44 43

jdrouet:
It seems like this search input missing is due to a race condition. I'll try to fix it in another PR.

usha-mandya:
OK, that's good to know, Jeremie. Can I merge this one then?

jdrouet:
LGTM (although it's my own PR...)

thaJeztah:
@jdrouet if the issue with the search is only in this PR, I think we should ideally fix it here.

Thinking now: could it be that a script looks for the "Guides" link specifically, and now that it moved, it no longer finds it?

jdrouet:
It's not only in this PR. It does the same on master.
Sometimes, this https://github.com/docker/docker.github.io/blob/master/js/docs.js#L347 is called after this https://github.com/docker/docker.github.io/blob/master/js/archive.js#L52 being called and that's why it doesn't display. Now, it's time to rewrite this to be async.

thaJeztah:
> It's not only in this PR. It does the same on master.
Sometimes, this https://github.com/docker/docker.github.io/blob/master/js/docs.js#L347 is called after this https://github.com/docker/docker.github.io/blob/master/js/archive.js#L52 being called and that's why it doesn't display. Now, it's time to rewrite this to be async.

ugh 😞 

Perhaps we should put a css `archive` class on the main `<html>` or `<body>` element that we can remove, instead of trying to un-hide each element.

(we should probably also just exclude the whole search if we're doing an archive build, but that's likely more work)

thaJeztah:
Quick try on my phone;

61245C24-4DE5-492D-8EA3-631157E88787
A715062B-F19E-470F-95C1-003951228FA0
DCBFC292-E705-4384-91F9-E63EB6A5E12C

thaJeztah:
Looks like the "samples" item wraps (but only visible if you expand the "hamburger menu").

The top-navigation is also hiding the first item in the TOC ("cli reference"), which we should fix (but not introduced in this PR)

usha-mandya:
> Perhaps we should put a css `archive` class on the main `<html>` or `<body>` element that we can remove, instead of trying to un-hide each element.
>
> (we should probably also just exclude the whole search if we're doing an archive build, but that's likely more work)

Is there any way we can fix the search bar for the landing page without rewriting archives.js?

@thaJeztah How does this impact the updates given that we'll be removing archives soon from our docs site?

thaJeztah:
> How does this impact the updates given that we'll be removing archives soon from our docs site?

We should no longer need that code once we no longer host archives (I think it's only to add the "warning this is an archive" banner). Actually we could potentially even remove it already, if we won't be publishing the current version of the docs as an archive. (Existing archives still have that could, so would be showing the banner)

But we _should_ keep generating (and hosting) the https://docs.docker.com/js/archives.json file for now, because that's used by the archives to detect if they are an archive.

usha-mandya:
The search bar works fine for me now. Tried on a different browser and that works fine too.

jdrouet:
Could we merge this and fix the search bar on another PR ?

PR #10817, opened 2020-05-15
landing page: set focus on search field

thaJeztah:
Also updates the CSS to reduce the "focus highlight" applied by browsers.

@usha-mandya @jdrouet we were discussing this option, so thought I'd open a quick PR for it to try "what it feels like". Feel free to close if this is not what we want 👍

### DIFF
--- a/_scss/_landing.scss
+++ b/_scss/_landing.scss
@@ -105,6 +105,11 @@ body#landing {
       height: 50px;
       line-height: 24px;
       padding: 13px 20px;
+
+      &:focus {
+        box-shadow: none;
+        outline: none;
+      }
     }
   }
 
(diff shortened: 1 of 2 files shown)


_scss/_landing.scss
```diff
@@ -105,6 +105,11 @@ body#landing {
       height: 50px;
       line-height: 24px;
       padding: 13px 20px;
+
+      &:focus {
+        box-shadow: none;
+        outline: none;
+      }
```
thaJeztah, line 112: Looks like this doesn't _fully_ hide the "focus" (border still turns "blue" on the "light" version of the website); didn't search if there's other options to fully disable it nowadays.

usha-mandya: We don't have to complete hide the focus. The current version looks good.

thaJeztah: Yeah, I wanted to at least get rid of the "glow", as it distracted from the overall layout on the page

PR #10819, opened 2020-05-15
Add note on Desktop installer size reduction

usha-mandya:
Added a note in Desktop 2.3.0.2 release notes about the reduction in the Desktop installer size.

### DIFF
--- a/docker-for-mac/release-notes.md
+++ b/docker-for-mac/release-notes.md
@@ -31,6 +31,7 @@ Docker Desktop introduces a new onboarding tutorial upon first startup. The Quic
 
 ### Bug fixes and minor changes
 
+- Reduced the size of the Docker Desktop installer from 708 MB to 456 MB.
 - Fixed bug where containers disappeared from the UI when Kubernetes context is invalid. Fixes [docker/for-win#6037](https://github.com/docker/for-win/issues/6037).
 - Fixed a file descriptor leak in `vpnkit-bridge`. Fixes [docker/for-win#5841](https://github.com/docker/for-win/issues/5841).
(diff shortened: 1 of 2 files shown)


usha-mandya:
> do we need a mention that some components are now downloaded on demand?

Not sure. @stephen-turner Should we mention this?

docker-for-mac/release-notes.md
```diff
@@ -31,6 +31,7 @@ Docker Desktop introduces a new onboarding tutorial upon first startup. The Quic
 
 ### Bug fixes and minor changes
 
+- Reduced the size of the Docker Desktop installer from 710 MB to 445 MB.
```
stephen-turner, line 34: 708 -> 456

docker-for-windows/release-notes.md
```diff
@@ -54,6 +54,7 @@ For information about Edge releases, see the [Edge release notes](edge-release-n
 
 **Other fixes**
 
+- Reduced the size of the Docker Desktop installer from 960 MB to 400 MB.
```
stephen-turner, line 57: 960 -> 409

stephen-turner:
> > do we need a mention that some components are now downloaded on demand?
> 
> Not sure. @stephen-turner Should we mention this?

I don't think we need to mention that.

docker-for-mac/release-notes.md
usha-mandya: Updated.

docker-for-windows/release-notes.md
usha-mandya: Updated.

PR #10826, opened 2020-05-18
Update note on Win Server support

usha-mandya:
Update the note in Desktop docs to clarify that Windows Server isn't supported on Docker Desktop.

### DIFF
--- a/docker-for-windows/faqs.md
+++ b/docker-for-windows/faqs.md
@@ -117,9 +117,14 @@ Yes, you can run VirtualBox along with Docker Desktop if you have enabled the [ 
 
 ## Windows requirements
 
-### How do I run Windows containers on Docker Desktop on Windows Server?
+### Can I run Docker Desktop on Windows Server?
 
-You can find a tutorial about running Windows containers on Windows Server in
+No, running Docker Desktop on Windows Server is not supported.
+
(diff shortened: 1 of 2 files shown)


docker-for-windows/faqs.md
```diff
@@ -117,9 +117,14 @@ Yes, you can run VirtualBox along with Docker Desktop if you have enabled the [
 
 ## Windows requirements
 
+### Can I run Docker Desktop on Windows Server?
+
+No, running Docker Desktop on Windows Server is not supported.
+
 ### How do I run Windows containers on Docker Desktop on Windows Server?
```
mat007, line 124: Do we want to change this as well?

usha-mandya: Could you provide more info? Do you mean updating the heading or combining both questions?

mat007: Oh sorry, I meant the heading looks wrong as it has Docker Desktop in it, maybe it should just be `How do I run Windows containers on Windows Server?`?

usha-mandya: Aah, yes. That makes sense.

usha-mandya: Updated

PR #10837, opened 2020-05-19
Landing page touch-ups

thaJeztah:
Some things I was playing with in the weekend for fun, so thought I'd push them in case we think these look good 🤷‍♂️ 

### scss: various small fixes and linting issues

- Add generic font classes where missing (except for FontAwesome cfonts)
- Combined some rules with their shorthand form
- Removed unneeded units (0px -> 0)
- Fix some invalid values
- Commented out empty styles in _github.scss

### Landing page: apply icons using CSS, and remove duplication

This patch:

- removes the icon images from the landing page, instead setting
  them as a background-image through css.
- replaced the "community-resources" and "play-with-docker" styles
  in favor of a conditional style (based on cards linking to external
  websites)
- changes the minimal height of cards for mobile devices (there was
  a big amount of padding on the smallest size, which didn't add
  much value).

Looks of the "community resources" cards before and after:

community-card-before
community-card-after

### Landing page: left-align icon on small screen sizes

Reduce the vertical space on small devices by putting the icon at the left of the text, instead of above.

Before and after:

icon-card-before
icon-card-after
 

### Landing page: reduce tab-width to prevent wrapping

The tabs on the "help by project" card were wrapping on iPhone, so reducing the padding to prevent this.

tabs-before
tabs-after

### DIFF
--- a/_scss/_github.scss
+++ b/_scss/_github.scss
@@ -24,13 +24,15 @@ github.com style (c) Vasily Polovnyov <vast@whiteants.net>
   font-weight: bold;
 }
 
+/*
 .hljs-number,
 .hljs-literal,
 .hljs-variable,
 .hljs-template-variable,
 .hljs-tag .hljs-attr {
-  // color: #008080;
+   color: #008080;
 }
+*/
 
 .hljs-string,
 .hljs-doctag {
@@ -61,33 +63,37 @@ github.com style (c) Vasily Polovnyov <vast@whiteants.net>
   font-weight: normal;
 }
 
+/*
 .hljs-regexp,
 .hljs-link {
-  // color: #009926;
+   color: #009926;
 }
 
 .hljs-symbol,
 .hljs-bullet {
-  // color: #990073;
+   color: #990073;
 }
 
 .hljs-built_in,
 .hljs-builtin-name {
-  // color: #0086b3;
+   color: #0086b3;
 }
+*/
 
 .hljs-meta {
   // color: #999;
   font-weight: bold;
 }
 
+/*
 .hljs-deletion {
-  // background: #fdd;
+   background: #fdd;
 }
 
 .hljs-addition {
-  // background: #dfd;
+   background: #dfd;
 }
+*/
 
 .hljs-emphasis {
   font-style: italic;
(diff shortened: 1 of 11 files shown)


jdrouet:
Looks like it introduced a problem on ipad size screen

Screen Shot 2020-05-19 at 11 12 00

thaJeztah:
Thx for checking; looks like I need to up the min-height a bit for that fold 👍

thaJeztah:
Updated; should be fixed now (pls check)

Also noticed that the tabs in the "help by product" section were wrapped; pushed a commit to reduce padding on small sizes

_scss/_github.scss
```diff
@@ -24,13 +24,15 @@ github.com style (c) Vasily Polovnyov <vast@whiteants.net>
   font-weight: bold;
 }
 
+/*
```
jdrouet, line 27: I found lot of commented code in this codebase. Maybe we should get rid of it

thaJeztah: Yes, I generally would clean up these. The reason I left them is that it looks like these were "overriding" the defaults in an "upstream" stylesheet. Keeping them commented makes it slightly easier to see what rules we removed/modified.

That said; I want to have another look at https://github.com/docker/docker.github.io/pull/10288 to remove highlightjs (and related styles) altogether, so that we only have a single location to style code-blocks / samples.

rgardziel:
Please leave the arrows on the cards that point to external resources outside of docs.docker. It is an interaction design best practice to signal to the user that an action would point them away from where they currently are.

thaJeztah:
@rgardziel that's how I interpreted the "arrows" (indicator that it's an external resource), and that's what this PR currently does. The "inconsistency" is that before this PR, _all_ cards under the "community-resources" and "play-with-docker" headings would have an arrow, but there's two cards in those sections that are not pointing to external sources, but to pages on docs.docker.com itself, and those no longer have an arrow with this PR.

Here's the "before" and "after" of those sections; the last card in both sections doesn't have an arrow, because it's linking to https://docs.docker.com/get-started/resources/

Before this PR

Screenshot 2020-05-20 at 16 59 38

After this PR

Screenshot 2020-05-20 at 17 17 49

thaJeztah:
(rebased the PR to get the latest changes from master in)

rgardziel:
ah, I missed that. Great catch.

PR #10848, opened 2020-05-20
Move Google Analytics to include, and prepare for making it optional

thaJeztah:
@usha-mandya @jdrouet PTAL

### DIFF
--- a/_includes/analytics/google_analytics.html
+++ b/_includes/analytics/google_analytics.html
@@ -1,0 +1,7 @@
+{% if include.GOOGLE_ID and include.GOOGLE_ID != '' %}<!-- Google Tag Manager -->
+  <script>(function(w,d,s,l,i){w[l]=w[l]||[];w[l].push({'gtm.start':
+  new Date().getTime(),event:'gtm.js'});var f=d.getElementsByTagName(s)[0],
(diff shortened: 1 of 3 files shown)


_includes/analytics/google_analytics.html
```diff
@@ -0,0 +1,7 @@
+{% if include.GOOGLE_ID and include.GOOGLE_ID != '' %}<!-- Google Tag Manager -->
+  <script>(function(w,d,s,l,i){w[l]=w[l]||[];w[l].push({'gtm.start':
+  new Date().getTime(),event:'gtm.js'});var f=d.getElementsByTagName(s)[0],
+  j=d.createElement(s),dl=l!='dataLayer'?'&l='+l:'';j.async=true;j.src=
+  'https://www.googletagmanager.com/gtm.js?id='+i+dl;f.parentNode.insertBefore(j,f);
+  })(window,document,'script','dataLayer','{{ include.GOOGLE_ID }}');</script>
```
thaJeztah, line 6: Kept the original formatting, but considering making this script a single-line (not wrapping it), so that it's less clutter on the page

thaJeztah:
FWIW, I noticed we don't have the "noscript" variant of GTM included in our pages (https://developers.google.com/tag-manager/quickstart); do we want to track users that have javascript disables? (probably not a large number, given that the site doesn't perform well with it disabled)

PR #10856, opened 2020-05-21
Add new includes file for Mouseflow

usha-mandya:
Adding a new includes file for Mouseflow

### DIFF
--- a/_includes/analytics/mouseflow.html
+++ b/_includes/analytics/mouseflow.html
@@ -1,0 +1,9 @@
+<script type="text/javascript">
+  window._mfq = window._mfq || [];
+  (function() {
+  var mf = document.createElement("script");
+  mf.type = "text/javascript"; mf.defer = true;
+  mf.src = "//cdn.mouseflow.com/projects/e9e8497f-ca8a-426e-b0a1-a78247e85aa4.js";
+  document.getElementsByTagName("head")[0].appendChild(mf);
+  })();
+</script>


_includes/analytics/mouseflow.html
```diff
@@ -0,0 +1,9 @@
+<script type="text/javascript">
+  window._mfq = window._mfq || [];
+  (function() {
+  var mf = document.createElement("script");
+  mf.type = "text/javascript"; mf.defer = true;
+  mf.src = "//cdn.mouseflow.com/projects/e9e8497f-ca8a-426e-b0a1-a78247e85aa4.js";
+  document.getElementsByTagName("head")[0].appendChild(mf);
+  })();
+</script>
```
thaJeztah, line 9: `type="text/javascript"` is redundant nowadays.

To make this script only include if we pass an ID, this could look something like

```suggestion
{% if include.MOUSEFLOW_ID and include.MOUSEFLOW_ID != '' %}
<script>
  window._mfq = window._mfq || [];
  (function() {
  var mf = document.createElement("script");
  mf.type = "text/javascript"; mf.defer = true;
  mf.src = "//cdn.mouseflow.com/projects/e9e8497f-ca8a-426e-b0a1-a78247e85aa4.js";
  document.getElementsByTagName("head")[0].appendChild(mf);
  })();
</script>
{% else %}<!-- No Mouseflow ID configured, Mouseflow is disabled -->{% endif %}
```

Then, in our `_config.yml` we can add a `mouseflow_id` variable, and add an `include` line on our landing page.


Let me make those changes and push it to a branch on my fork

thaJeztah:
@usha-mandya pushed a branch here; https://github.com/docker/docker.github.io/compare/master...thaJeztah:mouseflow_patch_suggestions?expand=1


I'm still wondering if we can handle this through Google Tag Manager though (as I think it should be possible to insert such scripts through GTM, which would be a lot more flexible)

usha-mandya:
@thaJeztah the Mouseflow tag has been set up on the main account. We should now add the GTM ID to the `_config.yml`  files.  Would you be able to update your branch https://github.com/docker/docker.github.io/compare/master...thaJeztah:mouseflow_patch_suggestions?expand=1?

Guess we don't need the `_includes/analytics/mouseflow.html` file? I can close this PR if we don't need an additional snippet. Thanks.

thaJeztah:
(just got off a call with @usha-mandya, and we managed to set up GTM to add this script, so we can likely close this PR 🎉)

PR #10858, opened 2020-05-21
Fix broken link to ucp install cmd docs

### LINKED ISSUE #10857: Link in UCP install page leads to 404 page
### Problem description
There is a link on this page https://docs.docker.com/ee/ucp/admin/install/ which leads to a 404 page.
The link is part of this excerpt:
>This runs the install command in interactive mode, so that you’re prompted for any necessary configuration values. To find what other options are available in the install command, including how to install UCP on a system with SELinux enabled, check the reference documentation.

The link point to https://docs.docker.com/reference/ucp/3.2/cli/install.md which shows the 404 page.

Deleting `install.md` shows a valid page - https://docs.docker.com/reference/ucp/3.2/cli/
but is not exactly what I need.

### Problem location

- I saw a problem on the following URL: https://docs.docker.com/ee/ucp/admin/install/

### Suggestions for a fix

- Update the link to point to a valid page like https://docs.docker.com/reference/ucp/3.2/cli/ or another `install` specific one if there is such.

- Add the `install.md` file wherever is needed, so that https://docs.docker.com/reference/ucp/3.2/cli/install.md becomes a valid page.

asankov:
The link in question, currently leads to https://docs.docker.com/reference/ucp/3.2/cli/install.md
which shows the 404 page.

The actual docs are located at https://docs.docker.com/reference/ucp/3.2/cli/install/

Fixes #10857

### Proposed changes

This PR replaces the wrong link with the right one

### DIFF
--- a/ee/ucp/admin/install/index.md
+++ b/ee/ucp/admin/install/index.md
@@ -83,7 +83,7 @@ To install UCP:
     This runs the install command in interactive mode, so that you're prompted
     for any necessary configuration values.  To find what other options are
     available in the install command, including how to install UCP on a system
-    with SELinux enabled, check the [reference documentation](/reference/ucp/3.2/cli/install.md).
+    with SELinux enabled, check the [reference documentation](/reference/ucp/3.2/cli/install/).
 
 > Note
 >


usha-mandya:

PR #10861, opened 2020-05-21
Add note on re-creating containers

usha-mandya:
Add a note on deleting and re-creating containers when after completing the caching process.


docker-for-mac/mutagen-caching.md
```diff
@@ -14,6 +14,10 @@ a continuous two-way sync of files between the host and containers using
 
 This page contains an example to show how the Mutagen feature should be used to sync files between the host and containers. It also gives some best practices to maximize performance.
 
+> **Important**
+>
+> After completing the caching process, make sure you delete and re-create your containers which were using the cached directories.
```
nebuk89, line 19: which you would like to make use of the cached directories

usha-mandya: I've made some slight changes to the suggested wording. PTAL.

PR #10870, opened 2020-05-22
Update install.md

ScottFred:
The install instructions are missing a critical step that requires the user to be added to the "docker-users" group.  Without this, the user will be prevented from running Docker containers.

### DIFF
--- a/docker-for-windows/install.md
+++ b/docker-for-windows/install.md
@@ -77,7 +77,10 @@ Looking for information on using Windows containers?
 
 4. When the installation is successful, click **Close** to complete the installation process.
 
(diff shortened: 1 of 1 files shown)


StefanScherer:
@ScottFred thank you for the hint. The Docker Desktop installer normally adds the user to the local docker-users group. But there might be situations when it doesn't work as intended.

Could you tell me more about your user accounts? Do you use eg. a different user account when the UAC popup appears?

usha-mandya:
@StefanScherer Modified the instructions. PTAL

StefanScherer:
@ScottFred I could reproduce the problem with two different accounts. Indeed, in that case you need to add your user account to docker-users group.

Thanks @usha-mandya for updating the docs.

PR #10872, opened 2020-05-22
fix description for ptrace in seccomp.md

JanKanis:
### Proposed changes

Fix the description for `ptrace` in `engine/security/seccomp.md`

The existing description leads the reader to believe that dropping CAP_SYS_PTRACE already blocks all ptraces. That is not true, and could give the reader a false view on this aspect of docker security. Dropping CAP_SYS_PTRACE only blocks ptracing arbitrary processes, ptracing child processes is still allowed.


usha-mandya:
The updated version LGTM. @thaJeztah Could you PTAL?

PR #10889, opened 2020-05-26
Updated to use new syntax.

gaurish:
### Proposed changes
Updated to use new syntax. without this,  getting the following error:

```
ERROR: The Compose file './docker-compose.yml' is invalid because:
Unsupported config option for services.redis: 'restart_policy'

```

```
$ docker-compose --version 
docker-compose version 1.25.5, build 8a1c60f6
```

### DIFF
--- a/compose/compose-file/index.md
+++ b/compose/compose-file/index.md
@@ -48,8 +48,7 @@ services:
       update_config:
         parallelism: 2
         delay: 10s
-      restart_policy:
-        condition: on-failure
+      restart: on-failure
 
   db:
     image: postgres:9.4


usha-mandya:
@ulyssessouza Could you PTAL?

thaJeztah:
What `version` did you specify in the compose-file? I think `restart:` is the [v2 syntax](https://docs.docker.com/compose/compose-file/compose-file-v2/#restart), and `restart_policy` for v3; https://docs.docker.com/compose/compose-file/#restart_policy

gaurish:
@thaJeztah I mentioned version: "3" in the compose-file.

thaJeztah:
`version: "3"` is equivalent to `3.0` (not "latest 3.x"), perhaps that explains it (it may have been added to a later version of the 3.x schema

gtardif:
Hi @gaurish, `restart` is a service-level field, and `restart-policy` is a part of the service-level `deploy` section (see https://docs.docker.com/compose/compose-file/#restart_policy). 
Did you try to use `restart_policy` at the service level ?

gtardif:
Closing as it seems this is a misuse of fields at the wrong level in compose file. Feel free to re-open if this is not the case.

PR #10890, opened 2020-05-26
Correct proxies documentation for Docker Desktop (windows)

mat007:
### Proposed changes

The proxies set in the Docker Desktop settings are actually not propagated in the containers automatically.
The rationale being that we want the same UX on Windows/Mac and as on Linux.

@usha-mandya once this has been merged I’ll do the same for the Mac docs.
@simonferquel can you take a quick look?

### Related issues (optional)

Closes https://github.com/docker/for-win/issues/3676


usha-mandya:
Made some minor updates. Looks good otherwise.

mat007:
Thanks, I’ll do the same for the Mac part of the docs!

PR #10902, opened 2020-05-27
Update dotnetcore.md

avichalchum:
Adding an example for .Net Core docker files as consumed by build agents and recommended by Microsoft best practices.

### Proposed changes

Changing .Net Core example documentation to add an example that includes best practices by Microsoft for CI builds.

### DIFF
--- a/engine/examples/dotnetcore.md
+++ b/engine/examples/dotnetcore.md
@@ -70,7 +70,7 @@ ENTRYPOINT ["dotnet", "aspnetapp.dll"]
 bin\
 obj\
 ```
-### Method 2 (CI recommended method):
+### Method 2 (build app outside Docker container):
 
 1.  Create a `Dockerfile` in your project folder.
 2.  Add the text below to your `Dockerfile` for either Linux or [Windows
--- a/engine/examples/dotnetcore.md
+++ b/engine/examples/dotnetcore.md
@@ -81,16 +81,16 @@ obj\
     [switching containers](../../docker-for-windows/index.md#switch-between-windows-and-linux-containers).
 3.  The `Dockerfile` assumes that your application is called `aspnetapp`. Change the `Dockerfile` to use the DLL file of your project. This method assumes that your project is already built and it copies the build artifacts from the publish folder. Refer to the Microsoft documentation on [Containerize a .Net Core app](https://docs.microsoft.com/en-us/dotnet/core/docker/build-container?tabs=windows#create-the-dockerfile){: target="blank" rel="noopener" class=“"}.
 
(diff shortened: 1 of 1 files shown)


gtardif:
Hi @avichalchum, 

Thanks for your proposal. The Dockerfile you propose is indeed simpler, but it assumes the application is already build by some external tool. 
For CI we definitely recommend to use the Dockerfile including [build steps](https://github.com/docker/docker.github.io/pull/10902/files#diff-ca8e05c3094b3abb0a6f3592cc8b676d801a676824944050085671f99bb65b1aR56) , so your build does not require additional tools to be installed, and is more easily reproduced. 

The simpler Dockerfile you propose could make the dev feedback loop quicker, but if you want to use this after you build your app in Visual Studio, you might as well run the app directly from visual studio for the quickest dev feedback loop, no need for an additional Dockerfile for this purpose.

avichalchum:
Hi @gtardif ,

Thank you for your feedback. My proposal was more around the build once, reuse multiple places principal and not for faster dev feedback. You're right, for a faster dev feedback, Visual Studio would be the fastest.

But if the same code artifacts are being used by multiple docker files and especially if the code build takes a long time (assuming greater than 15 minutes), then rather than build it for each dockerfile every time, the prod deploy CI tools will build the artifacts once and use those for the various dockerfiles. 

Additionally, most production pipeline CI's could be running static/dynamic code analysis, vulnerability analysis, e.t.c. after the code is built by the pipeline, so building it again inside the dockerfile would then be redundant and add time.

I believe it is due to these reasons Microsoft also proposes a similar dockerfile: [https://docs.microsoft.com/en-us/dotnet/core/docker/build-container?tabs=windows#create-the-dockerfile](Microsoft dockerfile example).

Please note than I am proposing replacing the build steps inside a dockerfile to that outside. I am proposing providing both methods, a user who wishes to build the code inside the dockerfile can use method 1 and a user who is using it as a part of the CI can use method 2.

engine/examples/dotnetcore.md
```diff
@@ -68,7 +70,29 @@ ENTRYPOINT ["dotnet", "aspnetapp.dll"]
 bin\
 obj\
 ```
+### Method 2 (CI recommended method):
```
gtardif, line 73:
```suggestion
### Method 2 (build app outside Docker container):
```

gtardif (commented):
Hi @avichalchum,
Thanks, I see your point. We also see a lot of projects where it's better to build as much as possible inside containers, to reduce installed tooling on CI machines and make build more reproducible, make it easier to maintain build tool versions, etc. 
I suggest we simply change the wording for your proposed method

avichalchum:
@gtardif thanks. I agree with your proposed wording and have accepted that.

gtardif (approved):

usha-mandya (approved):

PR #10911, opened 2020-05-28
handle archive feature toggling with css

jdrouet:
Signed-off-by: Jérémie Drouet <jeremie.drouet@gmail.com>

### Proposed changes

Toggle archive features only with css to fix race condition around loading `/js/archives.json`

### DIFF
--- a/_includes/head.html
+++ b/_includes/head.html
@@ -62,7 +62,6 @@
   <script>
   // Default to assuming this is an archive and hiding some stuff
   // See js/archive.js and js/docs.js for logic relating to this
-  var isArchive = true;
   var dockerVersion = 'v{{ site.docker_ce_version }}';
   </script>
 </head>
--- a/js/archive.js
+++ b/js/archive.js
@@ -6,28 +6,45 @@ layout: null
 if (window.navigator.onLine) {
   var suppressButterBar = false;
   /* This JSON file contains a current list of all docs versions of Docker */
-  $.getJSON("/js/archives.json", function(result){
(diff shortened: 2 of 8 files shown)


_includes/head.html
```diff
@@ -62,7 +62,6 @@
   <script>
   // Default to assuming this is an archive and hiding some stuff
   // See js/archive.js and js/docs.js for logic relating to this
-  var isArchive = true;
   var dockerVersion = 'v{{ site.docker_ce_version }}';
```
thaJeztah, line 65: not for this PR, but this one kept confusing me (we're setting it both here, and in the `_body.tml`

js/archive.js
```diff
@@ -6,28 +6,45 @@ layout: null
...
-                               '<span style="z-index: 1001" class="dropdown">';
-    var listStart = '<ul class="dropdown-menu" role="menu" aria-labelledby="archive-menu">';
-    var listEnd = '</ul>';
-    var outerDivEnd = '</span></div></div></div>';
+  $.getJSON("/js/archives.json", function (result) {
+    var outerDivStart =
+      '<div id="archive-butterbar"><div class="container"><div style="text-align: center"><span id="archive-list">This is <b><a href="https://docs.docker.com/docsarchive/" style="color: #254356; text-decoration: underline !important">archived documentation</a></b> for Docker&nbsp;' +
+      dockerVersion +
+      '. Go to the <a style="color: #254356; text-decoration: underline !important" href="https://docs.docker.com/">latest docs</a> or a different version:&nbsp;&nbsp;</span>' +
```
thaJeztah, line 13: So, if we're not gonna be publishing the current (19.03) version of the docs as an archive, do we need this script at all? Basically this script would only be actively running on archived versions

usha-mandya: Yup, that's correct. As per the earlier discussions, the current version will not be archived and there will be only one version of the docs on docs.docker.com

usha-mandya:
@thaJeztah As we have a plan to create a new PR to remove the archives, should we merge this one? The fixes in this PR impact https://github.com/docker/docker.github.io/pull/10967#issuecomment-641093607

PR #10927, opened 2020-05-29
Added wp-content to the docker-compose Wordpress example

DougBeney:
Hi! I can't think of any reason someone wouldn't want wp-content as a volume - whether for development or production.

Let me know your thoughts. I could understand why you wouldn't include `/var/www/html` as a whole - because there's a lot of wp core files, but wp-content  is something people would want to keep. Even if not for development purposes, for backup purposes.

From here, I think the user would have a pretty safe and usable Wordpress setup.


DougBeney:
Any update on this?

PR #10940, opened 2020-06-01
Add Engine v19.03.11 release notes

usha-mandya:
Release notes for Docker Engine version 19.03.11

### DIFF
--- a/engine/release-notes/index.md
+++ b/engine/release-notes/index.md
@@ -27,7 +27,24 @@ for Docker Engine.
 
 ### Network
 
-- Disable IPv6 Router Advertisements to prevent address spoofing. [CVE-2020-13401](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2020-13401)
+Disable IPv6 Router Advertisements to prevent address spoofing. [CVE-2020-13401](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2020-13401)
+
+**Description**
+
+In the Docker default configuration, the container network interface is a virtual ethernet link going to the host (veth interface).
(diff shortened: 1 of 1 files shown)


justincormack:
You can link to https://github.com/kubernetes/kubernetes/issues/91507

Also can use this text


In the Docker default configuration the container network interface is a virtual ethernet link going to the host (veth interface).
In this configuration, an attacker able to run a process as root in a container can send and receive arbitrary packets to the host using the CAP_NET_RAW capability (present in the default configuration).

If IPv6 is not totally disabled on the host (via ipv6.disable=1 on the kernel cmdline), it will be either unconfigured or configured on some interfaces, but it’s pretty likely that ipv6 forwarding is disabled, ie /proc/sys/net/ipv6/conf/*/forwarding == 0.
Also by default, /proc/sys/net/ipv6/conf/*/accept_ra == 1. The combination of these 2 sysctls means that the host accepts router advertisements and configures the IPv6 stack using them.

By sending “rogue” router advertisements from a container, an attacker can reconfigure the host to redirect part or all of the IPv6 traffic of the host to the attacker controlled container.
Even if there was no IPv6 traffic before, if the DNS returns A (IPv4) and AAAA (IPv6) records, many HTTP libraries will try to connect via IPv6 first then fallback to IPv4, giving an opportunity to the attacker to respond.
If by chance the host has a vulnerability like last year’s RCE in apt (CVE-2019-3462), the attacker can now escalate to the host.

As CAP_NET_ADMIN is not present by default for Docker containers, the attacker can’t configure the IPs they want to MitM, they can’t use iptables to NAT or REDIRECT the traffic, and they can’t use IP_TRANSPARENT.
The attacker can however still use CAP_NET_RAW and implement a tcp/ip stack in user space.

champtar:
@justincormack hope I haven't made any spelling errors :P

PR #10956, opened 2020-06-04
change div to a for interactive element

jdrouet:
Signed-off-by: Jérémie Drouet <jeremie.drouet@gmail.com>

### Proposed changes

Format search script and use `a` instead of `div` for interactive elements


usha-mandya:
@jdrouet could you please split format changes from the search updates so it's easier for review and merge the PR?

PR #10996, opened 2020-06-14
Update the version and change the RDBMS as described in the example for Rails.

madai0517:
### Proposed changes
Update versions of Ruby on Rails and change the Database from PostgreSQL to Change to MySQL 8.0.

Because
* Ruby2.5+Rails5 is old.
* There are more users of MySQL than PostgreSQL.
* There wasn't much information on building Rails+MySQL 8.0 with Docker.

### DIFF
--- a/compose/rails.md
+++ b/compose/rails.md
@@ -5,7 +5,7 @@ title: "Quickstart: Compose and Rails"
 ---
 
 This Quickstart guide shows you how to use Docker Compose to set up and run
-a Rails/PostgreSQL app. Before starting, [install Compose](install.md).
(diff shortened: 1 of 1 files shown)


madai0517:
If you want a new "View the Rails welcome page!" image, here it is Use.

Rails Example 2020-06-08 3 07 23

craig-osterhout:
Thanks for contributing. The rails sample app has been taken out of the docs and moved to the Awesome Compose repo to hopefully facilitate better maintenance. If the issues still exists, please help open an issue there, https://github.com/docker/awesome-compose/blob/master/official-documentation-samples/rails/README.md

Issue #10984, opened 2020-06-10
Link to "Authorization plugins" broken


ptxmac:
### Problem description

From https://docs.docker.com/engine/security/https/ there's a small paragraph about authorization plugins:

> Authorization plugins offer more fine-grained control to supplement authentication from mutual TLS. In addition to other information described in the above document, authorization plugins running on a Docker daemon receive the certificate information for connecting Docker clients.

That start of this is a link to https://docs.docker.com/engine/security/extend/plugins_authorization which doesn't exists

### Problem location

- I saw a problem on the following URL: https://docs.docker.com/engine/security/https/

craig-osterhout:
Thanks for sharing the issue. It appears that it was fixed. If the issue still persists, please let us know.

PR #10997, opened 2020-06-14
Fix 404 Not Found - Broken Link

jfernancordova:
### Proposed changes
[Authorization plugins](../extend/plugins_authorization) Link  was removed because it does not exist.

### Related issues (optional)
#10984

### DIFF
--- a/engine/security/https.md
+++ b/engine/security/https.md
@@ -93,7 +93,7 @@ Now, generate the signed certificate:
     Getting CA Private Key
     Enter pass phrase for ca-key.pem:
 
-[Authorization plugins](../extend/plugins_authorization) offer more
+Authorization plugins offer more
 fine-grained control to supplement authentication from mutual TLS. In addition
 to other information described in the above document, authorization plugins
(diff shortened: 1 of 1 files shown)


jfernancordova:
Actually it exists [plugins_authorization](https://docs.docker.com/engine/extend/plugins_authorization/), but the markdown can't be found in the repository, so in order to avoid missing 404, I create this PR. thx

engine/security/https.md
```diff
@@ -93,7 +93,7 @@ Now, generate the signed certificate:
     Getting CA Private Key
     Enter pass phrase for ca-key.pem:
 
-[Authorization plugins](../extend/plugins_authorization) offer more
+Authorization plugins offer more
```
thaJeztah, line 96: Looks like the URL has to be one level up; the page that's linked to is fetched during build of the documentation, so perhaps good to just use an absolute path;

```suggestion
[Authorization plugins](/engine/extend/plugins_authorization/) offer more
```

jfernancordova:
I get it, thank you @thaJeztah

PR #11000, opened 2020-06-16
Improve description for secret file handling

chrisgrf:
### Proposed changes

The secret files are still present in the Docker containers, but they do not hold any data/content/text. This is opposed to the current description, which states that the secret files are not present in the Docker containers.

### DIFF
--- a/develop/develop-images/build_enhancements.md
+++ b/develop/develop-images/build_enhancements.md
@@ -157,8 +157,8 @@ RUN --mount=type=secret,id=mysecret,dst=/foobar cat /foobar
 ```
 
 This Dockerfile is only to demonstrate that the secret can be accessed. As you
-can see the secret printed in the build output. The final image built will not
-have the secret file:
+can see the secret printed in the build output. The final image built will have the secret file,
+but without any content:
 
 ```console
 $ docker build --no-cache --progress=plain --secret id=mysecret,src=mysecret.txt .


develop/develop-images/build_enhancements.md
```diff
@@ -157,8 +157,8 @@ RUN --mount=type=secret,id=mysecret,dst=/foobar cat /foobar
 ```
 
 This Dockerfile is only to demonstrate that the secret can be accessed. As you
-can see the secret printed in the build output. The final image built will not
-have the secret file:
+can see the secret printed in the build output. The final image built will have the secret file,
+but without any content:
```
thaJeztah, line 161: I'm a bit on the fence on what we should write here. The reason the file exists the final image is that (in order to mount the file during build), an empty file is created to act as a mount-point. Ideally, that file would be cleaned up afterwards, but currently that's not the case.

I'd like to avoid the implementation details here; perhaps instead, we should update the example to use a target-path that's a tmpfs (e.g. change `dst=/foobar` to `dst=/run/foobar`) so that the file doesn't show up in the final image.

@tonistiigi WDYT?

tonistiigi: just using `/run/foobar` shouldn't work as `/run` is not tmpfs automatically. But generally I agree that there is no rule that empty file would be created. It depends on the parent mounts.

Perhaps:

"Note that with due to how mounting in Linux works, if the secrets are created in persistent directories the current implementation may leave behind empty stub files where the secret used to be. This behavior will likely change in the future and should not be relied upon."

chrisgrf: I would suggest a more passive phrasing than claiming the file is deleted. I was confused when I checked the directory and the file still existed while the documentation claimed something else. The documentation should comfort users that even though they might still see the file there is no chance that their secrets are still accessible.

PR #11016, opened 2020-06-18
from include to includes in Mac section.

imorti:

### DIFF
--- a/compose/install.md
+++ b/compose/install.md
@@ -51,7 +51,7 @@ Python package manager or installing Compose as a container.
 
 ### Install Compose on macOS
 
-**Docker Desktop for Mac** and **Docker Toolbox** already include Compose along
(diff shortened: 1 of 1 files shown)


compose/install.md
```diff
@@ -51,7 +51,7 @@ Python package manager or installing Compose as a container.
 
 ### Install Compose on macOS
 
-**Docker Desktop for Mac** and **Docker Toolbox** already include Compose along
+**Docker Desktop for Mac** and **Docker Toolbox** already includes Compose along
```
thaJeztah, line 54: I don't think it's correct, because it's "Docker Desktop for Mac and Docker Toolbox" (plural), in which case "include" is correct

usha-mandya:
Thanks @thaJeztah. 

@imorti Thank you for taking the time to update the docs. I'll have to close this PR without accepting your change as the docs look correct as they are today.

PR #11023, opened 2020-06-23
Update insecure.md

arcosx:
### Proposed changes

Some known issues indicate that this configuration needs to be explained more clearly when using buildkit.

### Related issues (optional)

https://github.com/moby/moby/issues/41059
https://github.com/moby/moby/issues/41080

### DIFF
--- a/registry/insecure.md
+++ b/registry/insecure.md
@@ -107,6 +107,24 @@ This is more secure than the insecure registry solution.
 
 This section lists some common failures and how to recover from them.
 
+### Known issue on buildkit
+
+When using [buildkit](https://github.com/moby/buildkit), if the registry is not configured with `https` mode, the `http` needs to be added. The specific reason can be found on [issue:41059](https://github.com/moby/moby/issues/41059),[issue:41080](https://github.com/moby/moby/issues/41080).
+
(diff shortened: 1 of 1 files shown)


usha-mandya:
@tiborvass Could you PTAL?

StefanScherer:
@tonistiigi Could you PTAL? Is there a fix coming or is it good to merge right now?

tonistiigi:
This has been updated in buildkit https://github.com/moby/buildkit/commit/67044875c4169d6663ac42e3b9e1619e232773ba . No sure about timeline when it will reach moby. I don't think this special section is needed for this, I'd rather just update the main example to include a protocol as there should never be a real-world case where tls vs http needs to be detected.

arcosx:
> This has been updated in buildkit [moby/buildkit@6704487](https://github.com/moby/buildkit/commit/67044875c4169d6663ac42e3b9e1619e232773ba) . No sure about timeline when it will reach moby. I don't think this special section is needed for this, I'd rather just update the main example to include a protocol as there should never be a real-world case where tls vs http needs to be detected.

 Yes, I also think this PR can be closed.Thanks everyone.

PR #11031, opened 2020-06-24
Add additional redirects for archives

thaJeztah:
Putting these after the existing redirects, so that ee-specific redirects in the archives take precedence.

### DIFF
--- a/_website-config-docs.json
+++ b/_website-config-docs.json
@@ -344,6 +344,45 @@
         "ReplaceKeyPrefixWith": null,
         "ReplaceKeyWith": "docker-enterprise/v3.0/dockeree-products/index.html"
       }
+    },
+    {
+      "Condition": {
+        "HttpErrorCodeReturnedEquals": null,
+        "KeyPrefixEquals": "v17.06/"
+      },
+      "Redirect": {
+        "HostName": "docs.docker.com",
+        "HttpRedirectCode": null,
+        "Protocol": "https",
+        "ReplaceKeyPrefixWith": "",
+        "ReplaceKeyWith": null
+      }
+    },
+    {
+      "Condition": {
+        "HttpErrorCodeReturnedEquals": null,
+        "KeyPrefixEquals": "v18.03/"
+      },
+      "Redirect": {
+        "HostName": "docs.docker.com",
+        "HttpRedirectCode": null,
+        "Protocol": "https",
+        "ReplaceKeyPrefixWith": "",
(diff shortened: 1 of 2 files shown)


thaJeztah:
@usha-mandya @StefanScherer PTAL

thaJeztah:
We'll have to test if they're added in the right position; I assumed that it processes redirects in order (so if there's an existing `/v18.09/ee/` redirect higher up the list, it would be processed first, and other `/v18.09/` will be processed by the later rule).

_website-config-docs.json
```diff
@@ -344,6 +344,45 @@
...
+    },
+    {
+      "Condition": {
+        "HttpErrorCodeReturnedEquals": null,
+        "KeyPrefixEquals": "v18.09/"
+      },
+      "Redirect": {
+        "HostName": "docs.docker.com",
+        "HttpRedirectCode": null,
+        "Protocol": "https",
```
thaJeztah, line 382: I saw in the docs that these are optional, so thinking of doing a follow-up to test if redirects work ok without these specified (in that case, we could possibly use a single config both for `docs-stage` and for `docs`

StefanScherer: I think I tried that, but maybe I did it wrong.
Anyhow the feedback loop is 
1. create PR
2. merge PR to master
3. test redirects on docs-stage.docker.com
4. revert PR if it broke something

StefanScherer: The other idea I had before adding two json files was, we could specify the HostName value when running the GitHub action https://github.com/docker/docker.github.io/blob/master/.github/actions/update-website-config/main.go that uploads the website config in the workflow https://github.com/docker/docker.github.io/blob/master/.github/workflows/build-master.yml#L27-L32

StefanScherer: I think I remember, when you leave that HostName away, then the internal URL to the S3 bucket is used after the redirect without the docker.com domain.

thaJeztah: Ah, that'd make sense yes, that's a pity

thaJeztah:
Tested this on staging and looks to work ok;

https://docs-stage.docker.com/v18.09/engine/install/ redirects to https://docs-stage.docker.com/engine/install/
and https://docs-stage.docker.com/v18.09/ee/ still redirects to the mirantis website

usha-mandya:
Redirect rules are live in prod. Tested this in prod and it all works fine.

PR #11059, opened 2020-06-29
Fixing ImageList Go code

andrealmar:
Original: on line 18 we have `images, err := cli.ImageList(context.Background(), types.ImageListOptions{})` which gives us an error of:

```
# command-line-arguments
./main.go:12:2: ctx declared but not used
```

### Proposed changes

*Fix:* on line 18 we need to have `images, err := cli.ImageList(ctx, types.ImageListOptions{})`

in order to fix the error outlined above

### DIFF
--- a/engine/api/sdk/examples.md
+++ b/engine/api/sdk/examples.md
@@ -471,7 +471,7 @@ func main() {
 		panic(err)
 	}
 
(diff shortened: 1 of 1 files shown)


usha-mandya:
Thank you for the update @andrealmar. 

@thaJeztah Could you PTAL?

PR #11066, opened 2020-07-01
add a note for non existence daemon.json

sobhanattar:
As `daemon.json` file does not exist in new installations , it's important to give a hint about adding one.


usha-mandya:
Thank you @sobhanattar for the update and @StefanScherer for the review.

PR #11071, opened 2020-07-01
Update image push reference

usha-mandya:
Made a minor update to the username (Docker ID) and description in the example


usha-mandya:
I can update the example here to say 'your-dockerid`. Users will be made aware of the username requirements when they create an ID in Docker Hub. Not sure we want to capture those requirements here.

PR #11077, opened 2020-07-02
Update linux-postinstall.md

cuzrawr:
replaced netstat to ss

### DIFF
--- a/engine/install/linux-postinstall.md
+++ b/engine/install/linux-postinstall.md
@@ -177,7 +177,9 @@ Configuring Docker to accept remote connections can be done with the `docker.ser
     ExecStart=/usr/bin/dockerd -H fd:// -H tcp://127.0.0.1:2375
     ```
 
-3. Save the file. ( this will create `/etc/systemd/system/docker.service.d/override.conf` )
+3. Save the file.
+
+    The file is saved as `/etc/systemd/system/docker.service.d/override.conf`.
 
 4. Reload the `systemctl` configuration.
 
--- a/engine/install/linux-postinstall.md
+++ b/engine/install/linux-postinstall.md
@@ -177,7 +177,7 @@ Configuring Docker to accept remote connections can be done with the `docker.ser
     ExecStart=/usr/bin/dockerd -H fd:// -H tcp://127.0.0.1:2375
     ```
 
-3. Save the file.
+3. Save the file. ( this will create `/etc/systemd/system/docker.service.d/override.conf` )
 
 4. Reload the `systemctl` configuration.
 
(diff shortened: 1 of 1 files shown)


engine/install/linux-postinstall.md
```diff
@@ -177,7 +177,7 @@ Configuring Docker to accept remote connections can be done with the `docker.ser
     ExecStart=/usr/bin/dockerd -H fd:// -H tcp://127.0.0.1:2375
     ```
 
-3. Save the file.
+3. Save the file. ( this will create `/etc/systemd/system/docker.service.d/override.conf` )
```
thaJeztah, line 180: I'm a bit on the fence if we should add this detail (a bit worried that it slightly distracts from the steps itself).


If we _do_ want to keep this, perhaps put it below the instruction itself (and avoid using future tense, which is generally avoided in technical documentation)


```suggestion
3. Save the file.

    The file is saved as `/etc/systemd/system/docker.service.d/override.conf`.
```

engine/install/linux-postinstall.md
```diff
@@ -191,11 +191,11 @@ Configuring Docker to accept remote connections can be done with the `docker.ser
     $ sudo systemctl restart docker.service
     ```
 
-6.  Check to see whether the change was honored by reviewing the output of `netstat` to confirm `dockerd` is listening on the configured port.
+6.  Check to see whether the change was honored by reviewing the output of `ss` to confirm `dockerd` is listening on the configured port.
```
thaJeztah, line 194: How common is it to have `ss` installed? I understand it mostly replaces `netstat`, but `netstat` (I think) is more common to have installed. Wondering if we'd need to have a note which package provides it (but not sure it's the same package name between RPM-based and Debian-like systems)

engine/install/linux-postinstall.md
cuzrawr: I thought it would be useful to someone.
If this violates docs standard then we throw away this.

engine/install/linux-postinstall.md
cuzrawr: As decribed here:
https://bugzilla.redhat.com/show_bug.cgi?id=1119297
And here:
https://github.com/inspec/inspec/issues/1655

We can found most linux distribution going to iproute rather thatn net-tools.

tianon: I have to admit I also wasn't familiar with `ss`, but it is installed (both on my host, and in my `network-toolbox` container), so I'll probably start adjusting my own usage. :sweat_smile:

For the benefit of our users though, we should probably add an extra note here, something like:

```suggestion
6.  Check to see whether the change was honored by reviewing the output of `ss` (or `netstat`) to confirm `dockerd` is listening on the configured port.
```

usha-mandya:
@thaJeztah Could you PTAL and let me know if this is ready to be merged? Thanks.

PR #11094, opened 2020-07-06
Update gettingstarted.md

### LINKED ISSUE #11093: Code and explanation are not in sync.
For this link,
https://docs.docker.com/compose/gettingstarted/

Code says that flask app is running in 6379 and in Dockerfile there is no port expose, in docker-compose the port mapping is done on 5000:5000. 
But the explanation says that we have to run the flask app in 5000, expose the correct port(5000) and then map the port properly from 5000:5000.

Its a bit misleading please fix one of Code or explanation.

[search.md](https://docs.docker.com/search/)

cerofrais:
1. Flask app was running in the wrong port.
2. 5000 port was not exposed in Dockerfile

closes #11093

### DIFF
--- a/compose/gettingstarted.md
+++ b/compose/gettingstarted.md
@@ -32,7 +32,7 @@ Define the application dependencies.
         from flask import Flask
 
         app = Flask(__name__)
-        cache = redis.Redis(host='redis', port=6379)
+        cache = redis.Redis(host='redis', port=5000)
 
 
         def get_hit_count():
@@ -90,6 +90,7 @@ following:
     COPY requirements.txt requirements.txt
     RUN pip install -r requirements.txt
     COPY . .
+    EXPOSE 5000
     CMD ["flask", "run"]
 
 This tells Docker to:
--- a/compose/gettingstarted.md
+++ b/compose/gettingstarted.md
@@ -32,7 +32,7 @@ Define the application dependencies.
         from flask import Flask
 
         app = Flask(__name__)
-        cache = redis.Redis(host='redis', port=5000)
+        cache = redis.Redis(host='redis', port=6379)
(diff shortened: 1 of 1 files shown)


compose/gettingstarted.md
```diff
@@ -32,7 +32,7 @@ Define the application dependencies.
         from flask import Flask
 
         app = Flask(__name__)
-        cache = redis.Redis(host='redis', port=6379)
+        cache = redis.Redis(host='redis', port=5000)
```
thaJeztah, line 35: This is the port for the redis container, which is 6379 by default, so I don't think this is correct

compose/gettingstarted.md
```diff
@@ -90,6 +90,7 @@ following:
     COPY requirements.txt requirements.txt
     RUN pip install -r requirements.txt
     COPY . .
+    EXPOSE 5000
```
thaJeztah, line 93: expose should not be needed, as it's only adding metadata to the image

compose/gettingstarted.md
cerofrais: My Mistake

compose/gettingstarted.md
cerofrais: Flask app is running at port 5000, docker-compose maps 5000 port in the docker container to 5000 post on the host machine. Doesn't the docker container need to expose the 5000 port? similar to (https://docs.docker.com/get-started/part2/#sample-dockerfile)

usha-mandya: Thank you @cerofrais. I've discussed with @thaJeztah. We also need to add a description of 'Expose'. I'll make this change in an incoming PR.

PR #11014, opened 2020-06-18
Remove archives

thaJeztah:

PR #11109, opened 2020-07-09
Fix missing "toggle" for right-hand navigation

thaJeztah:
This button was accidentally removed when the archives were removed in ba85012f47efdc41e30ad9e9f7a9a579d05f6382 (https://github.com/docker/docker.github.io/pull/11014)

This brings back that toggle button:

Screenshot 2020-07-09 at 12 32 42

We should have a look at re-designing this button, as it's not very clear (and easy to be missed), but that's good for a follow-up

### DIFF
--- a/_includes/navigation.html
+++ b/_includes/navigation.html
@@ -26,4 +26,7 @@
 
         </ul>
     </div>
+    <div class="ctrl-right">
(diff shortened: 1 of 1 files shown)


thaJeztah:
@usha-mandya PTAL

PR #11112, opened 2020-07-09
Publish updates from master

usha-mandya:

### DIFF
--- a/engine/context/aci-integration.md
+++ b/engine/context/aci-integration.md
@@ -155,16 +155,13 @@ The Docker ACI Integration CLI adds support for running and managing containers 
 You can install the new CLI using the install script:
 
 ```console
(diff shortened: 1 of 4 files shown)


engine/context/aci-integration.md
```diff
@@ -155,16 +155,13 @@ The Docker ACI Integration CLI adds support for running and managing containers
 You can install the new CLI using the install script:
 
 ```console
-curl -L https://github.com/docker/aci-integration-beta/releases/download/v0.1.4/install.sh | sh
+curl -L https://github.com/docker/aci-integration-beta/blob/main/scripts/install_linux.sh | sh
```
thaJeztah, line 158: we should use `-fsSL` for the curl command, and probably document downloading the script instead of piping it directly.

Something for a follow up

PR #11170, opened 2020-07-21
Add req on ARN and resource format

usha-mandya:
Adding a prerequisite on opting into the AWS account into the new ARN and resource format

### DIFF
--- a/engine/context/ecs-integration.md
+++ b/engine/context/ecs-integration.md
@@ -205,6 +205,12 @@ $ docker ecs version
 Docker ECS plugin 0.0.1
 ```
 
+## FAQ
+
+**What does the error `this tool requires the "new ARN resource ID format"` mean?**
+
+This error message means that your integration requires the new ARN resource ID format for ECS. To learn more, see [Migrating your Amazon ECS deployment to the new ARN and resource ID format](https://aws.amazon.com/blogs/compute/migrating-your-amazon-ecs-deployment-to-the-new-arn-and-resource-id-format-2/){: target="_blank" class="_"}.
+
 ## Feedback
 
 Thank you for trying out the Docker ECS Integration beta release. Your feedback is very important to us. Let us know your feedback by creating an issue in the [ecs-plugin](https://github.com/docker/ecs-plugin){: target="_blank" class="_"} GitHub repository.


engine/context/ecs-integration.md
```diff
@@ -33,6 +33,8 @@ To deploy Docker containers on ECS, you must meet the following requirements:
 
 2. Ensure you have an AWS account.
 
+3. Log into the AWS console and opt your AWS account into the new ARN resource ID format. For more information, see [Migrating your Amazon ECS deployment to the new ARN and resource ID format](https://aws.amazon.com/blogs/compute/migrating-your-amazon-ecs-deployment-to-the-new-arn-and-resource-id-format-2/){: target="_blank" class="_"}.
```
chris-crone, line 36: @metcalfc do you think we need this as a general instruction or has the change been around for long enough that we can have it as part of the FAQ?

metcalfc: FAQ is fine.

metcalfc: I think if were put the error message and fix in the faq that would be great.

usha-mandya: > @metcalfc do you think we need this as a general instruction or has the change been around for long enough that we can have it as part of the FAQ?

@chris-crone Do you mean the Desktop FAQ? I don't think we have an FAQ for the partner integration docs yet.

chris-crone: I was thinking just for the ECS integration. We could put it on the GitHub repo like [we did for ACI](https://github.com/docker/aci-integration-beta#faq), I'm not sure what makes it visible enough for users.

metcalfc: @usha-mandya I think it would be reasonable to just put a heading at the bottom of this doc.

```FAQ

1. What does `this tool requires the "new ARN resource ID format"` mean?

This integration requires the new ARN resource ID format for ECS. To learn more , see [Migrating your Amazon ECS deployment to the new ARN and resource ID format](https://aws.amazon.com/blogs/compute/migrating-your-amazon-ecs-deployment-to-the-new-arn-and-resource-id-format-2/){: target="_blank" class="_"}.. To enable it log into the AWS console and opt your AWS account into the new ARN resource ID format here: https://console.aws.amazon.com/ecs/home?#/settings.

usha-mandya: @metcalfc Updated. PTAL

justincormack:
You should also note that this setting is per region, so you have to make the change on every region you might want to use.

PR #11171, opened 2020-07-21
Add info on converting to an annual plan

usha-mandya:
Add instructions on how to convert a monthly team plan to an annual plan.


pkennedyr:
The team-specific change looks good. We should also add a docs update for switching from a monthly Pro plan to an annual plan.

pkennedyr:
LGTM

PR #11178, opened 2020-07-23

ftvda:
COPY command copies files from the host to the client, doesn't it?


usha-mandya:
@ftvda The existing description is correct as the command `COPY` adds files from the client's current directory as per the examples and the build context used in this topic. Unfortunately, I'll have to close this PR without merging your changes due to this reason.

ftvda:
I misunderstood the word "docker client". 
Thank you for replying!

PR #11180, opened 2020-07-24
Fix header color to be consistent

thaJeztah:

### DIFF
--- a/_includes/body-landing.html
+++ b/_includes/body-landing.html
@@ -1,4 +1,4 @@
-<body ng-app="Docker" ng-controller="DockerController" id="landing">
+<body ng-app="Docker" ng-controller="DockerController" id="landing" class="landing">
   <nav class="navbar header">
     <div class="container">
       <div class="navbar-header">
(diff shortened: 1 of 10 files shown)


thaJeztah:

_includes/body-landing.html
```diff
@@ -1,4 +1,4 @@
-<body ng-app="Docker" ng-controller="DockerController" id="landing">
+<body ng-app="Docker" ng-controller="DockerController" id="landing" class="landing">
```
thaJeztah, line 1: I think the id, and the `ng-` attributes can be removed now, but will look separately at those.

StefanScherer:
Agreed, I also could see the same color in the header. LGTM

I also tried on mobile, and I have some findings, but unrelated to this PR, I saw that on docs.docker.com as well.

docs-mobile

thaJeztah:
Oh, yes, I think I saw the "black" color for the header somewhere, and looked like it was possibly done on purpose, but admit I didn't check on my phone how it looked

usha-mandya:
Thanks Stefan. I checked the light and dark modes but didn't check the responsive design on the Netlify preview.

PR #11205, opened 2020-07-31
Update fedora.md

jjasghar:
### Proposed changes

Removed the `-y` from the different example commands, it seemed inconsistent, so this way you have to say `Y` to each removal and installation.

### DIFF
--- a/engine/install/fedora.md
+++ b/engine/install/fedora.md
@@ -77,7 +77,7 @@ Install the `dnf-plugins-core` package (which provides the commands to manage
 your DNF repositories) and set up the **stable** repository.
 
 ```bash
-$ sudo dnf -y install dnf-plugins-core
+$ sudo dnf install dnf-plugins-core
 
 $ sudo dnf config-manager \
     --add-repo \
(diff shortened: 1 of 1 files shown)


thaJeztah:
Can you explain why it's inconsistent? I know many people use these steps also as base for scripts, in which case waiting for user input wouldn't be useful. The text introducing the code explains what the code does, so from that perspective, I'm not sure it needs the extra confirmation.

jjasghar:
@thaJeztah it seems that the `-y` was used in some but not all of the commands. To your point, people copy and paste these commands without thinking about them. If you want to just do a `-y` everywhere in essence you are doing a `curl || bash` into their systems. Having removed the `-y` you force the person to say `yes` to the installation of any of the dependencies and anything else that might come in.

This is just a simple sanity check, if you have the `-y` you might as well create a "tl;dr" section with all the commands as one bash script so people don't even need to read the article.

Hopefully, this makes sense, thanks for reaching out.

usha-mandya:
@jjasghar Thanks for the PR. The `master` branch has been renamed to `main`. Some of the existing PRs have been automatically closed as a result of this change. Could you create a new PR against our latest docs if the suggested changes are still relevant? Thank you.

PR #11209, opened 2020-08-02
examples.md: Actually use `ctx` variable in Go example

fhemberger:
### Proposed changes

"List and manage containers" Go example:
`ctx := context.Background()` was defined but never used.
Follow the structure of the other examples instead.


thaJeztah:
thanks!

Issue #7622, opened 2018-11-06
Built in web server shouldn't be used


pupeno:
In this document:

[compose/django.md](https://docs.docker.com/compose/django/)

You end up with a Docker image that uses Django's built in server. Since Docker is very easy to ship to production, this is quite dangerous, as this server should never be used in prod:

https://docs.djangoproject.com/en/2.1/ref/django-admin/#runserver

> DO NOT USE THIS SERVER IN A PRODUCTION SETTING. It has not gone through security audits or performance tests. (And that’s how it’s gonna stay. We’re in the business of making Web frameworks, not Web servers, so improving this server to be able to handle a production environment is outside the scope of Django.)

PR #11228, opened 2020-08-09
Add note about setup being not viable for production

themasch:
### Proposed changes

I think there should be at least a small hint that this guide is intended as a bare-bones development environment and should not be rolled into production like this. 

### Related issues (optional)

Relates to #7622

### DIFF
--- a/compose/django.md
+++ b/compose/django.md
@@ -87,10 +87,10 @@ and a `docker-compose.yml` file. (You can use either a `.yml` or `.yaml` extensi
 
     > Note:
     >
-    > This uses the build in development server to run your application on port 8000.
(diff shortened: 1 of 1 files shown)


themasch:
Hi @usha-mandya, thanks for improving and merging this pull request so fast. 
But I think there is an error in the markdown syntax now, leading to the link not being rendered correctly: 
https://docs-stage.docker.com/compose/django/#define-the-project-components

I think the new line break in line 91 causes this?

usha-mandya:
Oops! Thanks for letting me know, @themasch. Will fix it in an incoming PR.

usha-mandya:
@themasch This has been fixed in PR-11236

PR #11233, opened 2020-08-11
Update running_ssh_service.md

Gerard-ODonnell:
1. The set function to allow root logins is commented by default, and the command here does not remove the comment, so root login would still fail. 
2. Increased clarification on the correct way to set environment variable within SSHd.


StefanScherer:
Thanks. But it actually works for me without any change. 
I'm about to close this PR, maybe we remove such examples in the future, because it's hard to maintain it everywhere.

PR #11234, opened 2020-08-11
Add Packet to the machine drivers index

displague:
Adds a link to the Packet docker machine driver:
<https://github.com/packethost/docker-machine-driver-packet>


dockertopia:
Hi @displague, thanks for this. In the meantime docker machine has been deprecated (see https://docs.docker.com/machine/).

PR #11257, opened 2020-08-17
Update index.md

msapaydin:
The dot at the end of the run command has confused me, as it caused an error, for clarity I think it is better to omit the dot.

### Proposed changes

Simply removed a "." from the end of the sentence of the docker command since I thought it may be confused with the "." corresponding to the current directory in linux and Mac.


usha-mandya (approved):

usha-mandya:
@msapaydin Thanks for your feedback. I see how this can be confusing for new users. I've slightly reordered the step instead of just removing the period from the step.

PR #11264, opened 2020-08-19
[install/ubuntu] remove apt update

arthurlogilab:
add-apt-repository  takes care of doing an apt update after adding a new repo

### Proposed changes

Simplify install by removing unnecessary step

### DIFF
--- a/engine/install/ubuntu.md
+++ b/engine/install/ubuntu.md
@@ -167,7 +167,6 @@ from the repository.
    Engine and containerd, or go to the next step to install a specific version:
 
     ```bash
-    $ sudo apt-get update
     $ sudo apt-get install docker-ce docker-ce-cli containerd.io
     ```
 


engine/install/ubuntu.md
```diff
@@ -167,7 +167,6 @@ from the repository.
    Engine and containerd, or go to the next step to install a specific version:
 
     ```bash
-    $ sudo apt-get update
```
thaJeztah, line 170: I don't think we can remove it here, because in the first steps above, we add the docker repository to the list of repositories; after adding, we need to run `apt-get update` to update the local package cache, so that `apt` knows which packages are available in that repository to install

thaJeztah (changes requested):
thanks for contributing! I don't think this change is correct (left a comment inline), so I'm closing this PR, but feel free to continue the conversation in case I overlooked something 😅

engine/install/ubuntu.md
arthurlogilab: When adding using `add-apt-repository`, the `apt update` is automatically launched in the single command. 

In very old versions (we would have to check which version of ubuntu is supported in your documentation/by docker) this wasn't the case.

PR #11297, opened 2020-08-26
Improve Toolbox to Desktop migration

djs55:
### Proposed changes

Previously we recommended converting the disk at the block level. This is a bit risky as any on-disk format change could prevent it from working.

It's safer to explicitly `docker commit`, then `docker save` and `docker load` containers and images.


usha-mandya (approved):
LGTM. Thanks @djs55.

PR #11299, opened 2020-08-26
Update ubuntu.md

john-visosky:
### Proposed changes

I was following the instructions in the "Uninstall Old Versions" section (near the top), and after doing the 'apt-get remove' step, I had to search to determine whether it was also safe to remove /var/lib/docker.  (It is if you don't have anything there you need).

I thought I'd save someone else the trouble of looking it up so added the changes in this pull request.

### DIFF
--- a/engine/install/ubuntu.md
+++ b/engine/install/ubuntu.md
@@ -42,7 +42,15 @@ $ sudo apt-get remove docker docker-engine docker.io containerd runc
 It's OK if `apt-get` reports that none of these packages are installed.
 
 The contents of `/var/lib/docker/`, including images, containers, volumes, and
-networks, are preserved. The Docker Engine package is now called `docker-ce`.
+networks, are preserved. If you do not need to save anything or you want to
+start with a clean installation, it is safe to remove this directory with the
+command:
+
+```bash
+$ sudo rm -rf /var/lib/docker
+```
+
+The Docker Engine package is now called `docker-ce`.
 
 ### Supported storage drivers
 
--- a/engine/install/ubuntu.md
+++ b/engine/install/ubuntu.md
@@ -46,8 +46,6 @@ networks, are preserved. If you do not need to save your existing data, and want
 start with a clean installation, refer to the [uninstall Docker Engine](#uninstall-docker-engine)
 section at the bottom of this page.
 
-The Docker Engine package is now called `docker-ce`.
-
 ### Supported storage drivers
 
 Docker Engine on Ubuntu supports `overlay2`, `aufs` and `btrfs` storage drivers.
--- a/engine/install/ubuntu.md
+++ b/engine/install/ubuntu.md
@@ -42,13 +42,9 @@ $ sudo apt-get remove docker docker-engine docker.io containerd runc
 It's OK if `apt-get` reports that none of these packages are installed.
 
 The contents of `/var/lib/docker/`, including images, containers, volumes, and
-networks, are preserved. If you do not need to save anything or you want to
-start with a clean installation, it is safe to remove this directory with the
-command:
-
-```bash
-$ sudo rm -rf /var/lib/docker
-```
+networks, are preserved. If you do not need to save your existing data, and want to
+start with a clean installation, refer to the [uninstall Docker Engine](#uninstall-docker-engine)
+section at the bottom of this page.
 
 The Docker Engine package is now called `docker-ce`.
 


usha-mandya:
Thank you @john-visosky.

@thaJeztah Could you PTAL?

engine/install/ubuntu.md
```diff
@@ -42,7 +42,15 @@ $ sudo apt-get remove docker docker-engine docker.io containerd runc
...
-networks, are preserved. The Docker Engine package is now called `docker-ce`.
+networks, are preserved. If you do not need to save anything or you want to
+start with a clean installation, it is safe to remove this directory with the
+command:
+
+```bash
+$ sudo rm -rf /var/lib/docker
+```
+
+The Docker Engine package is now called `docker-ce`.
```
thaJeztah, line 53: This feels a bit out of place now as a separate line, perhaps it should be removed altogether 🤔

engine/install/ubuntu.md
```diff
@@ -42,7 +42,15 @@ $ sudo apt-get remove docker docker-engine docker.io containerd runc
...
 
 The contents of `/var/lib/docker/`, including images, containers, volumes, and
-networks, are preserved. The Docker Engine package is now called `docker-ce`.
+networks, are preserved. If you do not need to save anything or you want to
+start with a clean installation, it is safe to remove this directory with the
+command:
+
+```bash
+$ sudo rm -rf /var/lib/docker
+```
```
thaJeztah, line 51: So, while I think it's correct, I'm a bit concerned about adding  this example in this section. This section is for situations where people are upgrading from either very old versions of docker, or if they previously had a distro-packaged version of docker installed.

My concern is that, while users _should_ read carefully, it's also easy to overlook why this line is here, and blindly copy/paste it to their terminal and end up with all (possibly important) data gone.

Perhaps instead refer them to the "uninstall docker" section if they want to do a clean install; https://docs.docker.com/engine/install/ubuntu/#uninstall-docker-engine

```suggestion
networks, are preserved. If you do not need to save your existing data, and want to
start with a clean installation, refer to the [uninstall Docker Engine](#uninstall-docker-engine)
section at the bottom of this page.
```

WDYT?

thaJeztah (changes requested):
Thanks for contributing! I left some thoughts inline; if we're updating the Ubuntu page, we should probably also update the other distros (debian, centos, fedora) accordingly

engine/install/ubuntu.md
john-visosky: Yes, I thought so, too.  However, I didn't want to remove it, and wasn't sure where a better place for it would be.  I'm just getting up to speed on docker, so...

engine/install/ubuntu.md
john-visosky: I agree, it's potentially dangerous.

Your suggested edit is good.

john-visosky:
@thaJeztah You suggested that the same edit should be done on the other distro pages.  Should that be done in this pull request?  Or create additional ones for each distro's page?

engine/install/ubuntu.md
john-visosky: @thaJeztah Do you want this line removed?

thaJeztah: Oh, sorry for the delay, yes, I think we could as well remove this line

engine/install/ubuntu.md
```diff
@@ -42,7 +42,11 @@ $ sudo apt-get remove docker docker-engine docker.io containerd runc
 It's OK if `apt-get` reports that none of these packages are installed.
 
 The contents of `/var/lib/docker/`, including images, containers, volumes, and
-networks, are preserved. The Docker Engine package is now called `docker-ce`.
+networks, are preserved. If you do not need to save your existing data, and want to
+start with a clean installation, refer to the [uninstall Docker Engine](#uninstall-docker-engine)
+section at the bottom of this page.
+
+The Docker Engine package is now called `docker-ce`.
```
thaJeztah, line 49:
```suggestion
```

thaJeztah (changes requested):
Thanks for updating; left one suggestion; we could merge that suggestion, and do the other distro's in a follow-up, or as part of this PR; let me know what you prefer 🤗

usha-mandya:
Committed the changes suggested by @thaJeztah. 

@john-visosky Would you be able to create a new PR to update the additional distros? Thanks.

thaJeztah (approved):
LGTM

PR #11304, opened 2020-08-27
Change wording for setting capabilities lists

allir:
### Proposed changes

Update the wording for allow/deny approach for setting capabilities.


thaJeztah (approved):
LGTM, thanks!

I still need to check our codebase as well; I expect there to be occurrences like this in our code, but haven't come round to that yet 😓

allir:
Yeah, I'm guessing everyone will find stuff like this here and there in docs and code 🙂

PR #11316, opened 2020-08-31
Make image order clear

karl-johan-grahn:
Make it clear which order the image should be read - this is not always clear (think people with native languages where reading order is not left-to-right by default)


usha-mandya (approved):
LGTM. Thank you @karl-johan-grahn.

PR #11324, opened 2020-09-01
Update wsl.md

SQLvariant:
Trying to clarify for the user that they should run the `wsl -l -v` command on Windows, not within the Linux distro.


usha-mandya:
Thanks for the update @SQLvariant. We've slightly modified the commands to ensure they work better in any WSL distros and Windows terminals.

usha-mandya (approved):

PR #11328, opened 2020-09-02
Remove Mutagen doc

usha-mandya:
Removing the Mutagen topic from the docs as the feature has been removed from Docker Desktop Edge 2.3.5.0.


StefanScherer (approved):
Makes sense.
LGTM

PR #11335, opened 2020-09-03
use .tgz instead of .tar.gz

nwtgck:
I found .tar.gz in the document should be .tgz because the file is .tgz as follows.
https://download.docker.com/linux/static/stable/x86_64/


ulyssessouza (approved):
LGTM

thaJeztah (approved):
LGTM, thanks!

usha-mandya (approved):
LGTM. Thank you @nwtgck.

PR #11369, opened 2020-09-14
Container base layer in jpg is 15.04

agbanagba:
### Proposed changes
Update the release version of base layer of container image to match what's in the jpg image files.


thaJeztah (approved):
LGTM, thanks!

PR #11381, opened 2020-09-16
fix the background color of the sidebar on 'medium' sized devices

### LINKED ISSUE #11366: The formatting of the sidebar is unreadable on smaller screens
### Problem description
On desktop when I have my screen split and I open the sidebar from the hamburger menu, it just overlays the current text without additional formatting, so I cannot read it. I have attached a quick video of it happening to better describe it.

https://drive.google.com/file/d/1hajRmJ5zpw46wf_tszl6lFmyjgZIdsvX/view?usp=sharing

### Problem location

- I saw a problem on the following URL: https://docs.docker.com/registry/deploying/

- Other: I'm guessing since there is a sidebar almost everywhere, then this issue would apply everywhere

n/a

### Suggestions for a fix
I'm taking a look to see if it is an easy fix

ingshtrom:
### Proposed changes
This fixes #11366 by adding the background color for the correct class name. It
seems that `.sidebar-home` used to be used a lot, but I couldn't find its usage
anywhere (when loading up the docs app), so I stuck to using `.sidebar`, which
seemed to work :smile:

### Related issues (optional)

Fixes #11366

### DIFF
--- a/_scss/_mobile.scss
+++ b/_scss/_mobile.scss
@@ -61,6 +61,7 @@
         position: fixed;
         width: 100%;
         z-index: 100;
+        background-color: $bg-sidebar;
     }
     .main-content {
         padding: 22px 35px 25px 30px;
@@ -117,16 +118,10 @@
     .header {
         height: 100px;
     }
-    .nav-secondary-tabs,
-    .nav-secondary {
-        background-color: $black;
-    }
-    .nav-secondary input[type=search] {
-        background: rgb(78, 77, 77) url(/images/search.png) no-repeat 10px 9px;
-    }
+    /*.nav-secondary-tabs,
     .nav-sidebar a {
         color: #a8a8a8;
-    }
+    }*/
     .tabs {
         float: left;
         margin: 0;
@@ -147,8 +142,8 @@
     }
     .sidebar,
     .sidebar-home {
-        background-color: #15212e;
-        border-right: 1px solid rgba(204,204,204,0.29);
+        background-color: $bg-sidebar;
+        /*border-right: 1px solid rgba(204,204,204,0.29);*/
         top: 51px;
         left: 0;
         display: block;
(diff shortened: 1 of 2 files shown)


usha-mandya:
Thanks @ingshtrom. Looks like the issue still exists

image

ingshtrom:
Oops! I fixed this, and then noticed that the header and side-bar changes color (black navbar, and darker sidebar) at the lowest resolutions. I fixed that, but maybe it was purposeful. Please let me know if it was and I can roll that back.

usha-mandya:
Thank you @ingshtrom. It looks fine to me. @StefanScherer Do you mind taking another look when you get a chance? Thanks.

_scss/_mobile.scss
```diff
@@ -117,16 +118,10 @@
     .header {
         height: 100px;
     }
-    .nav-secondary-tabs,
-    .nav-secondary {
-        background-color: $black;
```
StefanScherer, line 122: Looks good so far, only thing I see is that on mobile the top bar is blue in light mode, but when I rotate my mobile to landscape it turns black. 
Previously on mobile it always was black. I think this change here was for black -> blue.
I like the blue background better in light mode.
But we should also fix it for landscape mode further down line 234 ff.

ingshtrom: Right, what I noticed was that on mobile the navbar (top bar) was black so I was trying to make it consistent with non-mobile views. I will change this for landscape mode as well. Nice find!

ingshtrom:
Oh, man.... I also noticed the search input changes based on the size of the screen... I'll fix that too.

ingshtrom:
PTAL. I think I got it this time. Mobile portrait and landscape are looking good. Just making a desktop screen smaller is looking good.

StefanScherer (approved):
LGTM 

Thanks Alex! Works fine now on mobile (portrait and landscape).

usha-mandya (approved):
LGTM. Thank you Alex.

PR #11374, opened 2020-09-15
Update Windows version for WSL 2

sliepie:
### Proposed changes
Update Windows version for WSL 2.
WSL 2 is backported to Windows 10 1903

PR #11391, opened 2020-09-17
Clarify the min WSL 2 Windows OS version 2004 for Stable and 1903 for Edge

StefanScherer:
### Proposed changes

This PR clarifies that only the Edge release of Docker Desktop supports WSL 2 on Windows 10, version 1903 or higher.
Regression of #11374 where we just reduced the OS version number from 2004 to 1903. 

Can be reverted once we have Stable 2.4 released. 

### Related issues (optional)

https://github.com/docker/for-win/issues/8549


usha-mandya (approved):
LGTM

PR #11393, opened 2020-09-17
logging: use separate column for default values

thaJeztah:
This makes it slightly easier to find what the default options are
for these logging drivers. Could possibly be added to some of the
other drivers as well.

@usha-mandya


usha-mandya:
Thank you @thaJeztah. The Description column looks a bit compressed now.

image

thaJeztah:
mmm, yes that doesn't look great. Though I checked, but perhaps only other page, or before I added the (unlimited)

PR #11396, opened 2020-09-18
Add missing external flag indicator in secrets

ChangingTerry:
### Proposed changes

This is required otherwise it tries to read a non-existing file and should make it clear for anyone following along

### Related issues (optional)

None


usha-mandya:
@ndeloof Could you PTAL?

ndeloof (approved):

PR #11399, opened 2020-09-19
bind-mounts.md: source (host) path must be absolute

jjlin:
The source path on the host system must be absolute. Using a relative path results in an error, e.g.

```
$ docker run -it -v ./sourcedir:/targetdir alpine
docker: Error response from daemon: create ./sourcedir: "./sourcedir" includes invalid characters for a local volume name, only "[a-zA-Z0-9][a-zA-Z0-9_.-]" are allowed. If you intended to pass a host directory, use absolute path.
See 'docker run --help'.
```


usha-mandya (approved):
LGTM. Thank you @jjlin.

PR #11414, opened 2020-09-23
remove examples for ENV key value without '='

thaJeztah:
The `ENV key value` form can be ambiguous, for example, the following defines a single env-variable (`ONE`) with value `"TWO= THREE=world"`:

    ENV ONE TWO= THREE=world

While we cannot deprecate/remove that syntax (as it would break existing Dockerfiles), we should reduce exposure of the format in our examples.

Also updating some code-blocks that were missing language-hints

### Related issues (optional)

relates to https://github.com/moby/buildkit/pull/1692#issuecomment-696823293, https://github.com/docker/cli/pull/2741


thaJeztah:
Need to verify if the code-hint changes/indentation is right, so waiting for Netlify preview

compose/django.md
```diff
@@ -62,25 +64,25 @@ and a `docker-compose.yml` file. (You can use either a `.yml` or `.yaml` extensi
 
 9. Add the following configuration to the file.
 
-    ```none
-     version: '3'
+    ```yaml
+    version: "{{ site.compose_file_v3 }}"
```
thaJeztah, line 68: Hmm.. this somehow adds a space in front of each line. I recall we had a similar issue at some point; https://github.com/docker/docker.github.io/pull/8105 (also somewhere else, but trying to find what the solution was for that)

thaJeztah:
ok, fixed the indentation, this should be ready for review.

To view without the indentation changes, use this link: https://github.com/docker/docker.github.io/pull/11414/files?w=1

@usha-mandya ptal

StefanScherer (approved):
Makes sense!

LGTM

StefanScherer:
Thanks @thaJeztah Good to know that.
Today I tried the VSCode Docker extension to create a Dockerfile for Node.js and it uses the old format, also their docs show some old examples. https://code.visualstudio.com/docs/containers/quickstart-node#_add-an-environment-variable-to-the-image

I try to create PR's in the extension and their docs to move to `ENV key=var`

thaJeztah:
Good catch, thanks @StefanScherer !

usha-mandya:
Thank you @StefanScherer @thaJeztah

StefanScherer:
I also created https://github.com/microsoft/vscode-docs/pull/4024 in the vscode docs repo.

PR #11425, opened 2020-09-25
Make the latest API reference page the canonical URL

thaJeztah:
We host multiple versions of the API reference. While older versions of the
API are still supported by the latest engine release, users should generally
refer to the latest version of the API.

This patch adds a "canonical" meta-tag to the API reference pages, and points
it to the latest version of the API.

Note that there's also a /engine/api/latest/ page, but I didn't pick that
URL, because it's a redirect to `/engine/api/v<current>/`, which Google
probably doesn't like.


thaJeztah:
@usha-mandya ptal

usha-mandya (approved):
LGTM.

PR #11429, opened 2020-09-26
Add indication to use docker create/run on port mapping #11420

### LINKED ISSUE #11420: Should specify this is using the "docker run" command
[config/containers/container-networking.md](https://docs.docker.com/config/containers/container-networking/)

For people really familiar with docker, this might be obvious, but for those just learning how docker works, the "-p" options should be specified as "docker run -p" to prevent possible confusion for those of us who are new to the whole docker thing.

Thanks!

maximillianfx:
Signed-off-by: Maximillian Fan Xavier <maximillianfx@gmail.com>

Proposed changes
Add some extra info to documentation to indicate the use of docker run/create when executing port mapping.

Related issues (optional)
Fixes #11420

### DIFF
--- a/config/containers/container-networking.md
+++ b/config/containers/container-networking.md
@@ -19,7 +19,11 @@ point of view of the container.
 
 ## Published ports
 
-By default, when you create/run a container using `docker create` or `docker run`, it does not publish any of its ports
+By default, when you create or run a container using `docker create` or `docker run`,
+it does not publish any of its ports to the outside world. To make a port available
+to services outside of Docker, or to Docker containers which are not connected to
+the container's network, use the `--publish` or `-p` flag. This creates a firewall
+rule which maps a container port to a port on the Docker host. Here are some examples.
 to the outside world. To make a port available to services outside of Docker, or
 to Docker containers which are not connected to the container's network, use the
 `--publish` or `-p` flag. This creates a firewall rule which maps a container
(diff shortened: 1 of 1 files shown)


usha-mandya:
Thank you for the update @maximillianfx.

@thaJeztah Could you PTAL?

config/containers/container-networking.md
```diff
@@ -19,7 +19,7 @@ point of view of the container.
 
 ## Published ports
 
-By default, when you create a container, it does not publish any of its ports
+By default, when you create/run a container using `docker create` or `docker run`, it does not publish any of its ports
```
thaJeztah, line 22: Perhaps change the `create/run` to "create or run". Also looks like this line is now quite a bit longer than 80 chars (which is what we try to use as width), so may need some wrapping;

```suggestion
By default, when you create or run a container using `docker create` or `docker run`,
it does not publish any of its ports to the outside world. To make a port available
to services outside of Docker, or to Docker containers which are not connected to
the container's network, use the `--publish` or `-p` flag. This creates a firewall
rule which maps a container port to a port on the Docker host. Here are some examples.
```

thaJeztah (changes requested):
thanks! I left a suggestion, but otherwise looks good

usha-mandya (approved):

usha-mandya:
@thaJeztah Looks like you need to approve the PR.

thaJeztah (approved):
LGTM, thanks!

PR #11430, opened 2020-09-26
Fix docker logs example #11398

### LINKED ISSUE #11398: Missing parameter in example
[engine/reference/commandline/logs.md](https://docs.docker.com/engine/reference/commandline/logs/)

Seemingly should be 
$ docker logs -f --until=2s test
instead of 
$ docker logs -f --until=2s
in the Examples section.

maximillianfx:
Signed-off-by: Maximillian Fan Xavier <maximillianfx@gmail.com>

### Proposed changes

Fix docker logs example using `until` flag

### Related issues (optional)

Fixes #11398

### DIFF
--- a/_data/engine-cli/docker_logs.yaml
+++ b/_data/engine-cli/docker_logs.yaml
@@ -108,7 +108,7 @@ examples: |-
   $ docker run --name test -d busybox sh -c "while true; do $(echo date); sleep 1; done"
   $ date
   Tue 14 Nov 2017 16:40:00 CET
-  $ docker logs -f --until=2s
+  $ docker logs -f --until=2s test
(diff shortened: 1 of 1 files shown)


usha-mandya:
@thaJeztah Could you PTAL?

thaJeztah:
Thanks! Change itself looks good, but these YAML files are generated from the docs that live in the CLI repository. While we could temporarily make fixes in the files here, I'd prefer to do them upstream, so that fixes don't get reverted if we update the generated versions.

I did a quick PR https://github.com/docker/cli/pull/2748 and backport https://github.com/docker/cli/pull/2749 with this change 👍

PR #11433, opened 2020-09-28
Make first apt example follow all best practices

mrexodia:
### Proposed changes

This is a bit of a draft, but I came across this issue in the documentation. When I searched for `docker apt-get` I found this page and copy-pasted the first example, but then I remembered it's best practice to install all packages at once and after that clear the `lists` to minimize the image size. Putting the example with all best practices first might help newer docker users to produce smaller images.

I'm not sure about the approach, perhaps I should just simply move the best example header to the top and leave the smaller examples alone.


usha-mandya (approved):
LGTM. Thank you @mrexodia

PR #11442, opened 2020-09-29
sort redirect_from in some files

thaJeztah:
This makes it easier to find if a redirect is already present, and to spot duplicates


usha-mandya (approved):
LGTM

PR #11449, opened 2020-09-30
Update container-networking.md

NullableInt:
### Proposed changes

I moved a partial sentence around, it made more sense like this compared to how it was.


usha-mandya (approved):
LGTM. Thank you @NullableInt

PR #11453, opened 2020-09-30
u with lsb_release gives upstream release codename

qumberrizvi:
### Proposed changes

When adding repo using `add-apt-repository` on a flavour of Ubuntu, `lsb_release` will give you the flavour's codename instead of Ubuntu's. Adding `-u` flag gives you the upstream codename instead which is more helpful.
E.g. On Elementary OS Hera `lsb_release -cs` returns `hera` but `lsb_release -cus` returns `bionic`.

### DIFF
--- a/engine/install/ubuntu.md
+++ b/engine/install/ubuntu.md
@@ -134,7 +134,7 @@ from the repository.
     ```bash
     $ sudo add-apt-repository \
        "deb [arch=amd64] {{ download-url-base }} \
-       $(lsb_release -cs) \
+       $(lsb_release -cus) \
        stable"
     ```
 
@@ -144,7 +144,7 @@ from the repository.
     ```bash
     $ sudo add-apt-repository \
        "deb [arch=armhf] {{ download-url-base }} \
-       $(lsb_release -cs) \
+       $(lsb_release -cus) \
        stable"
     ```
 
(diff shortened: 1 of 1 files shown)


usha-mandya:
@thaJeztah Could you PTAL?

qumberrizvi:
image_2020_10_01T13_51_13_566Z-1

Here's a comparison on Element OS 5.1 Hera - built on Ubuntu 18.04 LTS (Bionic Beaver).

thaJeztah:
Thanks for the suggestion, but I don't think this will work, because the `-u` option is only available on Ubuntu/Debian derivatives, but not on Ubuntu/Debian itself.

If you try to run this on Ubuntu, it will produce an error;

```
lsb_release -cus
Usage: lsb_release [options]

lsb_release: error: no such option: -u
```

qumberrizvi:
Oh I see. Thanks, I did not know that. I guess we could check the output if bash returns non 0 exit code and then try the other option (`command && commandIfZero || commandIfNonZero`). But that would perhaps be too cumbersome for a user?

thaJeztah:
Yes, I don't think we should make things too complicated, and instead have users set the right distro manually; https://docs.docker.com/engine/install/#other-linux-distributions.

I know we made some attempts in the past to make the instructions work for all distros (and similar approaches for the install script on get.docker.com), but there's too many variables at play

qumberrizvi:
Ah, I'll close this, then. Thanks. 😊

thaJeztah:
Thank you! Sorry that we weren't able to merge your PR ❤️

PR #11455, opened 2020-10-01
Fix names of storage account and file share

satoryu:
### Proposed changes

Fix options for `docker volume create` since its names were not correct. 

### Related issues (optional)

Notthing


usha-mandya (approved):
LGTM. Thank you @satoryu San :)

PR #11472, opened 2020-10-04
Added label for menu toggle to improve site accessibility on screen readers

123aswin123:
The toggle button on the top right of the individual pages was flagged by [Google's lighthouse](https://developers.google.com/web/tools/lighthouse) as a sub-optimal component for users on screen readers. The exact flag was:

> Links do not have a discernible name
Link text (and alternate text for images, when used as links) that is discernible, unique, and focusable improves the navigation experience for screen reader users. Learn more.
Failing Elements
a#menu-toggle

To address this issue, I have added an [aria-label](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-label_attribute) with the value "Current page's menu toggle", feel free to change it to a more suitable string as you see fit. 

**Here is a screenshot of the menu component I am speaking about:**

The Toggle Menu

This is the last of major the accessibility issues that were spotted on docker's documentation website. Thank you!

### DIFF
--- a/_includes/navigation.html
+++ b/_includes/navigation.html
@@ -29,6 +29,6 @@
         </ul>
     </div>
     <div class="ctrl-right">
-        <a href="javascript:void(0)" id="menu-toggle"><i class="fa fa-indent" aria-hidden="true"></i></a>
+        <a href="javascript:void(0)" id="menu-toggle" aria-label="Current page's menu toggle"><i class="fa fa-indent" aria-hidden="true"></i></a>
     </div>
 </div>


123aswin123:
Based on my [discussion](https://github.com/docker/docker.github.io/commit/fbfd22c08fbcd418f85ea9bad4667a7e91bb95ac#commitcomment-42947009) with @thaJeztah , I have also removed the unused `<i class="fa fa-indent" aria-hidden="true"></i>` element as it was put in there to improve accessibility, but unfortunately it did nothing.

The `<i class="fa fa-indent" aria-hidden="true"></i>` has been replaced with the cleaner [aria-label](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-label_attribute) to achive the goal of making this particular toggle menu more accessible.

usha-mandya:
Thank you @123aswin123. 

@thaJeztah Could you PTAL?

_includes/navigation.html
```diff
@@ -27,6 +27,6 @@
         </ul>
     </div>
     <div class="ctrl-right">
-        <a href="javascript:void(0)" id="menu-toggle" aria-label="Current page's menu toggle"><i class="fa fa-indent" aria-hidden="true"></i></a>
+        <a href="javascript:void(0)" id="menu-toggle" aria-label="Current page's menu toggle"></a>
```
thaJeztah: Looks like this tag is actually used to show the icon, so I think we should remove this commit 😅

thaJeztah (changes requested):
Thanks! I tried this PR without the second commit, and then it looked good. Let me push to this branch to remove the second commit, then we can merge 👍

thaJeztah (approved):
LGTM, I pushed to the branch, and removed the second commit; all looking good now, thanks!

_includes/navigation.html
123aswin123: That kind of makes sense! I never knew an <i> tag would be used to build a menu, they said the <i> tag was used to italicize text 😃. My whole life was a lie 🤯.

PR #11482, opened 2020-10-06
Revamp how the dark mode theme is managed in the codebase

### LINKED ISSUE #11468: Flash of incorrect theme (FOIT) occurs when viewing docs in dark mode
### Problem description

With the dark mode theme selected, switching between some pages, causes a "flash of incorrect theme" or "FOIT" in which the screen flashes bright white for a split second before switching to the dark theme. This is annoying and possibly seizure inducing for some.

### Problem location

For example, if you go to https://docs.docker.com/engine/ and then click the second item in the left menu, "Install", the flash of incorrect theme happens.

allejo:
Happy Hacktoberfest :jack_o_lantern:  Moby devs and fans :whale:

### Proposed changes

This PR completely changes how the website handles the dark mode by using [a `@media prefers-color-scheme` query](https://caniuse.com/?search=prefers-color-scheme) instead of a huge overrides file that had conflicting or overlapping CSS rules.

I have introduced a new `themer` mixin, which you will now use when you want to introduce a different color for the other theme; all in one place instead of separate Sass files. You could also add as many different themes as you'd like in this mixin too.

Why: I _really_ hated quick flash of the white theme whenever JS wouldn't kick in fast enough.

### Related issues (optional)

Fixes #11468


jeffslofish:
These changes make it work great in dark mode, however, in light mode, now there is a brief flash of dark mode when switching pages. 

It seems to have just pushed the "flash of incorrect theme" from dark mode to light mode.

allejo:
> It seems to have just pushed the "flash of incorrect theme" from dark mode to light mode.

This is mostly true. This PR changes the stylesheet to favor your operating system's light/dark mode. You'll only see the "flash of incorrect theme" when you purposely invert the theme from your OS' settings; e.g. your OS is in dark mode and you choose to use the site's light mode theme.

allejo:
yo @thaJeztah, is this PR something you all would be interested in...? There are a lot of conflicts in this PR now due to recent changes, which still don't make use of a CSS media query. I don't want to put in any more effort if this isn't something you all want.

allejo:
k. guess no one cares

PR #11486, opened 2020-10-06
Document manually installing rootlesskit

JasonMFry:
I'm assuming this is done in the install script, but if you're manually
installing the binaries you also need to manually install rootlesskit.

### DIFF
--- a/engine/security/rootless.md
+++ b/engine/security/rootless.md
@@ -173,6 +173,8 @@ extract `docker-rootless-extras-<version>.tgz`. The archive can be extracted
 under an arbitrary directory listed in the `$PATH`. For example, `/usr/local/bin`,
 or `$HOME/bin`.
 
+You will need to [install](https://github.com/rootless-containers/rootlesskit#setup) rootlesskit.
+
 ### Nightly channel
 
 To install a nightly version of the Rootless Docker, run the installation script


engine/security/rootless.md
```diff
@@ -173,6 +173,8 @@ extract `docker-rootless-extras-<version>.tgz`. The archive can be extracted
 under an arbitrary directory listed in the `$PATH`. For example, `/usr/local/bin`,
 or `$HOME/bin`.
 
+You will need to [install](https://github.com/rootless-containers/rootlesskit#setup) rootlesskit.
```
AkihiroSuda, line 176: No, this is included in `docker-rootless-extras-<version>.tgz`

AkihiroSuda (changes requested):
Doesn't seem correct

usha-mandya:
@JasonMFry Thanks for the PR. I'll have to close this one without merging as the rootlesskit is included in `docker-rootless-extras-<version>.tgz` as mentioned by @AkihiroSuda.

engine/security/rootless.md
JasonMFry: It's been a few weeks and I don't recall the details, but I'm pretty sure it didn't just work for me. But I no longer have the Linux machine I was using when doing this, so I can't really help further :(

PR #11487, opened 2020-10-06
use jekyll configuration for edit urls

thaJeztah:
The existing implementation was broken, and didn't override the URLs (due to leading, trailing slashes the paths were not matched).

Instead of using the custom "not_edited_here.yaml", set the edit-url as front-matter variable through the _config.yml


_includes/body.html
```diff
@@ -1,3 +1,8 @@
+{%- if page.edit_url -%}
+    {%- assign edit_url = page.edit_url -%}
+{%- else -%}
+    {%- assign edit_url = "https://github.com/docker/docker.github.io/edit/master/" | append: page.path -%}
+{%- endif -%}
```
thaJeztah, line 5: Added this to the "body" include, as we don't need it on the landing page

StefanScherer (approved):
SGTM
Good idea!

usha-mandya (approved):
LGTM

PR #11488, opened 2020-10-06
Generate page title from H1 if needed

thaJeztah:
This is a very hacky way to extract the page title from pages that do not have
front-matter yaml, but have a H1 header. We need to take (id-) attributes into
account, so some hacking is needed.

Note that there's also a Jekyll plugin that features similar functionality, but
it requires additional dependencies, and we only have a few pages that need
this, so for now using this hack.

First two commits are some basic cleaning up

### DIFF
--- a/_includes/head.html
+++ b/_includes/head.html
@@ -16,9 +16,10 @@
           display: block;
       }
   </style>
-  {% if site.google_analytics != '' %}{% include analytics/google_analytics.html GOOGLE_ID=site.google_analytics %}{% endif %}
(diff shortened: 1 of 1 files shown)


thaJeztah:

thaJeztah:
follow-up PR: https://github.com/docker/docker.github.io/pull/11489

StefanScherer (approved):
SGTM

usha-mandya (approved):
Thank you @thaJeztah

PR #11489, opened 2020-10-06
Fix missing descriptions

thaJeztah:
follow-up to https://github.com/docker/docker.github.io/pull/11488

- Add missing descriptions, and move reference index page
- _includes/head.html: generate description for reference pages
  This generates a description for pages that don't have one;
    - for reference docs, try the "long" description
    - fallback to "short" description
    - finally, fallback to taking the first 30 words from the page content

### DIFF
--- a/_includes/head.html
+++ b/_includes/head.html
@@ -14,6 +14,16 @@
     {%- assign c = b | split: '>' | last -%}
     {%- assign page_title = c | strip_html | strip | truncatewords: 10 -%}
 {%- endif -%}
+{%- if page.description == nil and page.datafile != nil and page.datafile != '' -%}
+    {%- assign yaml_data = site.data[page.datafolder][page.datafile] -%}
+    {%- if yaml_data.long and yaml_data.long != '' -%}
+        {%- assign page_description = page.title | append: ': ' | append: yaml_data.long | strip_html | strip | truncatewords: 30 -%}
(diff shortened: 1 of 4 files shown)


_includes/head.html
```diff
@@ -14,6 +14,16 @@
     {%- assign c = b | split: '>' | last -%}
     {%- assign page_title = c | strip_html | strip | truncatewords: 10 -%}
 {%- endif -%}
+{%- if page.description == nil and page.datafile != nil and page.datafile != '' -%}
+    {%- assign yaml_data = site.data[page.datafolder][page.datafile] -%}
+    {%- if yaml_data.long and yaml_data.long != '' -%}
+        {%- assign page_description = page.title | append: ': ' | append: yaml_data.long | strip_html | strip | truncatewords: 30 -%}
+    {%- elsif yaml_data.short and yaml_data.short != '' -%}
```
thaJeztah, line 21: FWIW, we can swap order of priority here if we want; the short description may be "too short", but the "long" description can contain HTML, tables, etc, so for some pages may contain some irrelevant info.

That's something we could tweak later of course; we can also manually create descriptions in the markdown-stubs (or update the short command description upstream in docker/cli)

thaJeztah:
@usha-mandya @StefanScherer PTAL

StefanScherer (approved):
LGTM

PR #11490, opened 2020-10-07
Add Kubernetes reset note in Docker Desktop release notes

mat007:
### Proposed changes

Adds Kubernetes reset note in Docker Desktop release notes, because we forgot to include it initially.


usha-mandya (approved):
LGTM. Thanks @mat007

PR #11496, opened 2020-10-07
Move open source registry to top-level side-bar navigation

thaJeztah:
This topic was moved under 'open source' because it confused users looking for the Enterprise products (DTR). This should no longer be a problem, so we should be able to move it back for easier discovery.

Also rephrased some TOC titles, and changed a page-title.


thaJeztah:
@usha-mandya @justincormack ptal

usha-mandya:
@thaJeztah The overview page states Docker Registry is open source. Wondering if it makes sense to retain it under Open source?

thaJeztah:
engine and cli are also open source, so not sure we need to make the distinction. Open to suggestions though: perhaps @justincormack has thoughts on this

justincormack:
Please do not do this.

PR #11498, opened 2020-10-07
Minor fixes to glossary

thaJeztah:

PR #11499, opened 2020-10-07
glossary: remove glossary search scripts

thaJeztah:
follow-up to https://github.com/docker/docker.github.io/pull/11498

This was added in https://github.com/docker/docker.github.io/commit/1a6874fcb4e95cd19449b11c2a8f142019820ea6 (https://github.com/docker/docker.github.io/pull/2662), but later got disabled, and now no longer is in use.

We can restore these scripts or re-implement if we ever decide to bring back similar functionality.


usha-mandya (approved):
Agree, makes sense to remove these until we bring back the glossary search functionality. Thank you.

PR #11501, opened 2020-10-07
Add missing relnotes

usha-mandya:

Adding entries that were missing from the Desktop Stable 2.4.0.0 release

### DIFF
--- a/docker-for-windows/release-notes.md
+++ b/docker-for-windows/release-notes.md
@@ -21,63 +21,68 @@ For information about Edge releases, see the [Edge release notes](edge-release-n
 Docker Desktop 2.4.0.0 contains a Kubernetes upgrade. Your local Kubernetes cluster will be reset after installing this version.
 
 ### New
-  - [Docker Compose CLI - 0.1.18](https://github.com/docker/compose-cli), enabling use of volumes with Compose and the Cloud through ECS and ACI.
-  - Docker introduces the new Images view in the Docker Dashboard. The images view allows users to view the Hub images, pull them and manage their local images on disk including cleaning up unwanted and unused images. To access the new Images view, from the Docker menu, select 'Dashboard' > 'Images'.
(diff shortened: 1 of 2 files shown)


docker-for-windows/release-notes.md
```diff
@@ -25,6 +25,7 @@ Docker Desktop 2.4.0.0 contains a Kubernetes upgrade. Your local Kubernetes clus
   - Docker introduces the new Images view in the Docker Dashboard. The images view allows users to view the Hub images, pull them and manage their local images on disk including cleaning up unwanted and unused images. To access the new Images view, from the Docker menu, select 'Dashboard' > 'Images'.
   - Docker Desktop now enables BuildKit by default after a reset to factory defaults. To revert to the old `docker build` experience, go to `Preferences` > `Docker Engine` and then disable the BuildKit feature.
   - [Amazon ECR Credential Helper](https://github.com/awslabs/amazon-ecr-credential-helper/releases/tag/v0.4.0)
+  - Docker Desktop now uses gRPC-FUSE for file sharing by default. This uses much less CPU than osxfs, especially when there are lots of file events on the host. To switch back to `osxfs`, go to **Settings** > **General** and disable gRPC-FUSE.
```
djs55: grpcfuse is not new on Windows in this release -- it was part of stable 2.3.x. I think we probably shouldn't mention the CPU problem here because the file event part of it seemed to be Mac-specific due to very large macOS filesystem event queue lengths. By contrast Windows seems very good at generating queue overflows and dropping events without our help :)

djs55 (changes requested):
I think we should drop the mention of grpcfuse for Windows but keep everything else.

docker-for-windows/release-notes.md
usha-mandya: Updated. Thanks.

stephen-turner (approved):
Thanks, LGTM

djs55 (approved):
LGTM!

PR #11505, opened 2020-10-08
_includes/head.html: remove unused ng-cloak styles

thaJeztah:
These styles seem to be used when using AngularJS;
https://docs.angularjs.org/api/ng/directive/ngCloak

> The ngCloak directive is used to prevent the AngularJS html template from
> being briefly displayed by the browser in its raw (uncompiled) form while
> your application is loading. Use this directive to avoid the undesirable
> flicker effect caused by the html template display.

And I don't think that's used anywhere currently, so let's remove


thaJeztah:
@usha-mandya ptal

usha-mandya (approved):
LGTM. Thank you

PR #11510, opened 2020-10-08
Remove glossary left-overs

thaJeztah:
Looks like removing the front-matter in f17ebae568a15033c914d2024f5e939dee89a4a8
caused the output to break, resulting in a JavaScript error.

Looking at where this file was used, it turned out that it was loaded, but
never used anywhere.

This commit removes the remaining parts of the glossary search functionality,
which was not used.


thaJeztah:
@StefanScherer ptal 🤗

StefanScherer (approved):
LGTM

PR #11511, opened 2020-10-08
Generate top-navigation instead of using javascript

thaJeztah:
This PR makes a couple of changes;

- Use the links defined in the `toc.yaml` both for the landing page and for the rest of the website, to prevent them getting out of sync (there's one exception for the "articles" tab, which is only used on the landing page, and "samples", which is not on the landing page, but is on all other pages
- Generate the header-links when building the site, instead of generate them using javascript. This way the top nav is functional also when javascript is disabled (and could help indexing the website in some cases)
- Some optimisations in the javascript

See individual commit messages for more details


StefanScherer (approved):
LGTM

usha-mandya (approved):

PR #11516, opened 2020-10-09
Move wsl 2 best practices

usha-mandya:

Moved the section below prereqs to emphasize the best practices that users must be aware of.


usha-mandya:
@stephen-turner Could you PTAL?

stephen-turner (changes requested):
Thank you, this looks much better. But I noticed that one of the prerequisites is also out of date:

1. Install Windows 10, version 2004 or higher. The Docker Desktop Edge release also supports Windows 10, version 1903 or higher.

We now have support for 1903 and 1909 on Stable too. Maybe we could change this at the same time.

stephen-turner (approved):
Thanks, LGTM now.

PR #11526, opened 2020-10-12
Add "rel=noopener" and fix canonical links

thaJeztah:
### Add rel="noopener" to external links

(see https://web.dev/external-anchors-use-rel-noopener/) Using noopener, as that addresses the security issue. "noreferer" blocks the REFERER header, which may still be useful for some target URLs.

### Use full URLs for canonical links

Canonical links are expected to be full URLs, not relative.

Screenshot 2020-10-12 at 12 25 09

For this to work, the Dockerfile had to be updated, because we're stripping the domain-name from links ("`<a href...`"), but the script currently also included "`<link rel='canonical' ..`" tags.

With the change, canonical links are left alone;

These hrefs will be replaced

    echo '<a class=foo href="https://docs.docker.com/foo">hello</a>' | sed -e 's#\(<a[^>]* href="\)https://docs.docker.com/#\1/#g'
    # <a class=foo href="/foo">hello</a>
    echo '<a href="https://docs.docker.com/foo">hello</a>' | sed -e 's#\(<a[^>]* href="\)https://docs.docker.com/#\1/#g'
    # <a href="/foo">hello</a>

But, for example, this one is left alone

    echo '<link rel="canonical" href="https://docs.docker.com/foo/bar" />' | sed -e 's#\(<a[^>]* href="?\)https://docs.docker.com/#\1/#g'
    # <link rel="canonical" href="https://docs.docker.com/foo/bar" />

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -54,7 +54,7 @@ COPY --from=upstream-resources /usr/src/app/md_source/. ./
 # API docs, based on the latest_engine_api_version parameter in _config.yml
 RUN ./_scripts/update-api-toc.sh
 RUN jekyll build -d ${TARGET} \
- && find ${TARGET} -type f -name '*.html' | while read i; do sed -i 's#href="https://docs.docker.com/#href="/#g' "$i"; done \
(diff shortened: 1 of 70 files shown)


thaJeztah:

StefanScherer (approved):
Makes sense. 
LGTM

usha-mandya (approved):
LGTM. Thanks.

PR #11535, opened 2020-10-13
Remove osxfs topics from Desktop docs

usha-mandya:


docker-for-mac/osxfs-caching.md
```diff
@@ -1,237 +0,0 @@
...
-   (permit delays before updates on the host appear in the container)
-
- * `delegated`: the container's view is authoritative
-   (permit delays before updates on the container appear in the host)
-
-## Examples
-
-Each of these configurations (`consistent`, `cached`, `delegated`) can be
-specified as a suffix to the
-[`-v`](../engine/reference/run.md#volume-shared-filesystems)
```
thaJeztah, line 98: Are the `consistent`, `cached`,  and `delegated` options no longer supported? Wondering if we document them elsewhere @djs55 ?

thaJeztah: (I know some of these have been a "hot topic", so if we remove the docs about them, we should have a look at either adding to the FAQ ("do I still need `consistent`, `cached`, `delegated` ?)

docker-for-mac/osxfs.md
```diff
@@ -1,369 +0,0 @@
...
-ownership metadata return the previously set values. Ownership-based permissions
-are only enforced at the macOS file system level with all accessing processes
-behaving as the user running Docker. If the user does not have permission to
-read extended attributes on an object (such as when that object's permissions
-are `0000`), `osxfs` attempts to add an access control list (ACL) entry that
-allows the user to read and write extended attributes. If this attempt fails,
-the object appears to be owned by the process accessing it until the extended
-attribute is readable again.
-
-### File system events
```
thaJeztah, line 95: These parts also still seem relevant

docker-for-mac/osxfs.md
```diff
@@ -1,369 +0,0 @@
...
-### Extended attributes
-
-Extended attributes are not yet supported.
-
-### Technology
-
-`osxfs` does not use OSXFUSE. `osxfs` does not run under, inside, or
-between macOS userspace processes and the macOS kernel.
-
-### SSH agent forwarding
```
thaJeztah, line 156: Same for this (which I don't think is specific to osxfs being the implementation)

thaJeztah (commented):
I think we should not remove these pages, but remove mentions of "osxfs" being the implementation, and keep all the descriptions that are generic

stephen-turner (approved):
These pages are causing trouble now, so I would like to remove them immediately. There may be a case for writing a similar page for gRPC FUSE but it will have to be mostly rewritten, which we don't have time to do on the same timescale.

usha-mandya:
Thank you @thaJeztah and @stephen-turner.

I'm merging this PR as this is causing issues to our users. I've created a Desktop ticket to track the docs work required for gRPC implementation.

docker-for-mac/osxfs-caching.md
BretFisher: So is caching not a thing in FUSE driver?

docker-for-mac/osxfs.md
BretFisher: All of this but the name of the extended attribute for FUSE is still true. Why was it removed? Just ran into this last week on latest edge.

BretFisher:
I'm one of the biggest DD fans, but FUSE isn't perfect yet, and simply removing docs without replacements, and giving us nothing to understand the years of osxfs moving to FUSE is frustrating. Spent days with a client on this only to find extended attributes are still a thing, and that caching/delegating "may" be gone...

PR #11541, opened 2020-10-14
Move polldaddy script to an include, and update some options

thaJeztah:
- use a config-variable for the ID
- exclude the script if no ID is set
- set default font and font-color

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -16,6 +16,7 @@ exclude: ["_scripts", "tests", "apidocs/layouts", "Gemfile", "hooks", "index.htm
 # Google Analytics, etc.
 # TODO these should only be set when building for production, or passed as an environment variable when building
 google_analytics: GTM-WL2QLG5
+polldaddy_id: 8453675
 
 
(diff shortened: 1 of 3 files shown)


_config.yml
```diff
@@ -16,6 +16,7 @@ exclude: ["_scripts", "tests", "apidocs/layouts", "Gemfile", "hooks", "index.htm
 # Google Analytics, etc.
 # TODO these should only be set when building for production, or passed as an environment variable when building
 google_analytics: GTM-WL2QLG5
+polldaddy_id: 8453675
```
thaJeztah, line 19: If we don't set these values, the analytics scripts and polldaddy will not be included. We should look if we can disable these when building the (netlify) previews, and when building locally. Not urgent, but something we could look at.

thaJeztah:
@StefanScherer @usha-mandya ptal

_config.yml
usha-mandya: Agree. We don't need the tracking script for Netlify previews and local builds.

PR #11542, opened 2020-10-14
js: move theme switcher to its own script, and use localstorage

thaJeztah:
rewrite the script to not depend on jQuery, so that it can be run as soon as possible.

Also switch to use localstorage instead of cookies, which is a more suitable mechanism for this, and use the same HTML include as was used on the landing-page for the whole site.


js/theme-switcher.js
```diff
@@ -0,0 +1,24 @@
+// Replacement for jQuery $(selector) and $.onready()
+const _ = s => document.querySelector(s);
+const ready = f => document.readyState !== 'loading' ? f() : document.addEventListener('DOMContentLoaded', f)
+
+const darkMode = window.matchMedia && window.matchMedia("(prefers-color-scheme: dark)")
+
+if (darkMode || (window.localStorage && localStorage.getItem("theme") === "night")) {
+    _("body").classList.add("night");
+}
```
thaJeztah, line 9: FWIW; I'm looking at some additional changes, and in a follow up may move this code inline, so that we perform the switch even earlier (preventing the "light flash" when we load).

js/theme-switcher.js
```diff
@@ -0,0 +1,24 @@
+// Replacement for jQuery $(selector) and $.onready()
+const _ = s => document.querySelector(s);
+const ready = f => document.readyState !== 'loading' ? f() : document.addEventListener('DOMContentLoaded', f)
+
+const darkMode = window.matchMedia && window.matchMedia("(prefers-color-scheme: dark)")
+
+if (darkMode || (window.localStorage && localStorage.getItem("theme") === "night")) {
```
thaJeztah, line 7: hm; hold on; looks like I made have made logic change here. guess there's another commit I didn't include; fixing now

thaJeztah:
Ok this is good for review again; got tricked by lighthouse, which by default resets localstorage, so didn't preserve the switch position 😂

thaJeztah:
Rebased

usha-mandya (approved):
LGTM. Thanks.

PR #11560, opened 2020-10-16
Improve font handling for best practices

thaJeztah:
preload fonts and add font-display: swap to prevent flickering;
see: https://www.freecodecamp.org/news/web-fonts-in-2018-f191a48367e8/
we only preload the "woff2" variants, as older formats (woff, eot) are only used
by older browsers, and we don't optimize for those. For fontawesome we "prefetch"
to allow the browser to pre-heat the cache, but not load it if the page does not
use it.


usha-mandya (approved):
LGTM

PR #11572, opened 2020-10-19
Improve Docker Desktop filesharing docs

djs55:
### Proposed changes

I updated the Docker Desktop for Mac filesharing documentation to add the same tips as Windows, now that the implementation is the same on Mac and Windows. I wonder if we should combine Mac and Windows into a "Desktop" section at some point, since more and more of the UX is the same (e.g. the dashboard)

I've also highlighted that  `docker volume`s are better on Desktop for
- raw performance
- Linux filesystem semantics (needed for things like databases)

and so should be used in preference to bind mounts for particular use-cases.

I added some basic text about compose to the general storage section too.


usha-mandya (approved):
LGTM. Thanks for updating these @djs55.

djs55:
Thanks!

PR #11585, opened 2020-10-20
addition of openSUSE

ddemaio:
### Proposed changes
I've provided documentation on how to install docker on openSUSE with some additional info for the index for docker on openSUSE arm

### DIFF
--- a/engine/install/index.md
+++ b/engine/install/index.md
@@ -46,6 +46,7 @@ and architectures:
 | [Fedora](fedora.md)   | [{{ yes }}](fedora.md) |                          | [{{ yes }}](fedora.md) |
 | [Raspbian](debian.md) |                        | [{{ yes }}](debian.md)   | [{{ yes }}](debian.md) |
 | [Ubuntu](ubuntu.md)   | [{{ yes }}](ubuntu.md) | [{{ yes }}](ubuntu.md)   | [{{ yes }}](ubuntu.md) |
+| [openSUSE](opensuse.md)   | [{{ yes }}](opensuse.md) | [{{ yes }}](opensuse.md)   | [{{ yes }}](opensuse.md) |
 
 ##### Other Linux distributions
 
--- a/engine/install/openSUSE.md
+++ b/engine/install/openSUSE.md
@@ -79,9 +79,15 @@ Verify docker is running:
 $ docker version
 
 ```
+Check to see container deamon is running: 
+
+```bash
+$ sudo docker ps
+
+```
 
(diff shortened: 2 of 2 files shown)


ddemaio:
Hi @thaJeztah would you mind merging this to add openSUSE? The checks passed. However, I don't see the opensuse.md in the preview option. Not sure why it's just showing the index.md preview

thaJeztah:
Thanks for contributing!

Unfortunately, I don't think we can merge this PR. Docker currently does not provide packages for openSUSE. While there are packages of docker in openSUSE's package repositories, those packages are maintained by the SUSE distro maintainers and may be built with non-standard configuration or from different source code as the packages we provide, and for which we cannot provide support.

Adding instructions for installing the distro packages in our documentation would raise the incorrect expectation that Docker provides support for those, which would not be the case.

That said, we are in the process of reviewing our build-matrix; we _did_ have packages for openSUSE and SUSE as part of Docker Enterprise, and some of the distros supported by Docker Enterprise may find their way back in our list of packages.

While those changes are not yet in place, we prefer users to follow the instructions under [Other Linux distributions](https://docs.docker.com/engine/install/#other-linux-distributions)

ddemaio:
> Thanks for contributing!
Thank you @thaJeztah 
> 
> That said, we are in the process of reviewing our build-matrix; we _did_ have packages for openSUSE and SUSE as part of Docker Enterprise, and some of the distros supported by Docker Enterprise may find their way back in our list of packages.
> 
Thanks for the info. I notice that Fedora is there with a 6 month release cycle, where openSUSE Leap is about a year cycle. Tumbleweed is a rolling release, so no sure how useful that would be for Docker, but perhaps it is for Kubernetes. SUSE packages it as well as openSUSE for Leap and Tumbleweed. https://software.opensuse.org/package/docker

 Leap and SUSE align, but your right that perhaps maybe the binaries are not the same with SUSE. If It's sources, then they will be the same. Just an FYI, for openSUSE Leap 15.3, it looks like it will be source and binary compatible with SUSE, so maybe you could share that info in the review of the build-matrix.

crydotsnake:
I use the Docker version from the openSUSE repository. However, a official docker version would be a loooot better..

thaJeztah:
>  However, a official docker version would be a loooot better..

The best way to get that prioritized, and get visibliity is to open a ticket in our public roadmap; https://github.com/docker/roadmap

Given that we currently don't have official packages for (open)SUSE, I'm closing this PR for now, but thanks again for contributing!

PR #11604, opened 2020-10-22
Less overrides

thaJeztah:
- remove unused "tree" metadata
- preload OpenSans font to prevent redraws
- update reading-time widget;
    - cleanup html and styles
    - skip reading time if it's a minute or less; the reading-time indication was not adding much value on short pages. The reader is likely able to make a better estimation how long it would take to read.
    - disable reading-time from command-line reference through configuration, instead of setting it on each page/stub
- removed some unneeded css rules that complicated the styles, and caused inconsistencies in font color, size and line-spacing
- removed "image with border" style, which was used for a single image only, and was barely visible


thaJeztah:
@usha-mandya ptal

usha-mandya:
@thaJeztah The page https://deploy-preview-11604--docsdocker.netlify.app/engine/deprecated/ appears broken in the Netlify preview. It looks good on the live site and on https://docs-stage.docker.com/engine/deprecated/. Not sure if this is related to the changes introduced in this PR.

thaJeztah:
That's expected; the "deprecated.md" page is an "upstream resource", and not built by Netlify

usha-mandya (approved):
LGTM.

PR #11644, opened 2020-10-27
docs: update supported Fedora versions

rafauke:
### Proposed changes

Update supported Fedora versions. Since Fedora 33 was released on 27.10.2020 I think it is a good idea to keep the documentation up-to-date. Ideally, this merge could go after https://github.com/docker/for-linux/issues/1114 is merged.


thaJeztah (approved):
LGTM

Issue #11646, opened 2020-10-27
Wrong Commands for Specifying Protocol in Port Publishing at Ingress Tutorial


mahs4d:
### Problem description

the commands provided for "Publish a port for TCP only or UDP only" are wrong and do not match their respective titles.

### Problem location

https://docs.docker.com/engine/swarm/ingress/#publish-a-port-for-tcp-only-or-udp-only

PR #11647, opened 2020-10-27
Update ingress.md

mahs4d:
titles for "Publish a port for TCP only or UDP only" subsections didn't match their corresponding codes.

### Proposed changes

in swarm mode tutorial -> ingress page, the codes and titles under "publish a port for tcp only or udp only" section didn't relate to each other. i proposed to fix this issue

### Related issues (optional)

https://github.com/docker/docker.github.io/issues/11646

### DIFF
--- a/engine/swarm/ingress.md
+++ b/engine/swarm/ingress.md
@@ -101,24 +101,6 @@ you publish both TCP and UDP ports, If you omit the protocol specifier,
 the port is published as a TCP port. If you use the longer syntax (recommended),
 set the `protocol` key to either `tcp` or `udp`.
 
-#### TCP only
-
-**Long syntax:**
-
-```bash
-$ docker service create --name dns-cache \
-  --publish published=53,target=53 \
-  dns-cache
-```
-
-**Short syntax:**
-
-```bash
-$ docker service create --name dns-cache \
-  -p 53:53 \
-  dns-cache
-```
-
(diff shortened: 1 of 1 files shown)


engine/swarm/ingress.md
```diff
@@ -101,7 +101,7 @@ you publish both TCP and UDP ports, If you omit the protocol specifier,
 the port is published as a TCP port. If you use the longer syntax (recommended),
 set the `protocol` key to either `tcp` or `udp`.
 
-#### TCP only
+#### TCP and UDP
```
usha-mandya, line 104: @mahs4d The existing heading looks correct as the following example demonstrates the default behavior, which is TCP only.

mahs4d: @usha-mandya you are right, i didn't know the default behavior :) i'll remove the PR

engine/swarm/ingress.md
```diff
@@ -119,14 +119,14 @@ $ docker service create --name dns-cache \
...
 ```
 
-#### TCP and UDP
+#### TCP only
 
 **Long syntax:**
 
 ```bash
 $ docker service create --name dns-cache \
   --publish published=53,target=53 \
```
usha-mandya, line 128: The heading TCP and UDP is correct for this section as L128 is for TCP (which is the default) and 129 is to set UDP

PR #11648, opened 2020-10-27
Update Hub pull rate limit information

binman-docker:
### Proposed changes

Adds further details on Hub pull rate limiting - link to FAQ, new error message, and info on checking headers.

Please do not merge until later this week, headers will not be live until the limits roll out. I will mark as Draft for now.

### DIFF
--- a/docker-hub/download-rate-limit.md
+++ b/docker-hub/download-rate-limit.md
@@ -54,7 +54,7 @@ To get a token with a user account (if you are authenticating your pulls) - inse
 
 > $ TOKEN=$(curl --user 'username:password' "https://auth.docker.io/token?service=registry-1.docker.io&scope=repository:ratelimitpreview/test:pull" | jq -r .token)
 
-Then to get the headers showing your limits, run this:
+Then to get the headers showing your limits, run this (keep in mind that requesting a manifest emulates a pull and will count against the limits):
 
 >$ curl -v -H "Authorization: Bearer $TOKEN" https://registry-1.docker.io/v2/ratelimitpreview/test/manifests/latest 2>&1 | grep RateLimit
 
--- a/docker-hub/download-rate-limit.md
+++ b/docker-hub/download-rate-limit.md
@@ -6,7 +6,7 @@ title: Download rate limit
 
 Docker has enabled download rate limits for pull requests on 
 Docker Hub. Limits are determined based on the account type. 
-For more information, see [Docker Hub Pricing](https://hub.docker.com/pricing){: target="_blank" rel="noopener" class="_"} and our [Resource Consumption FAQs](https://www.docker.com/pricing/resource-consumption-updates){: target="_blank" rel="noopener" class="_"}.
+For more information, see [Resource Consumption FAQs](https://www.docker.com/pricing/resource-consumption-updates){: target="_blank" rel="noopener" class="_"} and [Docker Hub Pricing](https://hub.docker.com/pricing){: target="_blank" rel="noopener" class="_"}.
 
 A user's limit will be equal to the highest entitlement of their
 personal account or any organization they belong to. To take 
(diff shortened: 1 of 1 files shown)


docker-hub/download-rate-limit.md
```diff
@@ -26,8 +26,46 @@ manifest requests.
...
+## How do I know my pull requests are being limited
+
+When you issue a pull request and you are over the limit for your account type, Hub will return a `429` response code with the following body when the manifest is requested:
+> You have reached your pull rate limit. You may increase the limit by authenticating and upgrading: https://www.docker.com/increase-rate-limits
+
+You will see this error message in the Docker CLI or Docker Engine logs.
+
+## How can I check my current rate
+
+Valid, non-rate-limited manifest API reqests to Hub will include the following rate limit headers in the response:
```
ingshtrom, line 40: > non-rate-limited manifest API requests to Hub
In testing we only return these headers for GET requests. Maybe we should return them for HEAD requests as well so that users can get information without increasing their rate limiting counter?

docker-hub/download-rate-limit.md
```diff
@@ -26,8 +26,46 @@ manifest requests.
...
+
+> RateLimit-Limit    
+RateLimit-Remaining
```
ingshtrom, line 43: Is it too much information to post also that we got this from the current IETF draft spec for rate limiting? Keep in mind that the current revision is set to expire on 11/27, so as it evolves we would want to update our implementation and/or update the docs--or just always point to the current revision we use.

docker-hub/download-rate-limit.md
binman-docker: For sure, let's file that as an improvement post-launch.

docker-hub/download-rate-limit.md
```diff
@@ -26,8 +26,46 @@ manifest requests.
...
+
+If you have a proxy or other layer in place that logs your requests, you can inspect the headers of these responses directly.
+
+Otherwise, you can use curl to view these. You will need `curl`, `grep`, and `jq` installed.
+
+To get a token anonymously (if you are pulling anonymously):
+
+> $ TOKEN=$(curl "https://auth.docker.io/token?service=registry.docker.io&scope=repository:ratelimitpreview/test:pull" | jq -r .token)
```
ingshtrom, line 51:
```suggestion
> $ TOKEN=$(curl "https://auth.docker.io/token?service=registry-1.docker.io&scope=repository:ratelimitpreview/test:pull" | jq -r .token)
```

docker-hub/download-rate-limit.md
```diff
@@ -26,8 +26,46 @@ manifest requests.
...
+
+To get a token with a user account (if you are authenticating your pulls) - insert your username and password in the following command:
+
+> $ TOKEN=$(curl --user 'username:password' "https://auth.docker.io/token?service=registry.docker.io&scope=repository:ratelimitpreview/test:pull" | jq -r .token)
```
ingshtrom, line 55:
```suggestion
> $ TOKEN=$(curl --user 'username:password' "https://auth.docker.io/token?service=registry-1.docker.io&scope=repository:ratelimitpreview/test:pull" | jq -r .token)
```

ingshtrom (changes requested):
Some questions and a few requested changes

docker-hub/download-rate-limit.md
binman-docker: I wouldn't want to just link to the spec, since we want to make this easy and succinct for users. We can certainly mention it's based on that spec though.

docker-hub/download-rate-limit.md
shawnaxsom: There's a typo here, "reqests"

docker-hub/download-rate-limit.md
```diff
@@ -26,8 +26,46 @@ manifest requests.
...
+>$ curl -v -H "Authorization: Bearer $TOKEN" https://registry-1.docker.io/v2/ratelimitpreview/test/manifests/latest 2>&1 | grep RateLimit
+
+Which should return something like this:
+
+> < RateLimit-Limit: 100;w=21600  
+< RateLimit-Remaining: 76;w=21600
+
+This means my limit is 100 per 21600 seconds (6 hours), and I have 76 pulls remaining.
+
+Please remember that these headers are best-effort and there will be small variations.
```
shawnaxsom, line 68: We should have a disclaimer that this ends up using up your pulls every 2nd request. (Is that intentional?)

shawnaxsom (changes requested):
A minor typo needs fixed, and I also think we should mention that the pull rate checks will use up the pull rate too. Otherwise some users may build automation that checks the headers every pull in their CI and not realize why their pulls ran out faster than expected.

docker-hub/download-rate-limit.md
ingshtrom:
```suggestion
Valid, non-rate-limited manifest API requests to Hub will include the following rate limit headers in the response:
```

docker-hub/download-rate-limit.md
ingshtrom: It is intentional because it is actually counting manifest requests and multi-architecture images can result in two manifests requests to actually pull the image, so we double the counter to make up for it.

shawnaxsom (dismissed):

usha-mandya (approved):

ingshtrom (approved):

PR #11649, opened 2020-10-28
Update ACI documentation

gtardif:
Signed-off-by: Guillaume Tardif <guillaume.tardif@docker.com>

### Proposed changes

Update ACI integration docs (moving info from https://github.com/docker/compose-cli/blob/main/docs/aci-integration.md)

### DIFF
--- a/engine/context/aci-integration.md
+++ b/engine/context/aci-integration.md
@@ -131,7 +131,8 @@ Containers can be removed using `docker rm`. Removing a running container will r
 
 > **Note**
 >
-The semantics of restarting a container on ACI are different to those when using a moby type context for local development. On ACI, the container will be reset to its initial state and started on a new node. This includes the container's filesystem so all state that is not stored in a volume will be lost on restart.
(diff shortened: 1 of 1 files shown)


engine/context/aci-integration.md
```diff
@@ -127,14 +126,18 @@ To stop and remove a container from ACI, run:
...
 ```
+
+Containers can be removed using `docker rm`. Removing a running container will require using the `--force` flag, or stopping the container with `docker stop` before removing.
+
 > **Note**
-> 
-> The stop command in ACI differs from the Moby stop command as a stopped 
-container will not retain its state when it is started again on ACI.
+>
+The semantics of restarting a container on ACI are different to those when using a moby type context for local development. On ACI, the container will be reset to its initial state and started on a new node. This includes the container's filesystem so all state that is not stored in a volume will be lost on restart.
```
chris-crone, line 134: nit: Not sure all users will know what moby is

```suggestion
The semantics of restarting a container on ACI are different to those when using a local Docker context for local development. On ACI, the container will be reset to its initial state and started on a new node. This includes the container's filesystem so all state that is not stored in a volume will be lost on restart.

```

chris-crone (dismissed):
LGTM

jt78-collab (approved):

usha-mandya (approved):
LGTM

PR #11652, opened 2020-10-28
Publish updates from master

usha-mandya:

### DIFF
--- a/engine/context/aci-integration.md
+++ b/engine/context/aci-integration.md
@@ -49,24 +49,28 @@ docker login azure
 ```
 
 This opens your web browser and prompts you to enter your Azure login credentials.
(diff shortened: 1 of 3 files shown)


engine/context/aci-integration.md
```diff
@@ -127,14 +126,18 @@ To stop and remove a container from ACI, run:
...
 ```
+
+You can remove containers using `docker rm`. To remove a running container, you must use the `--force` flag, or stop the container using `docker stop` before removing it.
+
 > **Note**
-> 
-> The stop command in ACI differs from the Moby stop command as a stopped 
-container will not retain its state when it is started again on ACI.
+>
+The semantics of restarting a container on ACI are different to those when using a local Docker context for local development. On ACI, the container will be reset to its initial state and started on a new node. This includes the container's filesystem so all state that is not stored in a volume will be lost on restart.
```
thaJeztah, line 134: is this missing a `> `?

usha-mandya: Aah, good catch. Fixed in https://github.com/docker/docker.github.io/pull/11654

thaJeztah (approved):
LGTM

PR #11670, opened 2020-10-31
Fix broken link in storage/volumes.md

daido1976:
### Proposed changes

Fix broken link in https://docs.docker.com/storage/volumes/#use-a-volume-with-docker-compose.


usha-mandya (approved):
LGTM. Thank you @daido1976 :)

PR #11671, opened 2020-11-01
Update up.md

gdennie:
An option's short form should not be used to identify them in descriptions since these are intentionally terse and as such don't provide the otherwise nice narrative and intuition of the long form.

I have replaced the short form with their long forms in this documentation. That includes the `docker-compose logs --f` out of band referent.


usha-mandya (approved):
@gdennie Thanks for the update. LGTM

PR #11086, opened 2020-07-03
Embed video tutorials within docs

usha-mandya:
Embed YouTube video tutorials within docs instead of pointing users to an external page.

PR #11682, opened 2020-11-02
docker-hub: remove unused images

thaJeztah:
These images were not used anywhere, so we might as well remove them.

I was confused for a bit about the video-thumbs being unused, but they were no longer needed since we now directly embed the video (https://github.com/docker/docker.github.io/pull/11086)


usha-mandya (approved):
LGTM. Thanks for making the repository lighter ;)

PR #11685, opened 2020-11-02
small addition to document expected behavior of parallel autobuilds

ingshtrom:
### Proposed changes

There was no documentation on the expected behavior of parallel builds, which is a line item on our pricing page.

PR #11688, opened 2020-11-02
parallel builds docs 2

ingshtrom:
### Proposed changes

I wanted to include [this change](https://github.com/docker/docker.github.io/pull/11685#discussion_r516102789), but saw it too late.

**I would also like to ensure that @arkodg approves this prior to merging--sorry I didn't make that clear before.** I know that main content is already there from #11685, but if I'm not 100% correct, he can fix that now 😄  

Thank you for the initial feedback @thaJeztah!


ingshtrom:
O.o I guess those changes did make it in. Sorry!

PR #11697, opened 2020-11-03
Straightforward typofix s/names/name

ryanmaclean:
Should be `name` here as it is an instruction.


usha-mandya:
@ryanmaclean Thanks for the update. Unfortunately, we'll have to close this PR without merging the suggested update as the existing text is grammatically correct.

PR #11698, opened 2020-11-04
engine/examples: remove sshd example

AkihiroSuda:
### Proposed changes

The ssh example was really bad; it even leaks the root password as a plain text in the image :(

This commit removes the sshd example and discourage people to dockerize sshd.

The `running_ssh_service.md` page itself is not removed because there
are hyperlinks on the Internet to that page.


justincormack (approved):

usha-mandya:
Thank you @AkihiroSuda.

PR #11701, opened 2020-11-04
Update systemd.md

kfortner:
Systemctl requires different escape characters than a URL does so a proxy var with special URL characters needs to have them escaped with the hex equivalent and the systemctl escapes.

A backslash in domain\user becomes domain%5Cuser but this breaks systemctl because of the % sign.

If you need to use systemctl with a url and special chars is domain%%5Cuser (Notice the DOUBLE %).

[Service]
Environment="HTTP_PROXY=http://domain%%5Cuser:complex%%23pass@proxy.example.com:8080/"

### DIFF
--- a/config/daemon/systemd.md
+++ b/config/daemon/systemd.md
@@ -119,6 +119,13 @@ you need to add this configuration in the Docker systemd service file.
     Environment="HTTP_PROXY=http://proxy.example.com:80"
     Environment="HTTPS_PROXY=https://proxy.example.com:443"
     ```
+        
+    If you have to use special characters in the proxy value they must be both URL encoded and escape the % signs with double %%.
+    
+    ```
+    [Service]
+    Environment="HTTP_PROXY=http://domain%%5Cuser:complex%%23pass@proxy.example.com:8080/"
+    ```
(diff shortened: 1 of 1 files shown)


shiguowang:
I meet this problem +1, It's really hard for me to figure out we need double % , please merge it if it's ok.

OliverScherf:
Why doensn't this get  merged? I wasted some time on this one.

dvdksn (approved):

dvdksn:
Thanks @kfortner!!

abhi10agarwal:
I am having the same issue docker cannot escape the special character '@' in the proxy.

Docker version v25.0.0
OS : RHEL 8.9

Please assist.

kfortner:
> I am having the same issue docker cannot escape the special character '@' in the proxy.
> 
> Docker version v25.0.0 OS : RHEL 8.9
> 
> Please assist.

I have since found a tool called systemd-escape that will build an escaped string. Take a look at the example below with the @ sign escaped.

```
systemd-escape testpwdwith@sign
testpwdwith\x40sign
```

PR #11707, opened 2020-11-05
Update rate limit examples to use HEAD

binman-docker:
### Proposed changes

We have enabled checking ratelimits via HEAD requests, which don't count against your limit, so I'm updating the docs to use HEAD rather than GET.

### DIFF
--- a/docker-hub/download-rate-limit.md
+++ b/docker-hub/download-rate-limit.md
@@ -40,7 +40,7 @@ You will see this error message in the Docker CLI or in the Docker Engine logs.
 
 ## How can I check my current rate
 
-Valid non-rate-limited manifest API requests to Hub will include the following rate limit headers in the response:
+Valid manifest API requests to Hub will usually include the following rate limit headers in the response:
 
 ```
 RateLimit-Limit    
@@ -64,7 +64,7 @@ $ TOKEN=$(curl --user 'username:password' "https://auth.docker.io/token?service=
 Then to get the headers showing your limits, run the following:
 
 ```
-$ curl -I -H "Authorization: Bearer $TOKEN" https://registry-1.docker.io/v2/ratelimitpreview/test/manifests/latest 2>&1 | grep RateLimit
+$ curl --head -H "Authorization: Bearer $TOKEN" https://registry-1.docker.io/v2/ratelimitpreview/test/manifests/latest 2>&1 | grep RateLimit || echo "No RateLimit headers found (pull not limited)"
 ```
 
 Which should return something like this:
@@ -77,6 +77,10 @@ RateLimit-Remaining: 76;w=21600
 This means my limit is 100 per 21600 seconds (6 hours), and I have 76 pulls remaining.
 
 > Remember that these headers are best-effort and there will be small variations.
+
+### I don't see any RateLimit headers
+
+If you do not see these headers, that means pulling that image would not count towards limiting. This could be because you are authenticated with a user associated with a Legacy/Pro/Team Docker Hub account, or because the image or your IP is unlimited in partnership with a publisher, provider, or open source organization.
 
 ## How do I authenticate pull requests
 
--- a/docker-hub/download-rate-limit.md
+++ b/docker-hub/download-rate-limit.md
@@ -40,14 +40,14 @@ You will see this error message in the Docker CLI or in the Docker Engine logs.
 
 ## How can I check my current rate
 
-When limiting starts, valid non-rate-limited manifest API requests to Hub will include the following rate limit headers in the response:
+Valid non-rate-limited manifest API requests to Hub will include the following rate limit headers in the response:
 
 ```
 RateLimit-Limit    
 RateLimit-Remaining
 ```
 
-If you have a proxy or other layer in place that logs your requests, you can inspect the headers of these responses directly. Otherwise, you can use curl to view these. You will need `curl`, `grep`, and `jq` installed.
+These headers will be returned on both GET and HEAD requests. Note that using GET emulates a real pull and will count towards the limit; using HEAD will not. To check your limits, you will need `curl`, `grep`, and `jq` installed.
 
 To get a token anonymously (if you are pulling anonymously):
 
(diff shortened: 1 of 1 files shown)


ingshtrom (dismissed):

docker-hub/download-rate-limit.md
```diff
@@ -40,14 +40,14 @@ You will see this error message in the Docker CLI or in the Docker Engine logs.
 
 ## How can I check my current rate
 
-When limiting starts, valid non-rate-limited manifest API requests to Hub will include the following rate limit headers in the response:
+Valid non-rate-limited manifest API requests to Hub will include the following rate limit headers in the response:
```
shawnaxsom, line 43: Is this clearly worded? "Non-rate-limited" almost sounds like it could be talking about if you are an unlimited pulls user.

docker-hub/download-rate-limit.md
```diff
@@ -61,17 +61,17 @@ To get a token with a user account (if you are authenticating your pulls) - don'
...
+$ curl -I -H "Authorization: Bearer $TOKEN" https://registry-1.docker.io/v2/ratelimitpreview/test/manifests/latest 2>&1 | grep RateLimit
 ```
 
 Which should return something like this:
 
 ```
-< RateLimit-Limit: 100;w=21600  
-< RateLimit-Remaining: 76;w=21600
+RateLimit-Limit: 100;w=21600
+RateLimit-Remaining: 76;w=21600
```
shawnaxsom, line 74: Maybe add a comment after this to be clear that it is expected that if you are authenticated as pro/team/legacy, you won't see these headers (I recall GitHub or CircleCI had to ask us to make sure)

docker-hub/download-rate-limit.md
```diff
@@ -61,17 +61,17 @@ To get a token with a user account (if you are authenticating your pulls) - don'
 $ TOKEN=$(curl --user 'username:password' "https://auth.docker.io/token?service=registry.docker.io&scope=repository:ratelimitpreview/test:pull" | jq -r .token)
 ```
 
-Then to get the headers showing your limits, run the following (keep in mind that requesting a manifest emulates a pull and will count against the limits):
+Then to get the headers showing your limits, run the following:
 
 ```
-$ curl -v -H "Authorization: Bearer $TOKEN" https://registry-1.docker.io/v2/ratelimitpreview/test/manifests/latest 2>&1 | grep RateLimit
+$ curl -I -H "Authorization: Bearer $TOKEN" https://registry-1.docker.io/v2/ratelimitpreview/test/manifests/latest 2>&1 | grep RateLimit
```
shawnaxsom, line 67: Consider using `--head` rather than `-I`, it's more explicit to the readers that it is a head request

shawnaxsom: What if we print out something informational if the RateLimit info isn't found?

This seems to work:

```
curl --head -H "Authorization: Bearer $TOKEN" https://registry-1.docker.io/v2/ratelimitpreview/test/manifests/latest | grep RateLimit || echo "No headers found (unlimited pulls)"

```

shawnaxsom (commented):
Some suggestions, no blockers

docker-hub/download-rate-limit.md
```diff
@@ -61,23 +61,27 @@ To get a token with a user account (if you are authenticating your pulls) - don'
...
+RateLimit-Remaining: 76;w=21600
 ```
 
 This means my limit is 100 per 21600 seconds (6 hours), and I have 76 pulls remaining.
 
 > Remember that these headers are best-effort and there will be small variations.
 
+### I don't see any RateLimit headers
+
+If you do not see these headers, that means pulling that image would not count towards limiting. This could be because you are authenticated with a user associated with a Legacy/Pro/Team Docker Hub account, or because the image or your IP is unlimited in partnership with a publisher, provider, or open source organization.
```
shawnaxsom, line 83: Minor note, "towards limiting", consider reminding that we're not talking about anti-abuse limits here?

docker-hub/download-rate-limit.md
```diff
@@ -61,23 +61,27 @@ To get a token with a user account (if you are authenticating your pulls) - don'
 $ TOKEN=$(curl --user 'username:password' "https://auth.docker.io/token?service=registry.docker.io&scope=repository:ratelimitpreview/test:pull" | jq -r .token)
 ```
 
-Then to get the headers showing your limits, run the following (keep in mind that requesting a manifest emulates a pull and will count against the limits):
+Then to get the headers showing your limits, run the following:
 
 ```
-$ curl -v -H "Authorization: Bearer $TOKEN" https://registry-1.docker.io/v2/ratelimitpreview/test/manifests/latest 2>&1 | grep RateLimit
+$ curl --head -H "Authorization: Bearer $TOKEN" https://registry-1.docker.io/v2/ratelimitpreview/test/manifests/latest 2>&1 | grep RateLimit || echo "No RateLimit headers found (pull not limited)"
```
shawnaxsom, line 67: One concern here:

```
curl --head -H "Authorization: Bearer $TOKEN" https://registry-1.docker.io/v2/ratelimitpreview/test/manifests/latest 2>&1 | grep RateLimit
```

If $TOKEN isn't set (or invalid), we're eating any errors.

We are hiding the fact that it returned:

```
HTTP/1.1 401 Unauthorized
```

shawnaxsom (changes requested):
One last concern, let me know your thoughts

docker-hub/download-rate-limit.md
binman-docker: Yeah that's true. I think it's better just to remove it and not try to be fancy, and rely on the next session talking about the lack of headers.

shawnaxsom: This seems to work:

```
curl --head -H "Authorization: Bearer $TOKEN" https://registry-1.docker.io/v2/ratelimitpreview/test/manifests/latest 2>&1 | grep "RateLimit\|HTTP.*[3-9][0-9][0-9]\|resolve" || echo "No RateLimit headers found (pull not limited)"
```

Checking for 300-500 errors (which can happen if $TOKEN is unset/invalid), or for "Could not resolve host"

binman-docker: Shawn and I discussed in Slack - better to remove it than give a complicated example that takes into account all possibilities. If people don't see headers they will hopefully read the next section a couple lines down.

docker-hub/download-rate-limit.md
```diff
@@ -61,23 +61,27 @@ To get a token with a user account (if you are authenticating your pulls) - don'
 $ TOKEN=$(curl --user 'username:password' "https://auth.docker.io/token?service=registry.docker.io&scope=repository:ratelimitpreview/test:pull" | jq -r .token)
 ```
 
-Then to get the headers showing your limits, run the following (keep in mind that requesting a manifest emulates a pull and will count against the limits):
+Then to get the headers showing your limits, run the following:
 
 ```
-$ curl -v -H "Authorization: Bearer $TOKEN" https://registry-1.docker.io/v2/ratelimitpreview/test/manifests/latest 2>&1 | grep RateLimit
+$ curl --head -H "Authorization: Bearer $TOKEN" https://registry-1.docker.io/v2/ratelimitpreview/test/manifests/latest 2>&1 | grep RateLimit
```
shawnaxsom, line 67: I think you still need to remove the grep here, changing it to:

```
curl --head -H "Authorization: Bearer $TOKEN" https://registry-1.docker.io/v2/ratelimitpreview/test/manifests/latest 2>&1
```

shawnaxsom (changes requested):
I think the curl is still eating errors isn't it, remove the pipe to grep?

shawnaxsom (approved):

usha-mandya (approved):
Thank you @binman-docker @ingshtrom @shawnaxsom

PR #11711, opened 2020-11-06
command error

aliray:
"volume-opt=o=10.0.0.10,rw,nfsvers=4,async" should be  "volume-opt=o=addr=10.0.0.10,rw,nfsvers=4,async"


usha-mandya (approved):
LGTM. Thanks

PR #11717, opened 2020-11-08
help user view the container running

carehart:
The current instructions lead the user through creating a context and running a sample nginx container, but it never shows the user how to go about viewing it. I have added that.

If anyone sees a need to change/improve the formatting of my proposed text, to some style used here, feel free to do that.

### DIFF
--- a/engine/context/aci-integration.md
+++ b/engine/context/aci-integration.md
@@ -103,7 +103,7 @@ docker run -p 80:80 nginx
 
 After you've switched to the `myacicontext` context, you can use `docker ps` to list your containers running on ACI.
 
-In the case of the demonstration nginx container you started above, you will see the result of the ps command showing a value labelled "PORTS" which will show both an IP address and port on which the container is running. For example, it may show `52.154.202.35:80->80/tcp`, in which case you can view the nginx welcome page by browsing `http://52.154.202.35`.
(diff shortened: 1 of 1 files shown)


carehart:
I found that some of what I mentioned (finding the IP address via the ps output) is discussed in the later section, "Exposing ports". Again feel free to revise what I proposed to mention that however you may prefer, to suit your preferences for making such related content references. Or perhaps you may feel it's not needed in this case.

engine/context/aci-integration.md
```diff
@@ -103,6 +103,8 @@ docker run -p 80:80 nginx
 
 After you've switched to the `myacicontext` context, you can use `docker ps` to list your containers running on ACI.
 
+In the case of the demonstration nginx container you started above, you will see the result of the ps command showing a value labelled "PORTS" which will show both an IP address and port on which the container is running. For example, it may show `52.154.202.35:80->80/tcp`, in which case you can view the nginx welcome page by browsing `http://52.154.202.35`.
+
```
gtardif, line 107:
```suggestion
In the case of the demonstration nginx container started above, the result of the ps command will display in column "PORTS" the IP address and port on which the container is running. For example, it may show `52.154.202.35:80->80/tcp`, and you can view the nginx welcome page by browsing `http://52.154.202.35`.

```

gtardif (approved):
LGTM, just a little rephrasing suggestion...

engine/context/aci-integration.md
carehart: Great revision. Thanks.

gtardif (approved):

PR #11729, opened 2020-11-10
Added health check docs, & details on updating apps

gtardif:
Signed-off-by: Guillaume Tardif <guillaume.tardif@docker.com>

### DIFF
--- a/engine/context/aci-integration.md
+++ b/engine/context/aci-integration.md
@@ -152,11 +152,14 @@ Name resolution between containers is achieved by writing service names in the `
 
 ## Updating applications
 
-From a deployed Compose application, you can update the application by re-deploying it with the same ACI group name: `docker compose up --project-name` with the Compose project name already used in previous deployments.
+From a deployed Compose application, you can update the application by re-deploying it with the same project name: `docker compose up --project-name PROJECT`.
 
-Updating an application means the ACI node will be reused, and the application will keep the IP address already allocated to expose ports, if any. ACI also has some limitations on what can be updated in an existing application (you will not be able to change CPU/memory reservation for example), in these cases you need to deploy a new application from scratch.
 
-Updating is the default behaviour if you invoke several times `docker compose up` on the same compose file, since the Compose project name is derived from the compose file location by default. You need to explicitely execute `docker compose down` before running again `docker compose up` in order to totally reset a Compose application.
+Updating an application means the ACI node will be reused, and the application will keep the same IP address that was previously allocated to expose ports, if any. ACI has some limitations on what can be updated in an existing application (you will not be able to change CPU/memory reservation for example), in these cases you need to deploy a new application from scratch.
+
+
+Updating is the default behavior if you invoke `docker compose up` on the an already deployed Compose file, since the Compose project name is derived from the directory where the Compose file is located by default. You need to explicitly execute `docker compose down` before running `docker compose up` again in order to totally reset a Compose application.
+
 
 ## Releasing resources
 
@@ -302,9 +305,12 @@ For Compose applications, you can specify the environment variables in the Compo
 
 ## Healthchecks
 
-You can specify container healthchecks either with `docker run --healthcheck-xxx` flags, or in a compose file with the healthcheck section.
-Healthchecks are converted to ACI LivenessProbe. ACI will run the healthcheck command periodically, and when it fails, the container will be terminated.
-Healthchecks must be used in addition to restart policies, to ensure the container is then restarted ; using no restart policy will lead to containers being killed but not restarted by healthchecks.
+You can specify a container health checks using either the `--healthcheck-` prefixed. flags with `docker run`, or in a Compose file with the healthcheck section of the service.
+
+Health checks are converted to ACI `LivenessProbe`s. ACI will run the health check command periodically, and if it fails, the container will be terminated.
+
+Health checks must be used in addition to restart policies, to ensure the container is then restarted on termination. The default restart policy for `docker run` is `no` which will not restart the container. The default restart policy for Compose is `any` which will always try restarting the service containers.
+
 
 Example using `docker run`:
 
@@ -312,7 +318,8 @@ Example using `docker run`:
 docker --context acicontext run -p 80:80 --restart always --health-cmd "curl http://localhost:80" --health-interval 3s  nginx
 ```
 
-Example using compose files:
+Example using Compose files:
+
 
 ```yaml
 services:
--- a/engine/context/aci-integration.md
+++ b/engine/context/aci-integration.md
@@ -154,12 +154,9 @@ Name resolution between containers is achieved by writing service names in the `
 
 From a deployed Compose application, you can update the application by re-deploying it with the same project name: `docker compose up --project-name PROJECT`.
 
+Updating an application means the ACI node will be reused, and the application will keep the same IP address that was previously allocated to expose ports, if any. ACI has some limitations on what can be updated in an existing application (you will not be able to change CPU/memory reservation for example), in these cases, you need to deploy a new application from scratch.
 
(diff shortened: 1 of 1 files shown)


engine/context/aci-integration.md
```diff
@@ -150,12 +150,20 @@ Name resolution between containers is achieved by writing service names in the `
 >
 > The current Docker Azure integration does not allow fetching a combined log stream from all the containers that make up the Compose application.
 
+## Updating applications
+
+From a deployed Compose application, you can update the application by re-deploying it with the same ACI group name: `docker compose up --project-name` with the Compose project name already used in previous deployments.
```
chris-crone, line 155:
```suggestion
From a deployed Compose application, you can update the application by re-deploying it with the same project name: `docker compose up --project-name PROJECT`.

```

engine/context/aci-integration.md
```diff
@@ -150,12 +150,20 @@ Name resolution between containers is achieved by writing service names in the `
 >
 > The current Docker Azure integration does not allow fetching a combined log stream from all the containers that make up the Compose application.
 
+## Updating applications
+
+From a deployed Compose application, you can update the application by re-deploying it with the same ACI group name: `docker compose up --project-name` with the Compose project name already used in previous deployments.
+
+Updating an application means the ACI node will be reused, and the application will keep the IP address already allocated to expose ports, if any. ACI also has some limitations on what can be updated in an existing application (you will not be able to change CPU/memory reservation for example), in these cases you need to deploy a new application from scratch.
```
chris-crone, line 157:
```suggestion
Updating an application means the ACI node will be reused, and the application will keep the same IP address that was previously allocated to expose ports, if any. ACI has some limitations on what can be updated in an existing application (you will not be able to change CPU/memory reservation for example), in these cases you need to deploy a new application from scratch.

```

engine/context/aci-integration.md
```diff
@@ -150,12 +150,20 @@ Name resolution between containers is achieved by writing service names in the `
 >
 > The current Docker Azure integration does not allow fetching a combined log stream from all the containers that make up the Compose application.
 
+## Updating applications
+
+From a deployed Compose application, you can update the application by re-deploying it with the same ACI group name: `docker compose up --project-name` with the Compose project name already used in previous deployments.
+
+Updating an application means the ACI node will be reused, and the application will keep the IP address already allocated to expose ports, if any. ACI also has some limitations on what can be updated in an existing application (you will not be able to change CPU/memory reservation for example), in these cases you need to deploy a new application from scratch.
+
+Updating is the default behaviour if you invoke several times `docker compose up` on the same compose file, since the Compose project name is derived from the compose file location by default. You need to explicitely execute `docker compose down` before running again `docker compose up` in order to totally reset a Compose application.
```
chris-crone, line 159:
```suggestion
Updating is the default behavior if you invoke `docker compose up` on the an already deployed Compose file, since the Compose project name is derived from the directory where the Compose file is located by default. You need to explicitly execute `docker compose down` before running `docker compose up` again in order to totally reset a Compose application.

```

engine/context/aci-integration.md
```diff
@@ -150,12 +150,20 @@ Name resolution between containers is achieved by writing service names in the `
...
+
+Updating an application means the ACI node will be reused, and the application will keep the IP address already allocated to expose ports, if any. ACI also has some limitations on what can be updated in an existing application (you will not be able to change CPU/memory reservation for example), in these cases you need to deploy a new application from scratch.
+
+Updating is the default behaviour if you invoke several times `docker compose up` on the same compose file, since the Compose project name is derived from the compose file location by default. You need to explicitely execute `docker compose down` before running again `docker compose up` in order to totally reset a Compose application.
+
 ## Releasing resources
 
-Single containers and Compose applications can be removed from ACI with 
-the `docker prune` command. The `docker prune` command removes deployments 
+Single containers and Compose applications can be removed from ACI with
```
chris-crone, line 163: Can't change this as a suggestion unfortunately but would be good to rephrase so that we say it's only stopped/terminated containers that are cleaned up by prune.

engine/context/aci-integration.md
```diff
@@ -292,6 +300,34 @@ does not delete the storage account, even when it has zero remaining file shares
 When using `docker run`, you can pass the environment variables to ACI containers using the `--env` flag.
 For Compose applications, you can specify the environment variables in the Compose file with the `environment` or `env-file` service field, or with the `--environment` command line flag.
 
+## Healthchecks
+
+You can specify container healthchecks either with `docker run --healthcheck-xxx` flags, or in a compose file with the healthcheck section.
```
chris-crone, line 305:
```suggestion
You can specify a container health checks using either the `--healthcheck-` prefixed. flags with `docker run`, or in a Compose file with the healthcheck section of the service.

```

engine/context/aci-integration.md
```diff
@@ -292,6 +300,34 @@ does not delete the storage account, even when it has zero remaining file shares
 When using `docker run`, you can pass the environment variables to ACI containers using the `--env` flag.
 For Compose applications, you can specify the environment variables in the Compose file with the `environment` or `env-file` service field, or with the `--environment` command line flag.
 
+## Healthchecks
+
+You can specify container healthchecks either with `docker run --healthcheck-xxx` flags, or in a compose file with the healthcheck section.
+Healthchecks are converted to ACI LivenessProbe. ACI will run the healthcheck command periodically, and when it fails, the container will be terminated.
```
chris-crone, line 306:
```suggestion
Health checks are converted to ACI `LivenessProbe`s. ACI will run the health check command periodically, and if it fails, the container will be terminated.

```

engine/context/aci-integration.md
```diff
@@ -292,6 +300,34 @@ does not delete the storage account, even when it has zero remaining file shares
 When using `docker run`, you can pass the environment variables to ACI containers using the `--env` flag.
 For Compose applications, you can specify the environment variables in the Compose file with the `environment` or `env-file` service field, or with the `--environment` command line flag.
 
+## Healthchecks
+
+You can specify container healthchecks either with `docker run --healthcheck-xxx` flags, or in a compose file with the healthcheck section.
+Healthchecks are converted to ACI LivenessProbe. ACI will run the healthcheck command periodically, and when it fails, the container will be terminated.
+Healthchecks must be used in addition to restart policies, to ensure the container is then restarted ; using no restart policy will lead to containers being killed but not restarted by healthchecks.
```
chris-crone, line 307:
```suggestion
Health checks must be used in addition to restart policies, to ensure the container is then restarted on termination. The default restart policy for `docker run` is `no` which will not restart the container. The default restart policy for Compose is `any` which will always try restarting the service containers.

```

chris-crone (approved):
Some minor changes but then LGTM

engine/context/aci-integration.md
```diff
@@ -292,6 +300,34 @@ does not delete the storage account, even when it has zero remaining file shares
...
+Healthchecks are converted to ACI LivenessProbe. ACI will run the healthcheck command periodically, and when it fails, the container will be terminated.
+Healthchecks must be used in addition to restart policies, to ensure the container is then restarted ; using no restart policy will lead to containers being killed but not restarted by healthchecks.
+
+Example using `docker run`:
+
+```console
+docker --context acicontext run -p 80:80 --restart always --health-cmd "curl http://localhost:80" --health-interval 3s  nginx
+```
+
+Example using compose files:
```
usha-mandya, line 315:
```suggestion
Example using Compose files:

```

engine/context/aci-integration.md
gtardif: It's actually any state that is not "running". ACI containers sometimes ends up in weird states that we can't all list. 
`The `docker prune` command removes deployments that are not currently running. ` is exactly what happens.

chris-crone: OK then SGTM

usha-mandya (approved):

PR #11738, opened 2020-11-11
Add examples for ps command to filter services by status

futtetennista:
### Proposed changes

There's no examples of filters anywhere in the docs. The only way to know which ones are valid apart from reading the source code. This PR doesn't aim to show all possible filters, just the one that I happened to need.


usha-mandya:
@aiordache Could you PTAL?

usha-mandya:
@futtetennista Thanks for the PR. Docker Compose reference is now part of Docker CLI reference. See https://docs.docker.com/engine/reference/commandline/compose_ps/. Docker CLI reference docs are autogenerated from the upstream repository. To suggest changes to the CLI reference, please open a PR in the upstream [CLI](https://github.com/docker/cli) repository.

PR #11741, opened 2020-11-11
remove unnecessary commands

lpmi-13:
these lines aren't needed anymore, and add extra layers to the final image.

### Proposed changes

This commit removes unnecessary `RUN mkdir` commands in the examples for django and rails.

These are needed anymore, and add extra layers to the final image.


usha-mandya (approved):
LGTM. Thank you!

PR #11744, opened 2020-11-12
Reference compose spec in compose file formats

aiordache:
Added compose spec tot the list of compose formats and redirect to compose spec GH page (implemented in Docker Compose >=v1.27.x).


_includes/content/compose-matrix.md
```diff
@@ -25,3 +26,6 @@ do not necessarily increment with each release. For example, Compose file format
 3.0 was first introduced in [Compose release
 1.10.0](https://github.com/docker/compose/releases/tag/1.10.0), and versioned
 gradually in subsequent releases.
+
+The latest Compose file format is defined by the [Compose Specification](https://github.com/compose-spec/compose-spec/blob/master/spec.md)
```
usha-mandya, line 30:
```suggestion
The latest Compose file format is defined by the [Compose Specification](https://github.com/compose-spec/compose-spec/blob/master/spec.md){:target="_blank" rel="noopener" class="_"}.
```

usha-mandya (dismissed):
LGTM. Thank you Anca. Will merge once @thaJeztah has had a chance to review the updates.

compose/compose-file/compose-file-v3.md
```diff
@@ -0,0 +1,2725 @@
+---
+description: Compose file reference
+keywords: fig, composition, compose, docker
+redirect_from:
+- /compose/yml
+- /compose/compose-file-v3/
```
thaJeztah, line 6: This redirect doesn't look correct, because this is the same location as the file is in now

_includes/content/compose-matrix.md
```diff
@@ -2,6 +2,7 @@ This table shows which Compose file versions support specific Docker releases.
 
 | **Compose file format** | **Docker Engine release** |
 |  -------------------    |    ------------------     |
+|  Compose specification  |       19.03.0+            |
```
thaJeztah, line 5: Docker 20.10 will ship with with compose-schema 3.9 (https://github.com/docker/cli/tree/v20.10.0-beta1/cli/compose/schema/data). Not sure how we'll address that, and it the compose-specification has all options that were added in that, so we probably need another matrix somewhere to list what's supported for `docker stack` commands.

I'm wondering if we need to come with a different name for this ("unified schema" ?) and/or if there will be releases of the compose-specification, otherwise it will be impossible to know what's part of the schema and what not.

compose/compose-file/compose-file-v3.md
```diff
@@ -0,0 +1,2725 @@
+---
+description: Compose file reference
```
thaJeztah, line 2: This should probably contain `v3` as well

compose/compose-file/index.md
```diff
@@ -3,16 +3,15 @@ description: Compose file reference
 keywords: fig, composition, compose, docker
 redirect_from:
 - /compose/yml
-- /compose/compose-file-v3/
-title: Compose file version 3 reference
+title: Compose file
```
thaJeztah, line 6: Looks like, because the new file (`compose-file-v3`) was added in the same commit as this file was modified, we'll be losing git history (`git blame`) won't work for the new file; could you split this PR into two commits?

- first commit to rename `index.md` to `compose-file-v3.md`
- second commit to add `index.md` (with the content that's in this PR), and to update `compose-file-v3.md` if needed?

I think it's important to make sure that git blame still works, as sometimes there's specific history attached to changes.

compose/compose-file/index.md
```diff
@@ -3,16 +3,15 @@ description: Compose file reference
...
 toc_max: 4
 toc_min: 1
 ---
 
 ## Reference and guidelines
 
-These topics describe version 3 of the Compose file format. This is the newest
-version.
+These topics describe the Docker Compose implementation of the Compose format.
+Docker Compose **1.27.0+** implements the format defined by the [Compose Specification](https://github.com/compose-spec/compose-spec/blob/master/spec.md). Previous Docker Compose versions have support for several Compose file formats – 1, 2, 2.x, and 3.x. The Compose specification is an unified 2.x and 3.x file format, aggregating properties accross these formats.
```
thaJeztah, line 14: Can we make sure that the version we're linking to matches the version that's used in docker-compose? (ideally, compose-spec would do releases; also see my other comment).

Also wondering if we could pull the actual spec/docs into this repository, so that we can have actual documentation on the file format (instead of users having to read on GitHub)

thaJeztah (dismissed):

usha-mandya (approved):

ldez:
This PR creates a regression: 

the item "Version 3" point to the https://docs.docker.com/compose/compose-file/ instead of https://docs.docker.com/compose/compose-file/compose-file-v3/

Screenshot_2020-12-18 Compose file

PR #11751, opened 2020-11-13
Aci ecs update

gtardif:
### Proposed changes

* Fix ACI flag name
* reorder ECS paragraphs, added command line examples for `compose logs`

### DIFF
--- a/engine/context/ecs-integration.md
+++ b/engine/context/ecs-integration.md
@@ -33,6 +33,8 @@ developers to seamlessly deploy Docker containers on Amazon ECS using a
 Compose file with the `docker compose up` command. The following sections
 contain instructions on how to deploy your Compose application on Amazon ECS.
 
+## Run an application on ECS
+
 ### Create AWS context
 
 Run the `docker context create ecs myecscontext` command to create an Amazon ECS Docker
@@ -49,7 +51,7 @@ myecscontext *
 default  Current DOCKER_HOST based configuration   unix:///var/run/docker.sock     swarm
 ```
 
-## Run Compose applications
+### Run a Compose application
 
 You can deploy and manage multi-container applications defined in Compose files
 to Amazon ECS using the `docker compose` command. To do this:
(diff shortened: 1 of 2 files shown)


engine/context/ecs-integration.md
```diff
@@ -86,7 +138,7 @@ arn:aws:secretsmanager:eu-west-3:12345:secret:DockerHubAccessToken
 
 Once created, you can use this ARN in you Compose file using using `x-aws-pull_credentials` custom extension with the Docker image URI for your service.
 
-```console
+```yaml
 version: 3.8
```
ndeloof, line 142: unrelated: we should remove version here to promote version-less compose-spec

ndeloof (approved):

engine/context/ecs-integration.md
gtardif: Yep, thx, updated

usha-mandya (dismissed):
LGTM. Thanks

engine/context/ecs-integration.md
```diff
@@ -119,7 +176,7 @@ If required, the initial file system can be customized using `driver-opts`:
 
 ```yaml
 volumes:
-  my-data: 
+  my-data:
```
nebuk89, line 179: Is this the simplest Compose file to create a new volume or are some of these optional?

engine/context/ecs-integration.md
```diff
@@ -49,7 +51,7 @@ myecscontext *
 default  Current DOCKER_HOST based configuration   unix:///var/run/docker.sock     swarm
```
nebuk89, line 51: Does the region not populate the description here?

engine/context/ecs-integration.md
```diff
@@ -73,6 +75,56 @@ their state using the `docker compose ps` command.
 - You can view logs from containers that are part of the Compose application
 using the `docker compose logs` command.
 
+## Rolling update
+
+Your ECS services are created with rolling update configuration. As you run
```
nebuk89, line 80: Can we start with a simple statement here, something like:
to update your application without interrupting production flow you can simple use Compose up.

nebuk89 (changes requested):
A few questions opened :)

engine/context/ecs-integration.md
gtardif: it does not actually for ECS. Also, since the context references AWS profile or env variables, region might change if you chenge the AWS profile definition or change env variables, so we can't easily use region in the description that is stored (duplicated) in the context

engine/context/ecs-integration.md
gtardif: updated

engine/context/ecs-integration.md
ndeloof: all optionals. Can write:
```yaml
volumes:
  my-data: {}
```
Would then be created with EFS defaults

gtardif: I added a sample compose service with the most basic volume definition possible, useful indeed

engine/context/ecs-integration.md
gtardif: creating a context from env vars, the description is actually set to "credentials read from environment", should we set this as example ?

nebuk89: Yeah I think so :)

nebuk89 (approved):
LGTM

PR #11756, opened 2020-11-16
$ sign removed

KritikaRanjan:


usha-mandya:
@KritikaRanjan Thank you for the PR. We are in the process of reviewing code examples across our docs and exploring the options of using a plugin to make it easier to copy/paste code blocks. Unfortunately, it is not possible to accept your PR at this stage as it will be inconsistent with the existing examples in our docs. I'll have to close this PR without merging for that reason. Feel free to continue the conversation after I close.

PR #11783, opened 2020-11-20
Add FS data volume for wordpress example

jlherren:
Without mounting a volume for the wordpress filesystem data, the user would lose any installed custom themes or plugins between down/up.

### DIFF
--- a/compose/wordpress.md
+++ b/compose/wordpress.md
@@ -52,6 +52,8 @@ Compose to set up and run WordPress. Before starting, make sure you have
          depends_on:
            - db
          image: wordpress:latest
+         volumes:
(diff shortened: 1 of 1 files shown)


compose/wordpress.md
```diff
@@ -52,6 +52,8 @@ Compose to set up and run WordPress. Before starting, make sure you have
          depends_on:
            - db
          image: wordpress:latest
+         volumes:
+           - wordpress_data: /var/www/html
```
usha-mandya, line 56:
```suggestion
           - wordpress_data:/var/www/html
```

jlherren: The rest of the file does not have spaces after the colon though...

PR #11789, opened 2020-11-23
Bump the notes on Docker support

usha-mandya:
- Moved the Support notes to the top of the page
- Updated Hub toc

### DIFF
--- a/_data/toc.yaml
+++ b/_data/toc.yaml
@@ -1240,6 +1240,8 @@ manuals:
     title: Webhooks
   - path: /docker-hub/slack_integration/
     title: Slack integration
+  - path: /docker-hub/download-rate-limit/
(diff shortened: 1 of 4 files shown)


nebuk89 (changes requested):
Awesome! For the URLs in the TOC we should have UTMs so we can track referrals.. or do we not need that with things going from Docs -> Hub for conversion?

usha-mandya:
> Awesome! For the URLs in the TOC we should have UTMs so we can track referrals.. or do we not need that with things going from Docs -> Hub for conversion?

@amnrahman Could you please let me know what these UTMs are so I can start adding these to track referrals?

nebuk89 (approved):
As discussed push this for now :)

PR #11818, opened 2020-12-01
Fix invalid compress value in the json-file docs

### LINKED ISSUE #9565: Fix compress option type in example
[config/containers/logging/json-file.md](https://docs.docker.com/config/containers/logging/json-file/)
Example shows bool type, but in fact it should be string with "enabled" or "disabled".
Docker daemon version:19.03.2, build 6a30dfc

luzfcb:
### Proposed changes

The current documentation suggests that `true` is a valid value for `compress` option of the `json-file`, when in fact the expected value is `enabled` and not `true`

### Related issues (optional)

Closes #9565


luzfcb:
Okay, per https://golang.org/pkg/strconv/ , `strconv.ParseBool` expect "true" or "false"

PR #11865, opened 2020-12-09
Compose cli reference docs

gtardif:
### Proposed changes

* Add reference docs previously hosted in https://github.com/docker/compose-cli/tree/main/docs
* Moved cloud integration docs under /cloud, instead of /engine/context. Happy to change this to another folder if someone has a better idea
* Added a redirect from the original md aci/ecs files, since this was referenced all over the place

(move is in a separate commit to make it easier to see content diff)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -20,6 +20,9 @@ ARG ENGINE_BRANCH="master"
 # Distribution
 ARG DISTRIBUTION_BRANCH="release/2.7"
 
+# Compose CLI
+ARG COMPOSE_CLI_BRANCH="main"
+
 ###
 # Set up base stages for building and deploying
 ###
@@ -35,6 +38,9 @@ ENV ENGINE_BRANCH=${ENGINE_BRANCH}
 ARG DISTRIBUTION_BRANCH
 ENV DISTRIBUTION_BRANCH=${DISTRIBUTION_BRANCH}
 
+ARG COMPOSE_CLI_BRANCH
+ENV COMPOSE_CLI_BRANCH=${COMPOSE_CLI_BRANCH}
+
 # Fetch upstream resources (reference documentation)
 # Only add the files that are needed to build these reference docs, so that these
 # docs are only rebuilt if changes were made to ENGINE_BRANCH or DISTRIBUTION_BRANCH.
@@ -45,6 +51,7 @@ WORKDIR /usr/src/app/md_source/
 COPY ./_scripts/fetch-upstream-resources.sh ./_scripts/
 ARG ENGINE_BRANCH
 ARG DISTRIBUTION_BRANCH
+ARG COMPOSE_CLI_BRANCH
 RUN ./_scripts/fetch-upstream-resources.sh .
 
 
--- a/_data/toc.yaml
+++ b/_data/toc.yaml
@@ -81,10 +81,20 @@ guides:
     title: Configure GitHub Actions
 - sectiontitle: Deploy your app to the cloud
   section:
-  - path: /engine/context/aci-integration/
+  - path: /cloud/aci-integration/
     title: Docker and ACI
-  - path: /engine/context/ecs-integration/
+  - path: /cloud/aci-container-features/
+    title: ACI container features
+  - path: /cloud/aci-compose-features/
+    title: ACI Compose features
+  - path: /cloud/ecs-integration/
     title: Docker and ECS
+  - path: /cloud/ecs-architecture/
+    title: Docker ECS integration architecture
+  - path: /cloud/ecs-compose-features/
+    title: ECS Compose features
+  - path: /cloud/ecs-compose-examples/
+    title: ECS Compose examples
 - sectiontitle: Run your app in production
   section:
   - sectiontitle: Orchestration
--- a/_scripts/fetch-upstream-resources.sh
+++ b/_scripts/fetch-upstream-resources.sh
@@ -6,6 +6,7 @@
 # which are usually set by the Dockerfile.
 : "${ENGINE_BRANCH?No release branch set for docker/docker and docker/cli}"
 : "${DISTRIBUTION_BRANCH?No release branch set for docker/distribution}"
+: "${COMPOSE_CLI_BRANCH?No release branch set for docker/compose-cli}"
 
 # Translate branches for use by svn
 engine_svn_branch="branches/${ENGINE_BRANCH}"
@@ -16,10 +17,15 @@ distribution_svn_branch="branches/${DISTRIBUTION_BRANCH}"
 if [ "${distribution_svn_branch}" = "branches/master" ]; then
 	distribution_svn_branch=trunk
 fi
+compose_cli_svn_branch="branches/${COMPOSE_CLI_BRANCH}"
+if [ "${compose_cli_svn_branch}" = "branches/main" ]; then
+	compose_cli_svn_branch=trunk
+fi
 
 # Directories to get via SVN. We use this because you can't use git to clone just a portion of a repository
 svn co "https://github.com/docker/cli/${engine_svn_branch}/docs/extend"              ./engine/extend || (echo "Failed engine/extend download" && exit 1)
 svn co "https://github.com/docker/docker/${engine_svn_branch}/docs/api"              ./engine/api    || (echo "Failed engine/api download" && exit 1)
+svn co "https://github.com/docker/compose-cli/${compose_cli_svn_branch}/docs"              ./cloud    || (echo "Failed compose-cli/docs download" && exit 1)
 svn co "https://github.com/docker/distribution/${distribution_svn_branch}/docs/spec" ./registry/spec || (echo "Failed registry/spec download" && exit 1)
 svn co "https://github.com/mirantis/compliance/trunk/docs/compliance"                ./compliance    || (echo "Failed docker/compliance download" && exit 1)
 
@@ -37,3 +43,6 @@ wget --quiet --directory-prefix=./registry/                     "https://raw.git
 # Remove things we don't want in the build
 rm -f ./engine/extend/cli_plugins.md # the cli plugins api is not a stable API, and not included in the TOC for that reason.
 rm -f ./registry/spec/api.md.tmpl
+rm -f ./cloud/README.md # readme to make things nice in the compose-cli repo, but meaningless here
+rm -f ./cloud/architecure.md # Compose-CLI architecture, unrelated to cloud integration
+rm -rf ./cloud/images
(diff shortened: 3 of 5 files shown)


gtardif:
cc @nebuk89 @thaJeztah moving reference cloud integration docs here, let me know if you see a better place to replace /engin/context

gtardif:
hmmm, need to fix headers in new reference md files I think, they appear as raw text

gtardif:
just rebased

engine/context/aci-integration.md
```diff
@@ -1,465 +1,5 @@
 ---
 title: Deploying Docker containers on Azure
 description: Deploying Docker containers on Azure
-keywords: Docker, Azure, Integration, ACI, context, Compose, cli, deploy, containers, cloud
-toc_min: 1
-toc_max: 2
+redirect_to: /cloud/aci-integration/
```
thaJeztah, line 4: Perhaps better to add a `redirect_from` in the new location, so that we don't have to keep this file (same for the other files)

thaJeztah: Also; could this be done with a _move_ of the file, so that we keep git history? (i.e. same commit that moves the file and adds the redirect)?

chris-crone (dismissed):
LGTM

_data/toc.yaml
```diff
@@ -67,9 +67,9 @@ guides:
     title: Configure GitHub Actions
 - sectiontitle: Deploy your app to the cloud
   section:
-  - path: /engine/context/aci-integration/
+  - path: /cloud/aci-integration/
     title: Docker and ACI
-  - path: /engine/context/ecs-integration/
+  - path: /cloud/ecs-integration/
```
thaJeztah, line 72: We're putting all these under `/cloud/`; is other content gonna be put there? Do we have a roadmap how we're gonna be unifying the docs related to this? (hard to describe; let me try); currently this is all about "compose integration with cloud". A logical location for it would be to be located under `/compose/xxx`, however, that's for the "python" compose, so might be confusing (unless we call out that it's for the "cloud" compose)

At some point, the go implementation will be equivalent in functionality, and we should probably be able to unify them; we should think of a plan how we move from "current situation" to "destination" (without moving files around too many times)

thaJeztah (changes requested):
If this content is closely tied to the code in the compose-cli repository, could we (instead  of moving the files here) fetch them when building the docs? Similar to how we do this for some other projects; https://github.com/docker/docker.github.io/blob/master/_scripts/fetch-upstream-resources.sh

That way, changes in the (reference) docs can be made together with the code changes, and will be updated automatically if we update the version from which to fetch the docs (and docs won't be out of sync with the actual code)

cloud/aci-compose-features.md
```diff
@@ -0,0 +1,241 @@
+---
```
thaJeztah, line 1: Looks like this page is not added to the TOC currently

cloud/aci-container-features.md
```diff
@@ -0,0 +1,85 @@
+---
```
thaJeztah, line 1: Same for this page

cloud/aci-compose-features.md
```diff
@@ -0,0 +1,241 @@
...
+title: ACI integration compose features
```
usha-mandya, line 2: Compose features?

cloud/ecs-integration.md
```diff
@@ -0,0 +1,536 @@
...
+Compose file with the `docker compose up` command. The following sections
+contain instructions on how to deploy your Compose application on Amazon ECS.
+
+## Run an application on ECS
+
+### Requirements
+
+AWS uses a fine-grained permission model, with specific role for each resource type and operation.
+
+To ensure that Docker ECS integration is allowed to manage resources for your Compose application, you have to ensure your AWS credentials grant access to following AWS IAM permissions:
```
nebuk89, line 44: Is there any guide we can point them to on how to do this, for a new user this is pretty scary as a list

nebuk89 (approved):
overall good and I am happy for us to ship now. I think the main thing to change would be making sure we have links to AWS/Azure guides on how to do the setup pieces or clear instructions around some of their stack pieces. I know it isn't up to maintain this but if we want 'new' users to try this out we need to avoid any gaps where we assume people know the cloud platform they are deploying to

engine/context/aci-integration.md
gtardif: Indeed, much better! fixed and re-organised commits

cloud/aci-compose-features.md
gtardif: fixed (and same in ecs)

cloud/aci-compose-features.md
```diff
@@ -0,0 +1,241 @@
...
+description: Reference list of compose ACI features
+keywords: Docker, Azure, Integration, ACI, Compose, cli, deploy, cloud
+toc_min: 1
+toc_max: 2
+---
+
+# Compose - Azure Container Instances mapping
+
+This document outlines the conversion of an application defined in a Compose file to ACI objects.
+At a high-level, each Compose deployment is mapped to a single ACI container group.
```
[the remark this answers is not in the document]
gtardif: fixed, and also add a link to azure resource groups in context creation help

cloud/ecs-integration.md
gtardif: fixed

cloud/aci-compose-features.md
gtardif: I only linked these from the aci-integration and ecs-integration pages, I think it's all right, but I can change that if yo prefer.

gtardif:
Updated with your comments.
@nebuk89 @thaJeztah OK with the current state ?

cloud/aci-compose-features.md
gtardif: After Ben comment, added pages to TOC

cloud/aci-container-features.md
gtardif: After Ben comment, added pages to TOC

gtardif:
Added TOC references after your last comment (slack msg) @nebuk89.

_scripts/fetch-upstream-resources.sh
```diff
@@ -16,10 +17,15 @@ distribution_svn_branch="branches/${DISTRIBUTION_BRANCH}"
...
 fi
+compose-cli_svn_branch="branches/main"
+if [ "${compose-cli_svn_branch}" = "branches/main" ]; then
+	compose-cli_svn_branch=trunk
+fi
 
 # Directories to get via SVN. We use this because you can't use git to clone just a portion of a repository
 svn co "https://github.com/docker/cli/${engine_svn_branch}/docs/extend"              ./engine/extend || (echo "Failed engine/extend download" && exit 1)
 svn co "https://github.com/docker/docker/${engine_svn_branch}/docs/api"              ./engine/api    || (echo "Failed engine/api download" && exit 1)
+svn co "https://github.com/docker/compose-cli/${engine_svn_branch}/docs/api"              ./cloud    || (echo "Failed compose-cli/docs download" && exit 1)
```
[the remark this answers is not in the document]
thaJeztah: Hmm.. looks like it's failing; ~could be that SVN support on GitHub not yet supports the "master -> main" mapping~ see comment below



```
#17 14.35 svn: E170000: URL 'https://github.com/docker/compose-cli/trunk/docs/api' doesn't exist
#17 14.35 Failed compose-cli/docs download
```

_scripts/fetch-upstream-resources.sh
```diff
@@ -16,10 +17,15 @@ distribution_svn_branch="branches/${DISTRIBUTION_BRANCH}"
 if [ "${distribution_svn_branch}" = "branches/master" ]; then
 	distribution_svn_branch=trunk
 fi
+compose-cli_svn_branch="branches/main"
+if [ "${compose-cli_svn_branch}" = "branches/main" ]; then
+	compose-cli_svn_branch=trunk
+fi
```
thaJeztah, line 23: Perhaps use an underscore instead of a hyphen in the variable name;

```suggestion
compose_cli_svn_branch="branches/${COMPOSE_CLI_BRANCH}"
if [ "${compose_cli_svn_branch}" = "branches/main" ]; then
	compose_cli_svn_branch=trunk
fi
```

_scripts/fetch-upstream-resources.sh
```diff
@@ -16,10 +17,15 @@ distribution_svn_branch="branches/${DISTRIBUTION_BRANCH}"
 if [ "${distribution_svn_branch}" = "branches/master" ]; then
 	distribution_svn_branch=trunk
 fi
+compose-cli_svn_branch="branches/main"
```
thaJeztah, line 20: This should contain the `COMPOSE_CLI_BRANCH` env-var;

```suggestion
compose-cli_svn_branch="branches/${COMPOSE_CLI_BRANCH}"
```

_scripts/fetch-upstream-resources.sh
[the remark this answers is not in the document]
thaJeztah: Ah, it's because it's trying `/docs/api` instead of `/docs` (fixing my suggestion above)

Dockerfile
```diff
@@ -20,6 +20,9 @@ ARG ENGINE_BRANCH="master"
 # Distribution
 ARG DISTRIBUTION_BRANCH="release/2.7"
 
+# Compose CLI
+ARG COMPOSE_CLI_BRANCH="main"
```
thaJeztah, line 24: We'll probably have to replace this with a specific tag/reference at some point, to prevent docs going out before the binary itself is updated, but should be ok for now

cloud/ecs-architecture.md
```diff
@@ -1,3 +1,10 @@
+---
+title: ECS integration architecture
+description: Mapping of Docker compose entities to Amazon constructs
+keywords: Docker, Amazon, Integration, ECS, Compose, architecture, mapping
+toc_min: 1
+toc_max: 2
+---
```
thaJeztah: We need to add this front-matter to the upstream repository now

Probably the `toc_min` and `toc_max` can be skipped

thaJeztah (changes requested):
left some comments; looks like Jekyll currently doesn't build the pages that were fetched from upstream because they don't have front-matter metadata in them yet (I suspect it's because the files in _this_ repository do have front-matter, and Jekyll doesn't like mixing with files that do and don't have them 🤔

_scripts/fetch-upstream-resources.sh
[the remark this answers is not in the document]
gtardif: Thx, fixed

_scripts/fetch-upstream-resources.sh
gtardif: updated

_scripts/fetch-upstream-resources.sh
gtardif: fixed

cloud/ecs-architecture.md
gtardif: Added for all files, now merged in compose-cli main branch

thaJeztah (approved):
LGTM

tested this locally, and looks good 👍 thanks!

PR #11874, opened 2020-12-10
Fix trust_delegation.md for target role

developer-guy:
I think the target role should be target/releases instead of the targets/targets.


usha-mandya:
@tiborvass Could you PTAL

developer-guy:
@tiborvass hello Tibor, is any update available about this issue?

tiborvass (approved):

PR #11895, opened 2020-12-12
Update release-note

fahedouch:
based on this comment https://github.com/moby/buildkit/issues/1892#issuecomment-743409990. the source tracking feature is only possible with `buildx` v0.5-rc which is not available in docker 20.10


usha-mandya:
@thaJeztah Could you PTAL

thaJeztah:
I think this should still be available when using `DOCKER_BUILDKIT=1`, which uses BuildKit compiled into the docker daemon

thaJeztah:
Oh, never mind; catching up on https://github.com/moby/buildkit/issues/1892#issuecomment-743409990

thaJeztah (approved):
LGTM

PR #11914, opened 2020-12-16
Update userns-remap.md

HareshChhelana:
Is it 231072 UID instead of 231073 right?


usha-mandya:
@HareshChhelana `231073` is correct in this instance as it is mapped to UID `1` and not UID `0`. I'll have to close this PR without merging for this reason.

usha-mandya:
Feel free to continue the conversation after I close

PR #11946, opened 2020-12-20
Add missing opening tag in Get-started > Multi-Container

GuillaumeOj:
### Proposed changes

Add a missing opening tag for a code snippet : https://docs.docker.com/get-started/07_multi_container/#running-our-app-with-mysql

PR #11971, opened 2020-12-23
Fix missing opening for code block

sroet:
While going through the tutorial I noticed one of the code blocks was missing formatting, this fixes that

### Proposed changes
Fix the code block that was missing the opening back-ticks ` ```bash ` this fixes it.


sroet:
closing as it is a duplicate of #11946

PR #11972, opened 2020-12-23
Fix incorrect link to compose-file-v3

mike-marcacci:
The navigation item for "Version 3" of the docker compose file referenced a nearly useless page, while the exceedingly thorough and useful v3 docs were only discoverable via Google.

### Proposed changes

Fixed link


thaJeztah:
Looks like this was fixed in https://github.com/docker/docker.github.io/pull/11959, so let me close this one; thanks for contributing!

Issue #11625, opened 2020-10-25
Use trusted.gpg.d instead of apt-key


he7d3r:
### Problem description

One of the steps at https://docs.docker.com/engine/install/ubuntu/#install-using-the-repository is to add Docker’s official GPG key like this:
```
$ curl -fsSL https://download.docker.com/linux/ubuntu/gpg | sudo apt-key add -
```
However, the last part generates the following warning (at least on Ubuntu 20.10):
> Warning: apt-key is deprecated. Manage keyring files in trusted.gpg.d instead (see apt-key(8)).

Maybe it is time to use `trusted.gpg.d` as suggested by the warning?

FelipeGangrel:
Same here

noahsbwilliams:
I'm experiencing the same issue, with the additional problem that when I try to manually add the gpg file to `/etc/apt/trusted.gpg.d/docker.gpg` and then run `sudo apt update` I get this error:

```stdout
The key(s) in the keyring /etc/apt/trusted.gpg.d/docker.gpg are ignored as the file has an unsupported filetype.
```

...with of course loads of other errors from not having a valid key.

devs-saifur-rahman:
This worked for me 

`curl -fsSL https://download.docker.com/linux/ubuntu/gpg | sudo apt-key --keyring /etc/apt/trusted.gpg.d/docker-apt-key.gpg add`

Also, if you are using groovy , you should try focal instead.

mcaci:
Hi @devs-saifur-rahman, thanks for this tip. I was stuck in the same point and it helped me to progress with the install. This may be a good update to be done in the docs.

denis-roy:
apt-key is deprecated and will not be available after Debian 11 / Ubuntu 22.04

Although adding keys directly to `/etc/apt/trusted.gpg.d`/ is suggested by apt-key deprecation message, as per [Debian Wiki](https://wiki.debian.org/DebianRepository/UseThirdParty) GPG keys for third party repositories should be added to `/usr/share/keyrings` and referenced with the `signed-by` option in the source.list.d entry.

#### ADD A KEY

```bash
# Adding an ASCII Armored key (.asc key)
curl -fsSL https://download.docker.com/linux/ubuntu/gpg | \
gpg --dearmor | \
sudo tee /usr/share/keyrings/docker-ce-archive-keyring.gpg > \
/dev/null

# Or if you prefer a one-liner
[...]

PR #11974, opened 2020-12-24
Update debian.md

wschwab:
### Proposed changes

Added workaround for `apt-key` deprecation errors.

The installation instructions for Debian use a plain `sudo apt-key add -` to 
add the gpg fingerprint, resulting in errors when trying to install on newer 
Debian builds, as detailed in #11625 . The workaround is copied from [this](https://github.com/docker/docker.github.io/issues/11625#issuecomment-735273469) comment.

### Related issues (optional)

#11625

### DIFF
--- a/engine/install/debian.md
+++ b/engine/install/debian.md
@@ -93,7 +93,16 @@ from the repository.
     ```bash
     $ curl -fsSL {{ download-url-base }}/gpg | sudo apt-key add -
     ```
-
+    Newer versions of Debian have deprecated using `apt-key` like this. If you 
+    are receiving the following error:
+    ```
+    Warning: apt-key is deprecated. Manage keyring files in trusted.gpg.d instead (see apt-key(8)).
+    ```
+    then you can avoid it by using the following instead:
+    ```
+    curl -fsSL https://download.docker.com/linux/debian/gpg | sudo apt-key --keyring /etc/apt/trusted.gpg.d/docker-apt-key.gpg add -
+    ```
+    
     Verify that you now have the key with the fingerprint
     `9DC8 5822 9FC7 DD38 854A  E2D8 8D81 803C 0EBF CD88`, by searching for the
     last 8 characters of the fingerprint.


denis-roy:
I noticed the workaround still make use of the deprecated apt-key tool.

I believe a more future-proof approach would be to discontinue use of apt-key altogether as described in [Debian's wiki](https://wiki.debian.org/DebianRepository/UseThirdParty) and summarized in [this comment](https://github.com/docker/docker.github.io/issues/11625#issuecomment-751388087)

engine/install/debian.md
```diff
@@ -93,7 +93,16 @@ from the repository.
...
     ```
-
+    Newer versions of Debian have deprecated using `apt-key` like this. If you 
+    are receiving the following error:
+    ```
+    Warning: apt-key is deprecated. Manage keyring files in trusted.gpg.d instead (see apt-key(8)).
+    ```
+    then you can avoid it by using the following instead:
+    ```
+    curl -fsSL https://download.docker.com/linux/debian/gpg | sudo apt-key --keyring /etc/apt/trusted.gpg.d/docker-apt-key.gpg add -
```
thaJeztah, line 103: Does this also work on older versions? If so, we should probably replace the existing instructions to use the new one

This should probably also use the variable;

```suggestion
    curl -fsSL {{ download-url-base }}/gpg | sudo apt-key --keyring /etc/apt/trusted.gpg.d/docker-apt-key.gpg add -
```

thaJeztah (changes requested):

engine/install/debian.md
denis-roy: Good day @thaJeztah 

As per my [previous comment](https://github.com/docker/docker.github.io/pull/11974#issuecomment-751405853), this solution still rely on the deprecated `apt-key` tool. 

#11990 offers a Debian-compliant way of dealing with `apt-key` deprecation that applies to Debian 9 / Ubuntu 16.04 onward.

Thank you :)

ankostis: I believe the verification command needs also ` --keyring /etc/apt/trusted.gpg.d/docker-apt-key.gpg`, below.

PR #11975, opened 2020-12-24
Fix markdown syntax in 07_multi_container.md

maww:
### Proposed changes

Fix markdown syntax


thaJeztah:
Oh! I just merged a similar fix in https://github.com/docker/docker.github.io/pull/12040, so this one will no longer be needed. Thanks for contributing!

PR #11988, opened 2020-12-26
Fixed duplicate image in the minecraft tutorial

viktor02:
The image of the minecraft start window from the first point is duplicated in the second point after "Click the “Add Server” button"

### Proposed changes

Replaced the duplicate image with the desired one in the minecraft tutorial.


thaJeztah (approved):
LGTM, thanks!

PR #11996, opened 2020-12-28
Update ubuntu.md

benkoshy:
### Why this PR?

I got confused. Is setting up the docker engine part of the instructions for installing via a repository? It's not easy to distinguish between the level 3 and level 4 markup headings. I thought I would number them explicitly to avoid confusion. A small change, but hopefully it is helpful. if not, pls feel free to close. chrs.

Ben


usha-mandya:
@benkoshy Thanks for the feedback. We are unable to accept your suggestions as they do not follow our standard heading format. However, we are looking to address this issue by updating some of the existing headings. Thanks again for your feedback. We really appreciate it. :)

PR #11838, opened 2020-12-02
Adding general docker getting started to docs

nebuk89:
### Proposed changes

Adding in Docker tutorial that matches other getting started experience so we have a single tutorial we can iterate on.

PR #12003, opened 2020-12-28
Update kube-deploy.md Fixing dead links

### LINKED ISSUE #12001: 404 at /get-started/kube-deploy/part2.md
I found a broken link : https://docs.docker.com/get-started/kube-deploy/part2.md
I arrived on this page through: https://docs.docker.com/get-started/kube-deploy/

xberna16:
Fixing dead links

### Proposed changes

Fix two dead links on kube-deploy page

### Related issues 

Fixes #12001

### DIFF
--- a/get-started/kube-deploy.md
+++ b/get-started/kube-deploy.md
@@ -7,7 +7,7 @@ description: Learn how to describe and deploy a simple application on Kubernetes
 ## Prerequisites
 
 - Download and install Docker Desktop as described in [Orientation and setup](index.md).
-- Work through containerizing an application in [Part 2](part2.md).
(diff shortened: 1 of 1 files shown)


thaJeztah:
Hm.. looks like this was introduced when the new getting started was added in https://github.com/docker/docker.github.io/pull/11838, which removed some of the old content.

@usha-mandya @nebuk89 we need to look at these, because the new getting started did not add redirects for the old URLs; not sure though if we can redirect, because the new content does not match the old one, so linking to the new pages probably doesn't provide the intended content; perhaps we need to restore some of the old content? (could be at a new URL)

thaJeztah:
Looks like https://github.com/docker/docker.github.io/pull/11960 also addresses this

PR #12007, opened 2020-12-28
Fix heading level in compose/completion.md

notpushkin:
### Proposed changes

Right now Mac section is at the same level as the surrounding Bash and zsh headings, while logically it should be nested inside the Bash one. This PR fixes that.


thaJeztah (approved):
LGTM, thanks!

PR #12055, opened 2021-01-06
Added release notes for Docker Desktop 3.0.4

mat007:
### Proposed changes

Added release notes for Docker Desktop 3.0.4

### DIFF
--- a/docker-for-mac/release-notes.md
+++ b/docker-for-mac/release-notes.md
@@ -16,10 +16,28 @@ This page contains information about the new features, improvements, known issue
 > Starting with Docker Desktop 3.0.0, Stable and Edge releases are combined into a single release stream for all users. Updates to Docker Desktop will now be available automatically as delta updates from the previous version. This means, when there is a newer version of Docker Desktop, it will be automatically downloaded to your machine. All you need to do is to click **Update and restart** from the Docker menu to install the latest update.
 {: .important }
 
+## Docker Desktop Community 3.0.4
+2021-01-06
+
+> [Download](https://desktop.docker.com/mac/stable/Docker.dmg)
+
+### Upgrades
+
+- [Docker Engine 20.10.2](https://docs.docker.com/engine/release-notes/#20102)
+
+### Bug fixes and minor changes
+
+- Avoid timeouts during `docker-compose up` by making cache invalidation faster. Fixes [docker/for-mac#4957](https://github.com/docker/for-mac/issues/4957).
+- Avoid generating a spurious filesystem DELETE event while invalidating caches. Fixes [docker/for-mac#5124](https://github.com/docker/for-mac/issues/5124).
+
+### Known issues
+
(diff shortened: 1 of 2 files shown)


djs55 (dismissed):
LGTM

docker-for-mac/release-notes.md
```diff
@@ -16,10 +16,23 @@ This page contains information about the new features, improvements, known issue
...
+
+> [Download](https://desktop.docker.com/mac/stable/Docker.dmg)
+
+### Upgrades
+
+- [Docker Engine 20.10.2](https://docs.docker.com/engine/release-notes/#20102)
+
+### Bug fixes and minor changes
+
+- Fixed a failure with mounting overlapping volumes. See [docker/for-mac#5157](https://github.com/docker/for-mac/issues/5157).
```
stephen-turner, line 30: This is not correct. 5157 was fixed in 3.0.3: it was just done by breaking 4957 and 5124. So the fix in 3.0.4 relative to 3.0.3 is that the latter two issues are now fixed (and 5157 isn't broken again, but we don't need to write that).

See the 3.0.3 release notes for more context.

docker-for-windows/release-notes.md
```diff
@@ -16,6 +16,19 @@ This page contains information about the new features, improvements, known issue
...
+
+> [Download](https://desktop.docker.com/win/stable/Docker%20Desktop%20Installer.exe)
+
+### Upgrades
+
+- [Docker Engine 20.10.2](https://docs.docker.com/engine/release-notes/#20102)
+
+### Bug fixes and minor changes
+
+- Fixed an issue causing Docker Desktop to fail to start. See [docker/for-win#955](https://github.com/docker/for-win/issues/9755).
```
stephen-turner, line 30: "Fixed an issue that could cause Docker Desktop to fail to start after upgrading to 3.0.0".

stephen-turner: Also typo: 9755 not 955.

docker-for-mac/release-notes.md
```diff
@@ -16,10 +16,23 @@ This page contains information about the new features, improvements, known issue
...
+
```
stephen-turner, line 31: I think the Alpine known issue is not yet fixed on this branch so we should carry it forward in the release notes.

stephen-turner (changes requested):

docker-for-mac/release-notes.md
mat007: Right, so what do we say here? That we fixed 4957 and 5124 again?

mat007: OK, that actually makes sense, I’m adding back
```
- Avoid timeouts during `docker-compose up` by making cache invalidation faster. Fixes [docker/for-mac#4957](https://github.com/docker/for-mac/issues/4957).
- Avoid generating a spurious filesystem DELETE event while invalidating caches. Fixes [docker/for-mac#5124](https://github.com/docker/for-mac/issues/5124).
```

stephen-turner (approved):
Looks good now.

PR #12059, opened 2021-01-07
Replace SIGTERM with SIGQUIT

mike-marcacci:
### Proposed changes

After much trial and error I discovered that these docs are incorrect, and that docker-compose is actually sending `SIGQUIT` instead of `SIGTERM` to initiate a graceful shutdown. This updates the docs to reflect the actual current behavior of docker-compose version 1.27.4, build 40524192.

### DIFF
--- a/compose/faq.md
+++ b/compose/faq.md
@@ -15,11 +15,11 @@ Yes - see [Controlling startup order](startup-order.md).
 
 ## Why do my services take 10 seconds to recreate or stop?
 
-Compose stop attempts to stop a container by sending a `SIGTERM`. It then waits
+Compose stop attempts to stop a container by sending a `SIGQUIT`. It then waits
 for a [default timeout of 10 seconds](reference/stop.md).  After the timeout,
 a `SIGKILL` is sent to the container to forcefully kill it.  If you
 are waiting for this timeout, it means that your containers aren't shutting down
-when they receive the `SIGTERM` signal.
+when they receive the `SIGQUIT` signal.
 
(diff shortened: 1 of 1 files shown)


thaJeztah:
Hmm.. do you have more details how you came to this? Do you have steps to reproduce how you determined the stop-signal that was used?

To my knowledge, docker-compose would do a `container stop` using the docker remote API, in which case this code is executed; https://github.com/moby/moby/blob/69b4fe406540c7989bd99b2244ed67ced96e17c7/daemon/stop.go#L45-L85

In that code, it requests the default stop-signal to use from the container (which can define a custom signal, either for a specific container, or defined in the container's image); https://github.com/moby/moby/blob/69b4fe406540c7989bd99b2244ed67ced96e17c7/daemon/stop.go#L45

```go
	stopSignal := container.StopSignal()
```

https://github.com/moby/moby/blob/69b4fe406540c7989bd99b2244ed67ced96e17c7/container/container.go#L497-L504

If the container does not define a custom stop-signal, it will use `signal.DefaultStopSignal`, which is `SIGTERM`; https://github.com/moby/moby/blob/69b4fe406540c7989bd99b2244ed67ced96e17c7/pkg/signal/signal_unix.go#L19-L20

```go
// DefaultStopSignal is the syscall signal used to stop a container in unix systems.
DefaultStopSignal = "SIGTERM"
```

thaJeztah:
Closing this for now, as I don't think this is correct, but feel free to comment if you have more information

Issue #9996, opened 2019-12-10
Old links at https://docs.docker.com/compose/environment-variables/


oj-lappi:
### Problem description

The url [https://docs.docker.com/compose/compose-file.md](https://docs.docker.com/compose/compose-file.md) no longer points to anything.
This url is linked to from [https://docs.docker.com/compose/environment-variables/](https://docs.docker.com/compose/environment-variables/)

The link is displayed as "Variable substitution"

I assume it should be pointing at [https://docs.docker.com/compose/compose-file/](https://docs.docker.com/compose/compose-file/) instead

### Problem location

- I saw a problem on the following URL: https://docs.docker.com/compose/environment-variables/

### Project version(s) affected

No version, online documentation

### Suggestions for a fix

Point the link at[https://docs.docker.com/compose/compose-file/#variable-substitution](https://docs.docker.com/compose/compose-file/#variable-substitution) instead.

Wallenstein61:
[compose/index.md](https://docs.docker.com/compose/)

Just came across the same problem: I cannot find any detailed information on "variable subsitution". The suggestion for a fix above 

> Point the link at https://docs.docker.com/compose/compose-file/#variable-substitution instead.

 seems not to work. There is no anchor `variable-substitution` on this page.

oj-lappi:
This issue is over a year old, I believe the documentation has changed structure since then.

thaJeztah:
opened https://github.com/docker/docker.github.io/pull/12060 to fix

PR #12060, opened 2021-01-07
(temporarily) replace compose-file deep links to v3 docs

thaJeztah:
fixes https://github.com/docker/docker.github.io/issues/9996

The documentation curently doesn't contain reference docs for the unified (version-less) compose-file syntax, so for now replacing all links to point to the v3 compose-file reference, which is still present.

### DIFF
--- a/compose/compose-file/compose-file-v1.md
+++ b/compose/compose-file/compose-file-v1.md
@@ -436,7 +436,7 @@ Links also express dependency between services in the same way as
 
 > **Note**
 >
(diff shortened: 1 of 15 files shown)


thaJeztah:
ping @StefanScherer @aiordache PTAL

StefanScherer (approved):
makes sense
LGTM
(netlify seems  to have problems, giving me 503's when I  tried to view some pages)

PR #12064, opened 2021-01-07
Fix formatting in templates of pages related with build command

### LINKED ISSUE #11331: Incorrect HTML output
[engine/reference/commandline/build.md](https://docs.docker.com/engine/reference/commandline/build/)

The HTML output is not correct in the explanation of the ssh option because of using greater/less than signs as they are.
I am not sure how should it be, but escape with backslashes seems to get the output properly.

The others are the same:  /engine/reference/commandline/{builder_build,image_build,buildx_build}

```yaml
diff --git a/_data/engine-cli/docker_build.yaml b/_data/engine-cli/docker_build.yaml
index 7d28b8f2b2..84f764de91 100644
--- a/_data/engine-cli/docker_build.yaml
+++ b/_data/engine-cli/docker_build.yaml
@@ -386,7 +386,7 @@ options:
   value_type: stringArray
   default_value: '[]'
   description: |
-    SSH agent socket or keys to expose to the build (only if BuildKit enabled) (format: default|<id>[=<socket>|<key>[,<key>]])
+    SSH agent socket or keys to expose to the build (only if BuildKit enabled) (format: default|\<id\>[=\<socket\>|\<key\>[,\<key\>]])
   deprecated: false
   min_api_version: "1.39"
   experimental: false
```

vpalmerini:
### Proposed changes
Fix formatting in templates of pages: `commandline/build`, `commandline/builder_build`, `commandline/buildx_build` and `commandline/image_build`.
The templates were not recognizing the `<` and `>` in the description of `--ssh` option, so this PR adds a `\` before each sign so the formatting can be preserved.

Before:
image

Now:
image

### Related issues
Closes #11331

### DIFF
--- a/_data/buildx/docker_buildx_build.yaml
+++ b/_data/buildx/docker_buildx_build.yaml
@@ -310,7 +310,7 @@ options:
   value_type: stringArray
   default_value: '[]'
   description: |
-    SSH agent socket or keys to expose to the build (format: default|<id>[=<socket>|<key>[,<key>]])
+    SSH agent socket or keys to expose to the build (format: default|\<id\>[=\<socket\>|\<key\>[,\<key\>]])
   deprecated: false
   experimental: false
   experimentalcli: false
--- a/_data/engine-cli/docker_build.yaml
+++ b/_data/engine-cli/docker_build.yaml
@@ -391,7 +391,7 @@ options:
   value_type: stringArray
   default_value: '[]'
   description: |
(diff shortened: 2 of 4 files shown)


thaJeztah (changes requested):
thanks for contributing! Unfortunately, I don't think this is the right fix, because these YAML files are generated from source, so manual updates to these files would be reverted when they are updated with a newly generated version.

Let me open a pull request with a fix to handle escaping these characters in our Jekyll template instead

thaJeztah:
opened https://github.com/docker/docker.github.io/pull/12083

vpalmerini:
Oh, makes sense. Thanks for the feedback

PR #12072, opened 2021-01-10
Fix ipvlan

bpmarinho:
Was setting `ipvlan`, since home router didn't support `macvlan`, and only started working when I noticed elsewhere that I needed to add the `-o parent=eth0` for `ipvlan` setup as it was set for `macvlan`.


usha-mandya:
@bpmarinho Thanks for the PR.

@arkodg Could you PTAL?

arkodg (approved):

PR #12076, opened 2021-01-11
Add language-specific guides

usha-mandya:

Adding Docker reference architecture docs


language/nodejs/run-tests.md
```diff
@@ -6,19 +6,7 @@ description: How to Build and Run your Tests using Node.js and Mocha frameworks
...
-We’ll use a fairly simple node.js application for this guide. To follow along, clone the following github repository.
-
-```shell
-$ git clone [ADD APPLICATION URL]
-```
-
-Our sample application uses an open source framework called Ronin.js for building simple REST services. It also has a mocks server that will accept JSON and return the same JSON when you send a GET request.
-
-So, for example, if we POST a JSON data structure to the endpoint /services/users and then perform a POST request to the same URI, we will receive the JSON which we first posted.
+Testing is an essential part of modern software development. Testing can mean a lot of things to different development teams. There are unit tests, integration tests and end-to-end testing. In this guide we take a look at running your unit tests in Docker. Let's assume we have defined Mocha tests in a `/test` folder within our application.
```
StefanScherer, line 9: NIT: It should be `test` folder, or `./test` folder to make it clear that it's a relative path to this sub directory.

language/nodejs/develop.md
```diff
@@ -0,0 +1,228 @@
...
+ mongodb:
+ mongodb_config:
+```
+
+This Compose file is super convenient as we do not have to type all the parameters to pass to the `docker run` command. We can declaratively do that in the Compose file.
+
+We are exposing port 9229 so that we can attach a debugger. We are also mapping our local source code into the running container so that we can make changes in our text editor and have those changes picked up in the container.
+
+One other really cool feature of using a Compose file, is that we have service resolution set up to use the service names. So we are now able to use `“mongo”` in our connection string. The reason we use mongo is because that is what we have named our mongo service in the Compose file as.
+
```
StefanScherer, line 148: _Add these steps to explain how to configure the debug step in the package.json:_

To be able to start our application in debug mode, we need to add a line to our `package.json` file to tell npm how to start our application in debug mode.

Open the `package.json` file and add the following line to the scripts section.

```
    "debug": "nodemon --inspect=0.0.0.0:9229 server.js"
```

As you can see we are going to use nodemon. Nodemon will start our server in debug mode and also watch for files that have changed and restart our server. Let’s add nodemon to our package.json file.

```
$ npm install nodemon
```

language/nodejs/run-tests.md
usha-mandya: Fixed

language/nodejs/develop.md
usha-mandya: Fixed.

StefanScherer (approved):
LGTM

PR #12142, opened 2021-01-20
clarification on .env file lookup vs project-directory

ndeloof:
### Proposed changes
clarification on `.env` file support in docker-compose and how it relates to project directory

see https://github.com/docker/compose/issues/7928


usha-mandya (approved):
LGTM

PR #12147, opened 2021-01-20
check for broken links

afeld:
Seeing [a number of issues around broken links in the docs](https://github.com/docker/docker.github.io/issues?q=is%3Aissue+is%3Aopen+broken+link); this should help avoid the issue.

### Proposed changes

This pull request adds broken link checking to continuous integration, which should point out broken links before those changes are merged.

n/a

### Related issues (optional)

[Many](https://github.com/docker/docker.github.io/issues?q=is%3Aissue+is%3Aopen+broken+link)

### DIFF
--- a/.github/workflows/build-pr.yml
+++ b/.github/workflows/build-pr.yml
@@ -14,6 +14,11 @@ jobs:
       - uses: actions/checkout@v2
       - name: build image
         run: docker build --target=current -t documentation:latest .
+      - name: check for broken links
+        uses: chabad360/htmlproofer@master
+        with:
+          directory: ./_site
+          arguments: --disable-external
       # Disabled netlify-deploy due to flakey 502 http errors
       # - name: copy static files
       #   if: github.event.pull_request.head.repo.fork == false


afeld:
Did a few spot checks of [the HTMLProofer output](https://github.com/docker/docker.github.io/pull/12147/checks?check_run_id=1738452012#step:7:15), and the failures seem to be legitimate (though redundant). Defer to the maintainers about where to go from here.

thaJeztah:
Thanks!! I'll try to have a look at the links that were reported broken.

We may also need to check if there's links that point to docs.docker.com (so would link to the online docs); does this action have a way to substitute those?

afeld:
A few ways to go about that, which aren't mutually exclusive:

- Switch them to relative/internal links
   - …in the repository
      - Recommend this, at the very least, since then links will work as expected in local development / preview site
      - Can be enforced via continuous integration
   - …during the continuous integration run, before the HTMLProofer check
      - Use `sed` to remove the `https://docs.docker.com/` on the Markdown files before build, or on the generated HTML after
- Include with [`--internal-domains docs.docker.com`](https://github.com/gjtorikian/html-proofer#configuration)

thaJeztah:
> Switch them to relative/internal links

Agreed, it's what's we try to do for links in this repository, that said, the occasional "full" URL finds its way in. There's also some content we pull in from other repositories; some of those may have links that (to make them "useful" for users browsing markdown files on GitHub) have full URLs to the online docs.

> Use sed to remove the https://docs.docker.com/ on the Markdown files before build, or on the generated HTML after

Yup, we do this in the Dockerfile (it's a bit ugly, but "does the job");
https://github.com/docker/docker.github.io/blob/cd5bc709406ef886503a2dee9a41e628385c62ca/Dockerfile#L76

That said, [as this famous answer on StackOverflow points out](https://stackoverflow.com/a/1732454); trying to modify HTML with regular expressions is, well "challenging", so there _might_ (and likely) will still be links that slip through.

> Include with `--internal-domains docs.docker.com`

Thanks! Yes, that looks interesting, and is roughly what I was looking for. Will have to check, but perhaps we'll have to include `docs-stage.docker.com` and `localhost:4000` to that list (depending on how the Dockerfile is built in CI). Perhaps not needed for those.


Overall, this looks **great** (thanks so much!); adding back a link-checker was still on my list of things to look in to; we did have this in place in the past (before we switched from Hugo to Jekyll (don't ask)), and some attempts were made since, with varying degrees of success.

I love ❤️ that the checker also checks for anchors, as those are often missed when content is moved around (or redirects are added)

Having a quick glance over the list of failures; most of those appear to be actual issues indeed (so something that needs to be fixed; some of the changes may have to be done in other repositories); given that we may not be able to fix them "all" immediately, perhaps this check could somehow be made "non-blocking" (notify that the check failed, but don't prevent merging) until we have a good "base".

I'll try to reserve some time to look more closely at the failures (feel free to open pull-requests if you see ones that can easily be fixed)

let me also ping @StefanScherer and @usha-mandya to have a look

afeld:
> this check could somehow be made "non-blocking"

You could point it at sub-directories and expand them out over time as they are fixed…?

PR #12206, opened 2021-01-30
The url of init db is error.

tmdhhl:
The url of init db is error.


thaJeztah:
Thanks for contributing!

Looks like this was fixed through https://github.com/docker/docker.github.io/pull/12397.

PR #12218, opened 2021-02-01
Clarify instructions

matt-continuousdelta:
* Eliminate unnecessary port change.
* Clarify code change and breakpoint location.


usha-mandya:
@matt-continuousdelta Thanks for the PR.

@StefanScherer Could you PTAL when you get a chance? Thanks.

StefanScherer (approved):
Hi @matt-continuousdelta 
Thank you very much. Your changes make a lot of sense and using the same port really makes it simpler for the reader.
I went through the instructions and they all look good.

PR #12224, opened 2021-02-01
Increase Windows 10 version requirement to 17134

mat007:
### Proposed changes

Increase Windows 10 version requirement to 17134.
16299 has been end of life since October 2020.


stephen-turner (approved):
Approved

PR #12230, opened 2021-02-02
Get started docs improvements

usha-mandya:

- Added a new section to download Docker and specific URLs for Mac and Window
- Updated headings to sentence style for consistency
- Updated numbered list
- Converted Pro-tip to 'Note' format
- Updated links to use relative paths instead of absolute URLs to docs.docker.com
- Added {:target="blank" rel="noopener" class=""} for external URLs
- Converted headings to task-oriented headings
- Updated the title and toc to match the new headings

### DIFF
--- a/get-started/02_our_app.md
+++ b/get-started/02_our_app.md
@@ -18,23 +18,22 @@ think about how it will work for a large team, multiple developers, etc.
 
 ![Todo List Manager Screenshot](images/todo-list-sample.png){: style="width:50%;" }
 
-
-## Getting our App
+## Get the app
 
 Before we can run the application, we need to get the application source code onto 
 our machine. For real projects, you will typically clone the repo. But, for this tutorial,
 we have created a ZIP file containing the application.
 
-1. [Download the App contents](https://github.com/docker/getting-started/tree/master/app). You can either pull the entire project or download it as a zip and extract the app folder out to get started with
+1. [Download the App contents](https://github.com/docker/getting-started/tree/master/app){:target="_blank" rel="noopener" class="_"}. You can either pull the entire project or download it as a zip and extract the app folder out to get started with
 
-1. Once extracted, use your favorite code editor to open the project. If you're in need of
-    an editor, you can use [Visual Studio Code](https://code.visualstudio.com/). You should
+2. Once extracted, use your favorite code editor to open the project. If you're in need of
+    an editor, you can use [Visual Studio Code](https://code.visualstudio.com/){:target="_blank" rel="noopener" class="_"}. You should
     see the `package.json` and two subdirectories (`src` and `spec`).
 
     ![Screenshot of Visual Studio Code opened with the app loaded](images/ide-screenshot.png){: style="width:650px;margin-top:20px;"}
     {: .text-center }
 
-## Building the App's Container Image
+## Build the app's container image
 
 In order to build the application, we need to use a `Dockerfile`. A
 Dockerfile is simply a text-based script of instructions that is used to
@@ -53,7 +52,7 @@ see a few flaws in the Dockerfile below. But, don't worry! We'll go over them.
 
     Please check that the file `Dockerfile` has no file extension like `.txt`. Some editors may append this file extension automatically and this would result in an error in the next step.
 
-1. If you haven't already done so, open a terminal and go to the `app` directory with the `Dockerfile`. Now build the container image using the `docker build` command.
+2. If you haven't already done so, open a terminal and go to the `app` directory with the `Dockerfile`. Now build the container image using the `docker build` command.
 
     ```bash
     docker build -t getting-started .
@@ -74,7 +73,7 @@ see a few flaws in the Dockerfile below. But, don't worry! We'll go over them.
 
     The `.` at the end of the `docker build` command tells that Docker should look for the `Dockerfile` in the current directory.
 
-## Starting an App Container
+## Start an app container
 
 Now that we have an image, let's run the application! To do so, we will use the `docker run`
 command (remember that from earlier?).
@@ -90,13 +89,13 @@ command (remember that from earlier?).
     background) and creating a mapping between the host's port 3000 to the container's port 3000.
     Without the port mapping, we wouldn't be able to access the application.
 
-1. After a few seconds, open your web browser to [http://localhost:3000](http://localhost:3000).
+2. After a few seconds, open your web browser to [http://localhost:3000](http://localhost:3000).
     You should see our app!
 
     ![Empty Todo List](images/todo-list-empty.png){: style="width:450px;margin-top:20px;"}
     {: .text-center }
 
-1. Go ahead and add an item or two and see that it works as you expect. You can mark items as
+3. Go ahead and add an item or two and see that it works as you expect. You can mark items as
    complete and remove items. Your frontend is successfully storing items in the backend!
    Pretty quick and easy, huh?
 
@@ -108,7 +107,6 @@ If you take a quick look at the Docker Dashboard, you should see your two contai
 (this tutorial and your freshly launched app container)!
 
 ![Docker Dashboard with tutorial and app containers running](images/dashboard-two-containers.png)
-
 
 ## Recap
 
--- a/get-started/02_our_app.md
+++ b/get-started/02_our_app.md
@@ -1,5 +1,5 @@
 ---
-title: "Our Application"
+title: "Sample application"
 keywords: get started, setup, orientation, quickstart, intro, concepts, containers, docker desktop
 redirect_from:
 - /get-started/part2/
--- a/get-started/07_multi_container.md
+++ b/get-started/07_multi_container.md
@@ -27,6 +27,8 @@ Remember that containers, by default, run in isolation and don't know anything a
 or containers on the same machine. So, how do we allow one container to talk to another? The answer is
 **networking**. Now, you don't have to be a network engineer (hooray!). Simply remember this rule...
 
+> **Note**
+>
 > If two containers are on the same network, they can talk to each other. If they aren't, they can't.
 
 ## Start MySQL
--- a/get-started/07_multi_container.md
+++ b/get-started/07_multi_container.md
@@ -3,6 +3,7 @@ title: "Multi container apps"
 keywords: get started, setup, orientation, quickstart, intro, concepts, containers, docker desktop
 description: Using more than one container in our application
 ---
+
 Up to this point, we have been working with single container apps. But, we now want to add MySQL to the
 application stack. The following question often arises - "Where will MySQL run? Install it in the same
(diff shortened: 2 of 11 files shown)


usha-mandya:
@thaJeztah Could you PTAL?

get-started/02_our_app.md
```diff
@@ -1,5 +1,5 @@
 ---
-title: "Our Application"
+title: "Sample application"
```
thaJeztah, line 2: Food for thought; for the page title itself, wondering if we should include "get started" on all of these pages. Doing so may help providing more context to the page.

Could be;

```
Get started: Sample application
```

or (if we want to include "what step" we're in)

```
Get started, step 1: Sample application
```


Currently, it shows:

Screenshot 2021-02-03 at 12 29 28




Alternatively, we could also look at modifying the template to add a suffix to the page-title (`<title>`) that differs from the H1, so that Google would show;

```
Sample application | Get started | Docker Documentation
```

or

```
Sample application | Docker Documentation: Get started
```

get-started/07_multi_container.md
```diff
@@ -165,21 +161,12 @@ The todo app supports the setting of a few environment variables to specify MySQ
...
->the variable and get the file contents.
-
+>**Warning**
+>
+>**Setting Connection Settings via Env Vars**
+>
+>While using env vars to set connection settings is generally ok for development, it is **HIGHLY DISCOURAGED** when running applications in production. Diogo Monica, the former lead of security at Docker, [wrote a fantastic blog post](https://diogomonica.com/2017/03/27/why-you-shouldnt-use-env-variables-for-secret-data/){:target="_blank" rel="noopener" class="_"} explaining why. A more secure mechanism is to use the secret support provided by your container orchestration framework. In most cases, these secrets are mounted as files in the running container. You'll see many apps (including the MySQL image and the todo app) also support env vars with a `_FILE` suffix to point to a file containing the variable. As an example, setting the `MYSQL_PASSWORD_FILE` var will cause the app to use the contents of the referenced file as the connection password. Docker doesn't do anything to support these env vars. Your app will need to know to look for the variable and get the file contents.
+{: .warning}
```
thaJeztah, line 169: Perhaps combine the "warning" with the next heading; something like

```markdown
> **Warning on using environment variables**
```

Looks like you un-wrapped this paragraph; for consistency, might want to add a space after each `>` as well;

```suggestion
> **Warning on using environment variables**
>
>
> While using env vars to set connection settings is generally ok for development,
> it is **HIGHLY DISCOURAGED** when running applications in production. Diogo Monica,
> the former lead of security at Docker, [wrote a fantastic blog post](https://diogomonica.com/2017/03/27/why-you-shouldnt-use-env-variables-for-secret-data/){:target="_blank" rel="noopener" class="_"}
> explaining why. A more secure mechanism is to use the secret support provided
> by your container orchestration framework. In most cases, these secrets are
> mounted as files in the running container. You'll see many apps (including the
> MySQL image and the todo app) also support env vars with a `_FILE` suffix to
> point to a file containing the variable. As an example, setting the
> `MYSQL_PASSWORD_FILE` var will cause the app to use the contents of the referenced
> file as the connection password. Docker doesn't do anything to support these
> env vars. Your app will need to know to look for the variable and get the file
> contents.
{: .warning}
```

get-started/08_using_compose.md
```diff
@@ -315,16 +310,13 @@ Now that we have our `docker-compose.yml` file, we can start it up!
     view the logs for a specific service, you can add the service name to the end of the logs command (for example,
     `docker-compose logs -f app`).
 
->**Pro tip**  Waiting for the DB before starting the app
```
thaJeztah, line 318: I think we can keep the "pro tip"

get-started/08_using_compose.md
```diff
@@ -315,16 +310,13 @@ Now that we have our `docker-compose.yml` file, we can start it up!
...
->
->When the app is starting up, it actually sits and waits for MySQL to be up and ready before trying to connect to it.
->Docker doesn't have any built-in support to wait for another container to be fully up, running, and ready
->before starting another container. For Node-based projects, you can use the 
->[wait-port](https://github.com/dwmkerr/wait-port) dependency. Similar projects exist for other languages/frameworks.
+    >**Waiting for the DB before starting the app**
+    >
+    >When the app is starting up, it actually sits and waits for MySQL to be up and ready before trying to connect to it. Docker doesn't have any built-in support to wait for another container to be fully up, running, and ready before starting another container. For Node-based projects, you can use the [wait-port](https://github.com/dwmkerr/wait-port){:target="_blank" rel="noopener" class="_"} dependency. Similar projects exist for other languages/frameworks.
```
thaJeztah, line 316: Looks like this was un-wrapped;

```suggestion
    > **Pro tip: wait for the DB before starting the app**
    >
    > When the app is starting up, it actually sits and waits for MySQL to be up
    > and ready before trying to connect to it. Docker doesn't have any built-in
    > support to wait for another container to be fully up, running, and ready
    > before starting another container. For Node-based projects, you can use
    > the [wait-port](https://github.com/dwmkerr/wait-port){:target="_blank" rel="noopener" class="_"}
    > dependency. Similar projects exist for other languages/frameworks.

```

get-started/index.md
```diff
@@ -60,14 +60,24 @@ Welcome! We are excited that you want to learn Docker.
...
 
+## Download Docker
+
+You can download and install Docker on multiple platforms. Choose your preferred operating system below.
+
+- [Mac](https://desktop.docker.com/mac/stable/Docker.dmg)
+
+- [Windows](https://desktop.docker.com/win/stable/Docker%20Desktop%20Installer.exe)
+
+- [Linux](../engine/install/index.md)
```
thaJeztah, line 79: Should be no blank line between bullets here.

Probably want to add a "blank" target to make sure the user doesn't navigate away from the page itself. Perhaps make the link captions more descriptive, e.g;


```suggestion
- [Download Docker Desktop for Mac](https://desktop.docker.com/mac/stable/Docker.dmg){:target="_blank" rel="noopener" class="_"}
- [Download Docker Desktop for Windows](https://desktop.docker.com/win/stable/Docker%20Desktop%20Installer.exe){:target="_blank" rel="noopener" class="_"}
- [Install docker Engine on Linux](../engine/install/index.md){:target="_blank" rel="noopener" class="_"}
```

Should these links point to the _install_ pages instead? (which contains information that may be relevant, such as system requirements);

```suggestion
- [Install Docker Desktop for Mac](../docker-for-mac/install.md){:target="_blank" rel="noopener" class="_"}
- [Install Docker Desktop for Windows](../docker-for-windows/install.md){:target="_blank" rel="noopener" class="_"}
- [Install docker Engine on Linux](../engine/install/index.md){:target="_blank" rel="noopener" class="_"}
```

get-started/index.md
```diff
@@ -60,14 +60,24 @@ Welcome! We are excited that you want to learn Docker.
...
-If you are looking for information on how to containerize an application using your favorite language, see [Language-specific getting started guides](/language).
+If you are looking for information on how to containerize an application using your favorite language, see [Language-specific getting started guides](../language/index.md).
 
 We also recommend the video walkthrough from DockerCon 2020.
 
 <iframe width="560" height="315" src="https://www.youtube-nocookie.com/embed/iqqDU2crIEQ?start=30" frameborder="0" allow="accelerometer; autoplay; encrypted-media; gyroscope; picture-in-picture" allowfullscreen></iframe>
 
 In this tutorial, you'll learn about creating and deploying Docker apps, including using multiple containers with a database, and using Docker Compose. You'll also deploy your containerized app to Azure.
 
+## Download Docker
```
thaJeztah, line 71: Perhaps

```suggestion
## Install Docker
```

or

```suggestion
## Download and install Docker
```

get-started/index.md
```diff
@@ -60,14 +60,24 @@ Welcome! We are excited that you want to learn Docker.
...
+
+You can download and install Docker on multiple platforms. Choose your preferred operating system below.
```
thaJeztah, line 73: Perhaps something like;

```suggestion
This tutorial assumes you have a current version of Docker installed on your
machine. If you do not have Docker installed, you can download and install Docker.
Choose your preferred operating system below for installation instructions:
```

thaJeztah (changes requested):
Thanks! I left some suggestion, and a "food for thought" comment 😅

get-started/07_multi_container.md
usha-mandya: Aah, I thought we've used the unwrapped version for Notes and other admonitions. The output looks similar though. Any specific reason for wrapping these?

I'll add a space after `>`

get-started/08_using_compose.md
usha-mandya: Believe this is the only section where we have Pro tips. I'd be inclined to use "Note" instead for consistency.

get-started/index.md
usha-mandya: These pages will be linked from the onboarding page. Would add the direct download links here instead of pointing to yet another page. I can add another link to Desktop pages for for installation instructions

get-started/07_multi_container.md
thaJeztah: Mostly the same as for other content; the wrapping helps with a) readability when reading the source, and b) reviews (being able to comment / make changes on a specific line)

get-started/02_our_app.md
usha-mandya: Adding Get started on each page would be a repetition. Updating the suffix might be a better option. Will think about this. Let's discuss during our next sync.

get-started/08_using_compose.md
thaJeztah: We have `Tip:` in a couple of places (e.g. https://docs.docker.com/config/containers/logging/configure/)

I think a "note" is more to highlight / go further into details of a change/example, whereas a "tip" is slightly more "consider this" (but not required to do, so "ok to skip" when reading)

get-started/index.md
thaJeztah: > These pages will be linked from the onboarding page. Would add the direct download links here instead of pointing to yet another page. I can add another link to Desktop pages for for installation instructions

So, we of course don't control "how" people arrive on this page (quite likely "through Google"), in which case this page is the first page they visit. From that perspective, I think it's important to provide the "install" step in context, instead of just the download links. (The installation page also, e.g., mentions that they agree to the EULA).

If we link from the on boarding page (the "designed" flow); does that flow first take them through the installation process? If so, users that would follow _that_ flow wouldn't need this section at all.

thaJeztah (approved):
LGTM

still a bit on the fence w.r.t. the "install" section 😅

usha-mandya:
Thanks @thaJeztah. Happy to discuss that further. I'd like to get the updates live today. We can discuss further updates and create a new PR for the changes.

PR #12234, opened 2021-02-03
update rootless docs

### LINKED ISSUE #12050: Conflicting documntation
[engine/install/linux-postinstall.md](https://docs.docker.com/engine/install/linux-postinstall/) says that rootless regime is experimental, but https://docs.docker.com/engine/security/rootless/ contradicts it, saying it is already matured.

### LINKED ISSUE #12053: Request docs changes: rootless installation
[engine/security/rootless.md](https://docs.docker.com/engine/security/rootless/)

https://github.com/docker/docker.github.io/blob/1976c2178c6317400fdbdc0933570c5ef9248ab6/engine/security/rootless.md

1) To install Rootless Docker as the "ROOT" user, see the Manual installation steps.

But the purpose of this page is installation without root privilege...

2) Manual installation (https://docs.docker.com/engine/security/rootless/#manual-installation)

Seems a placeholder, people don't know how to manually install without dig https://get.docker.com/rootless

3) It's better to provide the Uninstallation as the complement. This will make user more confident about the installation.

AkihiroSuda:
## _includes/install-script.md, engine/install/linux-postinstall.md
- Remove "Rootless mode is currently available as an experimental feature."
  Close #12050

###  "Prerequiresites" section
- Remove information about old distros (Debian 9, CentOS 7.5-7.6)

###  "Distribution-specific hint" section
- Tabified (`<div class="tab-content" />`)

### "Known limitations" section
- Kernel 5.11 supports rootless overlayfs, without the Ubuntu/Debian patch. https://github.com/torvalds/linux/commit/459c7c565ac36ba09ffbf24231147f408fde4203

###  "Install" section
- Promote RPM/DEB installation over TGZ installation.
  See docker/roadmap#188

### "Uninstall" section
- Add "Uninstall" section.
  Close #12053

### "Usage" section
- Added more information about systemd
- Move `nsenter` tips to "Tips for debugging" subsection under "Troubleshooting" section

### "Best practice" section
- Remove guide for `lxc-user-nic` network driver due to immaturity.
  Will be brought back in future.
  See rootless-containers/rootlesskit#138 .

### "Troubleshooting" section
- Add a guide for "can't open lock file /run/xtables.lock: Permission denied" (SELinux).
  See moby/moby#41230

- Add a guide for "failed to register layer: ApplyLayer exit status 1 ..." (NFS).
  Close docker/for-linux#1172

- Improve guides for slirp4netns.

- Remove v19.03 information (e.g., "cgroup v2 is unsupported, use cgroup v1")

# Preview
https://deploy-preview-12234--docsdocker.netlify.app/engine/security/rootless/

### DIFF
--- a/engine/security/rootless.md
+++ b/engine/security/rootless.md
@@ -50,16 +50,25 @@ testuser:231072:65536
 
 > Note: We recommend that you use the Ubuntu kernel.
 
-#### Ubuntu
+<ul class="nav nav-tabs">
+  <li class="active"><a data-toggle="tab" data-target="#hint-ubuntu">Ubuntu</a></li>
+  <li><a data-toggle="tab" data-target="#hint-debian">Debian GNU/Linux</a></li>
+  <li><a data-toggle="tab" data-target="#hint-arch">Arch Linux</a></li>
+  <li><a data-toggle="tab" data-target="#hint-opensuse">openSUSE</a></li>
+  <li><a data-toggle="tab" data-target="#hint-centos8-and-fedora">CentOS 8 and Fedora</a></li>
+  <li><a data-toggle="tab" data-target="#hint-centos7">CentOS 7</a></li>
+</ul>
+<div class="tab-content">
 
+<div id="hint-ubuntu" class="tab-pane fade in active" markdown="1">
 - No preparation is needed.
 
 - `overlay2` storage driver  is enabled by default
   ([Ubuntu-specific kernel patch](https://kernel.ubuntu.com/git/ubuntu/ubuntu-bionic.git/commit/fs/overlayfs?id=3b7da90f28fe1ed4b79ef2d994c81efbc58f1144)).
 
 - Known to work on Ubuntu 16.04, 18.04, and 20.04.
-
-#### Debian GNU/Linux
+</div>
+<div id="hint-debian" class="tab-pane fade in" markdown="1">
 - Add `kernel.unprivileged_userns_clone=1` to `/etc/sysctl.conf` (or
   `/etc/sysctl.d`) and run `sudo sysctl --system`.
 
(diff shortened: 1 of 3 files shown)


AkihiroSuda:
cc @thaJeztah @tonistiigi @tiborvass

usha-mandya:
@thaJeztah Could you PTAL?

soinu:
This solution / recommended installation guide does not look like the best solution. `docker-ce-rootless-extras` depends on `docker-ce`, therefore I need to install rootful docker in order to get rootless docker installed.
By default the rootful daemon is started too, something you definitely don't want on a rootless installation.

thaJeztah:
> therefore I need to install rootful docker in order to get rootless docker installed

Yes, it's a bit of a pain that Linux does not have a concept of "per-user" installs (when using package managers). The downside of the alternative is, that using statically-linked binaries may not be best-recommendation either

> By default the rootful daemon is started too, something you definitely don't want on a rootless installation.

We should probably add a step that describes how to disable the rootful daemon (for the `.deb` case, as RPMs don't automatically start after install (but the `get.docker.com` script may do this, in which case, probably for RPM as well).

soinu:
Is a "system wide" systemd service (e.g. in /etc/systemd/system) with User= flag supported? In any case, I think it should be mentioned in the docs.

AkihiroSuda:
Updated https://deploy-preview-12234--docsdocker.netlify.app/engine/security/rootless/


> This solution / recommended installation guide does not look like the best solution.

Tabified "With packages" and "Without packages" to weaken the "recommendation" tone. 
https://deploy-preview-12234--docsdocker.netlify.app/engine/security/rootless/#install

> We should probably add a step that describes how to disable the rootful daemon (for the .deb case, as RPMs don't automatically start after install (but the get.docker.com script may do this, in which case, probably for RPM as well).

Added in "Install" section https://deploy-preview-12234--docsdocker.netlify.app/engine/security/rootless/#install

> Is a "system wide" systemd service (e.g. in /etc/systemd/system) with User= flag supported? In any case, I think it should be mentioned in the docs.

No. Explained in the "Usage" section. https://deploy-preview-12234--docsdocker.netlify.app/engine/security/rootless/#usage

konstruktoid:
> This solution / recommended installation guide does not look like the best solution. `docker-ce-rootless-extras` depends on `docker-ce`, therefore I need to install rootful docker in order to get rootless docker installed.
> By default the rootful daemon is started too, something you definitely don't want on a rootless installation.

I brought this up in https://github.com/docker/roadmap/issues/188#issuecomment-771873492 as well, and "the solution" (aka the way I do it) is to use the static built packages or disable/mask the rootful Docker service after installation.

I did a Ansible example write-up about this at https://github.com/konstruktoid/ansible-docker-rootless#package-docker-installation

soinu:
> No. Explained in the "Usage" section. https://deploy-preview-12234--docsdocker.netlify.app/engine/security/rootless/#usage

> Starting Rootless Docker as a systemd-wide service (/etc/systemd/system/docker.service) is not supported, even with the User= > directive.

Thanks. Can you explain why this is not supported (probably enough if it is explained here on GitHub)?

AkihiroSuda:
Because we need to have a user session for creating XDG_RUNTIME_DIR (/run/user/$UID) and delegating cgroup on cgroup v2 hosts

soinu:
> we need to have a user session for creating XDG_RUNTIME_DIR (/run/user/$UID)

We could use any other directory? Does it really have to be /run/user/?

> delegating cgroup on cgroup v2 

This sounds like a limitation of systemd. I tried to research this a bit but couldn't find anything helpful so far. Can you explain this a bit more or link an article? Is work being done to support this?


While I understand that systemd with --user is helpful on systems where you don't have root access it's really annoying on systems where you do have root access and need such services in multiple accounts. In addition, with support for a USER flag we would no longer need lingering

AkihiroSuda:
That topic is beyond the scope of this PR, and should be better discussed in https://github.com/moby/moby/issues

AkihiroSuda:
@thaJeztah PTAL?

engine/security/rootless.md
```diff
@@ -50,72 +50,71 @@ testuser:231072:65536
...
 
-#### Ubuntu
-
+<ul class="nav nav-tabs">
+  <li class="active"><a data-toggle="tab" data-target="#hint-ubuntu">Ubuntu</a></li>
+  <li><a data-toggle="tab" data-target="#hint-debian">Debian GNU/Linux</a></li>
+  <li><a data-toggle="tab" data-target="#hint-arch">Arch Linux</a></li>
+  <li><a data-toggle="tab" data-target="#hint-opensuse">openSUSE</a></li>
+  <li><a data-toggle="tab" data-target="#hint-centos8-and-fedora">CentOS 8 and Fedora</a></li>
+  <li><a data-toggle="tab" data-target="#hint-centos7">CentOS 7</a></li>
```
thaJeztah, line 59: Given that we don't ship packages for Arch Linux or openSUSE, perhaps we should consider removing those instructions. At least, it may be good to move them as last tabs (and perhaps order the tabs base on "most common")

We can do that in a follow-up though; I see some other possible improvements, so may reserve some time to do a follow up (I'll ask you to review those 👍)

thaJeztah (approved):
LGTM! sorry for the delay, I started reviewing this, but saw I didn't submit 😓

PR #12237, opened 2021-02-03
engine/security: add more description about ssh://

AkihiroSuda:
### Proposed changes

Previously, the `ssh://` helper was only mentioned in `engine/security/index.md`.

The `ssh://` helper is now documented in "Protect the Docker daemon socket" (`engine/security/protect-access.md`, nee `engine/security/https.md`).

### DIFF
--- a/engine/install/linux-postinstall.md
+++ b/engine/install/linux-postinstall.md
@@ -151,7 +151,7 @@ the [Docker CLI Reference](/engine/reference/commandline/dockerd/) article.
 > understand the security implications of opening docker to the network. If steps are not taken to secure the connection, 
 > it is possible for remote non-root users to gain root access on the host. For more information on how to use TLS 
 > certificates to secure this connection, check this article on 
-> [how to protect the Docker daemon socket](../security/https.md).
+> [how to protect the Docker daemon socket](../security/protect-access.md).
 {: .warning}
 
 Configuring Docker to accept remote connections can be done with the `docker.service`
--- a/_data/toc.yaml
+++ b/_data/toc.yaml
@@ -205,7 +205,7 @@ guides:
         title: Docker security
       - path: /engine/security/non-events/
         title: Docker security non-events
-      - path: /engine/security/https/
+      - path: /engine/security/protect-access/
         title: Protect the Docker daemon socket
       - path: /engine/security/certificates/
         title: Using certificates for repository client verification
--- a/engine/security/certificates.md
+++ b/engine/security/certificates.md
@@ -6,7 +6,7 @@ redirect_from:
 title: Verify repository client with certificates
 ---
 
-In [Running Docker with HTTPS](https.md), you learned that, by default,
+In [Running Docker with HTTPS](protect-access.md), you learned that, by default,
 Docker runs via a non-networked Unix socket and TLS must be enabled in order
 to have the Docker client and the daemon communicate securely over HTTPS.  TLS ensures authenticity of the registry endpoint and that traffic to/from registry is encrypted.
 
@@ -92,4 +92,4 @@ If the Docker registry is accessed without a port number, do not add the port to
 ## Related information
 
 * [Use trusted images](trust/index.md)
-* [Protect the Docker daemon socket](https.md)
+* [Protect the Docker daemon socket](protect-access.md)
--- a/engine/security/https/README.md
+++ b/engine/security/https/README.md
@@ -2,7 +2,7 @@
 published: false
 ---
 
-This is an initial attempt to make it easier to test the examples in the https.md
+This is an initial attempt to make it easier to test the TLS (HTTPS) examples in the protect-access.md
 doc.
 
 At this point, it is a manual thing, and I've been running it in boot2docker.
--- a/engine/security/https/parsedocs.sh
+++ b/engine/security/https/parsedocs.sh
@@ -1,7 +1,7 @@
 #!/bin/sh
 
 echo "#!/bin/sh"
-cat ../https.md | awk '{if (sub(/\\$/,"")) printf "%s", $0; else print $0}' \
+cat ../protect-access.md | awk '{if (sub(/\\$/,"")) printf "%s", $0; else print $0}' \
         | grep '   $ ' \
         | sed 's/    $ //g' \
         | sed 's/2375/7777/g' \
--- a/engine/security/index.md
+++ b/engine/security/index.md
@@ -116,7 +116,7 @@ Note that even if you have a firewall to limit accesses to the REST API
 endpoint from other hosts in the network, the endpoint can be still accessible
 from containers, and it can easily result in the privilege escalation.
 Therefore it is *mandatory* to secure API endpoints with 
-[HTTPS and certificates](https.md).
+[HTTPS and certificates](protect-access.md).
 It is also recommended to ensure that it is reachable only from a trusted
 network or VPN.
 
--- a/engine/security/protect-access.md
+++ b/engine/security/protect-access.md
@@ -1,17 +1,84 @@
 ---
-description: How to setup and run Docker with HTTPS
-keywords: docker, docs, article, example, https, daemon, tls, ca,  certificate
+description: How to setup and run Docker with SSH or HTTPS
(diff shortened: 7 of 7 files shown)


AkihiroSuda:
cc @tonistiigi

engine/install/linux-postinstall.md
```diff
@@ -151,7 +151,7 @@ the [Docker CLI Reference](/engine/reference/commandline/dockerd/) article.
 > understand the security implications of opening docker to the network. If steps are not taken to secure the connection, 
 > it is possible for remote non-root users to gain root access on the host. For more information on how to use TLS 
 > certificates to secure this connection, check this article on 
-> [how to protect the Docker daemon socket](../security/https.md).
+> [how to protect the Docker daemon socket](../security/protect-socket.md).
```
thaJeztah, line 154: Thinking out loud; would `secure-api`, `protect-access` or something along those lines be more descriptive? (as it's not necessarily the "socket" we're protecting, but protecting access to the engine API.

Perhaps we should move this warning to an include-file, so that we can more easily re-use it (thinking of adding the same / similar warning on https://docs.docker.com/engine/api/)

engine/security/protect-socket.md
```diff
@@ -1,17 +1,41 @@
...
 By default, Docker runs through a non-networked UNIX socket. It can also
-optionally communicate using an HTTP socket.
+optionally communicate using SSH or a TLS (HTTPS) socket.
 
-If you need Docker to be reachable through the network in a safe manner, you can
-enable TLS by specifying the `tlsverify` flag and pointing Docker's
+## Use SSH to protect the Docker daemon socket
+
+The `docker` CLI can talk to a remote Docker daemon via SSH,
+by specifying `$DOCKER_HOST` to `ssh://<USERNAME>@<HOST>`.
```
thaJeztah, line 17: Perhaps would be good to;

- use a different username than `root` as example (to encourage setting up a user for docker; https://docs.docker.com/engine/install/linux-postinstall/)
- should we describe that `username` is optional (as it can be configured in `~/.ssh/config`) ?
- _slightly_ prefer (e.g.) `host1.example.com` to prevent confusing the above from being an e-mail address (`root@example.com` would be easily confused to be an e-mail address)
- I think we should describe using `docker context` as the "main" method to do this (assuming this is not for an "ad-hoc" situation)

engine/security/protect-socket.md
```diff
@@ -1,17 +1,41 @@
...
+
+```console
+$ export DOCKER_HOST=ssh://root@example.com
+$ docker ps
+```
+
+The support for SSH is available since Docker Engine 18.09.
+
```
thaJeztah, line 25: I think we can remove this, as 18.09 is EOL, so any "current" version of docker now supports it (people shouldn't be using older versions as they're insecure now)

```suggestion
```

engine/security/protect-socket.md
```diff
@@ -1,17 +1,41 @@
...
-enable TLS by specifying the `tlsverify` flag and pointing Docker's
+## Use SSH to protect the Docker daemon socket
+
+The `docker` CLI can talk to a remote Docker daemon via SSH,
+by specifying `$DOCKER_HOST` to `ssh://<USERNAME>@<HOST>`.
+
+```console
+$ export DOCKER_HOST=ssh://root@example.com
+$ docker ps
+```
```
thaJeztah, line 22: Perhaps something like this:

> **Note**
>
> The given `USERNAME` must have permissions to access the docker socket on the
> remote machine. Refer to [manage Docker as a non-root user](../../install/linux-postinstall.md#manage-docker-as-a-non-root-user)
> to learn how to give a non-root user access to the docker socket.

The following example creates a [`docker context`](../../context/working-with-contexts.md)
to connect with a remote `dockerd` daemon on `host1.example.com` using SSH, and
as the `docker-user` user on the remote machine:

```console
$ docker context create \
    --docker host=ssh://docker-user@host1.example.com \
    --description="Remote engine" \
    my-remote-engine

my-remote-engine
Successfully created context "my-remote-engine"
```

After creating the context, use `docker context use` to switch the `docker` CLI
to use it, and to connect to the remote engine:

```console
$ docker context use my-remote-engine
my-remote-engine
Current context is now "my-remote-engine"

$ docker info
<prints output of the remote engine>
```

Use use the `default` context to switch back to the default (local) daemon:

```console
$ docker context use default
default
Current context is now "default"
```

Alternatively, use the `DOCKER_HOST` environment variable to temporarily switch
the `docker` CLI to connect to the remote host using SSH. This does not require
creating a context, and can be useful to create an ad-hoc connection with a different
engine:

```console
$ export DOCKER_HOST=ssh://docker-user@host1.example.com
$ docker info
<prints output of the remote engine>
```

thaJeztah: Markdown for the above (GitHub "suggestions" didn't like the nested code-fences 😞)

    > **Note**
    >
    > The given `USERNAME` must have permissions to access the docker socket on the
    > remote machine. Refer to [manage Docker as a non-root user](../../install/linux-postinstall.md#manage-docker-as-a-non-root-user)
    > to learn how to give a non-root user access to the docker socket.

    The following example creates a [`docker context`](../../context/working-with-contexts.md)
    to connect with a remote `dockerd` daemon on `host1.example.com` using SSH, and
    as the `docker-user` user on the remote machine:

    ```console
    $ docker context create \
        --docker host=ssh://docker-user@host1.example.com \
        --description="Remote engine" \
        my-remote-engine

    my-remote-engine
    Successfully created context "my-remote-engine"
    ```

    After creating the context, use `docker context use` to switch the `docker` CLI
    to use it, and to connect to the remote engine:

    ```console
    $ docker context use my-remote-engine
    my-remote-engine
    Current context is now "my-remote-engine"

    $ docker info
    <prints output of the remote engine>
    ```

    Use use the `default` context to switch back to the default (local) daemon:

    ```console
    $ docker context use default
    default
    Current context is now "default"
    ```

    Alternatively, use the `DOCKER_HOST` environment variable to temporarily switch
    the `docker` CLI to connect to the remote host using SSH. This does not require
    creating a context, and can be useful to create an ad-hoc connection with a different
    engine:

    ```console
    $ export DOCKER_HOST=ssh://docker-user@host1.example.com
    $ docker info
    <prints output of the remote engine>
    ```

thaJeztah (changes requested):
Thanks! I left some thoughts/suggestions.

Wondering if the Dockerfile and shell-scripts are still used anywhere 🤔 (if they're used by external people, we may need to check if that's a problem, as we won't have redirects for them)

AkihiroSuda:
Thanks, updated

> Wondering if the Dockerfile and shell-scripts are still used anywhere 🤔 (if they're used by external people, we may need to check if that's a problem, as we won't have redirects for them)

Reverted Dockerfile and scripts

engine/security/protect-socket.md
AkihiroSuda: updated

engine/security/protect-socket.md
AkihiroSuda: updated

engine/security/protect-socket.md
AkihiroSuda: updated

engine/install/linux-postinstall.md
AkihiroSuda: Changed to protect-access

thaJeztah:
Oh! found one more occurrence of the old URL in the TOC; https://github.com/docker/docker.github.io/blob/60624cfb4a44268db19715a9ce6cf6c4a3b112f9/_data/toc.yaml#L208

Other than that, this looks great now, thanks! (Guess it can be moved out of draft?)

AkihiroSuda:
Thanks, fixed

AkihiroSuda:
Preview is here: https://deploy-preview-12237--docsdocker.netlify.app/engine/security/protect-access/

thaJeztah (approved):
LGTM!

This is a great enhancement, thank you!

Reading the page, there's a couple of other things that I think could use some enhancement, but adding that to my list to work on if I find time 😅

PR #12242, opened 2021-02-03
Add remark about non-glibc based distros in WSL docs

simonferquel:
Signed-off-by: Simon Ferquel <simon.ferquel@docker.com>

### Proposed changes

This adds a remark about the glibc dependency in Docker Desktop WSL backend.

### DIFF
--- a/docker-for-windows/wsl.md
+++ b/docker-for-windows/wsl.md
@@ -77,6 +77,8 @@ Ensure you have completed the steps described in the Prerequisites section **bef
 
     Optionally, select any additional distributions you would like to enable WSL 2 on.
 
+    Please note that the WSL Integrations bits running in your distro depend on glibc. This make it difficult to run it on musl-based distros such as Alpine Linux.
+
     ![WSL 2 Choose Linux distro](images/wsl2-choose-distro.png)
 
 8. Click **Apply & Restart**.
--- a/docker-for-windows/wsl.md
+++ b/docker-for-windows/wsl.md
@@ -77,7 +77,7 @@ Ensure you have completed the steps described in the Prerequisites section **bef
 
     Optionally, select any additional distributions you would like to enable WSL 2 on.
 
-    Please note that the WSL Integrations bits running in your distro depend on glibc. This make it difficult to run it on musl-based distros such as Alpine Linux.
+    Please note that the WSL Integrations bits running in your distro depend on glibc. This make it difficult to run it on musl-based distros such as Alpine Linux. Alpine users can use [this package](https://github.com/sgerrand/alpine-pkg-glibc) to deploy a glibc side by side with musl to run the integration.
 
     ![WSL 2 Choose Linux distro](images/wsl2-choose-distro.png)
 
--- a/docker-for-windows/wsl.md
+++ b/docker-for-windows/wsl.md
@@ -77,7 +77,9 @@ Ensure you have completed the steps described in the Prerequisites section **bef
 
     Optionally, select any additional distributions you would like to enable WSL 2 on.
 
-    Please note that the WSL Integrations bits running in your distro depend on glibc. This make it difficult to run it on musl-based distros such as Alpine Linux. Alpine users can use [this package](https://github.com/sgerrand/alpine-pkg-glibc) to deploy a glibc side by side with musl to run the integration.
+    > **Note**
+    >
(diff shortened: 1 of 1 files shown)


docker-for-windows/wsl.md
```diff
@@ -77,6 +77,8 @@ Ensure you have completed the steps described in the Prerequisites section **bef
 
     Optionally, select any additional distributions you would like to enable WSL 2 on.
 
+    Please note that the WSL Integrations bits running in your distro depend on glibc. This make it difficult to run it on musl-based distros such as Alpine Linux.
```
stephen-turner, line 80: Are there any instructions to make it work? I think we should either say that it won't work, or here are the instructions, not just that it will be "difficult".

docker-for-windows/wsl.md
```diff
@@ -77,6 +77,8 @@ Ensure you have completed the steps described in the Prerequisites section **bef
...
+
```
stephen-turner, line 81: Another comment about this section while you're editing it. We talk on lines 74 and 80 about "WSL integration" and on line 78 about "enabling WSL 2". Isn't it really "Docker integration" and "enabling Docker" in that distro? I think that would be clearer.

stephen-turner (changes requested):

docker-for-windows/wsl.md
simonferquel: Right. I am adding a link to the project we use in CI

docker-for-windows/wsl.md
simonferquel: For now we distinguish:
- Enabling the WSL 2 backend (running docker within WSL 2)
- Enabling the WSL integration (which is running the proxy to enable the docker experience within any of the user's WSL distros).

Please note that in the docs we refer to terms present in the UI. If we want to change that (and I agree with you, the vocabulary is confusing, and was obviously chosen by a french software engineer rather than by a native english PM :D), we need to do it both in UI (settings screen) and docs in sync.

stephen-turner: Well, I only just thought of it too. I think we should call it the "Docker-WSL integration" (in both UI and docs): then it's clear from both directions. I'll make a ticket. cc @usha-mandya

usha-mandya: Agree. We should clarify that it is the Docker integration within the WSL distro. Not sure if we should hold this PR up until the changes are in.

docker-for-windows/wsl.md
```diff
@@ -77,6 +77,8 @@ Ensure you have completed the steps described in the Prerequisites section **bef
 
     Optionally, select any additional distributions you would like to enable WSL 2 on.
 
+    Please note that the WSL Integrations bits running in your distro depend on glibc. This make it difficult to run it on musl-based distros such as Alpine Linux. Alpine users can use [this package](https://github.com/sgerrand/alpine-pkg-glibc) to deploy a glibc side by side with musl to run the integration.
```
usha-mandya, line 80:
```suggestion
    > **Note**
    >
    > WSL Integration components running in your distro depend on glibc. This can cause issues when running WSL 2 on musl-based distros such as Alpine Linux. Alpine users can use the [alpine-pkg-glibc(https://github.com/sgerrand/alpine-pkg-glibc){:target="_blank" rel="noopener" class="_"}  package to deploy a glibc with musl to run the integration.
```

usha-mandya: @simonferquel made some suggestions. PTAL and let me know if this sounds correct?

docker-for-windows/wsl.md
stephen-turner: Agreed, we can do it separately from this PR.

docker-for-windows/wsl.md
mat007: nit: double space before `package`, not sure it shows up in the final output (probably not).

usha-mandya: Fixed. Thanks Mathieu :)

simonferquel:
@stephen-turner do you still have changes in mind (aside from the vocabulary fix we need to plan - with fixing that in the desktop UI as well)

usha-mandya (approved):
LGTM.

stephen-turner:
No, no more changes, it looks good to me now.

stephen-turner (approved):

PR #12266, opened 2021-02-06
Self-signed certificate with Kubernetes fix

novakivanovski:
In Go 1.15 there is deprecation for using Common name:
https://golang.google.cn/doc/go1.15#commonname

This causes the self-signed cert to be rejected by some Kubernetes platforms:
https://stackoverflow.com/questions/64814173/how-do-i-use-sans-with-openssl-instead-of-common-name

Fix is to add this field to the cert.


StefanScherer (approved):
LGTM
Thank you @novakivanovski !

PR #12280, opened 2021-02-08
Change "WSL 2 integration" to "Docker-WSL integration"

stephen-turner:
As discussed privately, change "WSL 2 integration" to "Docker-WSL integration" because the former is confusing when in the context of doing something to a WSL distro.


usha-mandya (approved):
LGTM. Thanks.

PR #12282, opened 2021-02-08
Show command how to install Rosetta 2

StefanScherer:
The preview builds of Docker Desktop for Apple silicon cannot show the Apple dialog to install Rosetta 2. This nice little helper dialog only pops up for Intel apps as described here in the Apple support pages: [If you need to install Rosetta on your Mac](https://support.apple.com/en-us/HT211861)

So instead Docker Desktop shows this dialog on an Apple silicon machine that hasn't Rosetta 2 installed:

docker-desktop-arm64-requires-rosetta-error

The link opens our troubleshoot page and we  should at least show the command to install Rosetta from a terminal.

In the long term we want to ship all binaries as native arm64 mac binaries to avoid this dependency,  but for the preview builds we still need this trick.

### Related issues (optional)

internal JIRA DESKTOP-3583

### DIFF
--- a/docker-for-mac/apple-m1.md
+++ b/docker-for-mac/apple-m1.md
@@ -23,7 +23,10 @@ Click the following link to download the Apple M1 tech preview build:
 The tech preview of Docker Desktop for Apple M1 currently has the following limitations:
 
 - The tech preview build does not update automatically. You must manually install any future versions of Docker Desktop.
-- You must install Rosetta 2 as some binaries are still Darwin/AMD64.
+- You must install Rosetta 2 as some binaries are still Darwin/AMD64. To install Rosetta 2 manually from the command line use this command:
+    ```
+    softwareupdate --install-rosetta
+    ```
 - The DNS name `host.docker.internal` only works if you add `--add-host=host.docker.internal:host-gateway` to the `docker run` command
 - The DNS name `vm.docker.internal` does not work.
 - Kubernetes does not initialize because of a missing DNS name.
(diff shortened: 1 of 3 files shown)


stephen-turner (changes requested):
I think this is the wrong place in the docs. I think the /go/ link should redirect to https://docs.docker.com/docker-for-mac/apple-m1/ and the command to install Rosetta should be added to that page instead. (That page already mentions that Rosetta is needed, but it would be useful to add the specific command).

StefanScherer:
I've updated this PR.
- The troubleshooting page no longer links to the roadmap ticket, but instead to the apple-m1 page to keep the reader in our docs pages.
- Moved the command how to install Rosetta 2 to the apple-m1 page.
- Changed the redirect from "/go/apple-silicon" to the apple-m1 page. Can be tested here: https://deploy-preview-12282--docsdocker.netlify.app/go/apple-silicon
-

StefanScherer:
BTW: We should rename the apple-m1 page to apple-silicon, later this year there will probably a M2, M3, ... 😅

usha-mandya:
> BTW: We should rename the apple-m1 page to apple-silicon, later this year there will probably a M2, M3, ... 😅

Apple still refer to this as M1 - https://www.apple.com/uk/mac/m1/, so wondering if we should use the same terminology for now and then change it to Apple silicon when we have a newer version?

stephen-turner:
Let's not change it for now because I'm not yet sure whether we will still need the page after the dev preview is over. At the moment it's mostly about restrictions in the dev preview. Maybe there will be more to say later, but maybe there won't.

stephen-turner (approved):
PR LGTM now

PR #12293, opened 2021-02-10
Fix typo in Dockerfile Best Practices ENV section

henare:
### Proposed changes

The example below this change shows that the ENV value can be dumped so I think this was a typo saying that the ENV value can't be dumped.

N/A

### Related issues (optional)

N/A


usha-mandya (approved):
Thanks.

PR #12302, opened 2021-02-10
Fixes for language-specific (Python) tutorial

mboonstra:
Proposed changes
Quick grammar fix: run -> runs

Set the mysql server to use the root password as used in the Flask application

Switched from latest mysql image to mysql:5.7 image, which accepts remote root logins (the current 8.0 version did not accept root logins from anywhere but localhost for me). The same image is used in the "Get Started"-tutorial.

Changed checking whether the MySQL server is running properly from using docker run to docker exec since you want to log in to the existing mysql container. Added the password as set by the previous docker run command.

Changed @app.route in App.py to initdb, the path used in later parts of the tutorial.

Changed mysql image in docker-compose from mysql to mysql:5.7 for reason listed earlier.

### DIFF
--- a/language/python/develop.md
+++ b/language/python/develop.md
@@ -35,26 +35,25 @@ Now we’ll create a network that our application and database will use to talk 
 $ docker network create mysqlnet
 ```
 
-Now we can run MySQL in a container and attach to the volumes and network we created above. Docker pulls the image from Hub and run it for you locally.
+Now we can run MySQL in a container and attach to the volumes and network we created above. Docker pulls the image from Hub and runs it for you locally.
 
 ```shell
 $ docker run -it --rm -d -v mysql:/var/lib/mysql \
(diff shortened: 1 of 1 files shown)


usha-mandya:
@StefanScherer Could you PTAL?

language/python/develop.md
```diff
@@ -203,7 +202,7 @@ services:
   - ./:/app
 
  mysqldb:
-  image: mysql
+  image: mysql:5.7
```
StefanScherer, line 205: Looks like we need to fix line 149:

```
$ pip3 freeze > requirements.txt
```

instead of `-r`

StefanScherer:
Thanks @mboonstra for your contribution. I'll test your changes again to see if it's ready to merge.

StefanScherer:
Thanks @mboonstra I'm carrying your contribution in #12559 and to some other changes to make the tutorial work.

PR #12359, opened 2021-02-19
Fixed typo

mlr07:
Changed database reference from MongoDB to MySQL


romil-soni (approved):

StefanScherer (approved):
Thank you @mlr07 for your contribution!
Oh yes, it's MySQL instead of MongoDB.
We have some other PR's that also fix this, but have more changes and they're not ready yet to merge, so we can go with your change first.
LGTM

Issue #12372, opened 2021-02-23
Add links to Docker Engine SDK and the HTTP API


adriancuadrado:
[develop/index.md](https://docs.docker.com/develop/)
In [this section](https://docs.docker.com/develop/#advanced-development-with-the-sdk-or-api) the document mentions that you can use the Docker Engine SDK and the HTTP API. The [prerequisites](https://docs.docker.com/develop/#prerequisites), [Develop new apps on Docker](https://docs.docker.com/develop/#develop-new-apps-on-docker) and [learn about language-specific development](https://docs.docker.com/develop/#learn-about-language-specific-app-development-with-docker) sections all have links to whatever they mention, I think it would make sense to do the same with the last section with Docker Engine SDK and the HTTP API.

PR #12374, opened 2021-02-23
Add links to Docker Engine SDK and the HTTP API

t3hami:
### Proposed changes

Add links to Docker Engine SDK and the HTTP API

### Related issues (optional)

Resolves: #12372

### DIFF
--- a/develop/index.md
+++ b/develop/index.md
@@ -32,7 +32,8 @@ most benefits from Docker.
 
 ## Advanced development with the SDK or API
 
-After you can write Dockerfiles or Compose files and use Docker CLI, take it to the next level by using Docker Engine SDK for Go/Python or use the HTTP API directly.
-
-- [Docker Engine SDK](../engine/api/sdk/){: target="_blank" rel="noopener" class="_"}
-- [HTTP API](../engine/api/latest/){: target="_blank" rel="noopener" class="_"}
+After you can write Dockerfiles or Compose files and use Docker CLI, take it to
+the next level by using Docker Engine SDK for Go/Python or use the HTTP API
+directly. Visit the [Develop with Docker Engine API](../engine/api/index.md)
+section to learn more about developing with the Engine API, where to find SDKs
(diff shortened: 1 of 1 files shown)


develop/index.md
```diff
@@ -33,3 +33,6 @@ most benefits from Docker.
 ## Advanced development with the SDK or API
 
 After you can write Dockerfiles or Compose files and use Docker CLI, take it to the next level by using Docker Engine SDK for Go/Python or use the HTTP API directly.
+
+- [Docker Engine SDK](../engine/api/sdk/){: target="_blank" rel="noopener" class="_"}
+- [HTTP API](../engine/api/latest/){: target="_blank" rel="noopener" class="_"}
```
thaJeztah, line 38: Thanks! Thinking we can;

- remove the "target=blank", because these links point to other locations in the documentation, and I don't think it's needed to open those in a new tab / window
- perhaps instead we should link to https://docs.docker.com/engine/api/, which (in the left-hand navigation) has links to both the SDK, the API reference, as well as providing a introduction to the API and links to SDK examples

Given that it would then be a single link, perhaps remove the bullet-list, and make it part of the paragraph itself;

```suggestion
After you can write Dockerfiles or Compose files and use Docker CLI, take it to
the next level by using Docker Engine SDK for Go/Python or use the HTTP API
directly. Visit the [Develop with Docker Engine API](../engine/api/index.md)
section to learn more about developing with the Engine API, where to find SDKs
for your programming language of choice, and to see some examples.
```

thaJeztah (changes requested):

thaJeztah:
@usha-mandya ptal (let me know if my suggestion looks good to you)

develop/index.md
usha-mandya: Yes, makes more sense to link to the API section so users can see links to API and SDK docs.

usha-mandya (approved):

thaJeztah (approved):
LGTM

PR #12406, opened 2021-02-26

pneumoman:
The docker engine seems to get confused if you don't  quote device=1,2 as shown  I.E. '"device=1,2"'


StefanScherer (approved):
Makes sense, thank you!
LGTM

thaJeztah (approved):
LGTM

this is indeed because it's using a CSV format, so the comma gets seen as separator between options. Perhaps it's possible to improve that, and add some custom handling to make it do the more "expected" thing when parsing

PR #12413, opened 2021-03-01
Replace deprecated option --fluent-async-connect with --fluent-async

dontub:
The option --fluent-async-connect is deprecated. --fluent-async shall be used instead.
https://github.com/moby/moby/blob/a602b052a9c285e9659e9ce007f2aa7f0a73812f/daemon/logger/fluentd/fluentd.go#L53


usha-mandya (approved):
@dontub Thanks for fixing this. LGTM.

PR #12414, opened 2021-03-01
Missing hint for possible CIDR notation in noProxy and misleading wildcard

araisch:
### Proposed changes

Misleading to talk about Wildcards after mention "IP and host", while support them only for hosts without any hint to CIDR..


usha-mandya (approved):
LGTM. Thanks @araisch

PR #12418, opened 2021-03-01
ci: Use ubuntu-18.04

StefanScherer:
Signed-off-by: Stefan Scherer <scherer_stefan@icloud.com>

Use `ubuntu-18.04` instead of `ubuntu-latest` to see if we can fix the s3 upload with it.

### DIFF
--- a/.github/workflows/build-master.yml
+++ b/.github/workflows/build-master.yml
@@ -8,7 +8,7 @@ on:
 jobs:
   build-static-page:
     name: build
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-18.04
     timeout-minutes: 10
     env:
       DOCKER_BUILDKIT: '1'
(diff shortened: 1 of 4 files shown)


.github/workflows/build-master.yml
```diff
@@ -8,7 +8,7 @@ on:
 jobs:
   build-static-page:
     name: build
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-18.04
```
thaJeztah, line 11: `ubuntu-latest` is 20.04? or is it 20.10?

What was the failure we saw?

thaJeztah (approved):
SGTM

.github/workflows/build-master.yml
usha-mandya: It was `Error: Process completed with exit code 255.` There was also a warning `Ubuntu-latest workflows will use Ubuntu-20.04 soon. For more details, see https://github.com/actions/virtual-environments/issues/1816`

PR #12437, opened 2021-03-04
Add M1 tech preview version number

usha-mandya:

Add the Docker Desktop tech preview version number


StefanScherer (approved):
Correct, the version number without the build number is enough.
Next preview won't be a "3.1.0".
LGTM

stephen-turner:
Thanks. It's good enough for now, as you've already merged it, but in future I'd like to use the build number because I can't guarantee we won't have two 3.2.0 previews, for example.

PR #12447, opened 2021-03-05
Add `lsb-release` to required debian packages.

akester:
### Proposed changes

Add `lsb-release` to list of required debian packages.

Certain installations may not have this.  Without this, the step where `/etc/apt/sources.list.d/docker.list` is created will fail and write garbled data to the file, potentially breaking apt.

NA

### Related issues (optional)

NA

### DIFF
--- a/engine/install/debian.md
+++ b/engine/install/debian.md
@@ -84,7 +84,8 @@ from the repository.
         apt-transport-https \
         ca-certificates \
         curl \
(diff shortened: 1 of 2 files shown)


thaJeztah:
Actually; let me push that to your branch so that we can get this merged and published

thaJeztah (approved):
LGTM

PR #12458, opened 2021-03-08
Django: Use docker ps to be in sync with the sample output

### LINKED ISSUE #12268: docs: Wrong path to manage.py and inconsistent docker running process listing command shown
[compose/django.md](https://docs.docker.com/compose/django/)

In the `docker-compose.yml` file, the path to `manage.py` should be `composeexample/manage.py` and not `manage.py`.

Also (it is such a trivial problem I was debating if i should be opening a separate ticket for this).

At the end of the section, it asks that you should use `docker container ls` to list running containers. But the example shows `docker ps`. I understand that it is equivalent, but just wanted to point out that from the point of view of documentation, this is inconsistent.

I can create a pull request if someone confirms that these fixes are needed.

StefanScherer:

Use `docker ps` command to be in sync with the sample output below.

Fixes #12268

### DIFF
--- a/compose/django.md
+++ b/compose/django.md
@@ -228,7 +228,7 @@ In this section, you set up the database connection for Django.
 
 5. List running containers.
 
-   In another terminal window, list the running Docker processes with the `docker container ls` command.
+   In another terminal window, list the running Docker processes with the `docker ps` or `docker container ls` command.
 
    ```console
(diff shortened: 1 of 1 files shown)


usha-mandya (approved):
LGTM

PR #12461, opened 2021-03-08
Fix docker-compose database volume configuration

irlcatgirl:
### Proposed changes

I modified the docker-compose.yml configuration related to the database server's volumes. Without this change any database data is immediately discarded. This is inconsistent with the documentation for Ruby on Rails with Docker Compose where it's retained (See https://docs.docker.com/compose/rails/) when restarting the application within compose.

N/A

### Related issues (optional)
N/A


usha-mandya (approved):
LGTM. Thanks @irlcatgirl :)

Issue #9745, opened 2019-10-28
noProxy and use of wildcards


egarbi:
[network/proxy.md](https://docs.docker.com/network/proxy/)

From reading the documentation for proxy configuration is not clear to me the difference between for instance "*.example.com" and ".example.com".
When exactly wildcards have to be used for noProxy settings?
Actually, for my own use case * (wildcard) notation doesn't seem to work.

PR #12463, opened 2021-03-09
Fix proxy documentation- * as wildcard is not supported

bport:
### Proposed changes

Updated proxy documentation
According to documentation we can use * as a wild card for noProxy, this is not working. 
And according to https://www.gnu.org/software/wget/manual/html_node/Proxies.html we need to prefix with a dot.

### Related issue
https://github.com/docker/docker.github.io/issues/9745

### DIFF
--- a/network/proxy.md
+++ b/network/proxy.md
@@ -24,8 +24,8 @@ configure it in different ways:
 
     You can optionally exclude hosts or ranges from going through the proxy
     server by setting a `noProxy` key to one or more comma-separated IP
-    addresses or hosts. Using the `*` character as a wildcard for hosts and using CIDR notation for IP addresses is supported as
-    shown in this example:
+    addresses or hosts. Prefixing domain with a dot `.` character, will act as a wildcard for hosts and using CIDR notation for IP addresses is supported as
+    shown in this example.
 
     ```json
     {
@@ -35,7 +35,7 @@ configure it in different ways:
        {
          "httpProxy": "http://192.168.1.12:3128",
          "httpsProxy": "http://192.168.1.12:3128",
-         "noProxy": "*.test.example.com,.example2.com,127.0.0.0/8"
+         "noProxy": ".test.example.com,.example2.com,127.0.0.0/8"
        }
      }
     }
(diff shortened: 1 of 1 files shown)


network/proxy.md
```diff
@@ -26,7 +26,7 @@ configure it in different ways:
 
     You can optionally exclude hosts or ranges from going through the proxy
     server by setting a `noProxy` key to one or more comma-separated IP
-    addresses or hosts. Using the `*` character as a wildcard for hosts and using CIDR notation for IP addresses is supported as
+    addresses or hosts. Prefixing domain with a dot `.` character, will act as a wildcard for hosts and using CIDR notation for IP addresses is supported as
```
usha-mandya, line 29:
```suggestion
    addresses or hosts. Prefixing domain with a dot `.` acts as a wildcard for hosts and using CIDR notation for IP addresses is supported as
```

usha-mandya (approved):

network/proxy.md
```diff
@@ -26,7 +26,7 @@ configure it in different ways:
 
     You can optionally exclude hosts or ranges from going through the proxy
     server by setting a `noProxy` key to one or more comma-separated IP
-    addresses or hosts. Using the `*` character as a wildcard for hosts and using CIDR notation for IP addresses is supported as
+    addresses or hosts. Prefixing domain with a dot `.` acts as a wildcard for hosts and using CIDR notation for IP addresses is supported as
```
thaJeztah, line 29: Hm.. yes, these are "complicated", as the supported formats are not formalised. Perhaps we should mention something about that (supported formats depend on the software that's using the `NO_PROXY` (and other `_PROXY`/ `_proxy` vars), and we could link to the Go documentation, and possibly copy some of that; https://pkg.go.dev/golang.org/x/net/http/httpproxy#Config

```
NoProxy represents the NO_PROXY or no_proxy environment
variable. It specifies a string that contains comma-separated values
specifying hosts that should be excluded from proxying. Each value is
represented by an IP address prefix (1.2.3.4), an IP address prefix in
CIDR notation (1.2.3.4/8), a domain name, or a special DNS label (*).
An IP address prefix and domain name can also include a literal port
number (1.2.3.4:80).
A domain name matches that name and all subdomains. A domain name with
a leading "." matches subdomains only. For example "foo.com" matches
"foo.com" and "bar.foo.com"; ".y.com" matches "x.y.com" but not "y.com".
A single asterisk (*) indicates that no proxying should be done.
A best effort is made to parse the string and errors are
ignored.
```

I seem to recall we did that somewhere (but it could be I started working on it, but never pushed a PR 🤔)

thaJeztah:
Looks like this needs a rebase; otherwise this looks good to me (we can look at the improvements in a follow-up)

bport:
> Looks like this needs a rebase; otherwise this looks good to me (we can look at the improvements in a follow-up)

rebase done

bport:
@thaJeztah can we already merge this version ?

PR #12464, opened 2021-03-09
Fix uninstall command for Debian

gtklocker:
### Proposed changes

apt-get remove with multiple packages will stop on the first package that is not installed

### DIFF
--- a/engine/install/debian.md
+++ b/engine/install/debian.md
@@ -33,7 +33,7 @@ Older versions of Docker were called `docker`, `docker.io`, or `docker-engine`.
 If these are installed, uninstall them:
 
 ```console
-$ sudo apt-get remove docker docker-engine docker.io containerd runc
+$ for pkg in docker docker-engine docker.io containerd runc; do sudo apt-get remove -y $pkg; done
 ```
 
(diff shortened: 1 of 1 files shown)


engine/install/debian.md
```diff
@@ -33,7 +33,7 @@ Older versions of Docker were called `docker`, `docker.io`, or `docker-engine`.
 If these are installed, uninstall them:
 
 ```console
-$ sudo apt-get remove docker docker-engine docker.io containerd runc
+$ for pkg in docker docker-engine docker.io containerd runc; do sudo apt-get remove -y $pkg; done
```
thaJeztah, line 36: I'm a bit concerned that the command becomes longer (and potentially more error-prone for users to follow); wondering if instead we should use `dpkg --remove` (as suggested in https://superuser.com/a/518871)

```suggestion
$ sudo dpkg --remove docker docker-engine docker.io containerd runc
```

thaJeztah: If we update, we should also update the ubuntu page with the same changes

thaJeztah (changes requested):

usha-mandya:
Closing this PR as there isn't any activity on the PR.

PR #12466, opened 2021-03-09
compose cli reference documentation

ndeloof:
### Proposed changes

include reference documentation for the `docker compose` subcommand introduced by docker/compose-cli
yamldocs files are generated using https://github.com/docker/compose-cli/pull/1407

close https://github.com/docker/compose-cli/issues/1404

### DIFF
--- a/_data/compose-cli/docker_compose_create.yaml
+++ b/_data/compose-cli/docker_compose_create.yaml
@@ -1,0 +1,51 @@
+command: docker compose create
+short: Creates containers for a service.
+long: Creates containers for a service.
+usage: docker compose create [SERVICE...]
+pname: docker compose
+plink: docker_compose.yaml
+options:
+  - option: build
+    value_type: bool
+    default_value: "false"
+    description: Build images before starting containers.
+    deprecated: false
+    experimental: false
+    experimentalcli: false
+    kubernetes: false
+    swarm: false
+  - option: force-recreate
+    value_type: bool
+    default_value: "false"
+    description: |
+        Recreate containers even if their configuration and image haven't changed.
+    deprecated: false
+    experimental: false
+    experimentalcli: false
+    kubernetes: false
+    swarm: false
+  - option: no-build
+    value_type: bool
+    default_value: "false"
+    description: Don't build an image, even if it's missing.
+    deprecated: false
+    experimental: false
+    experimentalcli: false
+    kubernetes: false
+    swarm: false
+  - option: no-recreate
+    value_type: bool
+    default_value: "false"
+    description: |
+        If containers already exist, don't recreate them. Incompatible with --force-recreate.
+    deprecated: false
+    experimental: false
+    experimentalcli: false
+    kubernetes: false
+    swarm: false
+deprecated: false
+experimental: false
+experimentalcli: false
+kubernetes: false
+swarm: false
+
--- a/_data/compose-cli/docker_compose_up.yaml
+++ b/_data/compose-cli/docker_compose_up.yaml
@@ -1,0 +1,187 @@
+command: docker compose up
+short: Create and start containers
+long: Create and start containers
+usage: docker compose up [SERVICE...]
+pname: docker compose
+plink: docker_compose.yaml
+options:
+  - option: abort-on-container-exit
+    value_type: bool
+    default_value: "false"
+    description: |
+        Stops all containers if any container was stopped. Incompatible with -d
+    deprecated: false
+    experimental: false
+    experimentalcli: false
+    kubernetes: false
+    swarm: false
+  - option: always-recreate-deps
+    value_type: bool
+    default_value: "false"
+    description: |
+        Recreate dependent containers. Incompatible with --no-recreate.
+    deprecated: false
+    experimental: false
+    experimentalcli: false
+    kubernetes: false
+    swarm: false
+  - option: attach-dependencies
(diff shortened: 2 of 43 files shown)


_data/compose-cli/docker_compose_create.yaml
```diff
@@ -0,0 +1,51 @@
...
+    deprecated: false
+    experimental: false
+    experimentalcli: false
+    kubernetes: false
+    swarm: false
+  - option: force-recreate
+    value_type: bool
+    default_value: "false"
+    description: |
+        Recreate containers even if their configuration and image haven't changed.
```
usha-mandya, line 21:
```suggestion
        Re-create containers even if their configuration and image haven't changed.
```

_data/compose-cli/docker_compose_create.yaml
```diff
@@ -0,0 +1,51 @@
...
+    deprecated: false
+    experimental: false
+    experimentalcli: false
+    kubernetes: false
+    swarm: false
+  - option: no-recreate
+    value_type: bool
+    default_value: "false"
+    description: |
+        If containers already exist, don't recreate them. Incompatible with --force-recreate.
```
usha-mandya, line 40:
```suggestion
        If containers already exist, don't create them. Incompatible with --force-recreate.
```

_data/compose-cli/docker_compose_up.yaml
```diff
@@ -0,0 +1,187 @@
...
+    deprecated: false
+    experimental: false
+    experimentalcli: false
+    kubernetes: false
+    swarm: false
+  - option: force-recreate
+    value_type: bool
+    default_value: "false"
+    description: |
+        Recreate containers even if their configuration and image haven't changed.
```
usha-mandya, line 79:
```suggestion
        Re-create containers even if their configuration and image haven't changed.
```

_data/compose-cli/docker_compose_up.yaml
```diff
@@ -0,0 +1,187 @@
...
+    deprecated: false
+    experimental: false
+    experimentalcli: false
+    kubernetes: false
+    swarm: false
+  - option: always-recreate-deps
+    value_type: bool
+    default_value: "false"
+    description: |
+        Recreate dependent containers. Incompatible with --no-recreate.
```
usha-mandya, line 22:
```suggestion
        Re-create dependent containers. Incompatible with --no-recreate.
```

ndeloof:
@usha-mandya all commands description are actually defined here https://github.com/docker/compose-cli/blob/main/cli/cmd/compose/create.go#L60 and are just a pure copy/paste for the docker-compose command we re-implemented. 
In the current sprints we focus on compatibility with docker-compose to offer a smooth migration to "compose-in-docker-cli" but we don't plan to "fix" legacy weirdnesses :P

usha-mandya:
Thanks. Understand that it's not a regression as these descriptions exist in the live docs - https://docs.docker.com/compose/reference/create/. It would be really good to fix these descriptions in the upstream repo. Let me know if I should create a ticket to get these fixed.

usha-mandya:
@thaJeztah Do you mind taking a quick look at the reference docs updates? Thanks.

thaJeztah:
Nice!

I built this locally, and this looks good as a starting point. Obviously it's still _really_ bare-bones (no examples, only description of the commands and list of the flags), but that's something we can work on. 

Screenshot 2021-03-09 at 18 36 00
Screenshot 2021-03-09 at 18 38 56
 

Perhaps we should look to add a include/banner on these pages to;

- Mention it's currently available on Docker Desktop, not (yet) by default on Linux
- Links to how to enable/disable
- Create a https://docs.docker.com/cloud/ landing page that we can link to, and on which we can write the things we want about that
- Possibly create [a `/go/` redirect](https://github.com/docker/docker.github.io/tree/master/go) that goes to that page (which we could use for in-product links if needed, or (for now) as a convenience URL if things are still moved around

I can have a look at how to implement the above with as-minimal-as-possible changes (perhaps it's possible to add the include through configuration).

Having the yaml files in a `compose-cli` directory is good for the current situation (not yet integrated in the cli); if integration of the "UX-side" of compose-cli is done, we can move those files together with the `cli` YAML files.

As to; adding examples etc, we could have a look at copying/extracting examples from the docker-compose docs (restructuring the markdown files to use the expected sections). Help me remind on that, and I'll try to reserve some time to look at that.

thaJeztah (approved):
left a comment above, but if we want this to be "live" ASAP, I'd be "ok" with merging as-is. If it's ok to wait for a "bit", I can have a look at adding the banner (but we'd need to think what needs to be on the "landing page" with instructions etc.

so .. "LGTM"

usha-mandya:
Thanks for the details @thaJeztah. 

@ndeloof Could you let us know if we want this to be live asap? If so, suggest we add a banner indicating that 'docker compose' is offered as a Tech Preview.  Something similar to https://docs.docker.com/compose/cli-command/. We can then do a follow up PR to update the files based on the suggestions above by Sebastiaan. Thanks.

ndeloof:
@usha-mandya as `compose` subcommand is already available to all desktop users, better get this documented asap. Can you manage adding the tech preview banner?

usha-mandya:
@ndeloof Sure. I'll create a follow-up PR to add the Tech Preview banner.

PR #12490, opened 2021-03-12
Fix minor typo in example of documentation

mossroy:
### Proposed changes

I suppose the directory where to extract postgresql should be called postgres instead of postgress


usha-mandya (approved):
LGTM. Thank you @mossroy

PR #12500, opened 2021-03-15
Update django.md

anp135:
Not very clear to understand why multiple django docker application utilize CPU in idle state. --noreload comment there will be more friendly.


anp135:
@usha-mandya could you approve PR?

usha-mandya:
@anp135 Thanks for the PR.  Please note that this is a sample document. The suggestion in this PR is not a common use case. I'll have to close the PR without accepting your suggestions for this reason.

PR #12504, opened 2021-03-15
update 07_multi_container.md: add -u root to mysql

vic-torr:
user login didnt work, so it's needed to specify user: root:secret.
docker exec -it <mysql-container-id> mysql -u root -p


usha-mandya (approved):
LGTM. Thanks @vic-torr

PR #12506, opened 2021-03-15
Remove Compose file v1 ref docs

usha-mandya:

- Remove Compose file version 1 docs as this has been deprecated
- Added a redirect to point to Compose spec topic


compose/compose-file/compose-file-v2.md
```diff
@@ -998,7 +998,7 @@ options:
 
 > Changed in [version 2](compose-versioning.md#version-2) file format.
 >
-> The `network_mode` option replaces the version 1 [net](compose-file-v1.md#net) option.
+> The `network_mode` option replaces the Compose file version 1 net option.
```
chris-crone, line 1001: Should we remove this or add `(deprecated)` so that people aren't confused about v1 here?

chris-crone (approved):
One question but otherwise LGTM!

compose/compose-file/compose-file-v2.md
usha-mandya: You are right. This piece of info doesn't add value to users. I think it's best to remove. Let me update the PR.

chris-crone (approved):
LGTM

PR #12538, opened 2021-03-21
doc: clarify the docs

### LINKED ISSUE #12537: Make description more clearly
Hi Docker support team,

I've been learning this chapter: [language/python/run-containers.md](https://docs.docker.com/language/python/run-containers/)

I was quite confused, didn't know that I need to open a new terminal before using the `curl` command.

Because we need to use the `curl` command when the container is running. If we press `Ctrl+C` then use the `curl` command, it's not the expectation, right?

Thanks & Best regards,

Anh Tran

longqua69:
### Proposed changes
I was quite confused while reading the document that I showed in the issue #12537

=> I clarify the document by this PR.

### Related issues (optional)
Closes #12537

### DIFF
--- a/language/python/run-containers.md
+++ b/language/python/run-containers.md
@@ -24,7 +24,7 @@ $ docker run python-docker
 
 After running this command, you’ll notice that you were not returned to the command prompt. This is because our application is a REST server and runs in a loop waiting for incoming requests without returning control back to the OS until we stop the container.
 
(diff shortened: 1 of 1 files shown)


StefanScherer (approved):
Thank you @longqua69 to clarify that a second terminal should be used here.

LGTM. 💯

PR #12552, opened 2021-03-23
Apple Silicon: inotify APIs don't work under qemu

djs55:
### Proposed changes

Added a note about `inotify` APIs not working on Apple Silicon when running Intel images (through `qemu` emulation)


usha-mandya (approved):
LGTM. Thanks @djs55

PR #12559, opened 2021-03-24
Fix Python tutorial

StefanScherer:
### Proposed changes

Fix the Python language guide https://docs.docker.com/language/python/
Use port 5000 that is the default port of the flask app.

Includes #12302


usha-mandya (approved):
LGTM. Thank you @StefanScherer :)

PR #12570, opened 2021-03-25
Remove a link from the footer

usha-mandya:

Removed a link to Docker Security which was pointing to an outdated page. Need to restore this when we have an update page.


StefanScherer (approved):
LGTM

PR #12591, opened 2021-03-29
Add a note about `ping` when using `qemu`

djs55:
### Proposed changes

Add a note about `ping` not working correctly when using `qemu` on Apple Silicon.


usha-mandya (approved):
LGTM

PR #12614, opened 2021-04-04
Adding sudo command

osagiestar:
When running the command to pull the development docker image that FSCI provides, you can encounter access denial as a non-root user, you can add sudo as prefix (before) the command: docker pull registry.gitlab.com/fsci/resources:debian-dev


osagiestar:
Spoken to Akshay as not meant to be on here

PR #12618, opened 2021-04-05
CSS styles amended for proper printing

SetTrend:
### Proposed changes

I amended CSS styles for printing, because the text currently is illegible when being printed from dark mode:

Current-Desired

### DIFF
--- a/_scss/_night-mode.scss
+++ b/_scss/_night-mode.scss
@@ -95,10 +95,21 @@ body.night {
         color: #f3f3f3;
         background-color: #0a121b !important;
         border: 1px solid $black;
+
+        @media print {
+            color: #444;
+            background-color: #f5f8fa !important;
+            border-width: 0;
+        }
     }
     code {
         color: #c1f1f0 !important;
         background: #0a121b !important;
+
+        @media print {
+            color: darkcyan !important;
+            background: #f5f8fa !important;
+        }
     }
     .nav-secondary .dropdown-btn {
         color: #d0e8ff;
@@ -184,7 +195,7 @@ body.night {
         color: #e8a0e8;
       }
       .k {
-        color: #da5ada;
+        color: #da5ada !important;
       }
     }
 
--- a/_scss/_perldoc.scss
+++ b/_scss/_perldoc.scss
@@ -86,3 +86,8 @@
 .highlight .vi { color: #00688B } /* Name.Variable.Instance */
 .highlight .vm { color: #00688B } /* Name.Variable.Magic */
 .highlight .il { color: #B452CD } /* Literal.Number.Integer.Long */
+
+@media print {
+.highlight .k { color: #8B008B !important } /* Keyword */
+.highlight .s { color: #CD5555 !important } /* Literal.String */
+}
--- a/_scss/_utilities.scss
+++ b/_scss/_utilities.scss
@@ -189,6 +189,8 @@ input[type=text] {
 }
 
 @media print {
+    #announcement-bar,
+    .reading-time,
     .col-nav,
     .col-toc,
     .footer,
--- a/_scss/_content.scss
+++ b/_scss/_content.scss
@@ -20,7 +20,7 @@
  * Code formatting *************************************************************
  */
 pre, code {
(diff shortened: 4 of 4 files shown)


_scss/_utilities.scss
```diff
@@ -189,6 +189,8 @@ input[type=text] {
 }
 
 @media print {
+    #announcement-bar,
```
thaJeztah, line 192: The announcement bar is added by our marketing team through Google Tag Manager; I don't think we should add this here, because we don't know in advance what ID they might be using (now, and in future); but I can ask the marketing team if they could add a class or something else to banners to prevent them being printed.

```suggestion
```

_scss/_perldoc.scss
```diff
@@ -86,3 +86,8 @@
 .highlight .vi { color: #00688B } /* Name.Variable.Instance */
 .highlight .vm { color: #00688B } /* Name.Variable.Magic */
 .highlight .il { color: #B452CD } /* Literal.Number.Integer.Long */
+
+@media print {
+.highlight .k { color: #8B008B !important } /* Keyword */
+.highlight .s { color: #CD5555 !important } /* Literal.String */
+}
```
thaJeztah, line 93: Looks like this is missing a newline at the end

```suggestion
}

```

thaJeztah: Do you recall why the `!important` was needed? (trying to avoid those if possible)

Also wondering if it would be better to move this to `_night-mode.scss`

_scss/_night-mode.scss
```diff
@@ -91,10 +91,21 @@ body.night {
         color: #f3f3f3;
         background-color: #0a121b !important;
         border: 1px solid $black;
+
+        @media print {
+            color: #444;
+            background-color: #f5f8fa !important;
+            border-width: 0;
+        }
```
thaJeztah, line 99: Instead of overriding these, perhaps it would be possible to make the "night-mode" style conditional (i.e., don't apply dark mode in "print")

thaJeztah (changes requested):
left some initial comments, but may want to revisit to check some

_scss/_utilities.scss
SetTrend: Yes, that would be great. (Please pardon my late reply.)

I think it'd be very reasonable to have it hidden when converting the documentation pages to PDF for printing. Having such announcement on every other printed page currently is, indeed, wasting paper and reading time.

_scss/_perldoc.scss
SetTrend: > Looks like this is missing a newline at the end

I now added the suggested newline character to the changes I'm going to commit with the next update.

> Do you recall why the !important was needed? (trying to avoid those if possible)

Sure, in the original `_night-mode.scss`, someone used `!important` quite abundant:

```scss
    blockquote {
        ...
        &.warning, &.warning p:first-child {
            color: $warning-color-night !important;
        }
        &.important {
            border-left-color: #f5ac45 !important;
        }
        &.important, &.important p:first-child {
            color: $important-color-night !important;
        }
    }
    ...
    pre {
        color: #f3f3f3;
        background-color: #0a121b !important;
        border: 1px solid $black;
    }
```

So, in order to be able to override these settings, I was forced to apply the `!important` modifier, too. I didn't want to take the effort to refactor the existing SCSS files.

> Also wondering if it would be better to move this to `_night-mode.scss`

That's a matter of style (no pun intended 😉). I chose the `_perldoc.scss` location so someone in the future may be taking the task to update the corresponding syntax highlight colors. He/she would then recognize the dark mode versions and alter them accordingly.

If you want me to move them to `_night-mode.scss` that'd be perfectly alright with me. I'm looking forward to your valuable suggestion.

_scss/_night-mode.scss
SetTrend: I don't think this will be easy to achieve. ... The original author didn't use a media query to select styles for dark mode. He/she used a class (`body.night`) and applied it conditionally to the `<body>` element using JavaScript. So, I cannot alter a media query to omit the `print` media.

PR #12645, opened 2021-04-09
Switch to a better URL

stephen-turner:
### Proposed changes

Switch to a more permanent URL (we already switched in the WSL 2 page, but missed this reference).


usha-mandya (approved):
LGTM

PR #12647, opened 2021-04-09
Reduce duplication by referencing existing rootless Docker docs

benmccann:
### Proposed changes

There are instructions about rootless install on https://docs.docker.com/engine/install/ubuntu/ that are duplicated between that page and https://docs.docker.com/engine/install/linux-postinstall/. This change reduces duplication and creates one canonical location for info about rootless installs.

As a user trying to install Docker in a rootless fashion just for regular use (i.e. not for contributing to Docker itself) I was very confused as to whether this was possible or how to do it. I was on https://docs.docker.com/engine/install/ubuntu/ and did a Ctrl+F for "root" and the only info was under the convenience script section, which wasn't really relevant to me as a user installing from the repository. It turns out this info lived on the "Post-installation steps for Linux" page. If that page had been linked to under the convenience script section I would have discovered the info more more easily and realized it was applicable to all Linux users and not just convenience script users. Another possible solution might be to move it out from the convenience script section and put it someplace relevance to any installation method. However, that seemed like an additional possibly slightly larger change, so I did it this way to keep the change a bit smaller.

### DIFF
--- a/_includes/install-script.md
+++ b/_includes/install-script.md
@@ -43,32 +43,13 @@ $ sudo sh get-docker.sh
 <...>
 ```
 
-If you would like to use Docker as a non-root user, you should now consider
-adding your user to the "docker" group with something like:
-
-```console
-$ sudo usermod -aG docker <your-user>
-```
-
-Remember to log out and back in for this to take effect!
-
-> **Warning**:
->
-> Adding a user to the "docker" group grants them the ability to run containers
-> which can be used to obtain root privileges on the Docker host. Refer to
-> [Docker Daemon Attack Surface](/engine/security/#docker-daemon-attack-surface)
-> for more information.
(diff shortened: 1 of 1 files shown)


AkihiroSuda (approved):

usha-mandya:
@thaJeztah Could you PTAL? Wondering if it makes sense to retain the URL to rootless mode docs, in addition to the existing update.

thaJeztah:
@usha-mandya yes, perhaps we need to clean this up a bit; I gave it a quick try, and pushed a follow-up as draft in https://github.com/docker/docker.github.io/pull/12740

thaJeztah (approved):
let me get this one merged, then I'll move my follow-up out of draft

LGTM, thanks!

PR #12652, opened 2021-04-11
Update 02_our_app.md

zzzzzzzzzx:


StefanScherer (changes requested):
Hello @zzzzzzzzzx,
Can you explain why this changes is needed?
The node base image's current directory is `/`, so there is no parent directory.
It's the first time I see `WORKDIR` with `..`, that's quite unusual.
I think we can close this PR, but would like to hear your thoughts.

Thanks.

usha-mandya:
Closing this due to inactivity. Feel free to continue the conversation after I close.

PR #12682, opened 2021-04-15
fixed linux service commands

europ:
Service restarting commands did not include the command itself.

PATH: `<hostname>/engine/install/linux-postinstall`


usha-mandya (approved):
Thank you @europ. LGTM

PR #12685, opened 2021-04-15
Add `compose images` command

aiordache:
Add `compose images` reference for both `docker compose` and `docker-compose`.


usha-mandya (approved):
LGTM

PR #12703, opened 2021-04-19
use Dockerfile "syntax" directive in examples

thaJeztah:
BuildKit allows using alternative Dockerfile syntaxes to introduce new features
without having to update Docker itself. The general recommendation is to always
specify a "syntax" directive in a Dockerfile, so that (if needed) older versions
of Docker can download the correct syntax to build the Dockerfile.

This updates our examples to include a syntax directive, to make users more familiar
with these directives, and to illustrate best-pracitces in our documentation.

Example: https://deploy-preview-12703--docsdocker.netlify.app/language/nodejs/build-images/#create-a-dockerfile-for-nodejs

Screenshot 2021-04-19 at 12 23 58
Screenshot 2021-04-19 at 12 24 38

### DIFF
--- a/develop/develop-images/build_enhancements.md
+++ b/develop/develop-images/build_enhancements.md
@@ -19,7 +19,8 @@ on performance, storage management, feature functionality, and security.
   information for building new images with a specified Dockerfile 
 
 For more information on build options, see the reference guide on the
-[command line build options](/engine/reference/commandline/build/).
+[command line build options](../../engine/reference/commandline/build.md) and
+the [Dockerfile reference](/engine/reference/builder/) page.
 
 
 ## Requirements
@@ -115,8 +116,15 @@ frontend. To override the default frontend, set the first line of the
 `Dockerfile` as a comment with a specific frontend image: 
 
 ```dockerfile
-# syntax = <frontend image>, e.g. # syntax = docker/dockerfile:1.0-experimental
+# syntax=<frontend image>, e.g. # syntax=docker/dockerfile:1
 ```
+
+The examples on this page use features that are available in `docker/dockerfile`
+version 1.2.0 and up. We recommend using `docker/dockerfile:1`, which always
+points to the latest release of the version 1 syntax. BuildKit automatically
+checks for updates of the syntax before building, making sure you are using the
+most current version. Learn more about the `syntax` directive in the
+[Dockerfile reference](/engine/reference/builder/#syntax).
 
 ## New Docker Build secret information
 
@@ -138,13 +146,11 @@ For example, with a secret piece of information stored in a text file:
 $ echo 'WARMACHINEROX' > mysecret.txt
 ```
 
-And with a Dockerfile that specifies use of a BuildKit frontend
-`docker/dockerfile:1.0-experimental`, the secret can be accessed. 
-
-For example:
+Within a Dockerfile that uses BuildKit frontend `docker/dockerfile:1.2` or up,
+the secret can be accessed using the `--mount` option:
 
 ```dockerfile
-# syntax = docker/dockerfile:1.0-experimental
+# syntax=docker/dockerfile:1
 FROM alpine
 
 # shows secret from default secret location:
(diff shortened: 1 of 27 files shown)


thaJeztah:
@chris-crone @usha-mandya ptal

/cc @tonistiigi

thaJeztah:
Note to self: I saw some other outdated information in our docs that (likely because the old / classic builder could do this) explains that each instruction in the Dockerfile creates a new layer; search for:

- "Each instruction creates one layer"
- "each of which creates a layer"

With recent versions of Docker (and with BuildKit), this is no longer true; many instructions, such as `ENV`, `LABEL`, do not create a new layer, and new layers are only created for Dockerfile instructions that modify the filesystem (`RUN` (potentially), `COPY`, `ADD`)

_config.yml
```diff
@@ -27,6 +27,8 @@ compose_file_v3: "3.9"
 compose_file_v2: "2.4"
 machine_version: "0.16.0"
 distribution_version: "2.7"
+# latest stable major.minor for dockerfile syntax; will be shown as "syntax=docker/dockerfile:1.2"
```
chris-crone, line 30: We could use `1` here so that it attaches to the latest v1 stable

tonistiigi (commented):
Also https://docs.docker.com/engine/reference/builder/#syntax needs update

develop/develop-images/build_enhancements.md
```diff
@@ -115,9 +116,16 @@ frontend. To override the default frontend, set the first line of the
...
-# syntax = <frontend image>, e.g. # syntax = docker/dockerfile:1.0-experimental
+# syntax=<frontend image>, e.g. # syntax=docker/dockerfile:1
 ```
 
+The examples on this page use features that are available in `docker/dockerfile`
+version 1.2.0 and up. We recommend using `docker/dockerfile:1`, which always
+points to the latest release of the version 1 syntax. BuildKit automatically
+checks for updates of the syntax before building, making sure you are using the
+most current version. Learn more about the `syntax` directive in the
+[Dockerfile reference](/engine/reference/builder/#syntax).
```
thaJeztah, line 127: Added this snippet

get-started/nodejs/build-images.md
```diff
@@ -87,9 +87,27 @@ Let’s walk through the process of creating a Dockerfile for our application. I
...
+must appear before any other comment, whitespace, or Dockerfile instruction in
+your Dockerfile, should be the first line in Dockerfiles.
 
 ```dockerfile
+# syntax=docker/dockerfile:1
+```
+
+We recommend using `docker/dockerfile:1`, which always points to the latest release
+of the version 1 syntax. BuildKit automatically checks for updates of the syntax
+before building, making sure you are using the most current version.
```
thaJeztah, line 103: Added the same recommendation here

thaJeztah: (We could probably remove this section to an include file for easier re-use))

engine/security/trust/trust_automation.md
```diff
@@ -75,6 +75,7 @@ you should set the environment variable `DOCKER_CONTENT_TRUST` either manually o
 in a scripted fashion. Consider the simple Dockerfile below.
 
 ```dockerfile
+# syntax=docker/dockerfile:1
```
thaJeztah, line 78: I removed the parameter from the `_config.yml`, and now have `dockerfile:1` hard-coded; I don't expect we'll reach v2 anytime soon, and if we do, a `find & replace` should be easy, so no need to use a parameter for it

thaJeztah (commented):
@chris-crone @tonistiigi updated (see my comments) PTAL

tonistiigi (approved):

get-started/nodejs/build-images.md
usha-mandya: Adding this to an includes file would be good. We can do this and a couple of other updates in a follow-up PR.

usha-mandya (approved):
LGTM

PR #12724, opened 2021-04-21
update old "ee" redirects to go to our homepage

thaJeztah:
Looks like all links on the Mirantis docs website are broken and now showing
a 404, so those redirects are no longer useful.

This changes the redirects to go to our homepage instead; we can make
additional tweaks and/or create more specific redirects in a follow-up.

### DIFF
--- a/_website-config-docs-stage.json
+++ b/_website-config-docs-stage.json
@@ -182,11 +182,11 @@
         "KeyPrefixEquals": "ee/licensing/"
       },
       "Redirect": {
-        "HostName": "docs.mirantis.com",
+        "HostName": "docs-stage.docker.com",
         "HttpRedirectCode": null,
         "Protocol": "https",
         "ReplaceKeyPrefixWith": null,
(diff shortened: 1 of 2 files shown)


thaJeztah:
@StefanScherer @usha-mandya PTAL

I updated _both_ the "staging" and "docs.docker.com" versions, but we should test this on staging before it's published.

I _hope_ this is the correct way to do this (replace "key" (URL) with "empty string" (homepage)), but definitely could use some eyes (and testing)

StefanScherer (approved):
LGTM !

StefanScherer:
we can merge it to master and then test in docs-stage.docker.com

thaJeztah:
Sounds good to me! @usha-mandya do we have other (urgent) PR's pending that need to be published? Otherwise I think we can merge and try

usha-mandya:
No other urgent PRs at the moment. We can merge and test this in staging.

PR #12728, opened 2021-04-21
Publish updates from master

usha-mandya:

### DIFF
--- a/_website-config-docs-stage.json
+++ b/_website-config-docs-stage.json
@@ -182,11 +182,11 @@
         "KeyPrefixEquals": "ee/licensing/"
       },
       "Redirect": {
-        "HostName": "docs.mirantis.com",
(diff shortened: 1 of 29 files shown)


StefanScherer (approved):
LGTM

StefanScherer:
Small nit: ReplaceKeyPrefixWith is now used with both `null` and `""`, but the outcome is the same. So, it's fine.

thaJeztah:
> Small nit: ReplaceKeyPrefixWith is now used with both null and "", but the outcome is the same. So, it's fine.

Thanks! I wasn't sure if it would treat `null` as "don't do anything" and `""` as "replace", so I verged on the "safe side"; we can update in a follow up (perhaps even _remove_ the fields that are empty?)

thaJeztah (approved):
LGTM

PR #12743, opened 2021-04-24
Fix samples redirect, and remove old sshd from TOC

thaJeztah:
- The /samples/wordpress/ location collided with an old redirect to the official images, so removed the redirect
- Removed the sshd example from the table and TOC, as the example   was removed, and now pointing to an alternative examples. The   page is kept for people that may have bookmarked it, but no  need to include it in the navigation.


thaJeztah:
@StefanScherer @usha-mandya ptal

StefanScherer (approved):
LGTM

PR #12746, opened 2021-04-25
`condition` is supported for `depends_on`

porterdarby:
Removed a statement about `condition` not being supported for `depends_on` in Version 3, as it is supported.

### Proposed changes

Remove the statement regarding Version 3 of the Docker Compose file not supporting `condition`. As seen in https://github.com/porterdarby/docker-compose-condition-test, the `condition` field works when used with the `depends_on` field for Docker Compose file with Version 3.


gtardif (approved):

gtardif:
Thx @porterdarby for the update !

usha-mandya (approved):

PR #12755, opened 2021-04-27
Doc currently_tagged filter for image list API

zappy-shu:
### Proposed changes

Adds the `currently_tagged` query parameter to the Advance Image Management - List Images API documentation. This filter has been added to the Docker Hub API already and allows consumers to list only images that have 0 or at least 1 current tags. Not including the query parameter will return images regardless of whether they have current tags or not.
This filter combines with the `status` filter


usha-mandya (approved):
LGTM

PR #12788, opened 2021-05-02
Remove trailing dot from filename

kadenscott:
### Proposed changes

There was a trailing dot, resulting in `.env.` instead of `.env`. This removes that dot.


thaJeztah (approved):
LGTM, thanks!

PR #12795, opened 2021-05-04
startup-order: fix example executing unquoted cmd

lightmare:
### Proposed changes

Using `"$@"` in variable assignment is very confusing, because in that context it behaves like `"$*"`.

`exec $cmd` performs field splitting and pathname expansion on the cmd string, which has surprising effect if some of the original parameters contain spaces or wildcards.

The correct way to execute command from positional parameters is `exec "$@"`

### DIFF
--- a/compose/startup-order.md
+++ b/compose/startup-order.md
@@ -69,7 +69,6 @@ script:
   
   host="$1"
   shift
-  cmd="$@"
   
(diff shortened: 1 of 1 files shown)


usha-mandya (approved):
LGTM. Thanks @lightmare

PR #12798, opened 2021-05-04
Removed -slim-buster from the Dockerfile

cwsaunders:
### Proposed changes

I removed the slim version of Python 3.8 because it is missing the mysqlclient package. Therefore, pip and pip3 are unable to complete the build.

### DIFF
--- a/language/python/build-images.md
+++ b/language/python/build-images.md
@@ -91,7 +91,7 @@ we would like to use for our application.
 ```dockerfile
 # syntax=docker/dockerfile:1
 
-FROM python:3.8-slim-buster
+FROM python:3.8
 ```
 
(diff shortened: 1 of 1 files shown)


thaJeztah:
Can you describe at what step it fails for you? I just tried the tutorial and it seemed to work for me, e.g. I inserted a record in the `widgets` table, and the "rest-server" container was able to access it:

```bash
curl http://localhost:5000/widgets
[{"name": "hello", "description": "world"}]
```

I'd rather not switch to the non-slim version if possible; it's a _lot_ larger:

Screenshot 2021-05-04 at 18 13 19


Generally it's recommended to only install in an image what's needed; which could involve updating the Dockerfile to install the mysql client if it's missing, but I'm wasn't able to reproduce the problem, so interested to hear what issue you were encountering.

craig-osterhout:
Thank you for contributing. Closing as stale. If the PR is still applicable, please provide the details requested in the previous comment.

Issue #12800, opened 2021-05-05
Spelling mistake in the dockerfile_best-practices file.


Roshan13046:
[develop/develop-images/dockerfile_best-practices.md](https://docs.docker.com/develop/develop-images/dockerfile_best-practices/)

The red underline text in the below image has to be changed from "thin" to "within".
![Screenshot (802)](https://user-images.githubusercontent.com/55108788/117122177-3aab7b00-adb3-11eb-94f5-07209be4bca8.png)

PR #12804, opened 2021-05-05
Remove excess word/typo reported in #12800

Kickimanjaro:
fixes https://github.com/docker/docker.github.io/issues/12800

This initially looked like a typo (reported in #12800) but I believe it is actually just an extra word that can be removed.


thaJeztah (approved):
LGTM

I guess it was originally added to try to describe that the writable layer is "lightweight" (only containing changes you made), but I agree it's adding more confusing than value, so let's remove it 👍

PR #12815, opened 2021-05-07
removed version 1 from composefileref

jerae-duffin:
Removed version 1 from Compose File Reference as it isn't supported any longer.


usha-mandya (approved):
LGTM

PR #12824, opened 2021-05-10
Desktop: move and cleanup "multi-arch" page

thaJeztah:
- desktop: move "multi-arch" page out of "for mac" section
  Both Docker Desktop for Mac and Docker Desktop for Windows provide
  this feature, so moving it out of the section that's specific to
  Docker Desktop for Mac.
- desktop: multi-arch: remove "install" section
  Docker Desktop 2.x is deprecated, so any current version of Desktop
  ships with buildx and the requirements to use multi-arch.
- desktop/multi-arch: various (markdown) touch-ups
  - remove mention of "experimental" for buildx
  - use "console" where possible
  - removed use of Bash "heredoc" in example, which may not work on
    shells other than Bash
  - wrap content to ~80 chars

The TOC now looks like:

Screenshot 2021-05-10 at 12 44 19

And the in-page TOC:

Screenshot 2021-05-10 at 12 57 54


thaJeztah:

PR #12831, opened 2021-05-11
cleanup toolbox migration page, and remove old "for-mac" page

thaJeztah:
The docker for mac "toolbox" migration page was not accessible as it
was redirected, so removing that page and related images.

This page contained instructions to backup images, which was described
elsewhere, so removing that part in favor of a link to the (better maintained)
content in the "desktop" section.

Also move the migration page that was located inside the "for windows"
directory to a "toolbox" directory, which matches the location inside the
TOC.

Finally; remove the "troubleshooting" entry from the TOC, as it was
redirecting to the migration page (and was under the "kitematic" section,
which was confusing).


usha-mandya (approved):
LGTM

PR #12835, opened 2021-05-11
Image management filter update

usha-mandya:

Update screenshots and copy for the new filter


FranchommeE (approved):
LGTM!

I am still tempted to nit pick on us saying the user can "sort and filter by tags", which feels a bit inaccurate to me. But a more accurate copy would probably be to complex to read. The UI is intuitive enough, that we can spare some details more in the Docs.

PR #12871, opened 2021-05-20
Add missing docker compose cli commands

tedhexaflow:
This PR adds the missing commands "docker compose port" and "docker compose restart".

### DIFF
--- a/_data/compose-cli/docker_compose.yaml
+++ b/_data/compose-cli/docker_compose.yaml
@@ -14,9 +14,11 @@ cname:
   - docker compose logs
   - docker compose ls
   - docker compose pause
+  - docker compose port
   - docker compose ps
   - docker compose pull
   - docker compose push
+  - docker compose restart
   - docker compose rm
   - docker compose run
   - docker compose start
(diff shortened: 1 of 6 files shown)


thaJeztah:
Thanks! I did a compare between the yaml files in the compose-cli repository for v1.0.14 and v1.0.15 (https://github.com/docker/compose-cli/tree/v1.0.14/docs/reference, https://github.com/docker/compose-cli/tree/v1.0.15/docs/reference), and it looks like there's various other changes that were not yet upstreamed to this repository.

release v1.0.15 is marked as a "pre-release" on GitHub (not sure if intentional), and current Docker Desktop looks to have v1.0.14, so perhaps we should make sure to document what's in there.

@gtardif @ndeloof could one of you have a look? (probably good to update the other yaml files as well, but want to make sure we're not documenting things that are not yet available to most users)

ndeloof:
Those yaml files are indeed generated on compose-cli repository and should not be manually edited here.
In an ideal universe, we would have set automation so an update PR is created to the docs repo any time we release compose-cli :P

We are waiting for 1.0.15 or later to be bundled in Docker Desktop before we update the public docs to avoid user confusion.

tedhexaflow:
The yaml files in this PR were generated from the compose-cli repo by running `go run docs/yaml/main/generate.go` and copied to this repo for the PR.

But I did not copy all yaml files. (only the 'port' and 'restart')

Please feel free to reopen and merge the PR if it is ok to merge only for those two missing commands.

Thanks. 🙏

PR #12872, opened 2021-05-20
Reorganize toc

thaJeztah:
- move "docker desktop" to the top of the "product manuals" section,
  and make it the default page
- Add a "Reference documentation" menu item to the TOC, which will be
  the "active" menu item when opening the reference section.
- move the "Samples home" menu item outside of the "samples" sub-menu,
  and add it at the top level as "samples overview", so that the "index"
  highlights that menu-item without opening the "samples" sub-menu. The
  "samples" submenu contains various samples which are a bit outdated,
  so probably shouldn't be _that_ prominent when opening the page.
  Moving the menu item also makes the "breadcrumb" navigation more
  natural (moving from a sample to "samples overview" navigates back
  to the "top" of the samples section.

With this change:

Screenshot 2021-05-20 at 19 24 21
Screenshot 2021-05-20 at 19 24 32
Screenshot 2021-05-20 at 19 24 53

### DIFF
--- a/_data/toc.yaml
+++ b/_data/toc.yaml
@@ -7,7 +7,7 @@ horizontalnav:
   path: /get-started/overview/
   node: guides
 - title: Product manuals
-  path: /engine/
+  path: /desktop/
   node: manuals
 - title: Reference
   path: /reference/
@@ -398,6 +398,8 @@ guides:
 - path: /docsarchive/
   title: Documentation archive
 reference:
+- path: /reference/
+  title: Reference documentation
 - sectiontitle: Command-line reference
   section:
   - sectiontitle: Docker CLI (docker)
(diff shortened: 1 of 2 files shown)


thaJeztah:
@usha-mandya ptal

usha-mandya:
@thaJeztah Thanks for the update. Looks good. The Product manuals card on the homepage still opens the Engine docs.

thaJeztah:
> the Product manuals card on the homepage still opens the Engine docs.

Oh, good catch; perhaps we probably used a "deep link" for that one as we don't have a `/manuals/` url or redirect yet. Let me fix that link

thaJeztah:
updated 👍

usha-mandya (approved):
LGTM

PR #12883, opened 2021-05-23
Dockerfile overview added.

SetTrend:
Currently, the navigation panels left and right of the content are are not visible to the reader when certain browser window widths are used (less than 1,000 pixels width).

Apart from that, the navigation tree doesn't come with any help about the different Dockerfile commands.

So, this page serves the purpose of a quick Dockerfile reference overview that can be pinned or bookmarked for easy reference.

### DIFF
--- a/_data/toc.yaml
+++ b/_data/toc.yaml
@@ -1082,8 +1082,12 @@ reference:
     path: /docker-hub/api/latest/
   - title: Registry API
     path: /registry/spec/api/
-- title: Dockerfile reference
-  path: /engine/reference/builder/
+- sectiontitle: Dockerfile reference
+  section:
+    - title: Dockerfile reference Overwiew
+      path: /engine/reference/dockerfile_toc/
+    - title: Dockerfile reference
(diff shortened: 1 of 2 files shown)


SetTrend:
@thaJeztah : Do you believe we'll get any progress here?

thaJeztah:
Thanks for the ping; I just recalled I had a work-in-progress branch on my machine to address this in a slightly different way. I pushed it as a PR in https://github.com/docker/docker.github.io/pull/13260 (but it's possible it needs more work)

With that PR, the Dockerfile commands will appear in the left-hand navigation, which (I think) may be more "natural", and with the added "bonus" that they will always be available on the left-hand, and not scroll out of view when reading the page.

thaJeztah:
Oh, I may have mis-read the diff; this is a separate page, correct?

Perhaps we should combine the two 🤔

SetTrend:
Exactly. I was hoping people could use my page for bookmarking it so they could quickly branch into what they're searching for.

PR #12886, opened 2021-05-24
Strengthen language about emulation being best effort only

stephen-turner:
### Proposed changes

Strengthen language about emulation being best effort only


usha-mandya (approved):
LGTM. Thank you @stephen-turner

PR #12900, opened 2021-05-26
[Ubuntu installation] replace apt-get with apt

kevinadhiguna:
### Proposed changes

I replaced `apt-get` with `apt` as it is the one that Linux distributions recommend. It gives you the tools you need to manage your packages. Most importantly, with fewer but easy-to-remember options, it is simpler to use. 

In addition, what I replaced are commands that install and remove packages :
```diff
- sudo apt-get install
+ sudo apt install
```

```diff
- sudo apt-get remove
+ sudo apt remove
```

```diff
- sudo apt-get purge
+ sudo apt purge
```

Any feedback will be appreciated, thanks!

References : 
- [Difference Between apt and apt-get Explained - It's Foss](https://itsfoss.com/apt-vs-apt-get-difference/)
- [What is the difference between apt and apt-get? - Ask Ubuntu](https://askubuntu.com/questions/445384/what-is-the-difference-between-apt-and-apt-get)

### DIFF
--- a/engine/install/ubuntu.md
+++ b/engine/install/ubuntu.md
@@ -41,10 +41,10 @@ Older versions of Docker were called `docker`, `docker.io`, or `docker-engine`.
 If these are installed, uninstall them:
 
 ```console
-$ sudo apt-get remove docker docker-engine docker.io containerd runc
+$ sudo apt remove docker docker-engine docker.io containerd runc
 ```
 
-It's OK if `apt-get` reports that none of these packages are installed.
+It's OK if `apt` reports that none of these packages are installed.
 
 The contents of `/var/lib/docker/`, including images, containers, volumes, and
 networks, are preserved. If you do not need to save your existing data, and want to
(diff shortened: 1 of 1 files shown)


crydotsnake (approved):

crydotsnake (approved):

crydotsnake:
@usha-mandya

thaJeztah:
Thank you for contributing!

We use `apt-get` instead of `apt` in these instructions so that  the example commands can be used by users who want to script these steps. While `apt` is intended for "interactive" use, `apt-get` is still the recommended tool for scripting (as it has a stable API, whereas `apt` does not).

Some earlier discussions on this can be found in https://github.com/docker/docker.github.io/pull/10547, https://github.com/docker/docker.github.io/pull/8135, and probably some other ones.


I'm closing this PR because of the above, but thanks again for contributing, and feel free to continue the conversation.

PR #12907, opened 2021-05-27
Fix authorization header doc in hub api

zappy-shu:
### Proposed changes

Quick fix to the Hub API spec that incorrectly says `Authentication` HTTP header when it should say `Authorization`


mark-dr (approved):

usha-mandya (approved):
LGTM

PR #12937, opened 2021-06-02
Indent webhook notifications Authorization header

matthewbalvanz-wf:
### Proposed changes

Since `Authorization` is intended to be a header applied to HTTP requests it should be indented to place it in the YAML map named `headers` instead of being a sibling of it.


usha-mandya (approved):
LGTM. Thank you @matthewbalvanz-wf

PR #12939, opened 2021-06-02
Docker Desktop on Mac Apple silicon: Remove release date and first official version number

StefanScherer:
Now that Docker Desktop for Mac on Apple silicon is GA and also has some newer patch releases, this page looks a bit confusing. We showed the first official release date and the first version number. Now we are at 3.3.3 and the next minor version bump is just around the corner.

I think it's time to clean the page a little bit.


usha-mandya (approved):
LGTM

PR #12952, opened 2021-06-03
Typo Fix for docker-compose up.

EarthSchlange:
### Proposed changes

Changes "docker compose up" to "docker-compse up" which is the correct command in documentation.


SethFalco (approved):

usha-mandya:
@EarthSchlange `docker compose` is indeed a command. See https://docs.docker.com/compose/cli-command/ and https://docs.docker.com/engine/reference/commandline/compose/ for details on the new `docker compose` command.

PR #12977, opened 2021-06-08
Alpine Linux is no longer under 5 MB

breun:
### Proposed changes

The docs say Alpine is under 5 MB, but `alpine:latest` is currently 5.61 MB. I changed the docs to say Alpine is under 6 MB.


usha-mandya (approved):
LGTM. Thank you ;)

PR #12981, opened 2021-06-08
Minor fixes in the run-tests.md

dhasanbasic:
Fixed line numbers in example, as two lines 57 were shown and it was not obvious which line needs to be changed.
Removed profile argument from example outputs, as the corresponding Dockerfile examples for the test image do not have it.


usha-mandya (approved):
LGTM. Thank you @dhasanbasic

PR #12985, opened 2021-06-09
Fix a typo by adding a quote

shokkunrf:
### Proposed changes
In the `Reference / Compose file reference / Version 3` section of the documentation, change the example about ports syntax.


usha-mandya (approved):
LGTM. Thanks for fixing this @shokkunrf :)

PR #12992, opened 2021-06-10
Publish updates from master

usha-mandya:


_data/compose-cli/docker_compose.yaml
```diff
@@ -1,22 +1,77 @@
 command: docker compose
 short: Docker Compose
-long: Docker Compose
+long: "You can use compose subcommand, `docker compose [-f <arg>...] [options] [COMMAND]
```
thaJeztah, line 3: I need to have a look at "why" it's not using the long form for these. Could be if the source markdown has special characters (`tab` most notably, or trailing whitespaces). I think I have some fixes in the YAML generator script for that as well.


Not important though; this works, but it's just not so readable.

thaJeztah (approved):
LGTM

_data/compose-cli/docker_compose.yaml
usha-mandya: Sure, thanks @thaJeztah.

PR #13015, opened 2021-06-14
Fix references to docker-compose command

manuelgrabowski:
Apologies if this is not the correct way to fix this. I just noticed on https://docs.docker.com/engine/reference/commandline/compose/ that the first example is using `docker-compose` instead of `docker compose`. It was very confusing trying to open a PR to fix it, as the "Edit this page" link leads into an (I think?) outdated repository, and the other option "Request docs changes" seems overly complicated for a simple copy&paste fix like this.


usha-mandya:
@manuelgrabowski The source files for CLI reference documentation exist in the GitHub repositories and that's why you were pointed to the upstream repo when you clicked **Edit this page**. 

@ndeloof what's the best way to get the upstream doc fixed?

ndeloof:
Source of thrust is https://github.com/docker/compose-cli/tree/main/docs/reference
I wish we automate creating sync PRs to this repo as part of the release process...

manuelgrabowski:
I opened a PR over there instead, thanks for the advice! :)

PR #13021, opened 2021-06-14
docker docker run -rm -d -p ... fix

cevheri:
docker run -rm .. command is incorrect. 
➜  docker docker run -rm -d -p 8080:8080 --name springboot-server java-docker
unknown shorthand flag: 'r' in -rm
See 'docker run --help'.
correct command
➜  docker docker run -d -p 8080:8080 --name springboot-server java-docker


cevheri:
incorrect PR

PR #13039, opened 2021-06-21
Authentication to GHCR with GITHUB_TOKEN

### LINKED ISSUE #13037: Authentication to GHCR is now possible via GITHUB_TOKEN
[ci-cd/github-actions.md](https://docs.docker.com/ci-cd/github-actions/)

In this documentation page, the user is instructed to create a Personal Access Token `password: ${{ secrets.GHCR_TOKEN }}`. I was confused on how to do this in the context of an organization, only to find out that there is actually an easier way now: just using `GITHUB_TOKEN`: https://docs.github.com/en/packages/working-with-a-github-packages-registry/migrating-to-the-container-registry-from-the-docker-registry#authenticating-to-the-container-registry

> The Container registry now supports `GITHUB_TOKEN` for easy and secure authentication in your workflows. If your workflow is using a personal access token (PAT) to authenticate to ghcr.io, then we highly recommend you update your workflow to use `GITHUB_TOKEN`.

The fix here would be rather straight-forward (replace `GHCR_TOKEN` with `GITHUB_TOKEN`, but I am not sure if this would have any side-effects in other pages.

I hope I am not just more confused through this process.

crazy-max:
Fixes #13037

### DIFF
--- a/ci-cd/github-actions.md
+++ b/ci-cd/github-actions.md
@@ -194,13 +194,18 @@ Now, go to GitHub and check your Actions
 
 ![Push tagged version](images/push-tagged-version.png){:width="500px"}
 
-Now, let’s set up a second GitHub action file to store our latest commit as an image in the GitHub registry. You may want to do this to:
+Now, let’s set up a second GitHub action file to store our latest commit as an image in the GitHub Container Registry. You may want to do this to:
 
 1. Run your nightly tests or recurring tests, or
 2. To share work in progress images with colleagues.
 
 Let’s clone our previous GitHub action and add back in our previous logic for all pushes. This will mean we have two workflow files, our previous one and our new one we will now work on.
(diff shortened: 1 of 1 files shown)


atrauzzi:
@crazy-max - You might want to update this one again as the URL `https://docs.github.com/en/packages/guides/enabling-improved-container-support` is now redirecting to `https://docs.github.com/en/packages/working-with-a-github-packages-registry/working-with-the-container-registry` which says nothing about enabling _improved container support_.

I'm trying to do a simple push to ghcr using the docker CLI command in a task, would love to see instructions on that as well.

crazy-max:
@atrauzzi Thanks for your feedback. I have fixed the link and also it looks like the _"Enable improved container support"_ is not a thing anymore in the GitHub settings (enabled by default now).

thaJeztah (approved):
LGTM, thanks!

PR #13040, opened 2021-06-21
Mention the necessity of `--network host`

Drugoy:
This example is, honestly, idiotic, because it doesn't even explain what task it solves.
No one needs to "just pass *_PROXY envs to created containers", as it is useless by itself.
People want to make containers have internet access via proxy, but the proxy is usually on the host machine's host or networks, not on docker containers networks!
The example described in the docs just tells how to configure containers query THEIR 127.0.0.1:3001, not host machine's 127.0.0.1:3001. This is useless!

### Proposed changes

I improved docs by adding an important mention of an extra step that is necessary to solve a common task (that the article should've explained how to solve, but instead failed to).

### DIFF
--- a/network/proxy.md
+++ b/network/proxy.md
@@ -47,6 +47,9 @@ configure it in different ways:
 
  2. When you create or start new containers, the environment variables are
     set automatically within the container.
+ 
+ 3. If the intention was to use host's 127.0.0.1:3001 - don't forget to
+    specify `--network host` when creating container.
 
 
 ## Use environment variables


usha-mandya:
@Drugoy Apologies for the inconvenience. We are looking to update this topic. Would it help if we add an example with a proxy server that has a different IP address? Thanks

Drugoy:
@usha-mandya
Probably it'll be better the way you suggested, but then please mention that this IP address should be outside container's docker network, because otherwise it's expected that you run your own proxy in a sibling container in the same docker network (which is unusual).

usha-mandya:
@Drugoy The changes are now live on the prod site. I'm closing this ticket here for that reason. Feel free to continue the conversation after I close. Thanks.

PR #13049, opened 2021-06-22
Docs for Volumes UI updates

usha-mandya:

Add updates on Volumes UI FE changes


mat007 (commented):
We have invert the tabs, so the first one is now **IN USE** and the second one **DATA**, maybe we should invert the descriptions here as well?

Also we will most likely ship the «Save as…» menu entry: when selected it prompts the user for a location on the host to save the file (or a zip if on a folder in the volume).
We should probably document it?

mat007 (approved):
Looks good!

PR #13056, opened 2021-06-23
Use custom Dockerfile for Dev Environments image

glours:
### Proposed changes
Use a dedicated Dockerfile to start a Dev Environments of the `docker.github.io` repo. 

First step is to `bundle install` and then you can run Jekyll with `bundle exec jekyll serve`

https://user-images.githubusercontent.com/705411/123113482-e2056f80-d43e-11eb-8e33-7f9834b4e419.mp4

### DIFF
--- a/.docker/Dockerfile.devenv
+++ b/.docker/Dockerfile.devenv
@@ -1,0 +1,3 @@
+FROM docker/dev-environments-ruby:stable-1
+RUN gem install bundler jekyll
+CMD ["bundle", "install"]
--- a/.docker/config.json
+++ b/.docker/config.json
@@ -1,3 +1,3 @@
 {
-    "image": "jekyll/jekyll"
+    "dockerfile": "Dockerfile.devenv"
 }


gtardif (approved):
Tested using Guillaume's forked repo as devenv, docs site live & all good from VSCode

glours:
I'll create a base image to support Ruby in Dev Environments, do not merge this PR for the moment

usha-mandya:
Sounds good. Thanks @glours

glours:
@usha-mandya the PR is ready to merge and use the Ruby base image for Dev Environments

usha-mandya:
Sounds good. Thanks @glours. 

@thaJeztah Do you mind taking a look before we merge this PR?

usha-mandya (approved):

nebuk89:
Looks good to me :) A little heavier weight than before on the image but still better and the sharing works :D

.docker/Dockerfile.devenv
```diff
@@ -0,0 +1,3 @@
+FROM docker/dev-environments-ruby:stable-1
+RUN gem install bundler jekyll
+CMD ["bundle", "install"]
```
thaJeztah, line 3: Wondering, won't we miss all the env-vars (and potentially custom things) that are in the main dockerfile? Would it be possible to "compose" a dev-environment friendly environment in which we copy the tools from `<some image>` for the integration? https://github.com/docker/docker.github.io/blob/77a86298a373d06802cc8096dca127c4dceb38df/Dockerfile#L14-L41

thaJeztah: Wondering if "dev environment" in this case should be "ruby + jekyll" or a dev-environment with "docker installed" (as the build process of this project is a docker container (which _happens_ to be using `jekyll` as one of it's components).

glours: Not sure to understand 🤔 , we just added a Dockerfile for the dev flow. 
This one is used for production, no?

thaJeztah: Depending on the build-args, the Dockerfile is for "production" or for local development/preview.

This repository (for the "development flow") has no dependencies, other than docker, because steps to build the documentation are defined in the Dockerfile, which can be built (and run) with `docker-compose up --build`.

In the Dockerfile, this includes Jekyll (with github-pages) (for some steps) as well as bash/curl for other steps, and runs an nginx server to view the results.

I think the discrepancy here with the "dev environment" container is that that container sets up a Ruby/Jekyll environment which will only provide "some" of those. From that perspective, a better reflection of the "host" environment (inside a dev-container) would be a dev-container containing the `docker` (and `docker compose`) CLI (with `docker-compose`).

PR #13058, opened 2021-06-24
Mention compose v2 beta again

mat007:
### Proposed changes

Mention compose v2 beta again.
Most users are not aware this is on-going and may not think of scrolling down to past releases to understand what is happening.


usha-mandya:
Makes sense. Thanks.

usha-mandya (approved):

PR #13131, opened 2021-07-06
Update hello-world ref to clarify it is source only

tianon:
The "hello-world" repository (somewhat intentionally) does not contain any instructions for building it, but the instructions to do so already followed this line, so this adjusts the wording to be more clear that the source code lives in the hello-world repository without implying that there are instructions for building there as well.

(Closes https://github.com/docker-library/hello-world/issues/90)


usha-mandya (approved):
LGTM

PR #13138, opened 2021-07-07
Add "-dev"-suffix to new image

Sammeeey:
### Proposed changes

I've added a "-dev"-suffix in the lines where we are working with the new python-docker image.
In this way we do not overwrite the previous one which we build in the [image part of the guide](https://docs.docker.com/language/python/build-images/).
In this way the can differentiate between the two different images and make more clear what happens.
Otherwise (with the current version of the docs) the previous image name would turn to "<none>".

### DIFF
--- a/language/python/develop.md
+++ b/language/python/develop.md
@@ -153,7 +153,7 @@ $ pip3 freeze > requirements.txt
 Now we can build our image.
 
 ```shell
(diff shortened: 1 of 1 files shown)


usha-mandya (approved):
LGTM. Thanks.

PR #13142, opened 2021-07-07
Adjusted sentence to make more clear what to do

Sammeeey:
### Proposed changes

I inserted the same sentence as in [Run Containers](https://docs.docker.com/language/python/run-containers/#overview).
This makes more clear to make curl requests in a new terminal.
Because when the container is running you will not be able to type commands in the same terminal window.

### Related issues (optional)

#13143


usha-mandya (approved):
LGTM. Thank you @Sammeeey

PR #13150, opened 2021-07-08
Update 08_using_compose.md

KolinHuang:
In the "Run the application stack" part, I think there is a mistake in writing. That is "If you don't already, you'll see output that looks like this...". I change it to "If you have done already, you'll see output that looks like this...".


usha-mandya (approved):
LGTM

PR #13155, opened 2021-07-08
Update prometheus.md

ecormaksin:
Discussed in the [StackOverFlow](https://stackoverflow.com/questions/53202260/cant-collect-docker-metrics-using-prometheus), I couldn't get docker endpoint metrics on Ubuntu 20.04 (with `"metrics-addr" : "127.0.0.1:9323"` on `/etc/docker/daemon.json`, `- targets: ['localhost:9323']` on `prometheus.yml`, `http://localhost:9090/targets/` on url). By default, a docker container cannot access via localhost, can it? So if it is true, I think that we should use `127.0.0.1` insted of `localhost`.

### DIFF
--- a/config/daemon/prometheus.md
+++ b/config/daemon/prometheus.md
@@ -36,7 +36,7 @@ If the file is currently empty, paste the following:
 
 ```json
 {
-  "metrics-addr" : "127.0.0.1:9323",
+  "metrics-addr" : "0.0.0.0:9323",
   "experimental" : true
 }
 ```
@@ -111,7 +111,7 @@ scrape_configs:
          # scheme defaults to 'http'.
 
     static_configs:
-      - targets: ['localhost:9323']
+      - targets: ['<docker0 inet address>:9323']
 ```
 
 </div><!-- linux -->
@@ -240,7 +240,7 @@ PS C:\> docker service create --replicas 1 --name my-prometheus
 </div><!-- windows -->
 </div><!-- tabs -->
 
-Verify that the Docker target is listed at http://localhost:9090/targets/.
+Verify that the Docker target is listed at http://127.0.0.1:9090/targets/.
 
 ![Prometheus targets page](images/prometheus-targets.png)
 


usha-mandya:
@thaJeztah Could you PTAL?

config/daemon/prometheus.md
```diff
@@ -240,7 +240,7 @@ PS C:\> docker service create --replicas 1 --name my-prometheus
 </div><!-- windows -->
 </div><!-- tabs -->
 
-Verify that the Docker target is listed at http://localhost:9090/targets/.
+Verify that the Docker target is listed at http://127.0.0.1:9090/targets/.
```
thaJeztah, line 243: I don't think `localhost` or `127.0.0.1` would make a difference (only difference I could think of is `localhost` potentially resolving to `::1` (IPv6 localhost).

config/daemon/prometheus.md
```diff
@@ -36,7 +36,7 @@ If the file is currently empty, paste the following:
 
 ```json
 {
-  "metrics-addr" : "127.0.0.1:9323",
+  "metrics-addr" : "0.0.0.0:9323",
```
thaJeztah, line 39: I'll need to have a closer look, but this changes the metrics to be accessible on "any" IP address, not just the loopback IP address.

Depending on the setup, this may be what's wanted, but it also means metrics (which potentially could include sensitive data?) are exposed publicly.

thaJeztah (changes requested):
Thanks for contributing; I will need to have a closer look at this. I see some other (potential) issues in this document (e.g. `docker.for.mac.host.internal` may no longer be current).

I _think_ it's also possible to scrape prometheus metrics using a unix-socket. For documentation purposes (and to allow a container to access), perhaps that would be a good option to document (and have an example for that).

A unix-socket would be more secure (as it's not accessible externally) _and_ would work around the "localhost inside container" vs "localhost of the daemon" issue.

PR #13170, opened 2021-07-12
Change Install Docs Order Related to Importance of the OS

rene-hermenau:
The order of importance in the docs when its comes to installing docker should be 
MAC, Linux, Windows or Linux, Mac, Windows

Don't put Windows systems install instructions before linux. Every serious developer is working with docker either in Linux or Mac env but not Windows.


usha-mandya:
Thanks for the PR. We also have a lot of Windows users based on our stats. Therefore, we'll have to close this PR without accepting the changes to retain the existing order. Thanks again for the PR and your ideas to make our Docs better :)

PR #13172, opened 2021-07-12
daemon.json does not exist in a fresh docker install - Update json-file.md

rene-hermenau:
The file daemon.json does not exist in a fresh docker install in Ubuntu. This needs to be explained in the docs.


usha-mandya (approved):
LGTM. Thank you @rene-hermenau

PR #13084, opened 2021-06-30
Updated content

jerae-duffin:
Updated content and images for PAT.

PR #13174, opened 2021-07-12
Add access tokens API documentation

technicallyjosh:
### Proposed changes

This PR adds API documentation for personal access tokens.

Also added improvements including
- improvements to slugs (ids) using more idiomatic swagger naming conventions for friendlier urls
- tag groups
- rate limiting section
- minor verbiage changes to existing docs for clarity

**This feature is not released yet. Please hold merging until the feature is live.**

### Related issues (optional)

General documentation updates: #13084

### DIFF
--- a/docker-hub/api/latest.yaml
+++ b/docker-hub/api/latest.yaml
@@ -1,11 +1,12 @@
 swagger: '2.0'
+
+host: hub.docker.com
+basePath: '/'
 schemes:
   - https
-host: hub.docker.com
-basePath: "/"
-produces:
-  - application/json
 consumes:
+  - application/json
+produces:
   - application/json
 info:
   title: Docker HUB API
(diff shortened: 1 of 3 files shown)


amineck (approved):
LGTM!

docker-hub/api/latest.yaml
```diff
@@ -18,37 +19,89 @@ info:
...
+      - `X-RateLimit-Limit` - The limit of requests per minute.
+      - `X-RateLimit-Remaining` - The remaining amount of calls within the limit
+      period.
+      - `X-RateLimit-Reset` - The unix timestamp of when the remaining resets.
+
+      If you have hit the limit, you will receive a response status of `429` and
+      the `X-Retry-After` header in the response.
+
+      The `X-Retry-After` header is a unix timestamp of when you can call the
+      API again.
```
ingshtrom, line 41:
```suggestion
      API again.
      
      
      Note: These rate limits are separate from anti-abuse and Docker Hub download, or pull, rate limiting. To learn more about Docker Hub pull rate limiting, see [Docker Hub Download rate limit](https://docs.docker.com/docker-hub/download-rate-limit/)
```

You might want to put a snippet in there to explicitly call out the difference between the API rate limiting. 

Come to think of it, I wonder if it would be goo to link from /docker-hub/download-rate-limit back to this page 🤔 . It might be helpful so that you can easily navigate between the rate limiting documentation we have.

technicallyjosh: Good call. Yeah would be good to link as well as differentiate the different rate-limiting we have.

technicallyjosh: Added the suggestion. Thanks!

docker-hub/api/latest.yaml
```diff
@@ -18,37 +19,93 @@ info:
...
+
+      ### Scopes
+
+      For each scope grouping (in this case "repo"), you only need to define 1
+      scope as any lower scopes are assumed. For example: If you define
+      `repo:write`, the API assumes the scope of both `repo:read` *and*
+      `repo:public_read` as well. If you were to define both `repo:write` *and*
+      `repo:read`, then `repo:read` is assumed by `repo:write` and ignored.
+
+      ***Reminder: Personal access tokens should be kept secret like a password
```
usha-mandya, line 73: Suggest: Treat your personal access token like your password and keep it secret. You cannot retrieve your token after  it is generated.

Can we also use a Note here instead of a reminder? Here's the format we use in Docs, but not sure if Swagger can handle this format

```
> **Note**
>
> <text>
```

docker-hub/api/latest.yaml
```diff
@@ -18,37 +19,93 @@ info:
...
-  - name: Images
-    x-displayName: Advanced image management
+  - name: access-tokens
+    x-displayName: Personal Access Tokens
+    description: |
+      The Personal Access Token endpoints lets you manage personal access
+      tokens. For more information, see
+      [Access Tokens](https://docs.docker.com/docker-hub/access-tokens/).
+
+      You can use a personal access token in place of a password in the
```
usha-mandya, line 60:
```suggestion
      You can use a personal access token instead of a password in the
```

usha-mandya:
@technicallyjosh Thanks for the PR. Left some minor comments. Looks good otherwise.

docker-hub/api/latest.yaml
technicallyjosh: I can update the text. Notes don't work in swagger

technicallyjosh:
@usha-mandya Thanks! I've updated per comments.

usha-mandya (approved):
LGTM

PR #13198, opened 2021-07-19
Fix/improve a docker-build example

filipjares:
The example makes more sense this way.


filipjares:
Oh and I see the other examples on the page are similar.

Would it make sense to do this replacements in these other examples as well?

usha-mandya:
@filipjares Thanks for the PR. The RUN command makes more sense in this context as it is about building the image. I'll have to close your PR without accepting the changes for this reason. Thanks again for your PR. Feel free to continue the conversation after I close.

Issue #13207, opened 2021-07-21
Step 6/9 : COPY *.go . When using COPY with more than one source file, the destination must be a directory and end with a /


stolpovski:
[language/golang/build-images.md](https://docs.docker.com/language/golang/build-images/)

thaJeztah:
Thanks for reporting. I suspect you are trying to build with the classic builder, which is a bit more strict on these.

Are you seeing the same problem when enabling buildkit? You can do so by setting the `DOCKER_BUILDKIT=1` environment variable; assuming you're using a Linux or macOS shell;

```bash
DOCKER_BUILDKIT=1 docker build <etc>.......
```

Or

```bash
export DOCKER_BUILDKIT=1
docker build <etc>.......
```

Other options are outlined in https://docs.docker.com/go/buildkit/



That said, I think it's good to be explicit, and to update the examples to have a trailing slash if a directory is intended to be the target.

thaJeztah:
Opened https://github.com/docker/docker.github.io/pull/13208

PR #13208, opened 2021-07-21
Fix Dockerfile COPY/ADD examples without a trailing slash

thaJeztah:
Fixes https://github.com/docker/docker.github.io/issues/13207

When copying files to a destination directory, the classic builder requires
the destination to have a trailing slash (to indicate the target is a _directory_.
not a filename). BuildKit is a bit more flexible in this, and will assume the
target is a directory, but users following the example with buildkit disabled
might see an error message, e.g.:

    Sending build context to Docker daemon  3.072kB
    Step 1/2 : FROM busybox
     ---> 69593048aa3a
    Step 2/2 : COPY *.go .
    When using COPY with more than one source file, the destination must be a directory and end with a /

It doesn't hurt to be explicit, so this patch updates some examples to prevent
this.

### DIFF
--- a/develop/develop-images/dockerfile_best-practices.md
+++ b/develop/develop-images/dockerfile_best-practices.md
@@ -170,13 +170,13 @@ context, refer to [exclude with .dockerignore](#exclude-with-dockerignore).
 > 
 > docker build -t myimage:latest -<<EOF
 > FROM busybox
-> COPY somefile.txt .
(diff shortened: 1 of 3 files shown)


thaJeztah:
@crazy-max @usha-mandya PTAL

crazy-max (approved):
LGTM

usha-mandya (approved):

PR #13223, opened 2021-07-23
Update build-images.md

rajendrasolleti:


crydotsnake (approved):

thaJeztah (changes requested):
Can you describe the problem you ran into? the `python:3.8-slim-buster` image is an official image on Docker Hub, and should automatically be pulled if it's not present on your local machine; https://hub.docker.com/_/python?tab=tags&page=1&ordering=last_updated&name=3.8-slim-buster

PR #13230, opened 2021-07-26
Update 06_bind_mounts.md

Segodnyaev:
### Proposed changes

This script crashing on 4th step, with error:
`error /app/node_modules/sqlite3: Command failed.
Exit code: 1
Command: node-pre-gyp install --fallback-to-build
Arguments: 
Directory: /app/node_modules/sqlite3`

```
You need to install the latest version of Python.
gyp ERR! find Python Node-gyp should be able to find and use Python.

```

Because docker image alpine hasn`t installed python and g++
On previous step in this Guides (https://docs.docker.com/get-started/02_our_app/) you used the command to preinstall python and g++ before calling "yarn install && yarn run ..."

### DIFF
--- a/get-started/06_bind_mounts.md
+++ b/get-started/06_bind_mounts.md
@@ -46,7 +46,7 @@ So, let's do it!
     docker run -dp 3000:3000 \
         -w /app -v "$(pwd):/app" \
         node:12-alpine \
-        sh -c "yarn install && yarn run dev"
+        sh -c "apk add --no-cache python g++ make && yarn install && yarn run dev"
     ```
 
     If you are using PowerShell then use this command:
@@ -55,7 +55,7 @@ So, let's do it!
     docker run -dp 3000:3000 `
         -w /app -v "$(pwd):/app" `
         node:12-alpine `
-        sh -c "yarn install && yarn run dev"
+        sh -c "apk add --no-cache python g++ make && yarn install && yarn run dev"
     ```
 
     - `-dp 3000:3000` - same as before. Run in detached (background) mode and create a port mapping


usha-mandya (approved):
LGTM. Thanks @Segodnyaev

get-started/06_bind_mounts.md
```diff
@@ -55,7 +55,7 @@ So, let's do it!
     docker run -dp 3000:3000 `
         -w /app -v "$(pwd):/app" `
         node:12-alpine `
-        sh -c "yarn install && yarn run dev"
+        sh -c "apk add --no-cache python g++ make && yarn install && yarn run dev"
```
thaJeztah, line 58: Looks like the bullet at line 65 also needs to be updated (as it explains the command).

Also considering if we should use the image we built in "part 2: sample application" for this, as that would already have these dependencies installed (or if we can update the to not require these dependencies, as it may distract from the main purpose).

SeanNesdoly: Installing `python` failed on an arm64 machine. After running the below command:
```
docker run -dp 3000:3000 \
        -w /app -v "$(pwd):/app" \
        node:12-alpine \
        sh -c "apk add --no-cache python g++ make && yarn install && yarn run dev"
```
This error was produced:

```
fetch https://dl-cdn.alpinelinux.org/alpine/v3.15/main/aarch64/APKINDEX.tar.gz
fetch https://dl-cdn.alpinelinux.org/alpine/v3.15/community/aarch64/APKINDEX.tar.gz
ERROR: unable to select packages:
  python (no such package):
    required by: world[python]
```

```suggestion
        sh -c "apk add --no-cache python2 g++ make && yarn install && yarn run dev"
```

jacobweber:
This problem exists elsewhere in the getting-started docs, e.g. in https://docs.docker.com/get-started/09_image_best/.

usha-mandya:
Looks like this issue has already been fixed. Please see the latest docs - https://docs.docker.com/get-started/06_bind_mounts/. Closing the PR for this reason.

PR #13235, opened 2021-07-27
add explicit instruction to stop local server

qmacro:
### Proposed changes

This relates to the [Test the application section of the Build images tutorial](https://docs.docker.com/language/nodejs/build-images/#test-the-application). In this section, `node server.js` is used to start the server locally. The reader is not instructed to stop the server. 

In the [Run containers](https://docs.docker.com/language/nodejs/run-containers/) tutorial that follows this one, the reader is instructed to start up the container, and try to access `http://localhost:8000/test` which is now being served from the Node.js process inside the container. The idea is that this should not work (without port forwarding) and they should see an error. However, if the local Node.js process from the `node server.js` invocation earlier, they won't see the error and may be confused.

So this modification adds an explicit instruction to the "Build containers" tutorial to stop the `node server.js` process. I used the same phrasing as in the "Run containers" tutorial, i.e. "press ctrl-c". 

Great tutorials, btw!

### DIFF
--- a/language/nodejs/build-images.md
+++ b/language/nodejs/build-images.md
@@ -72,12 +72,22 @@ $ curl http://localhost:8000/test
 
 Switch back to the terminal where our server is running. You should now see the following requests in the server logs.
 
-```
+```console
 2020-XX-31T16:35:08:4260  INFO: POST /test
 2020-XX-31T16:35:21:3560  INFO: GET /test
 ```
 
-At this stage, you're done with testing this server script locally, so stop it running by going back to the first terminal session where you started it up, and press ctrl-c.
+Great! We verified the application works. At this stage, you're done with testing the server script locally.
+
+Press `CTRL-c` from within the terminal session where the server is running to stop it.
+
+```console
+2021-08-06T12:11:33:8930  INFO: POST /test
+2021-08-06T12:11:41:5860  INFO: GET /test
+^Cshutting down...
+```
+
+We will now continue to build and run the application in Docker.
 
 ## Create a Dockerfile for Node.js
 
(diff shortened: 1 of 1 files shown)


language/nodejs/build-images.md
```diff
@@ -77,6 +77,8 @@ Switch back to the terminal where our server is running. You should now see the
 2020-XX-31T16:35:21:3560  INFO: GET /test
 ```
 
+At this stage, you're done with testing this server script locally, so stop it running by going back to the first terminal session where you started it up, and press ctrl-c.
```
thaJeztah, line 80: Thanks! Yes, it makes sense to be sure it's stopped.

Perhaps we should show what it looks like after pressing that, and guide into the next part. Something like;

    Great! We verified the application works. At this stage, you're done with testing
    the server script locally.
    
    Press `CTRL-c` from within the terminal session where the server is running to
    stop it.
    
    ```console
    2021-08-06T12:11:33:8930  INFO: POST /test
    2021-08-06T12:11:41:5860  INFO: GET /test
    ^Cshutting down...
    ```
    
    We will now continue to build and run the application in Docker.


(sorry, couldn't make the a GitHub "suggestion", as it didn't like the nested "fences" (```)). Happy to push those changes to your branch though.

thaJeztah: /cc @usha-mandya

qmacro: I've added your suggestions to the PR 👍

thaJeztah (approved):
LGTM, thanks!

thaJeztah:
@usha-mandya PTAL

usha-mandya (approved):
LGTM. Thank you @qmacro :)

PR #13262, opened 2021-07-31

shivdyl22:
Updated ENV variable usage

updated ENV from $PATH to $PG_VERSION

### Proposed changes

An incorrect ENV variable was used.

### DIFF
--- a/develop/develop-images/dockerfile_best-practices.md
+++ b/develop/develop-images/dockerfile_best-practices.md
@@ -644,7 +644,7 @@ version bumps are easier to maintain, as seen in the following example:
 ENV PG_MAJOR=9.3
 ENV PG_VERSION=9.3.4
 RUN curl -SL https://example.com/postgres-$PG_VERSION.tar.xz | tar -xJC /usr/src/postgres && …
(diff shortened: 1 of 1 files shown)


develop/develop-images/dockerfile_best-practices.md
```diff
@@ -644,7 +644,7 @@ version bumps are easier to maintain, as seen in the following example:
 ENV PG_MAJOR=9.3
 ENV PG_VERSION=9.3.4
 RUN curl -SL https://example.com/postgres-$PG_VERSION.tar.xz | tar -xJC /usr/src/postgres && …
-ENV PATH=/usr/local/postgres-$PG_MAJOR/bin:$PATH
+ENV PATH=/usr/local/postgres-$PG_MAJOR/bin:$PG_VERSION
```
thaJeztah, line 647: I don't think this is correct; the intent here is to prepend the postgres's binary location to the existing `$PATH`.

Can you describe what problem you ran into?

thaJeztah (changes requested):

PR #13265, opened 2021-08-01
Mention about use of `|` or `&&` in one's commands

akauppi:
I think the root documentation should have a little bit more about what's allowed and what not in the `command`. I find it handy to place even complicated commands, within the YAML file (instead of spreading them into external `.sh` files), just like we sometimes do in `npm` `package.json`.

Anyhow, the fact that `|` or `&&` quietly chops my command was surprising.

The content of the PR is just a suggestion.

### DIFF
--- a/compose/compose-file/compose-file-v3.md
+++ b/compose/compose-file/compose-file-v3.md
@@ -450,6 +450,14 @@ The command can also be a list, in a manner similar to
 command: ["bundle", "exec", "thin", "-p", "3000"]
 ```
 
+> Note when using pipes (`|`) or `&&` in your command
+>
+> If you want to bind multiple commands, instead of placing them in an external `.sh` file, use the below syntax. Otherwise, your command gets quietly chopped at the `|` or `&&`. `-o pipefail` makes sure that a failing command fails the whole pipe (and thereafter, the composed Docker stack).
+>
+>```yaml
+>command: bash -o pipefail -c 'echo a | grep -v a'
+>``` 
+
 ### configs
 
 Grant access to configs on a per-service basis using the per-service `configs`


compose/compose-file/compose-file-v3.md
```diff
@@ -450,6 +450,14 @@ The command can also be a list, in a manner similar to
...
 ```
 
+> Note when using pipes (`|`) or `&&` in your command
+>
+> If you want to bind multiple commands, instead of placing them in an external `.sh` file, use the below syntax. Otherwise, your command gets quietly chopped at the `|` or `&&`. `-o pipefail` makes sure that a failing command fails the whole pipe (and thereafter, the composed Docker stack).
+>
+>```yaml
+>command: bash -o pipefail -c 'echo a | grep -v a'
+>``` 
+
```
thaJeztah, line 460: We should probably avoid using `bash` in the example, because it won't be available in many images, or at least show a "non-bash" alternative. Perhaps use `sh -c` instead; that leaves us with the `pipefail` option, which wasn't part of the POSIX spec (so a shell _may_ not support it), but it looks like it was accepted, and (I _think_) added to the standard (current version of `dash` doesn't support it yet though)

"exec-form" (JSON notation) should also work (perhaps worth mentioning);

```yaml
command: ["sh", "-o", "pipefail", "-c", "echo a | grep -v a"]
```
I'm a bit on the fence if we actually need to make it a note, or if it could just be a regular paragraph, e.g.

    The container exits once the command that is specified completes. If multiple
    commands are combined (`|` or `&&`), this means the container exits after the
    first command completes, unless the commands are executed inside a shell, or
    combined in a script (`.sh`) that's included in the container. Use the syntax
    below if you need to bind multiple commands, and do not want to create a `.sh`
    script:

    ```yaml
    command: sh -c 'echo a | grep -v a'
    ```

    Or, using "exec-form" (list);

    ```yaml
    command: ["sh", "-c", "echo a | grep -v a"]
    ```

    If the container provides `bash` as a shell, use the `-o pipefail` option to
    make sure that a failing command fails the whole pipe:
    
    ```yaml
    command: bash -o pipefail -c 'echo a | grep -v a'
    ```

usha-mandya:
Thanks for the PR. Closing this based on Sebastiaan's comment above. Feel free to continue the conversation after I close.

PR #13271, opened 2021-08-02
add postgresql-dev, python3-dev to Dockerfile

vildan-valeev:
add postgresql-dev, python3-dev to Dockerfile, because could not install psycopg2-binary package from requirements.txt

### DIFF
--- a/samples/django.md
+++ b/samples/django.md
@@ -32,6 +32,8 @@ and a `docker-compose.yml` file. (You can use either a `.yml` or `.yaml` extensi
    FROM python:3
    ENV PYTHONUNBUFFERED=1
    WORKDIR /code
+   RUN apk update \
+    && apk add postgresql-dev gcc python3-dev musl-dev
    COPY requirements.txt /code/
    RUN pip install -r requirements.txt
    COPY . /code/


samples/django.md
```diff
@@ -32,6 +32,8 @@ and a `docker-compose.yml` file. (You can use either a `.yml` or `.yaml` extensi
    FROM python:3
    ENV PYTHONUNBUFFERED=1
    WORKDIR /code
+   RUN apk update \
+    && apk add postgresql-dev gcc python3-dev musl-dev
```
thaJeztah, line 36: This doesn't look correct. The example here uses `python:3` as a base image, which is an alias for `python:3-buster`, and based on a Debian base image, not alpine; https://github.com/docker-library/python/blob/d18aeb0a2f1e52d8ff82dfe57104943174a0d8ab/3.9/buster/Dockerfile

Screenshot 2021-08-06 at 12 04 05

I suspect you were trying to optimise the example and create a smaller image by using alpine, but the current Dockerfile looks correct

thaJeztah (changes requested):

thaJeztah:
Thanks for contributing! I'm closing this PR per my comment above, but feel free to continue the conversation in case I missed something 👍

PR #13275, opened 2021-08-03
Add not about NFS mounts for rootless data-root

seemethere:
### Proposed changes

NFS mounts are not currently supported for rootless data-root so there should probably be a note about it so no one gets confused when it doesn't work.

Errors for running rootless containers when your data-root is an NFS mount look like:

```
docker: failed to register layer: ApplyLayer exit status 1 stdout:  stderr: open /root/.bash_logout: permission denied.
```


thaJeztah (approved):
SGTM, thanks!

@AkihiroSuda PTAL (is this something the daemon could detect?)

AkihiroSuda (changes requested):
I don’t think this is specific to rootless

AkihiroSuda:
Opened https://github.com/docker/docker.github.io/pull/13282

PR #13299, opened 2021-08-06
"console-idate" shell examples

thaJeztah:
This replaces the use of `bash` and `shell` where suitable, to allow easier copy/pasting of shell examples without copying the prompt or process output.


usha-mandya (approved):
LGTM

PR #13305, opened 2021-08-08
Update install.md

doggy8088:
### Proposed changes

Use curl with [jq](https://stedolan.github.io/jq/) to get the latest tag name from GitHub. No need to specify tag name manually.

### DIFF
--- a/compose/install.md
+++ b/compose/install.md
@@ -131,6 +131,12 @@ also included below.
 
     > To install a different version of Compose, substitute `{{site.compose_version}}`
     > with the version of Compose you want to use.
+  
+    If you have [jq](https://stedolan.github.io/jq/) installed, you can use this command to install the latest stable release of Docker Compose.
+  
+    ```console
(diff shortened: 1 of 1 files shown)


compose/install.md
```diff
@@ -131,6 +131,12 @@ also included below.
 
     > To install a different version of Compose, substitute `{{site.compose_version}}`
     > with the version of Compose you want to use.
+  
+    If you have [jq](https://stedolan.github.io/jq/) installed, you can use this command to install the latest stable release of Docker Compose.
+  
+    ```console
+    $ sudo curl -L "https://github.com/docker/compose/releases/download/$(curl -s 'https://api.github.com/repos/docker/compose/releases' | jq '[.[] | select(.prerelease==false)][0] | .tag_name' -r)/docker-compose-$(uname -s)-$(uname -m)" -o /usr/local/bin/docker-compose
```
thaJeztah, line 138: I don't think `jq` is needed for this; you can use the `latest` URL, which redirects to the right location;

```bash
$ curl -L "https://github.com/docker/compose/releases/latest/download/docker-compose-$(uname -s)-$(uname -m)" -o /usr/local/bin/docker-compose
```

thaJeztah (changes requested):
Thanks for contributing!

I think similar changes were discussed in the past, and for our documentation, we prefer to explicitly set the release we want to describe in the docs instead of the "latest" release which may be on GitHub. (related: https://github.com/docker/docker.github.io/pull/7061, https://github.com/docker/docker.github.io/pull/8397)

PR #13309, opened 2021-08-09
Remove Kitematic docs, as it's deprecated

thaJeztah:
Screenshot 2021-08-09 at 12 57 24


usha-mandya:
Thanks @thaJeztah. Should we add the Kitematic deprecation notice in the Desktop release notes?

thaJeztah:
> Thanks @thaJeztah. Should we add the Kitematic deprecation notice in the Desktop release notes?

Hmm... good one; I think it's been removed for some time now (would have to check which release removed it?) I see [DESKTOP-1736] was related to removing it from the menu, but not sure which release that was, and if there's other bits removing it in later releases. @mat007 do you know?

mat007:
Kitematic was removed from Desktop ages ago, I think it’s too late to release note it 😁

usha-mandya (approved):
LGTM

PR #13369, opened 2021-08-17
Add tooltips and links to reference docs to Dockerfile codeblocks

thaJeztah:
This is an initial "experiment" on making codeblocks more useful; it's
just a quick implementation, and can use a lot of improvements, but we
can see if it's useful in its current form already (as a starting point).

More context below:

With example blocks being parsed/highlighted by rouge, we should be able
to pick keywords from them; for example, in the highlighted Dockerfile
examples, we can show tool-tips (even a hover-card) with more details
about the command.

Shell examples may be a bit more involved, but we could still detect;

- adjacent keywords (`docker` + `run` => `docker run`)
- with the yaml-docs we have (we could generate JSON and upload them
  as a "database"),   we could even do an AJAX call to dynamically
  fetch  flag  descriptions, etc.

This very quick and dirty test adds tooltips to Dockerfile commands
in examples. When hovering, it shoulds a tooltip ("click for  more
information about this command"), and when clicking, navigates to the
Dockerfile reference.

Ideally, we wouldn't navigate away from the current page for initial
details (instead, we should present a rich hoverbox / pane), to provide
the user with more details without interrupting the article they were
reading.

Some screenshots:

A Dockerfile example in our docs:

Screenshot 2021-08-17 at 21 46 22

When hovering over a keyword:

Screenshot 2021-08-17 at 21 46 42

And when clicking on the keyword, the user is navigated to the reference docs:

Screenshot 2021-08-17 at 21 47 16

### DIFF
--- a/js/docs.js
+++ b/js/docs.js
@@ -215,4 +215,14 @@ ready(() => {
         const group = $(this).attr("data-group");
         $(`.nav-tabs > li > a[data-group='${ group }']`).tab("show");
     });
+
+    $('.language-dockerfile span.k').tooltip({
+        title: function() {
+            let c = this.textContent;
+            this.style.cursor = 'help';
+            $(this).on('click', () => { window.location.href = "/engine/reference/builder/#"+c.toLowerCase()});
+            return 'Learn more about the "'+ c + '" Dockerfile command.'
+        },
+        placement: "auto"
+    });
 });


thaJeztah:
@usha-mandya @crazy-max @tonistiigi As mentioned; a really quick 'n dirty write-up, but perhaps a fun thing to work on as a hackday thing. Combined with some structured (json file) source of data, we could possibly make it more generic than it currently is.

thaJeztah:
Turns out I was not the only one with such ideas; @crazy-max pointed me to https://squidfunk.github.io/mkdocs-material/reference/code-blocks/#adding-annotations 😂

crazy-max:
That looks like a great improvement @thaJeztah!

thaJeztah:
I love the mkdocs example, because that's roughly what I had in mind for the "richer" snippets. Let me add a screenshot here to show;

Screenshot 2021-08-17 at 22 05 05

usha-mandya:
This is a fantastic idea @thaJeztah. Thanks for this! We can extend this further to add hints and tooltips for Docker terms.

thaJeztah:
I rephrased the commit message (and description above) a bit, and moved it out of draft

For others watching this; we discussed merging this as an initial "experiment". It's functional, although not very "polished" yet, but we can use it as a starting point, and also see if users respond to the new feature.

The change itself is small, and if hell breaks loose, is easy to reverts (although I don't expect it to cause issues)

@crazy-max @usha-mandya PTAL

crazy-max (approved):

usha-mandya (approved):
LGTM

PR #13373, opened 2021-08-18
fix(yaml): Avoid bool type error

rpatterson:
### Proposed changes

Docs contained an invalid example:

```
ERROR: The Compose file './docker-compose.yml' is invalid because:                                                                                                                   
services.*.restart contains an invalid type, it should be a string 
```

Use quotes to avoid YAML's implicit type conversion.  Someone may also want to do an audit of [other type conversions that might cause problems](https://www.rpatterson.net/blog/docker-gotchas/#yaml-needs-some-zen) and/or quote all string values to avoid surprising behavior.

https://docs.docker.com/compose/compose-file/compose-file-v3/


rpatterson:
Oops, somehow I got mixed up between the v2 and v3 versions of the pages, this is handled correctly in the v3 docs (though not optimally IMO).  This change is still a minor improvement, however, so I'll leave it open but feel free to close if you disagree.

PR #13381, opened 2021-08-20
Updated docs for s390x for latest version

anujajakhade:
### Proposed changes
Added generic fix for enabling the openSUSE repo for SLES.
Added SLES support in btrfs driver document.
Added Ubuntu, SLES and RHEL 8 support in rootless docker document.


thaJeztah (changes requested):
I left some suggestions; let me try push them here, otherwise (if that doesn't work) I may push a "carry" PR

PR #13410, opened 2021-08-25
guides: use include for "create a Dockerfile" sections

thaJeztah:
Yet another one of my "I had this in a branch on my machine" pull requests 😂

This is trying to unify wording used in these sections, to allow maintaining it in a single location.


_includes/guides/create-dockerfile.md
```diff
@@ -0,0 +1,39 @@
+A Dockerfile is a text document that contains the instructions to assemble a
+Docker image. When we tell Docker to build our image by executing the `docker build`
+command, Docker reads these instructions, executes them, and creates a Docker
+image as a result.
```
thaJeztah, line 4: We may want to touch-up this section a bit; some versions of this section described it as "executed in sequence" or "one-by-one". With BuildKit, steps in a Dockerfile can be executed in parallel in some cases, so it's not all "sequential".

Explaining "building a graph" etc. may be out of scope for this though, so I kept it as-is, but perhaps we can find a way to include that information (or link to somewhere that describes it)

_includes/guides/create-dockerfile.md
```diff
@@ -0,0 +1,39 @@
...
+must appear before any other comment, whitespace, or Dockerfile instruction in
+your Dockerfile, and should be the first line in Dockerfiles.
+
+```dockerfile
+# syntax=docker/dockerfile:1
+```
+
+We recommend using `docker/dockerfile:1`, which always points to the latest release
+of the version 1 syntax. BuildKit automatically checks for updates of the syntax
+before building, making sure you are using the most current version.
```
thaJeztah, line 39: Possibly we could link to the reference, and add a snippet about "alternative syntaxes", but I wasn't sure if that would still be "in scope" for these guides

thaJeztah:
@chris-crone @crazy-max ptal

_includes/guides/create-dockerfile.md
```diff
@@ -0,0 +1,39 @@
+A Dockerfile is a text document that contains the instructions to assemble a
```
thaJeztah, line 1: Oh, perhaps worth mentioning; I tried to distinguish "Dockerfile" (the "concept") from `Dockerfile` (the _file_), and only formatting  `Dockerfile` as `code` if it's referring to the _file_. It's tricky to make that distinction everywhere though.

chris-crone (approved):
LGTM

_includes/guides/create-dockerfile.md
```diff
@@ -0,0 +1,39 @@
...
+when parsing the Dockerfile, and allows older Docker versions with BuildKit enabled
+to upgrade the parser before starting the build. [Parser directives](/engine/reference/builder/#parser-directives)
+must appear before any other comment, whitespace, or Dockerfile instruction in
+your Dockerfile, and should be the first line in Dockerfiles.
+
+```dockerfile
+# syntax=docker/dockerfile:1
+```
+
+We recommend using `docker/dockerfile:1`, which always points to the latest release
```
crazy-max, line 37:
```suggestion
We recommend using `docker/dockerfile:1` as [Dockerfile frontend](https://github.com/moby/buildkit/blob/master/frontend/dockerfile/docs/syntax.md), which always points to the latest release
```

thaJeztah: Ideally, we would have the "docker file" reference unified. Currently part of it lives in `docker/cli`, and part of it (front-end syntax) lives in moby/buildkit.

That way users wouldn't have to head to GitHub to find the missing bits (but we may have to add additional styles for the docs to annotate options (a "badge"?) that explains they are buildkit-only, and require syntax X or up.

PR #13433, opened 2021-08-29
Fix install options cards styling. Resolves #13432

### LINKED ISSUE #13432: [Styling] Install options card content goes out of card container
### Problem description
The install options card content on page [get-docker](https://docs.docker.com/get-docker/) goes out of its container because the `height: 250px;` given to it is too small for the content.

### Problem location

- I saw a problem on the following URL: [Get Docker](https://docs.docker.com/get-docker/)

- Other: <DETAILS>

Visual Representation:

image

### Suggestions for a fix
I'll link a Pull Request to fix this!

danBamikiya:
### Proposed changes

This fixes the styling issue the install options cards have on [Get Docker](https://docs.docker.com/get-docker/) page.

### Related issues (optional)
Resolves #13432

@thaJeztah

### DIFF
--- a/_scss/_content.scss
+++ b/_scss/_content.scss
@@ -69,7 +69,7 @@ code {
     text-align: center;
     margin: 0 8px 15px;
     float: left;
-    height: 250px;
+    height: 270px;
     width: 100%;
     background: $bg-component;
     img {
--- a/get-docker.md
+++ b/get-docker.md
@@ -19,7 +19,7 @@ significantly reduce the delay between writing code and running it in production
 You can download and install Docker on multiple platforms. Refer to the following
 section and choose the best installation path for you.
 
-<div class="component-container">
+<div class="component-container" style="margin-bottom: 30px;">
     <!--start row-->
     <div class="row">
         <div class="col-sm-12 col-md-12 col-lg-4 block">


thaJeztah:
Thanks! Looks like the problem depends on the window width. This change does help for some widths, but "moves" the problem, as for smaller widths (before the blocks wrap), the text still looks to run out of the box;

Screenshot 2021-09-02 at 00 25 50


These styles definitely need some cleaning up (or a redesign for these tiles), but trying to have a quick fix,  was having a quick look if https://stackoverflow.com/a/19695851/1811501 helps to address it short term.

That solution (with some other changes) works to address this particular problem;

Screenshot 2021-09-02 at 00 26 01


But just before one of the break-points, the blocks wrap, and now are not the same _width_, which is also a bit "yuk" 😓 
Screenshot 2021-09-02 at 00 30 33

Screenshot 2021-09-02 at 00 30 53


Let me push my changes as a draft PR, in case you want to look if that solution may work (with some additional changes)

thaJeztah:
Pushed my branch in https://github.com/docker/docker.github.io/pull/13452

danBamikiya:
Hi @thaJeztah I just pushed a commit to your branch that should resolve the yucky width issue.

PR #13441, opened 2021-08-31
Update with modern references/packages

Sameer711:
It doesnt work with postgresql 9.3 anymore since the package is not available any longer. I have updated the references to work


craig-osterhout:
Thanks for sharing the fix. Closing this PR as this content will be removed from the docs repo.

PR #13444, opened 2021-09-01
_config.yml: remove temporary URL parameters

thaJeztah:
These URLs are now all final, so remove the paramters and move them inline.


usha-mandya (approved):
LGTM. Thank you @thaJeztah

PR #13452, opened 2021-09-01
fix vertical size of "get docker" blocks

thaJeztah:
closes https://github.com/docker/docker.github.io/pull/13433
fixes https://github.com/docker/docker.github.io/issues/13432

This attempts to fix the problem where (on smaller windows), the text
flows out of the blocks.

This uses one of the solutions mentioned in https://stackoverflow.com/a/19695851/1811501,
with some additonal style changes.

Unfortunately, it introduces another issue, where (just before a break-point; after
the right-hand navigation is hidden, but before we switch to "mobile" / "small",
the blocks are wrapped, but no longer have equal _widths_.

See https://github.com/docker/docker.github.io/pull/13433#issuecomment-910843789

### DIFF
--- a/_scss/_content.scss
+++ b/_scss/_content.scss
@@ -59,17 +59,28 @@ code {
  */
 
 .component-container {
-    padding: 0;
-    margin: 0 0 50px;
-    width: 100%;
-}
+  padding:  0;
+  margin:   0 0 50px;
+  width:    100%;
 
-.component {
-    padding: 15px 25px 5px 15px;
+  .block {
+    padding: 0 15px 10px 15px;
+  }
+
+  .row {
+    display: flex;
+    flex-wrap: wrap;
+  }
+  .row > [class*='col-'] {
+    display: flex;
+    flex-direction: column;
+  }
+
+  .component {
+    padding: 15px 25px 15px 15px;
     text-align: center;
-    margin: 0 8px 15px;
     float: left;
-    height: 250px;
+    height: 100%;
     width: 100%;
     background: $bg-component;
     img {
@@ -80,11 +91,12 @@ code {
       justify-content: center;
     }
     h2 {
-      font-size: 22px;
+      font-size: 18px;
     }
     h2, h3, p {
       margin: 0;
     }
+  }
 }
 
 a.anchorLink {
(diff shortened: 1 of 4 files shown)


danBamikiya:
Hi @thaJeztah thanks for the detailed review/feedback. 
The yucky block widths happen on `max-width: 760px` because the `.block` elements have the class `col-sm-12` which applies
```css
@media (min-width: 768px)
.col-sm-12 {
    width: 100%;
}
```
but no `width: 100%` is applied to them at widths below the `786px` width that why they are all having whatever (yucky) width can contain their content not a unified width.

Iterating upon your pushed branch I added the class `col-xs-12` to the `.block` items which fixes the issue and gives them all a unified width at widths below `786px`

<h2>Before</h2>

docker_docs_website_issue_1_fix_2

<h2>After</h2>

 | docker_docs_website_issue_1_fix_1

danBamikiya:
I'll push my commit to your branch now.

danBamikiya:
Just pushed it!

thaJeztah:
Thank you! Sorry for the delayed response, it was getting late (in my timezone) and I had to drop off.

I'll give it a try with your changes. I was trying some variations locally, but then noticed they conflicted with the "Language-specific guides" overview (https://docs.docker.com/language/)

danBamikiya:
Oh, that's OK! 
Let me know if the changes I made still doesn't fix the issue or brings up more bugs. 
I'll be glad to help!

thaJeztah:
> Oh, that's OK!
> Let me know if the changes I made still doesn't fix the issue or brings up more bugs.
> I'll be glad to help!

Thanks for your help! Had to go into some meetings, but I think I fixed the "language guides" overview as well now

Preview is at:

- https://deploy-preview-13452--docsdocker.netlify.app/get-docker/
- https://deploy-preview-13452--docsdocker.netlify.app/language/

thaJeztah:
@usha-mandya PTAL

usha-mandya (approved):
LGTM. Thank you @danBamikiya and @thaJeztah

danBamikiya:
@thaJeztah @usha-mandya You're welcome. Glad I could help!

Issue #13430, opened 2021-08-29
Not able to copy given code


shyamal-jadav:
[language/nodejs/develop.md](https://docs.docker.com/language/nodejs/develop/)

![Capture](https://user-images.githubusercontent.com/26648851/131259200-b220fe9c-a220-4b80-b745-be46c7edd4ad.PNG)

thaJeztah:
Thanks for reporting! I'll have a look at this.

For context; we tried to make the code-blocks easier to copy without accidentally copying the command _output_ that's shown in some examples, but it looks like in this case, the code-highlighter detects the "payload" to be "output", and therefore doesn't allow copying.

thaJeztah:
Opened https://github.com/docker/docker.github.io/pull/13456 to fix it

PR #13456, opened 2021-09-02
language/nodejs: un-wrap JSON in example to allow copying

thaJeztah:
The JSON string was wrapped, causing the highlighter to consider it "output",
which cannot be selected/copied in "console" blocks.

This unwraps the JSON so that it can be copied.

fixes https://github.com/docker/docker.github.io/issues/13430


usha-mandya (approved):
LGTM

Issue #13435, opened 2021-08-30
inheriting anonymous volumes with --volumes-from prevents auto-remove


figaw:
[storage/volumes.md](https://docs.docker.com/storage/volumes/)

This part is incorrect.

> To automatically remove anonymous volumes, use the --rm option. For example, this command creates an anonymous /foo volume. When the container is removed, the Docker Engine removes the /foo volume but not the awesome volume.
```
 docker run --rm -v /foo -v awesome:/bar busybox top
```

It's possible to inherit an anonymous volume, that's not automatically removed when the container is removed. I'm not sure if this is a bug in Docker, because I guess it makes sense. But at least the documentation is slightly off (or just incomplete), and I'm not entirely sure how to rephrase it.

# Example

1. Start a container with an anonymous volume that should be automatically removed upon exit with,

```
docker run -it --rm -v /mnt/tmp --name my-alpine alpine sh
```

2. Start another container (NB: without `--rm`) that inherits the volume with,

```
docker run -it --volumes-from my-alpine alpine sh
```

3. Exit the first container and verify that the volume is still there, because it's in use by the second container.
4. Exit the second container, and verify that the volume is still there, (presumably because the second container hasn't been removed)
5. Remove the container with `rm`, not using `rm -v`. And the anonymous volume is leftover, because it wasn't removed with the initial container, and the new container wasn't removed with the `-v` flag.

thaJeztah:
Thanks for reporting. I think this was by design. Generally, anonymous volumes are often used as "ephemeral" storage, but once it's referenced by multiple containers that becomes more ambiguous. The `--volumes-from` _copies_ the volume definitions from the original container; for the second container, the volume is considered a "regular" volume (albeit with a very ugly name). I recall this was also the reason to settle on naming them "anonymous" volumes (the term "ephemeral" was used occasionally), as they're not always _ephemeral_ (naming is hard!).

I agree that this may be worth documenting; if you have suggestions how to phrase / explain this (perhaps we need a heading that explains anonymous volumes a bit more in depth 🤔), contributions are definitely welcome!

figaw:
Thanks for the clarification! I did expect it to be more like an.. "undocumented funny" rather than a bug 😄
Maybe it could be added as an "NB" like on this page <https://docs.docker.com/storage/bind-mounts/#configure-the-selinux-label>

Text from: https://docs.docker.com/storage/volumes/#remove-anonymous-volumes

# Remove anonymous volumes
[...]

PR #13471, opened 2021-09-07
Add anonymous volumes autoremove info

figaw:
Anonymous volumes aren't removed if another container adopts them with `--volumes-from`, see: https://github.com/docker/docker.github.io/issues/13435

### Proposed changes

Updated documentation to match what happens.


usha-mandya:
@djs55 Could you PTAL?

djs55 (approved):
LGTM -- I think this is indeed what happens. The only thing I'm not sure about is how commonly used `--volumes-from` is. Maybe it should be "note" rather than "important". A minor detail though.

usha-mandya (approved):
LGTM

PR #13484, opened 2021-09-08
Change distribution URL to distribution/distribution

milosgajdos:
### Proposed changes

`distribution` project has been donated to CNCF and subsequently moved to a dedicated GH org called `distribution` See [here](https://github.com/distribution/distribution/). This PR updates the URL from `docker/distribution` to `distribution/distribution`.


StefanScherer (approved):
LGTM !

PR #13495, opened 2021-09-10
Fix broken link to Windows path troubleshooting

camerondurham:
### Proposed changes

Fix broken link  link to Windows Path Troubleshooting header. The link to `#path-conversions-on-windows` seems to not work since it is not a high enough header, for example: https://docs.docker.com/desktop/windows/troubleshoot/#path-conversions-on-windows

Changed to point here and called out **Path Conversions on Windows** section: https://docs.docker.com/desktop/windows/troubleshoot/#volumes

n/a

### Related issues (optional)

n/a


usha-mandya (approved):

usha-mandya:
Thank you @camerondurham. I've added a direct link to the Path conversion on Windows section.

PR #13514, opened 2021-09-13
check for broken links

thaJeztah:
carry of https://github.com/docker/docker.github.io/pull/12147 - rebase and trying to tweak configuration


thaJeztah:
Trying to make it ignore the anchor links for the API (because the API reference is generated client-side with JavaScript, so it won't be able to check them).

```
- ./_site/engine/swarm/manage-nodes/index.html
  *  linking to internal hash #operation/ServiceCreate that does not exist (line 316)
     <a href="/engine/api/v1.31/#operation/ServiceCreate"></a>
  *  linking to internal hash #operation/ServiceCreate that does not exist (line 316)
```

Not sure what syntax to use to configure that (docs mention it accepts a string or regex, but so far it doesn't work) 🤔

.github/workflows/build-pr.yml
```diff
@@ -14,6 +14,17 @@ jobs:
...
+      - name: copy files to host
+        run: docker build --target=deploy-source --output=./_site .
+      - name: check for broken links
+        uses: chabad360/htmlproofer@master
+        with:
+          directory: ./_site
+          # for available options, refer to:
+          # - https://github.com/gjtorikian/html-proofer
+          # - https://github.com/gjtorikian/html-proofer/blob/main/bin/htmlproofer
+          arguments: --disable-external --internal-domains="docs.docker.com,docs-stage.docker.com,localhost:4000" --file-ignore="/^./_site/engine/api/.*$/,./_site/registry/configuration/index.html,./_site/engine/reference/commandline/buildx_build/index.html,./_site/engine/reference/commandline/buildx_create/index.html" --url-ignore="/^/docker-hub/api/latest/.*$/,/^/engine/api/v.+/#.*$/,/^/glossary/.*$/"
```
thaJeztah, line 26: We need to remove the buildx excludes once they're fixed (https://github.com/docker/buildx/pull/765 is merged, but only on master)

PR #13517, opened 2021-09-13
Fix various broken (anchor) links

thaJeztah:
relates to https://github.com/docker/docker.github.io/pull/13514

Still some links to fix in "upstream resources", but with this, the number of failures is reduced considerably;

```
HTML-Proofer found 106 failures!
```

### DIFF
--- a/_data/compose-cli/docker_compose_events.yaml
+++ b/_data/compose-cli/docker_compose_events.yaml
@@ -19,7 +19,7 @@ long: |-
     }
     ```
 
-    The events that can be received using this can be seen [here](/engine/reference/commandline/events/#object-types).
+    The events that can be received using this can be seen [here](https://docs.docker.com/engine/reference/commandline/events/#object-types).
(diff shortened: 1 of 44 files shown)


_data/compose-cli/docker_compose_events.yaml
```diff
@@ -19,7 +19,7 @@ long: |-
     }
     ```
 
-    The events that can be received using this can be seen [here](/engine/reference/commandline/events/#object-types).
+    The events that can be received using this can be seen [here](https://docs.docker.com/engine/reference/commandline/events/#object-types).
```
thaJeztah, line 22: Temporary fix (needs to be fixed in upstream, or need to figure out why it prepends `/engine/reference/commandline/` to the URL)

usha-mandya: Yup, makes sense

usha-mandya (approved):
LGTM

PR #13539, opened 2021-09-15
Random names to match the random name shown in the screenshot

mountainash:
### Proposed changes

Random names to match the random name shown in the screenshot at https://docs.docker.com/desktop/dev-environments/#start-a-single-container-dev-environment

Makes reading the docs a little less confusing for a first-timer.


usha-mandya (approved):
Makes sense. Thank you @mountainash

PR #13551, opened 2021-09-16
Remove outdated sentence

smanilov:
The output following the sentence no longer contains what the sentence
claims.

### Proposed changes

Just a documentation change. Removed a sentence that is misleading, as it refers to output that seems to no longer be there (there is no mention of "latest" in the cited output.)


usha-mandya (approved):
LGTM. Thank you @smanilov

PR #13555, opened 2021-09-17
Updated yum -y install yum-utils to yum -y upgrade

brianrkleis:
Docker requires more updated libraries than yum-utils. I've tested yum-utils in a new Virtual Machine without updating any libs, and couldn't get docker working. For new environments this is a better command and for any environment.

### DIFF
--- a/engine/install/centos.md
+++ b/engine/install/centos.md
@@ -83,7 +83,7 @@ Install the `yum-utils` package (which provides the `yum-config-manager`
 utility) and set up the **stable** repository.
 
 ```console
-$ sudo yum install -y yum-utils
+$ sudo yum -y upgrade
 
 $ sudo yum-config-manager \
     --add-repo \


engine/install/centos.md
```diff
@@ -83,7 +83,7 @@ Install the `yum-utils` package (which provides the `yum-config-manager`
 utility) and set up the **stable** repository.
 
 ```console
-$ sudo yum install -y yum-utils
+$ sudo yum -y upgrade
```
thaJeztah, line 86: This looks to be addressing different goals;

```bash
 sudo yum install -y yum-utils
```

Is added here to make sure `yum-config-manager` is installed (which is used below).

`yum upgrade` will update / upgrade _all packages and dependencies_ installed on the machine. While this _could_ be a good recommendation, this would be somewhat out of scope for the installation instructions.

What's the exact problem you ran into? Did the installation fail, or did it warn that a newer version of a dependency was needed?

thaJeztah:
closing this, as I don't think this change is needed per my comment above

PR #13590, opened 2021-09-24
Simplify wrapper script for running multiple processes

sheerun:
### Proposed changes

It seems that wrapper script for running multiple processes can be simplified to something like this


StefanScherer (approved):
Thank you @sheerun, that is a very elegant solution.
LGTM

PR #13600, opened 2021-09-26
Update multi-arch.md

rbrtmrtn:
Minor style fix (Arm => ARM)

### DIFF
--- a/desktop/multi-arch.md
+++ b/desktop/multi-arch.md
@@ -32,7 +32,7 @@ like the `arm32v7` or `ppc64le` variants of the busybox image.
 
 ## Build multi-arch images with Buildx
 
-Docker is now making it easier than ever to develop containers on, and for Arm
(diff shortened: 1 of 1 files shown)


usha-mandya:
@rbrtmrtn Thanks for the PR. Arm uses 'Arm' and `Armv7` on their website and documentation. I'll have to close this PR without merging for this reason.

image

rbrtmrtn:
Hi @usha-mandya, thank you for taking the time to look at this PR!

The reason I updated this from Arm => ARM is that the uppercase version is also used in the docs:

> Because of this, you can run an ARM container, like the arm32v7 or ppc64le variants of the busybox image.

Googling "arm cpu", I get more results for ARM than Arm, including Wikipedia. I can understand if the Docker team wants to use the more stylized Arm, but just a friendly flag this should probably be consistent throughout the docs :)

PR #13637, opened 2021-10-04
Fix version invocation for v2

mat007:
### Proposed changes

To display the version with v2 we use a command `version` instead of a flag `--version`.


usha-mandya (approved):
LGTM

PR #13657, opened 2021-10-06
fix download link

programmerq:

### DIFF
--- a/compose/cli-command.md
+++ b/compose/cli-command.md
@@ -79,7 +79,7 @@ from the [project release page](https://github.com/docker/compose/releases){:tar
 
     ```console
     $ mkdir -p ~/.docker/cli-plugins/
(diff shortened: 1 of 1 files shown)


compose/cli-command.md
```diff
@@ -79,7 +79,7 @@ from the [project release page](https://github.com/docker/compose/releases){:tar
 
     ```console
     $ mkdir -p ~/.docker/cli-plugins/
-    $ curl -SL https://github.com/docker/compose/releases/download/v2.0.0/docker-compose-linux-x86_64 -o ~/.docker/cli-plugins/docker-compose
+    $ curl -SL https://github.com/docker/compose/releases/download/v2.0.0/docker-compose-linux-amd64 -o ~/.docker/cli-plugins/docker-compose
```
thaJeztah, line 82: I think this was fixed in v2.0.1, which switched back to x86_64; https://github.com/docker/compose/releases/tag/v2.0.1

So perhaps instead we should update the link to v2.0.1

PR #13663, opened 2021-10-07
Updated PAT

jerae-duffin:
Updated scopes and images


usha-mandya:
@jerae-duffin Could you take screenshots without the cursor and optimize the screenshots? I think there's an option to do avoid capturing the cursor in SnagIt. Also, for the border, could you use the color `#C7C7C7`?

PR #13664, opened 2021-10-07
Updated steps

jerae-duffin:
Added confirmation step after json file


jerae-duffin:
@usha-mandya Updated with new error image and restructured steps

PR #11640, opened 2020-10-27
Theme: refactor colors and some fixes/updates

thaJeztah:
Continuing to "upstream" changes I have in my working branches 😅 

- _scss: remove some unused variables
- _scss: change colors to HSL
  While more verbose, HSL colors (Hue, Saturation, Lightness) allow for easier selecting color variations that only differ in lightness, but have the same hue. This conversion also reveals some slight differences in hue between existing color-variations that I'll address in a follow-up commit.
- _scss: tweak inconsistencies in hue
  Changes should be mostly invisible, except for a very slight increase in contrast in the "night" theme between the body- background and left/right navigation blocks.
- _scss: define variable for landing footer color
- _scss: some refactoring of landing-page cta-banners
- update http://dockr.ly links to use TLS

PR #13680, opened 2021-10-12
Update most links to use https by default

PeterDaveHello:
### Proposed changes

Update most links to use https by default for better security practice

### Related issues (optional)

- #11640
- d405b17ca2a394ccb2a473d22eb918147e2291e6

### DIFF
--- a/_data/engine-cli/docker_config_inspect.yaml
+++ b/_data/engine-cli/docker_config_inspect.yaml
@@ -6,7 +6,7 @@ long: |-
   By default, this renders all results in a JSON array. If a format is specified,
   the given template will be executed for each result.
 
-  Go's [text/template](http://golang.org/pkg/text/template/) package
+  Go's [text/template](https://golang.org/pkg/text/template/) package
   describes all the details of the format.
 
   For detailed information about using configs, refer to [store configuration data using Docker Configs](https://docs.docker.com/engine/swarm/configs/).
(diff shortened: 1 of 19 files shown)


_data/engine-cli/docker_config_inspect.yaml
```diff
@@ -6,7 +6,7 @@ long: |-
   By default, this renders all results in a JSON array. If a format is specified,
   the given template will be executed for each result.
 
-  Go's [text/template](http://golang.org/pkg/text/template/) package
+  Go's [text/template](https://golang.org/pkg/text/template/) package
```
thaJeztah, line 9: The YAML files in the `_data/engine-cli` directory are generated from MarkDown source files in the CLI repository. For example, this page is generated from https://github.com/docker/cli/blob/master/docs/reference/commandline/config_inspect.md

So changes made in these YAML files will be overwritten when these files are regenerated and updated. Perhaps you're interested in opening a pull request with these changes in the CLI repository? https://github.com/docker/cli

When doing so, be sure to include a DCO sign-off line (`Signed-Off-By`) line in the commit message (see https://github.com/docker/cli/blob/master/CONTRIBUTING.md#sign-your-work), otherwise CI will fail in that repository.

I'm also happy to make those changes if you want

thaJeztah (approved):
Thanks! Changes LGTM, but left a comment about some generated YAML files that were updated (which would also need to be updated in the upstream CLI repository to prevent the changes from being reverted at some point)

PR #13696, opened 2021-10-14
Make is clearer how to update from Desktop 4.1.0 from the menu

mat007:
### Proposed changes

Make is clearer how to update from Desktop 4.1.0 from the menu


desktop/mac/release-notes/index.md
```diff
@@ -29,7 +29,7 @@ This page contains information about the new features, improvements, known issue
 
 - Fixed a bug where a `Docker.qcow2` file would be ignored on upgrade and a fresh `Docker.raw` used instead, resulting in containers and images disappearing. If a system has both files (due to the previous bug) then the most recently modified file will be used, to avoid recent containers and images disappearing again. To force the use of the old `Docker.qcow2`, delete the newer `Docker.raw` file. Fixes [docker/for-mac#5998](https://github.com/docker/for-mac/issues/5998).
```
rfay, line 30:
```suggestion
- When upgrading from 4.1.0 to 4.1.1, note that **the menu does not change to "Update and restart"** so you can just wait for the download to complete (icon changes) and then select "restart". This bug is fixed (below) in 4.1.1, for future upgrades. 
- Fixed a bug where a `Docker.qcow2` file would be ignored on upgrade and a fresh `Docker.raw` used instead, resulting in containers and images disappearing. If a system has both files (due to the previous bug) then the most recently modified file will be used, to avoid recent containers and images disappearing again. To force the use of the old `Docker.qcow2`, delete the newer `Docker.raw` file. Fixes [docker/for-mac#5998](https://github.com/docker/for-mac/issues/5998).
```

desktop/mac/release-notes/index.md
```diff
@@ -29,7 +29,7 @@ This page contains information about the new features, improvements, known issue
...
 - Fixed update notification overlay sometimes getting out of sync between the Settings button and the Software update button in the Dashboard.
-- Fixed menu entry to install a downloaded new Desktop update. When an update is ready to install `Restart` should change to `Update and restart`.
+- Fixed menu entry to install a downloaded new Desktop update. When an update is ready to install `Restart` should change to `Update and restart`. When updating from 4.1.0 you will need to select `Restart` due to this bug.
```
rfay, line 32: I recommend just stating at the top of the release note that this problem exists and explaining it separately. Added a suggestion to that effect.

desktop/mac/release-notes/index.md
usha-mandya: Should we just update to "When upgrading from 4.1.0" without specifying the version users are upgrading to?

mat007: It’s when updating from 4.1.0 to any version, as the bug is in 4.1.0.

PR #13701, opened 2021-10-15
[FIX] update documentation

Timer91:
### Proposed changes

- add `PYTHONDONTWRITEBYTECODE` environment variable
> Python will no longer write these files to disk, and your development environment will remain nice and clean.

- rename pip package `psycopg2-binary` to `psycopg2`:
> `psycopg2-binary` and `psycopg2` both give us the same code that we interact with. The difference between the two is in how that code is installed in our computer.
If we use `psycopg2-binary` we have the following issue when `docker-compose up`: 
```
$> django.db.utils.OperationalError: SCRAM authentication requires libpq version 10 or above`
```

- add environment variable to `db` in docker-compose for a better usage
- use `os.environment` in `setting.py` to get those previous variables

### DIFF
--- a/samples/django.md
+++ b/samples/django.md
@@ -75,12 +75,6 @@ and a `docker-compose.yml` file. (You can use either a `.yml` or `.yaml` extensi
        image: postgres
        volumes:
          - ./data/db:/var/lib/postgresql/data
-       environment:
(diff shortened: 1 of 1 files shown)


usha-mandya (approved):
LGTM. Thank you @Timer91

PR #13702, opened 2021-10-15
Clarify when last-modified time is considered for caching

rpygithub:
### Proposed changes

This change updates the "Leverage build cache" section of the "Dockerfile best practices" to make it explicit that the last-modified time of image file(s) is considered first, before examining file content. Otherwise, the reader might form an impression that the last-modified time of file(s) is not considered and checksums are calculated in any case. See [this comment](https://github.com/moby/buildkit/issues/1058#issuecomment-934473074) on a BuildKit issue for an example where this caused confusion.

### DIFF
--- a/develop/develop-images/dockerfile_best-practices.md
+++ b/develop/develop-images/dockerfile_best-practices.md
@@ -379,12 +379,13 @@ image. The basic rules that Docker follows are outlined below:
   of the child images is sufficient. However, certain instructions require more
   examination and explanation.
 
-- For the `ADD` and `COPY` instructions, the contents of the file(s)
-  in the image are examined and a checksum is calculated for each file.
-  The last-modified and last-accessed times of the file(s) are not considered in
-  these checksums. During the cache lookup, the checksum is compared against the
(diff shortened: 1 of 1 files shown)


crazy-max:
Ok I digged up a bit around that phrasing and the actual behavior in BuildKit vs the legacy builder. As we want to make BuildKit the default experience I'm good with this change but I think we could instead add an "info/warn block" under that paragraph about the current BuildKit caching behavior. We could also point to https://github.com/moby/buildkit/issues/1058 to explain it could be handled when implemented at frontend level.

knkarthik:
This is an important piece of information and I would very much want to see this in the docs! 

cc: @rpygithub @crazy-max

tonistiigi:
I think the current proposal is misunderstanding the behavior and the comment it links to. There are two completely separate topics: the build context upload from local directory to the daemon and then content-based cache checksumming for `ADD/COPY` commands. The first uses file metadata comparisons between build invocations for the files in the build context(Dockerfiles are always uploaded). The second uses a checksum algorithm that includes file metadata(except timestamp) and file content.

PR #13711, opened 2021-10-16
Explain the use of the `COPY` command in the NodeJS example

### LINKED ISSUE #13669: Redundant line in Node.js app Dockerfile example
[language/nodejs/build-images.md](https://docs.docker.com/language/nodejs/build-images/)

```
COPY ["package.json", "package-lock.json*", "./"]
RUN npm install --production
COPY  .  .
```

The second COPY copies the source again after the first COPY's last expression.

danBamikiya:
### Proposed changes

The explanation of the `COPY` command given in the NodeJS language guide does not correspond with the example.
The explanation says the `COPY` command takes in two parameters but goes on to use three parameters in the example without explaining that it can take in multiple `src` parameters. This is a confusing explanation/example:

docker_docs_website_issue_close_1

This PR explicitly shows that the `COPY` command takes in two parameters `src` and `dest` and there can be multiple `src` parameters but they should be separated by a comma for example: `COPY ["<src1>", "<src2>", ..., "<dest>"]`
It also explicitly shows readers that the last parameter is always the destination path avoiding confusions like #13669

### Related issues (optional)

Closes #13669

cc @usha-mandya

### DIFF
--- a/language/nodejs/build-images.md
+++ b/language/nodejs/build-images.md
@@ -130,10 +130,14 @@ WORKDIR /app
 
 Usually the very first thing you do once you’ve downloaded a project written in Node.js is to install npm packages. This ensures that your application has all its dependencies installed into the `node_modules` directory where the Node runtime will be able to find them.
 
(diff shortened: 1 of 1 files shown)


crydotsnake (dismissed):

usha-mandya (approved):
LGTM

usha-mandya:
@danBamikiya Thanks for the update. It looks good :)

PR #13717, opened 2021-10-20
docker-hub: various touch-ups and some rephrasing

thaJeztah:
- moved note about open source projects into the CTA; this reduces the number of
  notes on this page (which was already quite large), and puts related information
  about upgrading/plans in a central location.
- changed note about limits to body text, and added mention of "business" plan.
- replaced https://docs.docker.com URLs with local links
- replaced some HTML links with links to the target markdown file
- add missing YAML and "console" code-hints
- wrapped long paragraphs/lines
- corrected the information about Amazon Linux

To be looked at; the recent changes added information about using "v2" compose
files, but compose (after having switched to the compose-spec) no longer uses
versions, so perhaps this needs to be corrected (we don't have local docs for
the compose-spec format though, so that makes it a bit more complicated).

### DIFF
--- a/docker-hub/builds/advanced.md
+++ b/docker-hub/builds/advanced.md
@@ -6,7 +6,8 @@ redirect_from:
 - /docker-cloud/builds/advanced/
 ---
 
-The following options allow you to customize your automated build and automated test processes.
+The following options allow you to customize your automated build and automated
+test processes.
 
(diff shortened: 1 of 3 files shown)


thaJeztah:
@usha-mandya ptal


LOL, wanted to work on some other changes, and found that I still had this on my machine and not committed yet.

PR #13753, opened 2021-10-29
Update install.md

Ap0them:
### Proposed changes

The command I found to work, as well as what I believe the modern command is is 'pip3 install'. 'pip install' didn't work for me

### Potential Issues

The only one I've seen so far is that the version can't be identified when running 'docker-compose --version'

### DIFF
--- a/compose/install.md
+++ b/compose/install.md
@@ -190,7 +190,7 @@ tutorial](https://docs.python-guide.org/dev/virtualenvs/) to get
 started.
 
 ```console
-$ pip install docker-compose
+$ pip3 install docker-compose
 ```
 
 If you are not using virtualenv,


thaJeztah:
@ulyssessouza @StefanScherer ptal

StefanScherer (approved):
That's reasonable for me, I'm not a Python expert. But that's what I also typically do when pip doesn't work :-)

Ap0them:
> That's reasonable for me, I'm not a Python expert. But that's what I also typically do when pip doesn't work :-)
> 
> 

It's definitely a bit of a bodge, but getting Docker running on Raspberry Pi is a pain & this is the only way I've gotten it to work.

ulyssessouza:
docker-compose V1 doesn't support python 2.x anymore. So if python version is 2.x we have to be explicit on it

PR #13781, opened 2021-11-03
Fix typo in HTTP proxy URL

mat007:
It’s HTTP here, not HTTPS. Besides the rest of the text has `http://proxy.example.com:3128`

### DIFF
--- a/desktop/windows/index.md
+++ b/desktop/windows/index.md
@@ -155,7 +155,7 @@ If you wish to set the proxy settings for your containers, you need to define
 environment variables for them, just like you would do on Linux, for example:
 
 ```ps
-> docker run -e HTTP_PROXY=https://proxy.example.com:3128 alpine env
+> docker run -e HTTP_PROXY=http://proxy.example.com:3128 alpine env
 
 PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
 HOSTNAME=b7edf988b2b5


djs55:
There might be a capitalization problem as well:
```
PS >  docker run -e HTTP_PROXY=http://127.0.0.1:3128 alpine wget http://www.docker.com/
Connecting to www.docker.com (143.204.198.101:80)
Connecting to www.docker.com (143.204.198.101:443)
saving to 'index.html'
index.html           100% |********************************| 98.6k  0:00:00 ETA
'index.html' saved

PS >  docker run -e http_proxy=http://127.0.0.1:3128 alpine wget http://www.docker.com/
Connecting to 127.0.0.1:3128 (127.0.0.1:3128)
wget: can't connect to remote host (127.0.0.1): Connection refused
```

mat007:
> There might be a capitalization problem as well:

It seems the docs are correct then? It’s `HTTP_PROXY`?
(Golang supports both but I think the Docker world mostly supports upper case, for some reason…)

djs55:
@mat007 I should have clarified: I wasn't running a proxy, so "Connection refused" was success :) `wget` ignored `HTTP_PROXY` but it respected `http_proxy` , so I suspect the "standard" capitalization is lower-case.

mat007:
Oh, right. I see what you mean now.
Yeah there is no real consensus on what the capitalization should be [it](https://unix.stackexchange.com/q/212894) [seems ](https://superuser.com/q/944958), but if `wget` only knows about the lowercase version then indeed the example must use that.

I’ll amend, thanks!

mat007:
Oh wait, the docs actually never mention `wget`… 🤔

usha-mandya:
We also have https://github.com/docker/docker.github.io/pull/13786 from Sebastiaan on Proxy settings

usha-mandya (approved):
LGTM

PR #8352, opened 2019-02-26
Update proxy.md

paigehargrave:
#6450

PR #13786, opened 2021-11-04
Rewrite section on configuring proxies for containers

thaJeztah:
- related: https://github.com/docker/docs/pull/8352

### DIFF
--- a/config/daemon/systemd.md
+++ b/config/daemon/systemd.md
@@ -1,6 +1,6 @@
 ---
 description: Controlling and configuring Docker using systemd
-keywords: docker, daemon, systemd, configuration
+keywords: dockerd, daemon, systemd, configuration, proxy, networking
 redirect_from:
   - /articles/host_integration/
   - /articles/systemd/
@@ -25,17 +25,34 @@ Docker with systemd. For this, install the two unit files (`service` and
 [the github repository](https://github.com/moby/moby/tree/master/contrib/init/systemd)
 to `/etc/systemd/system`.
 
-## HTTP/HTTPS proxy
+### Configure the Docker daemon to use a proxy server {#httphttps-proxy}
 
-The Docker daemon uses the `HTTP_PROXY`, `HTTPS_PROXY`, and `NO_PROXY`
-environmental variables in its start-up environment to configure HTTP or HTTPS
-proxy behavior. You can't configure these environment variables using the
-`daemon.json` file.
+The Docker daemon uses the following environment variables in
+its start-up environment to configure HTTP or HTTPS proxy behavior:
 
-This example overrides the default `docker.service` file.
+- `HTTP_PROXY`
+- `http_proxy`
+- `HTTPS_PROXY`
+- `https_proxy`
+- `NO_PROXY`
+- `no_proxy`
+
+In Docker Engine version 23.0 and later versions, you may also configure proxy
+behavior for the daemon in the [`daemon.json` file](./index.md#configure-the-docker-daemon):
+
+```json
+{
+  "http-proxy": "http://proxy.example.com:3128",
+  "https-proxy": "https://proxy.example.com:3129",
+  "no-proxy": "*.test.example.com,.example.org,127.0.0.0/8"
+}
+```
+
+These configurations override the default `docker.service` systemd file.
 
 If you are behind an HTTP or HTTPS proxy server, for example in corporate
-settings, you need to add this configuration in the Docker systemd service file.
+settings, the daemon proxy configurations must be specified in the systemd
+service file, not in the `daemon.json` file or using environment variables.
 
 > **Note for rootless mode**
 >
@@ -65,7 +82,7 @@ settings, you need to add this configuration in the Docker systemd service file.
 
    ```systemd
    [Service]
-   Environment="HTTP_PROXY=http://proxy.example.com:80"
+   Environment="HTTP_PROXY=http://proxy.example.com:3128"
    ```
 
    If you are behind an HTTPS proxy server, set the `HTTPS_PROXY` environment
@@ -73,7 +90,7 @@ settings, you need to add this configuration in the Docker systemd service file.
 
    ```systemd
    [Service]
-   Environment="HTTPS_PROXY=https://proxy.example.com:443"
+   Environment="HTTPS_PROXY=https://proxy.example.com:3129"
    ```
 
    Multiple environment variables can be set; to set both a non-HTTPS and a
@@ -81,8 +98,8 @@ settings, you need to add this configuration in the Docker systemd service file.
 
    ```systemd
    [Service]
-   Environment="HTTP_PROXY=http://proxy.example.com:80"
-   Environment="HTTPS_PROXY=https://proxy.example.com:443"
+   Environment="HTTP_PROXY=http://proxy.example.com:3128"
+   Environment="HTTPS_PROXY=https://proxy.example.com:3129"
    ```
 
    > **Note**
@@ -92,7 +109,7 @@ settings, you need to add this configuration in the Docker systemd service file.
    >
    > ```
    > [Service]
-   > Environment="HTTP_PROXY=http://domain%%5Cuser:complex%%23pass@proxy.example.com:8080/"
+   > Environment="HTTP_PROXY=http://domain%%5Cuser:complex%%23pass@proxy.example.com:3128/"
    > ```
 
 3. If you have internal Docker registries that you need to contact without
@@ -117,8 +134,8 @@ settings, you need to add this configuration in the Docker systemd service file.
 
    ```systemd
    [Service]
-   Environment="HTTP_PROXY=http://proxy.example.com:80"
-   Environment="HTTPS_PROXY=https://proxy.example.com:443"
+   Environment="HTTP_PROXY=http://proxy.example.com:3128"
+   Environment="HTTPS_PROXY=https://proxy.example.com:3129"
    Environment="NO_PROXY=localhost,127.0.0.1,docker-registry.example.com,.corp"
    ```
 
@@ -135,7 +152,7 @@ settings, you need to add this configuration in the Docker systemd service file.
    ```console
    $ sudo systemctl show --property=Environment docker
 
-   Environment=HTTP_PROXY=http://proxy.example.com:80 HTTPS_PROXY=https://proxy.example.com:443 NO_PROXY=localhost,127.0.0.1,docker-registry.example.com,.corp
+   Environment=HTTP_PROXY=http://proxy.example.com:3128 HTTPS_PROXY=https://proxy.example.com:3129 NO_PROXY=localhost,127.0.0.1,docker-registry.example.com,.corp
    ```
 
 </div>
@@ -152,7 +169,7 @@ settings, you need to add this configuration in the Docker systemd service file.
 
    ```systemd
    [Service]
-   Environment="HTTP_PROXY=http://proxy.example.com:80"
+   Environment="HTTP_PROXY=http://proxy.example.com:3128"
    ```
 
    If you are behind an HTTPS proxy server, set the `HTTPS_PROXY` environment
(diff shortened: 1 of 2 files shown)


thaJeztah:
Just pushing this as a draft; this is _very_ unfinished (need to rewrite the last section and also mention upper/lowercase etc)

network/proxy.md
```diff
@@ -4,44 +4,49 @@ description: How to configure the Docker client to use a proxy server
...
-     "proxies":
-     {
-       "default":
-       {
+     "proxies": {
+       "default": {
          "httpProxy": "http://192.168.1.12:3128",
-         "httpsProxy": "http://192.168.1.12:3128",
-         "noProxy": "*.test.example.com,.example2.com,127.0.0.0/8"
+         "httpsProxy": "https://192.168.1.12:3129",
```
mat007, line 40: I think it is supposed to be `http` here, quoting @djs55:

> Confusingly a HTTPS proxy is a HTTP server which accepts CONNECT host:port HTTP requests and establishes a tunnel, which the client then sends TLS back and forth but the proxy doesn't understand (in some ways this is easier than in the HTTP case where the proxy has a bunch of rules about what to do with headers)

network/proxy.md
```diff
@@ -50,15 +55,32 @@ configure it in different ways:
...
+
+ 
+This method makes the image less portable, so if you have Docker 17.07
 or higher, you should [configure the Docker client](#configure-the-docker-client)
 instead.
 
-| Variable      | Dockerfile example                                | `docker run` example                                |
-|:--------------|:--------------------------------------------------|:----------------------------------------------------|
-| `HTTP_PROXY`  | `ENV HTTP_PROXY="http://192.168.1.12:3128"`          | `--env HTTP_PROXY="http://192.168.1.12:3128"`          |
-| `HTTPS_PROXY` | `ENV HTTPS_PROXY="https://192.168.1.12:3128"`        | `--env HTTPS_PROXY="https://192.168.1.12:3128"`        |
```
mat007, line 62: Ah that one had `https` though… 🤔

network/proxy.md
thaJeztah: I think both should work, although it _could_ depend on the implementation.

It's really a "fuzzy" area, without any (good) standards. e.g. Golang also made a silly (and breaking) change and started to ignore `HTTP_PROXY` for https connections; https://docs.docker.com/engine/release-notes/#201010

thaJeztah:
I need to collect more changes and put them together, e.g. https://github.com/docker/docker.github.io/pull/8352 and https://github.com/docker/docker.github.io/pull/12463 (which also isn't correct, as "it depends")

And describe that handling of `NO_PROXY` is quite different (depending on what software uses it), same as for upper/lowercase [cURL](https://curl.se/libcurl/c/CURLOPT_NOPROXY.html), [wget](https://www.gnu.org/software/wget/manual/html_node/Proxies.html), and [Golang](https://pkg.go.dev/golang.org/x/net/http/httpproxy) don't agree on those.

network/proxy.md
djs55: I think
- the `httpProxy` (`HTTP_PROXY`) will be used to proxy http:// connections
- the `httpsProxy` (`HTTPS_PROXY`) will be used to proxy https:// connections

where the connection to the proxy can either be plaintext `http://proxy:3128` or itself using TLS `https://proxy:3128`. So there can be 0, 1 or 2 layers of TLS.

If the underlying connection is TLS, then the only reason to use TLS to talk to the proxy too is to protect the authorization header which could include a base64-encoded password. The destination server is sent in the clear even with TLS (as a Server Name Indication) so the `HTTP CONNECT <server:443>` part is not secret.

So I think the examples are ok. My main problem with the docs as they are is that there are a confusing number of places where proxies can be configured. I think this PR makes an improvement though!

usha-mandya:
@thaJeztah Is this PR still relevant? Thanks.

thaJeztah:
Reopening for now (still in draft), but still something to finish.

dvdksn:
@thaJeztah I continued on some of the changes you started in this PR. PTAL and let me know if there's more stuff that we should add/change in the scope of this

thaJeztah:
Oh! Thanks, sorry forgot about this one; gonna give this one a look Today.

I see there's one link failure; perhaps we need a custom id / anchor somewhere;

```
========================================================================
engine/reference/commandline/cli/index.html
  hash does not exist --- engine/reference/commandline/cli/index.html --> /config/daemon/systemd/#httphttps-proxy
engine/reference/commandline/dockerd/index.html
  hash does not exist --- engine/reference/commandline/dockerd/index.html --> /config/daemon/systemd/#httphttps-proxy
========================================================================
```

dvdksn:
Fixed the broken link upstream in docker/cli#4249 - will need to backport that one

network/proxy.md
```diff
@@ -6,68 +6,154 @@ keywords: network, networking, proxy, client
...
+the client connects to. To configure the proxies for individual daemons,
+specify the proxy configuration under `proxies.<hostname>` in the client
+configuration file, where hostname is the address of the daemon.
+
+For example:
+
+```json
+{
+ "proxies": {
+   "https://docker-daemon1.example.com": {
```
thaJeztah, line 112: I _think_ this one needs to be `tcp://docker-daemon1.example.com` (but would have to verify that).

Do you think (for a follow-up?) it would be useful to show an example as well where both "default" and "other is configured?

```bash
cat ~/.docker/config.json
{
	"proxies": {
		"default": {
			"httpProxy": "http://192.168.1.12:1111"
		},
		"unix:///Users/thajeztah/.docker/run/docker.sock": {
			"httpProxy": "http://192.168.1.12:2222"
		}
	}
}
```

And we could show examples, including how it interacts with `docker context`;

```bash
docker context ls
NAME                TYPE                DESCRIPTION                               DOCKER ENDPOINT                                   KUBERNETES ENDPOINT   ORCHESTRATOR
default *           moby                Current DOCKER_HOST based configuration   unix:///var/run/docker.sock
desktop-linux       moby                                                          unix:///Users/thajeztah/.docker/run/docker.sock


# uses "default" (if set)
docker run --rm  alpine sh -c 'env | grep _PROXY'
HTTP_PROXY=http://192.168.1.12:1111

# uses proxy for "unix:///Users/thajeztah/.docker/run/docker.sock"
docker --host unix:///Users/thajeztah/.docker/run/docker.sock run --rm  alpine sh -c 'env | grep _PROXY'
HTTP_PROXY=http://192.168.1.12:2222

# also uses proxy for "unix:///Users/thajeztah/.docker/run/docker.sock"
docker --context=desktop-linux run --rm  alpine sh -c 'env | grep _PROXY'
HTTP_PROXY=http://192.168.1.12:2222
```


(and, yes, we should integrate that with `docker context` itself, and have it be part of the context's options)


☝️ I should mention that it looks like there's currently a bug in the CLI, and setting a "per-host" proxy for a `ssh://` connection doesn't work (it tries to look up using our "magic" `http://docker.example.com` hostname, which is used internally.

network/proxy.md
```diff
@@ -6,68 +6,154 @@ keywords: network, networking, proxy, client
...
-
-| Variable      | Dockerfile example                                | `docker run` example                                |
-|:--------------|:--------------------------------------------------|:----------------------------------------------------|
-| `HTTP_PROXY`  | `ENV HTTP_PROXY="http://192.168.1.12:3128"`          | `--env HTTP_PROXY="http://192.168.1.12:3128"`          |
-| `HTTPS_PROXY` | `ENV HTTPS_PROXY="https://192.168.1.12:3128"`        | `--env HTTPS_PROXY="https://192.168.1.12:3128"`        |
-| `FTP_PROXY`   | `ENV FTP_PROXY="ftp://192.168.1.12:3128"`            | `--env FTP_PROXY="ftp://192.168.1.12:3128"`            |
-| `NO_PROXY`    | `ENV NO_PROXY="*.test.example.com,.example2.com"` | `--env NO_PROXY="*.test.example.com,.example2.com"` |
+You can configure global proxy configurations for builds and containers using
+the Docker client configuration file, located in `~/.docker/config.json`.
+Builds and containers automatically inherit this configuration.
```
thaJeztah, line 30: Do we need a link to the "per-daemon" config (for those reading this that may need different configs?) or would that be clear enough because it's on the same page?

network/proxy.md
```diff
@@ -6,68 +6,154 @@ keywords: network, networking, proxy, client
...
+
+```dockerfile
+FROM alpine
+ARG HTTPS_PROXY
+RUN apk update    # uses httpsProxy from ~/.docker/config.json
+```
+
+### Configure proxy settings per daemon
+
+The `proxies.default` key configures the proxy settings for all daemons that
```
thaJeztah, line 102: Slightly wondering if `proxies.default` could lead to users using that as a literal key. Perhaps we should look at slightly different phrasing.

network/proxy.md
```diff
@@ -6,68 +6,154 @@ keywords: network, networking, proxy, client
...
+Embedding proxy settings in images also poses a security risk, as the values
+may include sensitive information.
+
+## About proxy environment variables
+
+There's no standard that defines how web clients should handle proxy 
+environment variables, or the format for defining them. If you're interested
+in learning more about the history of these variables, check out this blog post
+on the subject, written by the GitLab team:
+[We need to talk: Can we standardize NO_PROXY?](https://about.gitlab.com/blog/2021/01/27/we-need-to-talk-no-proxy/){: target="_blank" rel="noopener"}.
```
thaJeztah, line 159: Should this be more at the start of the page? I also see that some part about the `noProxy` format (wildcard etc) was removed. Perhaps that's ok because there is no common standard 🤔 (just thinking if we can give some guidelines somehow)

config/daemon/systemd.md
```diff
@@ -25,17 +25,34 @@ Docker with systemd. For this, install the two unit files (`service` and
 [the github repository](https://github.com/moby/moby/tree/master/contrib/init/systemd)
 to `/etc/systemd/system`.
 
-## HTTP/HTTPS proxy
+### Configure the Docker daemon to use a proxy server
```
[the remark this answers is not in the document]
thaJeztah: @dvdksn ☝️ this could be a good alternative (also because there may be links "in the wild" pointing to this section)

config/daemon/systemd.md
```diff
@@ -25,17 +25,34 @@ Docker with systemd. For this, install the two unit files (`service` and
...
-This example overrides the default `docker.service` file.
+- `HTTP_PROXY`
+- `http_proxy`
+- `HTTPS_PROXY`
+- `https_proxy`
+- `NO_PROXY`
+- `no_proxy`
+
+In Docker Engine version 23.0 and later versions, you may also configure proxy
+behavior for the daemon using the using the `daemon.json` file:
```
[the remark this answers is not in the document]
dvdksn: oh no, not again, the horror of trying to describe where to edit this file depending on your installation :D

config/daemon/systemd.md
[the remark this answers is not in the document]
dvdksn: Yes, I'll use this.

config/daemon/systemd.md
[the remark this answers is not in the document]
dvdksn: not yet... need to revisit

network/proxy.md
dvdksn: That's a cool example. Let's add it in a follow-up - for now I can try to clarify the confusion with the key names, and maybe add both a default and a "named" daemon.

TBH, I find **contexts** a confusing topic overall. Maybe that's an area worth spending some time trying to clarify.

network/proxy.md
dvdksn: I'll remove the word "global" as it doesn't really make sense in the way I thought about it. It's per client, and/or per client+context/daemon. But I feel it's a bit too complicated to describe these variations here, I'm just gonna refer to it as client config.

network/proxy.md
dvdksn: Good call, I'll try moving it up on the page. It's a nice callout to have as an intro.

thaJeztah (commented):
LGTM ❤️ 

(I can't approve the PR because I opened it, but if you can LGTM it @dvdksn, then we can merge)

dvdksn (approved):

PR #13787, opened 2021-11-04
Note the max-retries option on the Start containers automatically page.

hawicz:
### Proposed changes

Include a mention of "max-retries" on https://docs.docker.com/config/containers/start-containers-automatically/ so it matches https://docs.docker.com/engine/reference/run/#restart-policies---restart

n/a

### Related issues (optional)

n/a


usha-mandya (approved):
LGTM. Thank you @hawicz

PR #13791, opened 2021-11-05
Made it easy to identify the video in landing page for the editors in HTML.

soubhikmandal2000:
### Proposed changes

Made it easy to identify the video on the landing page for the editors in HTML.

### DIFF
--- a/_includes/landing-page/popular-videos.html
+++ b/_includes/landing-page/popular-videos.html
@@ -6,6 +6,7 @@
         <p>Watch the most popular videos from DockerCon.</p>
       </div>
     </div>
+    <!-- How to Get Started with Docker -->
     <div class="row">
       <div class="col-xs-12 col-md-6 col-lg-12">
         <a class="video-item" href="https://youtu.be/iqqDU2crIEQ?t=30" target="_blank" rel="noopener">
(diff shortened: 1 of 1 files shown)


_includes/landing-page/popular-videos.html
```diff
@@ -6,19 +6,22 @@ <h5 class="title">Popular videos</h5>
         <p>Watch the most popular videos from DockerCon.</p>
       </div>
     </div>
+    <!-- How to Get Started with Docker -->
     <div class="row">
       <div class="col-xs-12 col-md-6 col-lg-12">
         <a class="video-item" href="https://youtu.be/iqqDU2crIEQ?t=30" target="_blank" rel="noopener">
           <img class="video-image" src="/images/video.svg" alt="video: How to get started with Docker" width="65" height="44" />
           <div class="video-title">How to get started with Docker</div>
```
thaJeztah, line 14: Hmm.. not sure if this adds much value; the same text is already a few lines below the comment, and adding these comments will actually introduce more work to update them together with the actual HTML.

Screenshot 2021-11-11 at 15 20 28

thaJeztah (changes requested):
thanks for contributing; I don't think we need this (left a comment inline), so I'll close this PR, but thanks again for contributing; it's appreciated 🤗

PR #13796, opened 2021-11-05
Fix typo

seungwonleee:
### Proposed changes

Fixed typo.

- Language-specific guides / Node.js / Run containers document
The problem is that the forwarding json data and the response json data are not properly distinguished and visible.

- Language-specific guides / Node.js / Run your tests document
The requested json data is seen as json response result data.

n/a

### Related issues (optional)

n/a


thaJeztah (approved):
LGTM, thanks!

PR #13797, opened 2021-11-05
Check if site_id is empty and skip delete command

StefanScherer:

### Proposed changes

Fix a case when the netlify site is already deleted. Fixes this situation: https://github.com/docker/docker.github.io/runs/4119974208?check_suite_focus=true
If no `site_id` could be retrieved, skip the sites:delete command.


usha-mandya (approved):
LGTM

PR #13801, opened 2021-11-07
chore: silent completion and avoid manual version update

froblesmartin:
### Proposed changes

Avoid showing the curl download info on every shell start.

As well, avoid having to manually update the docker compose version and use the current installed version

Signed-off-by: Francisco Robles Martín <f.robles.martin@pm.me>

### DIFF
--- a/compose/completion.md
+++ b/compose/completion.md
@@ -21,8 +21,8 @@ available.
 2. Place the completion script in `/etc/bash_completion.d/`.
 
    ```console
-   $ sudo curl \
-       -L https://raw.githubusercontent.com/docker/compose/{{site.compose_version}}/contrib/completion/bash/docker-compose \
+   $ sudo curl -s \
+       -L https://raw.githubusercontent.com/docker/compose/$(docker-compose version --short)/contrib/completion/bash/docker-compose \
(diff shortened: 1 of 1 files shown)


froblesmartin:
Squash this PR when merging.

thaJeztah (changes requested):
I don't think we should make these changes;

> Avoid showing the curl download info on every shell start.

The instructions show how to download the file _once_ (for the current version of compose), not to put the whole command in your bash profile (which would download it every time).

Given that the intent is to download it once, I don't think we should remove the progress output (it's useful for the user to see the file is downloaded).

As to the `docker-compose --version`; we use the `{{site.compose_version}}` option in our docs to set the version of compose we currently ship; with the migration of `docker-compose` to `docker compose` (v2), I think we should avoid potential errors for users that use v2, but haven't enabled the `docker-compose` alias.

froblesmartin:
> The instructions show how to download the file _once_ (for the current version of compose), not to put the whole command in your bash profile (which would download it every time).
> 
> Given that the intent is to download it once, I don't think we should remove the progress output (it's useful for the user to see the file is downloaded).

Oh yeah, I did understand it wrong when following it.

PR #13813, opened 2021-11-09
Update description of -m option

onkarjanwa:
### Proposed changes

The container memory uses does not go beyond the limit set to `-m` option. So I believe the description should be "If you set this option, the maximum allowed value is 6m (6 megabyte)." instead of "If you set this option, the minimum allowed value is 6m (6 megabyte).".


config/containers/resource_constraints.md
```diff
@@ -77,7 +77,7 @@ Most of these options take a positive integer, followed by a suffix of `b`, `k`,
...
-| `-m` or `--memory=`    | The maximum amount of memory the container can use. If you set this option, the minimum allowed value is `6m` (6 megabyte).                                                                                                                                                                                                                                                                      |
+| `-m` or `--memory=`    | The maximum amount of memory the container can use. If you set this option, the maximum allowed value is `6m` (6 megabyte).                                                                                                                                                                                                                                                                      |
```
thaJeztah, line 80: The intent was to describe that values lower than `6m` are not allowed (e.g., setting `--memory=1m` will produce an error).

```bash
docker run --rm --memory=1m alpine

docker: Error response from daemon: Minimum memory limit allowed is 6MB.
See 'docker run --help'.
```


@usha-mandya perhaps you have a suggestion to better phrase that?

usha-mandya: @thaJeztah  I think the text "minimum allowed value" is pretty self-explanatory. Perhaps we could clarify by adding  "If you use this option, set the value to at least 6m (6 megabytes)."

usha-mandya (approved):

PR #13826, opened 2021-11-10
Fix python3 dependency

StefanScherer:

### Proposed changes

Fixes a problem in the getting started guide, alpine has made it more clear which version of Python you want to install. So we have to use `python3` instead of `python`.

Related fix in the docker/getting-started repo is https://github.com/docker/getting-started/pull/218


thaJeztah (approved):
LGTM

Issue #13547, opened 2021-09-16
Ubuntu installation documentation is out of date, at least for Linux Mint


jcald1:

thaJeztah:
Thanks for reporting; yes, we were planning to remove those instructions, as they are outdated, and we don't test on Ubuntu and Debian derivatives, so probably better to remove the instructions than to keep them and provide incorrect information.

PR #13837, opened 2021-11-12
engine: remove outdated instructions for derivatives

thaJeztah:
closes https://github.com/docker/docker.github.io/issues/13547

The information about which distro version to use to match upstream
Debian / Ubuntu versions was outdated. Given that Docker does not
officially support, nor tests, installing these packages on Ubuntu
and Debian derivatives, it's better to remove it, and leave it to
the user to find the correct codename; also see
https://docs.docker.com/engine/install/#other-linux-distributions


usha-mandya (approved):
LGTM

PR #13882, opened 2021-11-23
Updated Wordpress Sample Markdown page

hkbertoson:
### Proposed changes

Added a tip for users that are on an M1 mac to add ```platform: linux/amd64``` underneath image of mysql:5.7
This is a quality of life change.

### DIFF
--- a/samples/wordpress.md
+++ b/samples/wordpress.md
@@ -36,6 +36,8 @@ Compose to set up and run WordPress. Before starting, make sure you have
 3.  Create a `docker-compose.yml` file that starts your
     `WordPress` blog and a separate `MySQL` instance with volume
     mounts for data persistence:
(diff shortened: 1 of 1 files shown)


StefanScherer:
Thanks @hkbertoson for the suggestion.
Indeed, it's a common problem to run `mysql` image on Apple Silicon machines.
We've discussed this topic and would go a different way to adjust the sample to use `mariadb` image which supports both `amd64` and `arm64` CPU's.

Please also have a look at the **Awesome Compose** repo where we have a working example to run Wordpress with MariaDB: https://github.com/docker/awesome-compose/tree/master/wordpress-mysql

hkbertoson:
Okay. Sounds great! Thank you so much. :)

PR #13889, opened 2021-11-25
Fixed typo

mat007:
Compose Cli is different than Compose, I don’t think we usually mention the former in release notes.


usha-mandya (approved):
LGTM

PR #13924, opened 2021-12-01
Clarify Advanced Wordpress section

jguady:
I believe this is the intended meaning, otherwise it is unclear where the name wp_db_secret is coming from when referenced as "keeping the same" by wp_db_password


usha-mandya (approved):
LGTM

usha-mandya (approved):
LGTM. Thank you

PR #13954, opened 2021-12-10
Publish updates from master

usha-mandya:

### DIFF
--- a/_data/glossary.yaml
+++ b/_data/glossary.yaml
@@ -1,10 +1,9 @@
 ﻿amd64: |
   AMD64 is AMD's 64-bit extension of Intel's x86 architecture, and is also
   referred to as x86_64 (or x86-64).
-aufs: |
-  aufs (advanced multi layered unification filesystem) is a Linux [filesystem](#filesystem) that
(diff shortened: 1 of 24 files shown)


_data/glossary.yaml
```diff
@@ -134,9 +132,9 @@ filesystem: |
 
   Examples :
 
-  - Linux : ext4, aufs, btrfs, zfs
+  - Linux : overlay2, extfs, btrfs, zfs
```
thaJeztah, line 135: `overlay2` is probably wrong in this context, as that's the name of a storage driver in Docker Engine, but the name in the kernel would be `overlayfs` 🤔

thaJeztah (approved):
LGTM (left one comment for a follow-up)

_data/glossary.yaml
StefanScherer: That‘s my fault 😅

thaJeztah: no worries! we need to consider what we want to do with the glossary altogether; I guess it's not viewed too often

PR #13956, opened 2021-12-10
Clarify which image architecture is run by default

nathanielks:
### Proposed changes

Given the list of architectures supported by the busybox image, it's confusing to reference "the `x86_64` variant is pulled and run" when there's no platform architecture that matches that name. Instead, changing the documentation to reference `amd64` makes more sense as there is no `linux/x86_64` architecture that's supported by Docker (AFAIK), but there _is_ `linux/amd64`.


usha-mandya (approved):
LGTM. Thank you :)

PR #14002, opened 2021-12-17
List features that don't work in air-gapped envs

usha-mandya:

Add an FAQ about features that don't work when Docker Desktop is offline


StefanScherer (approved):
LGTM

PR #14004, opened 2021-12-18
The "Edge" Channel has been added

f1linux:
I added "Edge" to channels.  Its' omission was causing confusion.  "Edge" was referenced in some documents, but not others leaving the reader to check this page and seeing it missing start Googling.

### DIFF
--- a/engine/install/index.md
+++ b/engine/install/index.md
@@ -94,10 +94,11 @@ These binaries are statically linked and can be used on any Linux distribution.
 
 ## Release channels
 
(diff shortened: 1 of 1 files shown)


thaJeztah:
Thanks for contributing; Edge releases were deprecated Two/Three years ago, and no longer receive updates, so we removed mentions of that channel from our docs.

I'll close this PR because of the above, but feel free to continue the conversation

f1linux:
Thanks for clearing up the confusion @thaJeztah -

Issue #13998, opened 2021-12-17
The docker run example in the GELF logging page has a typo


ItsCollinT:
[config/containers/logging/gelf.md](https://docs.docker.com/config/containers/logging/gelf/)

The sample is:

 docker run \
      --log-driver gelf –-log-opt gelf-address=udp://1.2.3.4:12201 \
      alpine echo hello world

but the first dash before log-opt is an incorrect character.

thaJeztah:
Nice catch! Looks like the first hyphen is an n-dash, instead of a regular hyphen; I opened https://github.com/docker/docker.github.io/pull/14013 to fix

PR #14013, opened 2021-12-20
logging/gelf: fix stray 'n-dash' instead of hyphen in example

thaJeztah:
fixes https://github.com/docker/docker.github.io/issues/13998


thaJeztah:
@StefanScherer ptal

StefanScherer (approved):
Nice catch!

thaJeztah:
Yes, these are easy to miss (and we did) 🤣
It was reported by someone; after that it was an easy find; also searched the rest of the files, but looks like this was the only one

Issue #13984, opened 2021-12-15
[seccomps] - Blocked syscall table in documentation differs from actual default.json reference code 


tonkapango:
[engine/security/seccomp.md](https://docs.docker.com/engine/security/seccomp/)

The syscalls that are blocked in the "Significant syscalls blocked by the default profile" are outdated as related to the default reference provided just above the table: here: 
 https://github.com/moby/moby/blob/master/profiles/seccomp/default.json

The recommendation is to update the table to reflect the actual syscalls that are blocked.

thanks!

thaJeztah:
Are there specific syscalls you were looking at? Note that https://github.com/moby/moby/blob/master/profiles/seccomp/default.json is the `master` branch; the current (20.10.x) release of the engine will use the profile from https://github.com/moby/moby/blob/20.10/profiles/seccomp/default.json, which may differ from that.

tonkapango:
thanks thaJeztah.  for example, clock_adjtime is called out as blocked by default profile, yet it is allowed in default profile.  (due to the kernel code actually checking the sys_time capability before allowing time modification).

others like clone, ioperm, kcmp, etc are not "effectively blocked" but are allowed given the presence of a certain capability.   Or others, eg. personality, are allowed given certain parameter values in the call.

thaJeztah:
Ah, yes, looks like the `clock_adjtime` sys call was added to the allow-list in https://github.com/moby/moby/pull/40929, but not updated in the docs in this repository.

> others like clone, ioperm, kcmp, etc are not "effectively blocked" but are allowed given the presence of a certain capability. Or others, eg. personality, are allowed given certain parameter values in the call.

Yes, I think the original intent was to keep the description limited to just the things that are blocked by "default", but some rules were made conditional based on kernel version or capabilities (e.g. https://github.com/moby/moby/pull/25567) to make the profile easier to use.

Perhaps it would make sense to add a section describing the "conditional" rules, and mention what capability "gates" them.


Looking at the page, some quick thoughts on other changes that could be made on that page;

- remove the note about "seccomp 2.2.1 which is not available on Ubuntu 14.04, Debian Wheezy, or Debian Jessie", because all of those distro versions are now EOL, and no longer supported by Docker Engine.
- update the "Run without the default seccomp profile" section to not use the `debian:jessy` image (EOL as mentioned)
- I think it may be good to also mention that seccomp (as well as other protection) is _disabled_ when running a container with `--privileged`
[...]

PR #14020, opened 2021-12-20
updated docs per issue thread 13984

tonkapango:
see thread here: https://github.com/docker/docker.github.io/issues/13984

### Proposed changes

per discussion on issue 13984, we noticed some discrepancies between blocked syscalls in the document page and what the code actually was blocking.   thaJeztah made a few more recommendations for updates.  So the following was changed:

1.  Note: seccomp profiles require seccomp 2.2.1 or later.
2. docker run --rm -it --security-opt seccomp=unconfined debian:bullseye
3. Note: seccomp is _disabled_ when contain the container is running with --privilege
4. conditional table was added
5. updated the counts to reflect what is in the default profile code. 

### Related issues (optional)
#13984

### DIFF
--- a/engine/security/seccomp.md
+++ b/engine/security/seccomp.md
@@ -18,17 +18,14 @@ $ grep CONFIG_SECCOMP= /boot/config-$(uname -r)
 CONFIG_SECCOMP=y
 ```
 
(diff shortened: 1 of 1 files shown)


usha-mandya:
@tonkapango The PR is still in the draft stage. Could you rebase the PR and let us know if this is ready for review?

PR #14030, opened 2021-12-22
Include the `vscode` explanation in dev envs

felipecruz91:
### Proposed changes

Update docs to include the `vscode` user in the Dockerfile section.

/cc @glours @rumpl @gtardif @chris-crone

### DIFF
--- a/desktop/dev-environments.md
+++ b/desktop/dev-environments.md
@@ -258,10 +258,13 @@ Next, you have to define the dependencies you want to include in your `Dockerfil
 While any images or Dockerfiles will include a non-root user with a UID/GID of `1000`, many base images and Dockerfiles do not. Fortunately, you can add a non-root user named `vscode`. If you were to include the Docker tooling (e.g. `docker` cli, `docker compose`, etc.) in the `Dockerfile.devenv`, you would need the `vscode` user to be included in the `docker` group. Finally, remember to set `sleep infinity` in the entrypoint to keep the dev container running indefinitely.
 
 ```dockerfile
+# syntax=docker/dockerfile:1
+
 FROM <your base image>
 
-RUN useradd -s /bin/bash -m vscode
-RUN groupadd docker && usermod -aG docker vscode
+RUN useradd -s /bin/bash -m vscode \
+ && groupadd docker \
+ && usermod -aG docker vscode
 
 USER vscode
 
--- a/desktop/dev-environments.md
+++ b/desktop/dev-environments.md
@@ -267,8 +267,6 @@ RUN useradd -s /bin/bash -m vscode \
  && usermod -aG docker vscode
 
 USER vscode
-
-ENTRYPOINT ["sleep", "infinity"]
 ```
 
 ## Specify a base image
--- a/desktop/dev-environments.md
+++ b/desktop/dev-environments.md
@@ -255,7 +255,7 @@ In this preview, Dev Environments support a JSON file which allows you to specif
 
 Next, you have to define the dependencies you want to include in your `Dockerfile.devenv`, alongside the following requisites:
 
-While some images or Dockerfiles will include a non-root user, many base images and Dockerfiles do not. Fortunately, you can add a non-root user named `vscode`. If you were to include the Docker tooling (e.g. `docker` cli, `docker compose`, etc.) in the `Dockerfile.devenv`, you would need the `vscode` user to be included in the `docker` group. Finally, remember to set `sleep infinity` in the entrypoint to keep the dev container running indefinitely.
+While some images or Dockerfiles will include a non-root user, many base images and Dockerfiles do not. Fortunately, you can add a non-root user named `vscode`. If you were to include the Docker tooling (e.g. `docker` cli, `docker compose`, etc.) in the `Dockerfile.devenv`, you would need the `vscode` user to be included in the `docker` group.
 
 ```dockerfile
 # syntax=docker/dockerfile:1
--- a/desktop/dev-environments.md
+++ b/desktop/dev-environments.md
@@ -255,7 +255,7 @@ In this preview, Dev Environments support a JSON file which allows you to specif
 
 Next, you have to define the dependencies you want to include in your `Dockerfile.devenv`, alongside the following requisites:
 
-While any images or Dockerfiles will include a non-root user with a UID/GID of `1000`, many base images and Dockerfiles do not. Fortunately, you can add a non-root user named `vscode`. If you were to include the Docker tooling (e.g. `docker` cli, `docker compose`, etc.) in the `Dockerfile.devenv`, you would need the `vscode` user to be included in the `docker` group. Finally, remember to set `sleep infinity` in the entrypoint to keep the dev container running indefinitely.
+While some images or Dockerfiles will include a non-root user, many base images and Dockerfiles do not. Fortunately, you can add a non-root user named `vscode`. If you were to include the Docker tooling (e.g. `docker` cli, `docker compose`, etc.) in the `Dockerfile.devenv`, you would need the `vscode` user to be included in the `docker` group. Finally, remember to set `sleep infinity` in the entrypoint to keep the dev container running indefinitely.
 
 ```dockerfile
 # syntax=docker/dockerfile:1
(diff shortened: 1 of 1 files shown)


rumpl (dismissed):

desktop/dev-environments.md
```diff
@@ -253,6 +253,21 @@ In this preview, Dev Environments support a JSON file which allows you to specif
 }
 ```
 
+Next, you have to define the dependencies you want to include in your `Dockerfile.devenv`, alongside the following requisites:
+
+While any images or Dockerfiles will include a non-root user with a UID/GID of `1000`, many base images and Dockerfiles do not. Fortunately, you can add a non-root user named `vscode`. If you were to include the Docker tooling (e.g. `docker` cli, `docker compose`, etc.) in the `Dockerfile.devenv`, you would need the `vscode` user to be included in the `docker` group. Finally, remember to set `sleep infinity` in the entrypoint to keep the dev container running indefinitely.
+
+```Dockerfile
```
felipecruz91, line 260:
```suggestion
```dockerfile
```

glours (approved):

gtardif (approved):

desktop/dev-environments.md
```diff
@@ -253,6 +253,21 @@ In this preview, Dev Environments support a JSON file which allows you to specif
...
 
+Next, you have to define the dependencies you want to include in your `Dockerfile.devenv`, alongside the following requisites:
+
+While any images or Dockerfiles will include a non-root user with a UID/GID of `1000`, many base images and Dockerfiles do not. Fortunately, you can add a non-root user named `vscode`. If you were to include the Docker tooling (e.g. `docker` cli, `docker compose`, etc.) in the `Dockerfile.devenv`, you would need the `vscode` user to be included in the `docker` group. Finally, remember to set `sleep infinity` in the entrypoint to keep the dev container running indefinitely.
+
+```dockerfile
+FROM <your base image>
+
+RUN useradd -s /bin/bash -m vscode
+RUN groupadd docker && usermod -aG docker vscode
```
thaJeztah, line 264: This will create 2 layers; perhaps we should combine these;

```suggestion
RUN useradd -s /bin/bash -m vscode \
 && groupadd docker \
 && usermod -aG docker vscode
```

Do we require `/bin/bash` or does `/bin/sh` work as well? `/bin/bash` won't be available by default in many images.

These should likely use `useradd` instead of `adduser`, because `adduser` is an optional install (`useradd` / `groupadd` are the default; see https://superuser.com/a/1543631)

When updating, also be sure to test them on `rpm` (e.g. `fedora` or `centos`), `apk` (alpine), and `deb` (`ubuntu` or `debian`) based base images; unfortunately the command flags differ a bit between distros 😞 (see the Moby code, which also has some variants for this); https://github.com/moby/moby/blob/7b9275c0da707b030e62c96b679a976f31f929d3/pkg/idtools/usergroupadd_linux.go#L65-L88

The Dockerfile for the nginx image may be useful to find the corresponding variants for alpine and debian;

- alpine: https://github.com/nginxinc/docker-nginx/blob/01561128671d4e33c039a654fe198c6259a2514c/mainline/alpine/Dockerfile#L16-L17
- debian: https://github.com/nginxinc/docker-nginx/blob/3a7105159a6c743188cb1c61e4186a9a59c025db/mainline/debian/Dockerfile#L16-L17

desktop/dev-environments.md
```diff
@@ -253,6 +253,21 @@ In this preview, Dev Environments support a JSON file which allows you to specif
 }
 ```
 
+Next, you have to define the dependencies you want to include in your `Dockerfile.devenv`, alongside the following requisites:
+
+While any images or Dockerfiles will include a non-root user with a UID/GID of `1000`, many base images and Dockerfiles do not. Fortunately, you can add a non-root user named `vscode`. If you were to include the Docker tooling (e.g. `docker` cli, `docker compose`, etc.) in the `Dockerfile.devenv`, you would need the `vscode` user to be included in the `docker` group. Finally, remember to set `sleep infinity` in the entrypoint to keep the dev container running indefinitely.
```
thaJeztah, line 258: Should the example explicitly use `1000` as `uid` ?

Also wondering; is the `vscode` user required, or would numeric `uid/gid` work?;

```dockerfile
USER 1000:1000
```

usha-mandya:
@felipecruz91 Is this ready fo review?

felipecruz91:
> @felipecruz91 Is this ready fo review?

There're still some valid points that @thaJeztah commented that I'd like to discuss with @glours.

desktop/dev-environments.md
felipecruz91: After a discussion with @glours, we want to avoid confusing the reader with the ID/GID `1000`. 

We'll remove that part and just mention adding a non-root user (regardless of the ID/GID) named `vscode`.

desktop/dev-environments.md
```diff
@@ -253,6 +253,24 @@ In this preview, Dev Environments support a JSON file which allows you to specif
...
```
felipecruz91, line 258:
```suggestion
While some images or Dockerfiles will include a non-root user, many base images and Dockerfiles do not. Fortunately, you can add a non-root user named `vscode`. If you were to include the Docker tooling (e.g. `docker` cli, `docker compose`, etc.) in the `Dockerfile.devenv`, you would need the `vscode` user to be included in the `docker` group. Finally, remember to set `sleep infinity` in the entrypoint to keep the dev container running indefinitely.
```

desktop/dev-environments.md
```diff
@@ -253,6 +253,24 @@ In this preview, Dev Environments support a JSON file which allows you to specif
 }
 ```
 
+Next, you have to define the dependencies you want to include in your `Dockerfile.devenv`, alongside the following requisites:
+
+While some images or Dockerfiles will include a non-root user, many base images and Dockerfiles do not. Fortunately, you can add a non-root user named `vscode`. If you were to include the Docker tooling (e.g. `docker` cli, `docker compose`, etc.) in the `Dockerfile.devenv`, you would need the `vscode` user to be included in the `docker` group. Finally, remember to set `sleep infinity` in the entrypoint to keep the dev container running indefinitely.
```
thaJeztah, line 258: > Finally, remember to set `sleep infinity` in the entrypoint to keep the dev container running indefinitely.

Perhaps a bit off-topic (just now noticing this part); setting `sleep infinity` in entrypoint seems like a bit of a large hammer (and makes the image/container harder to use).

IIUC, we (can) control how the container is started, correct? I'm wondering if instead we could just make the container run with different options to "make it work".

When trying to run a container "detached" but the container's process is not a long-living process (e.g. has `/bin/sh` or `/bin/bash` as entrypoint), starting it detached will make it exit immediately; this is because the shell detects there's no terminal attached, so it has nothing to do;

```bash
docker run -d busybox
eb94e6e28b41bc5bb6f651cc54e3aab5a46a56d805f3787eeb9451a54301af47

docker ps -a

CONTAINER ID   IMAGE                        COMMAND                  CREATED          STATUS                      PORTS                            NAMES
eb94e6e28b41   busybox                      "sh"                     15 seconds ago   Exited (0) 12 seconds ago                                    serene_keldysh
```

However, starting the same image with a TTY attached (`-t` or `-it`) will keep it running; the shell detects that a TTY is attached, so keeps waiting; here's a busybox container that's started with both `-d` (detached) and `-t` (a tty);

```bash
docker run -dt busybox
5b7e989747625b023144a7c5ab0af092ea0708f13bcdc58994bdb591d7a19e22

docker ps -a
CONTAINER ID   IMAGE                        COMMAND                  CREATED              STATUS                          PORTS                            NAMES
5b7e98974762   busybox                      "sh"                     3 seconds ago        Up 2 seconds                                                     nice_black
eb94e6e28b41   busybox                      "sh"                     About a minute ago   Exited (0) About a minute ago                                    serene_keldysh
```

glours: We can remove the sentence about setting the entrypoint to `sleep infinity` to not confuse the user, the Dev Env backend server is taking care of this.

felipecruz91, line 258:
```suggestion
While some images or Dockerfiles will include a non-root user, many base images and Dockerfiles do not. Fortunately, you can add a non-root user named `vscode`. If you were to include the Docker tooling (e.g. `docker` cli, `docker compose`, etc.) in the `Dockerfile.devenv`, you would need the `vscode` user to be included in the `docker` group.
```

thaJeztah (approved):
LGTM, thanks for updating!

PR #14033, opened 2021-12-22
updated SSO banner

jerae-duffin:
Updated SSO banner and DNS info


izzychu (dismissed):

thaJeztah:
Thanks; I'll try to have a look later; I want to have a quick look at the images, as it looks like the previous versions were scaled down (probably to better match the size as shown on the page); and the new ones are quite a bit larger, so perhaps I can scale them a bit

single-sign-on/index.md
```diff
@@ -35,20 +39,40 @@ To configure SSO, log into [Docker Hub](https://hub.docker.com){: target="_blank
...
 
 ![SSO SAML](images/sso-saml.png){:width="500px"}
 
-### Domain
+### Domain control
 
-1. Click **Add Domain** and specify the email domains that are allowed to authenticate via your server.
-    Note: This should include all email domains users will use to access Docker. Public domains are not permitted, such as gmail.com, outlook.com, etc. Also, the email domain should be set as the primary email.
-2. Click **Send Verification** to receive an email for the domains you have specified and verify your domain.
+Click Add Domain and specify the corporate domain you’d like to manage with SSO. Domains should be formatted without protocol or www information, for example, yourcompany.com.
```
thaJeztah, line 52: I'm wondering if we should include something describing that the domain they're looking for is the domain that will be used by their users as part of their e-mail. Perhaps that makes it easier to understand (companies may have various domains, e.g., Docker has `docker.io`, `dckr.io`, and `docker.com`, but (I think) only `docker.com` is relevant in this case as that's the domain used for e-mail addresses that employees use.

That way we could describe that if users will use `user@company-mail.org`, then they need to use that domain (everything after `@`).


I'm not sure though what needs to be entered if (less common) their users use e-mail addresses on subdomains. For example, I know IBM used per-department subdomains at some point (`name@linux.vnet.ibm.com`). Do they need to enter `ibm.com` in that case, or `linux.vnet.ibm.com` ?

Let me try get an answer to that last one ^^

thaJeztah (commented):
thanks! I had a look at the images, but trying to resize them didn't make a huge difference in their file size, so 👍 let's keep those as-is.

Left some comments below

single-sign-on/index.md
jerae-duffin: @thaJeztah in line 56 we mention the users and access to Docker and the domains that aren't permitted. I can check to see if there are other instances where they would use a domain such as:  linux.vnet.ibm.com

thaJeztah (approved):
LGTM (I'm ok with addressing the "domain" bit in a follow-up if needed)

shawnaxsom (approved):

PR #14042, opened 2021-12-23
added workaround for different distribution

3apa3a-b-ta3e:
Some distribution like Oracle Linux for aarch64 architecture reports their version as "Linux" instead of usual "linux" and this one-liner doesn't work.

### DIFF
--- a/compose/install.md
+++ b/compose/install.md
@@ -132,7 +132,7 @@ also included below.
 1.  Run this command to download the current stable release of Docker Compose:
 
     ```console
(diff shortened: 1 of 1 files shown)


StefanScherer:
Thanks @3apa3a-b-ta3e , but I wonder why this is needed. 
It would fail for Mac which returns Darwin with an upper case letter. Anyway the URL's are case insensitive (server side) and it should work without the `tr`.

usha-mandya:
Thanks @StefanScherer.

@3apa3a-b-ta3e Closing this ticket as the existing docs look good. Feel free to add more information on the ticket to continue the conversation.

Issue #6891, opened 2018-06-16
custom docker0 config example broken


FlorianLudwig:
[engine/userguide/networking/default_network/custom-docker0.md](https://docs.docker.com/engine/userguide/networking/default_network/custom-docker0/)

Using the example from the documentation:
```
{
  "bip": "192.168.1.5/24",
  "fixed-cidr": "192.168.1.5/25",
  "fixed-cidr-v6": "2001:db8::/64",
  "mtu": 1500,
  "default-gateway": "10.20.1.1",
  "default-gateway-v6": "2001:db8:abcd::89",
  "dns": ["10.20.1.2","10.20.1.3"]
}
```

which is inconsistent in itself and wouldn't work.

```
Error starting daemon: Error initializing network controller: Error creating default "bridge" network: auxilairy address: (DefaultGatewayIPv4:10.20.1.1) must belong to the master pool: 192.168.1.0/24
```

L-Hudson:
Closing this ticket due to its age, and the impending refactor. If you think this is in error, feel free to reopen. Thanks!

PR #14053, opened 2021-12-27
Fix bridge example

tatref:
### Proposed changes

Fix bridge example in docs

### Related issues (optional)

This is a followup of #6891

`default-gateway` must be included inside the `bip` network, otherwise restarting docker gives:

    dockerd: failed to start daemon: Error initializing network controller: Error creating default "bridge" network: auxiliary address: (DefaultGatewayIPv4:10.20.1.1) must belong to the master pool: 192.168.1.0/24

I moved `bip` to be the 1st address  (.1), and `default-gateway` to be the last one (.254).

I also changed `fixed-cidr` to .0/25 instead of .5/25, as I think this makes more sense: the fixed network is the first 128 addresses of the network (0-127).

### DIFF
--- a/network/bridge.md
+++ b/network/bridge.md
@@ -219,11 +219,11 @@ the settings you need to customize.
 
 ```json
 {
-  "bip": "192.168.1.5/24",
-  "fixed-cidr": "192.168.1.5/25",
+  "bip": "192.168.1.1/24",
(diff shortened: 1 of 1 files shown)


usha-mandya (approved):
LGTM

PR #14060, opened 2021-12-30
Fix anchor locations

halkeye:
### Proposed changes

Currently, if you go to https://docs.docker.com/compose/release-notes/#1290 the header (h*) is actually cutover by the header (top bar). 

Ex: https://docs.docker.com/compose/release-notes/#1290

image

vs

image

N/A: Pure docs site change

### Related issues (optional)

N/A


brianrobt (dismissed):
lgtm

timja:
https://deploy-preview-14060--docsdocker.netlify.app/compose/release-notes/#1290 looks the same to me as https://docs.docker.com/compose/release-notes/#1290 ?

halkeye:
> https://deploy-preview-14060--docsdocker.netlify.app/compose/release-notes/#1290 looks the same to me as https://docs.docker.com/compose/release-notes/#1290 ?

it does look like when coming in from external site, it does sorta work, but when you click on the chain icon next to header, it jumps to under the topbar.

halkeye:
So my guess is that its https://github.com/docker/docker.github.io/blob/master/js/docs.js#L165

thaJeztah:
Thanks for looking into this; yes, I recall the layout was a bit quirky in this area and collected a good amount of legacy code, css, and "oh,let's apply a quick hack here and there" over the years. I tried to clean up some of that, but it's been a good 7 years or so since I did any front-end work 😅, so there may be some remaining issues (ideally, we should do a good cleanup / overhaul / rewrite of the docs "theme").

I'll try to reserve some time soon to have a look at your changes, and to check if it doesn't affect some of the other "hacks" that are in place 😬

halkeye:
We essentially had the same issue on jenkins.io which Is how I learned about it. I think with this you can probably safely get rid of most of the hacks, but I didn't want to mess with very much, CSS tends to be brittle on long running sites :)

PR #14070, opened 2022-01-06
Updates Step 9 `docker inspect` commands in Copy-on-write strategy section of About storage drivers page

lmshipp:
There is an error in the example commands of Step 9 in the copy on write strategy section of the [About storage drivers](https://docs.docker.com/storage/storagedriver/#the-copy-on-write-cow-strategy) page. The first command should reference the `acme/my-base-image:1.0` image and the second command should reference the `acme/my-final-image:1.0`.

### DIFF
--- a/storage/storagedriver/index.md
+++ b/storage/storagedriver/index.md
@@ -339,7 +339,7 @@ layers are the same.
 
     {% raw %}
     ```console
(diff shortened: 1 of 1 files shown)


thaJeztah (approved):
LGTM

PR #14079, opened 2022-01-08
incorrect download link

KY-7:
For example: /download/v2.2.3/ (correct) instead of  /download/2.2.3/ (incorrect)


usha-mandya (approved):
LGTM

PR #14080, opened 2022-01-08
mysql volume path is not correct

mek-12:
Path should be like below:
/etc/todos

### DIFF
--- a/get-started/08_using_compose.md
+++ b/get-started/08_using_compose.md
@@ -259,7 +259,7 @@ services:
   mysql:
     image: mysql:5.7
     volumes:
-      - todo-mysql-data:/var/lib/mysql
+      - todo-mysql-data:/etc/todos
     environment:
       MYSQL_ROOT_PASSWORD: secret
       MYSQL_DATABASE: todos


StefanScherer:
Thanks @mek-12, but the idea is to have the whole database folder in the volume, and `/var/lib/mysql` is the default folder inside the container.
Could you explain a bit more why this change is needed? It seems fine to me as it is right now.

mek-12:
@StefanScherer. Hi Mr. Stefan. I am sorry for my english, because not perfect. As for the review issue;
I'm new to Docker. And I when run this docker-compose  Then I was getting an error. Error as I remember: "todos is not found. "According to my previous notes, when I changed the volume thinking it had access, the error went away. Thanks for your feedback.

rayedbw:
Hi @mek-12, the change you are proposing is not correct. MySQL by default writes its data files to `/var/lib/mysql`. In the example, a volume named `todo-mysql-data` is being mounted to the correct location where MySQL will store its data. I don't see anything wrong here. Please take a look at the [MySQL Docker page](https://hub.docker.com/_/mysql) for mor information. Here's a screenshot from the relevant section.

docker-mysql-data

usha-mandya:
@StefanScherer @rayedbw Thanks for weighing in.

@mek-12 Closing this issue for the reasons described in the previous comments. Feel free to continue the conversation after I close

rayedbw:
Thank you @mek-12 for raising the topic. Thank you @StefanScherer for your input and thank you @usha-mandya for validating my understanding 🙂

PR #14083, opened 2022-01-10
Update 02_our_app.md: Fix copy command

sarahchima:
Running `Copy .  .` leads to a file not found error because the files are copied to the wrong path(see the image attached). 

Screenshot 2022-01-10 at 07 41 48

Running `Copy .  /` fixes this.

### DIFF
--- a/get-started/02_our_app.md
+++ b/get-started/02_our_app.md
@@ -47,7 +47,7 @@ see a few flaws in the Dockerfile below. But, don't worry. We'll go over them.
    FROM node:12-alpine
    RUN apk add --no-cache python2 g++ make
    WORKDIR /app
-   COPY . .
+   COPY . /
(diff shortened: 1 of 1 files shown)


StefanScherer:
Thanks @sarahchima, but I don't think it should be.
Maybe the problem was that the downloaded zip file had a toplevel `app` folder. The idea of the guide is to create the `Dockerfile` in the same folder as the `package.json`, then there is `WORKDIR /app` and then the `COPY . .` copies the sources to /app inside the image.
I'm thinking how we can improve the guide to make it more clear. Do you have any idea?

usha-mandya:
Thanks @StefanScherer

@sarahchima Closing this issue as the existing docs look good. Feel free to continue the conversation after I close

PR #14086, opened 2022-01-10
Update rhel.md

the-src:
The dockerrepo only provides "centos" (not rhel) as distribution. 

So the url must like `https://download.docker.com/linux/centos/docker-ce.repo`

More: https://access.redhat.com/discussions/6249651

### DIFF
--- a/engine/install/rhel.md
+++ b/engine/install/rhel.md
@@ -82,7 +82,7 @@ from the repository.
 
 #### Set up the repository
 
-{% assign download-url-base = "https://download.docker.com/linux/rhel" %}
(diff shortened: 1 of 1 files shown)


thaJeztah:
Thank you for contributing! Unfortunately, I don't think this change is correct; while we don't have packages for RHEL for x86, we do publish packages that have been built by IBM for s390x; the "prerequisites" section contains a note about that;
https://docs.docker.com/engine/install/rhel/#prerequisites


Screenshot 2022-01-11 at 13 59 47

For other architectures, the CentOS packages may be an alternative for RHEL

I'm closing this PR because of the above, but feel free to continue the conversation

PR #14092, opened 2022-01-11
go/buildx redirect

crazy-max:
This URL will used in https://github.com/docker/cli/pull/3314 to direct users to instructions on installing buildx. Maybe `go/buildx-install` would be better?

The last commit updates the install instructions for buildx (will also be changed upstream on buildx repo). Atm the instructions for `Manual download` redirect to the GitHub releases page but in the future we might change that to use our `download.docker.com` front. See https://github.com/docker/docker-ce-packaging/issues/618.

cc @chris-crone @tonistiigi @thaJeztah @errordeveloper

### DIFF
--- a/buildx/working-with-buildx.md
+++ b/buildx/working-with-buildx.md
@@ -14,24 +14,66 @@ multiple nodes concurrently.
 
 ## Install
 
-Docker Buildx is included in Docker Desktop and Docker Linux packages when
-installed using the [DEB or RPM packages](../engine/install/index.md).
+### Windows and macOS
 
-You can also download the latest `buildx` binary from the
-[Docker buildx](https://github.com/docker/buildx/releases/latest){:target="_blank" rel="noopener" class="_"} releases page
-on GitHub, copy it to `~/.docker/cli-plugins` folder with name
-`docker-buildx` and change the permission to execute:
+Docker Buildx is included in [Docker Desktop](../desktop/index.md) for Windows
+and macOS.
 
-```console
-$ chmod a+x ~/.docker/cli-plugins/docker-buildx
-```
+### Linux packages
 
-Here is how to use buildx inside a Dockerfile through the
+Docker Linux packages also include Docker Buildx when installed using the
+[DEB or RPM packages](../engine/install/index.md).
+
+### Manual download
+
+> **Important**
+>
+> This section is for unattended installation of the buildx component. These
+> instructions are mostly suitable for testing purposes. We do not recommend
+> installing buildx using manual download in production environments as they
+> will not be updated automatically with security updates.
+>
+> On Windows and macOS, we recommend that you install [Docker Desktop](../desktop/index.md)
+> instead. For Linux, we recommend that you follow the [instructions specific for your distribution](#linux-packages).
+{: .important}
+
+You can also download the latest binary from the [releases page on GitHub](https://github.com/docker/buildx/releases/latest){:target="_blank" rel="noopener" class="_"}.
+
+Rename the relevant binary and copy it to the destination matching your OS:
+
+| OS       | Binary name          | Destination folder                       |
+| -------- | -------------------- | -----------------------------------------|
+| Linux    | `docker-buildx`      | `$HOME/.docker/cli-plugins`              |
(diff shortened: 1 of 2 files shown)


buildx/working-with-buildx.md
```diff
@@ -14,24 +14,42 @@ multiple nodes concurrently.
...
+
+Docker Linux packages also include Docker Buildx when installed using the
+[DEB or RPM packages](../engine/install/index.md).
+
+### Manual download
+
+You can also download the latest binary from the [releases page on GitHub](https://github.com/docker/buildx/releases/latest){:target="_blank" rel="noopener" class="_"}.
+
+Rename the relevant binary for your OS to `docker-buildx` and copy it to
+`$HOME/.docker/cli-plugins`.
```
thaJeztah, line 32: I guess if users use `sudo` to run `docker`, they should pick one of the instructions below, correct?

We also may need to add some information about which path to pick in general, as they're processed in a specific order, which means that users could inadvertently "mask" the version that comes with their install with an older version (and never update)

buildx/working-with-buildx.md
```diff
@@ -14,24 +14,42 @@ multiple nodes concurrently.
...
+
+Or copy it into one of these folders for installing it system-wide:
+
+* `/usr/local/lib/docker/cli-plugins` OR `/usr/local/libexec/docker/cli-plugins`
+* `/usr/lib/docker/cli-plugins` OR `/usr/libexec/docker/cli-plugins`
+
+Might require to make it executable with `chmod +x`:
+
+```shell
+$ chmod +x ~/.docker/cli-plugins/docker-buildx
```
thaJeztah, line 42: Do we need manual instructions for Windows (and macOS) as well?

We should probably link this content from; https://docs.docker.com/engine/install/binaries/ as well

buildx/working-with-buildx.md
```diff
@@ -14,24 +14,42 @@ multiple nodes concurrently.
...
+### Dockerfile
+
+Here is how to install and use Buildx inside a Dockerfile through the
 [`docker/buildx-bin`](https://hub.docker.com/r/docker/buildx-bin) image:
 
 ```dockerfile
 FROM docker
-COPY --from=docker/buildx-bin /buildx /usr/libexec/docker/cli-plugins/docker-buildx
+COPY --from=docker/buildx-bin:latest /buildx /usr/libexec/docker/cli-plugins/docker-buildx
 RUN docker buildx version
```
thaJeztah, line 53: Should this (and other instructions) include a `docker buildx install` step, and a way to verify that `docker buildx install` correctly replaced `docker build` to be backed by buildx?

thaJeztah:
> Maybe go/buildx-install would be better?

The goal is to have a short, memorable, easy-to-type URL (mostly because copy/paste from the command-line may not always work, so users should have an easy way to manually type the URL), so shorter is better, but I'm fine either way.

buildx/working-with-buildx.md
crazy-max: > I guess if users use sudo to run docker, they should pick one of the instructions below, correct?

Correct. We might stipulate that this is for an advanced usage and like you said warn users about it. Maybe add a warning bloc about it, wdyt?

buildx/working-with-buildx.md
crazy-max: Yes indeed. Maybe another page specific to plugins so we could use the same for others like compose, etc..

buildx/working-with-buildx.md
crazy-max: We already have a dedicated section below that talks about it `Set buildx as the default builder`. We might update it when https://github.com/docker/cli/pull/3314 is merged/released.

crazy-max:
> > Maybe go/buildx-install would be better?
> 
> The goal is to have a short, memorable, easy-to-type URL (mostly because copy/paste from the command-line may not always work, so users should have an easy way to manually type the URL), so shorter is better, but I'm fine either way.

Sure SGTM

buildx/working-with-buildx.md
crazy-max: Updated let me know what you think.

thaJeztah (approved):
LGTM, thanks!

PR #14096, opened 2022-01-11
Add migration V1 -> V2 info

ulyssessouza:
### Proposed changes
Add V1 -> V2 migration notes


StefanScherer (approved):
LGTM

mat007 (commented):
Nothing about environment variables, or variable substitution?
Are we assuming V2 should have the same behaviour us V1 now?

ulyssessouza:
@mat007 The differences I had on my list are already fixed. I'm maybe missing something, but from what I remember that's conceptually the same now. If there are differences they should be considered as bugs

mat007 (approved):

ulyssessouza:
@mat007 And also, we can do that incrementally with followup PRs like this one

PR #14111, opened 2022-01-14
Add release notes for Desktop 4.4.3

mat007:
### Proposed changes

Added release notes for Desktop 4.4.3.
This is a Windows only patch release to fix https://github.com/docker/for-win/issues/12495

I copied the release notes for 4.4.2 and added the bug fix description at the end of the relevant section.

### Related issues (optional)

https://github.com/docker/for-win/issues/12495

### DIFF
--- a/desktop/windows/release-notes/index.md
+++ b/desktop/windows/release-notes/index.md
@@ -19,14 +19,53 @@ This page contains information about the new features, improvements, known issue
 
 Take a look at the [Docker Public Roadmap](https://github.com/docker/roadmap/projects/1){: target="_blank" rel="noopener" class="_"} to see what's coming next.
 
-## Docker Desktop 4.4.2
-2022-01-13
+## Docker Desktop 4.4.3
+2022-01-14
 
 > Download Docker Desktop
 >
 > [For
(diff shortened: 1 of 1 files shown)


thaJeztah:
Should we just mention the regression fix in the release notes, and mention it's otherwise identical to 4.4.2 `<link to 4.4.2 anchor>` ?

Will there be a Docker Desktop for Mac release as well (same bits, only different version number?) (wondering if it's confusing if they're not at the same version - have we had this before?)

usha-mandya:
Yea, we've had similar situations before where we've issued a release for Mac only (3.0.1, 3.0.2, 3.03). I know we haven't called out the details in Day1, Day2 patch releases, but I think if it's beneficial to call out the New/Upgrades, etc for users who directly land on 4.2.3.

mat007:
Yes, it’s how we’ve handled similar situations in the past.
But I’m happy to amend if we think there is a better way.
@stephanierifai any opinion?

thaJeztah:
Yeah, my train of thought was that the "full release notes" are just a small scroll away (especially if we mention it's identical otherwise), and having the "fix" clearly put in the release notes for this release would make it stand out better.

mat007:
@usha-mandya I have fixed the link to download 4.4.2

usha-mandya (approved):
LGTM

PR #14131, opened 2022-01-17
Update install.md

EB8F9A:
linux docker-compose download path error


usha-mandya:
Thank you for the PR. This is fixed as part of #14151. Closing this PR without accepting your changes as the updates are live on docs.docker.com

PR #14133, opened 2022-01-17
FIX: fixed download docker-compose not found

sansroman:


usha-mandya:
Thank you for the PR. This is fixed as part of #14151. Closing this PR without accepting your changes as the updates are live on docs.docker.com

PR #14147, opened 2022-01-18
Update install.md: Fixed urls to download releases

karimmakhloufi:
It seems that github changed the urls to download a release and that the `v` before the tag must not be added anymore.

For example this link works: https://github.com/docker/compose/releases/download/1.29.2/docker-compose-Linux-x86_64

But this one, generated by the command in the readme file does not work anymore: https://github.com/docker/compose/releases/download/v1.29.2/docker-compose-Linux-x86_64

I'm maybe missing something but I think we should remove the `v` from the urls to make the download command work again

### DIFF
--- a/compose/install.md
+++ b/compose/install.md
@@ -37,7 +37,7 @@ Python package manager or installing Compose as a container.
 > Install a different version
 >
 > The instructions below outline installation of the current stable release
-> (**v{{site.compose_version}}**) of Compose. To install a different version of
+> (**{{site.compose_version}}**) of Compose. To install a different version of
 > Compose, replace the given release number with the one that you want. For instructions to install Compose 2.0.0 on Linux, see [Install Compose 2.0.0 on Linux](cli-command.md#install-on-linux).
 >
 > Compose releases are also listed and available for direct download on the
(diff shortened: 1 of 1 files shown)


rumpl:
The issue is that some of the tags in compose have the v t prefix and some don’t. I guess a fix would be to create tags with the prefix where it’s missing

karimmakhloufi:
@rumpl you seem to be right, the `v` appeared  after the version 2.0 https://github.com/docker/compose/releases/
Changing tags will break existing permalinks, I still think we should update the documentation with a working link. If we choose to suggest the installation of v2, change the version from 1.29.2 to v2.2.3, if we prefer users to install version 1.29.2 for now: merge this PR to remove the `v` in the url.

What do you think ?

usha-mandya:
@karimmakhloufi Thanks for the PR. 

We've added the correct download URLs for the existing Compose version and also added a pointer to the download link for Compose v2 users. Please see the Linux section in https://docs.docker.com/compose/install/#install-compose.

Closing this ticket here as the Docs are updated. Feel free to continue the conversation after I close.

PR #14156, opened 2022-01-18
Update protect-access.md with remote host Docker version requirement

stayclassychicago:
### Proposed changes

Update docs, expanding the note to include the minimum Docker version
that needs to be installed on the remote host.

This makes the existing error message more transparent:
>please make sure the URL is valid, and Docker 18.09 or later is installed on the remote host


thaJeztah:
In general, our docs reflect currently supported versions of docker; docker 18.09 has reached EOL more than three years ago; did you run into this issue on an even older version? (as those versions are no longer supported)

stayclassychicago:
Understood., closing issue.  Unfortunately we are slow to upgrade and have not completed the migration to 18.09; which is where we ran into this.

PR #14161, opened 2022-01-21
Reorder words in about section

asandoval95:
Changes "... drivers that a use copy-on-write filesystem."  to "... drivers that use a copy-on-write filesystem."


usha-mandya (approved):
LGTM. Thank you :)

PR #14166, opened 2022-01-24
cgroups: fix typo, soften language around upgrading to v2

djs55:
### Proposed changes

1. Fix a typo in a release note (sorry, my mistake!)
2. Soften the language around updating containers from cgroup v1 to v2. I expect that v1 will continue to work, but that v2 is preferred (since it is where all the development is happening in the upstream Linux kernel). It's likely that some future features will depend on v2, but nothing critical to current use.

### Related issues (optional)

See discussion here: https://github.com/docker/for-mac/issues/6073


mat007 (approved):
Thanks!

PR #14167, opened 2022-01-24
Release notes for Desktop 4.4.4

mat007:
### Proposed changes

Adding release notes for Desktop 4.4.4.

This is a Windows only point release, only the first security item and the last bug fix item are new.

### DIFF
--- a/desktop/windows/release-notes/index.md
+++ b/desktop/windows/release-notes/index.md
@@ -23,14 +23,55 @@ This page contains information about the new features, improvements, known issue
 
 Take a look at the [Docker Public Roadmap](https://github.com/docker/roadmap/projects/1){: target="_blank" rel="noopener" class="_"} to see what's coming next.
 
-## Docker Desktop 4.4.3
-2022-01-14
+## Docker Desktop 4.4.4
+2022-01-24
 
 > Download Docker Desktop
 >
 > [For
 > Windows](https://desktop.docker.com/win/main/amd64/Docker%20Desktop%20Installer.exe?utm_source=docker&utm_medium=webreferral&utm_campaign=docs-driven-download-win-amd64){:
 > .button .primary-btn }
+
+### Security
+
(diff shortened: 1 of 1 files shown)


stephen-turner (changes requested):
I'm not sure that this idea of copying all the release notes from the previous point release works. I do understand that we're trying to cater for users who don't read back, but it's now become hard for users who have taken the intermediate release to understand what has changed in this release.

I think it would be reasonable to expect users to read back for "what's new" information, and we only list what's new since the last point release, but for us to repeat "known issues" if they are continuing. But I'd like to hear @usha-mandya's pov on it too.

usha-mandya:
@stephen-turner As the percentage of users who've taken the intermediate release was relatively small, the release notes were copied from the previous releases. I'd also like to understand how we display this info in product where users are notified about the availability of a new release through Docker Desktop. @mat007 ?

mat007:
>  I'd also like to understand how we display this info in product where users are notified about the availability of a new release through Docker Desktop.

It’s entirely separate, what they see is the «Release Highlights» that are set in Omakase.
For 4.4.3 and 4.4.4 we’ve copied the highlights from 4.4.2.

stephen-turner:
For the in-product release highlights, I agree it makes sense to repeat them.

desktop/windows/release-notes/index.md
```diff
@@ -34,6 +34,47 @@ Take a look at the [Docker Public Roadmap](https://github.com/docker/roadmap/pro
 
 ### Security
 
+- Fixed [CVE-2022-23774](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2022-23774) where arbitrary files could be moved on the system.
```
stephen-turner, line 37: Do we normally credit the reporters of vulnerabilities in the release notes? (Two independent reports in this case). Or only in the CVE?

usha-mandya: I don't think we've credited the reporters of vulnerabilities in previous release notes. We link to the CVEs for details.

usha-mandya (approved):
LGTM

StefanScherer (approved):
LGTM

PR #14179, opened 2022-01-27
Fix x-aws-keys field name

stevenpitts:
The current compose-cli documentation suggests listing SecretsManager secret JSON keys in a `keys` field, but this is not correct.
The correct field is `x-aws-keys`.

The `keys` field as written is not supported by the [compose spec](https://github.com/compose-spec/compose-spec/blob/master/spec.md), and attempting most `docker compose` commands would result in failure if one named the field as written.

### Proposed changes

Replace the incorrect `keys` field with `x-aws-keys`.

### DIFF
--- a/cloud/ecs-integration.md
+++ b/cloud/ecs-integration.md
@@ -370,7 +370,7 @@ services:
 secrets:
   foo:
     name: "arn:aws:secretsmanager:eu-west-3:1234:secret:foo-ABC123"
-    keys:
+    x-aws-keys:
       - "bar"
 ```
 


stevenpitts:
Tagging @ndeloof as they appear to be responsible for all compose-cli issues

stevenpitts:
> Help us merge your changes more quickly by adding details and setting metadata (such as labels, milestones, and reviewers) over at the right-hand side.

That sure would be a good suggestion if I had the permissions to do that

usha-mandya (approved):
LGTM

PR #14201, opened 2022-02-01
Changed log-opt to log-opts

ValorNaram:
### Proposed changes

Changed text `log-opt` to `log-opts` because this is how the JSON key must be named in `daemon.json`. `log-opt` is wrong and using it leads docker to fail with its starting routine.


usha-mandya (approved):
LGTM. Thank you @ValorNaram :)

PR #14203, opened 2022-02-01
Add another reason for using a VM

stephen-turner:
Add note about controlling the kernel and OS inside the VM


djs55 (approved):
LGTM

usha-mandya (approved):
LGTM

PR #14212, opened 2022-02-02
Update the docs page ratings display

usha-mandya:

Update the docs page ratings display


thaJeztah (approved):
LGTM (from looking at the change - the netlify preview doesn't show poll daddy, so we'll have to double check after it's published)

PR #14214, opened 2022-02-02
Update 04_sharing_app.md

bartduisters:
There is no 'Docker commands' section ont he right side. There is an image with the command below, hence the change.


usha-mandya (approved):
Thank you :)

PR #14215, opened 2022-02-03
Update index.md to present clearer overview persisted versus in-memory storage

softwarepronto:
There was no transition or clear delineation between persisted storage to in-memory storage. The main paragraph was confusing because it discussed persisted storage and then just mention the Linux/Windows options for supporting in-memory file storage.


usha-mandya (approved):
LGTM. Thank you :)

PR #14218, opened 2022-02-03
Linux: update url for tech preview build

aiordache:
We have a new build with small fix for running `docker scan` without any manual user configuration


usha-mandya (approved):
LGTM

PR #14225, opened 2022-02-03
ASP.NET Core does not run on .NET Framework

breggles:
ASP.NET Core is almost a complete rewrite and an ASP.NET Core app does not run on .NET Framework. See e.g. https://docs.microsoft.com/en-us/aspnet/core/introduction-to-aspnet-core?view=aspnetcore-6.0#why-choose-aspnet-core.


usha-mandya (approved):
LGTM. Thank you

PR #14228, opened 2022-02-04
Move example of symbolic link closer to related text

jrudel:
### Proposed changes

This is to co-locate the text explaining why a symbolic link might be required with the example of the symbolic link creation. This will make it more clear that the symbolic link example is related to the note immediately above.


jrudel:
Hi @usha-mandya, thank you for taking a look. It looks like the checks have passed, but I am unable to merge because I am not an authorized user. Are you able to merge this for me?

PR #14234, opened 2022-02-06
Update 07_multi_container.md

bartduisters:
### Proposed changes

Add a script that will work on Apple Silicon, this will save time of people who are following along.


StefanScherer (approved):
Thanks @bartduisters for your contribution.
I only make some minor changes to enable better syntax highlighting.
And I was surprised that `linux/x86_64` also works 😅 , but I've changed it to the more common `linux/amd64`.

PR #14236, opened 2022-02-06
Update overview.md

haripriya1909:
In THE  DOCKER PLATFORM section , there is a small grammatical error in  second line it should be "The isolation and security allows you to run many
containers simultaneously on a given host." (allow->allows) not "The isolation and security allow you to run many
containers simultaneously on a given host."


usha-mandya (approved):
@haripriya1909 Thank you :)

PR #14256, opened 2022-02-11
Fix port number in Getting Started images

danBamikiya:
### Proposed changes
This pull request #14248 used port 8000 the examples in the [compose/gettingstarted](https://docs.docker.com/compose/gettingstarted/) page but the images are still using port 5000 so I switched the images to ones using port 8000.


danBamikiya:
@StefanScherer Hi, I was observing the changes and noticed the images still used port 5000 which will be confusing for readers, here's a PR that switches the images to ones using port 8000, is this okay?

danBamikiya:
> Thanks @danBamikiya! Yes, that's a good catch. Of course the screen shots should display port 8000. Could I ask you if you can provide screenshots with a drop shadow similar to the old ones? Thanks.

Hi @StefanScherer I'm so sorry for the late response, I didn't initially know how to do this & I got busy with work. I've made the requested changes using Figma & I'll love for your review incase there's anymore changes I can make. Thanks!

usha-mandya (approved):
@danBamikiya Thank for updating this. Appreciate it :D

danBamikiya:
@usha-mandya You're welcome! :)

PR #14280, opened 2022-02-17
Fix permissions in buildkit ssh forwarding example.

relsqui:
### Proposed changes

Updates the permissions in the [buildkit ssh forwarding example](https://docs.docker.com/develop/develop-images/build_enhancements/#using-ssh-to-access-private-data-in-builds) to set the `.ssh` directory to 0700 instead of 0600. Directories need the executable bit set to be traversible.


usha-mandya (approved):
LGTM. Thank you @relsqui

Issue #11060, opened 2020-06-29
htpasswd  has disappeared 


scottstensland:
[registry/deploying.md](https://docs.docker.com/registry/deploying/)

binary htpasswd  used to work OK as per your docs however now it has disappeared when using your

    docker run --entrypoint htpasswd registry:2 -Bbn ${local_username}  ${local_reg_pwd}  > ${GKE_DOCKER_REG_INFO}/auth/htpasswd

if you pin to version 2.7.0 it does work

    docker run --entrypoint htpasswd registry:2.7.0 -Bbn ${local_username}  ${local_reg_pwd}  > ${GKE_DOCKER_REG_INFO}/auth/htpasswd 

for details see

 https://github.com/GoogleContainerTools/jib/pull/2538/commits/f816c837e34eb389c2cdee1bc9a2918c5d2e33e3 

https://github.com/GoogleContainerTools/jib/pull/2539 

https://github.com/docker/distribution-library-image/issues/106 

please update your docs to tell us what is now the new proper usage to create password file as your doc gives bad advice

raylua2566:
Not work for `2.7.1` again.

sekhemrekhutawysobekhotep:
true ... avoid using 2.7.1 as its broken instead use 2.7.0 for same reason as above OP has

scottstensland:
here is corresponding stackoverflow ticket where folks have problems due to this unresolved issue

https://stackoverflow.com/questions/62531462/docker-local-registry-exec-htpasswd-executable-file-not-found-in-path/62531540#62531540

craig-osterhout:
This looks like it was fixed in https://github.com/docker/docker.github.io/commit/d39b0d8e6e047ffcf5d0e6eecf45b3495437ceb8

If the issue still persists, please let us know.

PR #14321, opened 2022-02-28
Add read-only and read-write user auth to Nginx recipe

MmAaXx500:
### Proposed changes

It is mentioned at the top of the page of the [nginx recipe](https://github.com/docker/docker.github.io/blob/c3cb5785f17896727fca90d71347bc08cccbfa2e/registry/recipes/nginx.md?plain=1#L35) that there somewhere would be an example about read-only and read-write user authentication. This is missing, so I do like to fill this gap with this.
I have also changed the image used by the password file creation example because `registry:2` does not have the `htpasswd` utility.

### Related issues
Related: #11060, docker/distribution-library-image#106 (and probably fixes these)

### DIFF
--- a/registry/recipes/nginx.md
+++ b/registry/recipes/nginx.md
@@ -134,8 +134,18 @@ Review the [requirements](index.md#requirements), then follow these steps.
           }
 
           # To add basic authentication to v2 use auth_basic setting.
-          auth_basic "Registry realm";
-          auth_basic_user_file /etc/nginx/conf.d/nginx.htpasswd;
+          set $auth_basic "Registry realm rw";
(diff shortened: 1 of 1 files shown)


usha-mandya:
@MmAaXx500 Thanks for the PR. The source files for Registry docs are now in the upstream distribution repository. Could you suggest your changes in https://github.com/distribution/distribution? Thanks. Closing this PR here. Feel free to continue the conversation after I close.

MmAaXx500:
@usha-mandya Thanks for the info, I'll open my PRs there. If these pages are moved there, I suggest that the "Edit this page" on the right side of the documentation page be changed to point to the new repository.

PR #14324, opened 2022-03-01
add endpoint to edit repository

eunomie:
### Proposed changes

Describe endpoint to update repository's settings:

- description
- full description (readme)
- public/private

### DIFF
--- a/docker-hub/api/latest.yaml
+++ b/docker-hub/api/latest.yaml
@@ -94,10 +94,10 @@ tags:
     x-displayName: Org Settings
     description: |
       The Org Settings API endpoints allow you to manage your organization's settings.
-  - name: repository
-    x-displayName: Repository
+  - name: repositories
+    x-displayName: Repositories
     description: |
-        The repository endpoint allows you to manage your repository's configuration like description.
+        The repository endpoints allow you to manage your repository's configuration like description.
 
 x-tagGroups:
   - name: General
@@ -111,7 +111,7 @@ x-tagGroups:
       - images
       - audit-logs
       - org-settings
-      - repository
+      - repositories
 
 definitions:
   UsersLoginRequest:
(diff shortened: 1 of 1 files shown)


docker-hub/api/latest.yaml
```diff
@@ -1,13 +1,7 @@
-swagger: '2.0'
```
technicallyjosh, line 1: Doing a major upgrade here breaks our current generation of some APIs and what is copied into here. So each time a change occurs in an API we then have to translate from swagger 2 to 3.

technicallyjosh (changes requested):
I'd request that we keep swagger 2 until we can get all our internal API docs (generated) into OpenAPI 3. Have you also looked at the effect you have on the Engine docs? They use Redoc as well for that generation and last time I tried upgrading Redoc, it broke lots of things over there.

technicallyjosh:
It would also be helpful if we kept major upgrades as separate PRs 😄

eunomie:
OK understood, I'll go back with swagger and a small hack to have it working in all cases.

docker-hub/api/latest.yaml
```diff
@@ -94,6 +94,10 @@ tags:
     x-displayName: Org Settings
     description: |
       The Org Settings API endpoints allow you to manage your organization's settings.
+  - name: repository
```
technicallyjosh, line 97: **question:** Shouldn't we call this "Repositories" (plural)? Will make it easier to add onto with more routes.

technicallyjosh (approved):
LGTM. Just had a question about the naming 😄

docker-hub/api/latest.yaml
eunomie: 👍 I'll update it

technicallyjosh (approved):
LGTM! Thanks!

usha-mandya (approved):
LGTM. Thank you @eunomie

PR #14338, opened 2022-03-03
Webhooks: remove images property, drop scientific notation for timestamps

ruggi:
### Proposed changes

This PR updates the webhooks documentation, removing the deprecated `images` property from the payload and dropping the scientific notation from the `pushed_at`, `date_created` properties since it's not used (e.g. `1.417566161e+09` → `1417566161`).


usha-mandya (approved):
LGTM. Thank you @ruggi

PR #14348, opened 2022-03-05
Fix static link in Compose V2 installation

NanoCode012:
### Proposed changes

The link on the doc (https://docs.docker.com/compose/cli-command/#install-on-linux) to install Compose V2 on Linux is static. As of right now, the command is hardcoded for linux x86_64 only as seen below. 
```bash
$ curl -SL https://github.com/docker/compose/releases/download/v2.2.3/docker-compose-linux-x86_64 -o $DOCKER_CONFIG/cli-plugins/docker-compose
```

This PR updates it to also accommodate different distro/arch via `$(uname -s)`  `$(uname -m)` respectively.


usha-mandya:
Thanks for the PR. Closing this PR as it's obsolete. Please see https://docs.docker.com/compose/install/linux/.

PR #14349, opened 2022-03-06
removed reference to code line number

giudega:
### Proposed changes
I removed the reference to the code line. I think that using a code line makes the documentation easily out of date, in fact after the latest changes on the Spring PetClinic project the code line that was 54 is now 74


usha-mandya (approved):
LGTM. Thank you

PR #14350, opened 2022-03-06
Add missing argument to registry launch command

leodotcloud:
### Proposed changes
The instructions do not work as the program launch command has a missing argument.


usha-mandya:
@thaJeztah Could you PTAL? Maybe we need a PR in the upstream distribution repo as well to fix this? https://github.com/distribution/distribution/blob/main/docs/recipes/osx/com.docker.registry.plist#L18

thaJeztah:
Yes, this should now be contributed in the upstream; https://github.com/distribution/distribution/tree/main/docs/recipes/osx
and if it's needed for the v2.x version of the registry, probably needs a cherry-pick into the upstream v2.x branch

PR #14357, opened 2022-03-08
Bump redoc to v2.0.0-rc.59

ndeloof:
### Proposed changes

Bump Redoc so we can use multiple example support offered by OpenAPI

### Related issues (optional)
https://github.com/moby/moby/pull/43344


ndeloof:
doens't work out of the box, will require more investigations
https://deploy-preview-14357--docsdocker.netlify.app/engine/api/v1.41/

PR #14363, opened 2022-03-08
The Alpine 3.5 image is no longer available

jonasbn:
### Proposed changes

The Alpine 3.5 image is no longer available, perhaps `latest` should be used.

REF:  [Official Alpine image on Docker Hub](https://hub.docker.com/_/alpine)

A versioned image could be used, but then this situation will simply occur again at some point.


usha-mandya (approved):
LGTM. Thank you @jonasbn

PR #14383, opened 2022-03-13
Add pipx to the alternative install options for docker-compose

### LINKED ISSUE #12964: Install docker-composer with pipx
[compose/install.md](https://docs.docker.com/compose/install/)

Add installation option using pipx. It is simple and easier to uninstall and upgrade.
(Note, I am not python expert! You may want to verify and validate this suggestion!)

Instead current directions:
"...If you install using pip, we recommend that you use a virtualenv because..."
With pipx it will be easier:
"...pipx solves this by creating a virtual environment for each package..."

```
# apt install python3-venv pipx; # my debian-buster needed only these two packages

# pipx ensurepath
Added /root/.local/bin to the PATH environment variable in /root/.bashrc

# pipx install docker-compose
```

More here: https://packaging.python.org/guides/installing-stand-alone-command-line-tools/

NautillusSs:
pipx does the same thing as the pip install option without the hassle of manually creating a virtualenv. The pipx option should be recommended over the pip option which applies only if the user wants to install unrelated additional packages to the same virtual environment.

### Proposed changes

Added the more straightforward pipx install option in the alternative install options for docker-compose. pipx does the same thing as the pip install option without the hassle of manually creating a virtualenv. The pipx option should be recommended over the pip option which applies only if the user wants to install unrelated additional packages to the same virtual environment.

Not applicable.

### Related issues (optional)

Fixes #12964.

### DIFF
--- a/compose/install.md
+++ b/compose/install.md
@@ -174,9 +174,27 @@ also included below.
 
 ### Alternative install options
 
+- [Install using pipx](#install-using-pipx)
 - [Install using pip](#install-using-pip)
 - [Install as a container](#install-as-a-container)
 
+#### Install using pipx
+
+> For `alpine`, the following dependency packages are needed:
+> `py-pip`, `python3-dev`, `libffi-dev`, `openssl-dev`, `gcc`, `libc-dev`, `rust`, `cargo`, and `make`.
+{: .important}
+
+Compose can be installed as an isolated python application from
+[pypi](https://pypi.python.org/pypi/docker-compose) using [pipx](https://pypa.github.io/pipx/). Isolation is recommended because many operating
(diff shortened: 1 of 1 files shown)


NautillusSs:
Updated page is here: https://deploy-preview-14383--docsdocker.netlify.app/compose/install/

usha-mandya:
@NautillusSs Thanks for the PR. We are aiming to recommend using Compose V2 - https://docs.docker.com/compose/cli-command/#installing-compose-v2. Compose V1 (and therefore, pip install instructions) may not be the default  going forward. Closing this PR without accepting your changes for this reason. Feel free to continue the conversation after I close.

PR #14389, opened 2022-03-15
Environment variable POSTGRES_NAME should be POSTGRES_DB for the db service

alamirfan:
### Proposed changes
POSTGRES_NAME should be POSTGRES_DB for the db service

For postgres service the environment variable name should be POSTGRES_DB if you want to create custom db name


usha-mandya (approved):
LGTM

PR #14391, opened 2022-03-15
update the explanation of ipam in compose file v3

flyingbarbarian:
### Proposed changes

I changed the explanation of ipam which is in Compose file reference Version 3.
I copied the explanation from Compose file reference Version 2.
Because v2 and v3 has the same specification (This is discussed in [this issue](https://github.com/docker/compose/issues/6569) ).

### Related issues (optional)

https://github.com/docker/compose/issues/6569

### DIFF
--- a/compose/compose-file/compose-file-v3.md
+++ b/compose/compose-file/compose-file-v3.md
@@ -2502,6 +2502,11 @@ which is optional:
 -   `config`: A list with zero or more config blocks, each containing any of
     the following keys:
     - `subnet`: Subnet in CIDR format that represents a network segment
+    - `ip_range`: Range of IPs from which to allocate container IPs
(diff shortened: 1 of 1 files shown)


ndeloof:
> v2 and v3 has the same specification

They don't. Just the Compose Specification makes this all obsolete. v3 file format documentation should not be updated, but should be archived ASAP to better advertise the compose specification

flyingbarbarian:
@ndeloof 

Thanks to your advice, I understand that the existence of the Compose Specification.
I understand that the Compose Specification is new format which includes old v3 format and old v2 format.
I understand that compose file should be written ,not referring to [the v3 reference](https://docs.docker.com/compose/compose-file/compose-file-v3/) , but referring to [the Compose Specification reference](https://github.com/compose-spec/compose-spec/blob/master/spec.md).

I really appreciate your help.

PR #14398, opened 2022-03-16
ipvlan: minor grammar fix

mkozjak:
### Proposed changes

Minor grammar fix


usha-mandya:
@mkozjak Thanks for the PR. The existing text looks good. It says "This result is easy access for external facing services as there is no need for port mappings in these scenarios.". I'll have to close your PR without merging for this reason.

mkozjak:
I mean, what "result"? Are you sure this is correct wording? Made no sense to me when I read it.

PR #14413, opened 2022-03-20
Multi-architecture Linux download instructions Compose V2

keunes:
### Proposed changes
Update to [compose/cli-command.md](https://docs.docker.com/compose/cli-command/) to make the download URL agnostic (or rather adapted to) the system architecture. The current/old URL doesn't work on Raspberry Pi 4 with RPi OS 64 bit.

`uname -m` take from the URL in the [V1 instructions](https://docs.docker.com/compose/install/#install-compose).

### Related issues (optional)
https://github.com/docker/compose/issues/8630#issuecomment-1073111846

### DIFF
--- a/compose/cli-command.md
+++ b/compose/cli-command.md
@@ -80,7 +80,7 @@ from the [project release page](https://github.com/docker/compose/releases){:tar
     ```console
     $ DOCKER_CONFIG=${DOCKER_CONFIG:-$HOME/.docker}
     $ mkdir -p $DOCKER_CONFIG/cli-plugins
-    $ curl -SL https://github.com/docker/compose/releases/download/v{{site.compose_v2_version}}/docker-compose-linux-x86_64 -o $DOCKER_CONFIG/cli-plugins/docker-compose
+    $ curl -SL https://github.com/docker/compose/releases/download/v{{site.compose_v2_version}}/docker-compose-linux-$(uname -m) -o $DOCKER_CONFIG/cli-plugins/docker-compose
     ```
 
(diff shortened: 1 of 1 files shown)


compose/cli-command.md
```diff
@@ -80,7 +80,7 @@ from the [project release page](https://github.com/docker/compose/releases){:tar
     ```console
     $ DOCKER_CONFIG=${DOCKER_CONFIG:-$HOME/.docker}
     $ mkdir -p $DOCKER_CONFIG/cli-plugins
-    $ curl -SL https://github.com/docker/compose/releases/download/v{{site.compose_v2_version}}/docker-compose-linux-x86_64 -o $DOCKER_CONFIG/cli-plugins/docker-compose
+    $ curl -SL https://github.com/docker/compose/releases/download/v{{site.compose_v2_version}}/docker-compose-linux-$(uname -m) -o $DOCKER_CONFIG/cli-plugins/docker-compose
```
usha-mandya, line 83: This probably won't work, because the names for architectures used for the downloads use a different format (e.g. aarch64 or armv6); https://github.com/docker/compose/releases

We're working on updating the documentation, because we now also provide .deb and .rpm packages for docker compose v2

keunes: Hi @usha-mandya,
Many thanks for taking a look at this. I am no expert and it may not work for all architectures indeed. My impression is that `x86_64` (e.g. on my laptop, which I'm writing this comment on) as well as `armv7` (e.g. on a RaspberryPi 3/4) are just fine. So while it might not work in every case (e.g. the ones you mention), the adjusted link will already work in more cases than the current link.

But if that's not good enough and you prefer to keep it as is, then for newbies like me it would be helpful to at least have a pointer/reminder to update the architecture in the link. Can save a hope frustration of a not-working installation, I can tell from experience :)

PR #14416, opened 2022-03-21
Move Compose file ref to the legacy node

usha-mandya:

Move Compose file version 2 and 3 to a new legacy node

### DIFF
--- a/_data/toc.yaml
+++ b/_data/toc.yaml
@@ -1040,12 +1040,16 @@ reference:
   section:
     - path: /compose/compose-file/
       title: Compose Specification
-    - path: /compose/compose-file/compose-versioning/
-      title: About versions and upgrading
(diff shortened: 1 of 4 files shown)


glours:
I'm still confused with the Compose file format matrix, does it make sense to keep it for the legacy part only?

usha-mandya:
Yes, that's not required. Also, the I think the page on versioning and upgrading should also move to the Legacy section. WDYT?

glours:
way better 👍  
And the integration of the specification looks nice! Did you made a lot of changes on the specification markdowns?
Thanks @usha-mandya

usha-mandya:
@glours Nothing much. A minor revamp to the overview section, updated the source files in upstream compose-spec and fixed a bunch of broken URLs.

glours (approved):
LGTM

PR #14417, opened 2022-03-21
Add new step 1 "Exit MySQL"

YgorPerez:
### Changes:
Added new step 1 In the "Connect to MySQL" section, "Exit MySQL", and substituted the enumeration of the following steps to 2 and 3

### Why:
If you don't exit MySQL with the command `exit`, you can't proceed to create the container on the next step

### DIFF
--- a/get-started/07_multi_container.md
+++ b/get-started/07_multi_container.md
@@ -110,13 +110,20 @@ address)?
 To figure it out, we're going to make use of the [nicolaka/netshoot](https://github.com/nicolaka/netshoot) container,
 which ships with a _lot_ of tools that are useful for troubleshooting or debugging networking issues.
 
-1. Start a new container using the nicolaka/netshoot image. Make sure to connect it to the same network.
+
+1. Exit MySQL if you haven't already (if you have typed any thing before press `;` and hit enter).
+
+    ```console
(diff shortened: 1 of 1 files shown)


get-started/07_multi_container.md
```diff
@@ -110,13 +110,20 @@ address)?
 To figure it out, we're going to make use of the [nicolaka/netshoot](https://github.com/nicolaka/netshoot) container,
 which ships with a _lot_ of tools that are useful for troubleshooting or debugging networking issues.
 
-1. Start a new container using the nicolaka/netshoot image. Make sure to connect it to the same network.
+
+1. Exit MySQL if you haven't already (if you have typed any thing before press `;` and hit enter).
```
usha-mandya, line 114: @YgorPerez Thanks for the suggestion. Instead of adding the exit step here. Could you add it below line 102 so the **Connect to MySQL** section is cleaner? Please use the following text:


Exit the MySQL shell to return to the shell on our machine.

 ```console
    mysql> exit
 ```

YgorPerez: Sure!

usha-mandya (approved):
LGTM

PR #14423, opened 2022-03-22
Add `,` to json file

YgorPerez:
Without the comma the json code will fail

### DIFF
--- a/language/nodejs/develop.md
+++ b/language/nodejs/develop.md
@@ -149,7 +149,7 @@ To start our application in debug mode, we need to add a line to our `package.js
 Open the `package.json` file and add the following line to the scripts section:
 
 ```json
-  "debug": "nodemon --inspect=0.0.0.0:9229 server.js"
+  , "debug": "nodemon --inspect=0.0.0.0:9229 server.js"
 ```
 
 As you can see, we are going to use nodemon. Nodemon starts our server in debug mode and also watches for files that have changed, and restarts our server. Let’s run the following command in a terminal to install nodemon into our project directory.


usha-mandya:
@crazy-max Is there a better way to add a custom script without manually modifying the json file?

crazy-max:
@usha-mandya [`scripts` is a map](https://docs.npmjs.com/cli/v8/using-npm/scripts) so I think the current assumption to add this script is correct. ` , "debug":` is confusing.

crazy-max:
We could also use jq with something like:

```console
jq '.scripts + {"debug": "nodemon --inspect=0.0.0.0:9229 server.js"}' < package.json
```

But don't think it's worth it.

thaJeztah:
@crazy-max yeah, so we were discussing "how can we make this less error-prone"? We can't show the whole `package.json` because it's generated (and way too large to show) - otherwise we could've shown "this is what your file should look like".

Editing JSON can be error-prone, so I was thinking "perhaps `rpm` has a command to update it" (something like `npm script add --name "debug" --command "the command to run"` so that it would just "work"

I guess the best alternative then would be to call out "make sure the JSON is valid after you edit it" (or something along those lines)

thaJeztah:
> because it's generated (and way too large to show)

Perhaps it's still an option to show just the `scripts` section of the file (if that section always looks the same) to show what it must look like after editing

crazy-max:
Yeah with https://github.com/docker/docker.github.io/pull/14423#issuecomment-1078946631 it would look like this in output:

```json
{
  "build": "ncc build src/main.ts --source-map --minify --license licenses.txt",
  "lint": "eslint src/**/*.ts __tests__/**/*.ts",
  "format": "eslint --fix src/**/*.ts __tests__/**/*.ts",
  "test": "jest --coverage",
  "all": "yarn run build && yarn run format && yarn test",
  "debug": "nodemon --inspect=0.0.0.0:9229 server.js"
}
```

with https://github.com/docker/login-action/blob/master/package.json

crazy-max:
Also if someone has `npx` installed he could:

```console
$ npx npm-add-script \
  -k "debug" \
  -v "nodemon --inspect=0.0.0.0:9229 server.js" \
  --force
```

usha-mandya:
@thaJeztah @crazy-max Thanks for weighing in on this.

@YgorPerez Thanks for the PR. Discussed this PR internally and decided not to go ahead with adding `,` before `debug` as it can cause confusion and scripting this would also mean that we are complicating this. Closing this PR here without accepting your changes for this reason. Feel free to continue the conversation after I close.

PR #14435, opened 2022-03-23
add /go/storage-driver/ redirect

thaJeztah:
Currently pointing to "select a storage driver", but we may be adding
a section about migrating to a different storage driver (for users
that are currently using a deprecated storage driver).

relates to https://github.com/moby/moby/pull/43378 https://github.com/moby/moby/pull/43378#discussion_r826267907

### DIFF
--- a/go/storage-driver.md
+++ b/go/storage-driver.md
@@ -1,0 +1,6 @@
+---
+title: Information on how configure storage drivers
+description: Instructions for configuring (or switching between) storage-drivers.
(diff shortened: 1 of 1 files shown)


usha-mandya (approved):
LGTM

melutovich:
As discussed in https://github.com/moby/moby/pull/43378 are there any better documentation/instructions planned for users that get hit by the daemon not starting up.  With v23 now live, users are already getting hit by this (whom I work with just ran into this)

PR #14448, opened 2022-03-25
Improve the shell download command for arch output

kamaslau:
Automaticly generates file url according to local arch info.

### DIFF
--- a/compose/cli-command.md
+++ b/compose/cli-command.md
@@ -80,7 +80,7 @@ from the [project release page](https://github.com/docker/compose/releases){:tar
     ```console
     $ DOCKER_CONFIG=${DOCKER_CONFIG:-$HOME/.docker}
     $ mkdir -p $DOCKER_CONFIG/cli-plugins
-    $ curl -SL https://github.com/docker/compose/releases/download/v{{site.compose_v2_version}}/docker-compose-linux-x86_64 -o $DOCKER_CONFIG/cli-plugins/docker-compose
+    $ curl -SL "https://github.com/docker/compose/releases/download/v{{site.compose_v2_version}}/docker-compose-linux-$(uname -m)" -o $DOCKER_CONFIG/cli-plugins/docker-compose
     ```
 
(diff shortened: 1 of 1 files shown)


compose/cli-command.md
```diff
@@ -80,7 +80,7 @@ from the [project release page](https://github.com/docker/compose/releases){:tar
     ```console
     $ DOCKER_CONFIG=${DOCKER_CONFIG:-$HOME/.docker}
     $ mkdir -p $DOCKER_CONFIG/cli-plugins
-    $ curl -SL https://github.com/docker/compose/releases/download/v{{site.compose_v2_version}}/docker-compose-linux-x86_64 -o $DOCKER_CONFIG/cli-plugins/docker-compose
+    $ curl -SL "https://github.com/docker/compose/releases/download/v{{site.compose_v2_version}}/docker-compose-linux-$(uname -m)" -o $DOCKER_CONFIG/cli-plugins/docker-compose
```
thaJeztah, line 83: This probably won't work, because the names for architectures used for the downloads use a different format (e.g. `aarch64` or `armv6`); https://github.com/docker/compose/releases


We're working on updating the documentation, because we now also provide `.deb` and `.rpm` packages for docker compose v2

kamaslau: Can we try wrap up the file name generation (platform, arch, file extension, etc.) progress in a installation shell file, like that of [grafana](https://raw.githubusercontent.com/grafana/agent/release/production/grafanacloud-install.sh)?

kamaslau: Also I've noticed that $(uname -s) outputs titlized "Linux" instead of "linux" that is used in actual downloadable file name.

thaJeztah: I don't think we should overload this, and perhaps should even consider simplifying the instructions to tell users where they can find the downloads, instead of trying to script it. These instructions will likely go away soon (when we rewrite the instructions to use the deb/rpm packages), or at least will be rewritten.

kamaslau: Well it certainly is a down-to-the-point way to just direct users to the warehouse, but a installation shell script could really save developers a bunch of time. Either way, it would be a nice improvement from what we have now. Thanks.

PR #14457, opened 2022-03-28
add a workflow to create a PR when Compose is released

glours:
### Proposed changes

The Compose reference API is owned by the [`docker/compose` repository](https://github.com/docker/compose), this PR add a workflow to create a PR to automatically update the upstream Compose API documentation


.github/workflows/create-pr-compose-release.yml
```diff
@@ -0,0 +1,42 @@
...
+          rm -rf /.ComposeReleaseSource
+          git add -A .
+      - name: Create PR
+        uses: peter-evans/create-pull-request@v4
+        with:
+          commit-message: Update Compose reference API documentation to ${{ github.event.inputs.release }}
+          signoff: true
+          branch: update-compose-doc-${{ github.event.inputs.release }}
+          delete-branch: true
+          title: Update Compose reference API doc - ${{ github.event.inputs.release }}
```
usha-mandya, line 33: @glours Is this CLI reference or API reference docs?

glours: This one https://github.com/docker/compose/tree/v2/docs/reference

usha-mandya:
@glours Should we also automatically bump the version number in https://github.com/docker/docker.github.io/blob/master/_config.yml#L26 as part of this PR? WDYT?

glours:
@usha-mandya I need to completely review this PR, currently I doesn't update the right documentation files and yes it should also update the compose version in the `_config.yml` file

.github/workflows/create-pr-compose-release.yml
```diff
@@ -0,0 +1,43 @@
...
+          repository: docker/compose
+          ref: ${{ github.event.inputs.release }}
+          path: ./ComposeReleaseSource
+      - name: Prepare PR
+        run: |
+          rm -rf ./_data/compose-cli/*
+          cp -rf /.ComposeReleaseSource/docs/reference/*.yaml ./_data/compose-cli/
+          rm -rf /.ComposeReleaseSource
+          sed '/compose_version:/s/\".*\"/\"${{ github.event.inputs.release }}\"/g' _config.yml
+          git add -A .
```
crazy-max, line 26: I think we could enhance this behavior and build from a remote reference instead so the logic in "Prepare PR" would be carry by compose repo.

Would be pretty much like this:

```yaml
name: Create a automatic PR on Compose Release
description: create a Pull Request with doc changes when a new version of Docker Compose is released

on:
  repository_dispatch:
    types: [compose_ref]

jobs:
  create_compose_doc_pr:
    runs-on: ubuntu-latest
    steps:
      -
        name: Checkout
        uses: actions/checkout@v3
      -
        name: Fetch content
        uses: docker/build-push-action@v3
        with:
          context: ${{ github.event.inputs.ref }}
          target: docs-dispatch
          output: ./_data/compose-cli
      - 
        name: Commit
        run: |
          git add -A .
      -
        name: Create PR
        uses: peter-evans/create-pull-request@v4
        with:
          commit-message: Update Compose reference API documentation to ${{ github.event.inputs.release }}
          signoff: true
          branch: update-compose-doc-${{ github.event.inputs.release }}
          delete-branch: true
          title: Update Compose reference API doc - ${{ github.event.inputs.release }}
          body: |
            Update the Compose reference API documentation to keep in sync with the latest release ${{ github.event.inputs.release }}
            
            - Auto-generated by [create-pull-request][1]
            [1]: https://github.com/peter-evans/create-pull-request
          labels: area/Compose
          draft: false
```

.github/workflows/create-pr-compose-release.yml
```diff
@@ -0,0 +1,43 @@
...
+      - name: Create PR
+        uses: peter-evans/create-pull-request@v4
```
crazy-max, line 28: For security purpose, pin external actions to a commit sha.

```suggestion
        uses: peter-evans/create-pull-request@f094b77505fb89581e68a1163fbd2fffece39da1 # v4.0.3
```

.github/workflows/create-pr-compose-release.yml
```diff
@@ -0,0 +1,43 @@
...
+        with:
+          commit-message: Update Compose reference API documentation to ${{ github.event.inputs.release }}
+          signoff: true
+          branch: update-compose-doc-${{ github.event.inputs.release }}
```
crazy-max, line 32: we should probably use an uuid as suffix just in case.

crazy-max (commented):
Overall the idea looks good. Need to discuss about one of my comments to mutualize and move some logic upstream to ease maintainability.

dockertopia (commented):
The only thing I would add as per offline discussion is that it would be great to have the same summary for the release we can see in the repo release page in the [docs](https://docs.docker.com/compose/release-notes/). The info would always get there on time and we would always be quicker to improve it when necessary with that starting point.

PR #14468, opened 2022-03-30
desktop/linux: Add instructions to remove old systemd units

p1-0tr:
Signed-off-by: Piotr Stankiewicz <piotr.stankiewicz@docker.com>


aiordache (approved):

desktop/linux/index.md
```diff
@@ -148,6 +148,8 @@ At the end of the installation process, `apt` displays an error due to installin
   N: Download is performed unsandboxed as root, as file '/home/user/Downloads/docker-desktop.deb' couldn't be accessed by user '_apt'. - pkgAcquire::Run (13: Permission denied)
   ```
 
+If you have installed one of the previous releases and install the new package over it (as opposed to removing the old package explicitly), you will have to make sure that `~/.config/systemd/user/docker-desktop.service` and `~/.local/share/systemd/user/docker-desktop.service` are removed.
```
usha-mandya, line 151:
```suggestion
If you have installed one of the previous releases and reinstall the new package over it (as opposed to removing the old package explicitly), you need to make sure that `~/.config/systemd/user/docker-desktop.service` and `~/.local/share/systemd/user/docker-desktop.service` are removed.
```

PR #14476, opened 2022-03-31
Publish updates from master

usha-mandya:


desktop/linux/index.md
```diff
@@ -107,6 +101,60 @@ Alternatively, open a terminal and run:
...
+
+You can intialize pass by using a gpg key. To generate a gpg key, run:
+
+``` console
+$ gpg --generate-key
+...
+GnuPG needs to construct a user ID to identify your key.
+
+Real name: Molly
+Email address: molly@docker.com
```
thaJeztah, line 118: We should use an `@example.com`  or `@example.org` domain for this; e-mail addresses `@docker.com` may actually land in some mail-boxes (and `example.com` is intended for this purpose.

desktop/linux/index.md
```diff
@@ -107,6 +101,60 @@ Alternatively, open a terminal and run:
...
+mkdir: created directory '/home/molly/.password-store/'
+Password store initialized for 7865BA9185AFA2C26C5B505669FC4F36530097C2
+```
+
+After signing in on the Docker Dashboard, you can check whether the login credentials are stored in the password store.
+
+```console
+molly@ubuntu:~$ docker-credential-desktop list
+{"https://index.docker.io/v1/":"molly","https://index.docker.io/v1//refresh-token":"v1.Mb66i2rHIBOjNVkKpKpv8zYQU-gwtNam_5RbYZmsUCQ_smw2WiliFrJlUw5rszR947C2iZ4QGAJ1g5opK1URwaY","https://index.docker.io/v1//user":"molly"}
+```
```
thaJeztah, line 143: Wondering if this step is needed, or if the dashboard would already show a message if it didn't work (running `docker-credential-desktop list` can show other credentials stored in `pass`, so usually not something you want to have printed in your shell.

usha-mandya: @aiordache WDYT?

thaJeztah: I was a bit on the fence; on one hand, it's "ok" to check that it works through this way, on the other hand, the `docker-credential-desktop` helper is not directly intended to be used manually (and when used wrong can be "fun"; e.g. yesterday I accidentally ran `docker-credential-desktop get`, and got stuck, because it expects input on `stdin`, and didn't handle signals to terminate 😂 


I'm wondering though if the token that's shown in your example actually _works_ on the command-line; I think we were testing yesterday, and the CLI can currently only function with a PAT (or regular password), but didn't work when trying to use the refresh token.

thaJeztah (approved):
LGTM, thanks @aiordache !

PR #14478, opened 2022-03-31
[WIP] test engine cli reference changes

thaJeztah:
This updates the yamldocs for the engine cli reference with updated
sources that include anchor tags for flags; this should make the
flags in the "options" table clickable and scroll to the relevant
section on the page (where present).

This was built from https://github.com/docker/cli/pull/3509, which
is not yet merged (and has changes for the upcoming engine release,
so cannot yet be published here).

### DIFF
--- a/_data/engine-cli/docker_attach.yaml
+++ b/_data/engine-cli/docker_attach.yaml
@@ -1,5 +1,6 @@
 command: docker attach
-short: Attach local standard input, output, and error streams to a running container
+short: |
+  Attach local standard input, output, and error streams to a running container
 long: |-
   Use `docker attach` to attach your terminal's standard input, output, and error
   (or any combination of the three) to a running container using the container's
@@ -68,6 +69,7 @@ options:
   value_type: string
   description: Override the key sequence for detaching a container
   deprecated: false
+  hidden: false
   experimental: false
   experimentalcli: false
   kubernetes: false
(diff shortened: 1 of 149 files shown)


thaJeztah:
ok; looks like we need to update the compose-file reference to account for the new anchors (and/or add more "custom" anchors to make those old links work if we want to);

```
- ./_site/compose/compose-file/compose-file-v2/index.html

  *  linking to internal hash #specify-isolation-technology-for-container---isolation that does not exist (line 466)
     <a href="/engine/reference/commandline/run/#specify-isolation-technology-for-container---isolation">Docker Engine docs</a>
  *  linking to internal hash #specify-isolation-technology-for-container---isolation that does not exist (line 1080)
     <a href="/engine/reference/commandline/run/#specify-isolation-technology-for-container---isolation">Docker Engine docs</a>
- ./_site/compose/compose-file/compose-file-v3/index.html
  *  linking to internal hash #configure-namespaced-kernel-parameters-sysctls-at-runtime that does not exist (line 2091)
     <a href="/engine/reference/commandline/run/#configure-namespaced-kernel-parameters-sysctls-at-runtime">configure namespaced kernel parameters (sysctls) at runtime</a>
...
```

thaJeztah:
Alright; this works; it's not _perfect_ yet, as there's many flags that don't have a dedicated section (yet), and there's some flags that are described in combination with other flags (e.g. grouping `--add-host` and `--rm-host` in the same paragraph), but it's  a start.


This page, for example, now has anchors for various flags; https://deploy-preview-14478--docsdocker.netlify.app/engine/reference/commandline/build/#options

We should also have a look at making the links more _visible_ in the Options table, as currently it's not clear which ones are clickable, and which ones are not.

Issue #14195, opened 2022-01-31
Documentation is running in Quirks Mode


ForbiddenEra:
### Problem description

The documentation appears to be running in Quirks Mode, at least in Firefox. This is due to there being a comment with the page generation date preceeding the `<!DOCTYPE html>` which is disallowed for HTML5 standard.

Just happened to have devtools open while browsing and noticed this.

![image](https://user-images.githubusercontent.com/7980490/151769327-e516c0cf-8d68-4046-8932-e79cec0566dc.png)

danBamikiya:
same for Microsoft Edge:

![image](https://user-images.githubusercontent.com/58262528/152686587-7286f142-430c-481d-98df-c05c1372d6da.png)

Is that really a problem tho? I read about [Quirks Mode](https://en.wikipedia.org/wiki/Quirks_mode) and it doesn't seem to have any disadvantages for the documentation.

thaJeztah:
Thanks for reporting; I knew "old browsers" were picky about this, but didn't know current browsers still did. I opened a PR to move the comment after the doctype, which should (hopefully) fix this; https://github.com/docker/docker.github.io/pull/14484

Preview of that PR is available at https://deploy-preview-14484--docsdocker.netlify.app

PR #14484, opened 2022-03-31
move "page generated" comment to prevent rendering in quirks mode

thaJeztah:
fixes https://github.com/docker/docker.github.io/issues/14195

As described on https://en.wikipedia.org/wiki/Quirks_mode, a comment
preceding the `<!DOCTYPE html>` can cause some browsers to render the
HTML in quirks mode.

This patch moves the comment after the doctype to prevent this.

### DIFF
--- a/_layouts/docs.html
+++ b/_layouts/docs.html
@@ -1,10 +1,10 @@
+<!DOCTYPE html>
 <!-- Page generated {{ site.time }}
 {%- if page.edit_url -%}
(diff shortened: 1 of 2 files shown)


thaJeztah:
@usha-mandya @jerae-duffin PTAL

jerae-duffin (approved):
Thanks @thaJeztah LGTM!

Issue #14146, opened 2022-01-18
Extra Dot


afbarksd:
[get-started/06_bind_mounts.md](https://docs.docker.com/get-started/06_bind_mounts/)

I think that "docker build -t getting-started .." should be "docker build -t getting-started .".

thaJeztah:
Thanks for reporting; yes, that example is formatted a bit awkward as it's part of a sentence; I opened https://github.com/docker/docker.github.io/pull/14485 to fix the formatting

PR #14485, opened 2022-03-31
reformat "get started" 06 - bind-mounts

thaJeztah:
fixes https://github.com/docker/docker.github.io/issues/14146

The "docker build" example was written as an in-line example, but directly
followed by a full-stop at the end of the sentence, which made it easy to
confuse the extra `.` to be part of the command to run.

This moves it to a code block to allow the example to be highlighted, and
less confusing.

Before this patch:

Screenshot 2022-03-31 at 23 54 37

With this patch:

Screenshot 2022-04-01 at 00 00 27


thaJeztah:
@usha-mandya @jerae-duffin PTAL

jerae-duffin (approved):
LGTM

PR #14495, opened 2022-04-02
WEB-1041: Adds TrustArc CMP tool to the footer area

jhratliff:
### Proposed changes
Adds the TrustArc Consent Management tool to the footer of the site.

### Screenshot
image

### Related issues (optional)
Addressses WEB-1041 
https://docker.atlassian.net/browse/WEB-1041

### DIFF
--- a/_includes/consent-management/trustarc.html
+++ b/_includes/consent-management/trustarc.html
@@ -1,0 +1,1 @@
+<script async="async" src="//consent.trustarc.com/notice?domain=docker.com&c=teconsent&js=nj&noticeType=bb&gtm=1"></script>
--- a/_includes/head.html
+++ b/_includes/head.html
@@ -33,6 +33,9 @@
   {%- if page.sitemap == false or site.GH_ENV == "gh_pages" %}
   <meta name="robots" content="noindex"/>
   {%- endif %}
+  {%- if jekyll.environment != 'production' -%}<!-- TrustArc CMP -->
+  <script async="async" src="//consent.trustarc.com/notice?domain=docker.com&c=teconsent&js=nj&noticeType=bb&gtm=1"></script>
+  {%- endif -%}
   {%- if jekyll.environment == 'production' and site.google_analytics != '' -%}{%- include analytics/google_analytics.html GOOGLE_ID=site.google_analytics -%}{%- endif -%}
   <title>{{ page.title | default: page_title }} | Docker Documentation</title>
   <meta name="description" content="{{ page.description | default: page_description | escape}}" />
--- a/_includes/head.html
+++ b/_includes/head.html
@@ -33,9 +33,7 @@
   {%- if page.sitemap == false or site.GH_ENV == "gh_pages" %}
   <meta name="robots" content="noindex"/>
   {%- endif %}
-  {%- if jekyll.environment != 'production' -%}<!-- TrustArc CMP -->
   <script async="async" src="//consent.trustarc.com/notice?domain=docker.com&c=teconsent&js=nj&noticeType=bb&gtm=1"></script>
-  {%- endif -%}
   {%- if jekyll.environment == 'production' and site.google_analytics != '' -%}{%- include analytics/google_analytics.html GOOGLE_ID=site.google_analytics -%}{%- endif -%}
   <title>{{ page.title | default: page_title }} | Docker Documentation</title>
   <meta name="description" content="{{ page.description | default: page_description | escape}}" />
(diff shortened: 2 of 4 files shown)


_includes/head.html
```diff
@@ -33,6 +33,7 @@
   {%- if page.sitemap == false or site.GH_ENV == "gh_pages" %}
   <meta name="robots" content="noindex"/>
   {%- endif %}
+  <script async="async" src="//consent.trustarc.com/notice?domain=docker.com&c=teconsent&js=nj&noticeType=bb&gtm=1"></script>
```
thaJeztah, line 36: We should probably make this conditional, so that it doesn't render when building the docs locally, when previewing, or if someone would run the docs elsewhere. (similar to what we do for the google analytics the line below

thaJeztah:
Also wondering; 

- is the large "button" style something we can influence? Wondering if we can make it smaller and/or better fit in the style of the website

- I noticed the same on our www; looks like the styling doesn't work well with our own styles (before accepting cookies); the banner moves outside of the window, and has a margin at the left;

Screenshot 2022-04-04 at 21 25 14

_includes/head.html
jhratliff: I'll update the PR with this change. Thanks for the input!

jhratliff:
@thaJeztah - with the production flag requirement in place, what is the best way to test the changes before deploying the code to production?

thaJeztah:
@jhratliff you can build the docs locally with the `JEKYLL_ENV=production` env-var set; see https://github.com/docker/docker.github.io/blob/c38f20e7638e43cf49059130901d37b4acb70fb3/docker-compose.yml#L3-L9


(looks like we should update those instructions to use `docker compose` (v2) instead of `docker-compose` 😅)

```bash
JEKYLL_ENV=production docker compose up --build
```

When that's built and running, you can try it in your browser at `http://localhost:4000`

thaJeztah:
Oh, just realised the instructions are in `CONTRIBUTING.MD` as well; https://github.com/docker/docker.github.io/blob/master/CONTRIBUTING.md#build-the-docs-with-deployment-features-enabled

jhratliff:
@thaJeztah - yes, saw both of those and was able to test the results locally. Is there a way to test the production-flagged build ahead of it going out to production? (I'll have another commit to update some of the style concerns)

thaJeztah:
So, the short answer is "no, we don't"; we do have a https://docs-stage.docker.com, but it does not have the "production" bits enabled (to prevent it being indexed, but also because it would sometimes trip up analytics and GTM).

I think we had a protected staging website in the past, but in practice, it didn't add much value, as the environment being not _exactly_ as "production" (VPN or password required, different domain) still didn't make it a full representation of docs.docker.com.


> I find it hard to test the user experience on local trying to emulate an IP address of a European origin. Since it's localhost... the IP address doesn't seem to trigger the same experience we'd expect for international users.

Silly question; does that mean that people from outside the EU cannot opt-out of cookies? I realize EU _requires_ it, and USA (e.g.) not, but wondering if this is configurable, and we can give people outside of the EU the same service if they don't want to be tracked.

thaJeztah:
I can quickly spin up a machine with docker and run it though; give me 3 minutes

_includes/consent-management/trustarc.html
```diff
@@ -0,0 +1 @@
+<script async="async" src="//consent.trustarc.com/notice?domain=docker.com&c=teconsent&js=nj&noticeType=bb&gtm=1"></script>
```
thaJeztah, line 1: Looks like this can only work when hosted on a `docker.com` domain; the script does;

```
(new Image(1,1)).src=("https://consent.trustarc.com/log".replace("http:","https:"))+"?domain=docker.com&country=nl&state=&behavior=implied&c="+
```

jhratliff: The new testing site looks like it would need to be HTTPS for this to work, right?

jhratliff: Answering a question above - people in EU have a different compliance standard. Everybody gets to opt out, EU is a bit more stringent - and California has their own compliance requirements as well. The IP address of the user determines which flow they experience.

jhratliff: @thaJeztah - do we have a way to set up the staging / test area as HTTPS so we can get the script to load?

thaJeztah: (for those reading along) done ☝️ (I sent a link through slack)

crazy-max:
housekeeping

PR #14510, opened 2022-04-06
added misssing "driver"

Mandraenke:
The example for setting the MTU on the swarm ingress network uses the wrong property:

`--opt com.docker.network.mtu=1200`

instead of

`--opt com.docker.network.driver.mtu=1200`


thaJeztah (approved):
LGTM, thanks!

PR #14516, opened 2022-04-06
Add initial SBOM docs

chris-crone:
### Proposed changes

Add initial `docker sbom` docs. It would be great to add some practical examples in a follow up like diffing images or checking for vulnerable libraries.

### Unreleased project version (optional)

The command will be available with Docker Desktop 4.7.0.

### Related issues (optional)

N/A

PR #14524, opened 2022-04-07
Terminate the console fix-width example for --output on the SBOM page

halprin:
### Proposed changes

Terminating the `--output` flag example on the [SBOM page](https://docs.docker.com/engine/sbom/#output-formatting-and-saving-outputs).  It currently isn't showing up correct in the documentation as seen below.

Screen Shot 2022-04-07 at 3 59 52 PM

_N/A._

### Related issues (optional)

Related to #14516 which created this documentation.


thaJeztah (approved):
LGTM, thanks!

PR #14530, opened 2022-04-08
Fix homebrew paths to work with Apple silicon macs as well.

yogurtearl:
### Proposed changes

on Apple silicon, `$(brew --prefix)` is `/opt/homebrew` and not `/usr/local`
This change makes it work on intel and apple silicon.

This is already done in the bash examples above in the doc. 

see https://docs.brew.sh/Installation


aevesdocker:
Closing ticket as page no longer exists/information has changed.

PR #14544, opened 2022-04-13
Explain important caveat in htpasswd tutorial

raxod502:
### Proposed changes

Documentation improvement. It took me almost an hour to track down https://github.com/distribution/distribution/issues/2817#issuecomment-762736783 as the reason my private Docker registry was not working, because of the lack of any error message pointing to the cause.

### Related issues (optional)

Relates to https://github.com/distribution/distribution/issues/2817, but does not really solve it. In my opinion, a full solution to that issue would involve Docker Registry outputting a relevant error message when the incorrect credential type is used, rather than silently accepting it and then failing all login attempts.

### DIFF
--- a/registry/deploying.md
+++ b/registry/deploying.md
@@ -402,6 +402,12 @@ secrets.
 > authentication to work.
 {:.warning}
 
+> **Warning**
+> The official registry image **only** supports htpasswd credentials in
+> bcrypt format, so if you omit the `-B` option when generating the credential
+> using htpasswd, all authentication attempts will fail.
+{:.warning}
+
 1.  Create a password file with one entry for the user `testuser`, with password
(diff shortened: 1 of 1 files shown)


milosgajdos:
Hey, @raxod502. Thanks for the PR. This is documented in https://github.com/distribution/distribution/blob/main/docs/deploying.md.

raxod502:
I see - but it also appears there is a version here. Is this one generated from the other one?

raxod502:
Oh, never mind, I see your comment in https://github.com/distribution/distribution/issues/2817#issuecomment-1097691211. OK, will submit PR against the right repo.

raxod502:
Filed https://github.com/distribution/distribution/pull/3630.

PR #14548, opened 2022-04-13
Expose level 4 headings in the right-hand toc

aevesdocker:
Changes to the way the right hand toc is displayed. Now displays level 4 headings so users will be able to see all the information that is covered in a page.


jerae-duffin (approved):
LGTM!

PR #14572, opened 2022-04-19
desktop: windows: fix highlighting of some command-line examples

thaJeztah:
Before:

Screenshot 2022-04-19 at 11 58 37

After:

Screenshot 2022-04-19 at 12 01 55

### DIFF
--- a/desktop/windows/install.md
+++ b/desktop/windows/install.md
@@ -124,22 +124,24 @@ Log out and log back in for the changes to take effect.
 
 ### Install from the command line
 
-After downloading **Docker Desktop Installer.exe**, run the following command in a terminal to install Docker Desktop:
+After downloading **Docker Desktop Installer.exe**, open a terminal, and change
+to the directory in which you downloaded the installer, then run the following
+command to install Docker Desktop:
 
-```
-"Docker Desktop Installer.exe" install
+```console
+C:\> "Docker Desktop Installer.exe" install
 ```
 
 If you’re using PowerShell you should run it as:
 
-```
-Start-Process '.\win\build\Docker Desktop Installer.exe' -Wait install
+```console
+PS> Start-Process '.\Docker Desktop Installer.exe' -Wait install
 ```
 
 If using the Windows Command Prompt:
 
-```
-start /w "" "Docker Desktop Installer.exe" install
+```console
+C:\> start /w "" "Docker Desktop Installer.exe" install
 ```
 
 The install command accepts the following flags:
@@ -149,8 +151,9 @@ The install command accepts the following flags:
 - `--backend=<backend name>`: selects the backend to use for Docker Desktop, `hyper-v` or `wsl-2` (default)
 
 If your admin account is different to your user account, you must add the user to the **docker-users** group:
-```
-net localgroup docker-users <user> /add
+
+```console
+C:\> net localgroup docker-users <user> /add
 ```
 
 ## Start Docker Desktop
--- a/_includes/configure-registry-json.md
+++ b/_includes/configure-registry-json.md
@@ -9,7 +9,9 @@ Desktop and access all their organizations.
 
 ### Windows
 
-On Windows, run the following command in a terminal to install Docker Desktop:
+After downloading **Docker Desktop Installer.exe**, open a terminal, and change
+to the directory in which you downloaded the installer, then run the following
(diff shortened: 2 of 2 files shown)


desktop/windows/install.md
```diff
@@ -126,20 +126,20 @@ Log out and log back in for the changes to take effect.
...
+```console
+C:\> "Docker Desktop Installer.exe" install
 ```
 
 If you’re using PowerShell you should run it as:
 
-```
-Start-Process '.\win\build\Docker Desktop Installer.exe' -Wait install
+```console
+PS> Start-Process '.\win\build\Docker Desktop Installer.exe' -Wait install
```
thaJeztah, line 136: Is the `.\win\build\` actually correct here? I assume a user would download it to their current directory, not a `win/build` subdirectory 🤔

thaJeztah: Also slightly confused, as there seem to be 3 examples and both the first and last seem to be for `cmd.exe`, or is the third (`start /w ....`) for a different scenario?

mat007: Oh good catch, no we should remove the `.\win\build\` prefix!

mat007: As for the use of `start` for when using the Windows Command shell, it’s because the installer is actually a GUI application (as opposed to a Console application, see e.g. https://stackoverflow.com/a/7613999/1195832).
It’s a super annoying trait of Windows applications, and here it means that once invoked the installer process gets detached and the shell goes back to the prompt right away, but the logs are still written to the shell console…
`start` is a way to get around that (`/w` means «wait for the application to exit»)…

thaJeztah:
@mat007 made some updates to remove the `build/`; I'm still a bit confused about the _three_ examples (should the first one be removed?) i.e. these three (is the first one the same as the last one? should we remove the first one?);

```console
C:\Users\Admin> "Docker Desktop Installer.exe" install
```

```console
PS> Start-Process '.\Docker Desktop Installer.exe' -Wait install
```

```console
C:\Users\Admin> start /w "" "Docker Desktop Installer.exe" install
```

mat007 (approved):
Looks good, thanks!

desktop/windows/install.md
```diff
@@ -124,22 +124,24 @@ Log out and log back in for the changes to take effect.
...
 
-After downloading **Docker Desktop Installer.exe**, run the following command in a terminal to install Docker Desktop:
+After downloading **Docker Desktop Installer.exe**, open a terminal, and change
+to the directory in which you downloaded the installer, then run the following
+command to install Docker Desktop:
 
-```
-"Docker Desktop Installer.exe" install
+```console
+C:\> "Docker Desktop Installer.exe" install
```
mat007, line 132: Ah here, I don’t think we should prefix the command with any shell prompt, as it seems to imply it’s the Windows Command prompt whereas it’s more a «here is the command to run in a generic manner». Then the two next examples are for specific terminals.

desktop/windows/install.md
```diff
@@ -149,8 +151,9 @@ The install command accepts the following flags:
 - `--backend=<backend name>`: selects the backend to use for Docker Desktop, `hyper-v` or `wsl-2` (default)
 
 If your admin account is different to your user account, you must add the user to the **docker-users** group:
-```
-net localgroup docker-users <user> /add
+
+```console
+C:\> net localgroup docker-users <user> /add
```
mat007, line 156: Here again, if we prefix with a command prompt it seems to say it’s not for e.g. PowerShell.
But actually the same command runs fine in PowerShell or even WinGW, …

desktop/windows/install.md
mat007: Also it doesn’t really have to at the root of the C:, it could be in any folder.

thaJeztah: The highlighter expects a prompt to be there (and we use prompts in all our examples), but yeah, it's tricky for powershell and command.exe, as "what to use" as prompt (as it defaults to have the path in there); on Linux we just omit that part (`$`), but what should we use for `command.exe` examples (or generic examples for Windows?)

mat007: I see. Then I suppose `C:\>` is fine for Windows Command prompt examples, at least users are probably familiar enough with it that it should be clear enough what this means.

PR #14576, opened 2022-04-19
troubleshooting: include the support form link in terminal section

djs55:
### Proposed changes

If Docker Desktop doesn't start properly, to file a support ticket the user needs to
1. gather diagnostics from the "Troubleshooting from the terminal" section
2. open the link to the support form which is elsewhere on the page.

This patch includes the link to the support form in the "Troubleshooting from the terminal" section to make it more complete.


thaJeztah (approved):
LGTM

PR #14593, opened 2022-04-22
Desktop: improve enable kubernetes section

djs55:
### Proposed changes

I'd like to link to the Kubernetes section to show how easy it is to enable. This PR fixes a few issues:

1. the step about the `docker-desktop` context can't be done until after Kubernetes has been enabled, because the context doesn't exist until then. So it can't be done as a pre-req, it has to be checked afterwards. (Note that you might have an old `docker-desktop` context lying around if you have ever enabled Kubernetes before, but new users won't have one)
2. to see Kube docker containers you need `docker ps` (or `docker container ls`) not `docker service ls`

and makes a few simplifications:

1. the note about seeing internal Kube containers is an advanced option so doesn't need to be presented until later -- this keeps the enable Kubernetes section small (reflecting how easy it is)
2. the note about managing conflicts with minikube is an uncommon case, so can be pushed to the kubectl section

### DIFF
--- a/desktop/kubernetes.md
+++ b/desktop/kubernetes.md
@@ -39,8 +39,12 @@ The status of Kubernetes shows in the Docker menu and the context points to
 
 ## Use the kubectl command
 
-The Kubernetes client command `kubectl` is included and configured to connect
-to the local Kubernetes server. If you have already installed `kubectl` and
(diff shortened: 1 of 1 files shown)


usha-mandya (approved):
LGTM. Thank you @djs55

PR #14601, opened 2022-04-25
fix "docker-compose" command

tigercosmos:


StefanScherer:
Thanks for the contribution. This is intended to use Compose V2 here with `docker compose up`. Current versions of Docker Desktop for Mac, Windows, and also Linux have the compose command as CLI plugin installed, and also newer version of the Linux engine bring a docker-compose-plugin package where you have the new Compose V2 as a `docker compose` command.
I'm closing your PR.

PR #14606, opened 2022-04-26
Clarification of ambiguous port mapping

imrangthub:
Clarification of ambiguous port mapping


craig-osterhout:
Thank you for contributing. The description was updated in another update and now explains which port is which.

PR #14608, opened 2022-04-26
Add leading `v` to ver number in env docs

devidw:
Other version string have the `v`, this one doesn't, this way it's equal in the block


usha-mandya (approved):
LGTM. Thank you @devidw :)

usha-mandya (approved):

PR #14622, opened 2022-04-27
Capitalizing the proper noun / product name

himythili:
'Docker Registry' instead of 'Docker registry'

### Proposed changes
'Docker Registry' instead of 'Docker registry' as Docker Registry is a product name


usha-mandya (approved):
Thank you @himythili :)

PR #14623, opened 2022-04-27
[Do not merge until 11 May] DKP-168 Remove support for Windows versions which are not supported by Microsoft anymore

mat007:
### Proposed changes

Remove support for obsolete Windows versions, according to https://docs.microsoft.com/en-us/windows/release-health/supported-versions-windows-client#servicing-channels

The actual end of support date for these is May 10, 2022. Do we want to wait until then before merging? @usha-mandya @stephen-turner ?

### Related issues (optional)

https://docker.atlassian.net/browse/DKP-168

### DIFF
--- a/desktop/windows/install.md
+++ b/desktop/windows/install.md
@@ -41,7 +41,7 @@ Your Windows machine must meet the following requirements to successfully instal
 ### WSL 2 backend
 
 - Windows 11 64-bit: Home or Pro version 21H2 or higher, or Enterprise or Education version 21H2 or higher.
-- Windows 10 64-bit: Home or Pro 2004 (build 19041) or higher, or Enterprise or Education 1909 (build 18363) or higher.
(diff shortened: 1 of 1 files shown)


usha-mandya:
@mat007 Thanks for the update. I'd suggest we wait until the actual end of servicing date as it's less than two weeks away from now.

stephen-turner (changes requested):
The edit is good, but we need to make the parallel change in the Hyper-V section too.

stephen-turner:
> @mat007 Thanks for the update. I'd suggest we wait until the actual end of servicing date as it's less than two weeks away from now.

+1

mat007:
> The edit is good, but we need to make the parallel change in the Hyper-V section too.

Oh good catch! 🤦 
Fixed!

stephen-turner (approved):
Thanks, approved the edit now, but let's hold until May 11.

ebriney (approved):
LGTM

PR #14624, opened 2022-04-27
Update install.md

FierceWindV:
Modify the docker compose version test command.


dockertopia:
@FierceWindV thanks for the PR. The page is following the latest syntax for Compose which is v2. Take a look at the note at the top of this page: https://docs.docker.com/compose/reference/. Cheers.

PR #14626, opened 2022-04-27
Update 06_bind_mounts.md

sexybug:
### Proposed changes

When I run the powershell command it gives an error like this:

> error An unexpected error occurred: "https://registry.yarnpkg.com/accepts/-/accepts-1.3.7.tgz: connect ECONNREFUSED 104.16.25.35:443".
> info If you think this is a bug, please open a bug report with the information provided in "/app/yarn-error.log".
> info Visit https://yarnpkg.com/en/docs/cli/install for documentation about this command.

I finally found out that the reason is that the npm resource repository cannot be accessed normally in China. 
Changing the resource repository solves.

### DIFF
--- a/get-started/06_bind_mounts.md
+++ b/get-started/06_bind_mounts.md
@@ -57,6 +57,15 @@ So, let's do it!
         node:12-alpine `
         sh -c "yarn install && yarn run dev"
     ```
+    
+    For Chinese developers, you may need to modify the resource repository to download yarn dependencies, the command needs to be like this:
+
+    ```powershell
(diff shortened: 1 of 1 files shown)


craig-osterhout:
Hi @sexybug, 
Thanks for getting involved and sharing the solution. 
Adding this information to the forums may be the best place for this currently as we do not have localized content. We try our best to direct users to the Community Slack and Forums when they encounter errors related to their environment and need support. If you're able to, please help other users by adding the solution at https://forums.docker.com/

PR #14631, opened 2022-04-28
change extensions from preview to beta

aevesdocker:


usha-mandya:
@thaJeztah The HTML proofer is flagging broken links from CLI reference docs to API docs though these files aren't updated as part of the current PR. I recall that we had a similar issue before. Could you PTAL?

usha-mandya (approved):
LGTM

PR #14638, opened 2022-04-29
update Compose version to v2.5.0 and reference Compose CLI documentation

glours:
### Proposed changes

Update Compose version and reference CLI documentation to match `v2.5.0` release of Docker Compose

### DIFF
--- a/_data/compose-cli/docker_compose.yaml
+++ b/_data/compose-cli/docker_compose.yaml
@@ -98,6 +98,9 @@ long: |-
   and so does `COMPOSE_PROFILES` environment variable for to the `--profiles` flag.
 
   If flags are explicitly set on command line, associated environment variable is ignored
+
+  Setting the `COMPOSE_IGNORE_ORPHANS` environment variable to `true` will stop docker compose from detecting orphaned
+  containers for the project.
 usage: docker compose
 pname: docker
 plink: docker.yaml
@@ -126,6 +129,7 @@ cname:
 - docker compose top
 - docker compose unpause
 - docker compose up
+- docker compose version
 clink:
 - docker_compose_build.yaml
 - docker_compose_convert.yaml
@@ -151,6 +155,7 @@ clink:
 - docker_compose_top.yaml
 - docker_compose_unpause.yaml
 - docker_compose_up.yaml
+- docker_compose_version.yaml
 options:
 - option: ansi
   value_type: string
@@ -158,6 +163,17 @@ options:
   description: |
     Control when to print ANSI control characters ("never"|"always"|"auto")
   deprecated: false
+  hidden: false
+  experimental: false
+  experimentalcli: false
+  kubernetes: false
+  swarm: false
+- option: compatibility
+  value_type: bool
+  default_value: "false"
+  description: Run compose in backward compatibility mode
+  deprecated: false
+  hidden: false
   experimental: false
   experimentalcli: false
   kubernetes: false
@@ -166,6 +182,7 @@ options:
   value_type: string
   description: Specify an alternate environment file.
   deprecated: false
+  hidden: false
   experimental: false
   experimentalcli: false
   kubernetes: false
@@ -176,6 +193,7 @@ options:
   default_value: '[]'
   description: Compose configuration files
   deprecated: false
+  hidden: false
   experimental: false
   experimentalcli: false
   kubernetes: false
@@ -185,6 +203,7 @@ options:
   default_value: "false"
   description: Do not print ANSI control characters (DEPRECATED)
   deprecated: false
+  hidden: true
   experimental: false
   experimentalcli: false
   kubernetes: false
@@ -194,6 +213,7 @@ options:
   default_value: '[]'
   description: Specify a profile to enable
   deprecated: false
+  hidden: false
   experimental: false
   experimentalcli: false
   kubernetes: false
@@ -204,6 +224,7 @@ options:
     Specify an alternate working directory
     (default: the path of the Compose file)
   deprecated: false
+  hidden: false
   experimental: false
   experimentalcli: false
   kubernetes: false
@@ -213,6 +234,7 @@ options:
   value_type: string
   description: Project name
   deprecated: false
+  hidden: false
   experimental: false
   experimentalcli: false
   kubernetes: false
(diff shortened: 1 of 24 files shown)


_data/compose-cli/docker_compose.yaml
```diff
@@ -233,6 +267,7 @@ options:
     Specify an alternate working directory
     (default: the path of the Compose file)
   deprecated: false
```
thaJeztah, line 269: We should update the code to mark this deprecated, that way it's also annotated as such, and shows a `deprecated` badge, like this;

Screenshot 2022-05-02 at 11 27 27


To mark it deprecated, there's two options;

actually deprecate it; this makes the flag hidden on the command-line and produces an error ("deprecated: deprecation message") when someone tries to use it; https://github.com/docker/cli/blob/6c9eb708fa6d17765d71965f90e1c59cea686ee9/cli/command/context/update.go#L54-L55


"soft" deprecate; adding a `deprecated` annotation; the flag remains to be functional, but the annotation is used in our docs to show that it's deprecated. Soft deprecation can be useful in situations where a flag should no longer be used, but where _fully_ removing it would cause too much of a backward-breaking change. In that case, you can also mark the flag as `hidden` in Cobra so that it doesn't show on the command line `--help` output to discourage new users from using it; example; https://github.com/docker/cli/blob/v20.10.14/cli/command/context/create.go#L69

_data/compose-cli/docker_compose_pull.yaml
```diff
@@ -49,10 +53,52 @@ options:
   default_value: "false"
   description: Pull without printing progress information
   deprecated: false
+  hidden: false
   experimental: false
   experimentalcli: false
   kubernetes: false
   swarm: false
+examples: |-
+  suppose you have this `compose.yaml`:
```
thaJeztah, line 62: We should look at;

- Adding a heading for each example
- Perhaps split the example into `docker compose pull` (without service specified) and one with a service (and adjust the compose file to have multiple services that use an image).

_data/compose-cli/docker_compose_run.yaml
```diff
@@ -83,16 +85,29 @@ options:
...
   kubernetes: false
   swarm: false
-- option: labels
+- option: interactive
+  shorthand: i
+  value_type: bool
+  default_value: "true"
+  description: Keep STDIN open even if not attached.
+  deprecated: false
+  hidden: false
```
thaJeztah, line 99: Looks like for `docker compose exec` we marked this flag `hidden`, but for `docker compose run` we made it visible (perhaps something to look into); or is this dynamically hidden/shown based on the auto-detected?

_data/compose-cli/docker_compose_exec.yaml
```diff
@@ -35,17 +37,30 @@ options:
...
   experimentalcli: false
   kubernetes: false
   swarm: false
 - option: no-TTY
   shorthand: T
   value_type: bool
-  default_value: "false"
+  default_value: "true"
   description: |
     Disable pseudo-TTY allocation. By default `docker compose exec` allocates a TTY.
```
thaJeztah, line 61: Is the TTY now auto-detected on `compose exec` as well? (like on `compose run`?) or is it hard-coded to "enabled"?

_data/compose-cli/docker_compose_restart.yaml
```diff
@@ -1,6 +1,16 @@
 command: docker compose restart
 short: Restart containers
-long: Restart containers
+long: |-
+  Restarts all stopped and running services.
```
thaJeztah, line 4: TIL it starts a stopped service as well 🤔 

(I assume it's not starting one-off containers started by `compose run`?)

_data/compose-cli/docker_compose_run.yaml
```diff
@@ -131,6 +148,17 @@ options:
   default_value: '[]'
   description: Publish a container's port(s) to the host.
   deprecated: false
+  hidden: false
+  experimental: false
+  experimentalcli: false
+  kubernetes: false
+  swarm: false
+- option: quiet-pull
```
thaJeztah, line 156: FWIW; on `docker run`, this option is just named `--quiet` (but the flag's description describes what "quiet" means for `docker run`); https://github.com/docker/cli/blob/7b4d7717b1d14dbaf251c343ba47b242d9e72eb6/cli/command/container/run.go#L61

_data/compose-cli/docker_compose_run.yaml
```diff
@@ -150,6 +179,18 @@ options:
...
+  experimentalcli: false
+  kubernetes: false
+  swarm: false
+- option: tty
+  shorthand: t
+  value_type: bool
+  default_value: "true"
+  description: Allocate a pseudo-TTY.
+  deprecated: false
+  hidden: true
```
thaJeztah, line 193: Also see my other comment w.r.t. `-i`; looks like we're not consistent. I _think_ `-i` is also enabled/disabled through the `-T` / `--no-TTY` option? Or does that only toggle `-t`?

_data/compose-cli/docker_compose_up.yaml
```diff
@@ -73,16 +78,7 @@ options:
   default_value: "false"
   description: 'Detached mode: Run containers in the background'
   deprecated: false
-  experimental: false
-  experimentalcli: false
-  kubernetes: false
-  swarm: false
-- option: environment
-  shorthand: e
```
thaJeztah, line 81: The `-e` / `--environment` option was deprecated/removed? I couldn't find that mentioned in the release notes on https://github.com/docker/compose/releases

thaJeztah (approved):
LGTM, but left some comments/questions for things to address in the docker/compose repository.

_data/compose-cli/docker_compose_exec.yaml
glours: It's autodetected 
https://github.com/docker/compose/blob/4cebef1bf11c9ded9e2bdd405071f9340ed79e59/cmd/compose/exec.go#L72

_data/compose-cli/docker_compose_run.yaml
glours: I will open a PR to rename to `quiet` and add an alias for `quiet-pull` to keep backward compatibility

_data/compose-cli/docker_compose_run.yaml
glours: Right! we should apply the same configuration for this tag  as done in the `exec` command

usha-mandya:
Discussed with @glours. He can do a PR to update this in the upstream repo.

_data/compose-cli/docker_compose_run.yaml
glours: I just checked and Compose isn't working the same way as the cli, we only have one flag which manage the pull option without displaying the output. 
We don't have the `pull` + another flag to manage the output just the `quiet-pull` one

PR #14641, opened 2022-05-01
updates to command-line reference layout

thaJeztah:
### reference: remove "enterprise_only" condition

This metadata condition was used for features that were only available
on Docker Enterprise Edition ("Docker EE"), and is no longer used.

### reference: remove "description" header, and use it for "extended description"

The "description" header included the short description of each command. In most
cases, this description was very short ("run a container"), and adding the extra
header before it only was adding extra noise.

This patch:

- removes the top "description" header
- renames the existing "extended description" header to "description" (a hidden
  "extended-description" anchor is added for backward compatibility)

As the extended description can be long (hopefully!), there may be a long distance
between the `Usage` section and the `Options` section. To help users navigate
to the list of available options, an extra line is printed if an extended
description is available for the command, including a link to the corresponding
section:

> Refer to the options section for an overview of available OPTIONS for this command.

### Before

before

### After

after

### reference: omit "parent command" section for "docker"

The "parent command" can be useful if the page describes a subcommand,
but generally having a link to `docker` may not provide much value.

### reference: hide "minimum API version" badges for old versions

The reference pages show badges for commands and options (flags) that require
a minimum API version. While this information can be useful if an option was
added in a recent version of the Docker Engine (and API), these badges are no
longer relevant to most users if the minimum required version is quite "old".

We assume users reading these pages to be on the current version, or at most
on the version before that (which is already "unsupported"). Users running
older versions have bigger problems on their hand, so we're not accounting for
those.

So, to reduce unnecessary clutter on the page, we only show the minimum required
API version if it requires a relatively recent version of the Engine.

A new "min_api_threshold" option was added in the `_config.yml`, which specifies
the minimum required API version for which we show a badge (currently: API v1.40,
or "Docker 19.03").

### Before

For commands:

command-api-ver-before

For options:

option-api-ver-before2

### After

For commands:

command-api-ver-after

For options:

option-api-ver-after2

### reference: update description for experimental cli features

Experimental CLI features are enabled by default since docker 20.10, so the
instructions for enabling the features were no longer needed. Also tweaked the
description a bit to align how we describe experimental features elsewhere.

### Before

experimental-before

### After

experimental-after


thaJeztah:
@usha-mandya ptal

usha-mandya (approved):
LGTM

PR #14642, opened 2022-05-02
Create dedicated Docker Build section

crazy-max:
As discussed with @usha-mandya and @thaJeztah, create a dedicated "Docker Build" section in "Manuals". This is a first step to put documentation related to build together.

I moved the install part in a dedicated page, I think it makes sense (specially for the /go redirect to install buildx).

In some follow-ups I will take a look at the "Guides" and move some build bits to "Manuals" or at least mutualize some of them.

I will also check how we can fetch buildx guides and add them to our docs: https://github.com/docker/buildx/tree/master/docs/guides as well as the Dockerfile syntax from BuildKit repo: https://github.com/moby/buildkit/blob/master/frontend/dockerfile/docs/syntax.md


crazy-max:
image

Ah looks like it needs to be a member of this repo. Maybe it will be too noisy. Will just add me in the meantime.

.github/CODEOWNERS
```diff
@@ -3,6 +3,8 @@
 
 # For more details, see https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners
 
+/build @dockertopia @crazy-max
```
dockertopia, line 6: I was waiting for the /build doc set to appear first to change the codeowners file, but thanks for this @crazy-max .

build/buildx/index.md
```diff
@@ -1,7 +1,9 @@
 ---
-title: Docker Buildx
+title: Buildx component
 description: Working with Docker Buildx
-keywords: Docker, buildx, multi-arch
```
dockertopia, line 4: How does this metadata work? Does this help search or help create related topics? Just wondering how "multi-arch" would work depending on how this metadata is picked-up.

_data/toc.yaml
```diff
@@ -1391,15 +1391,22 @@ manuals:
...
   - path: /engine/context/working-with-contexts/
     title: Docker Context
   - path: /engine/scan/
     title: Docker Scan
   - path: /engine/sbom/
     title: Docker SBOM (Experimental)
 
+- sectiontitle: Docker Build
+  section:
+  - sectiontitle: Buildx component
```
dockertopia, line 1403: I don't remember if we discussed it before, but I'm not sure why we need this second level at least now before we have more content.

dockertopia (commented):
As a general comment I would like to take a few of days to review these new pages and discuss offline the next stages re guides and expansion of this doc set. 
But if you'd rather have a more simple review now to just get this new section into master and then improve on it with additional PRs, PLMK.

build/buildx/index.md
crazy-max: Keywords are like tags that will be used to enhance search results and are added at the bottom of the page:

image

image

They are also added as meta tag for SEO https://github.com/docker/docker.github.io/blob/9294453e5fe816824ba855e92eae3fc5b16c9c11/_includes/head.html#L39 but I don't think this is relevant anymore as it's not used by search engines anymore [like Google](https://webmasters.googleblog.com/2009/09/google-does-not-use-keywords-meta-tag.html):

> […] keyword meta tags quickly became an area where someone could stuff often-irrelevant keywords without typical visitors ever seeing those keywords. Because the keywords meta tag was so often abused, many years ago Google began disregarding the keywords meta tag.

Bing went one step further in 2011 when they [announced](https://searchengineland.com/the-meta-keywords-tag-lives-at-bing-why-only-spammers-should-use-it-96874) they use the tag as a spam signal.

I think we should remove the meta keywords in the future but would need to put in place a client-side search implementation, omitting the need for the integration of third-party services (Google in our case), which might not be compliant with privacy regulations. Moreover, search even works offline, allowing users to download our documentation. I think [lunr](https://lunrjs.com/) could be a good candidate for this. See for example my POC with MkDocs using a client-side search: https://crazymax.dev/docker-docs/

cc @usha-mandya @thaJeztah

_data/toc.yaml
crazy-max: Yes we might not need a new section for now. I also added bake guides in a sub section in this PR as it's a quite extensive guide: https://github.com/docker/buildx/pull/1140. Let me know what you think about it.

crazy-max:
> As a general comment I would like to take a few of days to review these new pages and discuss offline the next stages re guides and expansion of this doc set. But if you'd rather have a more simple review now to just get this new section into master and then improve on it with additional PRs, PLMK.

I will add a simple Overview page for the build section first in "Manuals > Docker Build" and after that I think we can merge and iterate over it. I keep you in touch.

build/buildx/index.md
```diff
@@ -1,7 +1,9 @@
 ---
-title: Docker Buildx
+title: Buildx component
```
dockertopia, line 2: Buildx is in other places referred to as the Docker CLI Buildx plugin, I think we should be consistent.

---------------------------------------
As for structure of the Build section (and page titles) here's a working proposal
----------------------------------------
It seems this will be a broad section, WDYT about (and since in Engine there is already a topic called Docker Buildx) structuring the content like:
Docker Build
1. (Build) Overview
2 About Buildx
 2.1. Installing Buildx
 2.2. About Buildx
 2.3. Using Buildx
 2.4. Using Bake

build/buildx/install.md
```diff
@@ -0,0 +1,76 @@
+---
+title: Install Docker Buildx
+description: How to install Docker Buildx
+keywords: build, buildx, buildkit
+---
+
+## Docker Desktop
+
+Docker Buildx is included in [Docker Desktop](../../desktop/index.md) for Windows, macOS, and Linux.
```
dockertopia, line 9:
```suggestion
Docker Buildx is a Docker CLI plugin and is shipped with [Docker Desktop](../../desktop/index.md). Docker Desktop is available for Windows, macOS, and Linux.
```

build/buildx/index.md
dockertopia: It would be great to be able to optimize the search indexes ourselves. It could take the docs experience to another level, when you'd search a term you'd already get the best matches from the team's perspective what are the best overall content candidates across types. 
This means we could, if somebody searched "docker build" make sure the number 1 hit on the list is the Docker Build docs index as well as curate results across categories (refs, guides, blog, etc.).

dockertopia (commented):
I have agreed with Kevin to review the section in general (and install and index files for build and buildx in particular) in a new build-docs branch.

PR #14645, opened 2022-05-02
Allow providing buildkit secrets with env

gitryder:
Wrote about functionality added in 20.10.0, that allows providing build secrets with env

### Proposed changes

- Starting with Docker v20.10.0, buildkit secrets allows providing build secrets with `env`, however this wasn't mentioned in the documentation and I only discovered it from [this](https://pythonspeed.com/articles/docker-build-secrets/) blog post.
- Ergo, this PR.

### DIFF
--- a/develop/develop-images/build_enhancements.md
+++ b/develop/develop-images/build_enhancements.md
@@ -188,6 +188,18 @@ $ docker build --no-cache --progress=plain --secret id=mysecret,src=mysecret.txt
 ...
 ```
 
+Starting from Docker 20.10, you can now provide secrets with `env` as an alternative for storing a secret value to a file. For example, if you have an environment variable MY_ENV, you can build the image with:
+```console
+docker build --secret id=foo,env=MY_ENV
+```
+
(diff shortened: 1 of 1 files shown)


thaJeztah:
@docker/build ptal

tonistiigi (approved):

usha-mandya:
@dvdksn @crazy-max The PR has conflicts as we've moved Build docs. Could you check if this is included in the latest docs?

crazy-max:
@usha-mandya I think we are good. We have secret usage example in `buildx build` ref docs: https://docs.docker.com/engine/reference/commandline/buildx_build/#secret. We also have one in Dockerfile ref but without `env` example: https://docs.docker.com/engine/reference/builder/#run---mounttypesecret

dvdksn:
Agree—the description in the `buildx build` reference describes this well. @gitryder if you feel that the existing documentation is not good enough, feel free to open an issue describing what improvements you need, and we can take it from there.

PR #14653, opened 2022-05-03
Fixed the description for handling the GitHub Container Registry

T145:
### Proposed changes

The documentation did not mention to also change where an image should be cached when pointing to the GitHub Container Registry instead of Docker Hub.

### DIFF
--- a/ci-cd/github-actions.md
+++ b/ci-cd/github-actions.md
@@ -247,18 +247,18 @@ Now let's change the Docker Hub login with the GitHub Container Registry one:
 ```
 {% endraw %}
 
-Remember to change how the image is tagged. The following example keeps ‘latest'
-as the only tag. However, you can add any logic to this if you prefer:
+Remember to change how the image is tagged and where it's cached. The following example is a
+functional equivalent for the aforementioned logic against the GitHub Container Registry:
 
 {% raw %}
 ```yaml
-  tags: ghcr.io/<username>/simplewhale:latest
+  tags: ghcr.io/${{ secrets.GITHUB_USERNAME }}/simplewhale:latest
+  cache-from: type=registry,ref=ghcr.io/${{ secrets.GITHUB_USERNAME }}/simplewhale:buildcache
+  cache-to: type=registry,ref=ghcr.io/${{ secrets.GITHUB_USERNAME }}/simplewhale:buildcache,mode=max
 ```
 {% endraw %}
 
-> **Note**: Replace `<username>` with the repository owner. We could use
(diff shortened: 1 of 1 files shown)


usha-mandya:
@crazy-max Could you PTAL?

ci-cd/github-actions.md
```diff
@@ -247,18 +247,18 @@ Now let's change the Docker Hub login with the GitHub Container Registry one:
...
-as the only tag. However, you can add any logic to this if you prefer:
+Remember to change how the image is tagged and where it's cached. The following example is a
+functional equivalent for the aforementioned logic against the GitHub Container Registry:
 
 {% raw %}
 ```yaml
-  tags: ghcr.io/<username>/simplewhale:latest
+  tags: ghcr.io/${{ secrets.GITHUB_USERNAME }}/simplewhale:latest
+  cache-from: type=registry,ref=ghcr.io/${{ secrets.GITHUB_USERNAME }}/simplewhale:buildcache
+  cache-to: type=registry,ref=ghcr.io/${{ secrets.GITHUB_USERNAME }}/simplewhale:buildcache,mode=max
```
crazy-max, line 257: Don't think this is the right place to introduce cache behavior with GitHub Actions in a getting started guide. Also we should use `gha` cache instead.

ci-cd/github-actions.md
```diff
@@ -247,18 +247,18 @@ Now let's change the Docker Hub login with the GitHub Container Registry one:
...
 ```
 {% endraw %}
 
-> **Note**: Replace `<username>` with the repository owner. We could use
-> {% raw %}`${{ github.repository_owner }}`{% endraw %} but this value can be mixed-case, so it could
-> fail as [repository name must be lowercase](https://github.com/docker/build-push-action/blob/master/TROUBLESHOOTING.md#repository-name-must-be-lowercase){:target="_blank" rel="noopener" class="_"}.
+> **Note**: The repository owner's username needs to be lowercase.
```
crazy-max, line 261: The current note looks better to me and also the link to upstream page is important to explain how to handle letter case for the username otherwise user might be confused.

crazy-max (commented):
@usha-mandya We are currently reviewing CI/CD documentation with @dvdksn so I suggest to wait before making any modifications on this page.

ci-cd/github-actions.md
T145: If you read the earlier documentation in this document, the caching behavior is already introduced. No instructions were provided when switching to the Github Container Registry.

ci-cd/github-actions.md
T145: I agree that including the link would be good since it explains the problem, and have updated the page to include it.

PR #14664, opened 2022-05-04
fix link to sdk docs

aevesdocker:


desktop/extensions.md
```diff
@@ -8,7 +8,7 @@ toc_max: 2
 
 Docker Extensions lets you use third-party tools within Docker Desktop to extend its functionality. There is no limit to the number of extensions you can install.
 
-Docker Community members and partners can use our [SDK](../extensions-sdk.md) to create new extensions. 
+Docker Community members and partners can use our [SDK](https://docs.docker.com/desktop/extensions-sdk/) to create new extensions.
```
usha-mandya, line 11: @thaJeztah wondering if it's possible to use relative links to content we fetch using the fetch upstream scripts?

thaJeztah: I think the link here should be;

```suggestion
Docker Community members and partners can use our [SDK](extensions-sdk.md) to create new extensions. 
```

If it's also within the `desktop` directory 🤔

aevesdocker: @thaJeztah just tried that locally at it didn't like it either :( still got a broken link

usha-mandya: Tried building locally with `extensions-sdk.md` but that fails as the file doesn't exist within the `desktop` directory. Here's the link to the Extensions entry `https://github.com/docker/docker.github.io/blob/master/_scripts/fetch-upstream-resources.sh#L34`

aevesdocker: It only seems to work if I use the full https one

aevesdocker: sounds good to me @usha-mandya

thaJeztah: Thanks for the link to the "fetch resources"; so I think the file is stored in the  `/desktop/extensions-sdk/` directory, which means the file is probably the `index.md` within that directory. Can you try if this works?

```suggestion
Docker Community members and partners can use our [SDK](../extensions-sdk/index.md) to create new extensions. 
```

thaJeztah: A, sorry, without `../` (edited my comment)

aevesdocker: @thaJeztah that worked! Thank you! Will change it now

usha-mandya (approved):
LGTM. Thank you @aevesdocker @thaJeztah

PR #14644, opened 2022-05-02
Add build guides from upstream

crazy-max:
requires #14642 and https://github.com/docker/buildx/pull/1087

Fetch build guides from upstream buildx repo. Use my fork in the meantime for testing.

PR #14685, opened 2022-05-06
Enhanced Dockerfile and update GHA workflows

crazy-max:
This is a follow-up of https://github.com/docker/docker.github.io/pull/14644#issuecomment-1114746180 to start working with the native netlify client to leverage the possibility to fetch remote resources.

This PR is just a prep to handle that because the current Dockerfile use a quite old jekyll image `starefossen/github-pages:198` (~3 years old) that doesn't support some Gem dependencies.

Instead of relying on a pre-built Jekyll image I switched to a vanilla ruby image.

I also vendor and push the Gemfile.lock in the repo as it should so we can pin to a specific Jekyll version as well as the plugins being used in our configuration. Atm it's Jekyll 4.2.2. I have created a dedicated stage to vendor our Gemfile in case we want to update and vendor the lock file ourselves:

```shell
$ make vendor
# or
$ docker buildx bake vendor
```

This is important so we are sync with our lock file while installing dependencies and deploying. Otherwise we could have unexpected behaviors if Jekyll or its plugins are updated. Same spirit as the `go.sum`.

I have also updated the GHA workflows accordingly. Also merge our `build-*.yml` altogether as there are some dedup logic that can be mutualized. Let me know what you think about this one.


docker-bake.hcl
```diff
@@ -0,0 +1,21 @@
...
+group "default" {
+  targets = ["release"]
+}
+
+target "release" {
+  target = "release"
+  args = {
+    JEKYLL_ENV = JEKYLL_ENV
+  }
+  no-cache-filter = ["upstream-resources"]
```
crazy-max, line 14: Thanks to the `no-cache-filter` there is no need to use `docker build --no-cache`. Just this specific stage will be invalidated.

crazy-max:
image

this required check could be removed if we merge this PR as the workflow is now mutualized to avoid dedup.

Dockerfile
```diff
@@ -1,108 +1,133 @@
...
+EOT
+
+FROM gem AS htmlproofer
+RUN --mount=type=bind,from=generate,source=/out,target=_site \
+  --mount=type=cache,target=/usr/local/bundle \
+  htmlproofer ./_site \
+    --disable-external \
+    --internal-domains="docs.docker.com,docs-stage.docker.com,localhost:4000" \
+    --file-ignore="/^./_site/engine/api/.*$/,./_site/registry/configuration/index.html" \
+    --url-ignore="/^/docker-hub/api/latest/.*$/,/^/engine/api/v.+/#.*$/,/^/glossary/.*$/"
```
crazy-max, line 111: I moved the htmlproofer check in our Dockerfile so we can also check locally.

crazy-max:
Since this PR and switch to Jekyll 4.2.2, there is a conflict being reported: https://app.netlify.com/sites/docsdocker/deploys/62783880dcb06f000a4f7d25#L243

image

Looks like we push an `index.html` file which collides with a markdown file: https://github.com/docker/docker.github.io/tree/master/docker-hub/api. I guess we should remove https://github.com/docker/docker.github.io/tree/master/docker-hub/api/latest

thaJeztah:
> Looks like we push an `index.html` file which collides with a markdown file: https://github.com/docker/docker.github.io/tree/master/docker-hub/api. I guess we should remove https://github.com/docker/docker.github.io/tree/master/docker-hub/api/latest

Yes, that looks a bit suspicious. https://github.com/docker/docker.github.io/blob/master/docker-hub/api/latest/index.html looks to have a redirect in it (but the redirect is to the page itself? Given that the `latest.md` includes a template/layout to use, I think the `latest.md` is the correct one.

Dockerfile
```diff
@@ -1,108 +1,115 @@
...
+# that is used by Netlify
+ARG RUBY_VERSION=2.6.10
+# Same as the one in Gemfile.lock
+ARG BUNDLER_VERSION=2.3.13
+
 ARG JEKYLL_ENV=development
+ARG DOMAIN=docs.docker.com
 
-# Engine
 ARG ENGINE_BRANCH="20.10"
```
thaJeztah, line 15: Should we use the new format to describe arguments (to allow the `docker buildx outline` feature to describe things?)

_config.yml
```diff
@@ -16,7 +16,19 @@ incremental: true
...
+exclude:
+  - _samples
+  - _scripts
+  - 404.html
+  - datacenter
+  - docker-bake.hcl
+  - ee
+  - index.html
+  - js/metadata.json
+  - LICENSE
```
thaJeztah, line 29: Does this exclude LICENSE from the final image? (wondering if we need to keep it)

Dockerfile
crazy-max: Yes we could but these args will be removed in this follow-up if we are ok with it. See https://github.com/docker/docker.github.io/pull/14688/commits/c20f8d42db17428e36a01ab1f198aeff0047c759

.github/workflows/build.yml
```diff
@@ -83,20 +83,11 @@ jobs:
...
-          targets: release
       -
         name: Check for broken links
-        uses: chabad360/htmlproofer@master
+        uses: docker/bake-action@v2
         with:
-          directory: ./_site
-          # for available options, refer to:
-          # - https://github.com/gjtorikian/html-proofer
-          # - https://github.com/gjtorikian/html-proofer/blob/main/bin/htmlproofer
```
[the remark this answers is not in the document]
crazy-max: Yes good point

_config.yml
crazy-max: Yes was not sure if it's relevant to have the LICENSE and README.md on the website actually:
* https://docs.docker.com/LICENSE
* https://docs.docker.com/README.md

Gemfile
```diff
@@ -1,43 +1,8 @@
-source "https://rubygems.org"
-
-# Update me once in a while: https://github.com/github/pages-gem/releases
-# Please ensure, before upgrading, that this version exists as a tag in starefossen/github-pages here:
-# https://hub.docker.com/r/starefossen/github-pages/tags/
-#
```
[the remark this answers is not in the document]
crazy-max: Not relevant anymore since we use a vanilla Ruby image.

Dockerfile
crazy-max: As discussed, will move the [JSON config for the plugin](https://github.com/docker/docker.github.io/blob/bd1eac0b7b0e413e4c260ea8f97c0e284a272080/_data/fetch-remote.json) in #14688 to the `_config.yml`

crazy-max:
You might be interested by this one @StefanScherer if you want to take a look

Gemfile.lock
```diff
@@ -0,0 +1,77 @@
...
+    rouge (3.28.0)
+    safe_yaml (1.0.5)
+    sassc (2.4.0)
+      ffi (~> 1.9)
+    terminal-table (2.0.0)
+      unicode-display_width (~> 1.1, >= 1.1.1)
+    unicode-display_width (1.8.0)
+
+PLATFORMS
+  x86_64-linux
```
thaJeztah, line 85: Will this affect things when building on an M1 ? (or do we not have an arm64 image either way?)

thaJeztah (approved):
LGTM, thanks!

Gemfile.lock
crazy-max: Afaik it will not break anything. It just says that this lock has been generated on this platform. We can add more of them which should be supported using `bundle lock --add-platform ...`. As I can see the official ruby image support a bunch of platforms for the `2.6.10` tag we use:

image

So I would say we could add them.

crazy-max: Ok tested on M1 and it just adds `aarch64-linux` to the list of platforms in the lock file. I've updated the lock file with it.

crazy-max: Also tried building on raspberry which seems to work fine. So I also added `arm-linux`.

thaJeztah: Oh! Living on the edge; building it on a rPI! 😂

thaJeztah (approved):
LGTM!

thaJeztah:
Thanks!

PR #14695, opened 2022-05-09
ddl: improve "overview" page

thaJeztah:
- add missing title
- add some keywords that may help with search
- wrap long lines to ~80 chars
- add a short introduction paragraph
- add link to virtiofs website
- add some links to DD for Windows and Mac.
- add some additional information on how DD4L and Engine differ (e.g., add  mention of the Dashboard and Preferences).
- extend description of "contexts", what they're for, and how to use them (within "context" of Docker Desktop for Linux).

Also made some tweaks to the "install" page;

- remove "update to terms" banner, as DD4L was not available before the change of terms
- slight tweaking of keywords

### Related issues (optional)

- follow-up to https://github.com/docker/docker.github.io/pull/14648

PR #14702, opened 2022-05-09
PR to fix further style guide issues on the new content that was added for DD4L.  

aevesdocker:
It also incorporates changes fed back in this ticket 
https://docker.atlassian.net/browse/POS-901

And some of the proposed changes in this PR:
https://github.com/docker/docker.github.io/pull/14695

Note: Any content that was copied and pasted from Mac/Windows docs was not touched/edited so as to have consistency across all three platforms until all the Docker Desktop docs can be reorganised and simplified.


desktop/linux/networking.md
```diff
@@ -46,12 +46,15 @@ See [Proxies](index.md#proxies).
 Docker Desktop for Linux allows you to use the host’s SSH agent inside a container. To do this:
 
 1. Bind mount the SSH agent socket by adding the following parameter to your `docker run` command:
-
-    `--mount type=bind,src=/run/host-services/ssh-auth.sock,target=/run/host-services/ssh-auth.sock`
+```console
+$ --mount type=bind,src=/run/host-services/ssh-auth.sock,target=/run/host-services/ssh-auth.sock
+```
```
thaJeztah, line 51: This example is a bit confusing, as it's a "partial" instruction (it's missing the `docker run`); I'm guessing this is why it wasn't put in a `console` block.

The full command would be something like

```console
$ docker run --mount type=bind,src=/run/host-services/ssh-auth.sock,target=/run/host-services/ssh-auth.sock <SOME IMAGE>
```

But perhaps that's too long (and the other options to use depend on what image you're running, etc).

Also looks like the code block needs to be indented to prevent the numbered list from being reset (next item is `1.` again)

desktop/linux/networking.md
```diff
@@ -46,12 +46,15 @@ See [Proxies](index.md#proxies).
...
 
 1. Add the `SSH_AUTH_SOCK` environment variable in your container:
 
-    `-e SSH_AUTH_SOCK="/run/host-services/ssh-auth.sock"`
+```console
+$ -e SSH_AUTH_SOCK="/run/host-services/ssh-auth.sock"
+```
```
thaJeztah, line 57: Same for this one

desktop/linux/install.md
```diff
@@ -138,7 +127,84 @@ Log out and log back in so that your group membership is re-evaluated.
...
+interact with Docker Desktop. On startup, Docker Desktop automatically sets its
+own context (`desktop-linux`) as the current context. This means that subsequent
+Docker CLI commands target Docker Desktop. On shutdown, Docker Desktop resets
+the current context to the `default` context.
+
+Use the `docker context ls` command to view what contexts are available on your
+machine. The current context is indicated with an asterisk (`*`);
+
+```console
+ $ docker context ls
```
thaJeztah, line 179: This one has a leading whitespace;

```suggestion
$ docker context ls
```

desktop/linux/install.md
```diff
@@ -138,7 +127,84 @@ Log out and log back in so that your group membership is re-evaluated.
...
+
+And use the `docker-desktop` context to interact with Docker Desktop:
+
+```console
+$ docker context use docker-desktop
+docker-desktop
+Current context is now "docker-desktop"
+```
+
+Refer to the [Docker Context documentation](/engine/context/working-with-contexts) for more details.
```
thaJeztah, line 204:
```suggestion
Refer to the [Docker Context documentation](../../engine/context/working-with-contexts.md) for more details.
```

desktop/linux/install/fedora.md
```diff
@@ -1,33 +1,25 @@
...
-[meet the prerequisites](#prerequisites), then
-[install Docker Desktop](#install-docker-desktop).
+This page contains information on how to install, launch and upgrade Docker Desktop on a Fedora distribution.
 
 ## Prerequisites
 
-Your Fedora distribution must meet the [system requirements](../install.md#system-requirements) to install and launch Docker Desktop successfully.
+To install Docker Desktop successfully, you must:
+- Meet the [system requirements](../install.md#system-requirements).
+- Have a 64-bit version of either Fedora 35 or Fedora 36.
```
thaJeztah, line 14: (I don't have a good solution, but let me at least leave a comment 😅)

In my experience, many users "don't read", and more than once, I encountered users not having read the supported versions of a distro. Perhaps the previous approach was "too verbose", but wondering how we can make the supported versions stand out more (for those users that only give it a "casual" look and overlook this). 🤔

desktop/linux/install/fedora.md
```diff
@@ -1,33 +1,25 @@
...
-- Fedora 36
+For a Gnome Desktop environment you must also install AppIndicator and KStatusNotifierItem [Gnome extensions](https://extensions.gnome.org/extension/615/appindicator-support/).
 
 ## Install Docker Desktop
 
-Recommended approach to install Docker Desktop on Ubuntu:
+To install Docker Desktop on Fedora:
 
-1. [Set up Docker's RPM repository](../../../engine/install/fedora.md#set-up-the-repository) 
+1. Set up [Docker's RPM repository](../../../engine/install/fedora.md#set-up-the-repository).
```
thaJeztah, line 22: Wondering if we should use "package repository" for these (instead of `RPM` / `DEB`) (not sure if those should be all-caps  🤔and using "package repository" avoids that 😂

thaJeztah (commented):
Thanks! Left some comments inline.


As to the "order of items" in the menu, I'm wondering if the "introduction" page should be used as the first item in all of the Desktop variants. Curently, the "index" page (https://deploy-preview-14702--docsdocker.netlify.app/desktop/linux/) highlights the third item in the list. While I understand the "expected" flow to be to _first_ install Desktop, another angle could be to look at it as

- "here's the overview of Desktop"
- if you haven't done so yet, head to the page below to install

For most users, the "installation" page would be something they'd visit only once (first time they install), after that, the index/overview/guide may be more relevant, and they'd only have to refer to the "install" page in the (less common) scenario that they need to re-install.

desktop/linux/networking.md
aevesdocker: Had to find a work around so the code block was still scrollable. Currently in the docs the code block crosses over the right-hand toc. I'm sure this is not the proper way to do it 😅

christophermclellan (approved):
Great, thanks Allie !

thaJeztah:
I see there's a merge conflict due to https://github.com/docker/docker.github.io/pull/14704 being merged (let me know if you need help on resolving)

aevesdocker:
Thank you for your feedback @thaJeztah. 

I completely agree with you about the order of items. This is something I'm hoping to revisit when I review the whole structure of Docker Desktop's docs. For now though, I think it's best to leave it as it is because this is how it currently is for Mac/Windows and perhaps is beyond the scope of this PR. 

Re: users not reading the prerequisites/ supported versions of the distros - you've described a universal problem 😂. I'm not 100% on how to make it extra clear. I felt my proposed change was an improvement to the original. Perhaps, again, this can be revisited further down the line - conscious we want to get this next iteration of improvements out before Dockercon! 

cc @usha-mandya

desktop/linux/install.md
```diff
@@ -138,7 +127,84 @@ Log out and log back in so that your group membership is re-evaluated.
...
+
+And use the `docker-desktop` context to interact with Docker Desktop:
+
+```console
+$ docker context use docker-desktop
+docker-desktop
+Current context is now "docker-desktop"
+```
+
+Refer to the [Docker Context documentation](../../engine/context/working-with-contexts) for more details.
```
thaJeztah, line 204: Link-checker was failing; at a glance, I think it's missing the `.md` extension here;

```suggestion
Refer to the [Docker Context documentation](../../engine/context/working-with-contexts.md) for more details.
```

desktop/linux/install.md
```diff
@@ -24,25 +14,25 @@ Welcome to Docker Desktop for Linux. This page contains information about Docker
...
-- systemd init system
+- systemd init system.
 
-- Gnome or KDE Desktop environment
+- Gnome or KDE Desktop environment.
 
 - At least 4 GB of RAM.
 
-Docker Desktop for Linux runs a Virtual Machine (VM). For more information on why, see [here](overview.md#why-docker-desktop-for-linux-runs-a-vm).
+Docker Desktop for Linux runs a Virtual Machine (VM). For more information on why, see [here](install.md#why-docker-desktop-for-linux-runs-a-vm).
```
usha-mandya, line 31:
```suggestion
Docker Desktop for Linux runs a Virtual Machine (VM). For more information, see [Why Docker Desktop for Linux runs a VM](install.md#why-docker-desktop-for-linux-runs-a-vm).
```

PR #14704, opened 2022-05-09
DDL: add `gnome-terminal` as prerequisite

aiordache:
The CLI button in the container menu from the Docker Dashboard is running a `gnome-terminal` cmd. If this is not installed, the Docker UI crashes.

### Proposed changes
For now, document the installation of gnome-terminal as a prerequisite. We'll remove it once we add it as a package dependency or update the UI to look for the default terminal of the host distro.


aevesdocker (approved):
LGTM

PR #14714, opened 2022-05-11
Update rails.md - removing Gemfile.lock

philkcw:
### Proposed changes

The Gemfile.lock does NOT need to be explicitly created. `bundle install` will generate it automatically. One should avoid editing the Gemfile.lock in general.

### DIFF
--- a/samples/rails.md
+++ b/samples/rails.md
@@ -22,7 +22,6 @@ FROM ruby:2.5
 RUN apt-get update -qq && apt-get install -y nodejs postgresql-client
 WORKDIR /myapp
 COPY Gemfile /myapp/Gemfile
-COPY Gemfile.lock /myapp/Gemfile.lock
 RUN bundle install
 
 # Add a script to be executed every time the container starts.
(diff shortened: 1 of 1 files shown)


samples/rails.md
```diff
@@ -22,7 +22,6 @@ FROM ruby:2.5
 RUN apt-get update -qq && apt-get install -y nodejs postgresql-client
 WORKDIR /myapp
 COPY Gemfile /myapp/Gemfile
-COPY Gemfile.lock /myapp/Gemfile.lock
```
philkcw, line 25: The Gemfile.lock could change from day to day, even with the same Gemfile. If the Gemfile.lock changes on another dev's machine, and the change is pulled to my machine, won't WORKDIR sync the Gemfile.lock over to the container?

philkcw:
@usha-mandya what do you think?

craig-osterhout:
Thanks for contributing. The rails sample app has been taken out of the docs and moved to the Awesome Compose repo to hopefully facilitate better maintenance. If the issues still exists, please help open an issue there, https://github.com/docker/awesome-compose/blob/master/official-documentation-samples/rails/README.md

Issue #14686, opened 2022-05-06
$(pwd) doesnt work on windows


allenross356:
[get-started/06_bind_mounts.md](https://docs.docker.com/get-started/06_bind_mounts/)

Running the command in step 2 on Windows
$ docker run -dp 3000:3000 \
     -w /app -v "$(pwd):/app" \
     node:12-alpine \
     sh -c "yarn install && yarn run dev"

gives following error:
docker: Error response from daemon: create $(pwd): "$(pwd)" includes invalid characters for a local volume name, only "[a-zA-Z0-9][a-zA-Z0-9_.-]" are allowed. If you intended to pass a host directory, use absolute path.
See 'docker run --help'.

craig-osterhout:
Hi @allenross356, thanks for sharing the issue. 
If you are using Windows, please run the command in PowerShell.
The first command should specify that it's intended for another shell and not CMD.

Issue #12662, opened 2021-04-12
Windows CMD did not accept $(pwd)


Jack-R-Long:
[get-started/06_bind_mounts.md](https://docs.docker.com/get-started/06_bind_mounts/)

Had to replace the `$(pwd)` with the CMD equivalent `%cd%` in the following command:
`docker run -dp 3000:3000 -w /app -v "$(pwd):/app" node:12-alpine sh -c "yarn install && yarn run dev"`

See error below:
docker: Error response from daemon: create $(pwd): "$(pwd)" includes invalid characters for a local volume name, only "[a-zA-Z0-9][a-zA-Z0-9_.-]" are allowed. If you intended to pass a host directory, use absolute path.
See 'docker run --help'

craig-osterhout:
Thanks for sharing this issue. The example command is for PowerShell, not CMD. Closing this issue as PowerShell is specified in the documentation. If the issue still persists, please let us know.

craig-osterhout:
Duplicate of  #14686

PR #14721, opened 2022-05-11
Fix for issue #14686 and #12662

craig-osterhout:
### Proposed changes

Some users on Windows were running the commands in Command Prompt, since the first command did not specify which shell or operating system (other than the prompt symbol).
Modified wording of second command to make it clearer that if using Windows, then users should use PowerShell.

### Related issues (optional)

Fix for #14686 and #12662


jerae-duffin (approved):
Thank you! @craig-osterhout LGTM

PR #14727, opened 2022-05-12
Update rootless.md

delanym:
### Proposed changes

After installing (and using) docker rootless, I had a ~/.docker directory that I needed to delete since I was getting

`Current context "rootless" is not found on the file system, please check your config file at /home/docker/.docker/config.json`

### DIFF
--- a/engine/security/rootless.md
+++ b/engine/security/rootless.md
@@ -217,6 +217,8 @@ Removed /home/testuser/.config/systemd/user/default.target.wants/docker.service.
 
 Unset environment variables PATH and DOCKER_HOST if you have added them to `~/.bashrc`.
 
+Remove directory `~/.docker`.
+
 To remove the data directory, run `rootlesskit rm -rf ~/.local/share/docker`.
 
 To remove the binaries, remove `docker-ce-rootless-extras` package if you installed Docker with package managers.


thaJeztah (changes requested):
Hm.. I think removing whole of `.docker/` may be a bit of a large hammer. The `~/.docker/` directory contains _all_ of the `docker` CLI configuration, so removing that could potentially delete other configs that may be important to the user. If this problem happens, the "correct" fix would be to update the `~/.docker/config.json` file and to remove the `currentContext` property (so that the default context is used). This should not happen though (under normal circumstances), so would only be needed if something went wrong.


I had a look at the `dockerd-rootless-setuptool.sh` script, and current versions of the script should contain the changes of https://github.com/moby/moby/issues/43020. With that change included, running `dockerd-rootless-setuptool.sh uninstall` _should_ run a `docker context rm -f rootless`, which will remove the context, and switch back to using the "default" context (the `docker context rm -f` should remove the `currentContext` setting from `~/.docker/config.json`);

- https://github.com/moby/moby/blob/v20.10.18/contrib/dockerd-rootless-setuptool.sh#L419-L422
- https://github.com/moby/moby/blob/v20.10.18/contrib/dockerd-rootless-setuptool.sh#L366-L369

If you're seeing the "Current context "rootless" is not found on the file system" error, that means that somehow the context itself was removed, but the `~/.docker/config.json` was not updated, which _should_ not happen if the script ran successfully, so I'm trying to come up with reasons why that could happen;

- If the `cli_ctx_exists` check failed, the `cli_ctx_rm` command wouldn't be executed
- `cli_ctx_rm` currently also looks to fail (even with `-f` / `--force`) if the current context is not found (something we should fix in the CLI)

Did you see an error when running the script?

PR #14728, opened 2022-05-12
Publish updates from master

usha-mandya:


dockertopia:
Nothing to add besides the topics the team previously discussed about needing to align on release notes (dates, "names for the releases"[for e.g. here we have a extra header for the major as "version 20.10"] of headers, how issues are grouped, text used in links for issues, etc.). Note: I'm not sure the term tab here applies or is to be revisited later when Allie defines a new term.

PR #14738, opened 2022-05-12
Update index.md

pkube:
Clarification of internal vs. external, and default vs. optional behavior.

### Proposed changes
Review the specific wording, but the clarification is a good idea.

### DIFF
--- a/compose/compose-file/index.md
+++ b/compose/compose-file/index.md
@@ -2031,8 +2031,7 @@ ipam:
 
 ### internal
 
-By default, Compose implementations MUST provides external connectivity to networks. `internal` when set to `true` allow to
-create an externally isolated network.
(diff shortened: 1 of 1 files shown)


glours (changes requested):
[This file is coming from the Compose-specification](https://github.com/compose-spec/compose-spec/blob/master/spec.md#internal), we can't merge the modification in this repo.
BTW the writing of a specification is more formal than a documentation, for example you can't remove the `MUST` verb which is really important from a specification point of view.

We should close this PR and you can open a new one in the [compose specification repository](https://github.com/compose-spec/compose-spec/), but you will have to care of keeping the formalism of the specification.

usha-mandya:
Closing the PR based on the comment above. Please open a PR in the [compose specification repository](https://github.com/compose-spec/compose-spec/).

PR #14752, opened 2022-05-16
Update django.md

mic-max:
I believe this is the reason but was only guessing as I went through this example. Please let me know if this is an inaccurate statement and I can update with a better explanation.

### Proposed changes

Add explanation for why `requirements.txt` is copied and installed before copying the entire directory over.
Does this effectively `COPY` the `requirements.txt` into the destination twice? Or will Docker or the operating system detect that it's the same file. I don't see anywhere in the [COPY documentation](https://docs.docker.com/engine/reference/builder/#copy)  anything explicitly saying it will overwrite existing files either.

### DIFF
--- a/samples/django.md
+++ b/samples/django.md
@@ -41,6 +41,8 @@ and a `docker-compose.yml` file. (You can use either a `.yml` or `.yaml` extensi
    This `Dockerfile` starts with a [Python 3 parent image](https://hub.docker.com/r/library/python/tags/3/).
    The parent image is modified by adding a new `code` directory. The parent image is further modified
    by installing the Python requirements defined in the `requirements.txt` file.
+   The requirements are copied and installed before copying the entire current
+   directory simply to fail fast if the installation fails.
 
 4. Save and close the `Dockerfile`.
 


StefanScherer:
Thanks @mic-max for your contribution.
The idea of the two `COPY` instructions is to have a fast inner loop cycle when you develop the application. The `COPY . /code/` instruction is probably very fast, so there is no advantage to fail fast.

The time consuming step is the `pip install` to download and install tall dependencies. When you develop an application it is more likely that you change code and not always the `requirements.txt` file. When you rebuild the image layer caching can speed up the build time to just skip the `pip install` when the `requirements.txt` file hasn't changed. The `docker build` then only has to copy the sources again to build the new image.

Answering your question, yes the `requirements.txt` file will be copied twice with the two COPY instructions, but this is typically only a very small file, so there's no need to make the second COPY line more unreadable to optimise this, the `COPY . /code/` is still a good practise.

Do you want to rephrase your PR to highlight the layer caching instead of the fail fast?

mic-max:
> Thanks @mic-max for your contribution. The idea of the two `COPY` instructions is to have a fast inner loop cycle when you develop the application. The `COPY . /code/` instruction is probably very fast, so there is no advantage to fail fast.
> 
> The time consuming step is the `pip install` to download and install tall dependencies. When you develop an application it is more likely that you change code and not always the `requirements.txt` file. When you rebuild the image layer caching can speed up the build time to just skip the `pip install` when the `requirements.txt` file hasn't changed. The `docker build` then only has to copy the sources again to build the new image.
> 
> Answering your question, yes the `requirements.txt` file will be copied twice with the two COPY instructions, but this is typically only a very small file, so there's no need to make the second COPY line more unreadable to optimise this, the `COPY . /code/` is still a good practise.
> 
> Do you want to rephrase your PR to highlight the layer caching instead of the fail fast?

Thanks for the explanation, that makes sense `requirements.txt` should be quite small :)

PR #14766, opened 2022-05-18
Improve WSL 2 docs

djs55:
### Proposed changes

- described that having additional Linux distros is recommended, but is not a pre-req. I've split the Install section into 2: the basic install which is all you need for `docker` to work on Windows; and a separate section on enabling it for user Linux distros since that's optional (but recommended)
- added a note describing that the built-in distros `docker-desktop` and `docker-desktop-data` are implementation details and not suitable for general use.
- re-ordered the page slightly so now it's: why do I want this, the pre-reqs, download link, install steps and then best-practices. Previously the best-practices were before the download which was (IMHO) too early, we'd like to help the user get installed first + not make it look too complicated
- removed references to a very old Docker Desktop 2.3.0.2 (but the link was to latest, which is fine)
- fixed a reference to "running natively without emulation" to "run without having to manage Virtual Machines" <- before WSL 2 the state of the art was to run Hyper-V VMs yourself; with WSL 2 there is a VM which Windows manages automatically

### DIFF
--- a/desktop/windows/wsl.md
+++ b/desktop/windows/wsl.md
@@ -46,7 +46,16 @@ Ensure you have completed the steps described in the Prerequisites section **bef
 
     If you have installed Docker Desktop on a system that supports WSL 2, this option will be enabled by default.
 5. Click **Apply & Restart**.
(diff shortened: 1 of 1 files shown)


usha-mandya:
@aevesdocker Coud you PTAL?

PR #14768, opened 2022-05-18
ci: split build and publish workflows and set up cache

crazy-max:
the new workflow introduced in #14685 takes quite some time to build (~5m). the gem stage is the culprit here as it takes almost 4m to build some native extensions. this PR introduces GitHub Cache backend for our workflows and now it takes ~37s to build: https://github.com/crazy-max/docker.github.io/runs/6492119592?check_suite_focus=true#step:4:1

image

also as discussed with @thaJeztah, split build workflow and create a dedicated publish one to have better separation of concerns.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -1,0 +1,75 @@
+name: publish
+
+on:
+  push:
+    branches:
+      - master
+      - published
+
+jobs:
+  publish:
+    runs-on: ubuntu-20.04
+    steps:
+      -
+        name: Prepare
+        run: |
+          JEKYLL_ENV=development
+          if [ "${{ github.ref }}" = "refs/heads/master" ]; then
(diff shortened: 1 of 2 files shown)


.github/workflows/publish.yml
```diff
@@ -0,0 +1,75 @@
+name: publish
+
+on:
+  push:
+    branches:
+      - master
+      - published
+
+jobs:
+  publish:
```
thaJeztah, line 10: Perhaps in a follow-up we should consider adding a check-for-broken-links check in this workflow as well, to make sure no links are broken before publishing

thaJeztah (approved):
LGTM (if green)

thaJeztah:
I guess it's green, but we need to update the config again 😂

thaJeztah:
Alright; let's get this one in

PR #14769, opened 2022-05-18
Update install.md

anelmusictest:
Dash missing


dockertopia:
@anelmusictest thanks for the PR. The example is using Compose V2 syntax, https://docs.docker.com/compose/install/compose-plugin/ please have a look a the note on that page.

PR #14688, opened 2022-05-07
jekyll: create plugin to fetch remote resources

crazy-max:
~needs #14685~

create a Jekyll plugin to fetch remote resources so they will be available as preview in Netlify when PR are created. also adds another plugin to update api toc.

PR #14784, opened 2022-05-20
jekyll: create plugin to update sitemap

crazy-max:
follow-up https://github.com/docker/docker.github.io/pull/14688

In the same spirit as the plugin to fetch remote resources, create a new one to update our sitemap.

Also enables sitemap generation in development so we can preview sitemap changes. Since Jekyll 4, sitemap generation is pretty well optimized and takes less than 3 seconds for our docs so I think we can enable it.

I also added another commit which adds a sitemap xsl, so we can easily browse our sitemap:

image

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,7 +10,7 @@ ARG RUBY_VERSION=2.6.10
 ARG BUNDLER_VERSION=2.3.13
 
 ARG JEKYLL_ENV=development
-ARG DOMAIN=docs.docker.com
+ARG DOCS_URL=http://localhost:4000
 
 # Base stage for building
 FROM ruby:${RUBY_VERSION}-alpine AS base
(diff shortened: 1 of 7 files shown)


crazy-max:
Here is the preview link for the sitemap: https://deploy-preview-14784--docsdocker.netlify.app/sitemap.xml

usha-mandya (approved):

Dockerfile
```diff
@@ -64,7 +63,7 @@ else
   )
 fi
 find ${TARGET} -type f -name '*.html' | while read i; do
-  sed -i 's#\(<a[^>]* href="\)https://${DOMAIN}/#\1/#g' "$i"
+  sed -i 's#\(<a[^>]* href="\)${DOCS_URL}/#\1/#g' "$i"
```
thaJeztah, line 66: Let me check this one; I think the intent here was to remove some hard-coded URLs in the docs (e.g. from upstream MarkDown files) that have a full `https://docs.docker.com/` URL in them

URLs in the pages _generated_ by Jekyll are expected to all be relative (so `<a href=/foo/bar/>`, without a domain); the intent of that is so that the generated pages can be deployed to "any" domain (staging/local/elsewhere) without links going to the "live" domain.

So perhaps that part was intentionally set to `docs.docker.com`

As to the `sitemap.xml` having `docs.docker.com` hard-coded; that may also have been intentional at the time, as we've had cases where Google found preview sites, and (although we set `noindex` headers), we wanted to avoid it finding the `sitemap.xml`, and as a result indexing those preview domains (and as a "best effort", we just pointed it to the live website, which was the "least bad" option.

PR #14795, opened 2022-05-21
fix broken urls in swagger files for engine api

crazy-max:
URLs of some images are broken for the engine api because of a missing leading slash: https://docs.docker.com/engine/api/v1.41/

image

I took the opportunity with this change to also move the logic in a dedicated plugin that will run in `post_read` (e.g., will be done after `pre_read` like the fetch remote resources plugin) and also added `./docker-hub/api/*.yaml` pattern which was not covered previously.

We didn't catch that because htmlproofer ignores this folder: https://github.com/docker/docker.github.io/blob/8d6bd6bdb5c1b7b23d196e62760b93b89a588661/Dockerfile#L77

### DIFF
--- a/_plugins/fetch_remote.rb
+++ b/_plugins/fetch_remote.rb
@@ -49,14 +49,6 @@ module Jekyll
         end
       end
 
-      Jekyll.logger.info "  Fixing up URLs in swagger files"
-      Dir.glob("./engine/api/*.yaml") do |file_name|
-        Jekyll.logger.info "    #{file_name}"
-        text = File.read(file_name)
-        replace = text.gsub!("https://docs.docker.com/", "")
-        File.open(file_name, "w") { |file| file.puts replace }
-      end
-
       end_time = Time.now
       Jekyll.logger.info "done in #{(end_time - beginning_time)} seconds"
     end
--- a/_plugins/fix_urls.rb
+++ b/_plugins/fix_urls.rb
@@ -1,0 +1,24 @@
+require 'jekyll'
+require 'octopress-hooks'
+
+module Jekyll
+
+  class FetchRemote < Octopress::Hooks::Site
+    def post_read(site)
+      beginning_time = Time.now
+      Jekyll.logger.info "Starting plugin fix_urls.rb..."
+
+      Jekyll.logger.info "  Fixing up URLs in swagger files"
+      Dir.glob(%w[./docker-hub/api/*.yaml ./engine/api/*.yaml]) do |file_name|
+        Jekyll.logger.info "    #{file_name}"
+        text = File.read(file_name)
+        replace = text.gsub!("https://docs.docker.com", "")
+        File.open(file_name, "w") { |file| file.puts replace }
+      end
+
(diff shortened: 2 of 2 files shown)


thaJeztah:
Do you know what caused this? Because the original file has a full URL; is one of the replace scripts replacing too much? https://github.com/moby/moby/blob/master/docs/api/v1.41.yaml#L27

crazy-max:
Removing the exclusion from htmlproofer leads to only one failure:

```
#15 0.340 
#15 0.340
#15 84.29 Ran on 1708 files!
#15 84.29
#15 84.29
#15 84.29 - ./_site/engine/api/v1.19/index.html
#15 84.29   *  linking to internal hash #bind-docker-to-another-host-port-or-a-unix-socket that does not exist (line 124)
#15 84.29      <a href="https://docs.docker.com/engine/reference/commandline/dockerd/#bind-docker-to-another-host-port-or-a-unix-socket">[Bind Docker to another host/port or a Unix socket](https://docs.docker.com/engine/reference/commandline/dockerd/#bind-docker-to-another-host-port-or-a-unix-socket)</a>
#15 84.29
#15 84.29 HTML-Proofer found 1 failure!
```

I think we could just exclude `./_site/engine/api/v1.19/index.html`.

_plugins/fetch_remote.rb
```diff
@@ -49,14 +49,6 @@ def pre_read(site)
         end
       end
 
-      Jekyll.logger.info "  Fixing up URLs in swagger files"
-      Dir.glob("./engine/api/*.yaml") do |file_name|
-        Jekyll.logger.info "    #{file_name}"
-        text = File.read(file_name)
-        replace = text.gsub!("https://docs.docker.com/", "")
```
thaJeztah, line 56: Ah yeah, replace should not strip the `/`

Are there other places where we do this?

crazy-max:
> Do you know what caused this? Because the original file has a full URL; is one of the replace scripts replacing too much? https://github.com/moby/moby/blob/master/docs/api/v1.41.yaml#L27

Yes in https://github.com/docker/docker.github.io/pull/14688, when moved to a Jekyll plugin, I forgot to add the leading slash like it was before: https://github.com/docker/docker.github.io/pull/14688/files#diff-4aad9ddf294771293cfaa8016eb98c5a91c56e3f7e45c608df4be2d519a68d2eL38

thaJeztah (approved):
SGTM

_plugins/fetch_remote.rb
crazy-max: I was checking other pages too with hardcoded `docs.docker.com` which leads to unexpected behavior for review because it doesn't check the actual domain with htmlproofer. I have another branch that will fix that and remove hardcoded root url as you suggested in https://github.com/docker/docker.github.io/pull/14784#discussion_r878382327

thaJeztah:
Just checked the preview on my desktop (looks like mobile view doesn't include an icon), and icon looks good now 👍

PR #14801, opened 2022-05-22
Update bridge.md

kaan-diren:
As it is mentioned, it is incorrect. There is no difference between default and user-based bridges in terms of attaching or dettaching on the fly.


kaan-aigoritma (approved):

oguzhalit (approved):

oguzhalit (commented):
.

PR #14802, opened 2022-05-22
Fix markdown syntax error

heitorlisboa:
### Proposed changes

Fixed an italic marker on the [_Compose specification_](https://docs.docker.com/compose/compose-file/) page.

Marker was `-TEXT_`, and fixed to `_TEXT_`


dockertopia (approved):
Thanks for the fix. It is appreciated!

PR #10305, opened 2020-02-17
Add markdown lint json file

usha-mandya:
Adding markdown lint json to flag numbered lists
Updated numbered lists to fix copy/paste issues

PR #14806, opened 2022-05-23
switch to ruby markdownlint and add ci validation

crazy-max:
follow-up https://github.com/docker/docker.github.io/pull/10305

like htmlproofer, switch to ruby markdownlint gem and add a matrix in the validate job to run markdownlint on ci.

```shell
$ make mdl
# or
$ docker buildx bake mdl
```

### DIFF
--- a/.mdlrc.style.rb
+++ b/.mdlrc.style.rb
@@ -1,0 +1,7 @@
+# https://github.com/markdownlint/markdownlint/blob/master/docs/RULES.md
+
+# Start with all built-in rules.
+all
+
+rule 'MD013', :line_length => -1
+rule 'MD030', :ol_multi => 2
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,3 +1,0 @@
-webpack.config.js
-server/index.js
-server/env.js
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -28,6 +28,12 @@ jobs:
 
   validate:
     runs-on: ubuntu-20.04
+    strategy:
+      fail-fast: false
+      matrix:
+        target:
+          - htmlproofer
+          - mdl
     steps:
       -
         name: Checkout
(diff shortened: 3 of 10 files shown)


.mdlrc.style.rb
```diff
@@ -0,0 +1,7 @@
+# https://github.com/markdownlint/markdownlint/blob/master/docs/RULES.md
+
+# Start with all built-in rules.
+all
+
+rule 'MD013', :line_length => -1
+rule 'MD030', :ol_multi => 2
```
thaJeztah, line 7: These can only be configured in a Ruby source file? Or would it still be possible to have it use a configuration file?

crazy-max: Yes rules must be a ruby source file: https://github.com/markdownlint/markdownlint/blob/master/docs/creating_rules.md

There is also the `.mdlrc` but that's just to define command line flags as config: https://github.com/markdownlint/markdownlint/blob/master/docs/configuration.md#mdl-configuration

crazy-max: Something like this:

```dockerfile
# mdl is a lint tool for markdown files
FROM gem AS mdl
COPY <<-"EOT" /tmp/mdlrc.style.rb
	# https://github.com/markdownlint/markdownlint/blob/master/docs/RULES.md
	# Start with all built-in rules.
	all
	rule 'MD013', :line_length => -1
	rule 'MD030', :ol_multi => 2
EOT
RUN --mount=type=bind,target=. \
  mdl --rules /tmp/mdlrc.style.rb .
```

thaJeztah: hm... that's a pity. I was hoping we could keep "config" and "code" a bit separate (and a JSON or YAML, or TOML file may be less "daunting' than editing a source-code file - and possibly easier to discover)

crazy-max: yeah that's something we have to deal with also in js like https://github.com/docker/build-push-action/blob/master/jest.config.ts

crazy-max:
rebased

thaJeztah (approved):
LGTM

PR #14813, opened 2022-05-23
Update to use "/etc/apt/keyrings" per APT maintainers recommendations

tianon:
See https://tracker.debian.org/news/1305679/accepted-apt-240-source-into-unstable/:

    * Install an empty /etc/apt/keyrings directory.
      This directory is intended to provide an alternative to
      /usr/share/keyrings for placing keys used with signed-by.

See also https://wiki.debian.org/DebianRepository/UseThirdParty?action=diff&rev2=47&rev1=46 (which was edited following a discussion with the APT maintainers about the expected usage):

> If future updates to the key will be managed by an apt/dpkg package as recommended below, then it SHOULD be downloaded into `/usr/share/keyrings` using the same filename that will be provided by the package. If it will be managed locally , it SHOULD be downloaded into `/etc/apt/keyrings` instead.

### DIFF
--- a/engine/install/debian.md
+++ b/engine/install/debian.md
@@ -92,7 +92,8 @@ from the repository.
 2.  Add Docker's official GPG key:
 
     ```console
-    $ curl -fsSL {{ download-url-base }}/gpg | sudo gpg --dearmor -o /usr/share/keyrings/docker-archive-keyring.gpg
+    $ sudo mkdir -p /etc/apt/keyrings
+    $ curl -fsSL {{ download-url-base }}/gpg | sudo gpg --dearmor -o /etc/apt/keyrings/docker.gpg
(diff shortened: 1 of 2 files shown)


thaJeztah:
Does that mean that potentially we can expect reports similar to https://github.com/docker/for-linux/issues/1349 ?

tianon:
Hmm, I don't think so -- that one looks like it's because we had both `docker-ce.list` _and_ `docker.list` with different values.

thaJeztah:
Ah, right makes sense; I thought it would cache the previous  value somewhere 👍

thaJeztah (approved):
LGTM

dockertopia (commented):
Acknowledged.

PR #14828, opened 2022-05-25
engine: remove some redundant information

thaJeztah:
### engine: remove instructions for "nightly" and "test" channels

The "nightly" channel hasn't been updated for a long time, so while it's
not actively updated, we should not show instructions for it.

The "test" channel is still updated in preparation of new releases, but
including the instructions in between the regular installation procedure
introduced quite some "noise" that's irrelevant to most users (who would
be installing the "stable" channel).

This patch removes the instructions to enable the nightly and test channels
from the installation instructions.

While making these changes, I also synced somme wording between distros
for consistency, and removed a stray section about storage-drivers, as
that's already outlined in the "optional post-installation steps".

### engine: remove warning about Ubuntu 16.04 "xenial" EOL

This warning has been in place for a Year, and while there's
still users running on this version of Ubuntu, it's not really
our responsibility to notify them about their distro version
being EOL.

### engine: move note about package split to 18.09 release notes

containerd, docker engine and the docker cli were split into separate
packages in the 18.09 release. We kept this note on the release-notes
for current releases, but by now it's mostly "old news", so move the
warning to the 18.09 release notes page (which was the release in which
this change was made).

### engine: remove some mentions of old Ubuntu and Engine versions

These were notes about distro versions that have been EOL for a long
time, and an old version of docker engine that's EOL as well.

### DIFF
--- a/engine/release-notes/18.09.md
+++ b/engine/release-notes/18.09.md
@@ -5,6 +5,19 @@ toc_max: 2
 skip_read_time: true
 ---
 
+> **Note:**
+>
+> With this release, the daemon, client and container runtime are now all shipped
+> in separate packages. When updating, you need to update all packages at the same
+> time to get the latest patch releases for each. For example, on Ubuntu:
+> 
+> ```console
+> $ sudo apt-get install docker-ce docker-ce-cli containerd.io
+> ```
+> 
+> See the [installation instructions](../install/index.md) for the corresponding
+> Linux distro for details.
+
 ## 18.09.9
 2019-09-03
 
--- a/engine/install/centos.md
+++ b/engine/install/centos.md
@@ -80,7 +80,7 @@ from the repository.
 {% assign download-url-base = "https://download.docker.com/linux/centos" %}
 
 Install the `yum-utils` package (which provides the `yum-config-manager`
-utility) and set up the **stable** repository.
+utility) and set up the repository.
 
 ```console
 $ sudo yum install -y yum-utils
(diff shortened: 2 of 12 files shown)


engine/release-notes/18.09.md
```diff
@@ -5,6 +5,14 @@ toc_max: 2
 skip_read_time: true
 ---
 
+> **Note:**
+>
+> The client and container runtime are now in separate packages from the daemon
+> in Docker Engine 18.09. Users should install and update all three packages at
+> the same time to get the latest patch releases. For example, on Ubuntu:
```
dockertopia, line 12:
```suggestion
> With this release, the daemon, client and container runtime are now all shipped in separate packages. So you > need to install and later update all three packages at the same time to get the latest patch releases. For example, in Ubuntu:
```

dockertopia (approved):
Besides my suggestions, looks good to me. If we no longer need to talk about all the different builds streamlining seems good 👯.

engine/release-notes/18.09.md
thaJeztah: I think we have that covered in the install page; for a "fresh" install, there should not be a need to specify all packages (they will be installed as dependency), so it's mostly for updating packages to force them being updated, but we mention those steps;


Screenshot 2022-05-26 at 00 20 39
Screenshot 2022-05-26 at 00 21 04

thaJeztah:
Tweaked the release-notes note based on your suggestions

Let me bring this one in, thanks!

engine/release-notes/18.09.md
dockertopia: Yes. We are telling users to remove all the packages ( to remove older installations ) in the instruction but the nexus behind the change (for some users at least) in number/structure of packages is the bit I thought was interesting mentioning here in addition to in the release notes. Although it's bean a while (18.xx) maybe it is still interesting.

thaJeztah: Yeah, I don't think many people will arrive at the 18.09 release notes; I mostly wanted to preserve the information (instead of removing it altogether).

PR #14850, opened 2022-05-29
Add a note that local is a bash keyword

crdotson:
The example will run from the command line, but not as part of a bash script.  It took me an embarrassing several minutes to figure out why my start script wouldn't run with the cryptic error "docker: invalid reference format," so I thought a one-line note was worth making in the documentation.

### DIFF
--- a/config/containers/logging/local.md
+++ b/config/containers/logging/local.md
@@ -54,6 +54,8 @@ $ docker run \
       alpine echo hello world
 ```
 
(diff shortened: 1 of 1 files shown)


aevesdocker (approved):

aevesdocker:
Thank you @crdotson for your contribution. I've approved and merged your PR. Have a great day.

Issue #14683, opened 2022-05-06
Using 'external' marked as deprecated by Compose v2.5


bfren:
[compose/networking.md](https://docs.docker.com/compose/networking/)

See https://github.com/docker/compose-cli/issues/1856

PR #13958, opened 2021-12-11
Update networking.md

andrewsurrey:

PR #14851, opened 2022-05-30
Update code example for pre-existing network

grappler:
Update the code example to match the new syntax.

### Related issues (optional)

Fixes https://github.com/docker/docker.github.io/issues/14683

### DIFF
--- a/compose/networking.md
+++ b/compose/networking.md
@@ -185,8 +185,8 @@ services:
   # ...
 networks:
   default:
-    external:
-      name: my-pre-existing-network
+    name: my-pre-existing-network
+    external: true
 ```
 
 Instead of attempting to create a network called `[projectname]_default`, Compose looks for a network called `my-pre-existing-network` and connect your app's containers to it.


dockertopia (commented):
@glours Could you PTAL? The sample code looks fine to me based on https://github.com/docker/docker.github.io/pull/13958/. TWISIT accepting changes in this PR would revert that update.

grappler:
@dockertopia Thank you for your review. I don't see any reasoning why https://github.com/docker/docker.github.io/pull/13958 was made. It appears to be incorrect for me. My suggested change was previously applied here: https://github.com/docker/docker.github.io/pull/6124

Please see the related comment: https://github.com/docker/compose-cli/issues/1856#issuecomment-1108552064

thaJeztah:
Hmm.. yes, https://github.com/docker/docker.github.io/pull/13958 looks a bit odd indeed; I think the `name` field was added in compose schema 2.3  and 3.5 (through this PR: https://github.com/docker/compose/pull/5448)

glours:
Hello @grappler 
This change isn't compliant with [the compose specification]( https://github.com/compose-spec/compose-spec/blob/master/spec.md#external)
@dockertopia we should update the network documentation page, especially [the pre-existing network section](https://docs.docker.com/compose/networking/#use-a-pre-existing-network)

glours (approved):
LGTM regarding the [Compose specification]( https://github.com/compose-spec/compose-spec/blob/master/spec.md#external)

glours:
@grappler Thanks for catching this one

dockertopia (approved):
Thanks @grappler. That was a good catch. Felt compelled to double check based on previous change to the file.

PR #14870, opened 2022-06-01
updated TOC for onboarding

craig-osterhout:
### Proposed changes
Updated TOC for Business onboarding to add more context that it is for Business and not for Team


jerae-duffin (approved):
LGTM

PR #14889, opened 2022-06-06
Update redoc and add DVP Data API doc

chrispatrick:
### Proposed changes

- Updates redoc to the latest version
- Updates existing Hub API doc to OAPI 3
- Adds new DVP Data API doc
- Links to DVP Data API doc from the "API Documentation" section
- Fixes a bug in URL processing


css/api-reference.scss
```diff
@@ -37,18 +56,14 @@ redoc {
...
+  a {
     color: $primary-links !important;
   }
 
-  .menu-item-title {
-    text-transform: none;
-  }
-
   // Disable request-type badges, see https://github.com/Rebilly/ReDoc/issues/61
   .operation-type {
```
technicallyjosh, line 68: suggestion: Remove this hack to remove it. I'd really like to see these. It makes it much easier when browsing for specific keywords.

technicallyjosh (approved):
Looks great! Had a suggestion. Thanks for the improvement!

css/api-reference.scss
chrispatrick: Hmm, worth considering, but I'd vote for doing that in a separate PR.

js/redoc.standalone.js
```diff
@@ -0,0 +1,1740 @@
+/*! For license information please see redoc.standalone.js.LICENSE.txt */
```
thaJeztah, line 1: Was `.standalone` the new name they use for this? Or was this file compiled by you? Does a minified version exist that we could use?

_plugins/fix_urls.rb
```diff
@@ -12,7 +12,7 @@ def post_read(site)
       Dir.glob(%w[./docker-hub/api/*.yaml ./engine/api/*.yaml]) do |file_name|
         Jekyll.logger.info "    #{file_name}"
         text = File.read(file_name)
-        replace = text.gsub!("https://docs.docker.com", "")
+        replace = text.gsub("https://docs.docker.com", "")
```
thaJeztah, line 15: What does this change do? I think the intent here is to make full URLs (including domain name) relative (no domain name), so that when hosting the docs on a preview domain (or locally) does not link to / use / depend on the live documentation at docs.docker.com.

chrispatrick: Yes, that should still work. The issue here is that it used to fail if there were no matches.

js/redoc.standalone.js
chrispatrick: Yeah, this is just taken from redoc. It looks minified already, apart from the line breaks for some reason.

thaJeztah: I was hoping to keep the same name, and just replace the file, but it is what it is then 😅 🤷‍♂️

docker-hub/api/dvp.yaml
```diff
@@ -0,0 +1,564 @@
...
+info:
+  title: DVP Data API
+  version: 1.0.0
+  x-logo:
+    url: https://docs.docker.com/images/logo-docker-main.png
+    href: /reference
+  description: |
+    The Docker DVP Data API allows Docker Verified Publishers to view image pull analytics data for their namespaces
+servers:
+  - url: https://hub-stage.docker.com/api/dvp/v1
```
thaJeztah, line 11: This contains an URL for staging; probably needs to be updated

docker-hub/api/dvp.yaml
```diff
@@ -0,0 +1,564 @@
+openapi: 3.0.0
+info:
+  title: DVP Data API
+  version: 1.0.0
+  x-logo:
+    url: https://docs.docker.com/images/logo-docker-main.png
+    href: /reference
+  description: |
+    The Docker DVP Data API allows Docker Verified Publishers to view image pull analytics data for their namespaces
```
thaJeztah, line 9: Should we add a link to the DVP programme? (https://docs.docker.com/docker-hub/publish/)

_data/toc.yaml
```diff
@@ -1030,6 +1030,8 @@ reference:
         title: v1.18 reference
   - title: Docker Hub API
     path: /docker-hub/api/latest/
+  - title: DVP Data API
+    path: /docker-hub/api/dvp/
```
thaJeztah, line 1034: I wonder if we should start to consider making `Docker HUB API` a sub-section under which we'll list the Docker Hub related APIs. I know we discussed that some time ago, but at the time we only had a single API, so no grouping was needed, but now that there's two (or three, if we would consider the "registry API" to be part), perhaps it makes sense to start doing so.

This would also allow for an `index.md` / overview to be added that outlines what APIs exist (and for which purpose), and possibly other content to introduce users to the APIs (and where to learn more), or highlight "changes" (if relevant), similar to the "Engine API" section.

Screenshot 2022-07-04 at 11 09 20

docker-hub/api/dvp.yaml
chrispatrick: I'm not actually sure if that gets rendered anywhere, but will update in case.

docker-hub/api/dvp.yaml
chrispatrick: Sure, no harm in that.

_data/toc.yaml
chrispatrick: Could do. Not sure who would be best placed to own that content though...

docker-hub/api/dvp.yaml
```diff
@@ -0,0 +1,562 @@
...
+          description: Success
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/NamespaceData'
+  /data/{namespace}:
+    get:
+      tags: [discovery]
+      summary: Get namespace
+      description: Gets metadata associated with specified namespace
```
mikeparker, line 48: Can we explain in plain english what this 'metadata' is and why a user wants it? Is it just the extra repos they have access to right now, with the option to extend further in future? Is it clear what the flow of using this data for subsequent calls is?

docker-hub/api/dvp.yaml
```diff
@@ -0,0 +1,562 @@
...
+      description: Gets a list of namespaces and repos which have data available
+      operationId: getNamespaces
+      responses:
+        '200':
+          description: Success
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/NamespaceData'
+  /data/{namespace}:
```
mikeparker, line 44: I think we were trying to move towards a more standardised url format with a preceding name for each variable

e.g. `/data/namespace/{namespace}`

Whilst more verbose, it allows us to add additional verbs in future such as `/data/users` without the API thinking that `users` is a  namespace. Might be worth asking on #hub-backend to see what the current state of thinking is.

chrispatrick: I suppose we could do that. I'm not entirely convinced it is necessary in such a simple API - what would `data/users` even refer to?

But I'm happy to make the change.

_data/toc.yaml
thaJeztah: Additional content/markdown (if we choose to add it) would be maintained in this repository, but can be added later. Changing the TOC to use a section would look something like;

```yaml
  - sectiontitle: Docker Hub API
    section:
      - title: Docker Hub API
        path: /docker-hub/api/latest/
      - title: DVP Data API
        path: /docker-hub/api/dvp/
  - title: Registry API
    path: /registry/spec/api/
  - sectiontitle: Docker Extension SDK API (Beta)
  # ...
```

chrispatrick: Sure, I meant more of whose responsibility would that be?

chrispatrick: And do we want to tackle that now or add it to the backlog for the next iteration?

thaJeztah: If we consider it a good change to have the Docker Hub API's grouped in the navigation, I would consider doing that as  part of this PR (as this changes the docs to have multiple docs for the hub API).

w.r.t "responsibility" (assuming you're talking about the other content (index page, change log, etc); I think it'll be difficult to make a "hard cut" between who's responsible; I'd consider if a shared responsibility for all teams working on Docker Hub APIs. I would expect maintenance for an overview page to be low (changes would be needed only if a new API is added, e.g., but of course depends on what we want to show on it). Perhaps someone from the documentation team could write a first draft for an index page.

chrispatrick: OK, we'll I've grouped it for now but haven't added an index page or anything.

chrispatrick:
Pushed some updates. 😄

thaJeztah:
Could you squash some of the commits, and group them into logical steps? I don't think we should squash _all_ commits down to one, but perhaps something like;

1. Update redoc to latest version
2. Update HUB API docs to OpenApi 3
3. Add separate DVP Data API docs in OpenAPI 3

chrispatrick:
Squashed down to 4 clearer commits

docker-hub/api/latest.yaml
```diff
@@ -1,104 +1,137 @@
...
   - name: resources
     x-displayName: Resources
-    description: |
+    description: >
       The following resources are available to interact with the documented API:
 
+
       - <a href="https://github.com/docker/hub-tool#readme" target="_blank">Docker Hub CLI tool</a> (currently experimental)
+
       - <a href="https://www.postman.com/devrel/workspace/docker-hub/collection/13191452-7d2f1375-9705-4db6-a47b-8235286e58a1" target="_blank">Postman Collection</a> (built by the Postman team)
```
thaJeztah, line 28: Not new, so probably best for a separate PR, but looks like this link is broken (no longer exists); https://www.postman.com/devrel/workspace/docker-hub/collection/13191452-7d2f1375-9705-4db6-a47b-8235286e58a1

thaJeztah (approved):
Overall this LGTM (but perhaps some more eyes in case I missed something wouldn't hurt).

I also did a quick check on the "Engine API" docs with this, to 

There's some minor tweaking we may want to do in the layout; looks like spacing is a bit different, and the new "content type" (which is nice!) looks to be overlapping a bit with the headings (which could be tweaked a bit), same for the `400`, `500`, xxx status codes, which also overlap with the heading (and seem to be (semi-)bold);

Old:

Screenshot 2022-07-04 at 17 48 10

New:

Screenshot 2022-07-04 at 17 42 06

The "query parameters" also seem to have lost the `?` tool-tip (not critical), and now using ALL CAPS (which may be a bit "shout-y" (although the old one had this for PARAMETERS, so perhaps not an issue). Spacing is a bit different here as well, with the left-hand column taking (too much?) space; causing some text to be wrapped in this example;

Old:

Screenshot 2022-07-04 at 17 43 25

New:

Screenshot 2022-07-04 at 17 44 51

docker-hub/api/latest.yaml
chrispatrick: Yep, looks like we should raise that as a separate issue.

chrispatrick:
I think some of the styling issues are just down to how our custom CSS interacts with the new redoc CSS. We can definitely look at improving it, but I don't think there are any functional issues currently.

For reference, this is the "default" styling - https://redocly.github.io/redoc/

PR #14897, opened 2022-06-08
Update index.md

aryzing:
### Proposed changes

Most people are going to run this on their local machine, makes sense to map to a port that is more likely to be available than 80.


aryzing:
I think we're missing the point here: the docs provide code that's clearly not going to work for most users. It would certainly help users learning docker having code (be it a less frequent port number or an explanation alerting them of the possible failure they'll encounter when using port 80) that works for them most of the time.

PR #14905, opened 2022-06-09
build: add guides

crazy-max:
follow-up #14644

~needs https://github.com/docker/buildx/pull/1140 so keeping in draft for now~

Fetch build guides ~and bake~ from upstream buildx repo in https://github.com/docker/buildx/tree/master/docs/guides

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -163,6 +163,17 @@ fetch-remote:
           - "!docs/README.md" # readme to make things nice in the compose-cli repo, but meaningless here
           - "!docs/architecture.md" # Compose-CLI architecture, unrelated to cloud integration
 
+  - repo: "https://github.com/docker/buildx"
+    ref: "master"
+    paths:
+      - dest: "build/guides"
+        src:
+          - "docs/guides/*.md"
+          - "!docs/guides/cicd.md" # CI/CD documentation is already handled in ci-cd/github-actions
+      - dest: "build/bake"
+        src:
+          - "docs/guides/bake/**"
+
   - repo: "https://github.com/distribution/distribution"
     default_branch: "main"
     ref: "release/2.7"
--- a/_data/toc.yaml
+++ b/_data/toc.yaml
@@ -1396,6 +1396,38 @@ manuals:
           title: Working with Buildx
         - path: /build/buildx/install/
           title: Install Buildx
+    - sectiontitle: Bake
+      section:
+        - path: /build/bake/
+          title: Overview
+        - path: /build/bake/file-definition/
+          title: File definition
+        - path: /build/bake/configuring-build/
+          title: Configuring builds
+        - path: /build/bake/hcl-funcs/
+          title: User defined HCL functions
+        - path: /build/bake/build-contexts/
+          title: Build contexts and linking targets
+        - path: /build/bake/compose-file/
+          title: Building from Compose file
+    - sectiontitle: Guides
+      section:
+        - path: /build/guides/cni-networking/
+          title: CNI networking
+        - path: /build/guides/custom-network/
+          title: Using a custom network
+        - path: /build/guides/custom-registry-config/
+          title: Using a custom registry config
+        - path: /build/guides/kubernetes-builder/
+          title: Kubernetes builder
+        - path: /build/guides/opentelemetry/
+          title: OpenTelemetry support
+        - path: /build/guides/registry-mirror/
+          title: Define a registry mirror
+        - path: /build/guides/remote-builder/
+          title: Remote builder
+        - path: /build/guides/resource-limiting/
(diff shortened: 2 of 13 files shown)


_config.yml
```diff
@@ -150,6 +150,17 @@ fetch-remote:
           - "!docs/README.md" # readme to make things nice in the compose-cli repo, but meaningless here
           - "!docs/architecture.md" # Compose-CLI architecture, unrelated to cloud integration
 
+  - repo: "https://github.com/docker/buildx"
+    ref: "master"
+    paths:
+      - dest: "build/guides"
+        src:
+          - "docs/guides/*.md"
+          - "!docs/guides/cicd.md" # CI/CD documentation is already handled in ci-cd/github-actions
```
dockertopia, line 159: I'm not sure what we're doing here, is this about removing something from the docs that is some repo? If so, I'm not sure that is the right way forward.

crazy-max: We are fetching markdown files from buildx repo in https://github.com/docker/buildx/tree/master/docs/guides (except ci-cd) and add them to `build/guides` folder here.
Will be available in `Manuals > Docker Build > Guides`.

crazy-max:
```
#16 140.2 - ./_site/build/guides/kubernetes-builder/index.html
#16 140.2   *  linking to internal hash #build-multi-platform-images that does not exist (line 231)
#16 140.2      <a href="https://docs.docker.com/buildx/working-with-buildx/#build-multi-platform-images">multi-architecture images</a>
#16 140.2   *  linking to internal hash #kubernetes-driver-1 that does not exist (line 227)
#16 140.2      <a href="https://docs.docker.com/engine/reference/commandline/buildx_create/#kubernetes-driver-1">buildx create</a>
```

Oh we have to update the links now that buildx has moved. Will do that.

crazy-max:
moving to draft as we might create stub files so we can maintain frontmatter section here. see #15165 for more info.

_data/toc.yaml
```diff
@@ -1396,6 +1394,40 @@ manuals:
...
+          title: File definition
+        - path: /build/bake/configuring-build/
+          title: Configuring builds
+        - path: /build/bake/hcl-funcs/
+          title: User defined HCL functions
+        - path: /build/bake/build-contexts/
+          title: Build contexts and linking targets
+        - path: /build/bake/compose-file/
+          title: Building from Compose file
+    - sectiontitle: Guides
```
jedevc, line 1411: Ahh this seems strange to me, but not sure whether there's a better alternative.

Having a section in Manuals called Guides, when we have a Guides category :thinking: I feel like it's better to maybe be consistent with the rest of our docs for now, and put it into Guides?

Other options:
- Don't call them "Guides" but something like "Advanced usage"? But then of course, what happens if we add more entry-level guides/tutorials, do those go into the guides section (splitting them up), or do we keep them together, which seems inconsistent with the rest of the docs.
- Rework the buildkit guides to be more like the guides in the rest of the docs, by making them more tutorial like, etc - maybe the [remote builder guide](https://github.com/docker/buildx/blob/master/docs/guides/remote-builder.md) is already like this in style?

_data/toc.yaml
```diff
@@ -1396,6 +1394,40 @@ manuals:
...
+        - path: /build/guides/kubernetes-builder/
+          title: Kubernetes builder
+        - path: /build/guides/multi-stage-build/
+          title: Multi-stage builds
+        - path: /build/guides/opentelemetry/
+          title: OpenTelemetry support
+        - path: /build/guides/registry-mirror/
+          title: Define a registry mirror
+        - path: /build/guides/remote-builder/
+          title: Remote builder
```
jedevc, line 1428: We should group this near the kubernetes builder. Maybe we also want to call it `<x>-driver` instead of `<x>-builder` - really not sure on that though.

build/buildx/index.md
```diff
@@ -157,25 +157,7 @@ FROM alpine
...
-request.
-
-There is also support for custom build rules from HCL/JSON files allowing
-better code reuse and different target groups. The design of bake is in very
-early stages, and we are looking for feedback from users. Let us know your
-feedback by creating an issue in the [Docker Buildx](https://github.com/docker/buildx/issues){:target="_blank" rel="noopener" class="_"}
-GitHub repository.
+## High-level build options with Bake
+
+Check out our guide about [Bake](../bake/index.md) to get started with the
```
jedevc, line 162: Do we want to call out that bake is still experimental here as well?

jedevc (commented):
This looks mostly good to me, I think we'd probably want to do some follow-ups to improve the consistency of our buildx docs - but it'll be good to have this info on the docs site I think.

PR #14906, opened 2022-06-09
dockerfile: add buildkit syntax reference

crazy-max:
depends on https://github.com/moby/buildkit/pull/2902 so keeping in draft

Fetch dockerfile reference from buildkit repo as well as the buildkit syntax in https://github.com/crazy-max/buildkit/tree/dockerfile-ref/frontend/dockerfile/docs (use my fork for now):

image

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -127,7 +127,6 @@ fetch-remote:
           - "docs/deprecated.md"
       - dest: "engine/reference"
         src:
-          - "docs/reference/builder.md"
           - "docs/reference/run.md"
       - dest: "engine/reference/commandline"
         src:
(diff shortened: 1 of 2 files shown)


dockertopia (approved):
@crazy-max I'm fine with publishing this asap to have it public somewhere as a starting point, but I think we're adding a lot of information into reference guides (e.g. this whole overview) that should be in the Manuals. References should be more like cheat sheets than dictionaries or Wikipedia articles. I also think that we need a product-pov (in how we organize the topics as well as) in naming features, options etc. instead of repo/project view.
I have also noticed that we call examples to items which are just samples and sometimes we don't even say what the sample illustrates.

crazy-max:
> @crazy-max I'm fine with publishing this asap to have it public somewhere as a starting point, but I think we're adding a lot of information into reference guides (e.g. this whole overview) that should be in the Manuals. References should be more like cheat sheets than dictionaries or Wikipedia articles. I also think that we need a product-pov (in how we organize the topics as well as) in naming features, options etc. instead of repo/project view. I have also noticed that we call examples to items which are just samples and sometimes we don't even say what the sample illustrates.

@dockertopia This PR is still draft. The idea here is to break reference in subsections but still need more work as you said. Can you look at https://github.com/docker/docker.github.io/pull/14948 instead? This is the one that just fetch the dockerfile reference docs from buildkit repo so we are aligned.

crazy-max:
clsoing this one since #14948 has been merged. will open a new one for new reference location when dockerfile changelog will be there.

PR #14910, opened 2022-06-09
build: move buildx to build section

crazy-max:
follow-up https://github.com/docker/docker.github.io/pull/14642

Moved "Docker Buildx" section under "Docker Engine" to the new "Docker Build" section and split buildx page with an overview page and an install page. I think it makes sense to have a dedicated install page, specially for the /go redirect to install buildx.

This is just a move without any drastic changes for now. @dockertopia You can make some changes to the install page in a follow-up as you suggested to have something similar to the compose page.

### DIFF
--- a/_data/toc.yaml
+++ b/_data/toc.yaml
@@ -1388,6 +1388,12 @@ manuals:
       title: Overview
     - path: /build/hellobuild/
       title: Hello Build
+    - sectiontitle: Buildx
+      section:
+        - path: /build/buildx/
+          title: Working with Buildx
+        - path: /build/buildx/install/
+          title: Install Buildx
 
 - sectiontitle: Docker Compose
   section:
--- a/build/buildx/install.md
+++ b/build/buildx/install.md
@@ -1,0 +1,78 @@
+---
+title: Install Docker Buildx
+description: How to install Docker Buildx
+keywords: build, buildx, buildkit
+---
+
+## Docker Desktop
+
+Docker Buildx is included in [Docker Desktop](../../desktop/index.md) for
+Windows, macOS, and Linux.
+
+## Linux packages
+
+Docker Linux packages also include Docker Buildx when installed using the
+[DEB or RPM packages](../../engine/install/index.md).
+
+## Dockerfile
+
+Here is how to install and use Buildx inside a Dockerfile through the
+[`docker/buildx-bin`](https://hub.docker.com/r/docker/buildx-bin) image:
+
+```dockerfile
+FROM docker
+COPY --from=docker/buildx-bin:latest /buildx /usr/libexec/docker/cli-plugins/docker-buildx
+RUN docker buildx version
+```
+
+## Manual download
+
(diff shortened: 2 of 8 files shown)


dockertopia (commented):
Shouldn't we combine this with the draft PR efforts in order to merge a first version of the new build section? Mostly I'm having the overview topics in mind.
On a smaller note, and on an revision level, the case use for buildx/Buildx is not consistent.

_data/toc.yaml
```diff
@@ -1388,6 +1388,12 @@ manuals:
       title: Overview
     - path: /build/hellobuild/
       title: Hello Build
+    - sectiontitle: Buildx component
```
jedevc, line 1391: Could we just title this "Buildx"?

build/buildx/install.md
```diff
@@ -0,0 +1,78 @@
...
+* `C:\Program Files\Docker\cli-plugins`
+
+> **Note**
+>
+> On Unix environments, it may also be necessary to make it executable with `chmod +x`:
+> ```shell
+> $ chmod +x ~/.docker/cli-plugins/docker-buildx
+> ```
+
+## Dockerfile
```
jedevc, line 60: Could we have this section above "Manual Packages"? It's such a short section, and I think putting it below the manual section means it's more likely to get missed by users (who might otherwise follow the manual installation guidelines).

jedevc (approved):

build/buildx/install.md
dockertopia: I had some high-level feedback for this PR, but I was in meetings. So, our build doc set first pages have changes since this. I think it poses a good opportunity to change some things.
So on the:
- order or pages: to keep inline "Install" types of topics follow the overview. So I would also move it up.
- structure of buildx: I think we should work on the content as if we already had the features list (simple and advanced use cases ready and you could link from there to here). So, even if, for now, they are going to be short pages. I think we make what users are trying to accomplish more discoverable by 2) Working with Buildx being only the intro (about config) and high-level options 3) promoting the sections to pages (so they show on the primary menu) 4) Check if the names of these pages are the best matches for what users would be searching. 
This will allow us to 1) have the driver for content to be tasks we want to accomplish with Build/x instead of a descriptive/system pov and 2) have room to expand on each of those use cases. 
So
Buildx
\ overview
\ install
\"config"
\working with builders
\multiplatform images

dockertopia (changes requested):
Inline: https://github.com/docker/docker.github.io/pull/14910#discussion_r924576665

build/buildx/install.md
dockertopia: I can delay some work and do these changes instead today. Do you mind if I commit to the branch?

crazy-max: Can we see that in a follow-up please? This PR is only about moving buildx and preserve the current state.

dockertopia (approved):
As agreed, making the proposed changes in a follow-up PR.

PR #14936, opened 2022-06-13
docs: context refers to desktop-linux

tomwillfixit:
The docker context ls command shows the output as desktop-linux but the context use command doesn't match this.

### Proposed changes

The output from the context ls command does not match the context use command later in the documentation.

### DIFF
--- a/desktop/linux/install.md
+++ b/desktop/linux/install.md
@@ -202,12 +202,12 @@ default
 Current context is now "default"
 ```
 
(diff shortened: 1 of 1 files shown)


usha-mandya:
@aiordache Could you PTAL?

aiordache (approved):
Thank you!

usha-mandya (approved):

PR #14965, opened 2022-06-18
fix networking reference

arimal199:

### DIFF
--- a/engine/swarm/networking.md
+++ b/engine/swarm/networking.md
@@ -15,7 +15,7 @@ A Docker swarm generates two different kinds of traffic:
 
 This topic discusses how to manage the application data for your swarm services.
 For more details about swarm networking in general, see the
-[Docker networking reference architecture](https://success.docker.com/Architecture/Docker_Reference_Architecture%3A_Designing_Scalable%2C_Portable_Docker_Container_Networks){: target="_blank" rel="noopener" class="_" }.
+[Docker networking reference architecture](https://docs.docker.com/network/){: target="_blank" rel="noopener" class="_" }.
 
 The following three network concepts are important to swarm services:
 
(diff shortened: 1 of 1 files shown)


engine/swarm/networking.md
```diff
@@ -15,7 +15,7 @@ A Docker swarm generates two different kinds of traffic:
 
 This topic discusses how to manage the application data for your swarm services.
 For more details about swarm networking in general, see the
-[Docker networking reference architecture](https://success.docker.com/Architecture/Docker_Reference_Architecture%3A_Designing_Scalable%2C_Portable_Docker_Container_Networks){: target="_blank" rel="noopener" class="_" }.
+[Docker networking reference architecture](https://docs.docker.com/network/){: target="_blank" rel="noopener" class="_" }.
```
thaJeztah, line 18: As we're now no longer leaving the docs website, I think it's ok to remove the `target=_blank`;

```suggestion
[Docker networking reference architecture](https://docs.docker.com/network/).
```

dockertopia: I'm fine with this change, but opening in another tab seems to be a behavior being used by us on other occasions other than when target link is for page outside Docker docs.

dockertopia (approved):
@arimal199 Thanks for the fix! 
@thaJeztah As agreed TOC fix following in a separate PR, thanks.

dockertopia (commented):
This image is missing. I'm removing this sentence as remedy/compromise in a separate PR (#15069).

dockertopia:
> @arimal199 Thanks for the fix! @thaJeztah As agreed TOC fix following in a separate PR, thanks.

PS - Your suggestion was added here (with some adjustments) #15069.

PR #14976, opened 2022-06-21
ci: workflow to consume dispatch events from compose repo

crazy-max:
closes https://github.com/docker/docker.github.io/pull/14457

As discussed with @glours in https://github.com/docker/docker.github.io/pull/14457#discussion_r878128542, workflow has been changed to use a remote context instead of cloning the repo.

* The "Fetch reference API" step will use the context and target from the payload to build the Docker image. In this case it will build and output to `./_data/compose_cli`. This way we avoid extra logic in docs repo so compose team can focus and maintain docs reference logic in their dockerfile.
* Then we commit changes and open a PR on docs repo

### DIFF
--- a/.github/workflows/dispatch-compose.yml
+++ b/.github/workflows/dispatch-compose.yml
@@ -1,0 +1,49 @@
+name: dispatch-compose
+
+on:
+  repository_dispatch:
+    types: [compose-docs]
+
+jobs:
+  api-reference:
+    runs-on: ubuntu-latest
+    steps:
+      -
+        name: Dump context
+        uses: crazy-max/ghaction-dump-context@v1
+      -
+        name: Checkout
+        uses: actions/checkout@v3
+      -
+        name: Prepare
+        run: |
+          rm -rf ./_data/compose-cli/*
+      -
+        name: Fetch reference API
+        uses: docker/build-push-action@v3
+        with:
+          context: ${{ github.event.client_payload.context }}
+          target: ${{ github.event.client_payload.target }}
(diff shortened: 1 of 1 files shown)


glours (approved):
🎉

thaJeztah:
Much magic. Much GHA.. I'm guessing this is fine?


One question I have / was wondering; so this action listens for events in the other repository; slightly wondering if it would make sense to do the reverse (so have an action in the compose repository that listens for these events, and that to open a PR in the docs repository? or is that a lot more complicated?

crazy-max:
>  open a PR in the docs repository?

Yes I think we can open a PR from a remote repository, I'm just afraid of some desync complications but it should work. I will try that.

crazy-max:
@thaJeztah Looks like it works: https://github.com/crazy-max/compose/runs/7045193447?check_suite_focus=true so no need to create a workflow on docs repo :tada:

thaJeztah:
Oh, nice! That is, of course, if you think that's a good idea (it felt slightly cleaner to trigger from the source repository)

crazy-max:
> Oh, nice! That is, of course, if you think that's a good idea (it felt slightly cleaner to trigger from the source repository)

Yes that's wayyy better as we don't need to take care of another workflow. Only drawback would be to manage conflicts but that's not an issue as `_data/compose` is not supposed to be manually changed.

thaJeztah:
> but that's not an issue as _data/compose is not supposed to be manually changed.

yup! that should not be an issue

PR #14977, opened 2022-06-22
Update ubuntu.md

cigdemkadakoglu:
Add Ubuntu service info

### DIFF
--- a/engine/install/ubuntu.md
+++ b/engine/install/ubuntu.md
@@ -149,9 +149,14 @@ from the repository.
     ```console
     $ sudo apt-get install docker-ce=<VERSION_STRING> docker-ce-cli=<VERSION_STRING> containerd.io docker-compose-plugin
     ```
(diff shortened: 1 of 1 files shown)


thaJeztah:
@cigdemkadakoglu could you describe if there was a problem you were trying to solve with this change? The number of steps included in the installation procedure is already quite long, so we try to minimise it to the steps that are actually needed for a regular installation.

Given that (as you mentioned in this PR) the service should be enabled and started automatically after installation, there should not be a need to verify if that's the case (the `docker run` following this already would fail if it's not working), so wondering if there was a specific situation you were trying to solve.

dockertopia:
Thanks for submitting this @cigdemkadakoglu, given the context and the information you have provided us so far we don't believe any change is needed. We're closing the ticket, but please feel free to get back to us to share what was the problem you were trying to solve and if that is resolved. Thanks.

cigdemkadakoglu:
Thank you!

PR #14981, opened 2022-06-22
extensions-sdk reorg (rebase)

thaJeztah:
this is a rebase of https://github.com/docker/docker.github.io/pull/14917, which had merge conflicts

What this does is;

- squashed all the commits from https://github.com/docker/docker.github.io/pull/14917
- added a separate commit that moves some of the files, so that the second commit can be used to review the changes that were made
- note that we should review if any of the changes made in these files should be restored; I think those were from https://github.com/docker/docker.github.io/commit/e1a8d3bbde62f144efd6e7c73dfc2d349146767f


desktop/extensions-sdk/dev-tools/api/reference/interfaces/Dialog.md
```diff
@@ -1,19 +1,11 @@
 ---
-title: Docker extension API reference
+title: Dialog
```
thaJeztah, line 2: Wondering; were these files re-generated or were these manual edits?

I also notice that there's no redirects for the old locations (slightly wondering if it's worth doing the move or if we should keep them at the old location)

thaJeztah:
Oh; before I fail to mention; if we merge this PR, we should not _squash_, and keep the first commit separate to make sure that git preserves history (otherwise it won't correlate the old file with the new one)

desktop/extensions-sdk/dev-tools/api/reference/interfaces/Dialog.md
aevesdocker: They were manual edits. I'd prefer to do the move and then create a separate PR to add the redirects of the old locations.

thaJeztah: ISTR these files were generated from source; if that's indeed the case, we should look how we can update the generate process to produce the files with the same changes, otherwise those changes will be reverted the next time we update with the generated docs; @gtardif could you confirm / have a look?

desktop/extensions-sdk/index.md
```diff
@@ -11,90 +11,52 @@ Use the resources in this section to create your own Docker Extension.
...
 Extensions are distributed through the Docker Hub.
 Development of extensions can be done locally without the need to push the extension to Docker Hub. See [Extensions distribution](extensions/DISTRIBUTION.md) for further details.
 
-## Prerequisites
+## How to use the resources in this section
 
-Before you create your own extension, you need a specific Docker Desktop build that comes with extension capabilities.
+The documentation is structured so that it matches the steps you need to take when creating your extension. There are two main parts to creating a Docker Extension; Build the foundations and then publish your extension. 
 
 From the [releases page](https://github.com/docker/extensions-sdk/releases/latest), install the Extensions CLI which is used to manage extensions later on.
```
aevesdocker, line 25:
```suggestion

```

aevesdocker (commented):
I've gone through each page that had the merge conflict and there is only one line that I wasn't expecting to be there. Everything else is ready to go and matches my original PR that has been approved by Guillaume

desktop/extensions-sdk/index.md
thaJeztah: Ah, my bad! removed 👍

desktop/extensions-sdk/dev-tools/api/reference/interfaces/Dialog.md
aevesdocker: Is this a blocker for merging this PR?

thaJeztah: I think we should at least know what will happen if those docs get updated yes, otherwise the changes will be reverted again (not sure exactly what options we have to make the generated docs match what we want it to look like here 😅)

gtardif: Yes indeed, these are generated pages (from the pinata repo). We can look at improving the generated docs form the SDK APIs, in the pinata repo, but I suggest we do this separately from this PR, so we can move forward with this one (so maybe just revert the changes from files in /api/reference in this PR, and do changes in API reference files separately)

thaJeztah:
Discussed with @aevesdocker to revert the `dev` -> `dev-tools` move, as it's not bringing too much advantages, but will require adding a lot of redirects; temporarily moved this back to draft, but will have a look at that

thaJeztah:
I reverted the `dev` -> `dev-tools` move, and moved the API reference changes to https://github.com/docker/docker.github.io/pull/14988 and opened an internal PR to update the docs generator (https://github.com/docker/pinata/pull/18357)

aevesdocker:
@thaJeztah I think the toc needs to be updated to reflect the dev-tools to dev move as I'm getting broken links on the preview

thaJeztah:
@aevesdocker thanks! Yes, I just noticed that, and pushed

Also wondering if we should change `api/overview.md` to `api/index.md`, but we can look at that in a follow-up

desktop/extensions-sdk/dev/overview.md
```diff
@@ -1,62 +0,0 @@
----
-title: Overview
```
thaJeztah, line 2: I see this page was removed, but there's no replacement or redirect; was that intentional?

aevesdocker: Yes, the content from this page was distributed to various other pages

thaJeztah:
arf; more to fix; looks like there's a redirect loop now

desktop/extensions-sdk/dev/overview.md
thaJeztah: I added a redirect to the https://docs.docker.com/desktop/extensions-sdk/ landing-page; users should likely be able to find where to go from there if they happened to have that old URL 👍

PR #14985, opened 2022-06-22
Fix Dead Link in the Compose specification documentation

BinaryWizard904:
Fix dead link to a Permissions Calculator in the Compose specification documentation.

### Proposed changes

Replace the dead link to the http://permissions-calculator.org with one to a [working copy](https://web.archive.org/web/20220310140126/http://permissions-calculator.org/) of the page hosted on the Internet Archive.

### DIFF
--- a/compose/compose-file/index.md
+++ b/compose/compose-file/index.md
@@ -1657,7 +1657,7 @@ the service's containers.
   service's task containers. Defaults to `source` if not specified.
 - `uid` and `gid`: The numeric UID or GID that owns the file within
   `/run/secrets/` in the service's task containers. Default value is USER running container.
-- `mode`: The [permissions](http://permissions-calculator.org/) for the file to be mounted in `/run/secrets/`
+- `mode`: The [permissions](https://web.archive.org/web/20220310140126/http://permissions-calculator.org/) for the file to be mounted in `/run/secrets/`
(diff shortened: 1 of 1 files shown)


BinaryWizard904:
@dockertopia Well for one that calculator only generates a 3 digit octal instead of the 4 digit octal the Compose specification documentation indicates we should use.  If there is an online calculator that produces the 4 digit

dockertopia:
> Fix dead link to a Permissions Calculator in the Compose specification documentation.
> 
> ### Proposed changes
> Replace the dead link to the http://permissions-calculator.org with one to a [working copy](https://web.archive.org/web/20220310140126/http://permissions-calculator.org/) of the page hosted on the Internet Archive.

True. Maybe this one could be an option https://www.killall.net/webmaster/chmod.php ? 
In the meantime we could use the archived URL version.

dockertopia (approved):

BinaryWizard904:
> True. Maybe this one could be an option https://www.killall.net/webmaster/chmod.php ? In the meantime we could use the archived URL version.

The archived version has the advantage of being able to copy-paste the results.  You can't do that with the killall version (or any other version I've found via searching for that matter).  Interface is also needlessly wide.  I know that sounds picky; I'm just not sure it would be an improvement over the archive version that's easier to use.

PR #14994, opened 2022-06-23
added what file type the github actions is

grant350:
I added this as it might be confusing for some people who dont write in yaml every day to know what file it is. it is best to explicitly state what file type it is


usha-mandya (approved):
LGTM. Thank you @grant350

PR #14999, opened 2022-06-24
Update debian.md

luozhenyu:
### Proposed changes
1. `docker-ce-cli containerd.io` is required by `docker-ce`. there is no need to declare to be installed.
2. `docker-compose-plugin` is not available.

### DIFF
--- a/engine/install/debian.md
+++ b/engine/install/debian.md
@@ -115,7 +115,7 @@ This procedure works for Debian on `x86_64` / `amd64`, `armhf`, `arm64`, and Ras
 
     ```console
     $ sudo apt-get update
-    $ sudo apt-get install docker-ce docker-ce-cli containerd.io docker-compose-plugin
+    $ sudo apt-get install docker-ce
     ```
 
     > Receiving a GPG error when running `apt-get update`?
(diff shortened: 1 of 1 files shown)


engine/install/debian.md
```diff
@@ -115,7 +115,7 @@ This procedure works for Debian on `x86_64` / `amd64`, `armhf`, `arm64`, and Ras
 
     ```console
     $ sudo apt-get update
-    $ sudo apt-get install docker-ce docker-ce-cli containerd.io docker-compose-plugin
+    $ sudo apt-get install docker-ce
```
thaJeztah, line 118: We explicitly include all packages here to make sure they're all updated; omitting the extra packages works on a fresh install, but users also use the same commands to update their installation.

Unfortunately, `apt-get install` does not automatically assume you want to update all dependencies to the latest version as well, which has led to users running outdated versions of (some of the) components. For that reason, we veer on the safe side, and include all packages in the instruction (also to make it more visible that there's multiple components that make up a full installation)

dockertopia: @luozhenyu regarding the error you got about the compose plugin package, have you set the Docker repository before executing the install instruction? 
You can find that info here: https://docs.docker.com/engine/install/debian/#install-using-the-repository.

Please let us know if both comments have helped you and if you were able to complete the installation.  Thanks.

dockertopia (commented):
@luozhenyu regarding the error you got about the compose plugin package, have you set the Docker repository before executing the install instruction?
You can find that info here: https://docs.docker.com/engine/install/debian/#install-using-the-repository.

Please let us know if both comments have helped you and if you were able to complete the installation. Thanks.

PR #15002, opened 2022-06-24
updated SCIM images

jerae-duffin:
Updated SCIM images and attributes step


docker-hub/scim.md
```diff
@@ -40,6 +40,8 @@ Before making SCIM configuration changes in your IdP, navigate to [Docker Hub](h
 
 ![scim-app-provisioning](images/scim-app-provisioning.png){:width="700px"}
 
+6. Remove the unwanted attributes from your mappings and fields from the Profile editor.
```
craig-osterhout, line 43: In the preview, this is displaying as step 1 instead of 6.

ivan-californias, line 43: Profile editor is used to modify custom user attributes, we don't need them to do that, they just need to remove them from the Attributes mappings
```suggestion
6. Remove all fields that are not supported from your Attributes Mappings.
```

And add this image a
Untitled
s reference of the attributes we do support:

jerae-duffin: Thank you, updated!

jerae-duffin: Thank you, updated!

craig-osterhout (approved):
LGTM

ivan-californias (approved):
👍  Thank you

PR #15019, opened 2022-06-27
Fix spelling

pokir:
### Proposed changes

Fixed word choice on the docker-compose file documentation.
"prevent... _for_ being portable" --> "prevent... _from_ being portable"


dockertopia (approved):
@pokir thanks for this fix!

PR #15023, opened 2022-06-28
Update wordpress.md

bcdady:
Update example to match https://github.com/docker/awesome-compose/blob/master/wordpress-mysql/compose.yaml, which works.

### Proposed changes

Update compose.yaml content to match https://github.com/docker/awesome-compose/blob/master/wordpress-mysql/compose.yaml, which works when run via `docker compose up -d`. The prior example yaml would result in an error message of "Top-level object must be a mapping".


usha-mandya (approved):
LGTM. Thank you @bcdady

PR #15026, opened 2022-06-28
Updating Compose release notes to 2.6.1

dockertopia:
Adding 2.6.1 Compose Release Notes.

### DIFF
--- a/compose/release-notes.md
+++ b/compose/release-notes.md
@@ -18,13 +18,13 @@ redirect_from:
 - Added support for setting secrets from environment variable. Fixes [#251](https://github.com/compose-spec/compose-spec/issues/251){:target="_blank" rel="noopener" class="_"}.
 - Added links to container create request. Fixes [#9513](https://github.com/docker/compose/issues/9513){:target="_blank" rel="noopener" class="_"}.
 - Fixed `compose run` to start only direct dependencies. Fixes [#9459](https://github.com/docker/compose/issues/9459){:target="_blank" rel="noopener" class="_"}.
-- Fixed `docker compose up` service not found errors when using `--no-deps` option. Fixes [#9427](https://github.com/docker/compose/issues/9427){:target="_blank" rel="noopener" class="_"}.
+- Fixed `docker compose up` 'service not found' errors when using `--no-deps` option. Fixes [#9427](https://github.com/docker/compose/issues/9427){:target="_blank" rel="noopener" class="_"}.
 - Fixed `compose down` to respect `COMPOSE_REMOVE_ORPHANS` environment variable. Fixes [#9562](https://github.com/docker/compose/issues/9562){:target="_blank" rel="noopener" class="_"}.
 - Fixed project-level bind mount volumes. Fixes [docker/for-mac#6317](https://github.com/docker/for-mac/issues/6317){:target="_blank" rel="noopener" class="_"}.
 - Fixed parsing of properties `deploy.limits.cpus` and `deploy.limits.pids` to respect floating-point values. Fixes [#9542](https://github.com/docker/compose/issues/9542){:target="_blank" rel="noopener" class="_"} and [#9501](https://github.com/docker/compose/issues/9501){:target="_blank" rel="noopener" class="_"}. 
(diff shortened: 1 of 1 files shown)


compose/release-notes.md
```diff
@@ -8,6 +8,51 @@ redirect_from:
...
+- Fixed `docker compose up` 'service not found' errors when using `--no-deps` option. Fixes [#9427](https://github.com/docker/compose/issues/9427){:target="_blank" rel="noopener" class="_"}.
+- Fixed `compose down` to respect `COMPOSE_REMOVE_ORPHANS` environment variable. Fixes [#9562](https://github.com/docker/compose/issues/9562){:target="_blank" rel="noopener" class="_"}.
+- Fixed project-level bind mount volumes. Fixes [docker/for-mac#6317](https://github.com/docker/for-mac/issues/6317){:target="_blank" rel="noopener" class="_"}.
+- Fixed parsing of properties `deploy.limits.cpus` and `deploy.limits.pids` to respect floating-point values. Fixes [#9542](https://github.com/docker/compose/issues/9542){:target="_blank" rel="noopener" class="_"} and [#9501](https://github.com/docker/compose/issues/9501){:target="_blank" rel="noopener" class="_"}. 
+- Fixed `compose ps` output to list all exposed ports. Fixes [#9257](https://github.com/docker/compose/issues/9527){:target="_blank" rel="noopener" class="_"}.
+- Fixed spelling mistakes in `compose ps` code.
```
milas, line 26: I don't think we need to include this, there's no user-facing impact/visibility.

compose/release-notes.md
```diff
@@ -8,6 +8,51 @@ redirect_from:
...
 
 
+## 2.6.1
+
+(2022-06-23)
+
+
+### Bug Fixes
+
+- Added support for setting secrets from environment variable. Fixes [#251](https://github.com/compose-spec/compose-spec/issues/251){:target="_blank" rel="noopener" class="_"}.
```
milas, line 18:
```suggestion
- Added support for setting secrets from environment variable. See [compose-spec/compose-spec#251](https://github.com/compose-spec/compose-spec/issues/251){:target="_blank" rel="noopener" class="_"}.
```
This is a feature rather than a fix, so it might actually make more sense for it to be separate from this section. But for consistency, I think specifying the repo (similar to `docker/for-mac#xxxx` below) is also nice since it's linking to something outside the main Compose repo

compose/release-notes.md
```diff
@@ -8,6 +8,51 @@ redirect_from:
...
+- Added links to container create request. Fixes [#9513](https://github.com/docker/compose/issues/9513){:target="_blank" rel="noopener" class="_"}.
+- Fixed `compose run` to start only direct dependencies. Fixes [#9459](https://github.com/docker/compose/issues/9459){:target="_blank" rel="noopener" class="_"}.
+- Fixed `docker compose up` 'service not found' errors when using `--no-deps` option. Fixes [#9427](https://github.com/docker/compose/issues/9427){:target="_blank" rel="noopener" class="_"}.
```
milas, line 21:
```suggestion
- Fixed `compose up` 'service not found' errors when using `--no-deps` option. Fixes [#9427](https://github.com/docker/compose/issues/9427){:target="_blank" rel="noopener" class="_"}.
```
For consistency: the others omit `docker` before the plugin

milas (approved):
lgtm - a few small suggestions but also fine with it as-is

compose/release-notes.md
dockertopia: So far I was including everything coming from the notes in the repo, as I had been told that was already a selection thinking of what was of interest to users on one hand. On another hand if they fix issues reported by users maybe it's a good reason to feature on the notes?

milas (approved):
(testing permissions)

glours (approved):
Just want to be sure we'll move the support of secret from environment variable to the #feature section otherwise agree with @milas NITs

compose/release-notes.md
dockertopia: OK, one small question would "Added links to container create request" not be an enhancement at the same level?

usha-mandya (approved):

PR #15027, opened 2022-06-29
"docker compose" instead "docker-compose"

angelcervera:
### Proposed changes

Installation is about v2 so docker compose is a plugin, but this section still uses the old `docker-compose` syntaxis.


dockertopia (approved):
Thanks for this @angelcervera!

PR #15044, opened 2022-07-03
docker scan: using an HTTP proxy

djs55:
### Proposed changes

Explain how to make `docker scan` use an HTTP proxy.

### DIFF
--- a/engine/scan/index.md
+++ b/engine/scan/index.md
@@ -121,7 +121,7 @@ Note that we do not currently have vulnerability data for your image.
 If your computer needs to use an HTTP proxy, then set the `HTTPS_PROXY` environment variable when using scan. For example:
 
 ```console
-$ export HTTPS_PROXY=http://username:password@proxy:port
+$ export HTTPS_PROXY=http://username:password@proxyAddress:port
 $ docker scan hello-world
 ```
 
--- a/engine/scan/index.md
+++ b/engine/scan/index.md
@@ -116,6 +116,24 @@ Licenses:          enabled
 Note that we do not currently have vulnerability data for your image.
 ```
 
+## Using an HTTP proxy
+
+If your computer needs to use an HTTP proxy, then set the `HTTPS_PROXY` environment variable when using scan. For example:
+
+```console
+$ export HTTPS_PROXY=http://username:password@proxy:port
+$ docker scan hello-world
+```
+
+or for powershell:
+
+```console
+> $env:HTTPS_PROXY = "http://username:password@proxy:port"
+> docker scan hello-world
+```
+
+Note that `docker scan` does not use the same proxy settings as Docker Desktop or the settings specified in `.docker/config.json` or the settings used by the Docker engine.
+
 ### Get a detailed scan report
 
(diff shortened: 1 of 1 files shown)


engine/scan/index.md
```diff
@@ -116,6 +116,24 @@ Licenses:          enabled
 Note that we do not currently have vulnerability data for your image.
 ```
 
+## Using an HTTP proxy
+
+If your computer needs to use an HTTP proxy, then set the `HTTPS_PROXY` environment variable when using scan. For example:
+
+```console
+$ export HTTPS_PROXY=http://username:password@proxy:port
```
vvoland, line 124: How about `address` or `proxyAddress` instead of `proxy`? Both here and for powershell.
So:
```suggestion
$ export HTTPS_PROXY=http://username:password@proxyAddress:port
```

engine/scan/index.md
```diff
@@ -116,6 +116,24 @@ Licenses:          enabled
...
+$ docker scan hello-world
+```
+
+or for powershell:
```
vvoland, line 128:
```suggestion
or for PowerShell:
```

engine/scan/index.md
```diff
@@ -116,6 +116,24 @@ Licenses:          enabled
 Note that we do not currently have vulnerability data for your image.
 ```
 
+## Using an HTTP proxy
+
+If your computer needs to use an HTTP proxy, then set the `HTTPS_PROXY` environment variable when using scan. For example:
```
vvoland, line 121: computer => environment?

vvoland (commented):
Maybe it would be also worth mentioning HTTP_PROXY or NO_PROXY?

thaJeztah:
I'm slightly on the fence on this, as `docker scan` is just _one_ subcommand of the Docker CLI (considering the fact that it happens to be a CLI plug-in an implementation detail). Setting up the CLI to use a proxy is something shared by every (sub)command of the CLI (e.g. `docker manifest` subcommands also make networking connections).

It's ok to mention that `docker scan` makes networking connections, and that this may require the proxy env-vars to be set up, but perhaps we should have a (reusable) snippet for that, which points to documentation for that. I know there's https://docs.docker.com/engine/reference/commandline/cli/#environment-variables, which mentions the proxy variables (but currently doesn't have a dedicated heading for this, so perhaps that's something we should add, and a bit more wording 🤔 

Alternatively (perhaps better) we could have a page describing all permutations of "proxy" to set up (cli, daemon, desktop, containers, build) and make the https://docs.docker.com/engine/reference/commandline/cli/#environment-variables section contain a link to that "for more details".

usha-mandya (approved):
LGTM

dockertopia:
> > Alternatively (perhaps better) we could have a page describing all permutations of "proxy" to set up (cli, daemon, desktop, containers, build) and make the https://docs.docker.com/engine/reference/commandline/cli/#environment-variables section contain a link to that "for more details".
> 
> I agree it would be ideal to have a single page which shows all the combinations, easily found by googling for "docker proxy". One of the problems at the moment is that google zooms in on one of the pages, and then the user is confused.

I guess this will be solved when we have the new page/section that will a better hit for the "docker proxy" search.

dockertopia (approved):
Thanks for the commit @vvoland.

djs55:
Thanks!

PR #15045, opened 2022-07-03
desktop proxies: clarify situation on Mac and Windows

djs55:
### Proposed changes

1. there are multiple different places proxies are configured (unfortunately). Clarify that multiple different configuration steps are needed depending on what you're trying to achieve.
2. there are behaviour differences between Mac and Windows (currently Mac performs transparent proxying but Windows does not)
3. rather than manually setting environment variables for containers we should use the CLI's new config.json support.

### Related issues (optional)

#15044

PR #15046, opened 2022-07-03
network/proxy: add a note pointing at docker engine and Docker Desktop

djs55:
### Proposed changes

Add references from network/proxy.md to the other places in the docs where we discuss HTTP proxy configuration.

The title is "Configure Docker to use a proxy server" which is generic and implies this is the page which describes everything. However the scope of the instructions is limited to containers:

> If your container needs ...

The limited scope is easy to miss.

There are other HTTP proxy cases, for example

- for Docker Desktop to log into Docker
- for pulling and pushing containers
- for docker scan

Add a note at the top of this page to clarify the scope and link to instructions for other cases.

### Related issues (optional)

#15044 #15045

### DIFF
--- a/network/proxy.md
+++ b/network/proxy.md
@@ -7,7 +7,8 @@ keywords: network, networking, proxy, client
 > **Note**
 >
 > This page describes how to configure the Docker CLI to configure proxies via environment variables in containers.
-> For information on configuring Docker Desktop to use HTTP/HTTPS proxies, see [proxies on Mac](../desktop/mac/index.md#proxies) and [proxies on Windows](../desktop/windows/index.md#proxies).
+> For information on configuring Docker Desktop to use HTTP/HTTPS proxies, see [proxies on Mac](../desktop/mac/index.md#proxies), [proxies on Windows](../desktop/windows/index.md#proxies), and [proxies on Linux](../desktop/linux/index.md#proxies).
+
 > If you are not running Docker Desktop, and have installed the Docker Engine in
 > other ways, refer to the "HTTP/HTTPS proxy" section in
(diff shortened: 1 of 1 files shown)


network/proxy.md
```diff
@@ -4,6 +4,12 @@ description: How to configure the Docker client to use a proxy server
 keywords: network, networking, proxy, client
 ---
 
+> **Note**
+>
+> This page describes how to configure the Docker CLI to configure proxies via environment variables in containers.
+> For information on configuring Docker Desktop to use HTTP/HTTPS proxies, see [proxies on Mac](../desktop/mac/index.md#proxies) and [proxies on Windows](../desktop/windows/index.md#proxies).
```
thaJeztah, line 10: Also need to add the DD4L link; https://docs.docker.com/desktop/linux/#proxies

That said;

- Perhaps the "proxies" section should be in a general page (looks like it's effectively the same for all variants)
- Currently the linked to section only describes "To configure HTTP proxies, switch on the Manual proxy configuration setting.", then continues to show a single example with `docker run -e ...`, and links back to this page for details (so we'd be sending users in a loop 😂); perhaps the `docker run -e` example should be removed there, and this page improved a bit.

usha-mandya: Thanks @thaJeztah. Added your suggestion. @aevesdocker is working on consolidating the Desktop docs. Adding a generic proxy page will be part of this effort. We can also do a follow-up PR for further improvements to the proxy page and to update based on your suggestion on adding a single example with `docker run -e`.

usha-mandya (approved):
LGTM

Issue #14124, opened 2022-01-15
CAP_PERFMON works in runc but not docker


stevenjswanson:
I've been trying to get CAP_PERFMON to work without success.

I've tried all the channel settings ("stable", "nightly", etc.) for a clean install from the latest get-docker.com script and consistently get:

```
# docker run --cap-add CAP_PERFMON hello-world
docker: Error response from daemon: invalid CapAdd: unknown capability: "CAP_PERFMON".
```
Poking around at `runc` suggests that my `runc` support `CAP_PERFMON`.  Here's the relevant section of my `config.json`, 

```                "capabilities": {
                        "bounding": [
                                "CAP_AUDIT_WRITE",
                                "CAP_KILL",
                                "CAP_NET_BIND_SERVICE",
"CAP_PERFMON"
                        ],
                        "effective": [
                                "CAP_AUDIT_WRITE",
                                "CAP_KILL",
                                "CAP_NET_BIND_SERVICE",
"CAP_PERFMON"
                        ],
                        "inheritable": [
                                "CAP_AUDIT_WRITE",
                                "CAP_KILL",
                                "CAP_NET_BIND_SERVICE",
"CAP_PERFMON"
                        ],
                        "permitted": [
                                "CAP_AUDIT_WRITE",
                                "CAP_KILL",
                                "CAP_NET_BIND_SERVICE",
"CAP_PERFMON"

                        ],
                        "ambient": [
                                "CAP_AUDIT_WRITE",
                                "CAP_KILL",
                                "CAP_NET_BIND_SERVICE",
"CAP_PERFMON"
                        ]
                },
```

and `runc run container1` completes successfully.

Here's what seems like the relevant version strings (I'm in Ubuntu 20.04 HWE):

```
# uname -a
Linux da-c3-small-x86-01 5.11.0-44-generic #48~20.04.2-Ubuntu SMP Tue Dec 14 15:36:44 UTC 2021 x86_64 x86_64 x86_64 GNU/Linux
```

```
# docker --version
Docker version 20.10.12, build e91ed57
```

```
# runc --version
runc version 1.0.2
commit: v1.0.2-0-g52b36a2
spec: 1.0.2-dev
go: go1.16.10
libseccomp: 2.5.1
```

vext01:
I have the same on Debian stable, running the Debian docker package (`20.10.5+dfsg1`).

```
$ docker run -ti --cap-add=CAP_PERFMON test
[...]

Issue #13731, opened 2021-10-25
Support for CAP_PERFMON, CAP_BPF, and CAP_CHECKPOINT_RESTORE


tallossos:
[engine/release-notes/index.md](https://docs.docker.com/engine/release-notes/)

Hello,
Few month back, the commit for disabling CAP_PERFMON, CAP_BPF, and CAP_CHECKPOINT_RESTORE got reverted (https://github.com/moby/moby/pull/42011), so now we should be able to create containers with these capabilities.
When this feature got disabled in the first place, it didn't get into the changelog (https://github.com/moby/moby/issues/42601#issuecomment-875736356), so one might think that it should work when it actually can't.
Please mention that using these capabilities should work in versions after this merge. 
Thanks.

jakubrak:
Still getting the error:
`
ubuntu@ubuntu21:~$ uname -r
5.13.0-20-generic

$ head -1 /etc/os-release
PRETTY_NAME="Ubuntu 21.10"

$ docker --version
Docker version 20.10.10, build b485636

$ docker run --cap-add PERFMON hello-world
docker: Error response from daemon: invalid CapAdd: unknown capability: "CAP_PERFMON".
See 'docker run --help'.`

stevenjswanson:
I've been trying to CAP_PERFMON to work without success.  Is there some branch or tag that has "this merge" (as mentioned above) that I need to be using?

I've tried all the channel settings ("stable", "nightly", etc.) for a clean install from the latest get-docker.com script and consistently get:

```
# docker run --cap-add CAP_PERFMON hello-world
docker: Error response from daemon: invalid CapAdd: unknown capability: "CAP_PERFMON".
```
Poking around at `runc` suggests that my `runc` support `CAP_PERFMON`.  Here's the relevant section of my `config.json`, 

```                "capabilities": {
                        "bounding": [
                                "CAP_AUDIT_WRITE",
                                "CAP_KILL",
                                "CAP_NET_BIND_SERVICE",
"CAP_PERFMON"
                        ],
                        "effective": [
                                "CAP_AUDIT_WRITE",
                                "CAP_KILL",
                                "CAP_NET_BIND_SERVICE",
"CAP_PERFMON"
                        ],
                        "inheritable": [
                                "CAP_AUDIT_WRITE",
                                "CAP_KILL",
                                "CAP_NET_BIND_SERVICE",
"CAP_PERFMON"
                        ],
                        "permitted": [
[...]

PR #15050, opened 2022-07-04
engine: remove CAP_PERFMON, CAP_BPF, CAP_CHECKPOINT_RESTORE from relnotes

thaJeztah:
- closes https://github.com/docker/docker.github.io/issues/13731
- closes https://github.com/docker/docker.github.io/issues/14124

These capabilities were reverted in https://github.com/moby/moby/commit/a38b96b8cdc4d345a050b417c4c492b75329e5a6 (https://github.com/moby/moby/pull/41563),
because the version of containerd and runc that shipped together with
docker 20.10.0 did not support them.

Current versions of containerd and runc support these, so possibly we can
include them again in a patch release, but otherwise support will come in
the next (22.06) release of docker engine.


dockertopia (commented):
My interpretation was you are repackaging the release assets for 20.10.0 reflecting mentioned reversal. If this is the case, is adding this line amidst the regular RNs of that release the best option? 
What do you think of adding this line as a note (with a date, that is going to be posterior to the release date) at the top of the page?

thaJeztah:
No, this is deleting an entry from the release notes, because the new feature was reverted/disabled before the 20.10.0 release, so didn't make it into the release, and should not have been included in the changelog

dockertopia (approved):
LGTM

PR #15079, opened 2022-07-07
Multi-arch: Change A1 to Graviton instances

otterley:
Change focus from legacy A1 EC2 instances to Graviton instances.


usha-mandya (approved):
LGTM. Thank you @otterley

PR #15082, opened 2022-07-07
Don't use quotes on HTTP_PROXY environment URLs

rfay:
### Proposed changes

Currently, I'm not sure how long, the quotes cause havoc with systemd and the config doesn't work. Tested with Ubuntu 22.04 where I ran into this.  The hint was from https://forums.docker.com/t/proxy-settings-do-not-work-for-either-etc-default-docker-or-systemd-drop-ins/68891/20

### DIFF
--- a/config/daemon/systemd.md
+++ b/config/daemon/systemd.md
@@ -92,7 +92,7 @@ you need to add this configuration in the Docker systemd service file.
 
     ```systemd
     [Service]
-    Environment="HTTP_PROXY=http://proxy.example.com:80"
+    Environment=HTTP_PROXY=http://proxy.example.com:80
     ```
 
     If you are behind an HTTPS proxy server, set the `HTTPS_PROXY` environment
@@ -100,7 +100,7 @@ you need to add this configuration in the Docker systemd service file.
 
     ```systemd
     [Service]
-    Environment="HTTPS_PROXY=https://proxy.example.com:443"
+    Environment=HTTPS_PROXY=https://proxy.example.com:443
     ```
     
     Multiple environment variables can be set; to set both a non-HTTPS and
@@ -108,8 +108,8 @@ you need to add this configuration in the Docker systemd service file.
 
     ```systemd
     [Service]
-    Environment="HTTP_PROXY=http://proxy.example.com:80"
-    Environment="HTTPS_PROXY=https://proxy.example.com:443"
+    Environment=HTTP_PROXY=http://proxy.example.com:80
+    Environment=HTTPS_PROXY=https://proxy.example.com:443
     ```
      
 3.  If you have internal Docker registries that you need to contact without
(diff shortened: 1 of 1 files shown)


config/daemon/systemd.md
```diff
@@ -92,24 +92,24 @@ you need to add this configuration in the Docker systemd service file.
 
     ```systemd
     [Service]
-    Environment="HTTP_PROXY=http://proxy.example.com:80"
+    Environment=HTTP_PROXY=http://proxy.example.com:80
```
thaJeztah, line 95: Hmmm, that's odd that it won't work; systemd documentation describes this as the supported syntax; https://www.freedesktop.org/software/systemd/man/systemd.exec.html#Environment

> Sets environment variables for executed processes. Each line is unquoted using the rules described in "Quoting" section in [systemd.syntax(7)](https://www.freedesktop.org/software/systemd/man/systemd.syntax.html#) and becomes a list of variable assignments. If you need to assign a value containing spaces or the equals sign to a variable, put quotes around the whole assignment. Variable expansion is not performed inside the strings and the "$" character has no special meaning. Specifier expansion is performed, see the "Specifiers" section in [systemd.unit(5)](https://www.freedesktop.org/software/systemd/man/systemd.unit.html#).

rfay: I don't see why it wouldn't work, but it sure worked fine after I removed those. And not at all before that. Ubuntu 22.04.

thaJeztah: The quotes may be important for situations where special characters are included in the proxy, e.g. if (shudder) the proxy requires authentication to be passed as part of the URL (`https://username:password@myproxy.com`), in which case the password may contain, e.g. `$`, wich would (attempted to) be substituted by systemd otherwise.

I tried the example with the quotes as they were used in the docs;

```bash
systemctl edit docker.service
```

Edit the file to add;

```
[Service]
Environment="HTTP_PROXY=http://proxy.example.com:80"
Environment="HTTPS_PROXY=https://proxy.example.com:443"
```

Check the `Environment` options in the service;

```bash
systemctl show docker.service | grep Environment
Environment=HTTP_PROXY=http://proxy.example.com:80 HTTPS_PROXY=https://proxy.example.com:443
```

Restart the docker service and check proxy config;

```bash
systemctl restart docker.service

docker info | grep Proxy
 HTTP Proxy: http://proxy.example.com:80
 HTTPS Proxy: https://proxy.example.com:443
```


_However_, it looks like the linked forum post used quoting incorrectly; systemd requires the quotes to be around the _whole_ `variable=value`, so `"HTTP_PROXY=https://example.com"`, not around the _value_ (`HTTP_PROXY="https://example.com"`).

Having those quotes in the wrong place _could_ mean that the proxy isn't working correctly.

thaJeztah: Oh! I see the example further down;

```
Environment="HTTP_PROXY=http://user:complex%23pass@proxy.example.com:8080/"
```

And indeed; that one is fully ignored by systemd;

```bash
systemctl show docker.service | grep Environment
Environment=HTTPS_PROXY=https://proxy.example.com:443
```

That looks to be a rather nasty bug in systemd?

thaJeztah: Oh! 🤦 I see the example uses `%` (`%23`), which still must be escaped because it's otherwise considered to be a [systemd specifier](https://www.freedesktop.org/software/systemd/man/systemd.unit.html#Specifiers); 

This fails

```unit file (systemd)
[Service]
Environment="HTTP_PROXY=http://user:pass%23word@proxy.example.com:8080"
```

This works (`%%` for a literal percentage)

```unit file (systemd)
[Service]
Environment="HTTP_PROXY=http://user:pass%%23word@proxy.example.com:8080"
```

PR #15085, opened 2022-07-08
Fix user flag in privileged-helper.md

gabriellavengeo:
### Proposed changes

Fix user flag - change form `-user` to `--user`


mat007:
@ebriney has the change been done in the code, or not yet?

PR #15086, opened 2022-07-08
move DD install pages

aevesdocker:


thaJeztah:
@aevesdocker @usha-mandya if we are still trying out new structure; perhaps we should not (yet) move / rename the actual pages, but only update the TOC to prevent too much churn (and redirects) in the URLs

PR #15090, opened 2022-07-09
Fix use of bullet list

peterhauke:
### Otherwise it would read like so:

To summarize, a container: Containers are isolated from each other and run their own software, binaries, and configurations.

### Alternatively, to keep it as a bullet point:

To summarize, a container: 
...
is isolated from other containers and runs its own software, binaries, and configurations.


craig-osterhout:
@peterhauke, thank you for creating this. Is it possible for you to update it to your alternative, keeping it in the bulleted list?

craig-osterhout (approved):
Thanks @peterhauke. LGTM

PR #15109, opened 2022-07-13
ci: update redirects with replace directive in s3 bucket

crazy-max:
Handles redirects generated by https://docs.docker.com/redirects.json in our s3 bucket by setting redirect requests for the objects. As our bucket is configured as a website, objects with redirects should have the proper 301 redirect status.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -53,6 +53,21 @@ jobs:
           AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
           AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
       -
+        name: Update redirects
+        run: |
+          for redirect in $(jq -c '. | to_entries | .[]' ./_site/redirects.json); do
+            from=$(echo "$redirect" | jq -r '.key')
+            to=$(echo "$redirect" | jq -r '.value')
+            from="${from%/}/index.html"
+            (
+              set -x
+              aws --region us-east-1 s3 cp "s3://${{ env.DOCS_S3_HOST }}${from}" "s3://${{ env.DOCS_S3_HOST }}${from}" --website-redirect "${to}" --metadata-directive REPLACE
+            )
+          done
+        env:
+          AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
+          AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+      -
         name: Update S3 website config
         uses: ./.github/actions/update-website-config
         with:


crazy-max:
@thaJeztah Is it https://github.com/docker/docker.github.io/blob/master/_data/redirects.csv still used?

thaJeztah:
Goooooood question. TBH I forgot that was there; looks like that was part of https://github.com/docker/docker.github.io/commit/e3cb7267064116579347db1864694b06cdf38e13, but I'd have to dig as well to see how (and _IF_) it's still used.

thaJeztah:
A quick search tells me it's no longer used 🤔 but please double check if you find some reference to it

thaJeztah:
Oh! Perhaps I left it as a reminder to "include these as `redirect_from`"

.github/workflows/publish.yml
```diff
@@ -52,6 +52,31 @@ jobs:
...
+            part1=$1
+            part2=$2
+            if [ "${part1: -1}" == "/" ]; then
+              part1=${part1}${part2}
+            else
+              part1=${part1}/${part2}
+            fi
+            echo $part1
+          }
+          for redirect in $(cat ./_site/redirects.json | jq -c '. | to_entries | .[]'); do
```
thaJeztah, line 68: I think the `cat` may not be needed; `jq` accepts filename(s) as positional argument; so something like this should work;

```bash
jq -c '. | to_entries | .[]' ./_site/redirects.jso
```

.github/workflows/publish.yml
```diff
@@ -52,6 +52,31 @@ jobs:
...
+            from=$(echo $redirect | jq '.key')
+            to=$(echo $redirect | jq '.value')
+            from=$(path_join $from "index.html")
+            (
+              set -x
+              aws --region us-east-1 s3 cp s3://${{ env.DOCS_S3_HOST }}${from} s3://${{ env.DOCS_S3_HOST }}${from} --website-redirect ${to} --metadata-directive REPLACE
```
thaJeztah, line 74:
```suggestion
              aws --region us-east-1 s3 cp "s3://${{ env.DOCS_S3_HOST }}${from}" "s3://${{ env.DOCS_S3_HOST }}${from}" --website-redirect "${to}" --metadata-directive REPLACE
```

.github/workflows/publish.yml
```diff
@@ -52,6 +52,31 @@ jobs:
         env:
           AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
           AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+      -
+        name: Update redirects
+        run: |
+          path_join() {
+            part1=$1
+            part2=$2
+            if [ "${part1: -1}" == "/" ]; then
```
thaJeztah, line 61: Shellcheck warns that `"${part1: -1}:` is non-POSIX; do we need to explicitly set `bash` as shell?

thaJeztah: Actually; I think `${part1%/}` would also work; basically: "normalise" by stripping the trailing `/` (if any), then append the `index.html`;

```bash
export foo="/hello/"
export bar="/hello"
echo "${foo%/}/index.html"
# /hello/index.html

echo "${bar%/}/index.html"
# /hello/index.html
```

.github/workflows/publish.yml
```diff
@@ -52,6 +52,31 @@ jobs:
...
+              part1=${part1}${part2}
+            else
+              part1=${part1}/${part2}
+            fi
+            echo $part1
+          }
+          for redirect in $(cat ./_site/redirects.json | jq -c '. | to_entries | .[]'); do
+            from=$(echo $redirect | jq '.key')
+            to=$(echo $redirect | jq '.value')
```
thaJeztah, line 70:
```suggestion
            from=$(echo "$redirect" | jq '.key')
            to=$(echo "$redirect" | jq '.value')
```

.github/workflows/publish.yml
```diff
@@ -52,6 +52,31 @@ jobs:
...
+            from=$(path_join $from "index.html")
```
thaJeztah, line 71: See above; this could be;

```suggestion
            from="${from%/}/index.html"
```

thaJeztah (changes requested):

thaJeztah (approved):
LGTM

I guess the only way to try this one is to merge and check [docs-stage.docker.com](https://docs-stage.docker.com) ?

crazy-max:
> LGTM
> 
> I guess the only way to try this one is to merge and check [docs-stage.docker.com](https://docs-stage.docker.com) ?

Yes indeed

PR #15113, opened 2022-07-13
Removes is_private from UpdateRepo docs

zappy-shu:
### Proposed changes

The UpdateRepository endpoint in the Hub API does not have the ability to set a repository's privacy. This removes mention of the field so as not to cause confusion


usha-mandya (approved):
LGTM. Thank you @zappy-shu

PR #15121, opened 2022-07-13
ENGDOCS-813 remove link to topic that no longer exists

craig-osterhout:
### Proposed changes

The information was removed with the support form migration. The current link now redirects to the Hub docs.

### Related issues (optional)

ENGDOCS-813


jerae-duffin (approved):
LGTM!

PR #14908, opened 2022-06-09
Build overview

dockertopia:
### Updates since original commits

- Review and proofreading

- Rewriting of the Overview page and new Hello Build page
Thanks @crazy-max and @jedevc for writing this with me, I think this version is a lot better and we incorporated a couple of great ideas.

  * Removed builders page only from TOC ( PR follow-up will ensue) - no need to review this page.
  * Refreshed Overview page's intro ( features highlights will follow in a separate PR)
  * Added new Hello Build page.

### Related PR
https://github.com/docker/docker.github.io/pull/14883

PR #15124, opened 2022-07-14
Remove unneeded <br> HTML tags

crazy-max:
follow-up https://github.com/docker/docker.github.io/pull/14908#discussion_r921417788

cc @usha-mandya 

### DIFF
--- a/language/golang/index.md
+++ b/language/golang/index.md
@@ -29,5 +29,3 @@ You can containerize your own Go application using the examples and resources pr
 Let's get started!
 
 [Build your Go image](build-images.md){: .button .outline-btn}
-
-<br />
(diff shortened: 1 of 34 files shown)


aevesdocker (approved):

language/golang/index.md
```diff
@@ -29,5 +29,3 @@ You can containerize your own Go application using the examples and resources pr
 Let's get started!
 
 [Build your Go image](build-images.md){: .button .outline-btn}
-
-<br />
```
thaJeztah, line 33: I see these were all added at the end of the page; do we know if there was a reason for this; e.g. was (or _is_) there perhaps a styling issue to fix to have some whitespace at the end of our pages?

crazy-max: I don't see why tbh:

Before:

image

After:

image

thaJeztah: The only reason I can come up with would be something related to the left-hand or right-hand navigation, where it would try to scroll to the right location, and potentially ends up in a loop if it isn't able to scroll to that 🤔 

But perhaps it's not an issue

usha-mandya (approved):
LGTM

PR #15125, opened 2022-07-14
Issue 15094 - how to access get started

### LINKED ISSUE #15094: Missing info in Getting Started: Start the Tutorial
In the *Start the Tutorial* section of https://docs.docker.com/get-started/,  a novice user is told to run this command:
```
$ docker run -d -p 80:80 docker/getting-started
```
They are however not told to open a browser at localhost:80,  so this command seems to do nothing,  which is confusing.

Also,  for those (like me) who already have a webserver running locally,  the above command will fail;  you need to specify another port, not the default of 80.

craig-osterhout:
### Proposed changes

Added info on how to access get started tutorial after running the image. Also added info about changing ports if users already have something listening on port 80.

### Related issues (optional)

Fixes #15094
ENGDOCS-817

### DIFF
--- a/get-started/index.md
+++ b/get-started/index.md
@@ -106,9 +106,9 @@ $ docker run -d -p 80:80 docker/getting-started
 
 You'll notice a few flags being used. Here's some more info on them:
 
-- `-d` - run the container in detached mode (in the background)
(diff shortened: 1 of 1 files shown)


jerae-duffin (approved):
LGTM

PR #15133, opened 2022-07-15
move s3 redirects to jekyll redirect_from

crazy-max:
follow-up https://github.com/docker/docker.github.io/pull/15109

This is in preparation for a follow-up PR to handle redirects directly in our Cloudfront distribution: https://github.com/crazy-max/docker.github.io/commit/d44d84060943631b7a73d8b9411364edcff8e434


index.md
```diff
@@ -39,4 +49,24 @@ redirect_from:
...
 - /engine/reference/commandline/cluster_update/
 - /engine/reference/commandline/cluster_version/
+- /v1.4/
+- /v1.5/
+- /v1.6/
+- /v1.7/
+- /v1.8/
+- /v1.9/
+- /v1.10/
+- /v1.10/
```
thaJeztah, line 59: I'm not sure these would work; the current redirects match by prefix (so `/v1.4/engine/reference/commandline/cli/` (e.g.) would redirect to `/engine/reference/commandline/cli/` etc.

crazy-max: Ohhh that's a very good point, didn't catch that we were using `KeyPrefixEquals`

crazy-max: I guess we have to handle that with the lambda edge function I was working on.

thaJeztah: Yeah, we could add some of those I guess; then again, the current ones works, so perhaps no need to spend too much time on that for now

PR #15147, opened 2022-07-18
Fix other dead link in compose doc

BinaryWizard904:
Turns out the dead-link to a permissions calculator fixed in PR #14985 only fixed one of two links.
This pull request fixes the other one.

### Related issues (optional)
PR #14985


dockertopia (approved):
Thanks for the fix @BinaryWizard904.

PR #14803, opened 2022-05-22
refactor and mutualize netlify and aws clients for deployment

crazy-max:
follow-up https://github.com/docker/docker.github.io/pull/14685

Refactor and mutualize our local actions in https://github.com/docker/docker.github.io/tree/master/.github/actions into a single one called `releaser`. this one is containerized so it can also been tested locally with `buildx bake` and doesn't require a GitHub Runner.

`netlify-deploy` and `netlify-clean` has been refactored and is using now the [official Go library](github.com/netlify/open-api) instead of the [npm CLI](https://github.com/docker/docker.github.io/blob/9294453e5fe816824ba855e92eae3fc5b16c9c11/.github/actions/netlify-deploy/Dockerfile#L4) so we are aligned with the AWS one that updates the website config and used the AWS Go SDK.

`update-website-config` had been refactored a bit and is now available as a subcommand of `releaser` app: `releaser aws update-config`. It's also updated to a more recent AWS Go SDK. In a follow-up we should update to the [AWS SDK v2](https://github.com/aws/aws-sdk-go-v2).

```shell
# update website config (staging)
$ AWS_REGION=us-east-1 \
  AWS_S3_BUCKET=docs.docker.com-stage-us-east-1 \
  AWS_S3_WEBSITE_CONFIG=_website-config-docs-stage.json \
  AWS_ACCESS_KEY_ID=foo \
  AWS_SECRET_ACCESS_KEY=bar \
  docker buildx bake aws-update-config
```

```shell
# remove site from Netlify
$ NETLIFY_AUTH_TOKEN=foo \
  NETLIFY_SITE_NAME=docker/docker.github.io/master \
  docker buildx bake netlify-remove
```

PR #15148, opened 2022-07-18
fix publish workflow

crazy-max:
fix an issue with #14803 when publish workflow is triggered in https://github.com/docker/docker.github.io/runs/7394165211?check_suite_focus=true#step:8:203


crazy-max:
Going YOLO with this one to avoid blocking publications :innocent:

thaJeztah (approved):
LGTM

PR #15151, opened 2022-07-18
cloudfront update with lambda function for redirects

crazy-max:
follow-up and take 2 of https://github.com/docker/docker.github.io/pull/15118

As replace directive for an s3 bucket is limited to 50 we can't use them. It's also really slow to propagate each of them with the AWS CLI (~30m).

This pr adds a new function to `releaser` that will update a lambda edge function to handle redirects directly in our cloudfront distribution.

I have created a [`lab` branch](https://github.com/crazy-max/docker.github.io/tree/lab) on my fork and also a S3 bucket with a Lambda Edge function attached to a new Cloudfront distribution on my personal AWS account to test this feature and so far so good:

```console
$ curl -s -D - https://dom3zg2ev0we5.cloudfront.net/get-started/part2/ -o /dev/null
HTTP/2 301 
content-length: 0
server: CloudFront
date: Mon, 18 Jul 2022 20:15:47 GMT
location: /get-started/02_our_app/
x-cache: LambdaGeneratedResponse from cloudfront
via: 1.1 d0229dbe69f77738f3ccab386a045ad8.cloudfront.net (CloudFront)
x-amz-cf-pop: CDG52-P2
x-amz-cf-id: _6ds2VEHsPPvW2w8IuT38Gqw8GmPoJCnMx4rN9MiU9OR1S72LUXDTQ==
```

Also looking at #15133, I have added a new `redirects-prefixes.json` file that handles redirects with prefixes so we can get rid of replaces in our s3 bucket with hardcoded domain inside it and have everything in place in our lambda edge function:

```console
$ curl -s -D - https://dom3zg2ev0we5.cloudfront.net/ee/sqdsqdsqdsqds -o /dev/null
HTTP/2 301 
content-length: 0
server: CloudFront
date: Mon, 18 Jul 2022 20:16:33 GMT
location: /
x-cache: LambdaGeneratedResponse from cloudfront
via: 1.1 8b20ff9a1799265d378bf510ac3db6de.cloudfront.net (CloudFront)
x-amz-cf-pop: CDG52-P2
x-amz-cf-id: KyHvFjoVvM9doj9iuCuAv4CYaxNoIG2GlhhB7KTwEcm2ntQLKDsVqQ==
```

* Cloudfront URL: https://dom3zg2ev0we5.cloudfront.net/
* Deployment demo: https://github.com/crazy-max/docker.github.io/runs/7397172669?check_suite_focus=true

```
[...]

PR #15153, opened 2022-07-19
allow search engines from crawling only on production

crazy-max:
while working on #15151, I found out that crawling for search engines was enabled for staging and netlify previews (pull requests): https://www.google.com/search?q=site%3Adocs-stage.docker.com


thaJeztah:
I think we set a `noindex` in the page's head on stage and preview;

Screenshot 2022-07-19 at 09 21 41


Looks like we picked the `sitemap` for that (maybe we should have something more explicit?) https://github.com/docker/docker.github.io/blob/38fec0d159134a9af7e8a3c226057a114b0622be/_includes/head.html#L33-L35

robots.txt
```diff
@@ -1,3 +1,13 @@
...
+{%- comment -%}
+This liquid template is necessary to only allow search engine crawling in
+production environment. We don't want staging or long-lived previews on Netlify
+to be indexed by search engines.
+{%- endcomment -%}
+
+{%- if jekyll.environment == 'production' -%}
 User-agent: *
 
 # Docker Engine archives
```
thaJeztah, line 13: Once we have proper redirects, we can probably remove all of these (I _think_ these were added when we were still hosting archives, but they also helped to prevent Google indexing archived URLs it found elsewhere with "new" content, but using the "old" URL - not 100% sure though).

robots.txt
```diff
@@ -26,3 +36,7 @@ Disallow: /ee/
 Disallow: /apidocs/v*
 
 Sitemap: https://docs.docker.com/sitemap.xml
+{%- else -%}
+User-agent: *
+Disallow: /
+{%- endif -%}
```
thaJeztah, line 42: Not sure if it's checking for that, but looks like the `%-` and `-%` on both sides causes no newline to be present at the end; perhaps (one of them) needs to be removed

crazy-max:
> Looks like we picked the `sitemap` for that (maybe we should have something more explicit?)

Hum yeah I think we should use `{%- if jekyll.environment == 'production' -%}` as cond

_includes/head.html
```diff
@@ -30,7 +30,7 @@
 {%- endif -%}
 <head>
   <meta http-equiv="Content-Type" content="text/html; charset=UTF-8">
-  {%- if page.sitemap == false or site.GH_ENV == "gh_pages" %}
+  {%- if jekyll.environment != 'production' %}
```
thaJeztah, line 33: We probably need to keep the `page.sitemap` as well; it's used for pages that are not removed, but no longer present in the TOC (and/or shouldn't be indexed, such as pages about deprecated things)

crazy-max: Oh yes indeed, fixed.

thaJeztah (approved):
LGTM, thanks!

PR #15164, opened 2022-07-19
Changing the structure of the section

dockertopia:
### Proposed changes
Changes as previously discussed in https://github.com/docker/docker.github.io/pull/14910#discussion_r924576665

### Related issues (optional)

https://github.com/docker/docker.github.io/pull/14910


jedevc:
Personally, I would rather we keep the pages together. If we measure pages in screen heights, then for me, then it works out to only around 1-1.5 screens, so I have to click between them to navigate instead of scrolling. I also think most of this information makes sense together in a broad overview - when we dive into the details of how each of these individual parts work, I think we should definitely split then, but maybe not here?

_data/toc.yaml
```diff
@@ -1392,11 +1392,15 @@ manuals:
...
-        - path: /build/buildx/
-          title: Working with Buildx
         - path: /build/buildx/install/
           title: Install Buildx
-
+        - path: /build/buildx/
+          title: Working with Buildx
+        - path: /build/buildx/multiple-builders/
+          title: Using multiple builders
+        - path: /build/buildx/
```
jedevc, line 1401: The navigation is broken here:
```suggestion
        - path: /build/buildx/multiplatform-images
```

dockertopia:
> Personally, I would rather we keep the pages together. If we measure pages in screen heights, then for me, then it works out to only around 1-1.5 screens, so I have to click between them to navigate instead of scrolling. I also think most of this information makes sense together in a broad overview - when we dive into the details of how each of these individual parts work, I think we should definitely split then, but maybe not here?

We're going to have more content for these pages, and we can get useful analytics if they are sitting in different pages. I honestly think this improves discoverability of content even if we need tweak the placement of some sections. Let's discuss in our next sync.

dockertopia:
As discussed in sync last week.

build/buildx/multiarch-images.md
```diff
@@ -0,0 +1,80 @@
+---
+title: Building multi-platform images
+description: Different strategies for building multiplatform images
+keywords: build, buildx, BuildKit, multiplatform images
+---
+
+
```
crazy-max, line 7: dup new line to remove

```suggestion

```

build/buildx/multiple-builders.md
```diff
@@ -0,0 +1,38 @@
+---
+title: Using multiple builders
+description: How to instantiate and work with multiple builders
+keywords: build, buildx, BuildKit, builders, build drivers
+---
+
+
```
crazy-max, line 7: same, dup new line

```suggestion

```

_data/toc.yaml
```diff
@@ -1392,11 +1392,14 @@ manuals:
...
-          title: Working with Buildx
         - path: /build/buildx/install/
           title: Install Buildx
-
+        - path: /build/buildx/
+          title: Working with Buildx
+        - path: /build/buildx/multiple-builders/
+          title: Using multiple builders
+        - path: /build/buildx/multiarch-images/
+          title: Bulding multi-platform images
```
crazy-max, line 1402:
```suggestion
          title: Building multi-platform images                  
```

build/buildx/multiarch-images.md
```diff
@@ -0,0 +1,80 @@
+---
+title: Building multi-platform images
+description: Different strategies for building multiplatform images
```
crazy-max, line 3: let's stay consistent in our wording:

```suggestion
description: Different strategies for building multi-platform images
```

build/buildx/multiarch-images.md
```diff
@@ -0,0 +1,80 @@
...
+keywords: build, buildx, BuildKit, multiplatform images
```
crazy-max, line 4:
```suggestion
keywords: build, buildx, buildkit, multi-platform images
```

_data/toc.yaml
```diff
@@ -1392,11 +1392,14 @@ manuals:
...
-        - path: /build/buildx/
-          title: Working with Buildx
         - path: /build/buildx/install/
           title: Install Buildx
-
+        - path: /build/buildx/
+          title: Working with Buildx
+        - path: /build/buildx/multiple-builders/
+          title: Using multiple builders
+        - path: /build/buildx/multiarch-images/
```
crazy-max, line 1401: We should be consistent:

```suggestion
        - path: /build/buildx/multi-platform-images/
```

Needs to rename `multiarch-images.md` to `multi-platform-images.md`

_data/toc.yaml
```diff
@@ -1392,11 +1392,14 @@ manuals:
       title: Hello Build
     - sectiontitle: Buildx
       section:
-        - path: /build/buildx/
-          title: Working with Buildx
         - path: /build/buildx/install/
           title: Install Buildx
-
+        - path: /build/buildx/
```
crazy-max, line 1397: Not sure about the order, looks confusing to me to have the parent page after a subpage.

As I can see with compose it's the first page: https://docs.docker.com/compose/ so we should be consistent.

crazy-max (changes requested):

_data/toc.yaml
dockertopia: It's true. It's called "multi-arch" in Desktop docs, and in conversation we do say multi-arch all the time.
I think most users would equate multi-platform with macOs/Linux/Windows. 
And architectures is clear that it refers to HW such as `arm`, `mips`, `ppc64le`, and even `s390x`.

Also,  taking for example this sentence in the original content:
"When you invoke a build, you can set the --platform flag to specify the target platform for the build output, (for example, linux/amd64, linux/arm64, or darwin/amd64)" or this command example "docker buildx build --platform linux/amd64,linux/arm64 "? 
We also say "An experimental package is available for arch-based distributions."  in Engine doc set.

Just trying to point out that it seems to be a combination of both. Platform is not enough to determine the arch (following the above interpretation): imho multi-architecture being more specific is a better candidate.

dockertopia (commented):
Left a couple of comments and committed line break removals and lowercased keywords.

_data/toc.yaml
crazy-max: Hum right, WDYT @tonistiigi @jedevc?

build/buildx/index.md
```diff
@@ -1,7 +1,7 @@
 ---
-title: Buildx
+title: Working with Buildx
 description: Working with Docker Buildx
-keywords: build, buildx, buildkit
+keywords: build, buildx, BuildKit
```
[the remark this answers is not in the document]
thaJeztah: Good question; I _think_ either way _may_ work, as long as we're consistently using the same.

Looks like the search auto-complete uses them case-insensitive, but it may be that they're used different elsewhere;

Screenshot 2022-07-26 at 11 04 03

"elsewhere" could be as part of an URL and/or anchor, in which case it's _possible_ that they're treated as "different tags". I guess "lowercase" would be the safest option for these 🤔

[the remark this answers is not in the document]
crazy-max: Ah indeed there is the auto-complete search but I was thinking about search engines and the `meta name="keywords"` that is case-insensitive. See also https://github.com/docker/docker.github.io/pull/14642#discussion_r878979334 about `keywords` meta tag abuse. So might only be relevant for our auto-complete.

_data/toc.yaml
jedevc: I've noticed this inconsistency as well, agreed it would be good to clarify this, but maybe out of scope for this PR, there's a *lot* of build docs that mix-and-match between the two, so I'd propose to pick something here that works, and then discuss separately later.

Personally, I think multi-platform is better when talking about building images. A couple reasons:
- This is what the OCI spec refers to - the CPU architecture is part of the platform (the OS is the other main component). 
- Buildkit will eventually (hopefully) support building windows images. Say we have an `amd64` linux image and a `amd64` windows image - in this case, it's not as accurate to talk about multi-arch images, even though the underlying mechanism is the same. Multi-platform is a nice catch-all.
- The buildx/buildkit flags are `--platform` - calling it multi-arch seems like it can be confusing, multi-platform seems more inline with existing behavior.

Maybe there are still use-cases for talking about multi-arch though, for example, maybe around QEMU, when we're using linux throughout, but emulating different CPU architectures.

> An experimental package is available for arch-based distributions

I think this is probably unrelated, I think this refers to [ArchLinux](https://archlinux.org/), not multi-architecture/platform support.

build/buildx/index.md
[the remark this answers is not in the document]
dockertopia: So, in your [previous example](https://github.com/docker/docker.github.io/pull/14642#discussion_r878979334) all of them are like Docker, BuildKit, etc.
image

Comment: I agree that making them all lowercase allows less room for creating multiple like (docker, Docker, dockerfile, Dockerfile) in terms of metadata, in terms of the presentation (they are visible) they make more sense with case. It would be great to decouple that. As in, write them in suitable case (rendering) and them being always interpretated in case insensitive way(metadata). 

Question for @usha-mandya do we then change all keywords to be lowercase in our docs sets?

_data/toc.yaml
dockertopia: Agreed on the Engine example. I revisited it and @jedevc is right, but a simple docs search renders this
image 
image
( some are community hits)

It seems for the discussion there's the case for both in content, when platform is the same and we do want to distinguish architectures, but that we'd like to introduce the feature as multi-platform as it's seen as capturing both.

But it leaves the case of harmonizing the remaining docs content.

_data/toc.yaml
crazy-max: Maybe `Overview` and `Install` is enough as it's already a subsection of `Buildx`. Don't think we need to repeat ourselves?

dockertopia: I see what you mean, that there's hierarchy providing context. But when we need to derive it from the folder context and there's multiple in the doc set, I think it's best for any "just" 'overview' or 'install' to be the ones for the main section (so doc set root index or install) for clear disambiguation in a glance.

_data/toc.yaml
```diff
@@ -1393,10 +1393,13 @@ manuals:
...
       section:
         - path: /build/buildx/
-          title: Working with Buildx
+          title: Buildx overview
         - path: /build/buildx/install/
           title: Install Buildx
-
+        - path: /build/buildx/multiple-builders/
+          title: Using multiple builders
+        - path: /build/buildx/multiplatform-images/
```
jedevc, line 1401:
```suggestion
        - path: /build/buildx/multi-platform-images/
```

build/index.md
```diff
@@ -27,3 +27,60 @@ BuildKit is the backend evolution from the Legacy Builder, it comes with new
...
+    * `docker-container` – spin up a dedicated builder in a docker container to unlock more advanced features like advanced caching, and multi-platform images.
+    * `kubernetes`– connect to your Kubernetes cluster to unlock more features and to run your builds at scale.
+    * `remote`–  manually manage your own BuildKit builder and connect directly to it.
+
+* **Optimizing builds with cache management**  
+Improve build performance by using a persistent shared build cache to avoid repeating costly operations such as package installations, downloading files from the internet, or code build steps.
+
+* **Creating build-once, run-anywhere with multi-architecture builds**
+Collaborate across platforms with one build artifact.  
+See [Build multi-platform images](buildx/multiplatform-images.md).
```
crazy-max, line 55:
```suggestion
See [Build multi-platform images](buildx/multi-platform-images.md).
```

crazy-max (changes requested):

build/buildx/index.md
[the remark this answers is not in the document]
usha-mandya: @dockertopia Agree on having a standard for keywords. Suggest we use lowercase going forward. 

We can create a ticket in the backlog to retrospectively update all keywords to be lowercase. It's not a high priority as the search results are case-insensitive.

build/index.md
dockertopia: I'm keeping the filename as two words for simplicity in xrefs. I think you'll find it's OK to not reflect the preferred spelling in the name for the file.
We can revisit this later both in terms of style guide (preferred spelling for hyphenated words so it's applied across doc sets) or simple shared guidelines for naming doc source files (even if just writing recording something that's been implicit so far).
The idea is whatever points of standardization coming up to be recorded and elevated so there's alignment across teams/doc sets and everyone on the docs teams has an opportunity to be clear on the matter. Again, the is not to dismiss the comments but to address them out of the scope of the PR.

build/buildx/index.md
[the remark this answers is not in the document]
dockertopia: Thanks @usha-mandya. Sounds good.

crazy-max (approved):
As discussed let's decide what do to for `multiplatform` vs `multiarch` `multi-platform` wording and fix references about this in our docs in a follow-up.

PR #15171, opened 2022-07-20
pic-fix

aevesdocker:


dockertopia (approved):
LGTM. 
On a note, just wondering if we want/need to keep source file names shorter.

aevesdocker:
yes agreed they should be shorter, but I didn't want to spend too much time messing around with 'inherited' images.

PR #15181, opened 2022-07-21
Adding additional note for apple silicon users

jk1510karti:
### Proposed changes

I have downloaded the Docker Desktop from the following page and the binary doesn't have the ".dmg" extension so I could not double-click and install it on my mac running on M1 chip. 

Please add this instruction to the primary documentation. Otherwise, update the release repository with a proper extension.


aevesdocker:
Hi @jk1510karti thank you for the PR. Could you tell me what browser you used to download Docker Desktop? I've tested it on Chrome and I see the .dmg extension

aevesdocker:
Hi @jk1510karti, thank you for responding. Other users have reported this issues to us (https://github.com/docker/for-mac/issues/6411https://github.com/docker/for-mac/issues/6411), however we've checked everything on our side and the download link looks good. We believe it's a bug in Firefox 102 and it should be fixed in a new release. For now you have to use a different browser or rename the file afterwards.

PR #15184, opened 2022-07-21
Issue-15170 specify arch for command

### LINKED ISSUE #15170: Specify the Device the User is Using Before Showing the Command
[get-started/06_bind_mounts.md](https://docs.docker.com/get-started/06_bind_mounts/#start-a-dev-mode-container):

The doc shows some commands for Apple device with Intel chips by default, which seems to assume the user is using Apple device with Intel chips, which is not always true, so for scenarios like below that have different commands for different devices, we'd better make it clear about the device type ahead of the actual command.

> Run the following command from the app directory **if you are using an Intel Mac**. We’ll explain what’s going on afterwards:
> 
> ```shell
>  $ docker run -dp 3000:3000 \
>      -w /app -v "$(pwd):/app" \
>      node:12-alpine \
>      sh -c "yarn install && yarn run dev"
> ```
> If you are using Windows then use this command in PowerShell:
> ```shell
>  PS> docker run -dp 3000:3000 `
>      -w /app -v "$(pwd):/app" `
>      node:12-alpine `
>      sh -c "yarn install && yarn run dev"
> ```
> If you are using an Apple silicon Mac or another ARM64 device, then use the following command.
> 
> ```shell
>  $ docker run -dp 3000:3000 \
>      -w /app -v "$(pwd):/app" \
>      node:12-alpine \
>      sh -c "apk add --no-cache python2 g++ make && yarn install && yarn run dev"
> ```

craig-osterhout:
### Proposed changes

In /get-started/06_bind_mounts.md:
- specified arch for command because it was ambiguous
- made wording and grammar more consistent among the commands
- changed codeblock from bash to console to be more copy-paste friendly

### Related issues (optional)

Fixes #15170 
ENGDOCS-830

### DIFF
--- a/get-started/06_bind_mounts.md
+++ b/get-started/06_bind_mounts.md
@@ -40,7 +40,9 @@ So, let's do it!
 
 1. Make sure you don't have any previous `getting-started` containers running.
 
-2. Run the following command from the app directory. We'll explain what's going on afterwards:
+2. Run the following command from the app directory. We'll explain what's going on afterwards.
+
+    If you are using an x86-64 Mac or Linux device, then use the following command.
 
     ```console
     $ docker run -dp 3000:3000 \
@@ -49,7 +51,7 @@ So, let's do it!
         sh -c "yarn install && yarn run dev"
     ```
 
-    If you are using Windows then use this command in PowerShell:
+    If you are using Windows, then use the following command in PowerShell.
 
     ```powershell
     PS> docker run -dp 3000:3000 `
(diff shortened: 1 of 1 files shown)


get-started/06_bind_mounts.md
```diff
@@ -40,7 +40,9 @@ So, let's do it!
 
 1. Make sure you don't have any previous `getting-started` containers running.
 
-2. Run the following command from the app directory. We'll explain what's going on afterwards:
+2. Run the following command from the app directory. We'll explain what's going on afterwards.
+
+    If you are using an x86-64 Mac or Linux device, then use the following command.
```
jerae-duffin, line 45: :

jerae-duffin (approved):
LGTM

PR #15198, opened 2022-07-24
volume-type-comparisons-mismatches

katxeus:
### Proposed changes
The volume type comparisons for "Populates new volume with container contents"	and Supports Volume Drivers had "No" for bind mounts instead of "Yes".
As bind mounts particularly control the exact mount-point on the host and used to provide additional data into containers like source code into the container for monitoring code changes.

### DIFF
--- a/get-started/06_bind_mounts.md
+++ b/get-started/06_bind_mounts.md
@@ -25,8 +25,8 @@ volume drivers are available to support other use cases ([SFTP](https://github.c
 | - | ------------- | ----------- |
 | Host Location | Docker chooses | You control |
 | Mount Example (using `-v`) | my-volume:/usr/local/data | /path/to/data:/usr/local/data |
-| Populates new volume with container contents | Yes | No |
-| Supports Volume Drivers | Yes | No |
+| Populates new volume with container contents | No | Yes |
+| Supports Volume Drivers | No | Yes |
 
 ## Start a dev-mode container
 


usha-mandya:
@thaJeztah Could you PTAL?

get-started/06_bind_mounts.md
```diff
@@ -25,8 +25,8 @@ volume drivers are available to support other use cases ([SFTP](https://github.c
 | - | ------------- | ----------- |
 | Host Location | Docker chooses | You control |
 | Mount Example (using `-v`) | my-volume:/usr/local/data | /path/to/data:/usr/local/data |
-| Populates new volume with container contents | Yes | No |
-| Supports Volume Drivers | Yes | No |
+| Populates new volume with container contents | No | Yes |
+| Supports Volume Drivers | No | Yes |
```
thaJeztah, line 29: Hm.. this doesn't look correct; only (named) volumes support volume drivers, and only volumes will copy data from the container to the volume (if it's empty); bind-mounts on the other hand take the files that are present on the host, and make them available to the container (but won't copy files that were present inside the container to the bind-mounted directory).

thaJeztah (changes requested):
Thanks for contributing; I don't think this change is correct, but I'm interested to hear what led to this pull request (or if there's improvement to make to explain the differences).

PR #15194, opened 2022-07-22
build: add bake section

crazy-max:
As discussed with @dockertopia and @jedevc, split bake and guides from #14905 as there is more work to do with guides.

Signed-off-by: CrazyMax <crazy-max@users.noreply.github.com>

PR #15218, opened 2022-07-26
jekyll: keep paragraph of reading time and use include

crazy-max:
While working on #15194 I found out there was no proper spacing after the title of some pages:

image
> https://deploy-preview-15194--docsdocker.netlify.app/build/bake/

Looking at the layout it seems it was done on purpose to handle the [estimated reading time](https://github.com/docker/docker.github.io/blob/f540f58541b02236262f93272790b706c2c48b24/_layouts/docs.html#L34) of page so rendering looks right:

image
> https://docs.docker.com/desktop/

This PR now keeps the paragraph even if the estimated reading time is not rendered so we are consistent across pages:

image

I also found out that this block was duplicated and available as an include in https://github.com/docker/docker.github.io/blob/f540f58541b02236262f93272790b706c2c48b24/_includes/read_time.html so switch to the include instead.

### DIFF
--- a/_layouts/docs.html
+++ b/_layouts/docs.html
@@ -28,12 +28,7 @@
                             {%- if page.advisory -%}
                             <blockquote>{{ site.data.advisories.texts[page.advisory] | markdownify }}</blockquote>
                             {%- endif -%}
-                            {%- unless page.skip_read_time == true -%}
-                            {%- assign words = content | number_of_words -%}
-                            {%- if words >= 360 -%}
-                            <p><em class="reading-time">Estimated reading time: {{ words | divided_by:180 }} minutes</em></p>
-                            {%- endif -%}
-                            {%- endunless -%}
+                            <p>{%- include read_time.html -%}</p>
                             {{ content }}
                             {%- unless page.notags == true -%}
                             {%- assign keywords = page.keywords | split:"," -%}
--- a/_includes/read_time.html
+++ b/_includes/read_time.html
@@ -1,4 +1,6 @@
+{%- unless page.skip_read_time == true -%}
 {%- assign words = content | number_of_words -%}
 {%- if words >= 360 -%}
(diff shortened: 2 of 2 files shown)


_layouts/docs.html
```diff
@@ -28,12 +28,7 @@ <h1>{{ page.title }}</h1>
                             {%- if page.advisory -%}
                             <blockquote>{{ site.data.advisories.texts[page.advisory] | markdownify }}</blockquote>
                             {%- endif -%}
-                            {%- unless page.skip_read_time == true -%}
-                            {%- assign words = content | number_of_words -%}
-                            {%- if words >= 360 -%}
-                            <p><em class="reading-time">Estimated reading time: {{ words | divided_by:180 }} minutes</em></p>
-                            {%- endif -%}
-                            {%- endunless -%}
+                            <p>{%- include read_time.html -%}</p>
```
thaJeztah, line 31: I guess alternatively, we could add a style for block quotes immediately following a H1;

```css
h1+blockquote {
  margin-top: 24px;
}
```

thaJeztah (approved):
LGTM, but left one suggestion we could try

usha-mandya (approved):
LGTM

_layouts/docs.html
crazy-max: Yes but looks like the `<blockquote>` is conditional?: https://github.com/docker/docker.github.io/blob/72a3a45c0e09a7c101d66dca13c7fae3916c96f3/_layouts/docs.html#L28

Maybe add a new css class to https://github.com/docker/docker.github.io/blob/72a3a45c0e09a7c101d66dca13c7fae3916c96f3/_layouts/docs.html#L26

crazy-max:
Let's get this one in, will see to enhance this with https://github.com/docker/docker.github.io/pull/15218#discussion_r930934027 in a follow-up.

PR #15223, opened 2022-07-27
jekyll(css): add experimental note

crazy-max:
While working on #15194, I found out the "note" block design was not really welcoming for new users:

image

This PR adds a new note type called `experimental` that will look like this:

image

image

Also needs to upgrade Font Awesome because the [vial icon](https://fontawesome.com/icons/vial?s=solid) is missing with the current release we have.

Not that I moved the Font Awesome library to an `assets` directory because accessing the resources from their css moved the fonts to a `webfonts` folder which is relative to their css. I didn't want to add another folder to root path of this repo as it's already bloated and I think it makes more sense to have the resources dedicated to the libraries in their respective folders so you know which css is linked to which font (same for javascript).

I think we should also do the same for other "assets" we have in `css`, `js`, `favicons`, `fonts` folders and move these into the `assets` folder in a follow-up.

Maybe also in the future use npm and a `package.json` to manage these dependencies. (cc @mark-dr)


thaJeztah (approved):
LGTM, I like it 👍

thaJeztah:
I recall I did some testing for redesign these in https://github.com/docker/docker.github.io/pull/11641 as well

crazy-max:
> I recall I did some testing for redesign these in #11641 as well

Ah looks neat! I'm using admonition on personal repos and I think they look great: https://squidfunk.github.io/mkdocs-material/reference/admonitions/#supported-types

image

thaJeztah:
Yeah, my intent with that PR was to have a clear "box" for these (as some are quite lengthy), without being _too much_ separate from the content (i.e., not a "in your face" background color, so that we also don't have to re-invent colour-schemes that "work" for each and every one of them). Also to make sure it's all just CSS, and not too much HTML markup needed.

usha-mandya:
@mark-dr Could you PTAL?

usha-mandya:
Thanks @crazy-max. Going forward, we need to use `{:.experimental}` to use the new style for experimental commands. Is this correct? Also, is there anything else we should do to get this style? TIA

crazy-max:
> Thanks @crazy-max. Going forward, we need to use `{:.experimental}` to use the new style for experimental commands. Is this correct? Also, is there anything else we should do to get this style? TIA

yes indeed, see https://github.com/docker/buildx/pull/1236

mark-dr:
The changes look good to me on a technical level, but we should be careful about Docs drifting away from the design system. Our existing alerts are based on the Alerts in this figma doc:

https://www.figma.com/file/R0GyrsN8tRxJIbti0KY6S3/%E2%9C%A8Components-%26-Patterns?node-id=2%3A8

In there, we use green to indicate success (i.e. the thing you just attempted was successful). Maybe this should be an info alert, or maybe even a "special" purple one. WDYT, @MCallaghanDocker?

mark-dr:
See conversation here - it's currently leaning toward blue being the right colour for this.

https://docker.slack.com/archives/C0324PUQMFF/p1658928599667639

crazy-max:
> we should be careful about Docs drifting away from the design system. Our existing alerts are based on the Alerts in this figma doc:
> 
> https://www.figma.com/file/R0GyrsN8tRxJIbti0KY6S3/%E2%9C%A8Components-%26-Patterns?node-id=2%3A8

Of course! Let me know what would work best for us.

> In there, we use green to indicate success (i.e. the thing you just attempted was successful). Maybe this should be an info alert

IMO there are two patterns to take into account. There is callouts and actionable/story states. Callouts being passive like this "experimental" note and "success" for actionable ones or describing a story.

> or maybe even a "special" purple one.

This one LGTM too.

mark-dr:
@crazy-max, how about blue (i.e. the "info" colour), with the vial?

crazy-max:
> @crazy-max, how about blue (i.e. the "info" colour), with the vial?

Yes sgtm!

crazy-max:
@mark-dr Here it is using the same palette as current `info` note.

image

image

mark-dr (approved):
Thanks for the color swap, looks good!

crazy-max:
Ok let's get this one in if ci ok

PR #15227, opened 2022-07-27
Fix linting :tada: 

jedevc:
### Proposed changes

Fix the linter to actually do the linting :smile:

The linter wasn't running at all, due to a mixup between the `--rules` and `--style` flag. This PR switches to the correct `--style` flag, and updates the style file with some appropriate linting rules.

As part of adding these linting rules (which shouldn't be too controversial), the issues they find in the code are fixed. These are split up per-patch, for ease-of-review, and for a nicer git-blame log.

As a follow-up, some more linting rules could be added, but for now, the checks enabled are ones that I considered to be "somewhat reasonable", could be fixed without too much effort, and shouldn't be an undue burden on future contributors.

### Related issues (optional)

#14806


crazy-max (approved):
LGTM thanks! PTAL @thaJeztah

jedevc:
The trailing spaces removal may not be right, and would like some more thoughts on it.

Apparently, kramdown supports using a trailing space like so:

```markdown
here is some text__
here is some more
```

Which renders as a line break in the middle of a paragraph. I think we rely on this behavior in some places, and there's too many instances to check through this manually and confirm if we'd be alright with changing the flow.

.mdlrc.style.rb
```diff
@@ -1,7 +1,14 @@
...
+rule 'MD011'
+rule 'MD018'
+rule 'MD019'
+rule 'MD020'
+rule 'MD021'
+rule 'MD029', :style => :ordered
+rule 'MD035'
+rule 'MD037'
+rule 'MD038'
+rule 'MD039'
```
jedevc, line 14: Use the linting rule alias names here.

desktop/extensions-sdk/dev/api/reference/interfaces/Docker.md
```diff
@@ -87,7 +87,7 @@ const containers = await ddClient.docker.listContainers();
 
 `Promise`<`unknown`\>
 
-___
+---
```
thaJeztah, line 90: I _think_ the `desktop/extensions-sdk/` directory (or at least _parts of it_) is generated from source, so these should be fixed in the code generating the Markdown.

(and at least temporarily excluded from listing)

/cc @gtardif

registry/deploying.md
```diff
@@ -414,7 +414,7 @@ secrets.
       --entrypoint htpasswd \
       httpd:2 -Bbn testuser testpassword > auth/htpasswd
     ```
-
```
thaJeztah, line 417: The `registry/` directory moved back to upstream https://github.com/distribution/distribution/pull/3397

(we still need to remove it here and/or set up a job to sync)

In the meantime, we should disable linting for this directory

thaJeztah (changes requested):
Thanks! overall, these _seem_ ok (there's a couple of "hidden" ones by GitHub that I may have missed); left some comments on "external" or "generated" content though

jedevc:
Changes:
- Have dropped the trailing whitespace commit for now, but might be worth investigation in a separate PR. 
- Excluded directories suggested by @thaJeztah 
- Have tidied the config file to use the aliases which should be clearer, and group them into categories.

dockertopia:
> The trailing spaces removal may not be right, and would like some more thoughts on it.
> 
> Apparently, Kramdown supports using a trailing space like so:
> 
> ```gfm
> here is some text__
> here is some more
> ```
> 
> Which renders as a line break in the middle of a paragraph. I think we rely on this behavior in some places, and there's too many instances to check through this manually and confirm if we'd be alright with changing the flow.

@docker/docs-team
This means two things:
* (In terms of linting) When we see two trailing spaces it could be intentional or accidental.
* (In terms of presentation) If linting removes all trailing spaces and no BR tags can be used, it means writers can only have paragraphs but never new lines.

#the case for using new lines
To avoid using br HTML tags I started using this allowed syntax in Kramdown. FYI since using this syntax I am now cleaning my files of incidental trailing spaces to remove any ambiguity.

Here's two examples:
image
It didn't make sense to create a section for each of the two files part of the same example (it would also not add great value or meaning as an element in the page TOC), but it made sense to clearly separate the sample files to comment on each one.

So the changing of the line gets the next text justified to the left, the same way as a section title or a Note or an Important element but without creating a lot of attention to it (and while being close enough to still be considered logically as part of the paragraph).
I'm using it here as well for the cross-references without having an overly conspicuous box of text (such as a Note).
image

I added some additional thoughts to a Notion page in our Authoring page.

thaJeztah:
> The trailing spaces removal may not be right, and would like some more thoughts on it.

Yes, we need to look into those; the _majority_ of them definitely didn't look _intentional_, and overall, depending on the "trailing whitespace" for line breaks (whitespace as in _actual `<space>`_) is troublesome (with many editors stripping trailing whitespace by default).

So yes, we should probably look if there's actual places where they're intentional, and see if those really need it (otherwise remove)

dockertopia:
> > The trailing spaces removal may not be right, and would like some more thoughts on it.
> 
> Yes, we need to look into those; the _majority_ of them definitely didn't look _intentional_, and overall, depending on the "trailing whitespace" for line breaks (whitespace as in _actual `<space>`_) is troublesome (with many editors stripping trailing whitespace by default).
> 
> So yes, we should probably look if there's actual places where they're intentional, and see if those really need it (otherwise remove)

If the only problem if they're are stripped in some occasions but not only or most. Then we just need to make sure that the output still is good also without the trailing spaces, instead of just cutting it. But if no other writer cares about this then maybe the discussion is unnecessary.

docker-hub/repos.md
```diff
@@ -47,9 +47,9 @@ images to this repository.
 
 2. Select a repository from the list, click **Settings** and then Delete Repository.
 
-  > **Note:**
-  >
-  > Deleting a repository deletes all the images it contains and its build settings. This action cannot be undone.
+    > **Note:**
+    >
+    > Deleting a repository deletes all the images it contains and its build settings. This action cannot be undone.
```
thaJeztah, line 52: Slightly leaning towards making this "note" an "important" / "warning", and putting it at the top, to make the numbered list more to-the-point. i.e. something like

```markdown
## Deleting a repository

> **Important**
>
> Deleting a repository deletes all the images it contains and its build settings. This action cannot be undone.
{: .important }


1. Sign into [Docker Hub](https://hub.docker.com){: target="_blank" rel="noopener" class="_"} and click **Repositories**.
2. Select a repository from the list, click **Settings** and then Delete Repository.
3. Enter the name of the repository to confirm the deletion and click **Delete**.
```


## Deleting a repository

> **Important**
>
> Deleting a repository deletes all the images it contains and its build settings. This action cannot be undone.
{: .important }


1. Sign into [Docker Hub](https://hub.docker.com){: target="_blank" rel="noopener" class="_"} and click **Repositories**.
2. Select a repository from the list, click **Settings** and then Delete Repository.
3. Enter the name of the repository to confirm the deletion and click **Delete**.

thaJeztah: Actually, given that the actual UX already shows a confirmation, perhaps an `info` is good enough, as it's more a "FYI"

(in general, we should reduce the number of "info", "warning" blocks, as sometimes they're basically just part of the description 😅 )

language/golang/build-images.md
```diff
@@ -427,9 +427,7 @@ content:
 ```dockerfile
 # syntax=docker/dockerfile:1
 
-##
 ## Build
-##
```
thaJeztah, line 432: heh. funny; it looks like it picked these up as Markdown, not as Dockerfile comments? (is that a bug in the linter?)

thaJeztah (commented):
some minor thoughts, but otherwise this all looks sane to me!

language/golang/build-images.md
jedevc: Yeah this is a bug as far as I can tell :cry:

I'll try and get a reproducible example at some point, and submit a patch upstream to fix when I have the time. It's strange, since this is inside a code block :confused: I think what's happened is it's picked up as a weird type of header? And then gone "this header isn't like any of the others", and thrown an error.

build/hellobuild.md
```diff
@@ -133,17 +133,22 @@ CMD flask run --host 0.0.0.0 --port 8000
 Go ahead and try this example in your local Docker installation or you can use [Play with Docker](https://labs.play-with-docker.com){:target="_blank" rel="noopener" class="_"} that provides you with a temporary Docker instance on the cloud.
 
 To test this example:
+
 1. Create a file hello.py with the content of sample A.
+
 2. Create a file named Dockerfile without an extension with the contents of sample B.
+
 3. From your Docker instance build it with `docker build -t test:latest .`
```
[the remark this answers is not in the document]
jedevc: This is documented [here](https://kramdown.gettalong.org/syntax.html#lists). Relevant quotes:

> The content of a list item is made up of either text or block-level elements. Simple list items only contain text like in the above examples. They are not even wrapped in a paragraph tag. If the first list text is followed by one or more blank lines, it will be wrapped in a paragraph tag

> The text of the last list item is also wrapped in a paragraph tag if all other list items contain a proper paragraph as first element. This makes the following use case work like expected, i.e. all the list items are wrapped in paragraphs

usha-mandya:
> If the only problem is they're are stripped in some occasions but not most. Then we just need to make sure that the output still is good also without the trailing spaces, instead of just cutting it. But if no other writer cares about this then maybe the discussion is unnecessary.

Agree. We need to consider any trailing spaces added intentionally. It looks like the ones being fixed by @jedevc in this PR weren't intentional. However, before we impose this MD rule globally, suggest we review this in the team sync.

build/hellobuild.md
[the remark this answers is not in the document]
jedevc: :facepalm: aha :smile:. Ideally we'd be able to explicitly mark GFM, which I don't think has some of these issues. But I can't seem to find support for this in markdown lint.

language/golang/build-images.md
thaJeztah: 😄 ah, yes, fun one indeed.

Changing it looks fine though for this (doesn't change the intent, so should be fine to change and make the linter happy)

build/hellobuild.md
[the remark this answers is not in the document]
thaJeztah: No worries; I guess it's fine for these (mostly interested in keeping the "short" numbered lists compact).

docker-hub/repos.md
jedevc: Could we circle back round to this one as a follow-up? I'd like to keep this stack of patches just minimal linting changes is possible, without any major changes in content.

thaJeztah: Yup, works for me!

thaJeztah (approved):
LGTM, thanks!

crazy-max (approved):

crazy-max:
@usha-mandya As discussed, this should be fine if you can PTAL, thanks

usha-mandya:
@crazy-max Thanks. I've invited @jedevc to our team Huddle tomorrow. Holding off the merge until then. Would like the team to look at the some of the rules so it's not a surprise when they create a PR.

PR #15231, opened 2022-07-27
dd: add missing redirect for old content

thaJeztah:
- relates to https://github.com/docker/docker.github.io/pull/15219
- fixes https://github.com/docker/docker.github.io/issues/15229
- fixes https://github.com/docker/docker.github.io/issues/15230
- fixes https://github.com/docker/docker.github.io/issues/15232
- fixes https://github.com/docker/docker.github.io/issues/15233
- fixes https://github.com/docker/docker.github.io/issues/15237
- fixes https://github.com/docker/docker.github.io/issues/15238

Adds redirects for content that was removed in cab9f914bf5185dedb8c4cdd8c7705e5c390c43d

I tried to pick the most logical location, or following existing redirects (some of them
I wasn't sure about, e.g. most of the old "landing" page URLs looked to be redirected to
"get-started", not "desktop overview").

Issue #15229, opened 2022-07-27
404 at /desktop/linux/


manoj701m:
I found a broken link : https://docs.docker.com/desktop/linux/#credentials-management

thaJeztah:
Thanks for reporting! Looks like https://github.com/docker/docker.github.io/pull/15219 removed the pages, but didn't add redirects.


/cc @aevesdocker @usha-mandya

thaJeztah:
opened https://github.com/docker/docker.github.io/pull/15231

PR #15219, opened 2022-07-27
ENGDOCS-846

aevesdocker:
Removes old pages from docker.github.io

PR #15236, opened 2022-07-27
Add /go/linux-credentials/ redirect

thaJeztah:
- relates to https://github.com/docker/docker.github.io/issues/15229
- relates to https://github.com/docker/docker.github.io/pull/15231
- relates to https://github.com/docker/docker.github.io/pull/15219

Docker Desktop for Linux currently has a link to;
https://docs.docker.com/desktop/linux/#credentials-management

which has moved, and didn't have a redirect, causing the link to break;
unfortunately we can't fix existing versions of Docker Desktop; redirects
are being added (but won't be able to redirect to the correct anchor).

This patch adds a https://docs.docker.com/go/linux-credentials/ redirect,
so that we can continue using that URL even if content moves around.

### DIFF
--- a/go/linux-credentials.md
+++ b/go/linux-credentials.md
@@ -1,0 +1,5 @@
+---
+title: How to set up and configure the pass credentials helper for Docker Desktop for Linux
+description: How to set up and configure the pass credentials helper for Docker Desktop for Linux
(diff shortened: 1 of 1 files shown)


craig-osterhout (approved):

thaJeztah:
Thanks! Let me get this one in to unblock the PR in docker desktop itself

PR #15239, opened 2022-07-27
Speed up inner loop

mikesir87:
### Proposed changes

After playing around with some docs changes, I was getting tired of having to rebuild and restart my containers every time I made a change. So, I decided to take a look at how to fix this. After all, it is a Jekyll project so we _should_ be able to use `jekyll serve`, right?

- Update our custom plugins to not produce file changes if nothing is being changed. Otherwise, we get in a constant file update loop
    - For the fetch_remote plugin, I added a `completion_marker` config option that serves as a filepath to determine if the fetch has already occurred. This prevents resource download from occurring on every file change.
- Add a new `dev` stage to the Dockerfile that starts jekyll serve. It uses a new `dev-start.sh` script to still support the production test use case (which adds the production config file)
- Update the `docker-compose.yaml` file to target this new `dev` stage and expose the livereload port

### Related issues (optional)

None

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -40,6 +40,11 @@ RUN bundle update \
 FROM scratch AS vendor
 COPY --from=vendored /out /
 
+# Serve the files using jekyll to enable file watching
+# and livereload for a faster inner loop
+FROM gem AS dev
+CMD ["./dev-start.sh"]
+
 # Build the static HTML for the current docs.
 # After building with jekyll, fix up some links
 FROM gem AS generate
--- a/_config.yml
+++ b/_config.yml
@@ -123,6 +123,7 @@ defaults:
 # Fetch upstream resources (reference documentation) used by _plugins/fetch_remote.rb
 # - repo is the GitHub repository to fetch from
 # - ref the Git reference
+# - completion_marker is a file path that, if it exists, will prevent a redownload from occurring
 # - paths is a list to the resources within the remote repository
 #   - dest is the destination path within the working tree
 #   - src is a list of glob source paths within the remote repository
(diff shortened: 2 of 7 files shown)


mikesir87:
Saw the build failed and it's because of the new `build/bake` docs... all of the same files exist locally _and_ in the remote repo to support the frontmatter. So, there's no way to really know if the files have already been fetched or not by simply using a marker file. Would be nice to have some mechanism to know we've fetched/setup the files without having to do so yet again.

mikesir87:
Same thing for the `https://github.com/moby/buildkit` repo too. The file it's loading already exists with the `fetch_remote` frontmatter.

mikesir87:
Ok. I think it's good to go for a review now. Bring on the feedback! 😄

usha-mandya:
Thank you @mikesir87. 

@crazy-max @thaJeztah Could you PTAL?

crazy-max:
@mikesir87 There are some conflicts, can you rebase please?

> * For the fetch_remote plugin, I added a `completion_marker` config option that serves as a filepath to determine if the fetch has already occurred. This prevents resource download from occurring on every file change.

We discussed about having better cache for `fetch_remote` plugin with @thaJeztah. Your `completion_marker` option might be good if current `ref` still has these files remotely otherwise it will break. It would also skip files newly added remotely.

As we currently download remote resources as zip and some references are not tied to a specific commit (e.g., using a branch like `master`) it's kinda hard to have proper caching. We should instead properly clone the repo and check for changes using git metadata.

> * Add a new `dev` stage to the Dockerfile that starts jekyll serve. It uses a new `dev-start.sh` script to still support the production test use case (which adds the production config file)

Live reload lgtm but I think we should create another PR for this one and first improve caching. WDYT?

usha-mandya:
@mikesir87 @crazy-max As we are gearing up to migrate Docs to Hugo, I'm wondering if we should bring in any changes from this PR to 'main' before the migration. WDYT?

PR #15215, opened 2022-07-26
fix commands to create registry.json (ENGDOCS-844)

craig-osterhout:
### Proposed changes

The Mac commands for creating a registry.json do not work. 
- `touch` does not create directories. Added `mkdir -p` to create directories. 
- The directory path has a space. Added quotes around directory path.

### Related issues (optional)
ENGDOCS-844

PR #15243, opened 2022-07-27
fix instructions for registry.json

thaJeztah:
- follow-up to https://github.com/docker/docker.github.io/pull/15215

The previous instructions fixed situations where the directory
and file didn't exist, but had some issues remaining;

The `>>` _appends_ the content to the file; this would work if the file didn't
exist (before `touch`'ing it) but if there happened to be a file already, it
would append the content to the existing file, resulting in invalid JSON.

e.g. running it twice (which may be accidentally);

```bash
sudo echo '{"allowedOrgs":["myorg"]}' >> "/Library/Application Support/com.docker.docker/registry.json"
sudo echo '{"allowedOrgs":["myorg"]}' >> "/Library/Application Support/com.docker.docker/registry.json"
```

Would result in;

```bash
sudo cat "/Library/Application Support/com.docker.docker/registry.json"
{"allowedOrgs":["myorg"]}
{"allowedOrgs":["myorg"]}
```

(which is invalid JSON)

The `sudo echo` also had some issues; the `sudo` only applied to the `echo`, and
not to the piped output (see https://unix.stackexchange.com/questions/1416/redirecting-stdout-to-a-file-you-dont-have-write-permission-on);

```bash
$ sudo touch "/Library/Application Support/com.docker.docker/registry.json"
$ sudo echo '{"allowedOrgs":["myorg"]}' > "/Library/Application Support/com.docker.docker/registry.json"
bash: /Library/Application Support/com.docker.docker/registry.json: Permission denied
```

Instead, using `tee` to run as privileged process, similar to the approach taken
in https://github.com/docker/docker.github.io/blame/c33d39a7ba80ad86e3d6f1bd38f21e530d58819e/engine/install/debian.md#L106

With that, the `touch` wouldn't be needed (the `mkdir` still would though);

```suggestion
$ sudo mkdir -p "/Library/Application Support/com.docker.docker"
$ echo '{"allowedOrgs":["myorg"]}' | sudo tee "/Library/Application Support/com.docker.docker/registry.json"
```

The new instructions will always overwrite the content with the new content,
replacing what's already there (that may need a small warning / note though) to
prevent the invalid JSON if someone runs the command multiple times, or if they
have to update the existing config with a new one.

Finally, some verification steps were added to verify content and permissions.

### DIFF
--- a/_includes/configure-registry-json.md
+++ b/_includes/configure-registry-json.md
@@ -72,7 +72,7 @@ On Mac, you can use the following methods to create a `registry.json` file.
 
 To automatically create a registry.json file when installing Docker Desktop, download `Docker.dmg` and run the following commands in a terminal from the directory containing `Docker.dmg`. Replace `myorg` with your organization's name.
 
-```bash
+```console
 $ sudo hdiutil attach Docker.dmg 
 $ sudo /Volumes/Docker/Docker.app/Contents/MacOS/install --allowed-org=myorg
 $ sudo hdiutil detach /Volumes/Docker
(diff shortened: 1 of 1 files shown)


thaJeztah:
@jerae-duffin @craig-osterhout PTAL 🤗

_includes/configure-registry-json.md
```diff
@@ -72,22 +72,40 @@ On Mac, you can use the following methods to create a `registry.json` file.
...
+```console
+$ sudo cat "/Library/Application Support/com.docker.docker/registry.json"
+{"allowedOrgs":["myorg"]}
 ```
 
-This creates the `registry.json` file at `/Library/Application Support/com.docker.docker/registry.json` and includes the organization information the user belongs to. Make sure this file can't be edited by the user, only by the administrator.
+Verify that the file has the expected permissions (`-rw-r--r--`) and ownership
+(`root` and `admin`):
+
+```bash
```
thaJeztah, line 106: derp. this should be `console` as well (to prevent the prompt from being copied)

thaJeztah: fixed

thaJeztah:
New version looks like;

Screenshot 2022-07-27 at 22 25 19

And mouse-selection works correctly without selecting the prompt (`$`);

Screenshot 2022-07-27 at 22 25 31

craig-osterhout (approved):
LGTM. Thanks @thaJeztah

thaJeztah:
Thanks for reviewing! Let me get this one in as well

PR #15250, opened 2022-07-28
jekyll: enable sitemap and local search for dev environment

crazy-max:
closes https://github.com/docker/docker.github.io/pull/14784
also related to https://github.com/docker/docker.github.io/pull/15179 as well as follow-up https://github.com/docker/docker.github.io/pull/15153

To test changes for local search but also the sitemap in our lab environment we need to enable these features. Since Jekyll 4, sitemap generation as well as metadata.json is pretty well optimized and takes less than 1 second for our docs so I think we can enable them for all environments which will allow us to check if nothing is broken on netlify environments before merging PRs.

Stats for sitemap and local search disabled:

```
#14 34.53 Build Process Summary:
#14 34.53 
#14 34.53 | PHASE      |    TIME |
#14 34.53 +------------+---------+
#14 34.53 | RESET      |  0.0004 |
#14 34.53 | READ       | 19.5737 |
#14 34.53 | GENERATE   |  1.1804 |
#14 34.53 | RENDER     | 12.7393 |
#14 34.53 | CLEANUP    |  0.0185 |
#14 34.53 | WRITE      |  0.2298 |
#14 34.53 +------------+---------+
#14 34.53 | TOTAL TIME | 33.7421 |
#14 34.53
#14 34.53
#14 34.53 Site Render Stats:
#14 34.54
#14 34.54 | Filename                                           | Count |     Bytes |   Time |
#14 34.54 +----------------------------------------------------+-------+-----------+--------+
#14 34.54 | _layouts/docs.html                                 |   712 | 23595.86K |  6.711 |
#14 34.54 | _includes/header.html                              |   712 |  1780.88K |  4.597 |
#14 34.54 | _includes/breadcrumbs.html                         |   712 |   301.95K |  4.532 |
#14 34.54 | _includes/toc_pure_liquid.html                     |   699 |   566.29K |  1.466 |
#14 34.54 | _includes/cli.md                                   |   243 |  1488.15K |  0.266 |
#14 34.54 | _includes/read_time.html                           |   712 |    21.14K |  0.175 |
#14 34.54 | _includes/head.html                                |   713 |  3098.72K |  0.172 |
#14 34.54 | _includes/navigation.html                          |   712 |  1179.25K |  0.045 |
#14 34.54 | _layouts/redirect.html                             |  1059 |   484.21K |  0.043 |
#14 34.54 | sitemap.xml                                        |     1 |     1.12K |  0.029 |
[...]

PR #15251, opened 2022-07-28
jekyll: fix not expanded DOMAIN env

crazy-max:
follow-up https://github.com/docker/docker.github.io/pull/15250#issuecomment-1198060836

fix regression from https://github.com/docker/docker.github.io/pull/14685 where `DOMAIN` env var is not expanded, resulting with a broken sitemap: https://github.com/docker/docker.github.io/pull/15250#issuecomment-1198060836

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -53,7 +53,7 @@ if [ "${JEKYLL_ENV}" = "production" ]; then
   (
     set -x
     bundle exec jekyll build --profile -d ${TARGET} --config _config.yml,_config_production.yml
-    sed -i 's#<loc>/#<loc>https://${DOMAIN}/#' "${TARGET}/sitemap.xml"
+    sed -i "s#<loc>/#<loc>https://${DOMAIN}/#" "${TARGET}/sitemap.xml"
   )
 else
   (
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -64,7 +64,7 @@ else
   )
 fi
 find ${TARGET} -type f -name '*.html' | while read i; do
-  sed -i 's#\(<a[^>]* href="\)https://${DOMAIN}/#\1/#g' "$i"
+  sed -i "s#\(<a[^>]* href=\"\)https://${DOMAIN}/#\1/#g" "$i"
```
thaJeztah, line 67: Have we checked if the existing escapes need to be double-escaped now? (e.g. `\)` to be `\\)`) (escaping and quoting and shells are hard!)

crazy-max: Looks good:

```console
$ DOMAIN=https://docs.docker.com
$ sed "s#\(<a[^>]* href=\"\)${DOMAIN}/#\1/#g" <<END
<a href="https://docs.docker.com/desktop/" id="manuals">Manuals</a>
END
<a href="/desktop/" id="manuals">Manuals</a>
```

thaJeztah: Thanks! Sorry, probably should've tried locally 😂 I know I ran into `sed` + regex + escaping fun in the past, so thought; let me leave a quick comment

thaJeztah (approved):
LGTM, thanks!

Dockerfile
crazy-max: > I know I ran into sed + regex + escaping fun in the past,

Yes me too!

PR #15255, opened 2022-07-28
Update 05_persisting_data.md

slwhitman:
Windows bash users need to add a second `/` to the path or they will get the error `cat: 'C:/Program Files/Git/data.txt': No such file or directory` and `ls: cannot access 'C:/Program Files/Git/': No such file or directory`


usha-mandya:
@slwhitman Thanks for the PR. The changes suggested aren't required by all Windows bash users. Therefore, we cannot accept the changes suggested. Please feel to comment after I close.

PR #15259, opened 2022-07-29
DD4L: Replace download links with per-distro install instruction links

p1-0tr:
### Proposed changes

Currently it's easy for users to skip the per-distro installation steps,
since direct download links for our packages are presented first. This
can lead to unnecessary frustration, since DD will not install if
pre-requisite steps are not followed. So instead of presenting package
download links in the most prominent spot, point users to per-distro
installation instructions.

### DIFF
--- a/desktop/install/debian.md
+++ b/desktop/install/debian.md
@@ -49,10 +49,10 @@ Recommended approach to install Docker Desktop on Debian:
 
 1. Set up [Docker's package repository](../../engine/install/debian.md#set-up-the-repository). 
 
-2. Download latest DEB package from the [release](../release-notes.md) page.
+2. Download latest [DEB package](https://desktop.docker.com/linux/main/amd64/docker-desktop-4.11.0-amd64.deb?utm_source=docker&utm_medium=webreferral&utm_campaign=docs-driven-download-linux-amd64).
 
 3. Install the package with apt as follows:
-    
+
 ```console
 $ sudo apt-get update
 $ sudo apt-get install ./docker-desktop-<version>-<arch>.deb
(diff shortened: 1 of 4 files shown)


aevesdocker:
Thanks for PR @p1-0tr. Can we add the download packages (as a button) to the Ubuntu, Fedora and Debian pages, so users don't have to navigate to the release notes to install. 

Also, I'm not sure about displaying the links to the pages as buttons. I think it would be better to do something like:

'Follow the links below to view the distro-specific installation instructions:
- [Ubuntu](link)
- [Debian](link)
- [Fedora](link)'

We could still wrap it in a note to make it pop out from the text.

desktop/install/debian.md
```diff
@@ -49,10 +49,12 @@ Recommended approach to install Docker Desktop on Debian:
 
 1. Set up [Docker's package repository](../../engine/install/debian.md#set-up-the-repository). 
 
-2. Download latest DEB package from the [release](../release-notes.md) page.
+2. Download latest DEB package:
```
aevesdocker, line 52: Sorry @p1-0tr , that was a bad idea of mine - this doesn't look great. Let's just keep it as it was with the links

desktop/install/linux-install.md
```diff
@@ -8,27 +8,26 @@ redirect_from:
 
  This page contains information about system requirements, download URLs, and instructions on how to install and update Docker Desktop for Linux.
 
-> Download Docker Desktop for Linux packages
+> Follow the links below to view the distro-specific installation instructions:
```
aevesdocker, line 11: Yes, this looks better, thank you

desktop/install/debian.md
p1-0tr: Done :) I made them direct download links, but inline

aevesdocker (approved):
Thank you @p1-0tr!

PR #15262, opened 2022-07-29
build: add drivers guides

crazy-max:
follow-up https://github.com/docker/buildx/pull/1233

also fixed a broken link from upstream repo (buildx).


build/buildx/drivers/remote.md
```diff
@@ -0,0 +1,20 @@
+---
+title: "Remote driver"
```
jedevc, line 2: I'm happy with the note at the beginning noting this as a pre-release feature - but I think we should definitely try and make sure we don't do this a lot.

Could we maybe consider just dropping this page for now? Although, I'm not sure how we'd modify the index to drop it from the list.

jedevc (approved):

build/buildx/drivers/remote.md
crazy-max: Sure we can drop it, lgtm

crazy-max: Argh if we drop this page, the `index.md` is not aligned :sweat_smile:

crazy-max: Let's keep it for now, I will change the `info` note to a warning one.

jedevc: Sounds good, it should get resolved quite soon, in the next bulidx release.

build/buildx/drivers/remote.md
```diff
@@ -0,0 +1,21 @@
...
+keywords: build, buildx, driver, builder, remote
+fetch_remote:
+  line_start: 2
+  line_end: -1
+---
+
+> Technical preview
```
usha-mandya, line 9: > Technical preview
>
> Remote driver is currently available as a Technical Preview.
> You can build a remote driver from the upstream [Buildx repository](https://github.com/docker/buildx#building){: target="_blank" rel="noopener" class="_"} or use
> the following command to download and install an edge release of Buildx:

usha-mandya, line 9: Also, should this be **Beta** and not Technical Preview so we are consistent with our terminology? See https://deploy-preview-14950--docsdocker.netlify.app/release-lifecycle/.

crazy-max: Ah thanks yes!

crazy-max: Looking at the statement, "Can be discontinued without notice" I'm not sure this is the case for the remote driver as this will still be released but not yet available.

I think we could also have "Technical preview" in this page if it makes sense:

> **Technical preview** means a service or software that is under active development and in stages of testing. Its purpose is to establish a feedback loop, allowing customers to influence its development and direction.

But in the meantime I will use Beta terminology.

crazy-max:
@usha-mandya thanks foo the review. Let me know if it LGTY now.

build/buildx/drivers/remote.md
usha-mandya: Yes, generally, we should avoid forward-looking statements in Docs. I guess we can tag it as Beta for now and update this page to remove the Beta flag when the remote drivers features goes GA.

usha-mandya (approved):
LGTM

PR #15264, opened 2022-07-29
DD4L: Document file sharing implementation

p1-0tr:
### Proposed changes

Describe file sharing implementation on Linux

### DIFF
--- a/desktop/install/linux-install.md
+++ b/desktop/install/linux-install.md
@@ -32,6 +32,8 @@ To install Docker Desktop successfully, your Linux host must meet the following 
 
 - At least 4 GB of RAM.
 
+- Enable configuring ID mapping in user namespaces, see [File sharing](#file-sharing).
+
 Docker Desktop for Linux runs a Virtual Machine (VM). For more information on why, see [Why Docker Desktop for Linux runs a VM](linux-install.md#why-docker-desktop-for-linux-runs-a-vm).
 
 > **Note:**
@@ -267,6 +269,58 @@ Docker Desktop for Linux runs a Virtual Machine (VM) for the following reasons:
     The VM utilized by DD4L uses [`virtiofs`](https://virtio-fs.gitlab.io){:target="_blank" rel="noopener" class="_"}, a shared file system that allows virtual machines to access a directory tree located on the host. Our internal benchmarking shows that with the right resource allocation to the VM, near native file system performance can be achieved with virtiofs.
 
     As such, we have adjusted the default memory available to the VM in DD4L. You can tweak this setting to your specific needs by using the **Memory** slider within the **Settings** > **Resources** tab of Docker Desktop.
+
+## File sharing
+
+Docker Desktop for Linux uses [virtiofs](https://virtio-fs.gitlab.io/){:target="_blank" rel="noopener"}{:target="_blank" rel="noopener"} as the
+default (and currently only) mechanism to enable file sharing between the host
+and Docker Desktop VM. In order not to require elevated privileges, without
+unnecessarily restricting operations on the shared files, Docker Desktop runs
+the file sharing service (`virtiofsd`) inside a user namespace (see
+`user_namespaces(7)`) with UID and GID mapping configured. As a result Docker
+Desktop relies on the host being configured to enable the current user to use
+subordinate ID delegation. For this to be true `/etc/subuid` (see `subuid(5)`)
+and `/etc/subgid` (see `subgid(5)`) must be present. Docker Desktop only
+supports subordinate ID delegation configured via files. Docker Desktop maps the
+current user ID and GID to 0 in the containers. It uses the first entry
+corresponding to the current user in `/etc/subuid` and `/etc/subgid` to set up
+mappings for IDs above 0 in the containers.
+
+| ID in container | ID on host                                                                       |
+| --------------- | -------------------------------------------------------------------------------- |
+| 0 (root)        | ID of the user running DD (e.g. 1000)                                            |
+| 1               | 0 + beginning of ID range specified in `/etc/subuid`/`/etc/subgid` (e.g. 100000) |
+| 2               | 1 + beginning of ID range specified in `/etc/subuid`/`/etc/subgid` (e.g. 100001) |
+| 3               | 2 + beginning of ID range specified in `/etc/subuid`/`/etc/subgid` (e.g. 100002) |
+| ...             | ...                                                                              |
+
+If `/etc/subuid` and `/etc/subgid` are missing, they need to be created.
+Both should contain entries in the form -
+`<username>:<start of id range>:<id range size>`. For example, to allow the current user
+to use IDs from 100000 to 165535:
+
+```console
+$ grep "$USER" /etc/subuid >> /dev/null 2&>1 || (echo "$USER:100000:65536" | sudo tee -a /etc/subuid)
+$ grep "$USER" /etc/subgid >> /dev/null 2&>1 || (echo "$USER:100000:65536" | sudo tee -a /etc/subgid)
+```
+
+To verify the configs have been created correctly, inspect their contents:
+
+```console
+$ echo $USER
+exampleuser
+$ cat /etc/subuid
+exampleuser:100000:65536
+$ cat /etc/subgid
+exampleuser:100000:65536
+```
+
+In this scenario if a shared file is `chown`ed inside a Docker Desktop container
+owned by a user with a UID of 1000, it shows up on the host as owned by
+a user with a UID of 100999. This has the unfortunate side effect of preventing
+easy access to such a file on the host. The problem is resolved by creating
+a group with the new GID and adding our user to it, or by setting a recursive
+ACL (see `setfacl(1)`) for folders shared with the Docker Desktop VM.
 
 ## Where to go next
 


desktop/install/linux-install.md
```diff
@@ -268,6 +270,17 @@ Docker Desktop for Linux runs a Virtual Machine (VM) for the following reasons:
...
+
+In case `/etc/subuid` and `/etc/subgid` are missing they will need to be created. Both should contain entries in the following form `<username>:<start of id range>:<id range size>`. E.g. to allow a user whose username is `docker` to use IDs from 100000 to 165535:
+
+    sudo echo 'docker:100000:65536' > /etc/subuid
+    sudo echo 'docker:100000:65536' > /etc/subgid
```
thaJeztah, line 280: This may not work; the `sudo` applies to the `echo`, but not to the pipe; https://unix.stackexchange.com/questions/1416/redirecting-stdout-to-a-file-you-dont-have-write-permission-on

```bash
sudo echo 'bla' > /etc/bla
bash: /etc/bla: Permission denied

echo 'bla' | sudo tee /etc/bla
bla

cat /etc/bla
bla
```

We should also use a `console` typehint here, and add a prompt (`$`);

    ```console
    $ echo 'docker:100000:65536' | sudo tee /etc/subuid
    $ echo 'docker:100000:65536' | sudo tee /etc/subgid
    ```

desktop/install/linux-install.md
```diff
@@ -268,6 +270,17 @@ Docker Desktop for Linux runs a Virtual Machine (VM) for the following reasons:
...
+
+In case `/etc/subuid` and `/etc/subgid` are missing they will need to be created. Both should contain entries in the following form `<username>:<start of id range>:<id range size>`. E.g. to allow a user whose username is `docker` to use IDs from 100000 to 165535:
```
thaJeztah, line 277: might be good to use something like `myusername` instead of `docker` for the examples; using `docker` is easy to be confused for "this must be `docker`" (some special username), whereas `myusername` is more verbose on "fill in your username here"

desktop/install/linux-install.md
p1-0tr: Thanks :) I have a tendency to forget that `sudo` does not work through pipes and redirections

desktop/install/linux-install.md
p1-0tr: Good point, I was wondering what would be a good example name to use. I usually go with putting stuff to fill in in angle brackets, but lately I've ran into cases where that confused someone.

thaJeztah: Yes, it remains tricky; indeed, use `<FOO>`, and users may try it literally (shell complaining about piping). Use something `ALLCAPS` and users may now think it must be uppercase (and/or it's case-insensitive).

We need to come with some conventions, and document them somewhere in the docs ("here's the conventions we use (describe what terms, and how we use example/placeholder values).

desktop/install/linux-install.md
thaJeztah: Same here! 😂 I recalled a similar thing in our install docs, and just the other day there was a similar issue, so I had it fresh in mind; https://github.com/docker/docker.github.io/pull/15243

desktop/install/linux-install.md
p1-0tr: Yeah, that sounds like a good idea. It could also help avoid some confusing situations when doing support :)

desktop/install/linux-install.md
```diff
@@ -268,6 +270,19 @@ Docker Desktop for Linux runs a Virtual Machine (VM) for the following reasons:
...
 
+## File sharing
+
+Docker Desktop for Linux uses [virtiofs](https://virtio-fs.gitlab.io/) as the default (and currently only) mechanism to enable file sharing between the host and Docker Desktop VM. In order not to require elevated privileges, without unnecessarily restricting operations on the shared files, Docker Desktop runs the file sharing service (`virtiofsd`) inside a [user namespace](https://man7.org/linux/man-pages/man7/user_namespaces.7.html) with UID and GID mapping configured. As a result Docker Desktop will rely on the host to be configured to enable the current user to use subordinate ID delegation. For this to be true [/etc/subuid](https://man7.org/linux/man-pages/man5/subuid.5.html) and [/etc/subgid](https://man7.org/linux/man-pages/man5/subgid.5.html) have to be present (Docker Desktop currently only supports subordinate ID delegation configured via files). In case of both Docker Desktop will use the first entry corresponding to the current user, to set up mappings for IDs above 0, and map the current users ID and GID to 0.
```
thaJeztah, line 275: Don't have time right now to do a full review, but some quick blurbs that may help other reviewers;


- perhaps we should refer to the man pages using (e.g.) `subuid(5)`; something like "see `subuid(5)`"
- for the external links, we may want to add `{:target="_blank" rel="noopener"}` to open in a new tab/window
- if possible, try to hard-wrap these paragraphs to (~) 80-chars; it makes reviewing easier (point to a specific passage), and we _try_ to stick with that in most parts
- avoid future tense


Generally, future tense should be avoided as readers read these docs as "do this, and docker does this". Adding future tense adds some cognitive overload ("if I do this, then docker will (eventually? when?) do this"

e.g.;

> As a result Docker Desktop will rely on the host

Can be (perhaps "as a result" can be skipped as well to make it more to the point);

> Docker Desktop relies on the host ..

Similar to "currently" (which may lead to the question: "are changes planned? will this be added?"); the docs
reflect the current state of things, so we can avoid some of those, e.g.;

> (Docker Desktop currently only supports subordinate ID delegation configured via files) In case of both Docker Desktop will use the first entry corresponding to the current user, to set up mappings for IDs above 0, and map the current users ID and GID to 0.

could be;

> Docker Desktop only supports subordinate ID delegation configured via files. In case of both Docker Desktop uses the first entry corresponding to the current user to set up mappings for IDs above 0, and map the current users ID and GID to 0.

The second sentence is a bit confusing (will have to read it a couple of times I think);

- "In case of both" (both what?)
- "the first  entry ... -> to set up mappings for IDs above 0"

Wondering if we need a simple example or a table to just show it, e.g.

| User ID in container | User ID on host     |
|----------------------|---------------------|
| 0 (root)             | 100000 (100000 + 0) |
| 1                    | 100001 (100000 + 1) |
| etc..                |                     |

thaJeztah: Coincidentally, @tianon was asking me recently about this area; perhaps he has some ideas as well 😅

desktop/install/linux-install.md
```diff
@@ -268,6 +270,19 @@ Docker Desktop for Linux runs a Virtual Machine (VM) for the following reasons:
...
+
+In case `/etc/subuid` and `/etc/subgid` are missing they will need to be created. Both should contain entries in the following form `<username>:<start of id range>:<id range size>`. E.g. to allow a user whose username is `myusername` to use IDs from 100000 to 165535:
+
+```console
+$ echo 'myusername:100000:65536' | sudo tee /etc/subuid
+$ echo 'myusernamer:100000:65536' | sudo tee /etc/subgid
+```
+
+In this scenario if a shared file is `chown`ed inside a Docker Desktop container to be owned by a user with a UID of 1000 it will show up on the host as owned by a user with a UID of 100999. This has the unfortunate side effect of preventing easy access to such a file on the host. The problem can be resolved by creating a group with the new GID and adding our user to it, or by setting a recursive [ACL](https://man7.org/linux/man-pages/man1/setfacl.1.html) on folders shared with the Docker Desktop VM.
```
tianon, line 284: IMO this is unexpected enough (from a user perspective) that we should _really_ reconsider whether this is the default behavior we want and/or whether this is the _only_ behavior we want to support -- the two obvious alternatives I see (either of which I personally would opt for) are running just `virtiofs` privileged (small security boundary, by design) or running it entirely unprivileged so that all shared files are "stuck" being owned by the host user which I think matches the behavior on Windows and macOS better, right?

(Sorry if this conversation already happened / is already happening elsewhere! :see_no_evil:)

As a side note (and probably the reason @thaJeztah pulled me into a thread here), I chuckled a little at "In order not to require elevated privileges," because setting this up is exactly the thing that _did_ require elevated privileges for me. :sweat_smile:

p1-0tr: > As a side note (and probably the reason @thaJeztah pulled me into a thread here), I chuckled a little at "In order not to require elevated privileges," because setting this up is exactly the thing that _did_ require elevated privileges for me. 😅

Yeah this is unfortunate :( On most distros this will be the default configuration now, so in a good number of cases users shouldn't need to do anything to configure subordinate IDs.

> IMO this is unexpected enough (from a user perspective) that we should _really_ reconsider whether this is the default behavior we want and/or whether this is the _only_ behavior we want to support -- the two obvious alternatives I see (either of which I personally would opt for) are running just `virtiofs` privileged (small security boundary, by design) or running it entirely unprivileged so that all shared files are "stuck" being owned by the host user which I think matches the behavior on Windows and macOS better, right?

I do agree that introducing a behaviour different to both, the native engine and DD on other OSes, should be avoided.

Option 1 would require asking the user for the `root` password every time we start (or they add a share), which is not great UX-wise (and goes against the requirement to run `root`-less), and has potential to adversely affect the host environment.

As for option 2 - it would be optimal, but is still a couple releases away from being possible. @fredericdalleau is working on a "fake owner" driver to use with virtiofs on macOS, where currently changing shared file ownership from the VM does not work, and changing ownership on the host file system, in the way we do on Linux, isn't an option. Once that's done, my intention is to make it the default behaviour on Linux.

As a middle ground I am considering having DD assign ACLs to shared folders at the moment of bind mounting. It would decrease the inconvenience of dealing with the actual rights for some files being changed, but introduces yet another assumption about the host environment.

desktop/install/linux-install.md
```diff
@@ -268,6 +270,47 @@ Docker Desktop for Linux runs a Virtual Machine (VM) for the following reasons:
...
+| ...             | ...                                                                              |
+
+If `/etc/subuid` and `/etc/subgid` are missing, they will need to be created.
+Both should contain entries in the form -
+`<username>:<start of id range>:<id range size>`. E.g. to allow a user whose
+username is `myusername` to use IDs from 100000 to 165535:
+
+```console
+$ echo 'myusername:100000:65536' | sudo tee /etc/subuid
+$ echo 'myusernamer:100000:65536' | sudo tee /etc/subgid
```
fredericdalleau, line 304: typo : `myusernamer`, should it be `mygroupname` to disambiguate?

fredericdalleau: I wonder if overlapping ranges for different users could cause security issues.

desktop/install/linux-install.md
```diff
@@ -268,6 +270,48 @@ Docker Desktop for Linux runs a Virtual Machine (VM) for the following reasons:
...
+
+If `/etc/subuid` and `/etc/subgid` are missing, they will need to be created.
+Both should contain entries in the form -
+`<username>:<start of id range>:<id range size>`. E.g. to allow a user whose
+username is `myusername` and group name is `mygroupname` to use IDs from 100000
+to 165535:
+
+```console
+$ echo 'myusername:100000:65536' | sudo tee /etc/subuid
+$ echo 'mygropname:100000:65536' | sudo tee /etc/subgid
```
fredericdalleau, line 305: there is still a typo, other than that, LGTM

desktop/install/linux-install.md
```diff
@@ -268,6 +270,48 @@ Docker Desktop for Linux runs a Virtual Machine (VM) for the following reasons:
...
+```
+
+In this scenario if a shared file is `chown`ed inside a Docker Desktop container
+to be owned by a user with a UID of 1000 it will show up on the host as owned by
+a user with a UID of 100999. This has the unfortunate side effect of preventing
+easy access to such a file on the host. The problem can be resolved by creating
+a group with the new GID and adding our user to it, or by setting a recursive
+ACL (see `setfacl(1)`) for folders shared with the Docker Desktop VM.
```
fredericdalleau, line 313: ACL looks like interesting , are there filesystems where they are not supported?

p1-0tr: Yes, there are, and IIRC some filesystem drivers can be built with ACL support disabled. Though most full featured file systems (extX, btrfs, appfs, ...) should support ACLs by default.

desktop/install/linux-install.md
fredericdalleau: @p1-0tr I notice the external links disappeared in the latest patches, was it intended?

desktop/install/linux-install.md
fredericdalleau: Sorry I was wrong, /etc/subgid contains a username on first column

desktop/install/linux-install.md
p1-0tr: Yes, I went with just referring to the relevant man pages (so that users can look them up locally), rather than relying on an external source that stores them.

desktop/install/linux-install.md
p1-0tr: np :)

desktop/install/linux-install.md
```diff
@@ -268,6 +270,47 @@ Docker Desktop for Linux runs a Virtual Machine (VM) for the following reasons:
...
+unnecessarily restricting operations on the shared files, Docker Desktop runs
+the file sharing service (`virtiofsd`) inside a user namespace (see
+`user_namespaces(7)`) with UID and GID mapping configured. As a result Docker
+Desktop relies on the host being configured to enable the current user to use
+subordinate ID delegation. For this to be true `/etc/subuid` (see `subuid(5)`)
+and `/etc/subgid` (see `subgid(5)`) have to be present. Docker Desktop only
+supports subordinate ID delegation configured via files. In case of both,
+`/etc/subuid` and `/etc/subgid`, Docker Desktop uses the first entry
+corresponding to the current user, to set up mappings for IDs above 0, and map
+the current users ID and GID to 0.
```
fredericdalleau, line 287:
```suggestion
supports subordinate ID delegation configured via files. Docker Desktop maps
the current user ID and GID to 0 in the containers. The mapping specified in
`/etc/subuid` and `/etc/subgid` are used for IDs above 0.
```

desktop/install/linux-install.md
```diff
@@ -268,6 +270,47 @@ Docker Desktop for Linux runs a Virtual Machine (VM) for the following reasons:
 
     As such, we have adjusted the default memory available to the VM in DD4L. You can tweak this setting to your specific needs by using the **Memory** slider within the **Settings** > **Resources** tab of Docker Desktop.
 
+## File sharing
+
+Docker Desktop for Linux uses [virtiofs](https://virtio-fs.gitlab.io/) as the
+default (and currently only) mechanism to enable file sharing between the host
+and Docker Desktop VM. In order not to require elevated privileges, without
+unnecessarily restricting operations on the shared files, Docker Desktop runs
```
fredericdalleau, line 278:
```suggestion
and Docker Desktop VM. In order not to require elevated privileges without
unnecessarily restricting operations on the shared files, Docker Desktop runs
```

fredericdalleau (commented):
Gah more nits

desktop/install/linux-install.md
```diff
@@ -268,6 +270,47 @@ Docker Desktop for Linux runs a Virtual Machine (VM) for the following reasons:
...
+
+| ID in container | ID on host                                                                       |
+| --------------- | -------------------------------------------------------------------------------- |
+| 0 (root)        | ID of the user running DD (e.g. 1000)                                            |
+| 1               | 0 + beginning of ID range specified in `/etc/subuid`/`/etc/subgid` (e.g. 100000) |
+| 2               | 1 + beginning of ID range specified in `/etc/subuid`/`/etc/subgid` (e.g. 100001) |
+| 3               | 2 + beginning of ID range specified in `/etc/subuid`/`/etc/subgid` (e.g. 100002) |
+| ...             | ...                                                                              |
+
+If `/etc/subuid` and `/etc/subgid` are missing, they will need to be created.
```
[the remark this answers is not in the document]
p1-0tr: No we don't, it suffers from us not having a sensible mechanism to report errors from the go backend to the user. So unfortunately it will only appear in logs and DD will be stuck in a starting state.

desktop/install/linux-install.md
```diff
@@ -268,6 +270,47 @@ Docker Desktop for Linux runs a Virtual Machine (VM) for the following reasons:
...
+Both should contain entries in the form -
+`<username>:<start of id range>:<id range size>`. E.g. to allow a user whose
+username is `myusername` to use IDs from 100000 to 165535:
+
+```console
+$ echo 'myusername:100000:65536' | sudo tee /etc/subuid
+$ echo 'myusername:100000:65536' | sudo tee /etc/subgid
```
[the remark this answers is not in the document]
p1-0tr: I am on the fence about this one, since it makes the user even less likely to read the commands and change what they need i.e. substitute `myusername` with the username they need to configure.

desktop/install/linux-install.md
p1-0tr: You can have multiple entries per user, we'll only use the first one, that's important for the user to understand. I'll try to rephrase this.

desktop/install/linux-install.md
p1-0tr: "without unnecessarily restricting operations on the shared files", is an interjection here, so it should go in between commas?

desktop/install/linux-install.md
fredericdalleau: Ha ok, I didn't understand it this way.

desktop/install/linux-install.md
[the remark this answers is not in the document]
fredericdalleau: What about "$USER", the thing is that tee without `-a` will overwrite the file.

desktop/install/linux-install.md
fredericdalleau: It didn't sound right to me as an interjection, but ok don't bother too much for a comma

desktop/install/linux-install.md
```diff
@@ -268,6 +270,47 @@ Docker Desktop for Linux runs a Virtual Machine (VM) for the following reasons:
...
+and Docker Desktop VM. In order not to require elevated privileges, without
+unnecessarily restricting operations on the shared files, Docker Desktop runs
+the file sharing service (`virtiofsd`) inside a user namespace (see
+`user_namespaces(7)`) with UID and GID mapping configured. As a result Docker
+Desktop relies on the host being configured to enable the current user to use
+subordinate ID delegation. For this to be true `/etc/subuid` (see `subuid(5)`)
+and `/etc/subgid` (see `subgid(5)`) have to be present. Docker Desktop only
+supports subordinate ID delegation configured via files. Docker Desktop maps the
+current user ID and GID to 0 in the containers. And it uses the first entry
+of `/etc/subuid` and `/etc/subgid`, corresponding to the current user, to set up
```
fredericdalleau, line 286: nit: the interjection still don't sound right to me

desktop/install/linux-install.md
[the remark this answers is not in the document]
p1-0tr: Sure, that works :)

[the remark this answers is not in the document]
thaJeztah: Question is; do we _want_ these to be appended? The risk there is that someone runs it twice and ends up with a configuration they don't want. The instructions should at least contain a validation step afterwards to check permissions of the file and its content.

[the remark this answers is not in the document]
p1-0tr: That's a good question. On one hand the instruction is for when the files are missing, so overwriting would be ok. But were the user not to read that we do not want to overwrite an existing config. So, Fredo's original suggestion would work better. I do say above that we will use the first entry in each of those files, and explain their format, so the implied validation step would be to have a look at their contents, and see if they match that. I can re-state that and give an example below, to make it more explicit.

desktop/install/linux-install.md
```diff
@@ -268,6 +270,58 @@ Docker Desktop for Linux runs a Virtual Machine (VM) for the following reasons:
...
+| ...             | ...                                                                              |
+
+If `/etc/subuid` and `/etc/subgid` are missing, they will need to be created.
+Both should contain entries in the form -
+`<username>:<start of id range>:<id range size>`. E.g. to allow the current user
+to use IDs from 100000 to 165535:
+
+```console
+$ test -f /etc/subuid || (echo "$USER:100000:65536" | sudo tee /etc/subuid)
+$ test -f /etc/subgid || (echo "$USER:100000:65536" | sudo tee /etc/subgid)
```
fredericdalleau, line 304: I'd suggest `grep "$USER" /etc/subuid`

p1-0tr: Done, I am weary that it's growing into magic one liner, thereby decreasing the likelihood that users will be motivated to understand what's happening. But I guess with the verification step suggested by Sebastian, this should not be too much of an issue.

desktop/install/linux-install.md
```diff
@@ -135,7 +137,7 @@ Log out and log back in so that your group membership is re-evaluated.
     >
     > If you do not agree to the terms, the Docker Desktop application will close and  you can no longer run Docker Desktop on your machine. You can choose to accept the terms at a later date by opening Docker Desktop.
     {: .important}
-
+ps-dd4l-add-filesharing-docs
```
fredericdalleau, line 140: that's weird, other than that LGTM

p1-0tr: emmm, tbh I have no idea how it got there

p1-0tr: done :)

fredericdalleau (approved):
LGTM

desktop/install/linux-install.md
```diff
@@ -268,6 +270,58 @@ Docker Desktop for Linux runs a Virtual Machine (VM) for the following reasons:
 
     As such, we have adjusted the default memory available to the VM in DD4L. You can tweak this setting to your specific needs by using the **Memory** slider within the **Settings** > **Resources** tab of Docker Desktop.
 
+## File sharing
+
+Docker Desktop for Linux uses [virtiofs](https://virtio-fs.gitlab.io/) as the
```
aevesdocker, line 275:
```suggestion
Docker Desktop for Linux uses [virtiofs](https://virtio-fs.gitlab.io/){:target="_blank" rel="noopener"}{:target="_blank" rel="noopener"} as the
```

desktop/install/linux-install.md
```diff
@@ -268,6 +270,58 @@ Docker Desktop for Linux runs a Virtual Machine (VM) for the following reasons:
...
+default (and currently only) mechanism to enable file sharing between the host
+and Docker Desktop VM. In order not to require elevated privileges, without
+unnecessarily restricting operations on the shared files, Docker Desktop runs
+the file sharing service (`virtiofsd`) inside a user namespace (see
+`user_namespaces(7)`) with UID and GID mapping configured. As a result Docker
+Desktop relies on the host being configured to enable the current user to use
+subordinate ID delegation. For this to be true `/etc/subuid` (see `subuid(5)`)
+and `/etc/subgid` (see `subgid(5)`) have to be present. Docker Desktop only
+supports subordinate ID delegation configured via files. Docker Desktop maps the
+current user ID and GID to 0 in the containers. And it uses the first entry
```
aevesdocker, line 285:
```suggestion
current user ID and GID to 0 in the containers. It uses the first entry
```

desktop/install/linux-install.md
```diff
@@ -268,6 +270,58 @@ Docker Desktop for Linux runs a Virtual Machine (VM) for the following reasons:
...
+
+| ID in container | ID on host                                                                       |
+| --------------- | -------------------------------------------------------------------------------- |
+| 0 (root)        | ID of the user running DD (e.g. 1000)                                            |
+| 1               | 0 + beginning of ID range specified in `/etc/subuid`/`/etc/subgid` (e.g. 100000) |
+| 2               | 1 + beginning of ID range specified in `/etc/subuid`/`/etc/subgid` (e.g. 100001) |
+| 3               | 2 + beginning of ID range specified in `/etc/subuid`/`/etc/subgid` (e.g. 100002) |
+| ...             | ...                                                                              |
+
+If `/etc/subuid` and `/etc/subgid` are missing, they will need to be created.
```
aevesdocker, line 297:
```suggestion
If `/etc/subuid` and `/etc/subgid` are missing, they need to be created.
```

desktop/install/linux-install.md
```diff
@@ -268,6 +270,58 @@ Docker Desktop for Linux runs a Virtual Machine (VM) for the following reasons:
...
+Both should contain entries in the form -
+`<username>:<start of id range>:<id range size>`. E.g. to allow the current user
```
aevesdocker, line 299:
```suggestion
`<username>:<start of id range>:<id range size>`. For example, to allow the current user
```

desktop/install/linux-install.md
```diff
@@ -268,6 +270,58 @@ Docker Desktop for Linux runs a Virtual Machine (VM) for the following reasons:
...
+to use IDs from 100000 to 165535:
+
+```console
+$ grep "$USER" /etc/subuid >> /dev/null 2&>1 || (echo "$USER:100000:65536" | sudo tee -a /etc/subuid)
+$ grep "$USER" /etc/subgid >> /dev/null 2&>1 || (echo "$USER:100000:65536" | sudo tee -a /etc/subgid)
+```
+
+to verify the configs have been created correctly, inspect their contents:
```
aevesdocker, line 307:
```suggestion
To verify the configs have been created correctly, inspect their contents:
```

desktop/install/linux-install.md
```diff
@@ -268,6 +270,58 @@ Docker Desktop for Linux runs a Virtual Machine (VM) for the following reasons:
...
+$ echo $USER
+exampleuser
+$ cat /etc/subuid
+exampleuser:100000:65536
+$ cat /etc/subgid
+exampleuser:100000:65536
+```
+
+In this scenario if a shared file is `chown`ed inside a Docker Desktop container
+to be owned by a user with a UID of 1000, it will show up on the host as owned by
```
aevesdocker, line 319:
```suggestion
owned by a user with a UID of 1000, it shows up on the host as owned by
```

desktop/install/linux-install.md
```diff
@@ -268,6 +270,58 @@ Docker Desktop for Linux runs a Virtual Machine (VM) for the following reasons:
...
+a user with a UID of 100999. This has the unfortunate side effect of preventing
+easy access to such a file on the host. The problem can be resolved by creating
```
aevesdocker, line 321:
```suggestion
easy access to such a file on the host. The problem is resolved by creating
```

desktop/install/linux-install.md
```diff
@@ -268,6 +270,58 @@ Docker Desktop for Linux runs a Virtual Machine (VM) for the following reasons:
...
+a group with the new GID and adding our user to it, or by setting a recursive
```
aevesdocker, line 322: Should it be '..the new GID and adding **your** user to it...'?

desktop/install/linux-install.md
```diff
@@ -268,6 +270,58 @@ Docker Desktop for Linux runs a Virtual Machine (VM) for the following reasons:
 
     As such, we have adjusted the default memory available to the VM in DD4L. You can tweak this setting to your specific needs by using the **Memory** slider within the **Settings** > **Resources** tab of Docker Desktop.
 
+## File sharing
+
+Docker Desktop for Linux uses [virtiofs](https://virtio-fs.gitlab.io/) as the
+default (and currently only) mechanism to enable file sharing between the host
+and Docker Desktop VM. In order not to require elevated privileges, without
```
aevesdocker, line 277:
```suggestion
and Docker Desktop VM. To remove the need for elevated privileges, without
```

desktop/install/linux-install.md
```diff
@@ -268,6 +270,58 @@ Docker Desktop for Linux runs a Virtual Machine (VM) for the following reasons:
...
+unnecessarily restricting operations on the shared files, Docker Desktop runs
+the file sharing service (`virtiofsd`) inside a user namespace (see
+`user_namespaces(7)`) with UID and GID mapping configured. As a result Docker
+Desktop relies on the host being configured to enable the current user to use
+subordinate ID delegation. For this to be true `/etc/subuid` (see `subuid(5)`)
+and `/etc/subgid` (see `subgid(5)`) have to be present. Docker Desktop only
```
aevesdocker, line 283:
```suggestion
and `/etc/subgid` (see `subgid(5)`) must be present. Docker Desktop only
```

aevesdocker (commented):
Some small picky changes @p1-0tr

aevesdocker:
is this ready to be merged now @p1-0tr ?

desktop/install/linux-install.md
p1-0tr: I changed it to "To avoid requiring". "remove" sounded to me like we had required elevated privileges previously, which is not the case. The current choice of implementation was more about avoiding an alternative that would, even though the choice comes with its own complications.

p1-0tr:
> is this ready to be merged now @p1-0tr ?

Yes :)

PR #15268, opened 2022-07-29
Add GitHub Actions annotations for linting

jedevc:
### Proposed changes

This PR introduces annotations for the failed linters - these are more easily human readable summaries of the errors encountered during the run, instead of requiring digging manually through the logs.

[For example](https://github.com/jedevc/docker.github.io/actions/runs/2761986305):

image

and inline with the Files Changed tab:

image

This should hopefully make the linter a little less painful to deal with, and should hopefully avoid needing to manually bring up the linting issues in reviews.

### Notes

The markdown linter is nice and easy - it has dedicated JSON output, with exact line numbers and everything.

The link checker is not so great. It only produces text output, which we need to parse with regex, but since it's run over the resulting html in `_site`, we can't easily work out the original markdown file and line to raise the issue on. For now, the extraction of the filename and line number occurs, but we deliberately comment them out since they won't work as expected.

### Related issues

This PR is dependent on #15227 to actually fix the linting first.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -45,9 +45,56 @@ jobs:
         name: Validate
         uses: docker/bake-action@v2
         with:
-          targets: ${{ matrix.target }}
+          targets: ${{ matrix.target }}-output
           set: |
+            *.cache-to=type=gha,scope=validate-${{ matrix.target }},mode=max
+            *.cache-from=type=gha,scope=validate-${{ matrix.target }}
             *.cache-from=type=gha,scope=build
+      -
+        name: Annotate
+        uses: actions/github-script@v6
+        with:
+          script: |
+            const fs = require('fs');
+            const results = fs.readFileSync('lint/results', 'utf-8');
+            if (results.length == 0) {
+              process.exit(0);
+            }
+
+            // print results
+            console.log(results);
+            process.exitCode = 1;
+
+            // construct annotations by parsing output
+            switch ("${{ matrix.target }}") {
+            case "htmlproofer":
+              const re = /^- (.+)\n  \*  (.+) \(line (\d+)\)\n(.+)$/gm;
+              while (true) {
+                const result = re.exec(results);
+                if (result === null) {
+                  break;
+                }
+
+                core.error(`${result[2]}\n${result[4]}`, {
+                  title: 'Link check failed',
+                  // file: result[1],
+                  // startLine: result[3],
+                });
+              }
+              break;
+            case "mdl":
+              const jsonResults = JSON.parse(results);
+              for (const result of jsonResults) {
+                const title = result.rule + (result.aliases.length > 0 ? ` (${result.aliases[0]})` : ``);
+                console.log(`${result.filename}:${result.line}; ${title} - ${result.description}`);
+                core.error(result.description, {
+                  title: title,
+                  file: result.filename,
+                  startLine: result.line,
+                });
+              }
+              break;
+            }
 
   # build-releaser job will just build _releaser app used for Netlify and
   # AWS deployment in publish workflow. it's just to be sure it builds correctly.
--- a/docker-bake.hcl
+++ b/docker-bake.hcl
@@ -41,10 +41,25 @@ target "htmlproofer" {
   output = ["type=cacheonly"]
 }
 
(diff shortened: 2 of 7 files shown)


.github/workflows/build.yml
```diff
@@ -26,14 +26,10 @@ jobs:
             *.cache-from=type=gha,scope=build
             *.cache-to=type=gha,scope=build,mode=max
 
-  validate:
```
crazy-max, line 29: Actually I think we could keep the matrix in the `validate` job to reduce workflow overhead and also keep required GitHub checks valid:

```yaml
  validate:
    runs-on: ubuntu-20.04
    strategy:
      fail-fast: false
      matrix:
        target:
          - htmlproofer
          - mdl
    steps:
      -
        name: Checkout
        uses: actions/checkout@v3
      -
        name: Set up Docker Buildx
        uses: docker/setup-buildx-action@v2
      -
        name: Validate
        uses: docker/bake-action@v2
        with:
          targets: ${{ matrix.target }}
          set: |
            *.cache-to=type=gha,scope=validate-${{ matrix.target }},mode=max
            *.cache-from=type=gha,scope=validate-${{ matrix.target }}
            *.cache-from=type=gha,scope=build
      -
        name: Annotate
        uses: actions/github-script@v6
        with:
          script: |
            const fs = require('fs');
            const results = fs.readFileSync('lint/results', 'utf-8');
            console.log(results);

            if ("${{ matrix.target }}" == "htmlproofer") {
              const re = /^- (.+)\n  \*  (.+) \(line (\d+)\)\n(.+)$/gm;
              while (true) {
                const result = re.exec(results);
                if (result === null) {
                  break;
                }
                core.error(`${result[2]}\n${result[4]}`, {
                  title: 'Link check failed',
                  // file: result[1],
                  // startLine: result[3],
                })
              }
              if (results.length > 0) {
                process.exitCode = 1
              }
            } else if ("${{ matrix.target }}" == "mdl") {
              for (const result of results) {
                core.error(result.description, {
                  title: result.rule + (result.aliases.length > 0 ? ` (${result.aliases[0]})` : ``),
                  file: result.filename,
                  startLine: result.line,
                })
              }
              if (results.length > 0) {
                process.exitCode = 1
              }
            }
```

jedevc: Agreed, that feels significantly better to me :+1:

.github/workflows/build.yml
```diff
@@ -45,9 +45,56 @@ jobs:
...
+            process.exitCode = 1;
+
+            // construct annotations by parsing output
+            switch ("${{ matrix.target }}") {
+            case "htmlproofer":
+              const re = /^- (.+)\n  \*  (.+) \(line (\d+)\)\n(.+)$/gm;
+              while (true) {
+                const result = re.exec(results);
+                if (result === null) {
+                  break;
```
crazy-max, line 75: I think it will break the switch but looks fine anyway.

jedevc: I think the `break` here should break out of the inner-most statement: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/break - at least that's my reading.

crazy-max: Ah indeed!

jedevc:
Marking as ready for review, but needs #15227 first :tada:

crazy-max (approved):
LGTM (putting back in draft to avoid merging by mistake)

usha-mandya (approved):

crazy-max (changes requested):
needs rebase

jedevc:
Have also added information on how to do the linting from within an IDE in this PR :tada:

docker-bake.hcl
```diff
@@ -41,12 +41,27 @@ target "htmlproofer" {
   output = ["type=cacheonly"]
 }
 
+target "htmlproofer-output" {
+  inherits = ["_common"]
+  target = "htmlproofer-output"
+  output = ["./lint"]
```
crazy-max, line 47: this folder should be added to `.gitignore`

crazy-max (approved):
LGTM thanks!

usha-mandya (approved):
LGTM

PR #15272, opened 2022-07-30
build: add docker driver guide

crazy-max:
follow-up https://github.com/docker/buildx/pull/1241

### DIFF
--- a/_data/toc.yaml
+++ b/_data/toc.yaml
@@ -1406,6 +1406,8 @@ manuals:
           section:
             - path: /build/buildx/drivers/
               title: Overview
+            - path: /build/buildx/drivers/docker/
+              title: Docker driver
             - path: /build/buildx/drivers/docker-container/
               title: Docker container driver
             - path: /build/buildx/drivers/kubernetes/
--- a/build/buildx/drivers/docker.md
+++ b/build/buildx/drivers/docker.md
@@ -1,0 +1,7 @@
+---
+title: "Docker driver"
+keywords: build, buildx, driver, builder, docker
+fetch_remote:
+  line_start: 2
+  line_end: -1
+---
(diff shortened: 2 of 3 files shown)


dockertopia (commented):
@crazy-max So docker driver content is coming from upstream while the other drivers are going to be in the drivers folder in the docs repo?

crazy-max:
> So docker driver content is coming from upstream while the other drivers are going to be in the drivers folder in the docs repo?

All drivers are from upstream: https://github.com/docker/buildx/tree/master/docs/guides/drivers

drivers folder in docs repo only contains stub files.

dockertopia:
> > So docker driver content is coming from upstream while the other drivers are going to be in the drivers folder in the docs repo?
> 
> All drivers are from upstream: https://github.com/docker/buildx/tree/master/docs/guides/drivers
> 
> drivers folder in docs repo only contains stub files.

I thought the idea was to consolidate build content in this doc set. Any reason we can't have this directly in the docs repo?

jedevc:
Keeping these kind of docs upstream in the buildx repo means that they live right next to the code that we work on, so we can make an effort to keep these up-to-date as we develop new features/work on bug fixes. Having them separated across repos means co-coordinating PRs, and will make it more likely to have out-of-date information.

I think it makes sense to have the high level information and such in the docker docs repo here, but then keep the buildx-specific/buildkit-specific content directly in their corresponding repositories, where users are already used to viewing that content.

dockertopia (approved):
From the perspective of the main docs base, it seems awkward to have so many exceptions regarding repo provenance for pages. It seems to be an easy answer for API documentation, but not as much for other topics such as this one. I'm new to these nuances, so although I appreciate your answer, I'm not entirely sure this is the way to go for build docs or any similar case. 
Since this PR won't change that (that was establishing by the previous PR for drivers) and just updates that info, I have no additional feedback for the specific scope of these changes. 
cc: @usha-mandya

PR #15273, opened 2022-07-30
hypen added

arslannoorpansota:


craig-osterhout:
Thanks for aligning the text with the command. It should all probably be updated to the v2 syntax without the hyphen.
The PR checks are stuck. Closing and re-opening to see if it fixes it.

craig-osterhout (approved):
LGTM to fix the consistency issue. As mentioned in previous comment, ideally should update all commands in the topic to newer syntax.

PR #15179, opened 2022-07-21
ci: add lab branch to publish workflow

crazy-max:
follow-up https://github.com/docker/docker.github.io/pull/15151

We are working with @VictorBersy to add a new environment to test advanced features on docs repo without impacting staging (https://docs-stage.docker.com/) and production (https://docs.docker.com/) environment.

This environment will be published to `https://docs-lab.docker.com/`.

I have created a lab branch where we are going to make some tests with https://github.com/docker/docker.github.io/pull/15151.

This PR adds the necessary configuration to our publish workflow to allow manual invocations on lab branch (see last commit).

Signed-off-by: CrazyMax <crazy-max@users.noreply.github.com>

PR #15276, opened 2022-07-31
remove most of absolute urls

crazy-max:
follow-up https://github.com/docker/docker.github.io/pull/15179 and https://github.com/docker/docker.github.io/pull/15250

Some links are broken in our lab environment because of absolute URLs being used and the `DOCS_URL` env var not being taken into account properly.

This PR removes most of the absolute URLs and use local `.md` files when available, otherwise a relative URL.


compose/release-notes.md
```diff
@@ -2113,8 +2113,8 @@ naming scheme accordingly before upgrading.
     You don't have to use it - your existing Compose files will run on Compose
     1.6 exactly as they do today.
 
-    Check the upgrade guide for full details:
-    https://docs.docker.com/compose/compose-file#upgrading
```
crazy-max, line 2117: This URL was not valid. We could not catch it with `htmlproofer` because not solved as link.

dockertopia: That's a line in the file from 2016, probably there were changes in `compose-file` affecting this. Did we have htmlproofer then?
You say "DOCS_URL env var not being taken into account properly.", how should we take this var into account? 
I'm unaware of this topic. cc: @usha-mandya

crazy-max: > You say "DOCS_URL env var not being taken into account properly.", how should we take this var into account?
> I'm unaware of this topic. cc: @usha-mandya

`DOCS_URL` has been added in https://github.com/docker/docker.github.io/pull/15250 to avoid hardcoded URLs that would not match the environment (stage vs lab vs prod). See for example the source of https://docs-stage.docker.com/ and you will see some URLs with `docs.docker.com` which is not right and should be `docs-stage.docker.com`.

tldr; this is an internal enhancement that should not impact your workflow.

desktop/previous-versions/edge-releases-mac.md
```diff
@@ -2287,7 +2287,7 @@ events or unexpected unmounts.
 
 **Bug fixes and minor changes**
 
-* Documentation moved to [https://docs.docker.com/desktop/mac/](../index.md)
+* Documentation moved to [{{ site.docs_url }}/desktop/mac/](../index.md)
```
usha-mandya, line 2290: Do we need `{{ site.docs_url }}` here? Wondering if it should be the same as other relative URLs on this page?

usha-mandya:
@crazy-max Added a minor comment on one of the Desktop pages. Looks good otherwise. Thank you!

desktop/previous-versions/edge-releases-mac.md
crazy-max: I think here it's a implicitly showing the absolute url but agree that `/desktop/mac/` is enough, will fix that.

crazy-max:
@usha-mandya Should be good, thanks for the review

compose/release-notes.md
dockertopia: Got it, so to get right base URL according with environment. Thanks for the explanation.

usha-mandya (approved):
LGTM. Thank you

PR #15283, opened 2022-08-01
Publish updates from master

aevesdocker:


dockertopia (changes requested):
For some reason this PR is picking up both approved and PRs requiring review.
We should look into this before publishing. Thanks.

dockertopia (approved):
So these are missing the green check icon (for approval) but looking into the PR they are approved and merged. In light of that removing the request for changes. LGTM
image

crazy-max:
> So these are missing the green check icon (for approval)

Green check icon does not mean there has been an approval or not but GitHub checks passed for this commit (aka GitHub Actions). As this PR has been merged as merge commit GitHub Actions will only validate the last commit (`d2fca0e`).

As long as it has been merged on master, it means each changes has been approved first through a PR which is protected for `master` and `published` branches with the following settings:

image

dockertopia:
Thanks for the clarification @crazy-max. 
When you hover a PR link you can see this;
 
image

There was  the doubt it if the check alongside the commit hash was for having passed the checks or approvals because as it is the exact same icon (screen grab above).

crazy-max:
@dockertopia Yes agree that the UI is misleading :disappointed:

PR #15288, opened 2022-08-01
Update Deactivate Accounts

jerae-duffin:
Updated steps for account or organization

### DIFF
--- a/docker-hub/deactivate-account.md
+++ b/docker-hub/deactivate-account.md
@@ -15,17 +15,13 @@ Before deactivating your Docker Hub account, please complete the following:
 
 2. If you belong to any organizations, remove your account from all of them.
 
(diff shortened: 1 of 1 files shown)


craig-osterhout (dismissed):
LGTM

amaderic:
For deactivating an Organization, we want to say:
"If you have an active subscription, downgrade it to the Docker Free Team subscription"

amaderic:
For deactivating an Account we want to say:
"If you have an active subscription, downgrade it to the Docker Personal subscription."

amaderic:
I also think this has been removed from Hub so we should omit it in docs: 
"If you have an enterprise license, download the key.
In Docker Hub, navigate to Your Account > Accounts Settings > Licenses. The download link will no longer be available after your account is disabled."

PR #15291, opened 2022-08-02
new release notes layout

crazy-max:
Adds a new utility called `release-notes` that I find more appealing and declarative for the end user, starting with build one. This utility can be reused by other projects. we might need to tweak a bit some fields to make it more generic and cover all use cases (cc @thaJeztah @usha-mandya):

image


_scss/_release-notes.scss
```diff
@@ -0,0 +1,54 @@
+.release-date {
```
mark-dr, line 1: I think there's work we could do in this file to keep things closer to the design system, but I think it's probably most efficient to let this in as it stands and then include it in a docs-wide theming review in the future.

crazy-max: Awesome! Let me know if you need anything!

_plugins/filters.rb
```diff
@@ -0,0 +1,11 @@
+module Jekyll
+  module CustomFilters
+    SPACE = ' '.freeze
+
+    def newline_to_space(input)
+      input.to_s.gsub(/\n/, SPACE)
+    end
```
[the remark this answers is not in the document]
crazy-max: > Couldn't you solve that in the YAML, by changing the `|` to a `>`?

Works fine with `>`

mark-dr (dismissed):
Looks great to me, nice UI design, really nice that it uses yaml data, awesome work! :+1:

crazy-max:
Made some changes to have a more generic release-notes layout so it can be used across our website:

* We can define components with their respective entries like Docker Desktop release notes does (For all platforms, For Mac, ...).
* Option to enable/disable GitHub link to full release notes
* Multiple tags support and allow arbitrary ones
* Header support to add arbitrary content before listing entries for each release
* New `upgrade` and `security` tag

I also split each release in its dedicated yaml definition. This way we could reuse this yaml file directly in Docker Desktop to display release notes for current version by just fetching it.

Here is an example with Docker Desktop 4.12 and 4.11: https://deploy-preview-15291--docsdocker.netlify.app/desktop/release-notes/

image

PR #15294, opened 2022-08-02
releaser: simplify access to env or secrets

crazy-max:
To avoid dealing with `cat` in our Dockerfile and also be able to retrieve secrets as env or file can be useful. We encounter that issue while working on lab branch as `AWS_SESSION_TOKEN` needs to be defined for this environment (OIDC auth) but can be optional for prod and stage.

### DIFF
--- a/_releaser/main.go
+++ b/_releaser/main.go
@@ -2,6 +2,8 @@ package main
 
 import (
 	"log"
+	"os"
+	"path/filepath"
 
 	"github.com/alecthomas/kong"
 )
@@ -28,4 +30,16 @@ func main() {
 			Summary: true,
 		}))
 	ctx.FatalIfErrorf(ctx.Run())
+}
+
+// getEnvOrSecret retrieves secret's value from secret file or env
+func getEnvOrSecret(name string) string {
+	if v, ok := os.LookupEnv(name); ok {
+		return v
+	}
+	b, err := os.ReadFile(filepath.Join("/run/secrets", name))
+	if err != nil {
+		return ""
+	}
+	return string(b)
 }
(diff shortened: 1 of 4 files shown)


_releaser/main.go
```diff
@@ -29,3 +30,19 @@ func main() {
 		}))
 	ctx.FatalIfErrorf(ctx.Run())
 }
+
+// getEnvOrSecret retrieves secret's value from secret file or env
+func getEnvOrSecret(name string) string {
+	if v, ok := os.LookupEnv(name); ok {
+		return v
+	}
+	secretPath := "/run/secrets/" + name
```
thaJeztah, line 39: I guess "in practice" this won't be an issue, but perhaps if we want to do it "correct";


```suggestion
	secretPath := filepath.Join("/run/secrets", name)
```

thaJeztah (commented):
one minor nit, but otherwise SGTM

_releaser/main.go
crazy-max: Done thanks

_releaser/main.go
```diff
@@ -29,3 +31,19 @@ func main() {
...
 	ctx.FatalIfErrorf(ctx.Run())
 }
+
+// getEnvOrSecret retrieves secret's value from secret file or env
+func getEnvOrSecret(name string) string {
+	if v, ok := os.LookupEnv(name); ok {
+		return v
+	}
+	secretPath := filepath.Join("/run/secrets", name)
+	if _, err := os.Stat(secretPath); err != nil {
```
thaJeztah, line 41: Guess this one is redundant, as we ignore any error below anyway 😁

crazy-max: Ah yes good catch, done.

thaJeztah (approved):
LGTM

PR #15322, opened 2022-08-05
Updated networking.md to remove sent word used twice

NishantBanat:
There was a small error where we see sent sent, so removed one of the sent word. Line 395.


dockertopia (approved):
Thank, @NishantBanat.

PR #15350, opened 2022-08-10
Addition of a 'Contribute' section with our Style Guide

aevesdocker:


contribute/style/grammar.md
```diff
@@ -0,0 +1,159 @@
...
+## Acronyms and initialisms
+
+An acronym is an abbreviation you would speak as a word, for example, ROM (for read only memory). Other examples include radar and scuba, which started out as acronyms but are now considered words in their own right.
+
+An initialism is a type of acronym that comprises a group of initial letters used as an abbreviation for a name or expression. If you were using the acronym in a spoken conversation, you would enunciate each letter: H-T-M-L for Hypertext Markup Language.
+
+### Best practice:
+
+- Spell out lesser-known acronyms or initialisms on first use, then follow with the acronym or initialism in parentheses. After this, throughout the rest of your page or document, use the acronym or initialism alone.
+> ‘You can use Single-Sign-on (SSO) to sign in to Notion. You may need to ask your administrator to enable SSO.’
```
dvdksn, line 17: Question: when we introduce an acronym in parentheses, do we capitalize the spelled-out word, as is done here? Even though we wouldn't normally? (single sign-on)

contribute/style/grammar.md
```diff
@@ -0,0 +1,159 @@
...
+
+## Bolds and italics
+
+Unless you're referring to Ul text or user-defined text, you should not add emphasis to text. Clear, front-loaded wording makes the subject of a sentence clear.
+
+### Best practice:
+
+- Don't use bold to refer to a feature name.
+- Use italics sparingly, as this type of formatting can be difficult to read in digital experiences.
+Notable exceptions are titles of books, newspapers or magazines.
```
dvdksn, line 31: More relevant examples?

```suggestion
Notable exceptions are titles of articles, blog posts, or specification documents.
```

contribute/style/grammar.md
```diff
@@ -0,0 +1,159 @@
...
+The following content elements should use sentence case:
+
+- Titles of webinars and events
+- Headings and subheadings in all content types
+- Calls to action
+- Headers in boxed text
+- Column and row headers in tables
+- Links
+- Sentences (of course)
+- Anything in the Ul including navigation labels, buttons, headings
```
dvdksn, line 46:
```suggestion
- Anything in the Ul including navigation labels, buttons, headings. Provided that the corresponding UI label uses sentence case; follow the capitalization style of the UI element.
```

contribute/style/grammar.md
```diff
@@ -0,0 +1,159 @@
...
+
+## Lists
+
+Lists are a great way to break down complex ideas and make them easier to read and scan.
+
+### Best practice:
+
+- Bulleted lists should contain relatively few words or short phrases. For most content types, limit the number of items in a list to five.
+- Don’t add commas (,) or semicolons (;) to the ends of list items.
+    - Some content types may use bulleted lists that contain 10 items, but it's preferable to break longer lists into several lists, each with its own subheading or introduction.
```
dvdksn, line 96: Remove indentation

```suggestion
- Some content types may use bulleted lists that contain 10 items, but it's preferable to break longer lists into several lists, each with its own subheading or introduction.
```

contribute/style/grammar.md
```diff
@@ -0,0 +1,159 @@
...
+
+### Commas
+
+- Use the serial or Oxford comma - a comma before the coordinating conjunction (and, or) in a list of three or more things.
+- If a series contains more than three items or the items are long, consider a bulleted list to improve readability.
+
+### Dashes and hyphens
+
+- Use the em dash (-) sparingly when you want the reader to pause, to create parenthetical statements, or to emphasize specific words or phrases. Always put a space on either side of the em dash.
+- Use an en dash (-) to indicate spans of numbers, dates, or time.
```
dvdksn, line 155: Finally I know which one to use where 👍🏻 :-D

contribute/style/recommended-words.md
```diff
@@ -0,0 +1,103 @@
...
+Use `earlier` when talking about version numbers.
+
+Use:
+
+_In Docker Desktop 4.1 and earlier._
+
+Instead of:
+
+_In Docker Desktop 4.1 and lower._
+
```
dvdksn, line 54:
```suggestion

#### execute

Avoid where possible. Use `run` if applicable.

```

dvdksn (commented):
Great work Allie!!

_data/toc.yaml
```diff
@@ -15,6 +15,9 @@ horizontalnav:
 - title: Samples
   path: /samples/
   node: samples
+- title: Contribute
+  path: /contribute/overview/
+  node: contribute
   # TODO unify navbbar links: homepage currently has a custom "articles" link
   hide_home: true
```
dvdksn, line 22: I don't think we should hide the contribute section from the landing page. But, it looks like that would also require some CSS changes, too. The home page already contains the additional "Articles" link, and adding contribute as well makes it rather wide. The media query looks like it's set to collapse the nav menu at 768px, which is narrower than the width of the navbar if we also add contribute.

Easy solution would be to remove the Articles (blog) link. Do we really need it? :D

dvdksn:
Also we might want to update CONTRIBUTING.md, and just put a link to the contribution section in there. I think that's better than removing the file entirely.

contribute/style/grammar.md
aevesdocker: This is actually a mistake in my draft. It should be single-sign on. Good spot! Will amend.

dvdksn (approved):
image

Issue #13553, opened 2021-09-17
404 at /desktop/mac/previous-versions/undefined


baz974:
I found a broken link : https://docs.docker.com/desktop/mac/previous-versions/undefined
I arrived on this page through: https://docs.docker.com/desktop/mac/previous-versions/

thaJeztah:
Thanks for reporting; we received a couple of reports for this URL, but so far I haven't been able to find the broken link itself; do you know what link you clicked on that page?

craig-osterhout:
From #15745

> Yandex translator apparently blocked the ability to download the dock gave an error shas directly everything swings.

usha-mandya:
@craig-osterhout Wondering if we can close this one?

usha-mandya:
Discussed with @thaJeztah. One option we discussed is to redirect the `.../undefined` pages to the main Docker Desktop release notes page where users can find the latest Desktop downloads. Based on our understanding of this issue, it only occurs when users arrive at a translated version of the page. I don't think we need to update our JS to fix this. Also, as we want our users to be on the latest version of Docker Desktop, redirection might be a better option.

PR #15352, opened 2022-08-10
Move Release Notes in TOC for consistency

dockertopia:
Moved Release Notes to last place in section's TOC for consistency in Engine and Compose.

### DIFF
--- a/_data/toc.yaml
+++ b/_data/toc.yaml
@@ -1339,9 +1339,6 @@ manuals:
       title: Install binaries
   - path: /engine/install/linux-postinstall/
     title: Optional post-installation steps
-  - path: /engine/release-notes/
-    title: Release notes
-  - sectiontitle: Previous versions
     section:
     - path: /engine/release-notes/19.03/
       title: Engine 19.03 release notes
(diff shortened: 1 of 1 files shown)


_data/toc.yaml
```diff
@@ -1339,9 +1339,6 @@ manuals:
       title: Install binaries
   - path: /engine/install/linux-postinstall/
     title: Optional post-installation steps
-  - path: /engine/release-notes/
-    title: Release notes
-  - sectiontitle: Previous versions
     section:
     - path: /engine/release-notes/19.03/
       title: Engine 19.03 release notes
```
thaJeztah, line 1344: You need to move the section itself as well (looks like currently only the section title was moved)

thaJeztah: Ohh.... interesting; so because of this, the TOC shows "undefined", and now I wonder if something like this is related to https://github.com/docker/docker.github.io/issues/13553

Screenshot 2022-08-11 at 14 23 36

dockertopia: Now we know what happens :). So the section was there but the indentation was off.

thaJeztah (approved):
LGTM, thanks!

PR #15370, opened 2022-08-12
Add link to official registry image in local mirror documentation.

Zandor300:
### Proposed changes

I was confused why the image wasn't referenced until I realized the image is called "registry". Adding a link will make it easier to understand.


dvdksn (approved):
Thanks a lot for helping us improve the documentation! 🐳

PR #15375, opened 2022-08-12
mark explicitly published port as string to avoid user confusion

glours:
Signed-off-by: Guillaume Lours <guillaume.lours@docker.com>

### Proposed changes
As string can be unquoted in Yaml, our example may confused user and let them assume that the `ports.published` attribut can be an integer. 

I explicitly wrapped the value with quotes to remove ambiguity 

### Related issues (optional)
https://github.com/docker/compose/issues/9306

### DIFF
--- a/compose/cli-command-compatibility.md
+++ b/compose/cli-command-compatibility.md
@@ -54,7 +54,7 @@ services:
     ports:
     - mode: ingress
       target: 80
-      published: 80
+      published: "80"
       protocol: tcp
 networks:
   default:


thaJeztah:
Did that change? Because it definitely was an integer before; if integers are no longer accepted that is a big breaking change; https://github.com/docker/cli/blob/master/cli/compose/schema/data/config_schema_v3.9.json#L226

jsoriano:
> Did that change?

It changed in docker compose 2.3.0, see https://github.com/docker/compose/issues/9306.

compose/cli-command-compatibility.md
```diff
@@ -54,7 +54,7 @@ services:
     ports:
     - mode: ingress
       target: 80
```
jsoriano, line 56: Is there a reason why `target` port is a number, but `published` port is a string?

glours: because you can use the range syntax to defined published port as we already mentioned in the [compose issue](https://github.com/docker/compose/issues/9306#issuecomment-1212185763)

glours:
@thaJeztah introduced in compose-go with this [PR](https://github.com/compose-spec/compose-go/pull/230), new versions of Compose can handle both, the issue comes when an older than `v2.3.0` version of Compose try to parse the `ports.published` port as an integer when it's a string (and `docker compose config` generate a string by default since version `v2.3.0`)
So no issue with the CLI, Compose handle that properly

compose/cli-command-compatibility.md
jsoriano: Ah ok, `target` cannot be a range, right? Thanks!

dockertopia (commented):
@milas, I have some questions about aligning compose file specification taking this PR into account, could you please take a look at this PR?

thaJeztah:
> and docker compose config generate a string by default since version v2.3.0)

could `config` produce an integer (without quotes) if it's only a number (and only fall-back to using a string if it was actually a port range?)

milas:
I'm not totally sure that changing this example will help ease confusion.

I think we should:
* Clarify in the Compose spec that `published` can be of type `number | string` with explanation of port ranges
  * For parsing, implementations MUST support string for both ranges and single ports
  * For serializing, can be implementation dependent
* Update `compose-go` to use `number` when single port, and `string` when range

I think it's clear from the linked issue that, ignoring compatibility issues, users find the "single port as a string in `published` while single port as a number in `target`" behavior undesirable, and given that `docker/compose` already supports both on _input_ and the spec is vague, aligning things around the least surprising behavior makes the most sense here IMO

PR #15381, opened 2022-08-13
Fix: remove unsupported Font Awesome icons from toggle theme

AntonyOnScript:
### Proposed changes
removed -o suffix from theme toggle icons, which was making it being empty. see the before x after my changes

image
image
### Related issues (optional)


aevesdocker (approved):
Thank you for the PR @AntonyOnScript

PR #15386, opened 2022-08-15
Fix dead link for Docker Support

tnir:
As `https://support.docker.com` looks unavailable any more, we can use `https://docs.docker.com/support/` instead.


dvdksn (approved):
Thanks!

PR #15392, opened 2022-08-16
build: multi-stage builds

crazy-max:
carry a commit from #14905 to move "multi stage build" page to `Build section > Building images`.

### DIFF
--- a/_data/toc.yaml
+++ b/_data/toc.yaml
@@ -128,8 +128,6 @@ guides:
       title: Dockerfile best practices
     - path: /develop/develop-images/build_enhancements/
       title: Build images with BuildKit
-    - path: /develop/develop-images/multistage-build/
-      title: Use multi-stage builds
     - path: /develop/develop-images/image_management/
       title: Manage images
     - path: /develop/develop-images/baseimages/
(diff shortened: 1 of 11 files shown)


jedevc:
Hm having a section called "guides" in the "manuals" feels confusing.

Since this relates to Dockerfiles, I wonder if we should have a `Dockerfile` section under `Docker Build`? This would be a good first page to put there, but I can easily imagine lots of other content ending up there, like our best practices.

crazy-max:
> Since this relates to Dockerfiles, I wonder if we should have a `Dockerfile` section under `Docker Build`?

That SGTM but what about our guides in https://github.com/docker/buildx/tree/master/docs/guides? I recall you had some ideas for a folder under `Docker Build` section for them.

jedevc:
> but what about our guides in https://github.com/docker/buildx/tree/master/docs/guides? I recall you had some ideas for a folder under Docker Build section for them

Good point :laughing: I think they probably belong in the Buildx section directly, or maybe in an Advanced section under that, if we really start to get too many. I'm not quite sure, I think some of those guides, probably need some love and care before we put them on the site, especially some of the older ones, just to bring them into line with the style of the ones we've already got there.

crazy-max:
@jedevc Moved to "Building images" section and rebased. PTAL.

jedevc (approved):

usha-mandya (approved):
LGTM

PR #15406, opened 2022-08-17
ci: remove publish workflow

crazy-max:
follow-up #15151

if tests on temporary envs https://docs-stage2.docker.com and https://docs2.docker.com are successful we can switch to the `deploy` workflow.

needs first to switch DNS for `docs-stage2.docker.com` to `docs-stage.docker.com` and `docs2.docker.com` to `docs.docker.com` (cc @VictorBersy)

keeping in draft for now.


crazy-max:
We can close this one as everything is already in place!

PR #15420, opened 2022-08-18
Word Change

gaoagong:
### Proposed changes
Changed word "simple" to "simply" as it is correct grammar.


craig-osterhout (approved):
Thanks, @gaoagong. LGTM.

PR #15429, opened 2022-08-19
Update develop.md

developer-freed:
When learning docker, when I start the container according to the example, mongodb will report a connection error, when I catch the error the container runs fine


craig-osterhout:
@developer-freed, thanks for contributing. This was updated based on an earlier issue.

Issue #12236, opened 2021-02-03
Please add "docker builder prune" to pruning section


truboxl:
[config/pruning.md](https://docs.docker.com/config/pruning/)

It is useful to be included into the list of items for pruning if you do a lot experimentation on building Dockerfile

https://docs.docker.com/engine/reference/commandline/builder_prune/

May also help clarifying https://github.com/docker/docker.github.io/issues/9738 by comparing pruning images and pruning layers

Issue #15205, opened 2022-07-25
Object pruning guide does not show how to query reclaimable sizes and does not mention build cache


wabain:
[config/pruning.md](https://docs.docker.com/config/pruning/)

Currently the pruning guide has at least two notable omissions: it does not mention the `docker system df` command, which allows one to see what memory can be reclaimed by pruning, and it does not mention `docker builder prune`, which is one of the subcommands aggregated into `docker system prune` (#12236 has already been raised for this).

In my case I had more than 100GB on a development machine which had been dedicated to dangling build cache. As it stands the guide did not help me to identify or remediate that storage usage (except through `docker system prune`, which would also prune stopped containers), although it seems like it would be exactly the right place to document the steps to follow.

PR #15437, opened 2022-08-19
config/pruning: add documentation on `docker builder prune`

RyanSquared:
### Proposed changes

This adds a section documenting the `docker builder prune` command in the pruning documentation.

### Related issues (optional)

Related: #15205
Fixes: #12236

### DIFF
--- a/config/pruning.md
+++ b/config/pruning.md
@@ -141,6 +141,40 @@ Other filtering expressions are available. See the
 [`docker network prune` reference](../engine/reference/commandline/network_prune.md)
 for more examples.
 
+## Prune build cache
+
+The `docker builder prune` command allows you to clean the build cache. By
+default, `docker builder prune` only cleans up build cache objects not attached
+to an image. To remove unused build cache objects:
+
+```console
+$ docker builder prune
+WARNING! This will remove all dangling build cache. Are you sure you want to continue? [y/N] y
+```
+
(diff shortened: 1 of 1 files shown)


jedevc (approved):

RyanSquared:
Is there anything else required before this can be merged?

jedevc:
LGTM, PTAL @crazy-max @thaJeztah

crazy-max (commented):
Fine to me but I think this is out of scope for the `Run you app in production` section. This should live in [Build section](https://docs.docker.com/build/) at some point or moved to [`builder prune` command](https://docs.docker.com/engine/reference/commandline/builder_prune/) reference.

RyanSquared:
I'm kinda not certain about that, because on one hand, you acre correct in that it is out of scope for running applications in production. However, I have come across this page several times when trying to clean up space using Docker, by using a search engine and looking up something like "docker clean up space", "docker remove images", and even "docker prune" (it's the second article for this one! just after `docker system prune`, which I skipped, as I confused it with `docker swarm` / `docker mcahine). The headline is appealing to people using search engines who just want an answer, so I believe this is the best place to put it, even if it may not be entirely relevant to people reading the manual in a linear fashion.

Maybe a similar section can be added to the builder section, but I think the SEO for this page is too valuable to not have such a critical piece of information missing, even if it's not *entirely* relevant to the chapter as a whole.

PR #15217, opened 2022-07-26
Add Fig as an autocomplete method

brendanfalk:
### Proposed changes

[Fig](https://fig.io) provides autocomplete for 300+ CLI tools including Docker, so we would love to have Fig listed as an autocomplete method in your docs. Thanks so much and please let me know if you have any questions!

PR #15444, opened 2022-08-22
Removing the completion.md file

dockertopia:
### Proposed changes
When autocompletion became unavailable in v2, this file was removed from the TOC. But the file is being found by search and users are trying to execute instructions on it or suggesting changes for this page. 

### Related issues (optional)
Influences answer to https://github.com/docker/docker.github.io/pull/15217.


dockertopia:
So removing the page for now was the suggested approach so far.
Another approach might be to keep it and add a note that it only be available for v1. This approach would also require the mentioned scripts be available for download at least for the versions the feature does exist (isolating global versions that are valid or that are valid for each context).

dockertopia:
Reviewed this with the whole team during today's sync. The consensus is to remove this page for now and then recover it once we have completed work to recover this feature.

aevesdocker (approved):

aevesdocker (approved):

PR #15467, opened 2022-08-23
[ENGDOCS-829] Add pull image to hub quickstart

craig-osterhout:
### Proposed changes

As part of re-organizing the Hub Repositories topic, we noticed info about pushing images exists in [Hub quickstart](https://docs.docker.com/docker-hub/) and [Repositories](https://docs.docker.com/docker-hub/repos/#pushing-a-docker-container-image-to-docker-hub), but info about pulling images does not exist in the Docker Hub topics.

Since we plan to re-organize the Repository topic and remove push, it doesn't make sense to add pull there. Instead, added to [Hub quickstart](https://docs.docker.com/docker-hub/).

### Related issues (optional)
ENGDOCS-829


jerae-duffin (approved):
Thanks Craig! LGTM

PR #15473, opened 2022-08-24
DD overview page improvements

aevesdocker:
Looking to make the DD overview page more dynamic

Note: Image icons will most likely need to change to align more with the Docker colours


_data/toc.yaml
```diff
@@ -1162,6 +1162,8 @@ manuals:
       title: Quick Start Guide and sign in
     - sectiontitle: Explore Docker Desktop
       section:
+        - path: /desktop/use-desktop/overview/
+          title: Overview
```
thaJeztah, line 1166: Can we make this `/desktop/use-desktop/` (without the `/overview/` suffix?) That way there's a logical hierarchy in the URLs;

- `/desktop/use-desktop/` (overview)
     - -> `/desktop/use-desktop/container/`
     - -> `/desktop/use-desktop/images/`
     - -> `/desktop/use-desktop/volumes/`
     - etc.


With that, I can be on a "details" page, and move "one level up" in the URL path, and arrive at the overview for the "explore" section; move one level up, and arrive at the overview for the "desktop" section, etc.

desktop/use-desktop/overview.md
```diff
@@ -0,0 +1,22 @@
+---
+description: Docker Dashboard
+keywords: Docker Dashboard, manage, containers, gui, dashboard, images, user manual
+title: Overview
+redirect_from:
+- /desktop/dashboard/
```
thaJeztah, line 6: See my comment above; this page would then be renamed to `index.md` (to be the index page for the section)

Also note that there's other redirects related to dashboard that were missing, and got added back in https://github.com/docker/docker.github.io/pull/15498; some of those may have to be moved as well

desktop/index.md
```diff
@@ -30,50 +34,85 @@ Docker Desktop includes:
...
 
-For information about new features, improvements, and bug fixes in Docker Desktop releases, see [Release notes](release-notes.md).
+<div class="component-container">
+    <!--start row-->
+    <div class="row">
+      <div class="col-xs-12 col-sm-12 col-md-12 col-lg-4 block">
+        <div class="component">
+             <div class="component-icon">
+                 <img src="/images/download-docker.svg" alt="Docker Desktop for Mac" width="70" height="70">
+             </div>
```
thaJeztah, line 64: I wonder if we can include these elements in our stylesheets; from the homepage / landing page, it looks like we already have some icons that we use for "download", "reference", etc., so possibly it would be good to re-use the same icons/elements for consistency;

Screenshot 2022-08-28 at 19 39 09

dvdksn:
I like the idea for using tabs on this page, it feels more dynamic than static headings. But I feel like we should probably do something to make it stand out more. Right now they are a bit anonymous. For example, we could adding borders to the tab items, or changing the background color:

Screenshot 2022-08-29 at 11 16 54

aevesdocker:
Thank you for the review @thaJeztah. Apologies it's taken a while to respond - other priorities took over and this PR got moved further down my to-do list. 

I have incorporated you suggestion in https://github.com/docker/docker.github.io/pull/15589. I had to create a new PR as there were lots of merge conflicts created when Kevin moved all the images to the assets folder. 

Talking of images/icons - I have replaced the icons with new ones that now match our Design System and work in both light and dark mode. I worked with one of the designers who picked more appropriate icons.  I'm not sure I agree with your point about reusing icons for consistency. I used the ones from the homepage just as placeholders. I feel that these are very specific to the homepage and we should use more demure icons in the rest of our docs.

PR #15477, opened 2022-08-24
Document legacy build vs BuildKit differences

### LINKED ISSUE #12231: Recommend Buildkit for using multi-stage builds
[develop/develop-images/multistage-build.md](https://docs.docker.com/develop/develop-images/multistage-build/)

When people started off using Docker a couple of versions ago, their default is set to not use Buildkit. However, Buildkit is required to properly execute `-t release` stages. Without Buildkit, other stages of the same Dockerfile are executed as well although they shouldn't.

Please add a hint that the described features are only properly usable with Buildkit enabled.

andreisilviudragnea:
### Proposed changes

I documented how legacy engine executes all stages of a multi-stage Dockerfile from the beginning of the Dockerfile up to the target stage, even if target stage does not depend on some intermediate stages. When BuildKit is enabled, this (possible bug) does not happen and only transitive dependency stages of the target stage get executed.

I also documented that BuildKit is not enabled by default on all platforms (it is on macOS, but not on Linux) and that multi-stage builds should always be executed with BuildKit enabled in order to avoid the problem described above.

### Related issues

Fixes #12231

### DIFF
--- a/build/building/multi-stage.md
+++ b/build/building/multi-stage.md
@@ -206,13 +206,13 @@ RUN g++ -o /binary source.cpp
 
 Multi-stage build syntax was introduced in Docker Engine 17.05.
 
-## Moby vs BuildKit engine differences
+## Differences between legacy build and BuildKit
 
-- With Moby engine, all stages from the beginning of the Dockerfile up to the specified stage are executed, even if 
-some stages are not dependencies of the target stage.
-- With BuildKit, only the transitive dependencies of the target stage are executed.
+The legacy Docker Engine builder processes all stages of a Dockerfile leading up to the selected `--target`. It will build a stage even if the selected target doesn't depend on that stage.
 
-For example, check the following Dockerfile:
+BuildKit only builds the stages that the target stage depends on.
+
+For example, given the following Dockerfile:
 
 ```dockerfile
 # syntax=docker/dockerfile:1
@@ -226,12 +226,10 @@ FROM base AS stage2
 RUN echo "stage2"
 ```
 
-- Output for BuildKit enabled:
+With BuildKit enabled, building the `stage2` target in this Dockerfile means only `base` and `stage2` are processed. There is no dependency on `stage1`, so it's skipped.
 
 ```console
 $ DOCKER_BUILDKIT=1 docker build --no-cache -f Dockerfile --target stage2 .
-```
-```console
 [+] Building 0.4s (7/7) FINISHED                                                                                                                                                                                
  => [internal] load build definition from Dockerfile                                                                                                                                                       0.0s
  => => transferring dockerfile: 36B                                                                                                                                                                        0.0s
@@ -246,12 +244,10 @@ $ DOCKER_BUILDKIT=1 docker build --no-cache -f Dockerfile --target stage2 .
  => => writing image sha256:f55003b607cef37614f607f0728e6fd4d113a4bf7ef12210da338c716f2cfd15                                                                                                               0.0s
 ```
 
-- Output for BuildKit disabled:
+On the other hand, building the same target without BuildKit results in all stages being processed:
 
 ```console
 $ DOCKER_BUILDKIT=0 docker build --no-cache -f Dockerfile --target stage2 .
-```
-```console
 Sending build context to Docker daemon  219.1kB
 Step 1/6 : FROM ubuntu AS base
  ---> a7870fd478f4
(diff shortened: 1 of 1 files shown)


andreisilviudragnea:
@craig-osterhout When will this be reviewed and merged?

usha-mandya:
@andreisilviudragnea Could you rebase the PR?

andreisilviudragnea:
@usha-mandya Done.

usha-mandya:
Thank you @andreisilviudragnea. One of our team members will take a look and get back to you.

build/building/multi-stage.md
```diff
@@ -205,3 +205,83 @@ RUN g++ -o /binary source.cpp
...
+FROM base AS stage2
+RUN echo "stage2"
+```
+
+- Output for BuildKit enabled:
+
+```console
+$ DOCKER_BUILDKIT=1 docker build --no-cache -f Dockerfile --target stage2 .
+```
+```console
```
dvdksn, line 234:
```suggestion
```

build/building/multi-stage.md
```diff
@@ -205,3 +205,83 @@ RUN g++ -o /binary source.cpp
...
+ => [base 2/2] RUN echo "base"                                                                                                                                                                             0.1s
+ => [stage2 1/1] RUN echo "stage2"                                                                                                                                                                         0.2s
+ => exporting to image                                                                                                                                                                                     0.0s
+ => => exporting layers                                                                                                                                                                                    0.0s
+ => => writing image sha256:f55003b607cef37614f607f0728e6fd4d113a4bf7ef12210da338c716f2cfd15                                                                                                               0.0s
+```
+
+- Output for BuildKit disabled:
```
dvdksn, line 249:
```suggestion
On the other hand, building the same target without BuildKit results in all stages being processed:
```

build/building/multi-stage.md
```diff
@@ -205,3 +205,83 @@ RUN g++ -o /binary source.cpp
...
+
+```console
+$ DOCKER_BUILDKIT=0 docker build --no-cache -f Dockerfile --target stage2 .
+```
+```console
```
dvdksn, line 254:
```suggestion
```

build/building/multi-stage.md
```diff
@@ -205,3 +205,83 @@ RUN g++ -o /binary source.cpp
...
+stage2
+Removing intermediate container bbc025b93175
+ ---> 09fc3770a9c4
+Successfully built 09fc3770a9c4
+```
+
+`stage1` gets executed when BuildKit is disabled, even if `stage2` does not depend on it.
+
+BuildKit is not enabled by default on all platforms. For example, while it is enabled on macOS, it is
+NOT on Linux.
```
dvdksn, line 283: 1) on what platforms is it enabled by default? (how about Windows?)
2) how can I check if I am using BuildKit by default?

build/building/multi-stage.md
```diff
@@ -205,3 +205,83 @@ RUN g++ -o /binary source.cpp
...
+
+Docker multi-stage builds should always be run with
+[BuildKit enabled](/develop/develop-images/build_enhancements/)
+to avoid the problem described above.
```
dvdksn, line 287:
```suggestion
Always run multi-stage builds with
[BuildKit enabled](/develop/develop-images/build_enhancements/)
for better performance.
```

dvdksn (changes requested):
I think saying "moby vs buildkit" is confusing, in part since buildkit is also a moby project 😅 I think I would prefer if we could refer to it as "legacy vs buildkit", as we do on [this page](https://docs.docker.com/develop/develop-images/build_enhancements/)

build/building/multi-stage.md
jedevc: 1. I think it's enabled by default on Docker Desktop, but not on our Linux packages - the default should be switching to BuildKit everywhere (with buildx :tada:) in the upcoming release of 22.06, see https://github.com/docker/cli/pull/3676.
2. I don't actually know :scream_cat: I know you can look at the configuration file for your installation, e.g. `/etc/docker/daemon.json`, and see if `buildkit` is enabled under the `features` key.

dvdksn: OK, gotcha! Let me suggest another edit based on this info.

dvdksn, line 283: Given that we are about to make BuildKit the default in the near future, we can simplify to this:

```suggestion
BuildKit is enabled by default if you use Docker Desktop.
```

build/building/multi-stage.md
andreisilviudragnea: While I agree with the better performance part, it also changes the semantics of the build.

This bug showed up in our build when enabling BuildKit on a Linux installation. Do you think this should be logged as a bug too? Is it a bug or a feature?

dvdksn (approved):
lgtm, thanks a lot @andreisilviudragnea

PR #15487, opened 2022-08-25
Install page improvement based on feedback

dockertopia:
Included workshop w/ PM feedback 
Remove Docker Desktop page and renamed instructions page and added redirects
Adjusted TOC
Made pages less wordy when possible and other multiple minor enhancements.


compose/install/index.md
```diff
@@ -3,30 +3,73 @@ description: How to install Docker Compose
 keywords: compose, orchestration, install, installation, docker, documentation
 title: Install Docker Compose
 toc_max: 2
+redirect_from:
+- /compose/compose-desktop/
 ---
 
-On this page you can find a summary of the available options for installing Docker Compose. 
+On this page you can find information on how to get and install Compose.
```
hyu, line 10:
```suggestion
```

I think this page is short and self-evident enough not to require this explanation.

compose/install/index.md
```diff
@@ -3,30 +3,73 @@ description: How to install Docker Compose
...
 
-## Compose prerequisites
+## Install Compose
 
-* Docker Compose requires Docker Engine.
-* Docker Compose plugin requires Docker CLI.
+**To get Compose:**
```
hyu, line 14:
```suggestion
```

This seems redundant with the header, but doesn't add much additional info.

compose/install/index.md
```diff
@@ -3,30 +3,73 @@ description: How to install Docker Compose
...
 
-## Compose installation scenarios
-You can run Compose on macOS, Windows, and 64-bit Linux. Check what installation scenario fits your needs. 
+* If you have Docker Desktop, you've got a full Docker installation, including Compose.
 
-Are you looking to:
+* If you don't have Docker yet installed, or just need to install Compose, check the options below:
```
hyu, line 18:
```suggestion
```

I think the "If you don't have Docker yet installed" should be included with the prior sentence about Docker Desktop. So that way, we address DD entirely with the first point:

> If you've previously installed Docker Desktop, then you already have a full Docker installation that includes Compose. (If not, download links for Docker Desktop by OS are in the table below.)



Then, that leaves the statement: 
> If you…just need to install Compose, check the options below:

This could be more helpful. Under what circumstances would the user need to install Compose outside Docker Desktop? And how do they tell the difference between whether they should install it as a plugin, or standalone?

(I address this point a bit in the "Table Description" feedback below)

compose/install/index.md
```diff
@@ -3,30 +3,73 @@ description: How to install Docker Compose
...
-     + [Setting up Docker's repository](compose-plugin#install-using-the-repository) and using it to install the compose plugin package.     
-     + Other scenarios, check the [Linux install](compose-plugin#installing-compose-on-linux-systems).
-  + __Windows Server__: If you want to run the Docker daemon and client directly on Microsoft Windows Server, follow the [Windows Server install instructions](compose-plugin#install-compose-on-windows-server).
+**Table description:**
+* **Linux(64bit):** *option1)* Install [Docker Desktop for Linux](../../desktop/install/linux-install.md) or, *option2)* [Install Docker Engine and the CLI](../../engine/install/#server) and then [install Compose](compose-linux.md).
+* **macOS:** [Install Docker Desktop](../../desktop/install/mac-install/).
+* **Windows:** [Install Docker Desktop](../../desktop/install/windows-install/).
+* **Windows Server:** [Install Docker Engine and the CLI](../../engine/install/binaries/#install-server-and-client-binaries-on-windows) and then [install Compose](#install-compose-on-windows-server).
+
```
hyu, line 33: This "table description" seems redundant with the information that's already in the table.  

Is this intended for accessibility, though? If so, that is unfortunately inconsistent with other instances of table usage. (e.g., on the DD [Linux Install](https://docs.docker.com/desktop/install/linux-install/#supported-platforms) page). 😞 

I would suggest that we remove from this section the repeated Docker Desktop download links, since it detracts from the simplicity and clarity of the links already in the table.

Without those links, it leaves the instructions for Linux and Windows Server. How about including them as clear sections?


### Note on Linux

```suggestion
### Note on Linux

On Linux(64bit), you may want to install Docker Engine separately from Compose, instead of the approach of installing everything with Docker Desktop. [This gives you more control over your install location and version of Compose.]

 1. [Install Docker Engine and the CLI](../../engine/install/#server) 
 2. [install Compose](compose-linux.md).


Alternately, you may want to [install Compose standalone], without the pre-requisite of installing the Docker Engine. 


###  Note on Windows Server
(This then replaces "Install Compose on Windows Server")
```


> [This gives you more control over your install location and version of Compose.]

I just put this here as a placeholder for trying to explain a common use case for bypassing the Docker Desktop install. 

Is there more we can say here to address "Under what circumstances would the user need to install Compose outside Docker Desktop? And how do they tell the difference between whether they should install it as a plugin, or standalone?"

compose/install/index.md
```diff
@@ -3,30 +3,73 @@ description: How to install Docker Compose
...
-## Compose prerequisites
+## Install Compose
 
-* Docker Compose requires Docker Engine.
-* Docker Compose plugin requires Docker CLI.
+**To get Compose:**
 
-## Compose installation scenarios
-You can run Compose on macOS, Windows, and 64-bit Linux. Check what installation scenario fits your needs. 
+* If you have Docker Desktop, you've got a full Docker installation, including Compose.
```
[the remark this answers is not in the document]
hyu: @dockertopia, is there a simple check the user can do to confirm they have docker compose? They should just be able to run `docker compose` right?

hyu (commented):
I really like this approach with the table! I added a bunch of specific feedback on the index page. I will go through the Linux install page when I have a chance.

_data/toc.yaml
```diff
@@ -1441,12 +1441,10 @@ manuals:
...
       - path: /compose/install/
-        title: Install Docker Compose 
-      - path: /compose/install/compose-desktop/
-        title: Install Compose plugin through Desktop 
-      - path: /compose/install/compose-plugin/
-        title: Install Docker Compose Plugin 
-      - path: /compose/install/uninstall/ 
+        title: Install Docker Compose
+      - path: /compose/install/compose-linux/
+        title: Install on Linux
```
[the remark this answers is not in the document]
dockertopia: I'm not sure how this affects SEO, the keywords in the front matter are there to optimize searched. I just thought there was a lot of context repetition and, since, there is not a lot of real-estate in the TOC (panel on the left) and the titles wrap after a couple of words I thought the TOC looked cleaner that way. I am also fine either way. Your call.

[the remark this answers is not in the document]
hyu: Sure, let's go with the shorter title.

compose/install/index.md
dockertopia: Agree. Just keeping what we're doing as a standard in the docs. Usually it helps with the weirdness of having content start right after the page's title (from a visual perspective).

dockertopia:
To clarify I thought checking if there's anything of interest to pick up from here https://github.com/docker/docker.github.io/pull/15494 or go with this scenario instead, it not make sense to address each individual review comment in this PR. If the goal is to dismiss the other scenario and just fine tune this one and can address all your feedback in comments/commits. Let me know how you want to proceed.

compose/install/index.md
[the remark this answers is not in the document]
dockertopia: See closing comment.

compose/install/index.md
dockertopia: See closing comment.

compose/install/index.md
dockertopia: See closing comment.

compose/install/index.md
dockertopia: See closing comment.

dockertopia:
So we're continuing the work in https://github.com/docker/docker.github.io/pull/15494, so all comments and feedback from everyone is addressed in the mentioned PR.

PR #15494, opened 2022-08-26
Install page improvement based on feedback (scenario2)

dockertopia:
Exploring @thaJeztah's feedback on this. Here's a second scenario:
* Giving more prominence to DD.
* Trying to portray more Compose as something you get with Docker.
* Including @hyu's feedback.

### Proposed changes
From scenario 1 POV:
* Changing the page flow, 1) to Linux (only plugin), 2) other installs (standalone per system).  
* Including About DD image

### Related PR
Scenario 1: https://github.com/docker/docker.github.io/pull/15487


dockertopia:
@Han to be good on my word, I wanted to consider thaJezta's feedback in a new scenario. Ths leaves you to appreciate all the information I've presented with so far. Although I would love if we could conclude this discussion as soon as possible and move forwards with publishing a fully reviewed version of this in the next 1-2 work days.

compose/install/compose-linux.md
```diff
@@ -0,0 +1,120 @@
...
+2. Update the `apt` package index, and install the _latest version_ of Docker Compose:
+
+    > Or, if using a different distro, use the equivalent package manager instructions.
+
+    ```console
+    $ sudo apt-get update
+    $ sudo apt-get install docker-compose-plugin
+    ```
+
+    Alternatively, to install a specific version of the Compose CLI plugin:
```
thaJeztah, line 44: I wonder into how much detail we want to go for installing older versions. I know we added the equivalent sections in the "engine" installation instructions for production scenarios (where enterprise users may want to "pin" to specific versions), but as this section already tries to be "generic" ("or if using a different distro"), perhaps we should keep the installation also generic ("install latest stable release"). In the end, "how to install a specific version of a package" is more of a general ["know how to work with your system's package manager"](https://linuxopsys.com/topics/install-specific-version-package-apt) topic.

compose/install/compose-other.md
```diff
@@ -0,0 +1,65 @@
...
+
+title: Other install Scenarios
+---
+
+## Install Compose Standalone
+
+### On Linux
+> **Compose standalone**
+>
+> Note that Compose standalone uses the _dash compose_ syntax instead of current's standard syntax (_space compose_).
```
hyu, line 14: Could an example be helpful here? 

```suggestion
> Note that Compose standalone uses the _dash compose_ syntax instead of current's standard syntax (_space compose_). For example: Use `docker-compose up` for Compose standalone, instead of `docker compose up`
```

compose/install/compose-linux.md
```diff
@@ -0,0 +1,120 @@
...
+      ```console
+      $ sudo apt-get install docker-compose-plugin=<VERSION_STRING>
+      ```
+    where `<VERSION_STRING>` is, for example,`2.3.3~ubuntu-focal`.
+
+3.  Verify that Docker Compose is installed correctly by checking the version.
+
+    ```console
+    $ docker compose version
+    Docker Compose version v2.3.3
```
thaJeztah, line 70: Somewhat related to my previous comment; if we remove the "install a specific version" section, we could change this to `docker compose --help` as "verify" step (`docker compose --help` should work after installing). That way we can avoid having to update the version each time 😅 

Alternatively, we could make the `v2.3.3` an "illustrative" version `<version>`

compose/install/index.md
```diff
@@ -2,31 +2,45 @@
...
+### Pre-existing Docker Installation
+
+If you have a Docker installation on Linux to which you want to add the Compose plugin, you can choose one of these options:
+
+* [Install using the repository](./compose-linux.md#install-using-the-repository)
+* [Install the plugin manually](./compose-linux.md#install-the-plugin-manually)
+
+### Other install scenarios
+
+For other options check, [Other install scenarios](./compose-other.md/).
```
hyu, line 42: The "check," looks like a typo. Also, suggestion for a less redundant phrasing:

```suggestion
See [Other Install Scenarios](./compose-other.md/).
```

_data/toc.yaml
```diff
@@ -1439,12 +1439,12 @@ manuals:
...
-      - path: /compose/install/compose-desktop/
-        title: Install Compose plugin through Desktop 
-      - path: /compose/install/compose-plugin/
-        title: Install Docker Compose Plugin 
-      - path: /compose/install/uninstall/ 
+        title: Install Docker Compose
+      - path: /compose/install/compose-linux/
+        title: Install on Linux
+      - path: /compose/install/compose-other/
+        title: Other scenarios install
```
hyu, line 1446: To me, "Other install scenarios" reads a bit more familiar than "Other scenarios install"

```suggestion
        title: Other install scenarios
```

compose/install/compose-linux.md
```diff
@@ -0,0 +1,120 @@
...
+    ```
+
+> **Note**
+>
+> To run Compose as a non-root user, see [Manage Docker as a non-root user](../../engine/install/linux-postinstall.md){:target="_blank" rel="noopener" class="_"}.
```
thaJeztah, line 75: This is a general topic for using `docker` as a non-root user; as the page starts with the assumption users already have a functional docker installation, this wouldn't be needed for most of those cases. That said, perhaps we should include this as a link in an "next steps" / "where to go next" section at the bottom of the page.

compose/install/compose-linux.md
```diff
@@ -0,0 +1,120 @@
...
+
+
+### Install the plugin manually
+
+> **Note**
+>
+> This option requires you to manage upgrades manually. We recommend setting up Docker's repository for an easier maintenance.
```
thaJeztah, line 82: Not for this section perhaps, and I'm somewhat counter-arguing my "general knowledge about your package manager" comment 😂 but maybe a "how to upgrade compose" section would be good (effectively just the same steps as installing; `apt-get update && apt-get install docker-compose` to upgrade to the latest version).

hyu (approved):
Thanks, Dora! I approve these changes to the page. I added some comments that are non-blocking. (They are minor. And to avoid too much delay from async communication if any of my points are controversial, I want you to go ahead and merge without changes so the bias can be on moving quickly).

compose/install/index.md
```diff
@@ -2,31 +2,45 @@
...
-* Docker Compose requires Docker Engine.
-* Docker Compose plugin requires Docker CLI.
+If you have Docker Desktop, you've got a full Docker installation, including Compose.
 
-## Compose installation scenarios
-You can run Compose on macOS, Windows, and 64-bit Linux. Check what installation scenario fits your needs. 
+You can check this by clicking on **About Docker Desktop** from the Docker menu ![whale menu](../../desktop/images/whale-x.png){: .inline}.
 
-Are you looking to:
+![about-desktop](../images/about_desktop.png){:width="750px"}
```
thaJeztah, line 18: Have you checked if we already have an "about desktop" screenshot somewhere? I'm wondering if there's already one somewhere that we can reuse; perhaps it would be good to have this screenshot within the `/desktop/images/` directory to encourage it to be re-used.

My concern with this particular screenshot is that (because it shows "versions"), we may get tempted to updated it with each release (which we should avoid if possible); perhaps we should consider blurring the versions to make it more stand out as "illustrative", and not "here's the version you should see".

_data/toc.yaml
```diff
@@ -1439,12 +1439,12 @@ manuals:
...
-        title: Install Docker Compose 
-      - path: /compose/install/compose-desktop/
-        title: Install Compose plugin through Desktop 
-      - path: /compose/install/compose-plugin/
-        title: Install Docker Compose Plugin 
-      - path: /compose/install/uninstall/ 
+        title: Install Docker Compose
+      - path: /compose/install/compose-linux/
+        title: Install on Linux
+      - path: /compose/install/compose-other/
```
thaJeztah, line 1445: Could we name these pages;

- `linux` (`/compose/install/linux/`)
- `other` (`/compose/install/other/`)

It makes the URLs a bit shorter (and possibly more predictable), and prevents duplicating `compose` in the URL (we're in the compose section, so the URL already starts with `/compose/`).

_data/toc.yaml
```diff
@@ -1531,6 +1531,8 @@ manuals:
     section:
       - path: /docker-hub/publish/
         title: Overview
+      - path: /docker-hub/publish/insights-analytics/
+        title: Insights & analytics
```
thaJeztah, line 1535: Unrelated change? (Looks like this was not meant to be part of this PR)

compose/install/index.md
dockertopia: Adding see after the comma.

_data/toc.yaml
```diff
@@ -1439,12 +1439,12 @@ manuals:
   - sectiontitle: Install Compose
     section:
```
thaJeztah, line 1440: An alternative here could be to move the `/compose/install/` to the top level. That way getting to the `install` page would be a "one click", instead of two (click to expand the sub-menu, then click to go to `install`). The section could then be renamed to (e.g. "Installation per platform" - similar to "Installation per distro" in the engine manual). Most users (users with Desktop installed) would never have to look inside that section.

### Install page at the top level

Screenshot 2022-08-29 at 17 42 59

### .. and for those that really need the other install methods..

Screenshot 2022-08-29 at 17 43 07

Note: For consideration; I also renamed the "Overview of Docker Compose" and "Install Docker Compose" in the TOC to just "overview" and "install".  That approach would provide a consistent navigation in each of the manual ("overview", "install", ...)

Screenshot 2022-08-29 at 17 46 18
Screenshot 2022-08-29 at 17 46 26
Screenshot 2022-08-29 at 17 46 30

thaJeztah (commented):
Overall, I like this (but I may be biased 😅) Left some thoughts / ramblings 😂

_data/toc.yaml
hyu: Good point about using "Overview" to make Compose subnav consistent with other sections! 

But, for this PR I don't think we should block on this suggestion of a separate "Install" link and another "Installation per platform" section.

| Dora's PR | Suggested Change |
| --- | --- |
| image | image | 


"Per platform" suggests you'll see a list of links by platform. But actually the content available is a kind of grab-bag of install-related topics: the Linux plugin install, "other scenarios" that group Linux standalone and Windows Server, and Uninstall. So, it doesn't really seem like an accurate group description.

My suggestion is that we stick with what Dora has in this PR, with everything under "Install Compose." We do have a similar precedent with the "Install Docker Desktop" info hierarchy:

| Dora's PR | Current Docs |
| --- | --- |
| image | image |


&nbsp;
&nbsp;
&nbsp;
That said…now I'm looking more closely at this, it does feel redundant to have:

```
Install Compose
  Install Docker Compose
  …
```

So how about this instead?

```
Install
  Install Docker Compose
  …
```

compose/install/compose-other.md
dockertopia: Collapsing addressed comment.

compose/install/index.md
dockertopia: Resolving addressed comment.

compose/install/compose-linux.md
dockertopia: If I understood correctly, you meant since we're providing instructions (as opposed to just how to setup the Docker repo) on how to install using the package manager, we might as well provide the instruction for upgrading? Correct?

compose/install/index.md
dockertopia: 1) I searched in the most obvi places desktop pages, and sources: desktop, get-started and root folder and didn't find any similar image.
2) I think we don't necessarily need to update it every release, depends on which solution we find acceptable. Maybe we could add a note/sentence saying this is an example for illustration purposes only taken at version x and is not intended to reflect the current state of affairs? WDYT? 
To blur it I need to add this to the backlog of a posterior round of edits.

compose/install/compose-linux.md
dockertopia: This is an important question, I would prefer to sync on this and probably save this for a second round of updates to page. As we will also have feedback from users soon.

_data/toc.yaml
thaJeztah: > But actually the content available is a kind of grab-bag of install-related topics

Yes, it definitely is, and I was somewhat hoping we could keep that hidden for most users (who have Desktop installed). Had a bit of a hard time coming up with a section title for that ("Advanced install"? "Other installation methods"?). 

> That said…now I'm looking more closely at this, it does feel redundant to have:

Yes, that was the other reason I was looking at changing.

> My suggestion is that we stick with what Dora has in this PR

... after all of the above said; I'm ok with refining after this if we don't (yet?) want to change. That is; assuming it will be changes in the TOC only (restructuring the menu); where possible we should avoid changing URLs / renaming files (as moving them around may also cause penalties from search engines / ranking). So if we address those before merging (e.g. https://github.com/docker/docker.github.io/pull/15494#discussion_r957487060), then all should be fine.

compose/install/compose-linux.md
dockertopia: Adding this to sync's agenda.

compose/install/compose-linux.md
dockertopia: Good point. Adding this to the batch of changed, I'll prompt you for another review when ready.

StefanScherer:
For the about box screenshot `compose/images/about_desktop.png` it would be good to have Compose V2 version shown instead of the 1.29.2 V1 version. :-)

compose/install/compose-linux.md
dockertopia: As discussed use `Docker Compose version <version>` as console output.

compose/install/compose-linux.md
dockertopia: As discussed leave essential instructions for update, install and check version.

compose/install/compose-linux.md
dockertopia: Suggestion to link to this topic http://localhost:4000/engine/install/linux-postinstall/#manage-docker-as-a-non-root-user.

compose/install/compose-linux.md
dockertopia: Add an upgrading topic, with above mentioned instructions.

_data/toc.yaml
dockertopia: Changed per suggestion.

_data/toc.yaml
dockertopia: Removed.

compose/install/compose-linux.md
dockertopia: Proposing another solution, where I edited the image with placeholder text.

_data/toc.yaml
dockertopia: * Changed the title of Overview of Compose (...) to follow what we're doing in other doc sets. 
* I haven't changed the Install section title - we can do it later on if we want all three doc sets to have that title instead.
* Changed the index to "Install overview" - this way it is also clearer there are no actual instructions in this page.
___
* Some inconsistencies in docs are not about docs, so it will be hard to solve them with docs.
* Engine installation is more information-dense and distributed across platforms/distros and will also need some consolidation IMHO.
* Personally, I find it confusing to have an "Install" page and then a "Install per distro" section below it. These are not independent topics, and so, do not merit a separate item in the TOC and should be grouped together.

compose/install/linux.md
```diff
@@ -0,0 +1,141 @@
...
+RPM-based distros:
+
+    ```console
+
+    $ sudo yum update
+    $ sudo yum install docker-compose-plugin
+
+    ```
+
+2. Update the `apt` package index, and install the _latest version_ of Docker Compose:
```
usha-mandya, line 55: Appears as 1. in the preview - https://deploy-preview-15494--docsdocker.netlify.app/compose/install/linux/#install-using-the-repository

compose/install/linux.md
```diff
@@ -0,0 +1,141 @@
...
+
+    Ubuntu, Dei:
```
usha-mandya, line 57: Nit: Debian

compose/install/linux.md
```diff
@@ -0,0 +1,141 @@
...
+    ```
+Where `vN.N.N` is a placeholder for the latest version.
+
+> **Note**
+>
+> To run Compose as a non-root user, see [Manage Docker as a non-root user](../../engine/install/linux-postinstall.md){:target="_blank" rel="noopener" class="_"}.
+
+#### Update Compose
+
+To update, do:
```
usha-mandya, line 78: Are we missing something here? Should this be, "To update Compose, run the following command" or something on those lines?

compose/install/other.md
```diff
@@ -0,0 +1,66 @@
...
+
+title: Other install Scenarios
+---
+
+## Install Compose Standalone
+
+### On Linux
+> **Compose standalone**
+>
+> Note that Compose standalone uses the _dash compose_ syntax instead of current's standard syntax (_space compose_).
```
usha-mandya, line 14: Nit: Typo current's

compose/install/index.md
```diff
@@ -2,31 +2,45 @@
...
+### Pre-existing Docker Installation
+
+If you have a Docker installation on Linux to which you want to add the Compose plugin, you can choose one of these options:
+
+* [Install using the repository](./linux.md#install-using-the-repository)
+* [Install the plugin manually](./linux.md#install-the-plugin-manually)
+
+### Other install scenarios
+
+For other options check, see [Other install scenarios](./other.md/).
```
usha-mandya, line 42: "For other options, see..."

hyu (approved):
LGTM!

compose/install/index.md
```diff
@@ -35,3 +49,7 @@ Are you looking to:
 - [Compose file reference](../compose-file/index.md)
 - [Sample apps with Compose](../samples-for-compose.md)
 
+## Feedback
+
+We have updated this page and would like to hear from you.
```
usha-mandya, line 54: How about "We'd love to hear from you about the improvements we've recently made to the Docker Compose installation instructions. Let us know what you think by completing this short [survey](<link>)."

CC @hyu

hyu: Good phrasing! I approve.

dockertopia:
> For the about box screenshot `compose/images/about_desktop.png` it would be good to have Compose V2 version shown instead of the 1.29.2 V1 version. :-)

@Stefan an image with placeholder text for the versions was added instead.

hyu (approved):
🚀

StefanScherer (approved):
LGTM

aevesdocker (approved):

PR #15506, opened 2022-08-28
Docs: update sample application initiation in venv

pedramardakani:
Until now, readers would install 'Flask' on their main system
python environment, freeze their entire packages, and grep only
the 'Flask' package! So they would be missing all the other
dependencies in their requirements file.

With this commit, a better practice is followed. First, create a
clean virtual environment, then install 'Flask' and freeze
everything into the requirements file.

Also, I recommend using 'python3 -m pip freeze' instead of 'pip3
freeze' while in the virtual environment. This is because
sometimes hidden quirks of 'pip' will ignore the actiavted
environment and freeze the entire python packages in the
system (e.g. in ArchLinux).


StefanScherer (approved):
Thanks @pedramardakani for the suggestion and the PR. LGTM

PMohanJ:
Hi @pedramardakani, in the snippet below, I think the first two commands are redundant.
```console
$ cd /path/to/python-docker
$ source .venv/bin/activate
(.venv) $ python3 -m flask run
```
like if the user/person is following along the documentation, he/she would already be in the activated environment.

PMohanJ:
Hi @pedramardakani
> But as the previous sentence in the documentations suggests, the user might open up a new terminal. In that case, some that are not very well acquainted with python virtual environments, might get confused that why their flask application is not running.

Considering the above case, as you said; it's better to leave it like that.
Good day!

PR #15513, opened 2022-08-29
add start command to run hello-world image

developer0hye:
### Proposed changes

I added start command before running `sudo docker run hello-world` because we can't run `docker image` after installing docker following guide without running `docker`.


dockertopia (approved):
Thanks @developer0hye!

PR #15527, opened 2022-08-31
Clarify bind-mount obscuring of non-empty directories in container

carehart:
The current wording implies that ANY bind mount into a non-empty directory will obscure the files in that directory, but while that's true of bind mounting a directory, it's not true if bind-mounting a single file (from my testing, in a linux container running Docker Desktop for Windows).

I simply added that minor (but important) clarification of "a directory".

### DIFF
--- a/storage/bind-mounts.md
+++ b/storage/bind-mounts.md
@@ -158,7 +158,7 @@ $ docker container rm devtest
 
 ### Mount into a non-empty directory on the container
 
-If you bind-mount into a non-empty directory on the container, the directory's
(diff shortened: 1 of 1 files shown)


usha-mandya:
Thank you @carehart. 

@djs55 Could you PTAL?

carehart:
Folks?

djs55:
@carehart thanks for the clarification!

usha-mandya (approved):
@carehart LGTM. Apologies for the delay!

carehart:
Great, thanks.

PR #15529, opened 2022-08-31
Update information about binding privileged ports on Mac without vmnetd

gabriellavengeo:
### Proposed changes

Fixes inaccurate detail about the limitations of running Docker Desktop on Mac without `vmnetd`.


djs55 (approved):
LGTM. (I wonder if we should tweak the getting-started instructions)

aevesdocker (approved):

PR #15545, opened 2022-09-01
Fix contents of the table with the last updates on precedence

ulyssessouza:


compose/envvars-precedence.md
```diff
@@ -43,4 +41,4 @@ Description for each column:
 * Command line - Environmental variable passed via `docker compose run -e <KEY[=[VAL]]>`.
 * Resolved as - This column expresses the result available in the container.
 
-**Note that the "`OS`" has precedence over "`.env` file" column for variable resolution in the other columns.**
```
dockertopia, line 46: This is the same sentence, right?

dockertopia (changes requested):
The table results look very different, does the precedence rule stated need any updates?
In order to understand how the variable is resolved, I believe it works best to keep the present column order and names. For CMD I think we should adjust the description for the column instead of changing the name. I think it provides a clearer description of the context.

compose/envvars-precedence.md
ulyssessouza: It's the same. The only thing that changed is that it has a new line in the end of the file.
This is a standard (on POSIX systems at least) to have a file ending with a new line.

ulyssessouza:
Yes. The contents needed some change. Also changed the order to have some sense with the proximity.
For the `CMD`, it only makes sense on Windows. It's an OS specific implementation of a Shell. So `CMD` doesn't mean anything on Mac or Linux. 
Also `CMD` doesn't relate to the first item on the order of precedence.

nicksieger:
Would it help to declare the the order for both the bullet paragraph (higher to lower)  as well as the table (left = lower, right = higher)?

> The list is in the order of decreasing precedence. Items at the top of the list take precedence over items below.

> The columns of the table from left to right are in the order of increasing precedence. A column has precedence over any column to its left.

nicksieger:
Also consider using "Host" instead of "OS"? To differentiate from the container environment.

glours (commented):
We should replace `Unset` by `-` to improve readability of the matrix. 
Also agreed with @nicksieger to specify that the OS env variables refere to the `Host` ones

glours:
> Would it help to declare the the order for both the bullet paragraph (higher to lower) as well as the table (left = lower, right = higher)?
> 
> > The list is in the order of decreasing precedence. Items at the top of the list take precedence over items below.
> 
> > The columns of the table from left to right are in the order of increasing precedence. A column has precedence over any column to its left.

@ulyssessouza is it possible to follow @nicksieger remark and keep the same priority order between the list and the table?

dockertopia:
So question:
**Is this the order and content we want for both table and bullet paragraph?**
1. Passed from the .env file.
2. Set in the Host OS environment.
3. Passed from/set in Container Image in the [ENV directive](../../engine/reference/builder/#env)
4. Passed from/set in `compose.yaml` service's configuration, from the [env_file key](../../compose/compose-file/#env_file).
5. Passed from/set in `compose.yaml` service's configuration, from the [environment key](../../compose/compose-file/#environment)
6. Passed from the command-line [`docker compose run --env <KEY[=[VAL]]>`](../../engine/reference/commandline/compose_run/#options).


And if so, summary of feedback is:
1. Align paragraph with table (re order / contents)
2. Replace UNSET for a dash (notation wise).
3. Where it now reads TAG should read $TAG.
4. Where it reads "OS environment" (table column name) should read "Host OS environment"

Thanks

glours (approved):
LGTM, thanks @ulyssessouza and @dockertopia 🙏

dockertopia (approved):

ulyssessouza:
The list of precedence is now the opposite of what it should be. Also, the first and second lines shouldn't be in the list, since they do not expose variables to the container.

dockertopia:
> The list of precedence is now the opposite of what it should be. Also, the first and second lines shouldn't be in the list, since they do not expose variables to the container. 

@ulyssessouza I made changes based on reviews from the team at large. Your comments would have been useful at this https://github.com/docker/docker.github.io/pull/15545#issuecomment-1239451975 point. 

It was discussed in the stand up that it would help to have in the bulleted paragraph and the tables aligned. You are now saying "the first and second lines shouldn't be in the list, since they do not expose variables to the container."
I'm confused by this because our explanation, especially taking the example is based on all those contexts and what we're setting (or not) for that variable, how the variable is resolved. 
Based on your comment:
1) If the result is what is in the Image, then we don't need 2 columns: Image and Resolved as. As the result would be check what is (column) Image.
2) From what I could understand the list is not about the contexts from where you can expose variables to the container, but in terms of interpolation all the "contexts" where env vars are defined and the order of precedence they take for Compose.

I can have the order of the bulleted list to reversed to reflect the logic of the table. But as for the other comments,  I'm not sure how to conciliate the different inputs. 

@ulyssessouza @glours @nicksieger do we need a brief sync for this?

glours:
Let's talk about this point during tomorrow standup 👍

PR #15552, opened 2022-09-02
ENGDOCS-967

aevesdocker:
Look and feel refresh of the right-hand toc. Makes it less dominant and the content of the page 'pop' more. 

This is part of the ongoing work to revamp/refresh docs.docker.com


usha-mandya (approved):
LGTM

PR #15553, opened 2022-09-02
[ENGDOCS-978] Hyphenate 'sign in' when noun

craig-osterhout:
### Proposed changes

Add a hyphen when "sign in" is used as a noun.

### Related issues (optional)
ENGDOCS-978


jerae-duffin (approved):
LGTM!

PR #15570, opened 2022-09-05
Updates from master

crazy-max:
closes #15569 

https://github.com/docker/docker.github.io/pull/15566 didn't take into account all changes from master and therefore conflicts were created on published branch: #15569 

This PR reverts #15566 and applies all changes from master so we are aligned.


dockertopia (approved):
Thanks for the fix,.

PR #15574, opened 2022-09-06
fix db to backend

niaomingjian:
fix db to backend in network - aliases


usha-mandya:
@niaomingjian Thanks for the PR. Could you create a PR to update the [upstream Compose spec docs](https://github.com/compose-spec/compose-spec/blob/master/spec.md) before we can accept the changes here.

PR #15577, opened 2022-09-06
Loss of keys - update of warnings to users

dockertopia:
Updating loss of keys warnings as per received request for change ENGDOCS-977.

### DIFF
--- a/engine/security/trust/index.md
+++ b/engine/security/trust/index.md
@@ -101,11 +101,6 @@ The following image depicts the various signing keys and their relationships:
 consumer that used a signed tag from this repository prior to the loss.
 {:.warning}
 
-
-
-
-
-
 You should back up the root key somewhere safe. Given that it is only required
 to create new repositories, it is a good idea to store it offline in hardware.
 For details on securing, and backing up your keys, make sure you
--- a/engine/security/trust/index.md
+++ b/engine/security/trust/index.md
@@ -98,9 +98,8 @@ The following image depicts the various signing keys and their relationships:
 > **WARNING**
 >
 > The root key once lost is not recoverable.
-> Correcting this loss requires intervention from [Docker
-> Support](/support/) to reset the repository state. This loss
-> also requires **manual intervention** from every consumer that used a signed
+> If you lose a key, send an email to [Docker Hub Support](mailto:hub-support@docker.com).
+> This loss also requires **manual intervention** from every consumer that used a signed
 > tag from this repository prior to the loss.
 {:.warning}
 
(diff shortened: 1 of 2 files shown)


engine/security/trust/index.md
```diff
@@ -97,13 +97,15 @@ The following image depicts the various signing keys and their relationships:
...
-> tag from this repository prior to the loss.
+>The root key once lost is not recoverable. If you lose any other key, send an email to [Docker Hub Support](mailto:hub-support@docker.com). This loss also requires manual intervention from every
+consumer that used a signed tag from this repository prior to the loss.
 {:.warning}
 
+
+
+
+
+
```
dvdksn, line 108: There are a lot of empty lines here, it's best if we only have two newlines per block. Some markdown renderers would render empty paragraphs here which would create a big margin.

engine/security/trust/trust_key_mng.md
```diff
@@ -69,16 +71,18 @@ for more details.
 
 Prior to Docker Engine 1.11, this feature was only in the experimental branch.
 
-## Lost keys
+## Key Loss
```
dvdksn, line 74: nit

```suggestion
## Key loss
```

engine/security/trust/trust_key_mng.md
```diff
@@ -69,16 +71,18 @@ for more details.
...
 
-If a publisher loses keys it means losing the ability to sign trusted content for
-your repositories.  If you lose a key, send an email to [Docker Hub
-Support](mailto:hub-support@docker.com) to reset the repository
-state.
+> **WARNING**
+>
+> If a publisher loses keys it means losing the ability to sign trusted content for the repositories in
```
dvdksn, line 78: I would probably steer clear of the term "trusted content" when we are talking about signed software. Use trusted content for DOI/DVP/DSOS.

```suggestion
> If a publisher loses keys it means losing the ability to sign images for the repositories in
```

dvdksn (changes requested):
A couple minor comments, but the phrase "trusted content" I think we should be careful with in this context, going fwd

engine/security/trust/index.md
dockertopia: I'm surprised the linter didn't get this. Isn't this being check there?

dockertopia: Addressed in commit 3.

engine/security/trust/trust_key_mng.md
dockertopia: Addressed in commit 3.

engine/security/trust/trust_key_mng.md
dockertopia: Addressed in commit 3.

dockertopia:
@dvdksn as discussed looped you in as I thought this content would interest you. Thanks for your suggestions, can you PTAL.

dvdksn (approved):
🌟🌟🌟🌟🌟

PR #15578, opened 2022-09-06
Fix documentation command to install .deb package

kln:
### Proposed changes
The command to install .deb packages is wrong


aevesdocker:
Closing issue as documentation is correct. Can install using both commands

PR #15591, opened 2022-09-07
Adding note promoting survey

dockertopia:
Although the survey's been up for a few days and we've had around 20k views, there's no answers to the survey. In talking with Erika, we think the reference to the survey is very subtle and one solution would be to give it some more attention.

### Proposed changes
Adding a "Give feedback" note and possibility of entering swag lottery.

### Related issues (optional)
#15494

### DIFF
--- a/compose/install/index.md
+++ b/compose/install/index.md
@@ -9,6 +9,10 @@ redirect_from:
 
 On this page you can find information on how to get and install Compose.
 
+> **Give Feedback!**
+>
+> [Give us feedback](#where-to-go-next) on this page and get entered into a lottery for Docker swag!
+
 ## Install Compose
 
 If you have Docker Desktop, you've got a full Docker installation, including Compose.
--- a/compose/install/index.md
+++ b/compose/install/index.md
@@ -9,7 +9,7 @@ redirect_from:
 
 On this page you can find information on how to get and install Compose.
 
-> **Give Feedback!**
+> **Give feedback!**
 >
 > [Give us feedback](#where-to-go-next) on this page and get entered into a lottery for Docker swag!
 


compose/install/index.md
```diff
@@ -9,6 +9,10 @@ redirect_from:
 
 On this page you can find information on how to get and install Compose.
 
+> **Give Feedback!**
+>
+> [Give us feedback](#where-to-go-next) on this page and get entered into a lottery for Docker swag!
```
usha-mandya, line 14: This should point to the 'Feedback' section. It currently points to another section.

Also some minor suggestion about wording: Provide feedback about Docker Compose installation instructions and enter into a lottery for Docker swag!

compose/install/index.md
```diff
@@ -9,6 +9,10 @@ redirect_from:
 
 On this page you can find information on how to get and install Compose.
 
+> **Give Feedback!**
```
usha-mandya, line 12:
```suggestion
> **Give feedback!**
```

compose/install/index.md
dockertopia: @usha-mandya 

This was not accidental,  if you use #feedback the section header will not be in view and it is confusing. 

**Using #feedback**
image

If I use the previous one I see the whole "Feedback" section header and all, so this works best for users to situate themselves. 

**Using the previous section #**
image

I think "this page" is shorter and is leaves it lighter (whereas comment on the instructions might seem a more involved process)and since the context is the page where the note is at the top of it does not require clarification. 
Also we use that wording in the section where the survey link is available from. Are you too set on this suggestion? 
If not, I would just adjust the note's case from title to sentence case.

usha-mandya:
Closing this as we are planning to implement a feedback mechanism.

PR #15603, opened 2022-09-08
Add build cache introduction docs

jedevc:
### Proposed changes

Add new introduction to caching for build. This includes a simplified explanation of how cache works, as well as how users can take advantage of this behavior to get efficient and quick builds for them and their teams.

I'm not quite sure how to do the diagrams that I want - I think just want to show a stack of layers, and then color code / add notes on them for when the cache is invalidated. I'm not sure what precedence there is in the docs already, but happy to use whatever, or even work on adding support for something like mermaid.js if that's needed :tada:

CC @dockertopia @crazy-max

Added page at [/build/cache](https://deploy-preview-15603--docsdocker.netlify.app/build/cache/)


crazy-max:
Awesome thanks @jedevc!

> I'm not quite sure how to do the diagrams that I want - I think just want to show a stack of layers, and then color code / add notes on them for when the cache is invalidated. I'm not sure what precedence there is in the docs already, but happy to use whatever, or even work on adding support for something like mermaid.js if that's needed 🎉

Don't think we have anything yet about cache management except this section https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#leverage-build-cache. As for diagrams I'm not sure we support mermaid ones with Jekyll atm. We could add the minimal mermaid script and add support for this lang with markdown syntax using `language-mermaid`. I think something like tonis has done in his cross comp post would be enough: 

image

crazy-max:
LGTM, maybe we should wait for #15618 to be merged so you can add it to this section?

dvdksn (changes requested):
Awesome update, it was really great to read. I submitted some change suggestions in a PR here: <https://github.com/jedevc/docker.github.io/pull/2>

dvdksn (approved):
💎

PR #15604, opened 2022-09-08
Update django.md

surfdemon:
### Proposed changes

the command "docker compose up" didn't work on my Mac. It gave error saying "ModuleNotFoundError: No module named 'psycopg2'" and failed to load. 

I had to run "docker compose up --build" and that worked.

### DIFF
--- a/samples/django.md
+++ b/samples/django.md
@@ -243,6 +243,10 @@ In this section, you set up the database connection for Django.
    > This value is **not** safe for production usage.  Refer to the
    > [Django documentation](https://docs.djangoproject.com/en/1.11/ref/settings/#allowed-hosts)  for more information.
 
+Note: if this errors with "ModuleNotFoundError: No module named 'psycopg2'", first you will need to run 
+  ```console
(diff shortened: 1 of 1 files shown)


craig-osterhout:
Hi @surfdemon, thank you for contributing. 
Can you help explain why the error appeared and why you had to add the `--build` flag to fix it?

surfdemon:
@craig-osterhout I don't know why it errored, I followed the instructions on the page and when I got to that command, it failed. I then searched on Google and after lots of pages that said to use psycopg2-binary instead of psycopg2 which didn't work I then found a page that said to do --build. I tried that and it worked, since doing that once I've not needed to add --build to the command.

craig-osterhout:
Thank you @surfdemon.
It sounds like there was a typo on that line in the Dockerfile that was fixed when you changed it and then changed it back.
In this case, `--build` was required to rebuild the image after it was corrected in the Dockerfile and is probably not the case for everyone else.
Closing this, but please feel free to reopen with any additional info.

PR #15618, opened 2022-09-10
build: building images section

crazy-max:
This changes follow the new hierarchy proposal that has been discussed internally the past few weeks.

First phase is creating a new `Building images` section under `Docker Build` section in manuals:
* `Hello Build` page has been moved to `Packaging your software` and reworked
* `Multi platform builds` page under `Buildx` section has been moved and reworked
* `Multi-arch` page under `Desktop` has been merged with `Multi platform builds`

@jedevc I'm not sure if drivers docs should be in `Building images` section actually. I think we can keep the `Buildx` section in the meantime.


build/index.md
```diff
@@ -28,56 +30,74 @@ and much improved functionality that can be powerful tools for improving your
 builds' performance or reusability of your Dockerfiles, and it also introduces
 support for complex scenarios.
 
-## Docker Build features
+Docker Build is way more than the `docker build` command and is not only about
+packaging your code, it's a whole ecosystem of tools and features that support
+you not only with common workflow tasks but also provides you with support for
+more complex and advanced scenarios:
+
```
jedevc, line 37: Maybe out of scope for this PR, but I think we need more details on what BuildKit is, since the rest of our docs refer to Buildx/BuildKit freely, so we should introduce these concepts as early as possible.

At the very least, we need to clarify:
- The client (build/buildx) - server (buildkit) architecture
- `docker build` vs `docker buildx build`
- Legacy vs BuildKit

I don't think we need much space for this, but we should go over it.

_data/toc.yaml
```diff
@@ -1389,14 +1387,18 @@ manuals:
...
+      section:
+        - path: /build/building/packaging/
+          title: Packaging your software
+        - path: /build/building/multi-platform/
+          title: Multi-platform images
     - sectiontitle: Buildx
       section:
         - path: /build/buildx/
-          title: Buildx overview
+          title: Overview
```
jedevc, line 1399: I'd like to get rid of the Buildx section if we can, but maybe we could approach that later.

- We should merge the multiple builders page into the drivers overview, since it's too small by itself, and duplicates some of the info.
- Buildx verview page could be merged to the build overview
- Buildx install could be made to a generic build install page, where we could also cover buildkit as well later

Then drivers is left by itself, so I think we should move it to Building Images.

build/building/multi-platform.md
```diff
@@ -0,0 +1,274 @@
...
+#16 pushing layers
+#16 pushing layers 3.6s done
+#16 pushing manifest for docker.io/username/demo:latest@sha256:f3b552e65508d9203b46db507bb121f1b644e53a22f851185d8e53d873417c48
+#16 pushing manifest for docker.io/username/demo:latest@sha256:f3b552e65508d9203b46db507bb121f1b644e53a22f851185d8e53d873417c48 1.4s done
+#16 DONE 5.6s
+```
+
+> **Note**
+> 
+> * `username` must be a valid Docker Hub username.
```
jedevc, line 214: I've started using "template" style text for things like this, see: https://github.com/docker/buildx/blob/master/docs/guides/drivers/docker-container.md?plain=1#L53. We could go for `<user>/<image>` or just `<image>` or something?

Not sure if we have a preferred style through the docs, but would be good to align on a single style for at least the build docs.

build/index.md
```diff
@@ -28,56 +30,74 @@ and much improved functionality that can be powerful tools for improving your
...
+## Building your images
 
-Docker Build is way more than your `docker build` command and is not only about packaging your code, it’s a whole ecosystem of tools and features that support you not only with common workflow tasks but also provides you with support for more complex and advanced scenarios.
-Here’s an overview of all the use cases with which Build can support you:
+### Packaging your software
```
jedevc, line 40: I quite liked the old style which did this as bullet points - I think the headers look a bit dense without a lot of content around them. The list style will definitely work a lot better as we rework the explanations and point them towards the new content we're working on.

_data/toc.yaml
crazy-max: Yes we should get ride of it and move bits around.

> We should merge the multiple builders page into the drivers overview, since it's too small by itself, and duplicates some of the info.

Yes that sounds good :+1:

> Buildx verview page could be merged to the build overview

As discussed yes good point

> Buildx install could be made to a generic build install page, where we could also cover buildkit as well later

Yes some sort of `Install` section with Buildx, BuildKit and other tooling looks good.

> Then drivers is left by itself, so I think we should move it to Building Images.

:+1:

build/index.md
crazy-max: Yes in `Docker Build > Overview` that sgtm

build/building/multi-platform.md
crazy-max: Good point, I'm not sure myself about the templating format across our docs. cc @usha-mandya

build/index.md
crazy-max: It's just easier to quickly look at headers because you have them available in the sidebar:

image

And also search engines make them referenced.

jedevc: Ah sounds good then :smile:

build/building/multi-platform.md
usha-mandya: @jedevc @crazy-max Looked at a couple of examples within Docs. We've used `<username>/<image>`. Suggest we use the same across Build docs

_data/toc.yaml
crazy-max: Merged buildx overview to main overview page. For multiple builders, I think you can move some bits to driver guide in buildx repo.

jedevc (approved):
LGTM :tada:

Will follow-up with the multiple builders content in buildx at some point this week :smile:

PR #15619, opened 2022-09-10
desktop: add buildx to docker desktop components list

crazy-max:
Looks like buildx was missing from components list included in Docker Desktop.


aevesdocker (approved):
Thank you @crazy-max

Issue #15617, opened 2022-09-10
GETTING STARTED app doesn't work (NEVER worked) - why is this ABSOLUTELY-CRITICAL demo always ABSOLUTELY GARBAGE ??? 


markvogt:
[get-started/02_our_app.md](https://docs.docker.com/get-started/02_our_app/)

I've followed this tutorial for building my own Docker image "from scratch" TO-THE-LETTER. 
I've:
CLONED the getting-started repo to a Win7 laptop running latest VSC
OBSERVED that the Sample Application documentation DOESN'T agree with what's ACTUALLY in the repo
(note: even after YEARS and YEARS and YEARS Docker simply CAN'T get this RIGHT, yet you want us to consider your product CREDIBLE -you all must have come from JAVA development backgrounds, yes? <they're the same way> ) 
The DOCUMENTATION says "MAKE a Dockerfile and use these few lines of code in it
I DO that, and attempt to build an image and GET an image, but spawning a CONTAINER from it yields an APP THAT DOESN'T WORK (throws errors) 

So next I decided to use the ORIGINAL (larger, more complex, still GARBAGE) Dockerfile... 
I CAN build an image
I CAN spawn a container from it... 
... and I STILL GET NOTHING when I browse to http://localhost:3000 

AMAZING how these magical, portable oh-so-simple "containerized" applications NEVER SEEM TO WORK. 

I honestly don't know WHY Docker continues to fascinate people so much. 

It certainly has NOTHING TO DO with the DOCKER DOCUMENTATION STAFF'S EFFORTS. 
IF anything their butchered efforts to make a "Getting Started" application has only made me want to go back to BUILDING APPLICATIONS FROM SCRATCH.

Luckily I'm not bitter. 

And I'm certainly NOT convinced ANYONE at Docker is EVER going to bother to read THIS issue for - oh - about 200 years... 

Data Scientist Mark in North Aurora IL

vladdoster:
I 100% agree that your frustration regarding incorrect documentation (especially since it is the get started example) is warranted. However, I don't think your blanket statements disparaging the organization are not helpful.

I value strong documentation and this oversight reflects poorly on Docker.

I've fixed the documentation.

https://github.com/docker/docker.github.io/pull/15620

NitishKumar06:
@markvogt What's the issue that you're facing in documentation? Do let me know about the stuff you're unable to understand.

NitishKumar06:
[...]

PR #15620, opened 2022-09-11
docs(02_our_app.md): update port to 8000

vladdoster:
### Proposed changes

The example should reflect the ports that are used to serve the [example application](https://github.com/docker/getting-started/blob/master/docker-compose.yml).

### Related issues (optional)

https://github.com/docker/docker.github.io/issues/15617

### Related PR (optional)

https://github.com/docker/getting-started/pull/302/files

Signed-off-by: Vladislav Doster <mvdoster@gmail.com>

### DIFF
--- a/get-started/02_our_app.md
+++ b/get-started/02_our_app.md
@@ -50,7 +50,7 @@ see a few flaws in the Dockerfile below. But, don't worry. We'll go over them.
    COPY . .
    RUN yarn install --production
    CMD ["node", "src/index.js"]
-   EXPOSE 3000
+   EXPOSE 8000
    ```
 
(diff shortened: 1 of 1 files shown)


craig-osterhout:
Hi @vladdoster, thank you for contributing.
The sample app for part 2 is in the `app` subfolder in that repo. There is no example dockerfile or compose file for the sample app.

The  compose file that you linked is not associated with the sample app. It is associated with the getting-started app. Same as the Dockerfile in the root of that repo. The getting-started app itself is a separate app that displays the tutorial. 

The fix is to probably create a separate repo for the example `app` to avoid confusion when pulling the entire repo, or briefly explain that the repo contains 2 apps when pulling the repo.

PR #15622, opened 2022-09-11
jekyll: set liquid error mode to strict

crazy-max:
While working on another PR I found out there are liquid warnings introduced by https://github.com/docker/docker.github.io/pull/15494 in the logs: https://github.com/docker/docker.github.io/runs/8288660605?check_suite_focus=true#step:4:491

```
#15 44.17     Liquid Warning: Liquid syntax error (line 59): Expected end_of_string but found id in "{{if eq .Name "compose"}}" in compose/install/uninstall.md
#15 44.17     Liquid Warning: Liquid syntax error (line 59): [:dot, "."] is not a valid expression in "{{.Path}}" in compose/install/uninstall.md
```

Looking at this page https://docs.docker.com/compose/install/uninstall/#inspect-the-location-of-the-compose-cli-plugin, format is indeed broken:

image

To avoid this kind of issues in the future, sets Liquid error mode to strict.

Also got the following issue while enabling new error mode for Liquid:

```
#16 17.79   Liquid Exception: Liquid syntax error (/src/_includes/cli.md line 136): Unexpected character \ in "{{'[],map[],false,0,0s,default,\'\',""' | split: ',' }}" in engine/reference/commandline/app.md
#16 17.79                     ------------------------------------------------
#16 17.79       Jekyll 4.2.2   Please append `--trace` to the `build` command
#16 17.79                      for any additional information or backtrace.
#16 17.79                     ------------------------------------------------
```

This is fixed with last commit by using `capture` instead of `assign`.

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -24,6 +24,10 @@ kramdown:
   syntax_highlighter: rouge
   toc_levels: 2..3
 
+# https://jekyllrb.com/docs/configuration/liquid/
+liquid:
+  error_mode: strict
+
 exclude:
   - _releaser
   - _samples
--- a/_includes/cli.md
+++ b/_includes/cli.md
@@ -133,7 +133,8 @@ For example uses of this command, refer to the [examples section](#examples) bel
   {%- endif -%}
   {% capture flag-orchestrator %}{% if option.swarm %}<span class="badge badge-info" data-toggle="tooltip" title="This option works for the Swarm orchestrator.">Swarm</span>{% endif %}{% if option.kubernetes %}<span class="badge badge-info" data-toggle="tooltip" title="This option works for the Kubernetes orchestrator.">Kubernetes</span>{% endif %}{% endcapture %}
   {% capture all-badges %}{{ deprecated-badge }}{{ experimental-daemon-badge }}{{ experimental-cli-badge }}{{ min-api }}{{ flag-orchestrator }}{% endcapture %}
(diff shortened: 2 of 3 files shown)


_config.yml
```diff
@@ -23,6 +23,10 @@ kramdown:
   syntax_highlighter: rouge
   toc_levels: 2..3
 
+# https://jekyllrb.com/docs/configuration/liquid/
+liquid:
+  error_mode: strict
```
thaJeztah, line 28: The only concern I have is that these errors may happen in external sources (eg reference YAMLdocs or Markdown we pull in), and while these should be addressed, we may not be able to do that immediately 🤔 


I'm surprised that the format string in compose didn't show up in the rendered output though; ISTR that liquid would just ignore the "invalid" liquid templates, and leave them as-is (so the warnings could be ignored if it wasn't really a liquid template)

crazy-max: > The only concern I have is that these errors may happen in external sources (eg reference YAMLdocs or Markdown we pull in), and while these should be addressed, we may not be able to do that immediately 🤔

We have the same issue with htmlproofer I guess, not sure what we can do about it except if upstream run the same validation steps like we do in buildx repo: https://github.com/docker/buildx/pull/1218

StefanScherer (approved):
SGTM

dvdksn (approved):
👍

PR #15624, opened 2022-09-11
Add instruction to remove older version of docker-compose-plugin

### LINKED ISSUE #15472: Also mention uninstalling `docker-compose` before installing.
The first step in the installation guide is to uninstall old versions of docker, in order go only get compatible packages from wherever you install from. Later, the `docker-compose-plugin` is installed, which would conflict with `docker-compose` (e.g. from ubuntu) if that is installed.

It would perhaps be nice to add uninstalling of `docker-compose` to the uninstall instructions.

[engine/install/ubuntu.md](https://docs.docker.com/engine/install/ubuntu/)

akankshakumari393:
### Proposed changes

Add instruction to remove older version of docker-compose-plugin if present

### Related issues (optional)

Fixes #15472

### DIFF
--- a/engine/install/ubuntu.md
+++ b/engine/install/ubuntu.md
@@ -44,11 +44,11 @@ Docker Engine is supported on `x86_64` (or `amd64`), `armhf`, `arm64`, and `s390
 
 ### Uninstall old versions
 
-Older versions of Docker were called `docker`, `docker.io`, or `docker-engine`.
+Older versions of Docker were called `docker`, `docker.io`, `docker-engine` or `docker-compose-pligin`.
 If these are installed, uninstall them:
 
 ```console
-$ sudo apt-get remove docker docker-engine docker.io containerd runc
+$ sudo apt-get remove docker docker-engine docker.io containerd runc docker-compose-plugin
 ```
 
 It's OK if `apt-get` reports that none of these packages are installed.


PMohanJ:
@akankshakumari393 there's a typo in your modifications.
On line number 47, it's written as `docker-compose-pligin` instead of `docker-compose-plugin`

engine/install/ubuntu.md
```diff
@@ -44,11 +44,11 @@ Docker Engine is supported on `x86_64` (or `amd64`), `armhf`, `arm64`, and `s390
 
 ### Uninstall old versions
 
-Older versions of Docker were called `docker`, `docker.io`, or `docker-engine`.
+Older versions of Docker were called `docker`, `docker.io`, `docker-engine` or `docker-compose-pligin`.
```
usha-mandya, line 47: @akankshakumari393 There's a typo: `plugin`. Also, I don't think any older version of Docker was called 'docker-compose-plugin`, therefore, I don't think it makes sense to add Docker Compose in this line. 

Perhaps we can add a new line recommending users to uninstall Docker Compose if it already exists. I believe in this case, they should uninstall `docker-compose` and not `docker-compose-plugin` ? @glours could you please confirm?

PR #15626, opened 2022-09-12
correct docker compose up command

pedrojgcesar:
On my pc the docker compose up command was without the dash I think on the wiki is wrong. The only think I changed was the dash on the comand I spend a lot of time until I figured out it was wrong.

### DIFF
--- a/get-started/08_using_compose.md
+++ b/get-started/08_using_compose.md
@@ -278,7 +278,7 @@ Now that we have our `docker-compose.yml` file, we can start it up!
    background.
 
     ```console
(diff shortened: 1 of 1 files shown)


craig-osterhout:
Hi @pedrojgcesar, thank you for contributing. The command also appears right above the codeblock in the step's explanation. Can you also help update that for consistency.

pedrojgcesar:
Yes, I will try to do that today

craig-osterhout:
Closing in favor of #15736

PR #15635, opened 2022-09-12
Update environment-variables.md

mindcreations:
It is not immediately evident or implicitly excluded whether the environment variables declared in the file referenced by *env_file* are not available for use in the Compose file or to configure Compose even if the first line says "You can pass multiple environment variables from an external file through to a service's containers..."

### DIFF
--- a/compose/environment-variables.md
+++ b/compose/environment-variables.md
@@ -176,6 +176,10 @@ web:
   env_file:
     - web-variables.env
 ```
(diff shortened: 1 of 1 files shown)


usha-mandya:
@ulyssessouza Could you PTAL?

ulyssessouza (approved):

aevesdocker (approved):

PR #15649, opened 2022-09-14
remove absolute links

crazy-max:
We should avoid using absolute links when we can, otherwise we can't detect broken links as it links to prod and also netlify preview, staging environment are not accurate when navigating. cc @andrei-z-albu @aevesdocker

Maybe this is something we should stipulate in contributing docs?


aevesdocker (approved):
Thanks @crazy-max. Will add it to contribute section asap

PR #15656, opened 2022-09-14
absolute link addition

aevesdocker:


contribute/components/links.md
```diff
@@ -24,6 +24,8 @@ An example of a link to an auto-generated reference page that we pull in during
   - Keep in mind that this link doesn't resolve until you merge the PR and
   your docs are published on [docs.docker.com](/).
 
+- It is best practice to use absolute links when linking to other docs pages. Otherwise broken links may not be picked up.
```
usha-mandya, line 27: > ..use absolute links when linking to other docs pages.

Do we mean other docs pages within our docs site? If so, should this be relative links?

aevesdocker: Typo! I meant *avoid the use of* absolute links. 🤦

aevesdocker, line 27:
```suggestion
- It is best practice to avoid the use of absolute links when linking to other docs pages. Otherwise broken links may not be picked up. 
```

usha-mandya (approved):

PR #15663, opened 2022-09-15
terminology.md

Bergmann5:
Updating registry


craig-osterhout:
Hi @Bergmann5, thanks for contributing. It looks like the wrong level heading is used in the update.
Also, does this term or definition appear anywhere in the documentation or product?

usha-mandya:
Closing the PR as it's not a valid suggestion. If this is still relevant, please create a new PR with more information.

PR #15666, opened 2022-09-15
[ENGDOCS-943] Create C# .NET language-specific guide

craig-osterhout:
### Proposed changes

Create C# .NET guide based on [blog post](https://www.docker.com/blog/building-multi-container-net-app-using-docker-desktop/).

https://deploy-preview-15666--docsdocker.netlify.app/language/

### Related issues (optional)
ENGDOCS-943


usha-mandya:
@StefanScherer Could you PTAL? Thanks

language/dotnet/build-images.md
```diff
@@ -0,0 +1,307 @@
...
+```
+
+Output similar to the following appears.
+
+```console
+Building...
+warn: Microsoft.AspNetCore.DataProtection.Repositories.FileSystemXmlRepository[60]
+    Storing keys in a directory '/root/.aspnet/DataProtection-Keys' that may not be persisted outside of the container. Protected data will be unavailable when the container is destroyed.
+warn: Microsoft.AspNetCore.Server.Kestrel[0]
+     Unable to bind to http://localhost:5000 on the IPv6 loopback interface: 'Cannot assign requested address'.
```
StefanScherer, line 90: This error seems odd.

For me the app always listened only on port 5016
```
PS C:\Users\stefan\dotnet-docker\src> dotnet run –urls http://localhost:5000
Building...
info: Microsoft.Hosting.Lifetime[14]
      Now listening on: http://localhost:5016
info: Microsoft.Hosting.Lifetime[0]
      Application started. Press Ctrl+C to shut down.
info: Microsoft.Hosting.Lifetime[0]
      Hosting environment: Development
info: Microsoft.Hosting.Lifetime[0]
      Content root path: C:\Users\stefan\dotnet-docker\src\
info: Microsoft.Hosting.Lifetime[0]
```
so in the end I just skipped the `-urls` parameter and still have the app running on port 5016. 🤔

language/dotnet/build-images.md
```diff
@@ -0,0 +1,307 @@
...
+COPY --from=build-env /publish .
+EXPOSE 80
+ENTRYPOINT ["dotnet", "myWebApp.dll"]
+```
+
+## .dockerignore file
+
+To make your build context as small as possible, add a .dockerignore file to your `dotnet-docker` folder and copy the following into it.
+
+```console
```
StefanScherer, line 209: The `console` seems to make the text block to be read-only. I can't copy paste the lines.

language/dotnet/build-images.md
```diff
@@ -0,0 +1,307 @@
...
+description: Learn how to build your first Docker image by writing a Dockerfile
+---
+
+{% include_relative nav.html selected="1" %}
+
+## Prerequisites
+
+Work through the orientation and setup in Get started [Part 1](../../get-started/index.md) to understand Docker concepts.
+
+{% include guides/enable-buildkit.md %}
```
StefanScherer, line 13: Maybe for a later PR. Nowadays buildkit is enabled by default in current Docker Desktop versions (IIRC it's default with Docker 20.10.x), so we probably don't need this include anymore.

language/dotnet/build-images.md
```diff
@@ -0,0 +1,307 @@
...
+ENTRYPOINT ["dotnet", "myWebApp.dll"]
+```
+
+## .dockerignore file
+
+To make your build context as small as possible, add a .dockerignore file to your `dotnet-docker` folder and copy the following into it.
+
+```console
+bin/
+obj/
```
StefanScherer, line 211: I got an error during the docker build. Running the application on the host created the `bin` and `obj` folders inside the src folder. So to ignore any `bin` and `obj` folder in any subdirectory, prepend the `**/`.

```suggestion
**/bin/
**/obj/
```

language/dotnet/build-images.md
```diff
@@ -0,0 +1,307 @@
...
+
+```dockerfile
+COPY src/*.csproj .
+RUN dotnet restore
+```
+
+Next, you'll need to copy the rest of your source files into the image. The line below will copy the files from the `src` directory on your local machine to a directory called `src` in the image.
+
+```dockerfile
+COPY src ./src
```
StefanScherer, line 147: The workdir already is `/src`, so the target in the COPY line must be `.`

```suggestion
COPY src .
```

language/dotnet/build-images.md
```diff
@@ -0,0 +1,307 @@
...
+Here's the complete Dockerfile.
+
+```dockerfile
+# syntax=docker/dockerfile:1
+
+FROM mcr.microsoft.com/dotnet/sdk:6.0 as build-env
+WORKDIR /src
+COPY src/*.csproj .
+RUN dotnet restore
+COPY src .
```
StefanScherer, line 195: Ah, here it's already fixed. 👍 I've copy-pasted the single lines 😅

language/dotnet/develop.md
```diff
@@ -0,0 +1,460 @@
...
+```console
+$ docker run \
+  --rm -d \
+  --network postgres-net \
+  --name dotnet-app \
+  -p 5000:80 \
+  dotnet-docker
+```
+
+Let's test that the application works and is connecting to the database. Using a web browser, access `http://localhost:5050`. A page similar to the following image appears.
```
StefanScherer, line 265: It's port 5000, not 5050, there also are some similar differences further down this page.

language/dotnet/run-containers.md
```diff
@@ -0,0 +1,170 @@
...
+
+```console
+$ docker run dotnet-docker
+```
+
+After running this command, you’ll notice that you were not returned to the command prompt. This is because our application is a server and runs in a loop waiting for incoming requests without returning control back to the OS until we stop the container.
+
+Let’s open a new browser and access `http://localhost:80`.
+
+As you'll see, the connection to our server was refused. This means, we were not able to connect to the localhost on port 80. This is expected because our container is running in isolation which includes networking. Let’s stop the container and restart with port 5050 published on our local network.
```
StefanScherer, line 29: Port 5000?

language/dotnet/run-containers.md
```diff
@@ -0,0 +1,170 @@
...
+
+To stop the container, press Ctrl-C. This will return you to the terminal prompt.
+
+To publish a port for our container, we’ll use the `--publish` flag (`-p` for short) on the `docker run` command. The format of the `--publish` command is `[host port]:[container port]`. So, if we wanted to expose port 80 inside the container to port 5050 outside the container, we would pass `5050:80` to the `--publish` flag. Run the container using the following command:
```
StefanScherer, line 33: Port 5000?

StefanScherer (changes requested):
I walked through the tutorial which is awesome!
Some minor things need changes, I left a few comments with small fixes where I struggled.

And I still don't know why the app on the host always only listens on port 5016, so maybe that's only a problem on my machine, it seems it ignores `-urls` or `--urls`, I tried several things, but without success. 🤷

language/dotnet/build-images.md
StefanScherer: Maybe use `shell` here, it seems that all the text is interpreted as the console prompt and therefore I can't select it.

language/dotnet/build-images.md
craig-osterhout: Thanks, Stefan. I missed a dash in the flag. It should be `--urls`.  I'm surprised it doesn't give an error. 
As for the output, it was copied from the original blog post. When I run the command, I get the same output as you. I'm on Windows too, but it looks like the blog writer wasn't. Maybe the error has to do with the operating system.
I'll update the flag and the output.

language/dotnet/build-images.md
craig-osterhout: Thank you. Will update to `shell`.

language/dotnet/build-images.md
craig-osterhout: Thanks, I'll remove it. It looks like it was recently removed from the other guides as well.

language/dotnet/build-images.md
craig-osterhout: Thank you. Will update.

language/dotnet/build-images.md
craig-osterhout: Thank you. Will update.

language/dotnet/build-images.md
craig-osterhout: Sorry, I rearranged the folders after initially writing it and forgot to update the single line.

language/dotnet/develop.md
craig-osterhout: Thank you. Will update this occurrence to 5000.

language/dotnet/run-containers.md
craig-osterhout: Thank you. Will update.

language/dotnet/run-containers.md
craig-osterhout: Thank you. Will update.

craig-osterhout:
Thank you @StefanScherer. The guide has been updated based on your feedback.

Port 5000 didn't initially work on my machine either. After a while, it started working. I'm not sure what the issue is. I reworded that part to ask the user to check the output to verify which port is used. 

PTAL to verify the changes.

StefanScherer (approved):
Thanks for the corrections. LGTM 🎉

PR #15676, opened 2022-09-16
[ENGDOCS-994] Update samples

craig-osterhout:
### Proposed changes
Updated samples page based on feedback.

### Related issues (optional)
ENGDOCS-994


craig-osterhout:
Closing and will create new PR. GitHub Actions continues to fail even after removing the offending line.

PR #15700, opened 2022-09-20
Fix hub api list tags page_size description

zappy-shu:
### Proposed changes

Fixes the description of the list-tags page size parameter


usha-mandya (approved):
LGTM. Thank you @zappy-shu

PR #15730, opened 2022-09-23
Fix typos, tone and brevity

firgun:
### Proposed changes

Change "tells a bunch of stuff" to "tells us a bunch of stuff"

While I don't think this is a grammatical issue, it reads better and is generally how someone would use it in a conversation (the docs have a very conversational tone, which is why I chose with "tells us" instead of replacing "tells" with "reveals", for example.)

Change "name" to "names", because a container can have multiple names (the docker ps command displays the column as "NAMES".  It's also more in line with the author using "ports" to refer to the list of exposed ports (even though there is only one exposed port.)

Change the following:

  "To stop the container, run the `docker stop` command which does just that, stops the container. You will need to pass 
  the name of the container or you can use the container ID."

To:

    "To stop the container, run the `docker stop` command, passing the container's name or ID."

Because I feel like the the former is overly repetitious and lengthy without being more understandable or providing more insight. I understand the docs tend to have a very conversational, easy-to-read tone, but I think the latter fits with the tone well enough and is nice and concise.

### DIFF
--- a/language/golang/run-containers.md
+++ b/language/golang/run-containers.md
@@ -97,9 +97,9 @@ CONTAINER ID   IMAGE            COMMAND             CREATED          STATUS     
 d75e61fcad1e   docker-gs-ping   "/docker-gs-ping"   41 seconds ago   Up 40 seconds   0.0.0.0:8080->8080/tcp   inspiring_ishizaka
 ```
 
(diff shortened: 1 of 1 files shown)


craig-osterhout (approved):
Thanks, @firgun. LGTM

PR #15731, opened 2022-09-23
Nested lists

craig-osterhout:
## Issue
There's a style issue with nested lists as seen below.
https://docs.docker.com/contribute/style/grammar/#best-practice-5
image

https://docs.docker.com/contribute/components/lists/#examples
image

## Proposed fix
Proposing to update the style somewhere to fix it. Not sure if this is the best place.

### DIFF
--- a/_scss/_utilities.scss
+++ b/_scss/_utilities.scss
@@ -227,10 +227,10 @@ input[type=text] {
  * lists **********************************************************************
  */
 
-ol li ol li{
+ol li ol li {
   list-style-type: lower-alpha;
 }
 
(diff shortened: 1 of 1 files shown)


craig-osterhout:
Hi @aevesdocker, I'm not sure if this is the right fix or in the right place. Usha mentioned that you've been working on a lot of other style updates, so you might be more familiar with this. PTAL if you have time.

aevesdocker (approved):
This looks like the right place to me @craig-osterhout. Thank you for fixing this!

PR #15732, opened 2022-09-23
Update image location

craig-osterhout:
Image location has changed. Updated to reflect new location.
https://docs.docker.com/docker-hub/api/latest/
image


craig-osterhout:
@thaJeztah, there's also an image issue in the Engine API docs.
https://docs.docker.com/engine/api/v1.41/
image

PR #15736, opened 2022-09-25
Update docker compose command

umar006:
### Proposed changes
Base on [Compose V2](https://docs.docker.com/compose/reference/), correct command for docker compose is without "-" character so I propose to update all current command to V2 by removing "-".

`docker-compose` -> `docker compose`

### Related issues
https://github.com/docker/docker.github.io/pull/15626


craig-osterhout (approved):
Thanks, @umar006. LGTM.

PR #15246, opened 2022-07-27
move resources to assets folder

crazy-max:
follow-up https://github.com/docker/docker.github.io/pull/15223#issue-1319334942

Font Awesome library has been moved to `assets` folder in #15223 because accessing the resources from their css moved the fonts to a webfonts folder which is relative to their css.

To keep things consistent with resources in this repo, I moved them to the assets folder as well (see https://github.com/crazy-max/docker.github.io/tree/move-assets/assets). It will make resources easier to discover and track and also avoid to bloat root dir of the repository. It also makes more sense to have the resources dedicated to the libraries in their respective folders so you know which css is linked to which font (same for javascript).

As discussed with @thaJeztah, we should also take a look in a follow-up to reorganize the images in this repo as some of them are not used anymore or are duplicated in other folders.

PR #15752, opened 2022-09-27
docker-hub: fix link to logo asset

crazy-max:
Looks like this one was missing in #15246

There is also the swagger api files fetched from moby repo that needs to be fixed on master, 20.10 and 22.06 branches.


usha-mandya (approved):
LGTM. Thank you @crazy-max

PR #15758, opened 2022-09-27
Update ubuntu.md

saberder:
Change the install order to make the docker service start during docker-ce installation

### Proposed changes

Changed the install order of Ubuntu apt packages to make the docker systemctl service start during installation of docker-ce.
Tested on Ubuntu 22.04

### DIFF
--- a/engine/install/ubuntu.md
+++ b/engine/install/ubuntu.md
@@ -120,7 +120,7 @@ from the repository.
 
     ```console
     $ sudo apt-get update
-    $ sudo apt-get install docker-ce docker-ce-cli containerd.io docker-compose-plugin
+    $ sudo apt-get install containerd.io docker-ce docker-ce-cli docker-compose-plugin
     ```
 
     > Receiving a GPG error when running `apt-get update`?


engine/install/ubuntu.md
```diff
@@ -120,7 +120,7 @@ from the repository.
 
     ```console
     $ sudo apt-get update
-    $ sudo apt-get install docker-ce docker-ce-cli containerd.io docker-compose-plugin
+    $ sudo apt-get install containerd.io docker-ce docker-ce-cli docker-compose-plugin
```
thaJeztah, line 123: Hm... I don't think the order in which packages are specified makes a difference; `apt-get` will always take the list of package you specified, then resolve dependencies between them, and base the order in which they're installed based on that.

You can see that when running both commands (without actually installing); here's the "before";

```bash
sudo apt-get install docker-ce docker-ce-cli containerd.io docker-compose-plugin
...
The following NEW packages will be installed:
  containerd.io docker-ce docker-ce-cli docker-ce-rootless-extras docker-compose-plugin docker-scan-plugin libltdl7 libslirp0 pigz slirp4netns
...
After this operation, 423 MB of additional disk space will be used.
Do you want to continue? [Y/n]
```

And here's the "after";

```bash
sudo apt-get install containerd.io docker-ce docker-ce-cli docker-compose-plugin
...
The following NEW packages will be installed:
  containerd.io docker-ce docker-ce-cli docker-ce-rootless-extras docker-compose-plugin docker-scan-plugin libltdl7 libslirp0 pigz slirp4netns
...
After this operation, 423 MB of additional disk space will be used.
Do you want to continue? [Y/n]
```

In both cases, `containerd.io` is listed before `docker-ce`. Even if they would be installed in a different order, the systemd unit defines a dependency on the `containerd` service, so it would start containerd before it starts the engine.

thaJeztah (changes requested):
Not sure this is needed (I left a comment in-line), could you describe the problem you ran into?

engine/install/ubuntu.md
saberder: I just tested this assumption on Ubuntu 22.04.
First I removed my installed docker packages with `sudo apt remove docker-ce docker-ce-cli docker-ce-rootless-extras docker-compose-plugin docker-scan-plugin containerd.io`
and `sudo systemctl daemon-reload`.
Then i tried installing it again using the two commands above.
The first (original) way of doing it using  `sudo apt-get install docker-ce docker-ce-cli containerd.io docker-compose-plugin` returned an error during docker-ce setup and the docker service didn't start: Screenshot from 2022-09-28 12-42-02
Then I uninstalled docker packages again and ran the second command `sudo apt-get install containerd.io docker-ce docker-ce-cli docker-compose-plugin` and it worked fine.

But then I repeated this test a few times, installing and uninstalling docker and running the above commands in different orders. It looks like there is a timer here that stops the service from restarting if the uninstall / install happens to close to each other. Just uninstalling and waiting a few minutes makes both the above commands work.

usha-mandya:
Thanks for the PR. Closing this PR here due to inactivity. Please create a new PR if the updates are still relevant. Thank you.

PR #15759, opened 2022-09-27
compose: update env file based on actual behavior

milas:
### Proposed changes
There has been a LOT of work on this area in Compose recently. As part of this, we discovered that the documentation has been incorrect for some time. That is, the documented behavior has often never been indicative of reality (across both v1 and v2).

We've standardized and cleaned things up, so the documentation now reflects current behavior. The Compose Spec requires similar corrections/reconciliation, but we're starting here to ensure that the more user-facing docs are in a good place.

### Related issues (optional)
* docker/compose#9806,
* docker/compose#9746,
* docker/compose#9704,
* docker/compose#9294

### DIFF
--- a/compose/env-file.md
+++ b/compose/env-file.md
@@ -22,7 +22,7 @@ The following syntax rules apply to environment files:
   - `VAR="VAL" # comment` -> `VAL`
 - Blank lines are ignored.
 - Unquoted and double-quoted (`"`) values will have [parameter expansion](#parameter-expansion) applied.
-- Single-quoted (`'`) values will be used literally.
+- Single-quoted (`'`) values are used literally.
   - `VAR='$OTHER'` -> `$OTHER`
   - `VAR='${OTHER}'` -> `${OTHER}`
 - Quotes can be escaped with `\`.
--- a/compose/env-file.md
+++ b/compose/env-file.md
@@ -4,29 +4,56 @@ keywords: fig, composition, compose, docker, orchestration, environment, env fil
 title: Declare default environment variables in file
 ---
 
-Compose supports declaring default environment variables in an environment file
-named `.env` placed in the project directory. Docker Compose versions earlier than `1.28`,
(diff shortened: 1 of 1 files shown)


compose/env-file.md
```diff
@@ -4,29 +4,56 @@ keywords: fig, composition, compose, docker, orchestration, environment, env fil
...
+  - `VAR='VAL'` -> `VAL`
+- Lines beginning with `#` are processed as comments and ignored.
+- Inline comments for unquoted values must be proceeded with a space.
+  - `VAR=VAL # comment` -> `VAL`
+  - `VAR=VAL# not a comment` -> `VAL# not a comment`
+- Inline comments for quoted values must follow the closing quote.
+  - `VAR="VAL # not a comment"` -> `VAL # not a comment`
+  - `VAR="VAL" # comment` -> `VAL`
+- Blank lines are ignored.
+- Unquoted and double-quoted (`"`) values will have [parameter expansion](#parameter-expansion) applied.
```
aevesdocker, line 24:
```suggestion
- Unquoted and double-quoted (`"`) values have [parameter expansion](#parameter-expansion) applied.
```

compose/env-file.md
```diff
@@ -4,29 +4,56 @@ keywords: fig, composition, compose, docker, orchestration, environment, env fil
...
+- Single-quoted (`'`) values will be used literally.
```
aevesdocker, line 25:
```suggestion
- Single-quoted (`'`) values are used literally.
```

aevesdocker (commented):
Thank you @milas! Added two picky review comments, which I've already committed and merged so as to amend the docs asap!

 I think it would be worth exploring a different way to display this information. Perhaps some of the points could have tables with a 'Correct' column and an 'incorrect' column, which might make things clearer. But this is certainly not a priority for now!

aevesdocker (approved):

PR #15763, opened 2022-09-27
Update overview.md

ltlombardi:
### Proposed changes
fixed a small typo.

### DIFF
--- a/get-started/overview.md
+++ b/get-started/overview.md
@@ -69,7 +69,7 @@ business needs dictate, in near real time.
 **Running more workloads on the same hardware**
 
 Docker is lightweight and fast. It provides a viable, cost-effective alternative
-to hypervisor-based virtual machines, so you can use more of your compute
+to hypervisor-based virtual machines, so you can use more of your computer
 capacity to achieve your business goals. Docker is perfect for high density
 environments and for small and medium deployments where you need to do more with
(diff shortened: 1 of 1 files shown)


get-started/overview.md
```diff
@@ -69,7 +69,7 @@ business needs dictate, in near real time.
 **Running more workloads on the same hardware**
 
 Docker is lightweight and fast. It provides a viable, cost-effective alternative
-to hypervisor-based virtual machines, so you can use more of your compute
+to hypervisor-based virtual machines, so you can use more of your computer
```
thaJeztah, line 72: Thanks for contributing! 

Looking at this, I think `compute` may have been intentional here ("compute" as in "compute  / computing capacity")

@usha-mandya @dvdksn wdyt? Perhaps there's better wording to prevent confusing?

dvdksn: Hey. Thanks for the PR. I agree @thaJeztah! What about something like this?

> It uses fewer resources than hypervisor-based virtual machines, allowing you to dedicate more of your server capacity towards running application code.

ltlombardi: :-) 
Yes, definitely should be "computing power" or  "capacity". 
Did another commit

dvdksn (approved):

PR #15775, opened 2022-09-29
Hardened Desktop and Admin Controls

aevesdocker:
Fixes merge issues in https://github.com/docker/docker.github.io/pull/15671


desktop/hardened-desktop/index.md
```diff
@@ -0,0 +1,55 @@
+---
+title: Hardened Desktop
+description: Overview of what Hardened Desktop is
+keywords: security, hardened desktop, enhanced container isolation, registry access management, admin controls, root access, administrators, docker desktop
+---
+>Note
+>
+>Hardened Desktop is available to Docker Business customers only.
+
+Hardened Desktop is a new security model for Docker Desktop. It's part of Docker's ongoing effort to increase Docker Desktop security without impacting the developer experience.
```
christophermclellan, line 10: I'd change to:

"Hardened Desktop is a new security model for Docker Desktop which is designed to provide admins with a simple and powerful way to improve their organizations security posture for containerised development, without impacting the great developer experience that Docker Desktop offers."

let's immediately convey the value of Hardened Desktop and who it's for. 'ongoing effort to increase security' sounds like we're going through a bit of a slog.

desktop/hardened-desktop/index.md
```diff
@@ -0,0 +1,55 @@
...
+
+The Hardened Desktop security model provides administrators with a simple and powerful way to increase the security of their containerised development and moves the ownership boundary for containers to the organization, meaning that any security controls administrators set cannot be altered by the user.
+
+It is designed specifically for organizations who don’t give developers root access to their machines, and who would like Docker Desktop to be within the organization’s centralized control.
```
christophermclellan, line 14: slight adjustment, i'd move this up to be the second paragraph and change to:

"This configuration is designed for security conscious organizations that don’t give their users root / admin access on their machines and wish to configure Docker Desktop to be within their organizations centralized control."

desktop/hardened-desktop/index.md
```diff
@@ -0,0 +1,55 @@
...
+description: Overview of what Hardened Desktop is
+keywords: security, hardened desktop, enhanced container isolation, registry access management, admin controls, root access, administrators, docker desktop
+---
+>Note
+>
+>Hardened Desktop is available to Docker Business customers only.
+
+Hardened Desktop is a new security model for Docker Desktop. It's part of Docker's ongoing effort to increase Docker Desktop security without impacting the developer experience.
+
+The Hardened Desktop security model provides administrators with a simple and powerful way to increase the security of their containerised development and moves the ownership boundary for containers to the organization, meaning that any security controls administrators set cannot be altered by the user.
```
christophermclellan, line 12: then would condense this to:

"The Hardened Desktop security model moves the ownership boundary for containers to the organization, meaning that any security controls set by admins cannot be altered by the user of Docker Desktop."

desktop/hardened-desktop/index.md
```diff
@@ -0,0 +1,55 @@
...
+
+It is designed specifically for organizations who don’t give developers root access to their machines, and who would like Docker Desktop to be within the organization’s centralized control.
+
+Hardened Desktop includes:
+- Admin Controls, which helps administrators to confidently manage and control usage of Docker Desktop.
```
christophermclellan, line 17: "...control the usage of Docker Desktop within their organization"

desktop/hardened-desktop/index.md
```diff
@@ -0,0 +1,55 @@
...
+- Enhanced Container Isolation, an admin setting that instantly enhances security by preventing containers from running as root in Docker Desktop’s Linux VM.
```
christophermclellan, line 18: can we extend this to also mention that it ensures admin controls cannot be modified by containers ?

desktop/hardened-desktop/index.md
```diff
@@ -0,0 +1,55 @@
...
+- Registry Access Management, which allows administrators to control the registries developers can access and can be locked in if  Enhanced Container Isolation is switched on by using Admin Controls.
```
christophermclellan, line 19: i don't think we need to mention the relationship with enhanced container isolation here, especially if we note that enhanced container isolation prevents admin controls from being modified (as per my feedback above)

desktop/hardened-desktop/index.md
```diff
@@ -0,0 +1,55 @@
...
+
+Docker plant to add more security enhancements to the Hardened Desktop security model over the coming months.
```
christophermclellan, line 21: i wouldn't specify 'over the coming months'. we'll be continuously adding enhancements here

desktop/hardened-desktop/index.md
```diff
@@ -0,0 +1,55 @@
+---
+title: Hardened Desktop
```
christophermclellan, line 2: looks great. some quite wordy feedback from me below but its mostly minor changes.

desktop/hardened-desktop/admin-controls/configure-ac.md
```diff
@@ -0,0 +1,129 @@
...
+## Step three: Re-launch and re-authenticate 
+Once you have created and configured `admin-settings.json`, developers need to quit Docker Desktop through the Whale menu, and then relaunch Docker Desktop and sign in to receive the changed settings.
+
+Docker doesn't automatically mandate that developers re-launch and re-authenticate once a change has been made so as not to disrupt your developers workflow. 
+
+## Example
+
+The following image displays an example `admin-settings.json` file:
+
+![admin-settings.json](../../images/admin-settings.PNG){:width="500px"}
```
christophermclellan, line 125: i'm sure you'll have noted this but let's not use this particular image when these docs goes live. i'd ask one of the Engineers to provide one where the details are correct.

desktop/hardened-desktop/admin-controls/index.md
```diff
@@ -0,0 +1,57 @@
+---
+description: admin controls for desktop
+keywords: admin controls, rootless, docker desktop, hardened desktop
+title: What is Admin Controls?
+--- 
+>Note
+>
+>Admin Controls is currently in [Early Access](../../../release-lifecycle.md#early-access-ea) and available to Docker Business customers only.
```
christophermclellan, line 8: Let's remove any mention of Early Access throughout the docs. We'll be immediately releasing all this functionality as GA.

desktop/hardened-desktop/admin-controls/index.md
```diff
@@ -0,0 +1,57 @@
...
+keywords: admin controls, rootless, docker desktop, hardened desktop
+title: What is Admin Controls?
+--- 
+>Note
+>
+>Admin Controls is currently in [Early Access](../../../release-lifecycle.md#early-access-ea) and available to Docker Business customers only. 
+
+Admin Controls is a feature that gives administrators the ability to enforce certain Docker Desktop settings for their organization. 
+
+With a few lines of JSON, administrators are able to enforce preferences for HTTP proxies, network settings, and the Docker Engine configuration. This saves significant time and cost in securing developer workflows.
```
christophermclellan, line 12: "admins are able to configure settings for Docker Desktop features such as proxies, network settings..."

desktop/hardened-desktop/admin-controls/index.md
```diff
@@ -0,0 +1,57 @@
...
+
+- For Organizations who wish to configure Docker Desktop to be within their organization's centralized control.
+- For Organizations who want to create a standardized Docker Desktop environment at scale.
+- For security conscious Docker Business customers who want to confidently manage their use of Docker Desktop within tightly regulated environments.
+
+### What can be set?
+
+Using the `admin-settings.json` file, administrators can:
+
+- Switch on Enhanced Container Isolation
```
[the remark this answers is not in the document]
christophermclellan: lets also link to the Enhanced Container Isolation docs here

desktop/hardened-desktop/admin-controls/index.md
christophermclellan: can we figure out a way to work Enhanced Container Isolation in here ? We're conveying that it's simple to configure admin settings and that it will thereby save time and cost. It would be good to also convey - 'and hey....when you enable Enhanced Container Isolation as part of this, it ensures that any configurations you create cannot be modified'

desktop/hardened-desktop/admin-controls/index.md
```diff
@@ -0,0 +1,57 @@
...
+- Enforce the use of WSL2 based engine or Hyper-V
+- Configure Docker Engine
+- Turn off Docker Desktop's ability to checks for updates
+- Turn off Docker Desktop's ability to send usage statistics
+
+For more details on the syntax and options administrators can set, see [Configure Admin Controls](configure-ac.md).
+
+### How is this different to the `settings.json` file?
+
+Using the `settings.json` file to pre-configure Docker Desktop settings means that developers own the `settings.json` file. They can therefore adjust any settings that their admins create, for example, modifying network and proxy controls.
```
christophermclellan, line 37: i'd change the first sentence to 'The settings.json is located in a directory that users can write to. As such, users can adjust any settings created by their admins in the settings.json, for example....'

desktop/hardened-desktop/admin-controls/index.md
```diff
@@ -0,0 +1,57 @@
...
+
+Next, you must [create and configure the admin-settings.json file](configure-ac.md).
+
+Once this is done, Docker Desktop users receive the changed settings when they re-launch Docker Desktop and re-authenticate to your organization. Docker doesn't automatically mandate that developers re-launch and re-authenticate once a change has been made, so as not to disrupt your developers workflow. 
+
+### What do users see when the settings are enforced?
+
+Docker Desktop users see a notification in the **Settings**, or **Preferences** if using a macOS, which states **Some settings are managed by your Admin**. 
+
+Any settings that are enforced, are grayed out in Docker Desktop and the user is unable to edit them, either via the Docker Desktop UI, CLI, or by modifying the Docker Desktop Linux VM.
```
christophermclellan, line 53: we need a screenshot here from Shadi or Trung

desktop/hardened-desktop/admin-controls/configure-ac.md
```diff
@@ -0,0 +1,129 @@
+---
+description: admin controls for desktop
+keywords: admin, controls, rootless, enhanced container isolation
+title: Configure Admin Controls
+--- 
+
+>Note
+>
+>Admin Controls is currently in [Early Access](../../../release-lifecycle.md#early-access-ea) and available to Docker Business customers only.
```
christophermclellan, line 9: same goes here, we need to remove any mention of Early Access

desktop/hardened-desktop/admin-controls/configure-ac.md
```diff
@@ -0,0 +1,129 @@
...
+
+## Prerequisite
+
+You need to [configure a registry.json to enforce sign-in](../../../docker-hub/configure-sign-in.md). For this configuration to take effect, Docker Desktop users must authenticate to your organization. 
+
+## Step one: Place the `admin-settings.json` file in the correct location
+
+Place the `admin-settings.json` file on your developers' machines in the following locations:
+
+- Mac: /Library/Application Support/com.docker.docker/admin-settings.json
```
christophermclellan, line 23: can we highlight these paths as code snippets for readability ?

desktop/hardened-desktop/admin-controls/configure-ac.md
```diff
@@ -0,0 +1,129 @@
...
+
+This page contains information about how administrators can configure Admin Controls to specify and lock configuration parameters to create a standardized Docker Desktop environment across the organization.
+
+Admin Controls is designed specifically for organizations who don’t give developers root access to their machines.
+
+## Prerequisite
+
+You need to [configure a registry.json to enforce sign-in](../../../docker-hub/configure-sign-in.md). For this configuration to take effect, Docker Desktop users must authenticate to your organization. 
+
+## Step one: Place the `admin-settings.json` file in the correct location
```
christophermclellan, line 19: Depending on what we're able to get done on time for launch, we might have an alternative way of 'placing the admin file in the correct location', which will be via an 'admin-settings' flag on the install CLI command for windows and mac.

desktop/hardened-desktop/admin-controls/configure-ac.md
```diff
@@ -0,0 +1,129 @@
...
+
+## Step two: Add the key value pairs for the admin controls you want to lock in
+
+>Note
+>
+>Some of the configuration parameters only apply to Windows. This is highlighted in the table below.
+
+The `admin-settings.json` file requires a nested list of configuration parameters, each of which must contain the  `locked` parameter. 
+If set to `true`, users are not able to edit this setting from Docker Desktop or the CLI. 
+If set to `false`, users can change this setting from Docker Desktop or the CLI by directly editing the `settings.json` file. If this setting is omitted, the default value is `false`.
```
christophermclellan, line 41: i think we should also mention here that admins don't have to enter every single key: value pair in the admin-settings.json for the feature to work correctly (note - i'm assuming that's the case but probably best to check with Manu !). 

e.g. if im an admin and i just want to configure proxy settings then i dont have to also enter values for 'useWindowsContainers', 'linuxVM', etc

desktop/hardened-desktop/admin-controls/configure-ac.md
```diff
@@ -0,0 +1,129 @@
...
+    "value": true
+  },
+}
+```
+
+| Parameter                        | Description                      |
+| :------------------------------- | :------------------------------- |
+| `configurationFileVersion`        | Specifies the version of the configuration file format.    |
+| `exposeDockerAPIOnTCP2375` |<span class="badge badge-info">Windows only</span> Exposes the Docker API on a specified port. If `value` is set to true, the Docker API is exposed on port 2375. Note: This is unauthenticated and should only be enabled if protected by suitable firewall rules.|
+| `proxy` | It is used for `http` and `https`.  If the port is custom, specify it in the property. |
```
christophermclellan, line 102: for keys like proxy, which take an object which has more keys, like "locked", "server", "port" etc, we should mention those things.

for each key throughout this doc that is not a boolean, we should mention the specific values that are accepted.

desktop/hardened-desktop/admin-controls/configure-ac.md
```diff
@@ -0,0 +1,129 @@
...
+| `dockerDaemonOptions`|If `value` is set to true, it overrides the options in the Linux daemon config file. See the [Docker Engine reference](../../../engine/reference/commandline/dockerd/#daemon-configuration-file). |
+| `vpnkitCIDR` |Overrides the network range used for vpnkit DHCP/DNS for `*.docker.internal`  |
+| (End of `linuxVM` section.)       |                                   |
+| `windowsContainers` | Parameters and settings related to `windowsContainers` options - grouped together here for convenience.                  |
+| `dockerDaemonOptions` | Overrides the options in the linux daemon config file. See the [Docker Engine reference](../../../engine/reference/commandline/dockerd/#daemon-configuration-file).|
+| (End of `windowsContainers` section.)    |                                   |
+|`disableUpdate`|If `value` is set to true, checking for and notifications about Docker Desktop updates is disabled.|
+|`analyticsEnabled`|If `value` is set to false, Docker Desktop does not send usage statistics to Docker. |
+
+## Step three: Re-launch and re-authenticate
```
christophermclellan, line 116: we should advise admins that after creating the admin-settings.json > quitting > re-launching / re-authenticating, that they should test out the functionality themselves locally to see if e.g. their new proxy settings work as expected.

desktop/hardened-desktop/index.md
ctalledo: Agree with Chris here.

desktop/hardened-desktop/index.md
ctalledo: Also Enhanced Container Isolation is not an admin-only setting; it can be set by developers (currently via the settings.json file, in the future probably through the UI). Admins can however configure the default value and lock the setting via the admin-settings.json file, just like other settings. (FYI @christophermclellan).

desktop/hardened-desktop/index.md
ctalledo: typo: "plant" -> "plans"

desktop/hardened-desktop/admin-controls/index.md
ctalledo: I would skip the "Expose daemon on tcp://localhost:2375" bullet in this list, because it actually reduces security of Docker Desktop.

desktop/hardened-desktop/enhanced-container-isolation/index.md
```diff
@@ -0,0 +1,61 @@
...
+Enhanced Container Isolation provides an additional layer of security that prevents containers from running as root in the Docker Desktop Linux VM. This ensures a strong container-to-host isolation and locks in any security configurations that have been created, for instance through Registry Access Management policies or with Admin Controls. 
+
+### Who is it for?
+
+- For organizations that want to prevent container attacks and reduce vulnerabilities.
+- For organizations that want ensure stronger container isolation that is easy and intuitive to implement on developers' machines.
+
+### What happens when Enhanced Container Isolation is switched on?
+
+When Enhanced Container Isolation is enabled using [Admin Controls](../admin-controls/index.md), developers can no longer:
```
ctalledo, line 20: This is worded in the negative sense, which appeals to admins but may scare developers. How about rewording it on a positive sense as follows?

"When Enhanced Container Isolation is enabled using Admin Controls, the following features are enabled:

* All user containers are automatically run in Linux User Namespaces (stronger isolation).
* The root user in the container maps to an unprivileged user at VM level.
* Users can continue using containers as usual, including bind-mounting host directories, volumes, networking configurations, etc. 
* Privileged containers work, but they are only privileged within the container's Linux User Namespace (not in the Docker Desktop VM).
* Containers can no longer share namespaces with the Docker Desktop VM (e.g., --network=host, --pid=host).
* Containers can no longer modify configuration files in the Docker Desktop VM.
* Containers become harder to breach (e.g., sensitive system calls are vetted and portions of /proc and /sys are emulated).

All of this is done automatically, without requiring any special user configuration and with minimal impact in performance and developer productivity. For more information on how Enhanced Container Isolation works, see [LINK-TO-SECTION].

desktop/hardened-desktop/enhanced-container-isolation/index.md
```diff
@@ -0,0 +1,61 @@
...
+```JSON
+{
+ "enhancedContainerIsolation": {
+    "value": true,
+    "locked": true
+    }
+}
+```
+
+Once this is done, Docker Desktop users receive the changed settings when they next authenticate to your organization on Docker Desktop. We do not automatically mandate that developers re-authenticate once a change has been made, so as not to disrupt your developers workflow.
```
ctalledo, line 53: We need indicate that a Docker Desktop re-launch is required too.

desktop/hardened-desktop/enhanced-container-isolation/index.md
```diff
@@ -0,0 +1,61 @@
...
+    "locked": true
+    }
+}
+```
+
+Once this is done, Docker Desktop users receive the changed settings when they next authenticate to your organization on Docker Desktop. We do not automatically mandate that developers re-authenticate once a change has been made, so as not to disrupt your developers workflow. 
+
+### What do users see when this setting is enforced?
+
+If Enhanced Container Isolation is enabled along with other settings in the `admin-settings.json`, users see a notification in the **Settings**, or **Preferences** using a macOS, which states **Some settings are managed by your Admin**.
```
ctalledo, line 57: Let's reword (the current wording applies to Admin Controls in general, not to Enhanced Container Isolation specifically).

I suggest:

"When Enhanced Container Isolation is enabled, users will see that containers run within a Linux User Namespace. For example:

```
$ docker run -it --rm alpine
/ # cat /proc/self/uid_map 
         0     100000      65536
```

This indicates that the container's root user (0) maps to unprivileged user (100000) in the Docker Desktop VM, and that the mapping extends for a range of 64K user-IDs. 

In contrast, without Enhanced Container Isolation the Linux user-namespace is not used:

```
$ docker run -it --rm alpine             
/ # cat /proc/self/uid_map                           
         0          0 4294967295
```

This means that the root user in the container (0) is in fact the root user in the Docker Desktop VM (0), reducing container isolation.

Note that Enhanced Container Isolation is in addition to other security techniques used by Docker (e.g., reduced Linux Capabilities, Seccomp, AppArmor, etc.) Refer to [LINK-TO-HOW-DOES-IT-WORK] for more info.

ctalledo (changes requested):
Some initial review comments. 

I'll work on adding more text to the "How does it work" and "FAQ/Known-issues" section tomorrow.

desktop/hardened-desktop/enhanced-container-isolation/index.md
christophermclellan: +1 ! would be great to call out all these positives.

desktop/hardened-desktop/enhanced-container-isolation/index.md
```diff
@@ -0,0 +1,61 @@
...
+>Note
+>
+>Enhanced Container Isolation is currently in [Early Access](../../release-lifecycle.md#early-access-ea) and available to Docker Business customers only. 
+
+Enhanced Container Isolation provides an additional layer of security that prevents containers from running as root in the Docker Desktop Linux VM. This ensures a strong container-to-host isolation and locks in any security configurations that have been created, for instance through Registry Access Management policies or with Admin Controls. 
+
+### Who is it for?
+
+- For organizations that want to prevent container attacks and reduce vulnerabilities.
+- For organizations that want ensure stronger container isolation that is easy and intuitive to implement on developers' machines.
```
christophermclellan, line 16: "that want to ensure"

desktop/hardened-desktop/enhanced-container-isolation/index.md
```diff
@@ -0,0 +1,61 @@
...
+- Bind mount and modify system files
+- Escape containers
+
+It also means all containers run unprivileged in the Docker Desktop Linux VM, in user namespaces. Root access to the Linux VM is removed, privileged containers cannot be run and there is no access to the host namespaces. As a result, it becomes impossible for users to alter any settings that have been locked in using [Admin Controls](../admin-controls/index.md).
+
+For more information on how Enhanced Container Isolation work, see [How does it work?](how-eci-works.md).
+
+>Important
+>
+>Enhanced Container Isolation is currently incompatible with WSL and does not protect Kubernetes pods. For more information on known limitations and workarounds, see [FAQS and known issues](faq.md).
```
christophermclellan, line 35: might be worth adding that we'll be working to address each of these incompatibilities in the near term.

desktop/hardened-desktop/enhanced-container-isolation/index.md
christophermclellan, line 57: Regarding _"If Enhanced Container Isolation is enabled along with other settings in the `admin-settings.json`, users see a notification in the **Settings**, or **Preferences** using a macOS, which states **Some settings are managed by your Admin**."_

Users won't actually see any of the above as Enhanced Container Isolation is not toggleable in the Docker Desktop UI. The only way to enable / disable is the via admin-settings.json, which only admins can use.

desktop/hardened-desktop/enhanced-container-isolation/index.md
```diff
@@ -0,0 +1,61 @@
...
+}
+```
+
+Once this is done, Docker Desktop users receive the changed settings when they next authenticate to your organization on Docker Desktop. We do not automatically mandate that developers re-authenticate once a change has been made, so as not to disrupt your developers workflow. 
+
+### What do users see when this setting is enforced?
+
+If Enhanced Container Isolation is enabled along with other settings in the `admin-settings.json`, users see a notification in the **Settings**, or **Preferences** using a macOS, which states **Some settings are managed by your Admin**. 
+
+As displayed in the image below, any settings that are enforced, are grayed out in Docker Desktop and the user is unable to edit them, either via the Docker Desktop UI, CLI, or by modifying the Docker Desktop Linux VM.
```
christophermclellan, line 59: Again, we won't need this paragraph as there's no corresponding UI element for ECI.

desktop/hardened-desktop/enhanced-container-isolation/how-eci-works.md
```diff
@@ -0,0 +1,21 @@
...
+
+Sysbox is an alternative runc used to create a container using the Linux kernel’s namespaces, cgroups, and other features. 
+
+Unlike the standard runc runtime, Sysbox enhances container isolation by using techniques such as enabling the Linux user-namespace on all containers, emulating portions of the proc filesystem and sysfs inside the container and vetting sensitive accesses between the container and the Linux kernel. This adds an extra layer of isolation between the container and the Linux kernel. 
+
+Without Enhanced Container Isolation, Docker Desktop has Docker Engine run as root with full capabilities inside a container that shares almost all namespaces with the Linux VM’s root user. Whilst this provides strong isolation between containers and the underlying host machine, it gives the container access to all the VM’s kernel resources and does not prevent Docker Desktop users from launching a container that runs as root in the Docker Desktop Linux VM, or from using insecure privileged containers. This brings Docker Desktop users closer to gaining privileged access to the underlying host.
+
+### How is this different to rootless mode in Docker Engine?
```
christophermclellan, line 19: I think we should flesh this out with some more questions that were addressed in the PR-FAQ. Specifically:

* With Hardened Desktop enabled, can the user still override the --runtime flag from the CLI ?
* With Hardened Desktop enabled, can the user still use the --privileged flag from the CLI ?
* Do I get any other benefits from Sysbox running under the hood ? (e.g. mention that users can run more advanced workloads with Sysbox) 

Cesar and Rodny will be able to guide on what should be included here.

desktop/hardened-desktop/enhanced-container-isolation/faq.md
```diff
@@ -0,0 +1,52 @@
...
+
+<hr>
+</div>
+<div id="tab4" class="tab-pane fade" markdown="1">
+
+#### Incompatibility with WSL
+Enhanced Container Isolation does not currently work when Docker Desktop runs on Windows with WSL/WSL2. This is due to some limitations of the WSL/WSL2 Linux Kernel. As a result, to use Enhanced Container Isolation on Windows, you must configure Docker Desktop to use Hyper-V.
+
+#### Kubernetes pods and extension containers are not yet protected
+ When Enhanced Container Isolation feature is enabled, Kubernetes pods and extension containers are not yet protected. A malicious or privileged pod or extension container can compromise the Docker Desktop Linux VM and bypass security controls.
```
christophermclellan, line 45: "When Enhanced Container Isolation is enabled"

desktop/hardened-desktop/enhanced-container-isolation/faq.md
```diff
@@ -0,0 +1,52 @@
...
+Privileged containers are typically used to run advanced workloads in containers, for example Docker-in-Docker, to perform kernel operations such as loading modules, or to access hardware devices. We aim to allow running advanced workloads, but deny the ability to perform kernel operations or access hardware devices.
+
+Allowing the `-–privileged` flag but restricting its impact within the container's user-namespace, it’s possible to do this.
+
+<hr>
+</div>
+<div id="tab4" class="tab-pane fade" markdown="1">
+
+#### Incompatibility with WSL
+Enhanced Container Isolation does not currently work when Docker Desktop runs on Windows with WSL/WSL2. This is due to some limitations of the WSL/WSL2 Linux Kernel. As a result, to use Enhanced Container Isolation on Windows, you must configure Docker Desktop to use Hyper-V.
```
christophermclellan, line 42: We could potentially mention here that via Admin Controls, admins can specify that Docker Desktop uses Hyper-V rather than WSL.

desktop/hardened-desktop/enhanced-container-isolation/faq.md
```diff
@@ -0,0 +1,52 @@
...
+
+No. With Hardened Desktop enabled, Sysbox is locked as the default (and only) runtime. If a user attempts to override the runtime by launching a container with the standard `runc` runtime, for example `docker run --runtime=runc`, container creation fails. 
+
+The reason `runc` is disallowed with Enhanced Container Isolation is because it allows users to run as root on the Docker Desktop Linux VM, thereby providing them with implicit control of the VM and the ability to modify the administrative configurations for Docker Desktop, for example.
+
+### With Enhanced Container Isolation enabled, can the user still use the `--privileged` flag from the CLI?
+
+Yes, with Enhanced Container Isolation the container is only privileged within its assigned Linux user-namespace. It is not privileged within the Docker Desktop Linux VM.
+
+For example, the container’s init process will have all Linux capabilities enabled, have read/write access to the kernel’s /proc and /sys, run without system call or other restrictions normally imposed by Docker on regular containers (for example, seccomp, AppArmor), and see all host devices under the container’s /dev directory.
```
christophermclellan, line 25: Can we highlight `proc`, `sys`, etc as code syntax ?

desktop/hardened-desktop/enhanced-container-isolation/faq.md
christophermclellan, line 45: We could also mention here that Kubernetes can be disabled via admin controls if admins are particularly concerned about this.

desktop/hardened-desktop/admin-controls/configure-ac.md
```diff
@@ -0,0 +1,129 @@
...
+| `enhancedContainerIsolation`  | If `value` is set to true, Docker Desktop prevents user containers from running as root and from being able to mount sensitive Docker Desktop configuration directories from the Docker Desktop VM. |
+|`useWindowsContainers` | <span class="badge badge-info">Windows only</span> If `value` is set to true, it switches Docker Desktop to toggle the Docker CLI to talk to the Windows daemon, enabling Windows containers. If false, switches Docker Desktop to toggle the Docker CLI to talk to the Linux daemon, enabling Linux containers. This overrides anything that may have been set at installation using the `--no-windows-containers` flag.|
+| `linuxVM` | Parameters and settings related to Linux VM options - grouped together here for convenience. |
+|`wslEngineEnabled`  |<span class="badge badge-info">Windows only</span> If `value` is set to true, Docker Desktop uses the WSL 2 based engine. This overrides anything that may have been set at installation using the `--backend=<backend name>` flag. It is also incompatible with Enhanced Container Isolation. See [Known issues](faq.md) for more information.|
```
christophermclellan, line 106: We need to more clearly denote that `wslEngineEnabled`, `dockerDaemonOptions` and `vpnkitCIDR` are nested values within the `linuxVM` section. The way this table [looks on  this page](https://deploy-preview-15775--docsdocker.netlify.app/desktop/hardened-desktop/admin-controls/configure-ac/) makes it seem like each of these value are at the same level in the JSON hierarchy.

desktop/hardened-desktop/admin-controls/configure-ac.md
christophermclellan: What happens to the windows values on macOS ? Can we call out that values that are irrelevant to your platform will be ignored ?

desktop/hardened-desktop/admin-controls/configure-ac.md
```diff
@@ -0,0 +1,129 @@
...
+| `dockerDaemonOptions`|If `value` is set to true, it overrides the options in the Linux daemon config file. See the [Docker Engine reference](../../../engine/reference/commandline/dockerd/#daemon-configuration-file). |
```
christophermclellan, line 107: as per [this comment](https://github.com/docker/docs/pull/15775/files#r984642121), for `dockerDaemonOptions` can we provide an example JSON snippet with the quotation marks in the JSON escaped (so it's clear to admins what sort of format will be valid).

desktop/hardened-desktop/admin-controls/configure-ac.md
christophermclellan, line 41: regarding `locked: false`, we should note that this is akin to setting a `factory default`. the logic for `locked: false` is as follows:

1. For new installs, `locked: false` will pre-populate the relevant settings in the Desktop UI

2. For machines on which Desktop is already installed and being used, `locked:false` is ignored. This is because existing users of Desktop may have already updated a setting (e.g. their proxy). In these instances, the setting created by the user will have been written to the relevant config file (e.g. `settings.json` or `daemon.json`) and respect that user preference. Unless of course the admin chooses to override by setting a `locked: true`.

Hope that makes sense ! Further context within the comments on [DKM-29](https://docker.atlassian.net/browse/DKM-29).

ctalledo: Regarding: "For machines on which Desktop is already installed and being used, locked:false is ignored .. because existing users ... may have already updated a setting"

What happens for settings that were not update by the user? Is `locked: false` ignored in that case?

christophermclellan: My understanding from previous chat with Manu, is that for settings that were not updated by the user, `locked:false` is also ignored in that case. Details are in the comments here - https://docker.atlassian.net/browse/DKM-29.

desktop/hardened-desktop/enhanced-container-isolation/faq.md
aevesdocker: @christophermclellan I thought that Kubernetes was not being implemented for 4.13? 🤔

christophermclellan: ha, yes, good point ! there was a chance that it might have been included as Manu has picked up some of the remaining items but that's now looking increasingly unlikely.

desktop/hardened-desktop/index.md
```diff
@@ -0,0 +1,55 @@
...
+description: Overview of what Hardened Desktop is
+keywords: security, hardened desktop, enhanced container isolation, registry access management, admin controls, root access, administrators, docker desktop
+---
+>Note
+>
+>Hardened Desktop is available to Docker Business customers only.
+
+Hardened Desktop is a new security model for Docker Desktop. It's designed to provide administrators with a simple and powerful way to improve their organizations security posture for containerised development, without impacting the developer experience that Docker Desktop offers.
```
christophermclellan, line 10: sorry, annoying feedback here, but I think we should change 'administrators' to 'admins' throughout the docs. 

'Administrators' seems too formal. Unless we have a style guide or something which says that 'administrators' is preferred terminology and it's used elsewhere throughout the docs.

desktop/hardened-desktop/index.md
```diff
@@ -0,0 +1,55 @@
...
+This configuration is designed for security conscious organizations who don’t give their users root or admin access on their machines, and who would like Docker Desktop to be within the organization’s centralized control.
+
+The Hardened Desktop security model moves the ownership boundary for containers to the organization, meaning that any security controls administrators set cannot be altered by the user of Docker Desktop.
+
+Hardened Desktop includes:
+- Admin Controls, which helps administrators to confidently manage and control the usage of Docker Desktop within their organization.
+- Enhanced Container Isolation, a setting that instantly enhances security by preventing containers from running as root in Docker Desktop’s Linux VM and ensures that configurations set using Admin Controls cannot be modified by containers.
+- Registry Access Management, which allows administrators to control the registries developers can access.
+
+Docker plans to continue adding more security enhancements to the Hardened Desktop security model.
```
christophermclellan, line 21: @ctalledo @aevesdocker - i wonder if we can add a little more detail here. e.g. `Security is a moving target and as such, Docker will be continuously adding enhancements to Hardened Desktop and it's underlying features'

desktop/hardened-desktop/admin-controls/index.md
```diff
@@ -0,0 +1,55 @@
...
+keywords: admin controls, rootless, docker desktop, hardened desktop
+title: What is Admin Controls?
+--- 
+>Note
+>
+>Admin Controls is available to Docker Business customers only. 
+
+Admin Controls is a feature that gives administrators the ability to configure certain Docker Desktop settings for their organization. 
+
+With a few lines of JSON, administrators can configure Docker Desktop settings for HTTP proxies, network settings, and the Docker Engine configuration. This saves significant time and cost in securing developer workflows. If administrators also enable [Enhanced Container Isolation](../enhanced-container-isolation/index.md), it ensures that any configurations set with Admin Controls cannot be modified by containers.
```
christophermclellan, line 12: '...settings for features such as proxies, network settings and the Docker Engine configuration (for the full list, see 'What can be set ?' below)

desktop/hardened-desktop/admin-controls/index.md
```diff
@@ -0,0 +1,55 @@
...
+- For Organizations who want to create a standardized Docker Desktop environment at scale.
+- For security conscious Docker Business customers who want to confidently manage their use of Docker Desktop within tightly regulated environments.
+
+## How does it work?
+
+Administrators can configure several Docker Desktop settings using the `admin-settings.json` file. This file is located within the Docker Desktop host and can only be accessed by users with root or admin privileges. 
+
+As the `admin-settings.json` overrides the `settings.json` file, which contains the settings configured by regular Docker Desktop users and is located in a directory users can write to, administrators can configure default values for several Docker Desktop settings and lock them in so regular users are prevented from modifying them.
+
+### What can be set?
```
christophermclellan, line 26: Can we change this to `What features can I configure with Admin Controls ?` ?

desktop/hardened-desktop/admin-controls/index.md
christophermclellan: can we remove the sentence about 'This saves significant time...workflows' and instead have something at the end of the previous sentence such as ", and the Docker Engine configuration, sparing significant effort in securing developer workflows"

not sure if 'sparing significant effort' is the right wording !

desktop/hardened-desktop/admin-controls/index.md
```diff
@@ -0,0 +1,55 @@
...
+
+For more details on the syntax and options administrators can set, see [Configure Admin Controls](configure-ac.md).
+
+### How do I set up and enforce Admin Controls?
+
+As an administrator, you first need to [configure a registry.json to enforce sign-in](../../../docker-hub/configure-sign-in.md). This is because your Docker Desktop users must authenticate to your organization for this configuration to take effect.
+
+Next, you must [create and configure the admin-settings.json file](configure-ac.md).
+
+Once this is done, Docker Desktop users receive the changed settings when they re-launch Docker Desktop and re-authenticate to your organization. Docker doesn't automatically mandate that developers re-launch and re-authenticate once a change has been made, so as not to disrupt your developers workflow.
```
christophermclellan, line 46: Instead of 're-launch' i think we explicitly need to state that its a 'Quit and Restart' of Docker Desktop that's required. @ctalledo what's the appropriate terminology ?

desktop/hardened-desktop/admin-controls/configure-ac.md
```diff
@@ -0,0 +1,144 @@
...
+
+## Step two: Configure the admin controls you want to lock in
+
+>Note
+>
+>Some of the configuration parameters only apply to Windows. This is highlighted in the table below.
+
+The `admin-settings.json` file requires a nested list of configuration parameters, each of which must contain the  `locked` parameter. You can add or remove configuration parameters as per your requirements.
+If set to `true`, users are not able to edit this setting from Docker Desktop or the CLI. 
+If set to `false`, the configuration value acts as a default value, but users can change this setting from Docker Desktop or the CLI by directly editing the `settings.json` file.
```
christophermclellan, line 41: OK so this + table below are great but I'd restructure slightly:

Paragraph 1 - "The `admin-settings.json` file requires...as per your requirements"

Paragraph 2 - "If `locked:true`, users will not be able to modify this setting via Docker Desktop.

Paragraph 3:

"Setting a value as `locked: false` is similar to setting a factory default, in that:

* For new installs, `locked: false pre-populated the relevant setting in the Docker Desktop UI, but users will be able to modify it as they please

* When Docker Desktop is already installed and being used, `locked: false is ignored. This is because existing users of Docker Desktop may have already updated a setting, which in turn will have been written to the relevant config file, for example the `settings.json` or `daemon.json`. In these instances, the user's preferences are respected and we do not alter these values. These can be controlled by the administrator by setting `locked:true`

christophermclellan: And then I'd remove the 'Addition locked: false logic' table entirely. Let's just pull out the relevant details from that (which are important) and make them visible.

desktop/hardened-desktop/admin-controls/configure-ac.md
```diff
@@ -0,0 +1,144 @@
...
+| &nbsp; &nbsp; &nbsp; &nbsp;`dockerDaemonOptions` | Overrides the options in the linux daemon config file. See the [Docker Engine reference](../../../engine/reference/commandline/dockerd/#daemon-configuration-file).|                                |
+|`disableUpdate`|If `value` is set to true, checking for and notifications about Docker Desktop updates is disabled.|
+|`analyticsEnabled`|If `value` is set to false, Docker Desktop does not send usage statistics to Docker. |
+
+## Step three: Re-launch and re-authenticate
+>Note
+>
+>Administrators should test the changes made through the `admin-settings.json` file locally to see if the settings work as expected.
+
+Once you have created and configured `admin-settings.json`, developers need to quit Docker Desktop through the Whale menu, and then relaunch Docker Desktop and sign in to receive the changed settings. Selecting **Restart** from the Whale menu isn't sufficient as it only restarts some components of Docker Desktop.
```
christophermclellan, line 132: We need to be careful here as this slightly nuanced.

For settings to take effect:

1. On a new install, developers just have to open Docker Desktop and authenticate to their org

2. On an existing install, developers have to quit (via the whale menu) and restart Docker Desktop. They must also ensure they are signed in to their org. (but they don't have to reauthenticate to their org)


@ctalledo please keep me honest here !

christophermclellan: btw - im not saying that we should structure it as a numbered list, just trying to flag how this works for either use case.

desktop/hardened-desktop/enhanced-container-isolation/index.md
```diff
@@ -0,0 +1,89 @@
...
+>Note
+>
+>Enhanced Container Isolation is available to Docker Business customers only. 
+
+Enhanced Container Isolation provides an additional layer of security that uses a variety of advanced techniques to harden container isolation without impacting developer productivity. 
+
+These techniques include:
+- Running all containers unprivileged (via the Linux user-namespace)
+- Restricting containers from modifying Docker Desktop VM settings
+- Vetting some critical system calls to prevent container escapes, and partially virtualizing portions of `/proc` and c/sys` inside the container for further isolation.
```
christophermclellan, line 16: c/sys requires the opening quotation

desktop/hardened-desktop/enhanced-container-isolation/index.md
```diff
@@ -0,0 +1,89 @@
...
+- Restricting containers from modifying Docker Desktop VM settings
+- Vetting some critical system calls to prevent container escapes, and partially virtualizing portions of `/proc` and c/sys` inside the container for further isolation. 
+
+This is all done automatically and with minimal performance impact. 
+
+Enhanced Container Isolation helps ensure a strong container-to-host isolation and locks in any security configurations that have been created, for instance through [Registry Access Management policies](../registry-access-management.md) or with [Admin Controls](../admin-controls/index.md). 
+
+>Note
+>
+> Enhanced Container Isolation is in addition to other security techniques used by Docker. For example, reduced Linux Capabilities, Seccomp, AppArmor
```
christophermclellan, line 24: ", Seccomp and AppArmor."

desktop/hardened-desktop/index.md
aevesdocker: @christophermclellan I wonder if this is actually the place for the sentence at all. This seems more like it belongs in marketing/blogs rather than docs which is meant to represent the 'here and now', rather than the 'what's coming soon'

christophermclellan: yeh, its a good point. ok let's leave it as is. i think having the original 'docker plans to continue adding...' is still useful as we want to convey that 'this is not completely, 100% hardened, and that we still have some things we want to work on'.

desktop/hardened-desktop/admin-controls/configure-ac.md
aevesdocker: Unable to have that in the table, markdown table unfortunately has very limited formatting options. Would the value Manu provided be enough?

desktop/hardened-desktop/admin-controls/index.md
```diff
@@ -0,0 +1,54 @@
...
+keywords: admin controls, rootless, docker desktop, hardened desktop
+title: What is Admin Controls?
+--- 
+>Note
+>
+>Admin Controls is available to Docker Business customers only. 
+
+Admin Controls is a feature that gives admins the ability to configure certain Docker Desktop settings for their organization. 
+
+With a few lines of JSON, admins can configure Docker Desktop settings for HTTP proxies, network settings, and the Docker Engine configuration, which reduces effort spent on securing developer workflows. If admins also enable [Enhanced Container Isolation](../enhanced-container-isolation/index.md), it ensures that any configurations set with Admin Controls cannot be modified by containers.
```
christophermclellan, line 12: Hey @aevesdocker , could you please adjust this sentence to note that there are more settings than just proxies / network / engine

_data/toc.yaml
```diff
@@ -1272,6 +1272,26 @@ manuals:
           title: Run Docker Desktop for Windows in a VM or VDI environment
         - path: /desktop/uninstall/
           title: Uninstall Docker Desktop
+    - sectiontitle: Hardened Desktop
+      section:
+        - path: /desktop/hardened-desktop/
+          title: Overview
+        - sectiontitle: Admin controls (Early Access)
```
christophermclellan, line 1279: Please remove 'Early Access'

_data/toc.yaml
```diff
@@ -1272,6 +1272,26 @@ manuals:
...
+          section:
+          - path: /desktop/hardened-desktop/admin-controls/
+            title: What is Admin Controls?
+          - path: /desktop/hardened-desktop/admin-controls/configure-ac/
+            title: Configure Admin controls
+        - sectiontitle: Enhanced Container Isolation (Early Access)
```
christophermclellan, line 1285: we also need to remove 'early access' here (and anywhere else throughout the docs)

desktop/hardened-desktop/admin-controls/index.md
```diff
@@ -0,0 +1,54 @@
...
+
+Administrators can configure several Docker Desktop settings using the `admin-settings.json` file. This file is located within the Docker Desktop host and can only be accessed by users with root or admin privileges. 
+
+Values that are set to `locked: true` within the `admin-settings.json` override any previous values set by users and ensure that these cannot be modified. For more information, see [Configure Admin Controls](../admin-controls/configure-ac.md#step-two-configure-the-admin-controls-you-want-to-lock-in).
+
+### What features can I configure with Admin Controls?
+
+Using the `admin-settings.json` file, admins can:
+
+- Enable [Enhanced Container Isolation](../enhanced-container-isolation/index.md)
```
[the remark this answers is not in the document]
christophermclellan: or have a sub-bullet point noting incompatibility with WSL

desktop/hardened-desktop/admin-controls/configure-ac.md
```diff
@@ -0,0 +1,130 @@
...
+  },
+}
+```
+
+| Parameter                        | Description                      |
+| :------------------------------- | :------------------------------- |
+| `configurationFileVersion`        | Specifies the version of the configuration file format.    |
+| `exposeDockerAPIOnTCP2375` |<span class="badge badge-info">Windows only</span> Exposes the Docker API on a specified port. If `value` is set to true, the Docker API is exposed on port 2375. Note: This is unauthenticated and should only be enabled if protected by suitable firewall rules.|
+| `proxy` | It is used for `http` and `https`.  If the port is custom, specify it in the property. |
+| `enhancedContainerIsolation`  | If `value` is set to true, Docker Desktop runs all containers as unprivileged, via the Linux user-namespace, prevents them from modifying sensitive configurations inside the Docker Desktop VM, and uses other advanced techniques to isolate them. For more information, see [Enhanced Container Isolation](../enhanced-container-isolation/index.md). |
```
christophermclellan, line 108: we should also note here that ECI is currently not compatible with WSL

desktop/hardened-desktop/admin-controls/configure-ac.md
```diff
@@ -0,0 +1,130 @@
...
+    }
+  },
+  "disableUpdate": {
+    "locked": false,
+    "value": false
+  },
+  "analyticsEnabled": {
+    "locked": false,
+    "value": true
+  },
```
christophermclellan, line 99: @aevesdocker can you remove the trailing comma here as this makes the JSON invalid ?

desktop/hardened-desktop/enhanced-container-isolation/index.md
```diff
@@ -0,0 +1,87 @@
...
+- Restricting containers from modifying Docker Desktop VM settings
+- Vetting some critical system calls to prevent container escapes, and partially virtualizing portions of `/proc` and `/sys` inside the container for further isolation. 
+
+This is done automatically and with minimal performance impact. 
+
+Enhanced Container Isolation helps ensure a strong container-to-host isolation and locks in any security configurations that have been created, for instance through [Registry Access Management policies](../registry-access-management.md) or with [Admin Controls](../admin-controls/index.md). 
+
+>Note
+>
+> Enhanced Container Isolation is in addition to other security techniques used by Docker. For example, reduced Linux Capabilities, Seccomp, AppArmor.
```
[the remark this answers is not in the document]
aevesdocker: Haven't added it in this section as it is already called out in the 'What happens when....' section,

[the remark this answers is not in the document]
christophermclellan: Roger that !

ctalledo (commented):
Some more doc comments.

desktop/hardened-desktop/enhanced-container-isolation/index.md
```diff
@@ -0,0 +1,92 @@
...
+
+When Enhanced Container Isolation is enabled using [Admin Controls](../admin-controls/index.md), the following features are enabled: 
+
+- All user containers are automatically run in Linux User Namespaces which ensures stronger isolation.
+- The root user in the container maps to an unprivileged user at VM level.
+- Users can continue using containers as usual, including bind-mounting host directories, volumes, networking configurations, etc.
+- Privileged containers work, but they are only privileged within the container's Linux User Namespace, not in the Docker Desktop VM.
+- Containers can no longer share namespaces with the Docker Desktop VM. For example, `--network=host`, `--pid=host`.
+- Containers can no longer modify configuration files in the Docker Desktop VM.
+- Containers become harder to breach. For example, sensitive system calls are vetted and portions of `/proc` and `/sys` are emulated.
```
rodnymolina, line 41: Looks like this section is technical enough for these details below, otherwise we need a separate section to cover this stuff as we need to explain why certain workflows won't work anymore for security reasons.

- The docker-engine's configuration file is automatically modified, even if matching clauses are already present, in which case they are overridden. Refer to [Docker Engine Configuration Guide](https://docs.docker.com/config/daemon) for details.
    * The `runtimes` section displays a single runtime: `sysbox-runc`. Notice that while `runc` is still the default runtime, this one is not utilized to launch regular containers.
    * The `entitlements` section is explicitly defined to forbid the use of `--network=host` and `--security-insecure` flags during `docker build` attempts.

desktop/hardened-desktop/admin-controls/configure-ac.md
```diff
@@ -0,0 +1,137 @@
...
+|`useWindowsContainers` | <span class="badge badge-info">Windows only</span> If `value` is set to true, it switches Docker Desktop to toggle the Docker CLI to talk to the Windows daemon, enabling Windows containers. If false, switches Docker Desktop to toggle the Docker CLI to talk to the Linux daemon, enabling Linux containers. This overrides anything that may have been set at installation using the `--no-windows-containers` flag.|
+| `linuxVM` | Parameters and settings related to Linux VM options - grouped together here for convenience. |
+| &nbsp; &nbsp; &nbsp; &nbsp;`wslEngineEnabled`  |<span class="badge badge-info">Windows only</span> If `value` is set to true, Docker Desktop uses the WSL 2 based engine. This overrides anything that may have been set at installation using the `--backend=<backend name>` flag. It is also incompatible with Enhanced Container Isolation. See [Known issues](faq.md) for more information.|
+| &nbsp;&nbsp; &nbsp; &nbsp;`dockerDaemonOptions`|If `value` is set to true, it overrides the options in the Linux daemon config file. See the [Docker Engine reference](../../../engine/reference/commandline/dockerd/#daemon-configuration-file). |
```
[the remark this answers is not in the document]
rodnymolina: Also, this is the proper link:

[Docker Engine Configuration Guide](https://docs.docker.com/config/daemon)

rodnymolina (commented):
Sorry, I'm late here. Excellent job with the documentation, wow!

I added a few comments below.

desktop/hardened-desktop/admin-controls/configure-ac.md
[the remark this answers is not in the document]
aevesdocker: Thanks @rodnymolina for the comments above! Have added the top comment. Re the broken link: some of the engine docs are stored in an upstream repo which is pulled in when the docs are built. It results in links appearing broken when we use relative links. It will resolve itself when the PR is merged

desktop/hardened-desktop/enhanced-container-isolation/index.md
aevesdocker: To be covered in the FAQs/Known issues section @ctalledo?

desktop/hardened-desktop/enhanced-container-isolation/faq.md
```diff
@@ -0,0 +1,52 @@
...
+<ul class="nav nav-tabs">
+  <li class="active"><a data-toggle="tab" data-target="#tab3">FAQs</a></li>
+  <li><a data-toggle="tab" data-target="#tab4">Known issues</a></li>
+</ul>
+<div class="tab-content">
+<div id="tab3" class="tab-pane fade in active" markdown="1">
+
+### With Enhanced Container Isolation enabled, can the user still override the `--runtime` flag from the CLI ?
+
+No. With Hardened Desktop enabled, Sysbox is locked as the default (and only) runtime. If a user attempts to override the runtime by launching a container with the standard `runc` runtime, for example `docker run --runtime=runc`, container creation fails.
```
rodnymolina, line 17: There's a correction needed here:

If a user attempts to override the runtime by launching a container with the standard `runc` runtime, for example `docker run --runtime=runc`, this request is ignored and the container is created through the Sysbox runtime.

desktop/hardened-desktop/enhanced-container-isolation/faq.md
```diff
@@ -0,0 +1,52 @@
...
+
+This makes running a privileged container with Enhanced Container Isolation much safer than a privileged container launched with the standard runc, which offers almost no isolation.
+
+### Why not just restrict usage of the `--privileged` flag with Enhanced Container Isolation?
+
+Privileged containers are typically used to run advanced workloads in containers, for example Docker-in-Docker, to perform kernel operations such as loading modules, or to access hardware devices. We aim to allow running advanced workloads, but deny the ability to perform kernel operations or access hardware devices.
+
+Allowing the `-–privileged` flag but restricting its impact within the container's user-namespace, it’s possible to do this.
```
rodnymolina, line 35: This sentence is missing something, maybe a "by" at the beginning...

"By allowing the `-–privileged` flag but restricting its impact within the container's user-namespace, it’s possible to do this."

desktop/hardened-desktop/enhanced-container-isolation/faq.md
```diff
@@ -0,0 +1,52 @@
...
+
+<hr>
+</div>
+<div id="tab4" class="tab-pane fade" markdown="1">
+
+#### Incompatibility with WSL
+Enhanced Container Isolation does not currently work when Docker Desktop runs on Windows with WSL/WSL2. This is due to some limitations of the WSL/WSL2 Linux Kernel. As a result, to use Enhanced Container Isolation on Windows, you must configure Docker Desktop to use Hyper-V. This can be enforced using Admin Controls. For more information, see [Admin Controls](../admin-controls/index.md).
+
+#### Kubernetes pods and extension containers are not yet protected
+When Enhanced Container Isolation is enabled, Kubernetes pods and extension containers are not yet protected. A malicious or privileged pod or extension container can compromise the Docker Desktop Linux VM and bypass security controls.
```
rodnymolina, line 45: Dev-environments is another category not being protected (for now).

desktop/hardened-desktop/admin-controls/index.md
```diff
@@ -0,0 +1,58 @@
...
+
+Admin Controls is a feature that helps admins to control certain Docker Desktop settings on client machines within their organization.
+
+With a few lines of JSON, admins can configure controls for Docker Desktop settings such as proxies and network settings. For an extra layer of security, admins can also use Admin Controls to enable [Enhanced Container Isolation](../enhanced-container-isolation/index.md) which ensures that any configurations set with Admin Controls cannot be modified by containers.
+
+### Who is it for? 
+
+- For Organizations who wish to configure Docker Desktop to be within their organization's centralized control.
+- For Organizations who want to create a standardized Docker Desktop environment at scale.
+- For security conscious Docker Business customers who want to confidently manage their use of Docker Desktop within tightly regulated environments.
```
dvdksn, line 18:
```suggestion
- For Docker Business customers who want to confidently manage their use of Docker Desktop within tightly regulated environments.
```

desktop/hardened-desktop/enhanced-container-isolation/index.md
```diff
@@ -0,0 +1,94 @@
...
+
+This indicates that the container's root user (0) maps to unprivileged user (100000) in the Docker Desktop VM, and that the mapping extends for a range of 64K user-IDs.
+
+In contrast, without Enhanced Container Isolation the Linux user-namespace is not used:
+
+```
+$ docker run -it --rm alpine             
+/ # cat /proc/self/uid_map                           
+         0          0 4294967295
+```
```
dvdksn, line 92: Make this command a one-liner?

```suggestion
```console
$ docker run -it --rm alpine cat /proc/self/uid_map                           
         0          0 4294967295
```
```

**Review by dvdksn — COMMENTED** [review]
Awesome work @aevesdocker -- I only have minor comments/suggestions.

PS: I think the image at `/desktop/images/admin-settings.PNG` can be removed?

File: desktop/hardened-desktop/settings-management/configure.md [hunk 56]
```diff
@@ -0,0 +1,139 @@
...
+- [Download and install Docker Desktop 4.13.0 or later](../../release-notes.md).
+- As an admin, you need to [configure a registry.json to enforce sign-in](../../../docker-hub/configure-sign-in.md). This is because this feature requires a Docker Business subscription and therefore your Docker Desktop users must authenticate to your organization for this configuration to take effect.
+
+### Step one: Create the `admin-settings.json` file and save it in the correct location
+
+1. Create a new, empty JSON file and name it `admin-settings`.
+2. Save the `admin-settings.json` file on your developers' machines in the following locations:
+
+    - Mac: `/Library/Application\ Support/com.docker.docker/admin-settings.json`
+    - Windows: `/ProgramData/DockerDesktop/admin-settings.json`
```
ctalledo, line 26: On Windows paths use "\" instead of "/". Please change to:

Windows: `C:\ProgramData\DockerDesktop\admin-settings.json`

ctalledo (changes requested):
nit: please fix Windows path for admin-settings.json file.

desktop/hardened-desktop/enhanced-container-isolation/index.md
```diff
@@ -0,0 +1,98 @@
...
+
+This indicates that the container's root user (0) maps to unprivileged user (100000) in the Docker Desktop VM, and that the mapping extends for a range of 64K user-IDs.
+
+In contrast, without Enhanced Container Isolation the Linux user-namespace is not used, the following displays:
+
+```
+         0          0 4294967295
+```
+
+This means that the root user in the container (0) is in fact the root user in the Docker Desktop VM (0) which reduces container isolation. If a process were to escape the container, it would find itself without privileges at the VM level. For further details, see [How Enhanced Container Isolation works](how-eci-works.md).
```
ctalledo, line 98: "If a process were to escape the container ..." -> "With Enhanced Container Isolation, if a process were to escape the container ..."

ctalledo (commented):
Some more suggestions for ECI (though it's looking pretty good already!)

desktop/hardened-desktop/enhanced-container-isolation/index.md
```diff
@@ -0,0 +1,98 @@
...
+$ docker run -it --rm alpine / # cat /proc/self/uid_map 
+```
+
+The following output displays:
+
+```
+         0     100000      65536
+```
+
+This indicates that the container's root user (0) maps to unprivileged user (100000) in the Docker Desktop VM, and that the mapping extends for a range of 64K user-IDs.
```
ctalledo, line 90: After this sentence please add:

"If you create more containers, the user-ID mapping will vary as each container gets an exclusive range of host User-IDs for isolation. This mapping automatically managed by Docker Desktop."

desktop/hardened-desktop/enhanced-container-isolation/index.md
ctalledo: Also, let's add this in a new paragraph:

Enhanced Container Isolation uses the Sysbox container runtime embedded in the Docker Desktop Linux VM (Docker Business Customers only). Therefore another way to determine if a container is running with Enhanced Container Isolation is by using `docker inspect`:

```
docker inspect --format='{{.HostConfig.Runtime}}' my_container
sysbox-runc
```

Without Enhanced Container Isolation, this command will instead output `runc` (the standard OCI runtime).

desktop/hardened-desktop/enhanced-container-isolation/index.md
```diff
@@ -0,0 +1,98 @@
...
+>**Note**
+>
+> Enhanced Container Isolation is in addition to other container security techniques used by Docker. For example, reduced Linux Capabilities, Seccomp, AppArmor.
+
+### Who is it for?
+
+- For organizations that want to prevent container attacks and reduce vulnerabilities.
+- For organizations that want to ensure stronger container isolation that is easy and intuitive to implement on developers' machines.
+
+### What happens when Enhanced Container Isolation is enabled?
```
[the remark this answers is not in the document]
aevesdocker: I'm gonna disagree with you here @ctalledo! I think users like to know exactly what they're getting into before they make up their minds about enabling a feature. Also, if I swapped the sections around, the following section 'what do users see' would seem out of place.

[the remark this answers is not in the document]
ctalledo: Sounds good @aevesdocker.

desktop/hardened-desktop/settings-management/configure.md
```diff
@@ -0,0 +1,139 @@
...
+    "exclude": []
+  },
+  "enhancedContainerIsolation": {
+    "locked": true,
+    "value": true
+  },
+  "useWindowsContainers": {
+      "locked": false,
+      "value": false
+  },
```
ebriney, line 73: this is not implemented in 4.13

desktop/hardened-desktop/settings-management/configure.md
```diff
@@ -0,0 +1,139 @@
...
+| `enhancedContainerIsolation`  |  | If `value` is set to true, Docker Desktop runs all containers as unprivileged, via the Linux user-namespace, prevents them from modifying sensitive configurations inside the Docker Desktop VM, and uses other advanced techniques to isolate them. For more information, see [Enhanced Container Isolation](../enhanced-container-isolation/index.md). Note: Enhanced Container Isolation is currently [incompatible with WSL](../enhanced-container-isolation/faq.md#incompatibility-with-wsl). |
+|`useWindowsContainers` | <span class="badge badge-info">Windows only</span> |If `value` is set to true, it switches Docker Desktop to toggle the Docker CLI to talk to the Windows daemon, enabling Windows containers. If false, switches Docker Desktop to toggle the Docker CLI to talk to the Linux daemon, enabling Linux containers. This overrides anything that may have been set at installation using the `--no-windows-containers` flag.|
```
ebriney, line 111: this is not implemented in 4.13

desktop/hardened-desktop/settings-management/configure.md
```diff
@@ -0,0 +1,139 @@
...
+>**Note**
+>
+>Administrators should test the changes made through the `admin-settings.json` file locally to see if the settings work as expected.
+
+For settings to take effect:
+- On a new install, developers need to launch Docker Desktop and authenticate to their organization.
+- On an existing install, developers need to quit Docker Desktop through the Docker menu, and then relaunch Docker Desktop and sign in to receive the changed settings. 
+  >**Important**
+  >
+  >Selecting **Restart** from the Docker menu isn't enough as it only restarts some components of Docker Desktop.
```
ebriney, line 131: NB: this is a design choice, we can change this if we think UX is not intuitive.
But as far as changing admin settings must not interrupt the developer workflow, I think it's works fine like that. WDYT @ctalledo

desktop/hardened-desktop/registry-access-management.md
```diff
@@ -0,0 +1,62 @@
...
+---
+
+>Note
+>
+>Registry Access Management is available to Docker Business customers only. 
+
+With Registry Access Management, administrators can ensure that their developers using Docker Desktop only access registries that are allowed. This is done through the Registry Access Management dashboard on Docker Hub. 
+
+Below are some example registries administrators can allow: 
+ - Amazon ECR
```
christophermclellan, line 16: Docker Hub should be the first item on this list. Can we also note that Docker Hub will be enabled by default ?

desktop/hardened-desktop/registry-access-management.md
```diff
@@ -0,0 +1,62 @@
...
+ - GitHub Container Registry
+ - Google Container Registry
+
+Administrators can ensure registries are locked in and cannot be edited by developers, if Enhanced Container Isolation is switched on with Enhanced Container Isolation. To learn more, see [Enhanced Container Isolation](enhanced-container-isolation/index.md).
```
christophermclellan, line 20: Double mention of Enhanced Container Isolation in this sentence.

desktop/install/mac-install.md
```diff
@@ -100,6 +100,10 @@ The `install` command accepts the following flags:
...
 - `--allowed-org=<org name>`: requires the user to sign in and be part of the specified Docker Hub organization when running the application
 - `--user=<username>`: Runs the privileged helper service once during installation, then disables it at runtime. This removes the need for the user to grant root privileges on first run. For more information, see [Privileged helper permission requirements](../mac/permission-requirements.md#permission-requirements){: target="_blank" rel="noopener" class="_"}. To find the username, enter `ls /Users` in the CLI.
+- `--admin-settings`: Automatically creates an `admin-settings.json` file which is used by admins to control certain Docker Desktop settings on client machines within their organization. For more information, see [Settings Management](../hardened-desktop/settings-management/index.md).
+  - It must be used together with the `--allowed-org=<org name>` flag. 
+  - For example:
+    `--allowedOrg=<org name> --admin-settings='{"configurationFileVersion": 2, "enhancedContainerIsolation": {"value": true, "locked": false}}'`
```
ebriney, line 106: sorry I did a typo it's `--allowed-org=<org name>` instead of  `--allowedOrg=<org name>`

desktop/install/windows-install.md
```diff
@@ -158,6 +158,10 @@ The install command accepts the following flags:
 - `--no-windows-containers`: disables Windows containers integration
 - `--allowed-org=<org name>`: requires the user to sign in and be part of the specified Docker Hub organization when running the application
 - `--backend=<backend name>`: selects the default backend to use for Docker Desktop, `hyper-v`, `windows` or `wsl-2` (default)
+- `--admin-settings`: Automatically creates an `admin-settings.json` file which is used by admins to control certain Docker Desktop settings on client machines within their organization. For more information, see [Settings Management](../hardened-desktop/settings-management/index.md).
+  - It must be used together with the `--allowed-org=<org name>` flag. 
+  - For example:
+    `--allowedOrg=<org name> --admin-settings='{"configurationFileVersion": 2, "enhancedContainerIsolation": {"value": true, "locked": false}}'`
```
ebriney, line 164: same here, `--allowed-org=<org name>` instead of `--allowedOrg=<org name>`

desktop/hardened-desktop/settings-management/configure.md
```diff
@@ -0,0 +1,137 @@
...
+
+    - Mac: `/Library/Application\ Support/com.docker.docker/admin-settings.json`
+    - Windows: `C:\ProgramData\DockerDesktop\admin-settings.json`
+    - Linux: `/usr/share/docker-desktop/admin-settings.json`
+
+    By placing this file in the above protected directories, end users are unable to modify it.
+
+    >**Note**
+    >
+    > It is assumed that you have the ability to push the `admin-settings.json` settings file to the locations specified above through a device management software such as [Jamf](https://www.jamf.com/lp/en-gb/apple-mobile-device-management-mdm-jamf-shared/?attr=google_ads-brand-search-shared&gclid=CjwKCAjw1ICZBhAzEiwAFfvFhEXjayUAi8FHHv1JJitFPb47C_q_RCySTmF86twF1qJc_6GST-YDmhoCuJsQAvD_BwE).
```
ebriney, line 36: we should mention that you can use command line flag also here now, no?

ebriney (changes requested):

desktop/hardened-desktop/settings-management/configure.md
aevesdocker: I added it in the lines above. second paragraph under 'How do I set up and enforce Settings Management'

ebriney (changes requested):

desktop/hardened-desktop/enhanced-container-isolation/index.md
```diff
@@ -0,0 +1,115 @@
...
+
+```
+         0          0 4294967295
+```
+
+This means that the root user in the container (0) is in fact the root user in the Docker Desktop VM (0) which reduces container isolation. The user-ID mapping varies with each new container, as each container gets an exclusive range of host User-IDs for isolation. User-ID mapping is automatically managed by Docker Desktop.
+
+With Enhanced Container Isolation, if a process were to escape the container, it would find itself without privileges at the VM level. For further details, see [How Enhanced Container Isolation works](how-eci-works.md).
+
+Since, Enhanced Container Isolation [uses the Sysbox container runtime](how-eci-works.md) embedded in the Docker Desktop Linux VM, another way to determine if a container is running with Enhanced Container Isolation is by using `docker inspect`:
```
MihaelaStoica, line 103: remove comma:  "Since, Enhanced Container Isolation uses" -> "Since Enhanced Container Isolation uses"

desktop/hardened-desktop/enhanced-container-isolation/how-eci-works.md
```diff
@@ -0,0 +1,21 @@
...
+
+Sysbox is an alternative `runc` used to create a container using the Linux kernel’s namespaces, cgroups, and other features. 
+
+Unlike the standard `runc` runtime, Sysbox enhances container isolation by using techniques such as enabling the Linux user-namespace on all containers, emulating portions of the `proc` filesystem and `sysfs` inside the container and vetting sensitive accesses between the container and the Linux kernel. This adds an extra layer of isolation between the container and the Linux kernel. 
+
+Without Enhanced Container Isolation, Docker Desktop has Docker Engine run as root with full capabilities inside a container that shares almost all namespaces with the Linux VM’s root user. Whilst this provides strong isolation between containers and the underlying host machine, it gives the container access to all the VM’s kernel resources and does not prevent Docker Desktop users from launching a container that runs as root in the Docker Desktop Linux VM, or from using insecure privileged containers. This brings Docker Desktop users closer to gaining privileged access to the underlying host.
+
+### How is this different to rootless mode in Docker Engine?
+
```
MihaelaStoica, line 20: I am assuming we will add something here, or remove the header?

MihaelaStoica (changes requested):

desktop/hardened-desktop/enhanced-container-isolation/how-eci-works.md
aevesdocker: Yes. Waiting for Cesar's additions.

ebriney (dismissed):

dvdksn (approved):
nå kjører vi!

PR #15777, opened 2022-09-29
move the uninstall content to a new page

aevesdocker:
Also removes the 'Update' content which is covered elsewhere


dvdksn (approved):
Good move!

PR #15774, opened 2022-09-29
rename docs repository

crazy-max:
Replace `https://github.com/docker/docker.github.io` with `https://github.com/docker/docs`. Also adds a new `repo` var in `_config.yml`.

Signed-off-by: CrazyMax <crazy-max@users.noreply.github.com>

PR #15786, opened 2022-09-29
fix: incorrect default edit URL

neersighted:
### Proposed changes
Fix the default edit URL after the repo rename.

The current URL includes the `/edit` path component twice, resulting in a 404.

### Related issues (optional)
Introduced in: #15774


crazy-max (approved):
Good catch thanks!

PR #15788, opened 2022-09-29
Remove redirects for samples

craig-osterhout:
### Proposed changes
New samples pages were added but some paths had old redirects.
This removes redirects for the new samples pages that have been added.


usha-mandya (approved):
LGTM

PR #15792, opened 2022-09-30
chore(vale): update word list

dvdksn:
verbiage updates

### DIFF
--- a/.github/vale/Vocab/Industry/accept.txt
+++ b/.github/vale/Vocab/Industry/accept.txt
@@ -1,6 +1,7 @@
 AWS
 Amazon
 Apple
+Azure( Blob Storage)?
 Artifactory
 BusyBox
 CentOS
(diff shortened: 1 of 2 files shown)


.github/vale/Vocab/Industry/accept.txt
```diff
@@ -1,6 +1,7 @@
 AWS
 Amazon
 Apple
+Azure( Blob Storage)?
```
usha-mandya, line 4: Is the ` ( Blob storage)?` intentional?

dvdksn: yeah it's an optional (acceptable) suffix to Azure, so it approves both Azure and Azure Blob Storage. We have the same for `GitHub( Actions)?`

usha-mandya: Makes sense. Thanks

usha-mandya (approved):
LGTM

PR #15804, opened 2022-10-04
Fix subu(g)id range to match context

pojiro:


pojiro:
I read and I realized misunderstanding the configure, excuse me

> This means that testuser is assigned a subordinate user ID range of 231072 and the next 65536 integers in sequence. 

ref https://docs.docker.com/engine/security/userns-remap/

So close this PR.

Issue #15806, opened 2022-10-04
change from apt-get  to apt


jeas560:
[desktop/install/ubuntu.md](https://docs.docker.com/desktop/install/ubuntu/)
change:
sudo apt-get install ./docker-desktop-<version>-<arch>.deb
to:
sudo apt install ./docker-desktop-<version>-<arch>.deb

thaJeztah:
Was there a specific issue you ran into when using `apt-get`? I know there's been some prior discussion on that; in general, we converge on using `apt-get` as (unlike `apt`), it has a stable API, and therefore is also suitable for users who want to script these commands. Also see https://github.com/docker/docs/pull/8135, and some issues/prs linking to that

usha-mandya:
Closing this issue as there's no response. Feel free to comment on the issue after I close.

PR #15807, opened 2022-10-04
Command to install package with apt on Ubuntu modified

NitishKumar06:
Fix: #15806

### DIFF
--- a/desktop/install/ubuntu.md
+++ b/desktop/install/ubuntu.md
@@ -54,7 +54,7 @@ Recommended approach to install Docker Desktop on Ubuntu:
 
 ```console
 $ sudo apt-get update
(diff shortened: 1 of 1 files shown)


thaJeztah (changes requested):
looks like this content moved, so the PR no longer applies.

I'm not sure though if we should update; I know there's been some prior discussion on that; in general, we converge on using `apt-get` as (unlike `apt`), it has a stable API, and therefore is also suitable for users who want to script these commands. Also see https://github.com/docker/docs/pull/8135, and some issues/prs linking to that

(although I see some `apt install` examples sneaked in; we should probably update those)

usha-mandya:
Thanks for the PR. Closing this PR as per Sebastiaan's comment above. Feel free to continue the conversation after I close.

PR #15808, opened 2022-10-05
Format `docker build` as a code sample

daniel-schroeder-dev:
### Proposed changes

To stay consistent with other uses in the docs, it would make sense for `docker build` to be formatted as code here.


aevesdocker (approved):
Thank you!

PR #15809, opened 2022-10-05
Update wording around removing a container

daniel-schroeder-dev:
### Proposed changes

The original used "simple" instead of "simply". I edited the sentence a bit more to make it read clearer and removed the use of "simple" altogether.


aevesdocker (approved):
Thanks @daniel-schroeder-dev!

PR #15817, opened 2022-10-06
Update "Mounts" example

daniel-schroeder-dev:
### Proposed changes

In the original example, the "Mounts" field returned by `docker container inspect [CONTAINER]` was not shown. My edit may include too much data, but we should at least show the "Mounts" field so folks know what to look for.


dvdksn:
Thanks a lot for helping identify this fix @daniel-schroeder-dev, I just pushed a small update that shortens the example.

PR #15826, opened 2022-10-07
Engdocs 1019 1021

aevesdocker:
This PR adds no new content. It merely takes the content that was there and moves it to more intuitive places.


hyu:
After reviewing the Netlify preview and the current docs, I am _so_ excited at the re-org in this PR! Here are the changes I observed:

- Split the currently overwhelming Overview page into…
  - …a more welcoming, shorter page with cards to let users jump into the most relevant next steps
  - …and a "Key features and use cases" for the rest of the content
- Consolidate Compose V2-related docs in its own section
- Rename the generic "Getting started" (which may seem a bit redundant with Overview) to "Try Docker Compose," which is more of a crisp CTA and more accurate to the content on the page.

All these seem like excellent improvements. Let me know if I missed anything!

hyu:
This is _clearly_ beyond the scope of this PR, but I wanted to mention as feedback that I attempted to click these icons and was disappointed when they didn't work as part of the larger "button" (Only the blue link itself is clickable)

image

Not at all an actionable request for this PR, but leaving it here in context — in case it can contribute to a larger discussion around this component.

aevesdocker:
Thank you for going through it so closely! So pleased you approve of the changes. 
Fixing the icon link is totally within this PR - in fact, it's something I should've done the first time around, but forgot 🤦 so fixing now!!

dvdksn (approved):
So good

star struck

PR #15829, opened 2022-10-08
Remove wording around saving widgets from `develop.md`

daniel-schroeder-dev:
### Proposed changes

There is no logic in that Flask app to save widgets, so I reworded the two sentences that referenced saving widgets. 

There are two routes:
1. `/widgets` -> Gets all the rows from the `widgets` table
2. `/initdb` -> Drops the DB/Table and rebuilds them from scratch


craig-osterhout (approved):
Thanks. LGTM.

PR #15833, opened 2022-10-08
Included an easy way to clone the app directory

Xavier577:
I included a line that describes how to use degit to clone the app directory in the getting started repo.

### Proposed changes

Use degit to clone the app directory from the getting-started repo. I did because degit has the ability to clone just the app directory without cloning the entire repo which is more intuisive.

### DIFF
--- a/get-started/02_our_app.md
+++ b/get-started/02_our_app.md
@@ -24,7 +24,7 @@ Before we can run the application, we need to get the application source code on
 our machine. For real projects, you will typically clone the repo. But, for this tutorial,
 we have created a ZIP file containing the application.
 
-1. Download the App contents from the [getting-started repository](https://github.com/docker/getting-started/tree/master){:target="_blank" rel="noopener" class="_"}. You can either pull the entire project or [download it as a zip](https://github.com/docker/getting-started/archive/refs/heads/master.zip) and extract the app folder out to get started with.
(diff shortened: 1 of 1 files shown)


Xavier577:
I was editing straight from GitHub but now I had to clone the repo and edit locally because I wanted to change and see my changes and not have to redeploy every time. I was having issues with excessive whitespace but I fixed it up and I think I am good to go, here is a screenshot of my change. Please do leave a feedback.

image

craig-osterhout:
Hi @Xavier577, thank you for looking in to this issue.
Installing additional apps that are only used in the tutorial can be a touchy point for some users.

We can also use git to pull only the app folder.

```
    $ git init
    $ git remote add -f origin https://github.com/docker/getting-started.git
    $ git config core.sparseCheckout true
    $ git sparse-checkout set app/
    $ git pull origin master
```

Alternatively, we can include step-by-step instructions on extracting the folder from the zip file for Mac/Linux and Windows.
I think the git steps are probably the better solution because it doesn't require installing additional apps, and it's quick and easy to copy-paste the same instructions for all platforms.

Xavier577:
Thank you for your feedback @craig-osterhout. I used degit becuase I figured most of the users of docker would most likely have nodejs and npm already installed along with the fact that the tutorial was a nodejs one, plus doing it with degit was just a one line command.Do you suggest add the git way of doing it instead?

thaJeztah:
I wonder if we should jump through these hoops; the repository shouldn't be too big as a whole, so

```bash
git clone https://github.com/docker/getting-started
cd app
```

doesn't feel _too bad_ to do 🤔

Xavier577:
yh you are right, it's just that there a number of people who wouldn't want to clone an entire repo just to use one directory. But I just thought it would be more intuisive. 😅

craig-osterhout:
Thanks again for contributing. Closing this as the consensus seems to be keep it the same.

PR #15836, opened 2022-10-09
Fix typo

akiraaso:
### Proposed changes

I think the `)` is a typo, so I've removed it.


thaJeztah (approved):
yup! that looks like a typo, thanks!

LGTM

Issue #4101, opened 2017-08-05
Docker secret custom directory


anuragagarwal561994:
### Problem description

No where in the documentation it is mentioned on how to create directories inside /run/secrets if we want our secret keys to be organised. But if we name our target to contain slashes (representing directories), actual directories are formed within /run/secrets and it works.

### Problem location

- I saw a problem on the following URL: https://docs.docker.com/compose/compose-file/#long-syntax-2

- I couldn't find the information I wanted. I expected to find it near the following URL: I could not find the information I wanted. I explored it my self, and felt that it can be important for others to know too.

- Other: 

### Project version(s) affected

Versions supporting docker secrets.

### Suggestions for a fix

The target option says that it represents the "The name of the file that will be mounted", but more elaborately it should be path that will be mounted.

mstanleyjones:
Docker 17.06 CE and higher support custom locations for secrets. See https://docs.docker.com/engine/swarm/secrets/.

Enquier:
This needs to be updated. It is still incorrect in the Compose File Reference Documentation

andrewchen5678:
why is this closed, docker-compose had the wrong documentation, it says "target: The name of the file to be mounted in /run/secrets/ in the service’s task containers. Defaults to source if not specified," but I just tested out a simple set up with target outside of the /run/secrets/ folder like `target: /etc/testfile`. It is not the first time I see documentation issues in docker compose and the tickets for it are closed w/o a fix.

PR #15854, opened 2022-10-11
Include support for absolute secret path

epuronta:
### Proposed changes

Added a note to compose docs that absolute path can be set with secret long syntax. This has been supported by the engine for a while.

### Related issues (optional)

#4101

### DIFF
--- a/compose/compose-file/index.md
+++ b/compose/compose-file/index.md
@@ -1654,7 +1654,7 @@ the service's containers.
 
 - `source`: The name of the secret as it exists on the platform.
 - `target`: The name of the file to be mounted in `/run/secrets/` in the
-  service's task containers. Defaults to `source` if not specified.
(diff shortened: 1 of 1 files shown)


usha-mandya:
@epuronta Thanks for the PR.  Could you create a PR to update the [upstream Compose spec docs](https://github.com/compose-spec/compose-spec/blob/master/spec.md) before we can accept the changes here?

epuronta:
Thanks for the comment @usha-mandya! I commented on [compose-spec#161](https://github.com/compose-spec/compose-spec/issues/161) and created a similar PR there. They seem to be blocked from updating the spec as the detail is not described in this doc though. Who could break the deadlock?

usha-mandya:
@glours could you PTAL?

glours:
@epuronta @usha-mandya I try to unblock the situation on the specification level, IHMO we can close this issue

PR #15857, opened 2022-10-12
ENGDOCS-1020

aevesdocker:
No new content added, just moving it around

### DIFF
--- a/_data/toc.yaml
+++ b/_data/toc.yaml
@@ -1521,13 +1521,13 @@ manuals:
   - sectiontitle: Install Docker Compose
     section:
       - path: /compose/install/
-        title: Install overview
+        title: Overview
       - path: /compose/install/linux/
-        title: Install on Linux
+        title: Install the Compose plugin
       - path: /compose/install/other/
-        title: Other install scenarios
+        title: Install the Compose standalone
       - path: /compose/install/uninstall/
-        title: Uninstall Docker Compose
+        title: Uninstall Compose
   - path: /compose/gettingstarted/
     title: Getting started
   - sectiontitle: Environment variables
--- a/compose/install/index.md
+++ b/compose/install/index.md
@@ -37,3 +37,8 @@ If you already have Docker Engine and Docker CLI installed, you can install the 
 ### Scenario three: Install the Compose standalone 
 
 You can [install the Compose standalone](other.md) on Linux or on Windows Server.
+
+>Note
+>
+>This install scenario is no longer supported.
+{: .important}
--- a/compose/install/index.md
+++ b/compose/install/index.md
@@ -1,54 +1,39 @@
 ---
 description: How to install Docker Compose
 keywords: compose, orchestration, install, installation, docker, documentation
-title: Install Docker Compose
+title: Overview
 toc_max: 3
 redirect_from:
 - /compose/compose-desktop/
 ---
 
-On this page you can find information on how to get and install Compose.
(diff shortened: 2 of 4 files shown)


hyu:
Looks good! Here are the changes I noticed:

- Simplified and standardized titles in header (nice!)
- Removed "About Docker Desktop" screenshot from overview, since its size was giving it undue importance.
- In Overview, frame install options as 3 scenarios. (I like that this explicit hierarchy directs most users to #1.)
- Removed "where to go next" from plugin install page, making it more consistent with the standalone install page.

All of these seem like solid improvements to me. Let me know if I missed anything!

&nbsp;

### Reference
(I found myself wondering what had changed in the recent iterations. Hope this is useful to you too!)


| Merged 7/2  | Merged 8/26 | Current PR |
| --- | --- | --- |
| [ENGDOCS 711](https://github.com/docker/docs/pull/14838) | [PR 15494](https://github.com/docker/docs/pull/15494) |
| image | image | image |

hyu:
This is a nit that's possibly out of scope!

...

When a tip box immediately follows a heading, there doesn't seem to be enough vertical space. For example, in this [plugin install](https://deploy-preview-15857--docsdocker.netlify.app/compose/install/linux/#install-the-plugin-manually) section:

image

&nbsp;
The spacing is better elsewhere in docs, where there's a line between the header and the tip. Like on [Install Docker Desktop](https://deploy-preview-15857--docsdocker.netlify.app/desktop/install/mac-install/#system-requirements):
image

hyu:
Not blocking, but I do wonder whether it's possible to offer a bit more context on why a user would choose to install Compose standalone? 

For Scenario two, we do imply that you might want the plugin if you already have Docker Engine and CLI, and don't want to install DD for whatever reason. Is there some extra bit of info we can similarly offer for Scenario three?

(I apologize that I myself don't actually know how to contextualize this! My vague understanding is that we don't much support this install route, but offer it nonetheless for legacy reasons? CCing @StefanScherer in case he might have more context.)

image

StefanScherer:
Yes, installing compose v2 standalone is only for legacy reasons, maybe users have lots of scripts that use `docker-compose` instead of `docker compose` and don't want to migrate or just don't know that there now is the plugin. The Windows Server use case is also probably a very rare one, that's probably only for test purposes.
Another use case might be to have the cli and compose installed on Mac or Windows to use them with a manually local VM or a remote Docker Engine.

Looking at package managers that help installing docker-compose show at least some interest on Mac [via homebrew](https://formulae.brew.sh/formula/docker-compose), but on  Windows I was never asked to update the docker-compose [chocolatey package](https://community.chocolatey.org/packages/docker-compose#versionhistory) beyond v1.

aevesdocker:
Thank you for the feedback @hyu  and extra context @StefanScherer. 

Re: adding extra context for scenario three. 
Seeing as we no longer support this option, and there are simply edge cases as to why someone would choose this option, I wonder if it's better simply to add just a warning note saying we no longer support it and make this scenario as unappealing as possible? Users who still decide this is the option for them surely know what they're doing and are advanced users

hyu:
This seems like the right approach!

> Seeing as we no longer support this option, and there are simply edge cases as to why someone would choose this option, I wonder if it's better simply to add just a warning note saying we no longer support it and make this scenario as unappealing as possible? Users who still decide this is the option for them surely know what they're doing and are advanced users.

PR #15862, opened 2022-10-12
Update Microsoft image link

ltlombardi:
### Proposed changes

Previous link does not exist anymore. Updated with the alternative one.
Mays need to update also the text


dvdksn (approved):
lgtm

PR #15865, opened 2022-10-13
Clarify dockerfile instructions

usha-mandya:

Carry suggestions from https://github.com/docker/docs/pull/9315/files

closes #9315

### DIFF
--- a/build/building/multi-stage.md
+++ b/build/building/multi-stage.md
@@ -19,7 +19,7 @@ Dockerfiles while keeping them easy to read and maintain.
 ## Before multi-stage builds
 
 One of the most challenging things about building images is keeping the image
-size down. Each instruction in the Dockerfile adds a layer to the image, and you
+size down. Each `RUN`, `COPY`, and `ADD` instruction in the Dockerfile adds a layer to the image, and you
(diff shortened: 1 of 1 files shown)


crazy-max (approved):
LGTM

@thaJeztah I recall there was a thread about a slight difference between BuildKit and legacy builder for some instructions creating layers but can't put my finger on it. Wonder if this was not about squash :thinking:

thaJeztah:
@crazy-max you probably mean https://github.com/moby/moby/issues/38903

thaJeztah:
Oh, and the other difference between BuildKit and classic; `VOLUME` in classic applies volumes during build, but BuildKit doesn't (only changes the image config, but doesn't attach volumes during build)

crazy-max:
> Oh, and the other difference between BuildKit and classic; `VOLUME` in classic applies volumes during build, but BuildKit doesn't (only changes the image config, but doesn't attach volumes during build)

Ah right that was it!

thaJeztah:
I think there's also `WORKDIR`, which for BuildKit follows `USER`, but for the classic builder doesn't

PR #15866, opened 2022-10-13
registry.json: show how to query file ACLs on Windows

djs55:
### Proposed changes

On the `registry.json` page we described how to check the file permissions on Mac and Linux but not Windows. This patch shows how to check the ACLs on Windows.


usha-mandya (approved):
LGTM

PR #15878, opened 2022-10-15
faqs: fix formatting issue with a bullet point

lizzthabet:
### Proposed changes

This PR fixes a very small formatting issue with a bullet point I noticed while reading the FAQs page.


aevesdocker (approved):
Thank you for spotting and fixing this @lizzthabet!

PR #15891, opened 2022-10-17
[Issue 13005] Add folder name for where to create Dockerfile

### LINKED ISSUE #13005: Clarify where Dockerfile should be made
[get-started/02_our_app.md](https://docs.docker.com/get-started/02_our_app/)

I was not sure where a new Dockerfile should be made in this tutorial.
Perhaps rather than saying "in the same folder as the file package.json", "in the getting-started-master/app subdirectory where package.json can be found" might be clearer. 

Also, I'd suggest to remove "If you haven’t already done so" in the step 2, because this isn't told anywhere in the tutorial. The tutorial should not assume the reader, who is an absolute Docker beginner, would do this.

I was confused about the location of newly written Dockerfile because there is already a Dockerfile in the unzipped top level directory. So I thought I need to create a new fresh directory and copy just package.json and write a new Dockerfile. If there were a sentence like "You will see a Dockerfile at the top level of the unzipped directory, but disregard it for now. You are going to write a new Dockerfile in the app directory.", I wouldn't have attempted to create a new directory.

craig-osterhout:
### Proposed changes

User unsure where to create `Dockerfile` given the reference of only `package.json`. Added additional reference to the `app` folder.
Removed `If you haven't done so` wording as it's not necessary and can cause confusion if users think they missed a step.

### Related issues (optional)

Fixes #13005

### DIFF
--- a/get-started/02_our_app.md
+++ b/get-started/02_our_app.md
@@ -40,7 +40,7 @@ Dockerfile is simply a text-based script of instructions that is used to
 create a container image. If you've created Dockerfiles before, you might
 see a few flaws in the Dockerfile below. But, don't worry. We'll go over them.
 
-1. In the `app` folder, the same location as the `packge.json` file, create a file named `Dockerfile` with the following contents.
+1. In the `app` folder, the same location as the `package.json` file, create a file named `Dockerfile` with the following contents.
 
    ```dockerfile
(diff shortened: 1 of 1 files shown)


usha-mandya (approved):
LGTM

PR #15892, opened 2022-10-17
[Issue 14057] Update output for example

### LINKED ISSUE #14057: Mistake in the example
### Problem location

- I saw a problem at the following URL: (https://docs.docker.com/language/python/run-containers/#stop-start-and-name-containers)

- Other: There is an example: docker stop wonderful_kalam
and after that: docker ps --all
where we can see that docker container continues working, but it must be stopped - we stopped it one command ago.

craig-osterhout:
### Proposed changes
The example says to stop the container, but the example output still shows the container running.
Update example output.

### Related issues (optional)
Fixes #14057


dvdksn:
I think this was just fixed in a community PR: 3a2e9e38b632a9dfb08da80b1b07f6d01158b75b

PR #13113, opened 2021-07-03
Typo fixes

MicaelJarniac:
### Proposed changes

Mostly typo fixes.

PR #15895, opened 2022-10-18
Typo fixes

MicaelJarniac:
### Proposed changes

Mostly typo fixes.

This was supposed to be #13113, but I believe I failed horribly when rebasing, and it ended up in a mess. I hope this time it works.

### DIFF
--- a/language/python/develop.md
+++ b/language/python/develop.md
@@ -141,7 +141,7 @@ if __name__ == "__main__":
     app.run(host ='0.0.0.0')
 ```
 
-We’ve added the MySQL module and updated the code to connect to the database server, created a database and table. We also created a route to fetch widgets. We now need to rebuild our image so it contains our changes.
+We’ve added the MySQL module and updated the code to connect to the database server, create a database and table. We also created a route to fetch widgets. We now need to rebuild our image so it contains our changes.
 
 First, let’s add the `mysql-connector-python` module to our application using pip.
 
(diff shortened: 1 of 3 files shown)


MicaelJarniac:
Okay, this time it seems to have worked. I have no idea what went wrong with the original PR. This PR has been created from the exact same branch, the only thing I did was to close the original one and open this new one, but for whatever reason, this one worked fine.

language/python/develop.md
```diff
@@ -141,7 +141,7 @@ if __name__ == "__main__":
     app.run(host ='0.0.0.0')
 ```
 
-We’ve added the MySQL module and updated the code to connect to the database server, create a database and table. We also created a route to fetch widgets. We now need to rebuild our image so it contains our changes.
+We’ve added the MySQL module and updated the code to connect to the database server, created a database and table. We also created a route to fetch widgets. We now need to rebuild our image so it contains our changes.
```
MicaelJarniac, line 144: I believe it should be "create", as in "we've updated the code to connect and create".

"Created" makes it sound like we did two things, we've updated the code and we've created a database and table.

But we did only one thing, we've updated the code. We've updated the code so that it can connect to the database and create a database and table.

MicaelJarniac: @usha-mandya

dvdksn (approved):
LGTM, thanks for the helping hand @MicaelJarniac !

usha-mandya:
Thank you @MicaelJarniac!

PR #15897, opened 2022-10-18
CPUB-406 Include repo-level summary in DVP API doc

chrispatrick:
### Proposed changes

Add repo-level summary info to DVP Data docs

### DIFF
--- a/docker-hub/api/dvp.yaml
+++ b/docker-hub/api/dvp.yaml
@@ -3,7 +3,7 @@ info:
   title: DVP Data API
   version: 1.0.0
   x-logo:
-    url: https://docs.docker.com/images/logo-docker-main.png
+    url: https://docs.docker.com/assets/images/logo-docker-main.png
     href: /reference
   description: |
      The Docker DVP Data API allows [Docker Verified Publishers](https://docs.docker.com/docker-hub/publish/) to view image pull analytics data for their namespaces. Analytics data can be retrieved as raw data, or in a summary format.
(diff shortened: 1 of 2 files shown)


docker-hub/api/dvp.yaml
```diff
@@ -12,7 +12,14 @@ info:
...
 
-     - For every namespace, repository, and reference (tag or digest):
+     There are two levels of summary data:
+
+     - **Tag/digest-level** - for every namespace, repository, and reference (tag or digest):
+       - Unique IP address count
+       - Pulls by tag count
+       - Pulls by digest count
+       - Version check count
+     - **Repository-level** - for every namespace and repository:
```
dvdksn, line 22: The data points for repo-level and tag-level are both the same, right? So I don't think we need to describe them twice. I'm aligning the descriptions here with what I put in the other doc.

```suggestion
     There are two levels of summary data:

    - Repository-level, a summary of every namespace and repository
    - Tag- or digest-level, a summary of every namespace, repository, and reference
      (tag or digest)
```

docker-hub/api/dvp.yaml
```diff
@@ -12,7 +12,14 @@ info:
...
        - Unique IP address count
        - Pulls by tag count
        - Pulls by digest count
```
dvdksn, line 25: Deindent this and make it a separate list, i.e:

```
The summary data formats contain the following data points:

- Unique IP address count
- Pulls by tag count
- Pulls by digest count
- Version check count
```

dvdksn (approved):
:shipit:

PR #15907, opened 2022-10-18
Last modified poc

craig-osterhout:
Do not merge
Simple proof of concept to visualize showing last updated or validated.


myfile.txt
```diff
@@ -0,0 +1,1904 @@
+.docker/Dockerfile.devenv
```
thaJeztah, line 1: Was this file meant to be included?

craig-osterhout: No, myfile.txt is just a remanent of the method I used to grab the commit dates for each file locally. This was implemented hastily to experience how it will look and feel. There's no intent to use/merge any of this. The implementation will need to be better designed if we decide to do it.

craig-osterhout:
Closing as just need netlify for the poc.

PR #15911, opened 2022-10-19
fix icon links and update old screenshot

aevesdocker:


desktop/index.md
```diff
@@ -78,15 +78,15 @@ using a secure repository.
         <div class="component">
              <div class="component-icon">
                  <img src="/assets/images/download.svg" alt="Download and install" width="70" height="70">
```
dvdksn, line 80: Looks like the install icon is missing the link

```suggestion
                 <a href="/desktop/install/mac-install"><img src="/assets/images/download.svg" alt="Download and install" width="70" height="70"></a>
```

aevesdocker: Yeh I missed this one out on purpose the descriptions below provide the link to each install page for Mac, Windows, Linux and I wasn't sure which one to attach to the icon

dvdksn: maybe we should remove the link for the h2 element too in that case? it changes the text color but I don't think it looks bad.

image
image

aevesdocker: I'm tempted to leave as is so users can access whichever is appropriate to them

aevesdocker: Fixing now!

dvdksn (approved):
:shipit:

PR #15919, opened 2022-10-19
Remove shell prompts ($) from commands

sharjeelaziz:
Remove shell prompts from commands so that they can be copied/pasted without edits.

### Proposed changes

It is easier to copy commands from the document if the command is not prefixed by the shell prompt $.


dvdksn:
hi @sharjeelaziz, thanks for taking the time to raise this PR. We'll look into fixing this problem for the docs via a global solution (i.e. escaping user prompt sign when copying, or removing all occurrences of the prompt everywhere). We want to fix this everywhere, so I am closing this PR in favor of a global fix.

PR #15926, opened 2022-10-20
Make console command copy/paste compliant

pgrange:
The `$` symbole to represent the prompt at the beginning of the command in this documentation make it impossible for someone to just copy/paste them and install Docker.

I removed them to make installation even more a no-brainer.


dvdksn:
hi @pgrange, thanks for taking the time to raise this PR. We'll look into fixing this problem for the docs via a global solution (i.e. escaping user prompt sign when copying, or removing all occurrences of the prompt everywhere). We want to fix this everywhere, so I am closing this PR in favor of a global fix.

PR #15937, opened 2022-10-21
Updated the commands

rehman4023:
### Proposed changes
the dollar symbols at the start of each command are causing an issue while installing docker, we need to remove these $.

I have just removed $ where ever its not necessary, now the script can just be copied and pasted to install docker

### DIFF
--- a/engine/install/debian.md
+++ b/engine/install/debian.md
@@ -82,9 +82,9 @@ from the repository.
     repository over HTTPS:
 
     ```console
-    $ sudo apt-get update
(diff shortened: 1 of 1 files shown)


dvdksn:
hi @rehman4023, thanks for taking the time to raise this PR. We'll look into fixing this problem for the docs via a global solution (i.e. escaping user prompt sign when copying, or removing all occurrences of the prompt everywhere). We want to fix this everywhere, so I am closing this PR in favor of a global fix.

PR #15947, opened 2022-10-21
Update code copy for Compose files

craig-osterhout:
### Proposed changes

When copying Compose file code examples, a space appears at the start of additional lines.
Updated code copy to remove a single space at the start of all lines.

### DIFF
--- a/assets/js/copy.js
+++ b/assets/js/copy.js
@@ -15,8 +15,8 @@ function copyCodeBlock(event) {
   const copyButton = event.currentTarget
   const codeBlock = copyButton.parentElement.querySelector("pre.highlight code")
   const code = codeBlock.innerText.trim()
-  // remove "$ " prompt at start of lines in code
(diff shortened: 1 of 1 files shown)


dvdksn:
I'm either misunderstanding the problem you describe, or I am not seeing the same thing. E.g. when copying examples from [this page](https://docs.docker.com/compose/profiles/), I don't see any additional whitespace in my clipboard. Where have you observed this issue?

craig-osterhout:
Thanks, @dvdksn. I only looked at the getting started guides. For example, https://docs.docker.com/get-started/08_using_compose/
Your example from the compose profiles page doesn't cause any extra spaces. 

The only quick difference that I notice is there's an extra line in the getting started guide.  I'll try to dig deeper.

craig-osterhout:
I think it's related to nesting the codeblock inside a step/list in markdown. I'm gonna close this PR since this will cause more problems for the non-nested stuff.

PR #15959, opened 2022-10-24
Contribute patch

aevesdocker:
This PR adds guidance on working with UI elements in content (with the help of Google developer docs) and add guidance on parentheses.


dvdksn (approved):
nice additions!

PR #15979, opened 2022-10-25
jekyll: remove polldaddy

crazy-max:
As discussed remove polldaddy as it's not being used anymore.


404.md
```diff
@@ -1,7 +1,6 @@
 ---
 title: "Sorry, we can't find that page"
 permalink: /404.html
-noratings: true
```
crazy-max, line 4: we might need to switch `noratings` to `skip_feedback` for hotjar though.

dvdksn (approved):
lgtm

PR #15783, opened 2022-09-29
added chmod command

brccabral:
for those that are new to linux and those that like to just copy commands

PR #15990, opened 2022-10-26
added chmod command

brccabral:
Hi,
After pull request #15783 failed I did a new fork and reapplied the change.
"for those that are new to linux and those that like to just copy commands"

### DIFF
--- a/compose/install/other.md
+++ b/compose/install/other.md
@@ -18,6 +18,9 @@ On this page you can find instructions on how to install the Compose standalone 
   $ curl -SL https://github.com/docker/compose/releases/download/{{site.compose_version}}/docker-compose-linux-x86_64 -o /usr/local/bin/docker-compose
   ```
 2. Apply executable permissions to the standalone binary in the target path for the installation.
+  ```console
+  $ sudo chmod +x /usr/local/bin/docker-compose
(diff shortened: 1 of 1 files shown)


compose/install/other.md
```diff
@@ -18,6 +18,9 @@ On this page you can find instructions on how to install the Compose standalone
   $ curl -SL https://github.com/docker/compose/releases/download/{{site.compose_version}}/docker-compose-linux-x86_64 -o /usr/local/bin/docker-compose
   ```
 2. Apply executable permissions to the standalone binary in the target path for the installation.
+  ```console
+  $ sudo chmod +x /usr/local/bin/docker-compose
```
thaJeztah, line 22: Was the `sudo` needed to make it work? (Not at my computer right now to try the steps, but curious if it's needed, or if it works without)

brccabral: Yes, `sudo` is needed because the previous `curl` is downloading the `docker-compose` in the root directory, `/usr/local/bin`. In fact, the `curl` command would also need `sudo` to begin with.

aevesdocker:
Closing PR as per Engineering's advice - only a very small portion of users will need this

PR #15993, opened 2022-10-26
jekyll: fix fetch depth for last_modification_date on deploy

crazy-max:
Changes introduced by https://github.com/docker/docs/pull/15267 for `last_modification_date` don't pick up the right commit date for pages in the `deploy` workflow: https://docs-stage.docker.com/sitemap.xml. That's because `actions/checkout` doesn't fetch the history by default but [Netlify does for previews](https://deploy-preview-15267--docsdocker.netlify.app/sitemap.xml). That's why we didn't catch this issue. Same applies with `fetch_depth` plugin, we need the history when deployed but we can keep `depth: 1` with local builds for performance.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -68,6 +68,8 @@ jobs:
       -
         name: Checkout
         uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
       -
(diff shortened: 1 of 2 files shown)


dvdksn (approved):
Thanks for fixing this so promptly!

PR #15994, opened 2022-10-27
display reading time and tags in toc nav

crazy-max:
~needs #15993~
~needs #15998~

Move tags and reading time from content body to the toc nav as it seems more appropriate to have it there. I also took the opportunity to move the edit page and request changes links to the metadata block instead of having it under "Contents" as it's not visible at first sight on some pages like https://docs.docker.com/engine/reference/builder/.

image

image

### DIFF
--- a/_includes/read_time.html
+++ b/_includes/read_time.html
@@ -1,6 +1,0 @@
-{%- unless page.skip_read_time == true -%}
-{%- assign words = content | number_of_words -%}
-{%- if words >= 360 -%}
-<em class="reading-time">Estimated reading time: {{ words | divided_by:180 }} minutes</em>
-{%- endif -%}
-{%- endunless -%}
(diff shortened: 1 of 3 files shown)


crazy-max:
Looking at pages like https://deploy-preview-15994--docsdocker.netlify.app/engine/reference/commandline/buildx_build/ and https://deploy-preview-15994--docsdocker.netlify.app/engine/reference/builder/, last modification date does not look correct. It seems it takes the stub file commit date instead of the remote resource. I will take a look.

crazy-max:
As discussed, remove last mod time from toc nav and add tags instead:

image

@usha-mandya @dvdksn Let me know if that SGTY

dvdksn:
IMO we could either replace the "Metadata" title with something like "Page details", or remove that first title from the box altogether.  Thoughts? @docker/docs-team

aevesdocker:
> IMO we could either replace the "Metadata" title with something like "Page details", or remove that first title from the box altogether. Thoughts? @docker/docs-team

I think for consistency there should be a title, in which case 'Page details' works

dvdksn (approved):
LGTM 💪🏻 

image

usha-mandya (approved):
LGTM. Thank you @crazy-max. Let's :shipit:

PR #16021, opened 2022-10-29
Replace docker-compose with docker compose

hamzanatek:

### DIFF
--- a/compose/features-uses.md
+++ b/compose/features-uses.md
@@ -65,7 +65,7 @@ runs, if it finds any containers from previous runs, it copies the volumes from
 the old container to the new container. This process ensures that any data
 you've created in volumes isn't lost.
 
-If you use `docker-compose` on a Windows machine, see
+If you use `docker compose` on a Windows machine, see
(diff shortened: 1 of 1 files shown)


NitishKumar06:
Please avoid raising PRs before raising an issue! If you've raised this PR with the intentions to solve an issue, Do mention it in the PR description as well.

Looking at the file changes, `docker-compose` is still the right option. So we don't need to change it.

usha-mandya:
Hey @NitishKumar06, thanks for your suggestion about adding a description to the PR. Having a description in the PR helps Docs maintainers to understand the updates required. 

It's absolutely fine for users to create PRs directly and suggest changes. There's no hard requirement to create issues before suggesting changes.  Creating a PR is more helpful as users not only specify that something's wrong, but they also suggest changes required and contribute directly to improve docs :)

Suggestions from @hamzanatek look correct as the Compose V2 command is `docker compose up`.  See. https://docs.docker.com/compose/compose-v2/ and https://docs.docker.com/engine/reference/commandline/compose/.

aevesdocker (approved):
Thank you @hamzanatek

PR #16024, opened 2022-10-29
Update prometheus.md

Zekia:
Replacing some strange IP host.docker.internal for windows

### Proposed changes

There was some internal IP wich wasn't working for me. Using `host.docker.internal`  should be the right way to contact host.


dvdksn (approved):
tyvm

PR #16025, opened 2022-10-30
Enable dark mode in browser UI

lex111:
### Proposed changes

We can change a browser's color scheme in according with used dark mode on the website. For example, then scrollbars will use dark colors which is good for the eyes, just fell the difference:

| Before   | After    |
| -------- | -------- |
| image | image|

PS: `color-scheme` will only work on the root `html` element, so I had to make appropriate replaces, but it should not be breaking changes.


dvdksn (approved):
Thank you @lex111 for this nice little improvement :) lgtm!

PR #16026, opened 2022-10-30
Update md for docker not in local bin dir

cristiangirlea:
### Proposed changes

If docker is installed with the plugin directly it may not be in the local bin directory.


aevesdocker:
have checked with engineering and they advise not to approve the PR. This occurs very rarely and if the previous steps have been followed correctly, this will not be necessary.

PR #16030, opened 2022-10-31
Extreme nit: Apple silicon

lucasgelfond:
This bugged me a lot while reading

Screen Shot 2022-10-30 at 7 22 54 PM

because it's not parallel ("Apple chip" and "Apple silicon" used right next to each other!)

Initially I changed both to "Apple chip" but it turns out that it's referenced it as "Apple silicon" elsewhere in the docs (this is also what Apple calls it), I initially capitalized each change (i.e. Apple silicon --> Apple silicon) but the latter is used on their site and these docs only have one instance of it (which I changed the case of)


aevesdocker (approved):
Thank you for catching this @lucasgelfond !

PR #16032, opened 2022-10-31
build: garbage collection page

crazy-max:
~needs #16036~
related to https://github.com/docker/cli/pull/2864 and https://github.com/moby/buildkit/issues/1169

Was wondering if we should split this page and add some bits to [dockerd reference](https://docs.docker.com/engine/reference/commandline/dockerd/) page and BuildKit repo but I think it's fine to have it there as we already have all kind of information about caching. We could also cross link in a follow-up.

Also adds a new `Cache` section. Let me know if that sgty.

### DIFF
--- a/build/building/cache/garbage-collection.md
+++ b/build/building/cache/garbage-collection.md
@@ -1,0 +1,90 @@
+---
+title: Garbage collection
+keywords: build, buildx, buildkit, garbage collection, prune
+---
+
+While [`docker build prune`](../../../engine/reference/commandline/builder_prune.md)
+or [`docker buildx prune`](../../../engine/reference/commandline/buildx_prune.md)
+commands run at once, garbage collection runs periodically and follows an
+ordered list of prune policies.
+
+Garbage collection runs in the BuildKit daemon. The daemon clears the build
+cache when the cache size becomes too big, or when the cache age expires. The
+following sections describe how you can configure both the size and age
+parameters by defining garbage collection policies.
+
+## Configuration
+
+Depending on the [driver](../drivers/index.md) used by your builder instance,
+the garbage collection will use a different configuration file.
+
+If you're using the [`docker` driver](../drivers/docker.md), garbage collection
+can be configured in the [Docker Daemon configuration](../../../engine/reference/commandline/dockerd.md#daemon-configuration-file).
+file:
+
+```json
+{
+  "builder": {
+    "gc": {
+      "enabled": true,
+      "defaultKeepStorage": "10GB",
+      "policy": [
+          {"keepStorage": "10GB", "filter": ["unused-for=2200h"]},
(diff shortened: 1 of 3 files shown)


build/building/cache/garbage-collection.md
```diff
@@ -0,0 +1,86 @@
+---
+title: Garbage collection
+keywords: build, buildx, buildkit, garbage collection, prune
+---
+
+While [`docker build prune`](../../../engine/reference/commandline/builder_prune.md)
+or [`docker buildx prune`](../../../engine/reference/commandline/buildx_prune.md)
+commands run at once, garbage collection runs periodically and follows an
+ordered list of prune policies.
+
```
dvdksn, line 10: I think it would be nice to give a little bit more context about where the GC runs and how it works before delving into config instructions.

```suggestion

Garbage collection runs in the BuildKit daemon. The daemon clears the build cache when the cache size becomes too big, or when the cache age expires. The following sections describe how you can configure both the size and age parameters by defining garbage collection policies.

```

build/building/cache/garbage-collection.md
```diff
@@ -0,0 +1,86 @@
...
+        All:            false
+        Keep Duration:  1440h0m0s
+        Keep Bytes:     26GB
+GC Policy rule#2:
+        All:            false
+        Keep Bytes:     26GB
+GC Policy rule#3:
+        All:            true
+        Keep Bytes:     26GB
+```
```
jedevc, line 74: Where does this come from, it looks like command output? (partly for curiosity, but also we should include the command used so that users can check their own policies themselves).

jedevc (commented):
Woo a page on buildkit gc :tada:

I wonder about page naming - maybe we could call it "Build cache garbage collection"? We have different types of gc in docker, with networks/volumes/containers/images and now build, so it might get confusing when searching for information?

I think it's good to have build cache gc information in the build cache section, but we should x-link it from wherever we talk about the other forms of cache (if that exists?).

build/building/cache/garbage-collection.md
crazy-max: Default policies:
* buildkitd: https://github.com/moby/buildkit/blob/master/cmd/buildkitd/config/gcpolicy.go
* dockerd: https://github.com/moby/moby/blob/master/builder/builder-next/worker/gc.go

jedevc: Oh awesome, ok, let's not put that command in then, since it's a bit fiddly :smile:

jedevc: > Maybe we could add this in buildx when inspecting a builder? :thinking:

Yes! That would be really useful info to pull through :tada:

dvdksn (approved):
🚢

PR #16036, opened 2022-10-31
build: builkit configure

crazy-max:
Adds `Configure` and `TOML configuration` pages in the BuildKit section.

CNI and registry examples come from buildx but I don't fetch them as I think they should just be in docs website and removed from buildx repo in a follow-up. (@jedevc let me know if that sgty)

Also adds `Color output controls` page to "customize" section.

### DIFF
--- a/build/buildkit/configure.md
+++ b/build/buildkit/configure.md
@@ -4,51 +4,56 @@ description: Learn how to configure BuildKit for your builder.
 keywords: build, buildkit, configuration, buildx, network, cni, registry
 ---
 
-If you [create a `docker-container` or `kubernetes` builder](../building/drivers/index.md)
-with Buildx, you can set a custom [BuildKit configuration](toml-configuration.md)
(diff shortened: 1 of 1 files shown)


build/customize/color-output-controls.md
```diff
@@ -0,0 +1,24 @@
+---
+title: Color output controls
+description: Learn how to control the color of BuildKit progress output.
+keywords: build, buildx buildkit, color, terminal
+---
+
+BuildKit and Buildx has support for modifying the colors that are used to output
```
jedevc, line 7:
```suggestion
BuildKit and Buildx have support for modifying the colors that are used to output
```

jedevc (approved):
> CNI and registry examples come from buildx but I don't fetch them as I think they should just be in docs website and removed from buildx repo in a follow-up

I'm ambivalent on this, I don't really mind - it's definitely more guide-like content instead of specific technical details which is defined in the toml definition.

dvdksn (approved):

PR #16062, opened 2022-11-03
Update examples.md for golang

msadiq058:
calling, `defer cli.Close()`
to close the underline transport used by the client

### Proposed changes

A better practice is to close all the resources used.
So proposed change is to call Close() method for cli to close the underline transport.

### DIFF
--- a/engine/api/sdk/examples.md
+++ b/engine/api/sdk/examples.md
@@ -57,6 +57,7 @@ func main() {
 	if err != nil {
 		panic(err)
 	}
+  defer cli.Close()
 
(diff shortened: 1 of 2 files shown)


dvdksn:
aside from an issue with indentation, lgtm. PTAL @vvoland :heart:

vvoland (commented):
LGTM after indentation is fixed :)

msadiq058:
@vvoland @dvdksn 
fixed the indentation and added one more fix in index.md also :)

dvdksn (approved):
Thank you so much for helping make the docs better! 🐳

PR #16076, opened 2022-11-04
docs: enhance html output

jawys:
### Proposed changes

- Removed some leading space in a code formatting because I found the following issue in html rendering:

screenshot of issues in networking docs

- Replaced double $ before terminal commands in code blocks for consistency

### DIFF
--- a/network/ipvlan.md
+++ b/network/ipvlan.md
@@ -330,7 +330,7 @@ $$ ip a show eth0
   gateway points to the containers `eth0` device. See below for CLI output
   of `ip route` or `ip -6 route` from inside an L3 container for details.
 
-The mode ` -o ipvlan_mode=l3` must be explicitly specified since the default
+The mode `-o ipvlan_mode=l3` must be explicitly specified since the default
 IPvlan mode is `l2`.
 
 The following example does not specify a parent interface. The network drivers
--- a/network/ipvlan.md
+++ b/network/ipvlan.md
@@ -7,7 +7,7 @@ keywords: network, ipvlan, l2, l3, standalone
 The IPvlan driver gives users total control over both IPv4 and IPv6 addressing.
 The VLAN driver builds on top of that in giving operators complete control of
 layer 2 VLAN tagging and even IPvlan L3 routing for users interested in underlay
-network integration. For overlay deployments that abstract away physical constraints 
+network integration. For overlay deployments that abstract away physical constraints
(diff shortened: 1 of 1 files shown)


jawys:
checked the deploy preview:
grafik
LGTM

craig-osterhout:
Thanks @jawys. I haven't looked too closely at it, so I may be wrong, it looks like the `$$` is there to differentiate running a command in the container as opposed to the host.
It probably should be made more clear using different syntax. We're currently trying to standardize this, so if any suggestions, please let us know.

The change for removing the leading space looks good. If you want to revert the `$$` changes, I can merge this.

jawys:
Oh I see, maybe I replaced the `$$` too fast :D

Previously I stumbled upon the following part, which looks indeed as if there is some mixup there:

grafik

---

What I like about the single `$` is the fact, that the "Copy code to clipboard" button removes the `$` from the code block which I can then paste into the terminal. This saves me time while quickly fiddling stuff out.

Nevertheless Ill fixup the commit and remove the replacements as you wish.

craig-osterhout:
Thanks @jawys.

> @craig-osterhout if you agree I`d like to add the following:
>
> @@ -461,7 +461,8 @@ Start a container and view eth0 and both v4 & v6 routing tables:
> ```console
> $ docker run --net=ipvlan140 --ip6=2001:db8:abc2::51 -it --rm alpine /bin/sh
> 
> -$ ip a show eth0
> +# Inside the container
> +$$ ip a show eth0

I suggest open this in a separate issue/PR because it will require more scrutiny or an SME to approve, which will take longer. 

Thanks again for the feedback about the copy to clipboard feature. We're currently trying track down all the different prompts and standardize them in order to fix this.

craig-osterhout (approved):
LGTM

jawys:
> I suggest open this in a separate issue/PR because it will require more scrutiny or an SME to approve, which will take longer.

Thanks @craig-osterhout. I now opened another PR (#16109) for these changes.

jawys:
> Thanks again for the feedback about the copy to clipboard feature. We're currently trying track down all the different prompts and standardize them in order to fix this.

Another suggestion for the container shell syntax: 
If one would use `#$`instead of `$$` as prefix, the "copy to clipboard" would help to just copy a comment into my host shell instead of the commands which shall be executed in the container's shell.

What do you think?

PR #16077, opened 2022-11-04
fixed typo: http -> https in https proxy URL

alodish:


dvdksn:
Hi and thanks for your PR. This is not an error. `http` in the example illustrates the protocol that the proxy uses. Hence I am closing this.

PR #16082, opened 2022-11-04
mac-install-refine

aevesdocker:
Based off of some feedback received that the Install on Apple silicon page can be merged with the Mac install page as there is already repeated info. 

This PR only restructures and moves content around - no new words have been added


dvdksn (commented):
I think we should also add rel=noopener and target=blank for external links

aevesdocker:
> I think we should also add rel=noopener and target=blank for external links

Agreed! I have it on my to-do list to go through all DD docs and make sure links, code blocks, formatting all align with the Style Guide. Am hoping to get to it today

dvdksn (approved):
:shipit:

PR #16087, opened 2022-11-04
eci-patch

aevesdocker:


desktop/hardened-desktop/enhanced-container-isolation/index.md
```diff
@@ -79,10 +79,16 @@ For this to take effect:
...
 
 ```
-$ docker run --rm alpine cat /proc/self/uid_map 
+$ docker run -it --rm alpine cat /proc/self/uid_map 
+```
+
+Then:
+
+```
+cat /proc/self/uid_map
```
dvdksn, line 91: This command is already added to the first command as the entrypoint for the container, meaning it will be run automatically and print out the user info

aevesdocker: ah ok, was asked by the PM to split it up. Will change back

aevesdocker: thank you!

desktop/hardened-desktop/enhanced-container-isolation/index.md
```diff
@@ -79,10 +79,10 @@ For this to take effect:
 
 When Enhanced Container Isolation is enabled, users see that containers run within a Linux user namespace. 
 
-To check, run:
+To check, first run:
 
 ```
-$ docker run --rm alpine cat /proc/self/uid_map 
+$ docker run -it --rm alpine cat /proc/self/uid_map
```
dvdksn, line 85: the `-it` flag is only needed if you want to start an interactive shell in the container. In this case, if we pass the entrypoint parameters to `docker run`, they are not needed.

PR #12083, opened 2021-01-12
_includes/cli.md: fix flag descriptions with < and >

thaJeztah:
fixes https://github.com/docker/docker.github.io/issues/11331
replaces / closes https://github.com/docker/docker.github.io/pull/12064

Some flag descriptions contain point-brackets to indicate required options, e.g.:

    --ssh stringArray   SSH agent socket or keys to expose to the build (format: default|<id>[=<socket>|<key>[,<key>]])

When rendering those options as HTML, those options were not visible as
they were rendered as a HTML element.

Given that flag-descriptions are not expected to have MarkDown or HTML
formatting, we can HTML-escape them to prevent this.

This patch escapes the flag-descriptions using liquid's `esacape` command.

Before this patch:

Screenshot 2021-01-12 at 16 20 38

After this patch:

Screenshot 2021-01-12 at 16 20 16

PR #16093, opened 2022-11-05
_includes/cli.md: do not escape flag descriptions

crazy-max:
revert changes https://github.com/docker/docs/pull/12083

`--ssh` flag with buildx is now properly escaped upstream but with #12083 it now also escapes inline code: https://docs.docker.com/engine/reference/commandline/buildx_build/

image

### DIFF
--- a/_includes/cli.md
+++ b/_includes/cli.md
@@ -138,7 +138,7 @@ For example uses of this command, refer to the [examples section](#examples) bel
     <td markdown="span">`--{{ option.option }}`{% if option.shorthand %} , `-{{ option.shorthand }}`{% endif %}</td>
     {%- endif %}
     <td markdown="span">{{ option-default }}</td>
-    <td markdown="span">{% if all-badges != '' %}{{ all-badges | strip }}<br />{% endif %}{{ option.description | strip | escape }}</td>
+    <td markdown="span">{% if all-badges != '' %}{{ all-badges | strip }}<br />{% endif %}{{ option.description | strip }}</td>
   </tr>
 {% endfor %} <!-- end for option -->
 </tbody>


dvdksn (dismissed):
lgtm!

thaJeztah:
Looks like we have to re-generate the YAMLdocs for some other pages from the CLI repository;

- https://deploy-preview-16093--docsdocker.netlify.app/engine/reference/commandline/build/  (`--ssh` flag) 
- https://deploy-preview-16093--docsdocker.netlify.app/engine/reference/commandline/run/ (`--user` - well, it _seems_ to work, because they're not considered HTML tags)
- https://deploy-preview-16093--docsdocker.netlify.app/engine/reference/commandline/commit/ (`--author`)
- https://deploy-preview-16093--docsdocker.netlify.app/engine/reference/commandline/container_prune/ (`--filter`)
- https://deploy-preview-16093--docsdocker.netlify.app/engine/reference/commandline/network_prune/ (`--filter`)
- https://deploy-preview-16093--docsdocker.netlify.app/engine/reference/commandline/system_prune/ (`--filter`)
- https://deploy-preview-16093--docsdocker.netlify.app/engine/reference/commandline/volumes_prune/ (`--filter`)


Screenshot 2022-11-05 at 17 25 27
Screenshot 2022-11-05 at 17 27 26
Screenshot 2022-11-05 at 17 28 48

crazy-max:
@thaJeztah Oh indeed good catch

crazy-max:
@thaJeztah @dvdksn PTAL. Should be good to go now.

thaJeztah:
Changes LGTM, but I now realise (and may have commented on the CLI side) that `docker build` no longer documents any BuildKit options (as we removed those, not hidden them). We need to find a solution to that, otherwise people may think you can only use those options with `docker buildx build`, and the goal was to make `docker build` now provide all the options that buildkit provides

crazy-max:
> but I now realise (and may have commented on the CLI side) that `docker build` no longer documents any BuildKit options (as we removed those, not hidden them). We need to find a solution to that, otherwise people may think you can only use those options with `docker buildx build`, and the goal was to make `docker build` now provide all the options that buildkit provides

Yes I think we should do smth, left a comment in https://github.com/docker/docs/pull/16416#issuecomment-1413908886 about it.

dvdksn (approved):

thaJeztah (approved):
oh! LGTM (thought I did); feel free to merge if we want to get this one in 👍

PR #16094, opened 2022-11-05
build: context page

crazy-max:
Adds build context information. In follow-up we could cross link to this page from `docker build` and `docker buildx build` command ref docs pages.


build/building/context.md
```diff
@@ -0,0 +1,168 @@
...
+for better performance with [BuildKit](../buildkit/index.md). The build context
+is processed recursively. So, a `PATH` includes any subdirectories and the
+`URL` includes the repository and its submodules.
+
+## `PATH` context
+
+When building with [BuildKit](../buildkit/index.md), which sets up an
+"interactive" session, only requested files that are needed by the build from
+the client are sent over the entire build context. On the other hand, the
+legacy builder will send the entire context (recursively) to the daemon.
```
dvdksn, line 25:
```suggestion
When building with [BuildKit](../buildkit/index.md), the client only sends
the files that the daemon requires to complete the build. This is in contrast
with the legacy builder, where the client always sends the entire context,
recursively, to the daemon.
```

build/building/context.md
```diff
@@ -0,0 +1,168 @@
...
+
+To increase the build's performance, exclude files and directories by adding
+a [`.dockerignore` file](../../engine/reference/builder.md#dockerignore-file) to
+the context directory.
```
dvdksn, line 29: The previous paragraph makes it sound like we don't need to use a `.dockerignore` file if we are using BuildKit, because only files used in the build are sent over to the daemon anyway. What role does the `.dockerignore` file play in this context?

dvdksn (commented):
looking forward to having this page in!

build/building/context.md
```diff
@@ -0,0 +1,168 @@
...
+succeeds, the directory is sent to the daemon as the context. Local copy gives
+you the ability to access private repositories using local user credentials,
+VPN's, and so forth.
+
+> **Note**
+>
+> If the `URL` parameter contains a fragment the system will recursively clone
+> the repository and its submodules using a `git clone --recursive` command.
+
+Git URLs accept context configuration in their fragment section, separated by a
```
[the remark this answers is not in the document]
dvdksn: actually my suggestion here is not good. I think it should read something like, "Git URLs accept a context configuration parameter in the form of a URL fragment."

build/building/context.md
crazy-max: > What role does the `.dockerignore` file play in this context?

With this instruction:

```dockerfile
COPY . .
```

All files will be sent anyway (BuildKit or legacy). The `.dockerignore` file can be useful to exclude some of them without having multi COPY instructions in your Dockerfile for the files/folders you need.

Also with BuildKit if you don't have `COPY` commands in the stages you are building there is no build context sent at all. If you have `COPY foo .` only file in build context for that build would be `foo` whereas legacy builder still send the whole context.

crazy-max: I've added more explanation on this page, let me know if it seems clear enough.

dvdksn (approved):
Nice

build/building/context.md
```diff
@@ -0,0 +1,206 @@
...
+In some cases you may want to send the entire context:
+
+```dockerfile
+FROM busybox
+WORKDIR /src
+COPY . .
+```
+
+But exclude some files or directories. You can use a [`.dockerignore`](../../engine/reference/builder.md#dockerignore-file)
+file to exclude them from being sent like:
```
jedevc, line 73:
```suggestion
You can use a [`.dockerignore`](../../engine/reference/builder.md#dockerignore-file) file to exclude some files or directories from being sent:
```

build/building/context.md
```diff
@@ -0,0 +1,206 @@
...
+> as it causes the build to transfer the entire contents of your hard drive to
+> the daemon.
+{:.warning}
+
+## `URL` context
+
+The `URL` parameter can refer to three kinds of resources:
+* [Git repositories](#git-repositories)
+* Pre-packaged [tarball contexts](#tarball-contexts)
+* Plain [text files](#text-files)
```
jedevc, line 93: We also support pre-built binaries.

Maybe "plain text files" might be better as "regular files" in the Unix sense?

dvdksn: Hmm. What's an **irregular** file then?

build/building/context.md
```diff
@@ -0,0 +1,206 @@
...
+
+### Git repositories
+
+When the `URL` parameter points to the location of a Git repository, the
+repository acts as the build context. The builder recursively pulls the
+repository and its submodules. The pulls ignore the commit history. A
```
jedevc, line 99: "The pulls ignore the commit history"

Can we add some clarification for this? Is this just removing the `.git` directory, or is it a shallow-clone, or something else?

build/building/context.md
jedevc: Ugh nope, yup that comment doesn't make sense :smile: Why would we be using a binary as a context.

jedevc (dismissed):

dvdksn (approved):

jedevc (approved):

PR #16101, opened 2022-11-06
jekyll: cleanup redirect only stub files and merge config

crazy-max:
Removes unnecessary stub files that were just used for redirects. This is now handled through a new plugin `pageless_redirects.rb`. Pageless redirects can be set in `_redirects.yml` file.

Also found out that scopes from `_config.yml` were overridden by the production one which doesn't have proper sitemap exclusions. Now scopes and exclusions configuration are merged in `_config.yml`.


dvdksn (approved):
259? files deleted in a single commit has to be one of the most beautiful things I have ever seen.

PR #16105, opened 2022-11-06
Clarify instantly that Docker Desktop for Linux uses its own context

goulashsoup:
### Proposed changes

Clarifying immediately that Docker Desktop for Linux has its own context would greatly improve the documentation because it would prevent further misunderstandings regarding allegedly missing images and containers, as well as problems regarding apps that use the default docker socket (`/var/run/docker.sock`) for communication (e.g. PhpStorm).

### DIFF
--- a/desktop/install/linux-install.md
+++ b/desktop/install/linux-install.md
@@ -11,6 +11,15 @@ redirect_from:
 > employees OR more than $10 million USD in annual revenue) requires a paid
 > subscription.
 
+>Important
+>
+>Be aware that Docker Desktop on Linux runs a Virtual Machine (VM), creates and uses a custom docker context `desktop-linux` on startup. 
+>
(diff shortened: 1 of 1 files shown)


desktop/install/linux-install.md
```diff
@@ -11,6 +11,15 @@ redirect_from:
 > employees OR more than $10 million USD in annual revenue) requires a paid
 > subscription.
 
+>Important
+>
+>Be aware that Docker Desktop on Linux runs a Virtual Machine (VM), creates and uses a custom docker context `desktop-linux` on startup.
```
aevesdocker, line 16:
```suggestion
>Docker Desktop on Linux runs a Virtual Machine (VM) so creates and uses a custom docker context `desktop-linux` on startup. 
```

aevesdocker (approved):
Thank you for the PR @goulashsoup

Issue #15874, opened 2022-10-14
the version property is deprecated


ver-alex:
[get-started/08_using_compose.md](https://docs.docker.com/get-started/08_using_compose/)

The top-level attribute version is marked as DEPRECATED since beginning of 2021 and should be removed from the example of the compose documentation.

https://github.com/compose-spec/compose-spec/commit/85d1bcfc07462fbfc645882a4ca1901012075cf6

NitishKumar06:
@ver-alex Would it be a good practice if we remove everything that includes about version?

ver-alex:
I think it is confusing to use a deprecated attribute and should be removed. Version is not used anymore if you are on a recent and updated version of docker compose.

As the compose spec describes this here: https://github.com/compose-spec/compose-spec/blob/master/spec.md#version-top-level-element

_Top-level version property is defined by the specification for backward compatibility but is only informative._

_A Compose implementation SHOULD NOT use this version to select an exact schema to validate the Compose file, but prefer the most recent schema at the time it has been designed._

NitishKumar06:
@ver-alex How about we change the version of docker-compose to the latest version?

ver-alex:
Since the version element is (or should not be) used by the compose parser it has no functionality and is not a necessary part in a compose file anymore. I would just remove it. Guess that over time no compose files will have the version property.
Also, since no version is stated, the documentation will not become outdated 😉.

NitishKumar06:
Would you mind raising a PR for this issue? Reviewers would be happy to see the changes you make.

PR #16127, opened 2022-11-09
remove deprecated version property

ver-alex:
### Proposed changes

Remove the version property from the examples since it is deprecated according to the compose-spec https://github.com/compose-spec/compose-spec/blob/master/spec.md#compose-file

### Related issues (optional)

#15874


glours (approved):
Thanks @ver-alex, sounds good to me

PR #16143, opened 2022-11-11
Refine wording in swarm tutorial

craig-osterhout:
### Proposed changes

"Alone" in the second bullet point can be ambiguous. For example, alone as a product or alone on one host. 

### Related issues (optional)
ENGDOCS-1082


craig-osterhout:
Closing. Further input is to consider removing.

PR #16166, opened 2022-11-15
Add code formatting checks to vale

craig-osterhout:
### Proposed changes

Add 3 code formatting checks to vale:
- Check if code fences have indentation that is not 0, 3, or 5.
- Check if code blocks contain multiple lines with command prompts ($, #).
- Check if code blocks have stuff (output?) following a command that does not have ` or \ to wrap the command, and is not another command with # or $.


dvdksn:
I spent some time with this, and I think that the we should stick to only having a lint rule for the indentation at this point.

craig-osterhout:
Thanks, @dvdksn. The 2 other styles have been removed.

dvdksn (approved):

Swetik1902 (commented):
Гл

PR #16174, opened 2022-11-17
SDK landing page reorg

aevesdocker:
This PR addresses feedback from users of the Extensions SDK docs. Currently, some of the key pages are buried in the toc. This PR brings out the pages on the landing page, for greater visibility.


dvdksn (approved):
nd

I wonder if at some point we should consider moving the SDK reference to the reference section? Also the `docker extension` subcommand should be a part of the CLI reference shouldn't it?

aevesdocker:
@dvdksn I've had the same thought. I think though that perhaps it would be better to wait until the SDK is out of the beta phase before moving it out of Desktop section?
And yes you're right, the subcommand should move. I'll add to the to-do list.

PR #16183, opened 2022-11-17
index.md: make port numbers consistent

zipperer:
replace a use of 8090 with a use of 8080


zipperer:
I saw only the 8080 mapping when I made this change. I did not see the 8090 mapping. facepalm.

PR #16201, opened 2022-11-22
Align with Design System

aevesdocker:


dvdksn:
doesn't seem like the font weights are being applied. Might be due to how the font is loaded?

aevesdocker:
Good catch, thanks @dvdksn! I think I fixed it. There are still some changes to be made, such as dark mode - haven't touched that yet.  Have a working session with Megan and Rebeca on Monday to sort this 🤞

dvdksn (commented):
There seems to be some alignment problem with the icons/switch on the landing page. 
Screenshot 2022-11-30 at 16 03 43

Some of the raw HTML blocks that we've added are slightly out of style, which is mostly fine but I found the nav bar for the getting started guides to be a little extra. Should we sort out the colors for these too?

dvdksn (approved):

crazy-max:
@aevesdocker Looks hard to read links in warning callouts: https://docs.docker.com/build/buildx/install/#manual-download

image

Before: https://docs-labs.docker.com/build/buildx/install/

image

crazy-max:
Also find hard to read long paragraphs: https://docs.docker.com/desktop/mac/permission-requirements/

image

I think we should increase letter-spacing: https://docs-labs.docker.com/desktop/mac/permission-requirements/

image

Would look like this with the new design:

```css
.content, p {
  line-height: 24px;
  clear: both;
  letter-spacing: 0.05em;
}
```

image

aevesdocker:
Thank you for the feedback @crazy-max  and for catching the color contrast issue - have new PR to fix this https://github.com/docker/docs/pull/16257 Text definitely looks much better with the increased letter spacing

PR #16216, opened 2022-11-25
A quick search for "symlink" on the docker issues will return hundred…

danshome:
A quick search for "symlink" on the docker issues will return hundreds of reported problems where the problem turned out to be missing symlinks that are necessary for Docker to function properly. Until the Docker install symlink issues are resolved, the problem needs to be acknowledged in the Known Issues for MacOS with the list of symlinks that need to exist.

### DIFF
--- a/desktop/troubleshoot/known-issues.md
+++ b/desktop/troubleshoot/known-issues.md
@@ -10,6 +10,24 @@ title: Known issues for Docker Desktop on Mac
 <div class="tab-content">
 <div id="tab3" class="tab-pane fade in active" markdown="1">
 <br>
+
+* For various reasons the symlinks necessary for the Docker to function properly don't get created during the Docker install process.
+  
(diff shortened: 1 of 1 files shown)


danshome:
Here is an existing Support Case related to this:
- Case #: 00047468
- Subject: unknown shorthand flag: 'f' in -f

aevesdocker:
@mat007 would you mind taking a quick look at this? Can this PR be merged?

mat007:
I believe this is supposed to have been fixed in 4.14.1, after being introduced in 4.14.0:
> Fixed a bug causing symlinks to not be created for the user if /usr/local/lib doesn’t already exist. Fixes [docker/for-mac#6569](https://github.com/docker/for-mac/issues/6569)

In https://docs.docker.com/desktop/release-notes/#bug-fixes-and-enhancements

aevesdocker:
Thanks for checking @mat007. Closing PR

PR #16224, opened 2022-11-25
Docs: Update Dockerfile base image and python version

shpiyu:
changed base image from node:12-alpine to node:19-alpine as new version of Jest requires node 18 or above.

### Proposed changes

Updated the base image in the Docker file from node:12-alpine to node:19-alpine because Jest expected a newer version.
Also updated python to python3, as python2 package was not found.


craig-osterhout:
Thank you for your contribution. This has been fixed by a larger scope update.

PR #16226, opened 2022-11-26
Example is outdated and throws errors

TimurGafarovR:
Above changes fix the error. Versions just have to be bumped up.


craig-osterhout:
Thank you for the contribution. This has been fixed by a larger scope update.

PR #16234, opened 2022-11-28
Use the correct error message when trying to inspect a nonexisting service

zombieleet:
### Proposed changes

Uses the correct error message when inspecting a non exiting service


zombieleet:
@dvdksn

dvdksn (approved):
thank you!

PR #16255, opened 2022-11-30
fix variable name typo

dannysauer:
### Proposed changes

Fix minor env var name typo


aevesdocker (approved):
Thank you @dannysauer!

PR #16257, opened 2022-12-01
docs-refresh2

aevesdocker:
Improves typography and link color contrast on important note

### DIFF
--- a/_scss/_typography.scss
+++ b/_scss/_typography.scss
@@ -152,6 +152,7 @@ h4, .h4 {
 p, .body1 {
   font-size: 1rem;
   line-height: 1.5;
+  letter-spacing: 0.05rem;
 }
 
 .body2 {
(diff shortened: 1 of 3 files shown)


_scss/_typography.scss
```diff
@@ -152,6 +152,7 @@ h4, .h4 {
 p, .body1 {
   font-size: 1rem;
   line-height: 1.5;
+  letter-spacing: 0.05rem;
```
crazy-max, line 155: @mark-dr I'm not sure if it should be a font property or letter-spacing change? WDYT? See https://github.com/docker/docs/pull/16201#issuecomment-1332933442

aevesdocker:
@crazy-max upon reflection, I've rolled back the letter-spacing change. It works alright in dark mode but for light mode it is not very good and arguably makes the text harder to read. 

I tried increasing the text from 14px to 16px instead, but this also is subpar and looks messy. 

I've settled on increasing the margin between paragraphs. This makes large chunks of text easier to scan. 

I think it just takes a bit of time for the eyes to adjust to the new font (which unfortunately was not chosen with docs in mind).

cc @usha-mandya

dvdksn (approved):

PR #16263, opened 2022-12-02
Adjust instructions for Docker Desktop on Mac uninstall from command line

mat007:
### Proposed changes

Adjusted the instructions for Docker Desktop on Mac uninstall from the command line.
The `--uninstall` flag was removed from the Docker Desktop application with 4.4.2, but we’ve never replaced it with anything else as of yet. The new command will be available in Docker Desktop 4.16.

### DIFF
--- a/desktop/uninstall.md
+++ b/desktop/uninstall.md
@@ -39,13 +39,13 @@ To uninstall Docker Desktop from your Mac:
 
 > Uninstall Docker Desktop from the command line
 >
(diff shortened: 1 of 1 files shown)


aevesdocker (approved):
Thank you @mat007

mat007:
Oh that’s not in the product yet actually. As the description mentioned, it’s only going to be available in the next version of Docker Desktop.

PR #16265, opened 2022-12-02
Desktop: update macOS min versions, post Ventura release

djs55:
### Proposed changes

Updated the Docker Desktop macOS version support statement.

The stated policy is latest + 2 previous which means Big Sur (11), Monterey (12) and Ventura (13).

Note that Apple seem to have dropped the dot and minor digit, see https://en.wikipedia.org/wiki/MacOS_version_history


aevesdocker (approved):
Thank you @djs55!

PR #16295, opened 2022-12-07
night-mode: don't override search background color

thaJeztah:
The color of the top-navigation for "light" and "night" mode is the same, so we don't need to override the color (for now). The old override used a hue that was slightly out of tone with the rest of the theme.

Before:

Screenshot 2022-12-07 at 12 34 54

After:

Screenshot 2022-12-07 at 12 37 33


thaJeztah:
@usha-mandya ptal

usha-mandya (approved):
LGTM. Thanks.

PR #16297, opened 2022-12-07
fix condition for hotjar include and move include

thaJeztah:
### fix condition for hotjar include

The condition checked if the hotjar-id wasn't empty, but did not
check if it was "set". Because of that, the script would still be
included, producing a JavaScript error.

Screenshot 2022-12-07 at 12 13 28

### move hotjar include to analytics

Moving it together with similar includes.


thaJeztah:
@crazy-max @usha-mandya ptal

thaJeztah:
Let's get this one in 👍

Issue #53, opened 2016-10-05
Feedback for: engine/reference/builder.md


nevil:
URL: engine/reference/builder.md

The format for this page is broken.

nevil:
Will be fixed by #47
Sorry for the noise.

PR #16318, opened 2022-12-11
Update context.md

hkchai:
Line #53, isn't the directory name should be '/src' instead of '.'?

### Proposed changes

Line #53, isn't the directory name should be '/src' instead of '.'?

### DIFF
--- a/build/building/context.md
+++ b/build/building/context.md
@@ -50,7 +50,7 @@ COPY foo .
 And this directory structure:
 
 ```
-.
+/src
 ├── Dockerfile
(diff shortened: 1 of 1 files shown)


crazy-max:
> Line #53, isn't the directory name should be '/src' instead of '.'?
> ### Proposed changes
> 
> Line #53, isn't the directory name should be '/src' instead of '.'?
> ### Related issues (optional)

Local context is not related to the working dir in the Dockerfile. Here we show the current directory structure where we type the build command.

PR #16320, opened 2022-12-12
build: move bake under the build section

crazy-max:
As discussed, we want to move some sections directly under "Docker Build" section. This PR moves the Bake one to start with.


dvdksn (approved):
❤️

PR #16323, opened 2022-12-12
build: docker-container custom network

crazy-max:
moved from https://github.com/docker/buildx/blob/master/docs/guides/custom-network.md

@dvdksn Maybe wording could be improved, let me know.

### DIFF
--- a/build/building/drivers/docker-container.md
+++ b/build/building/drivers/docker-container.md
@@ -139,6 +139,44 @@ $ docker buildx build \
 > slower than native builds. Compute-heavy tasks like compilation and
 > compression/decompression will likely take a large performance hit.
 
+## Custom network
+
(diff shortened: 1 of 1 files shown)


dvdksn (approved):
Thanks Kevin, I think the wording is good 👍🏻 but, it occurred to me while reading this: I don't think we have any information about networking for build(x)? Like, what problem this can help solve and when you would use it (integration testing? using a proxy? idk)

Or am I missing something?

jedevc (approved):
LGTM :tada:

CC @tianon, I remember discussing this option (and it's lack of docs) previously :smile:

crazy-max:
> Like, what problem this can help solve and when you would use it (integration testing? using a proxy? idk)

It can be useful if you want to connect to an extra service in the same network. Like a db to run integration tests for example.

PR #16324, opened 2022-12-12
build: buildkit resource limiting

crazy-max:
moved from https://github.com/docker/buildx/blob/master/docs/guides/resource-limiting.md

### DIFF
--- a/build/buildkit/configure.md
+++ b/build/buildkit/configure.md
@@ -178,3 +178,37 @@ $ docker buildx create --use --bootstrap \
   --driver-opt "image=buildkit-cni:local" \
   --buildkitd-flags "--oci-worker-net=cni"
 ```
+
+## Resource limiting
+
+### Max parallelism
+
+You can limit the parallelism of the BuildKit solver, which is particularly useful
+for low-powered machines, using a [BuildKit configuration](toml-configuration.md)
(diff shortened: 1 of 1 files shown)


build/buildkit/configure.md
```diff
@@ -178,3 +178,37 @@ $ docker buildx create --use --bootstrap \
...
+  --config /etc/buildkitd.toml
+```
+
+### Limit on TCP connections
+
+We are also limiting TCP connections to **4 per registry** with an additional
+connection not used for layer pulls and pushes. This limitation will be able to
+manage TCP connection per host to avoid your build being stuck while pulling
+images. The additional connection is used for metadata requests
+(image config retrieval) to enhance the overall build time.
```
dvdksn, line 212: Is this a fair rewording?

```suggestion
TCP connections are limited to 4 simultaneous connections per registry for pulling
and pushing images, plus one additional connection dedicated to metadata requests. 
This connection limit prevents your build from getting stuck while pulling images.
The dedicated metadata connection helps reduce the overall build time.
```

dvdksn (changes requested):

jedevc:
LGTM, once @dvdksn's comments are resolved :+1:

dvdksn (approved):

PR #16405, opened 2022-12-27
Add missing `container` notation

tuxiqae:

### DIFF
--- a/config/formatting.md
+++ b/config/formatting.md
@@ -84,7 +84,7 @@ $ docker inspect --format "{{lower .Name}}" container
 
 {% raw %}
 ```console
-$ docker inspect --format '{{split .Image ":"}}'
+$ docker inspect --format '{{split .Image ":"}}' container
 ```
 {% endraw %}
 


config/formatting.md
```diff
@@ -84,7 +84,7 @@ $ docker inspect --format "{{lower .Name}}" container
 
 {% raw %}
 ```console
-$ docker inspect --format '{{split .Image ":"}}'
+$ docker inspect --format '{{split .Image ":"}}' container
```
dvdksn, line 87: Just a small suggestion to add angle brackets for the container name placeholder to make it appear more like a placeholder than a sub-command

```suggestion
$ docker inspect --format '{{split .Image ":"}}' <container>
```

tuxiqae: I wholeheartedly agree, just bear in mind that those changes should be made to the rest of the examples as well and not only to these one.

dvdksn: oh yeah, I didn't check the rest of the doc. We can merge it like this and I can take the brackets in a follow-up.

tuxiqae: Should be quite quick to implement, want me to just fix it all in this PR?

dvdksn: sure, if you dont mind! thank you!

dvdksn (approved):
Bringing this one in, and then we can fix the placeholder style in a followup

PR #16413, opened 2022-12-31
Fix Windows kubectl.exe path in kubernetes.md

ninjaoxygen:
Kubectl.exe path has a DOS prompt ">" character in it, but then refers to configuring this as part of PATH. If the prompt character is copied into the PATH environment variable, it will not work.


mat007 (approved):
Good catch, thanks!

aevesdocker (approved):

PR #16419, opened 2023-01-02
Update Example workflows page

Robert-96:
### Proposed changes

Rename `secrets.DOCKERHUB_USERNAME` and `secrets.DOCKERHUB_PASSWORD` to `secrets.DOCKER_HUB_USERNAME` `secrets.DOCKER_HUB_ACCESS_TOKEN`. This way the tutorial from the [Introduction to GitHub Actions](https://docs.docker.com/build/ci/github-actions/) and the [Example workflows](https://docs.docker.com/build/ci/github-actions/examples/) page are using the same secret names.

### DIFF
--- a/build/ci/github-actions/examples.md
+++ b/build/ci/github-actions/examples.md
@@ -38,8 +38,8 @@ jobs:
         name: Login to Docker Hub
         uses: docker/login-action@v2
         with:
-          username: ${{ secrets.DOCKER_HUB_USERNAME }}
-          password: ${{ secrets.DOCKER_HUB_ACCESS_TOKEN }}
+          username: ${{ secrets.DOCKERHUB_USERNAME }}
+          password: ${{ secrets.DOCKERHUB_TOKEN }}
       -
         name: Login to GitHub Container Registry
         uses: docker/login-action@v2
@@ -121,8 +121,8 @@ jobs:
         if: github.event_name != 'pull_request'
         uses: docker/login-action@v2
         with:
-          username: ${{ secrets.DOCKER_HUB_USERNAME }}
-          password: ${{ secrets.DOCKER_HUB_ACCESS_TOKEN }}
+          username: ${{ secrets.DOCKERHUB_USERNAME }}
+          password: ${{ secrets.DOCKERHUB_TOKEN }}
       -
         name: Login to GHCR
         if: github.event_name != 'pull_request'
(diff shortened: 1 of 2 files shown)


jedevc:
Hey @Robert-96, thanks so much for the contribution, nice catch! :tada:

Consistency is good here, but I think we might want to aim to be consistent with the docs we already have for [docker/build-push-action](https://github.com/docker/build-push-action) and [docker/login-action](https://github.com/docker/login-action].

In these repos we use `DOCKERHUB_USERNAME` and `DOCKERHUB_TOKEN`:

```yaml

jobs:
  login:
    runs-on: ubuntu-latest
    steps:
      -
        name: Login to Docker Hub
        uses: docker/login-action@v2
        with:
          username: ${{ secrets.DOCKERHUB_USERNAME }}
          password: ${{ secrets.DOCKERHUB_TOKEN }}
```

I think if we change the [Introduction to GitHub Actions](https://docs.docker.com/build/ci/github-actions/) and the [Example workflows](https://docs.docker.com/build/ci/github-actions/examples/) to this, then we have consistency across *all* of our docs, which should make it a lot easier to move from the introduction, into the examples, and then into the reference documentation.

build/ci/github-actions/examples.md
```diff
@@ -919,10 +919,10 @@ jobs:
           tags: user/app:latest
       -
         name: Update repo description
-        uses: peter-evans/dockerhub-description@v2
+        uses: peter-evans/dockerhub-description@v3
         with:
-          username: ${{ secrets.DOCKERHUB_USERNAME }}
-          password: ${{ secrets.DOCKERHUB_PASSWORD }}
+          username: ${{ secrets.DOCKER_HUB_USERNAME }}
+          password: ${{ secrets.DOCKER_HUB_ACCESS_TOKEN }}
```
crazy-max, line 925: This one is on purpose as this action requires plain password.

```suggestion
          password: ${{ secrets.DOCKER_HUB_PASSWORD }}
```

crazy-max (changes requested):

build/ci/github-actions/examples.md
jedevc: Since this is different, probably worth adding a comment to indicate why :+1:

Robert-96: In the docs is mentioned that a PAT or a password is required: 

> Docker Hub password or [Personal Access Token](https://docs.docker.com/docker-hub/access-tokens/) with `read/write/delete scope`.

https://github.com/peter-evans/dockerhub-description#action-inputs

I managed to run the action successfully with a PAT.

Robert-96:
@jedevc I updated all the examples to use `DOCKERHUB_USERNAME` and `DOCKERHUB_TOKEN`.

build/ci/github-actions/examples.md
crazy-max: Ohh didn't know we could use a PAT now! Nice

jedevc (approved):

crazy-max (approved):

PR #16428, opened 2023-01-04
doc: add documentation for `READY` state

zombieleet:
### Proposed changes

The `READY` state for worker nodes was omitted from the documentation

https://github.com/moby/swarmkit/blob/48dd89375d0a3e8a333c52bce1e6297e0b192d06/api/types.proto#L509


dvdksn (approved):
TYVM for catching this!

PR #16429, opened 2023-01-04
update version validation command

gnguedescelf:


craig-osterhout:
Thanks for sharing the PR. The current command, `docker compose version`, appears to be correct based on the latest version.
If you ran into an issue running the command, ensure you have the latest version of Docker Desktop installed. If you need help troubleshooting, please seek [support](https://docs.docker.com/support/).

PR #16446, opened 2023-01-05
ENGDOCS-1063

aevesdocker:
This PR makes some minor adjustments mostly aligning the text to the Style Guide

### DIFF
--- a/compose/production.md
+++ b/compose/production.md
@@ -14,15 +14,15 @@ up your application, you can run Compose apps on a Swarm cluster.
 
 ### Modify your Compose file for production
 
-You probably need to make changes to your app configuration to make it ready for
-production. These changes may include:
+you may need to make changes to your app configuration to make it ready for
+production. These changes could include:
 
 - Removing any volume bindings for application code, so that code stays inside
   the container and can't be changed from outside
 - Binding to different ports on the host
 - Setting environment variables differently, such as reducing the verbosity of
   logging, or to specify settings for external services such as an email server
-- Specifying a restart policy like `restart: always` to avoid downtime
+- Specifying a restart policy like [`restart: always`](compose-file/index.md#restart){: target="_blank" rel="noopener" class="_" } to avoid downtime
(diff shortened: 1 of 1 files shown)


compose/production.md
```diff
@@ -14,24 +14,23 @@ up your application, you can run Compose apps on a Swarm cluster.
...
+You may need to make changes to your app configuration to make it ready for
+production. These changes might include:
 
 - Removing any volume bindings for application code, so that code stays inside
   the container and can't be changed from outside
 - Binding to different ports on the host
 - Setting environment variables differently, such as reducing the verbosity of
   logging, or to specify settings for external services such as an email server
-- Specifying a restart policy like `restart: always` to avoid downtime
+- Specifying a restart policy like [`restart: always`](compose-file/index.md#restart){: target="_blank" rel="noopener" class="_" } to avoid downtime
```
dvdksn, line 25: our current stylesheets do not provide any visual cue as to whether code elements are hyperlinked (they appear in the same color/style whether or not they are linked). Doesn't mean we can't link them, just that nobody is likely to realize that these are links. If we feel this link is important, I would put the link text separately. For example:

```suggestion
- Specifying a restart policy like `restart: always` to avoid downtime, see [command reference](compose-file/index.md#restart){: target="_blank" rel="noopener" class="_" }
```

dvdksn (approved):
non-blocking nit, otherwise LGTM!

compose/production.md
aevesdocker: Ah yes, good catch. As it's not an essential link I'm gonna be cheeky and just merge. But noted. How could we amend the stylesheet to make it clear? I guess one way to make the link stand out is with underlines...I know the Design System team raised this as something they would like, but it risks making the pages look awful IMHO

dvdksn: On the hugo POC site, [linked code still changes color](https://hugodocker.netlify.app/build/), but in a way that doesn't negatively affect the readability/contrast of the text (imo)

PR #16450, opened 2023-01-06
ENGDOCS-1061

aevesdocker:
PR aligns the 'Enable GPU access' page in Compose with the Style guide, makes sure it is referring to the Compose spec

To-do:

- [x] ask about 'Use of service `runtime` property from Compose v2.3 format (legacy)'

### DIFF
--- a/compose/gpu-support.md
+++ b/compose/gpu-support.md
@@ -10,19 +10,6 @@ The examples in the following sections focus specifically on providing service c
 You can use either `docker-compose` or `docker compose` commands.  
 See also, [Compose command compatibility with docker-compose](cli-command-compatibility.md){: target="_blank" rel="noopener" class="_" }.
 
-### Use of service `runtime` property from Compose v2.3 format (legacy)
-
-Docker Compose v1.27.0+ switched to using the Compose Specification schema which is a combination of all properties from 2.x and 3.x versions. This re-enabled the use of service properties as [runtime](/compose-file/compose-file-v2.md#runtime) to provide GPU access to service containers. However, this does not allow to have control over specific properties of the GPU devices.
-
-```yaml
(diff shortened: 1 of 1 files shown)


compose/gpu-support.md
```diff
@@ -4,46 +4,36 @@ keywords: documentation, docs, docker, compose, GPU access, NVIDIA, samples
...
-- [device_ids](https://github.com/compose-spec/compose-spec/blob/master/deploy.md#device_ids){:target="_blank" rel="noopener" class="_"} - value specified as a list of strings representing GPU device IDs from the host. You can find the device ID in the output of `nvidia-smi` on the host.
-- [driver](https://github.com/compose-spec/compose-spec/blob/master/deploy.md#driver){:target="_blank" rel="noopener" class="_"} - value specified as a string (eg. `driver: 'nvidia'`)
-- [options](https://github.com/compose-spec/compose-spec/blob/master/deploy.md#options){:target="_blank" rel="noopener" class="_"} - key-value pairs representing driver specific options.
+- `capabilities`. This value specifies as a list of strings (eg. `capabilities: [gpu]`). You must set this field in the Compose file. Otherwise, it returns an error on service deployment.
+- `count`. This value specified as an integer or the value `all` representing the number of GPU devices that should be reserved ( providing the host holds that number of GPUs). If no `count`is set, all GPUs available on the host are used by default.
```
dvdksn, line 20: Whitespace police 🚓 

```suggestion
- `count`. This value specified as an integer or the value `all` representing the number of GPU devices that should be reserved (providing the host holds that number of GPUs). If no `count` is set, all GPUs available on the host are used by default.
```

dvdksn (approved):

PR #16456, opened 2023-01-06
SDK form

aevesdocker:
Adds a feedback form the extensions-sdk docs


amyb12345 (approved):
LGTM

_includes/extensions-form.md
```diff
@@ -0,0 +1 @@
+Already built an extension? [Let us know about your experience](https://docs.google.com/forms/d/e/1FAIpQLSenjK8KiBiOrwCpq06u_iVEZGv4tfTEpfmY_cQDmsqJrHL9Gw/viewform).
```
thaJeztah, line 1: It looks like we're starting to use such feedback forms more frequently; these URLs will likely be temporary, and removed after some time.

I'm wondering if we should start using "permalink" redirects for these, like we do for the `/go/` redirects that we use when linking from our products; https://github.com/docker/docs/blob/380bf39b943201da19b537a71872c7a718d60922/_redirects.yml#L79-L93.

If we don't want to "pollute" the `/go/` namespace too much with such temporary URLs, we could consider to use another prefix (`/feedback/xxxx`) to be slightly clearer on intent, although `/go/` should still be an option, as long as it's somewhat clear what the purpose is/was; e.g. `/go/extensions-sdk-feedback/`.

Using those "friendly URL" redirects allows for;

- The URLs to be easy-to-use / communicate (`docs.docker.com/go/extensions-sdk-feedback/` is easier to remember than `docs.google.com/forms/<some very long ID>`)
- Once the form is no longer used, we can redirect the URLs to elsewhere (which could be just the extensions SDK overview page).
- If we need a new feedback form in future for the same purpose, we can re-use the old redirect to point to the new feedback form.

Of course we cannot prevent users referencing the actual (docs.google.com....) URL, but as long as we don't communicate the direct URL, and only communicate the "friendly" redirect-URL, that will be less common.

thaJeztah: Let me quickly ping @crazy-max to check if using a different (e.g. `/feedback/`) prefix would complicate things from a technical perspective (if we would decide to use that instead of `/go/`). Also wondering if "sub-directories" (`/go/feedback/<some easy name>`) would work with our current implementation (that way we keep things inside the `/go/` "namespace", but still have a bit of separation between "feedback" and other kind of redirects).

crazy-max: I think `/feedback/` is more explicit :+1:

Can be done like samples:

```yaml
# feedback
"https://docs.google.com/forms/d/e/1FAIpQLSenjK8KiBiOrwCpq06u_iVEZGv4tfTEpfmY_cQDmsqJrHL9Gw/viewform":
  - /feedback/extension/
```

aevesdocker: Thank you @thaJeztah, this is a much neater way of doing it and the benefits are clear. Have edited the PR accordingly 😃

_redirects.yml
```diff
@@ -550,3 +550,7 @@
 "https://hub.docker.com/_/zookeeper/":
   - /samples/zookeeper
   - /samples/library/zookeeper
+
+# feedback
+"https://docs.google.com/forms/d/e/1FAIpQLSenjK8KiBiOrwCpq06u_iVEZGv4tfTEpfmY_cQDmsqJrHL9Gw/viewform":
+  - /feedback/extension/
```
thaJeztah, line 556: I see it's missing a newline; not sure if this would cause issues, but perhaps better safe than sorry for that one 😅 

```suggestion
  - /feedback/extension/

```

thaJeztah (commented):
Thanks for updating! Left one comment for the missing newline, but looks good other than that 👍

_includes/extensions-form.md
```diff
@@ -0,0 +1 @@
+Already built an extension? [Let us know about your experience](https://docs.docker.com/feedback/extension/).
```
thaJeztah, line 1: Oh, this could probably use `/feedback/extension/` without the domain name;

```suggestion
Already built an extension? [Let us know about your experience](/feedback/extension/).
```

_includes/extensions-form.md
```diff
@@ -0,0 +1 @@
+Already built an extension? [Let us know about your experience](/feedback/extension/).
```
dvdksn, line 1: nit: since we are always calling this partial from a blockquote, I would move the `> ` syntax to be part of the include instead of duplicating it when reusing the snippet.

aevesdocker: valid nit - corrected!

_includes/extensions-form.md
```diff
@@ -0,0 +1,2 @@
+> Already built an extension? [Let us know about your experience](https://docs.docker.com/feedback/extension/).
+{: .important}
```
thaJeztah, line 2: (Not introduced in this PR) we should have a look at the link-colour for the "important" styles (blue on red looks a bit .. meh)

Screenshot 2023-01-09 at 15 01 47


Haven't tried what it'd look like, but wondering if it looks better if we split the first part to a separate line, so that it'd become a heading in the block;


```suggestion
> Already built an extension?
>
> [Let us know about your experience](https://docs.docker.com/feedback/extension/).
{: .important}
```

thaJeztah (approved):
LGTM, thanks! (probably best to "squash and merge" this one)

PR #16460, opened 2023-01-07
engine: update reference docs to use new anchor tags

thaJeztah:
- relates to / in preparation of https://github.com/docker/docs/pull/16416
- relates to https://github.com/docker/cli/pull/3924
- relates to / depends on https://github.com/docker/cli/pull/3525

This updates the engine reference documentation with the changes from https://github.com/docker/cli/pull/3525. We're using the new anchor tags for the upcoming 23.0 release, and this PR prepares changes related to that, so that updating the docs for v23.0.0 will be easier (all things already updated)..

Once these changes are in, we can also fix the buildx docs to point to the new anchor tags, and remove the htmlproofer exclusion.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -66,12 +66,11 @@ EOT
 # htmlproofer checks for broken links
 FROM gem AS htmlproofer-base
 # FIXME(thaJeztah): remove temporary exclusion rule for buildx_build once anchor links are updated.
-# FIXME(thaJeztah): remove temporary exclusion rule for commandline/cli once https://github.com/docker/cli/pull/3525 is merged.
 RUN --mount=type=bind,from=generate,source=/out,target=_site <<EOF
   htmlproofer ./_site \
     --disable-external \
     --internal-domains="docs.docker.com,docs-stage.docker.com,localhost:4000" \
-    --file-ignore="/^./_site/engine/api/.*$/,./_site/registry/configuration/index.html,./_site/engine/reference/commandline/buildx_build/index.html,./_site/engine/reference/commandline/cli/index.html" \
+    --file-ignore="/^./_site/engine/api/.*$/,./_site/registry/configuration/index.html,./_site/engine/reference/commandline/buildx_build/index.html" \
     --url-ignore="/^/docker-hub/api/latest/.*$/,/^/engine/api/v.+/#.*$/,/^/glossary/.*$/" > /results 2>&1
   rc=$?
   if [[ $rc -eq 0 ]]; then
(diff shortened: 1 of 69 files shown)


_config.yml
```diff
@@ -151,9 +151,10 @@ defaults:
 #   - dest is the destination path within the working tree
 #   - src is a list of glob source paths within the remote repository
 fetch-remote:
-  - repo: "https://github.com/docker/cli"
+  # FIXME(thaJeztah): temporarily using branch from https://github.com/docker/cli/pull/3525 to fix htmlproofer validation
+  - repo: "https://github.com/thaJeztah/cli"
     default_branch: "master"
-    ref: "20.10"
+    ref: "20.10_backport_anchor_tags"
```
thaJeztah, line 157: - This is temporary for this PR, and will be removed before merging (after https://github.com/docker/cli/pull/3525 is merged)

thaJeztah (commented):
the "fix anchor links" commit still needs to be squashed before merging as well

thaJeztah:
I did a quick test with only the second commit of this PR to check if htmlproofer  would fail, but looks good, so https://github.com/docker/cli/pull/3525 is probably good to go

thaJeztah:
This one should be ready for review now

@crazy-max @rumpl @dvdksn PTAL

crazy-max:
Just an observation: I Was looking at the Options table in https://deploy-preview-16460--docsdocker.netlify.app/engine/reference/commandline/build/ and I don't see any generated link on the name like `--ulimit` that should go to https://deploy-preview-16460--docsdocker.netlify.app/engine/reference/commandline/build/#-set-ulimits-in-container---ulimit.

thaJeztah:
@crazy-max you mean the TOC (right-hand side)? I guess this is because Jekyll doesn't understand the `<a name=...>` tags, or perhaps we need to update it to take the quotes into account? (If we did those ourselves) 🤔 let me check

thaJeztah:
@crazy-max It's probably this one that needs updating; https://github.com/docker/docs/blob/1f4f22c5c4859c50d64e9750f7b48c1a87b73c56/_includes/toc_pure_liquid.html#L1-L17

crazy-max:
> @crazy-max you mean the TOC (right-hand side)? I guess this is because Jekyll doesn't understand the `<a name=...>` tags, or perhaps we need to update it to take the quotes into account? (If we did those ourselves) 🤔 let me check

Not the TOC, the [Options](https://deploy-preview-16460--docsdocker.netlify.app/engine/reference/commandline/build/#options) table where `--ulimit` should be a link like https://github.com/docker/buildx/blob/master/docs/reference/buildx_build.md#options? Maybe this is not needed with the sidebar TOC actually.

thaJeztah:
Let me bring this one in 👍

PR #16467, opened 2023-01-09
ENGDOCS-1060a

aevesdocker:
PR aligns the 'Using service profiles' page in Compose with the Style guide, makes sure it is referring to the Compose spec


dvdksn (approved):
Really liking reference information section on these pages!

PR #16468, opened 2023-01-09
ENGDOCS-1062

aevesdocker:
PR aligns the 'Extend services' page in Compose with the Style guide, makes sure it is referring to the Compose spec

Addresses:
https://github.com/docker/docs/issues/16388

### DIFF
--- a/compose/extends.md
+++ b/compose/extends.md
@@ -6,14 +6,14 @@ title: Share Compose configurations between files and projects
 
 Compose supports two methods of sharing common configuration:
 
-1. Extending an entire Compose file by
+1. Extend an entire Compose file by
    [using multiple Compose files](extends.md#multiple-compose-files)
-2. Extending individual services with [the `extends` field](extends.md#extending-services) (for Compose file versions up to 2.1)
+2. Extend individual services with [the `extends` field](extends.md#extending-services) (for Compose file versions up to 2.1)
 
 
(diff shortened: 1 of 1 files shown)


compose/extends.md
```diff
@@ -6,14 +6,14 @@ title: Share Compose configurations between files and projects
 
 Compose supports two methods of sharing common configuration:
 
-1. Extending an entire Compose file by
+1. Extend an entire Compose file by
    [using multiple Compose files](extends.md#multiple-compose-files)
-2. Extending individual services with [the `extends` field](extends.md#extending-services) (for Compose file versions up to 2.1)
+2. Extend individual services with [the `extends` field](extends.md#extending-services) (for Compose file versions up to 2.1)
```
dvdksn, line 11: Maybe we could clarify that this option is no longer valid. Not sure I succeded but you catch my drift.

```suggestion
2. Extend individual services with [the `extends` field](extends.md#extending-services). This option is only usable in older Compose file versions, up to version 2.1.
```

aevesdocker: Oo actually this is wrong. It's available in the current spec as well. Have fixed

aevesdocker, line 11:
```suggestion
2. Extend individual services with [the `extends` field](extends.md#extending-services)
```

dvdksn (approved):
Nice!

Issue #16481, opened 2023-01-09
Issue when cloning to case-insensitive file system


mdavis-xyz:
# Steps to reproduce

On a Mac, with APFS (case-insensitive) file-system, clone the repo.

# Expected behavior

Normal clone

# Actual behavior

```
warning: the following paths have collided (e.g. case-sensitive paths
on a case-insensitive filesystem) and only one from the same
colliding group is in the working tree:

  'desktop/extensions-sdk/extensions/images/open-share.PNG'
  'desktop/extensions-sdk/extensions/images/open-share.png'
```

The clone is successful. So as long as I don't need those files I can do what I want. But what if I did want to touch those files?

# Problem location

  'desktop/extensions-sdk/extensions/images/open-share.PNG'
  'desktop/extensions-sdk/extensions/images/open-share.png'

# Suggestions for a fix

The two files have the same hash, and only the lower-case one is used by `../share.md`. So we should delete the uppercase one. I'll submit a PR for that.

aevesdocker:
Closing as PR for fix is now merged

PR #16482, opened 2023-01-09
remove redundant image with case-insensitive name-clash

mdavis-xyz:
### Proposed changes

The same image was present twice, once as `.png` and once as `.PNG`. (Same hash)

This results in issues when cloning to a case-insensitive file-system.

### Related issues (optional)

#16481


aevesdocker (approved):
thank you @mdavis-xyz

PR #16486, opened 2023-01-10
Update networkingcontainers.md

ronaldosc:
The syntax  '_--net=my_bridge_'  is not the same indicated when run Docker cli help.

So the correct is:
```
--network my_bridge
```


dvdksn:
hi, thank you for contributing to Docker docs. The `--net=<name>` syntax is valid, just like `--network <name>` (and `--network=<name>`) so no need to change this command,

PR #16498, opened 2023-01-11
doc: specify the initial gMSA example works only for windows containers

zombieleet:


zombieleet:
@dvdksn

dvdksn (approved):
LGTM, thanks!

PR #16465, opened 2023-01-09
DKP-539 Added release notes for Docker Desktop 4.16

mat007:
### Proposed changes

Added release notes for Docker Desktop 4.16

PR #16506, opened 2023-01-12
DD4.16 linux builds

aevesdocker:


mat007 (approved):
Oh good catch, that should have been part of #16465
I’ll update our release process so that we don’t forget it next time.

aevesdocker:
No problem, I didn't want to bother you with that since I had the time :)

dvdksn (approved):
LGTM!

PR #16508, opened 2023-01-12
Fix 4.16.0 release year

erkro1:


mat007 (approved):
Oh good catch!

aevesdocker (approved):
Thank you @erkro1

PR #16509, opened 2023-01-12
Update 05_persisting_data.md

KostiantynO:
### Proposed changes
  Added one slash '/' to exec command from `bash` in windows - to use absolute POSIX path.


craig-osterhout:
Thanks for sharing the PR. Closing this. This topic was modified recently and the issue was addressed in an alternate way.

PR #16520, opened 2023-01-13
Update generated reference docs with extension API 0.3.3 (Desktop 4.16)

gtardif:
Signed-off-by: Guillaume Tardif <guillaume.tardif@gmail.com>

### DIFF
--- a/desktop/extensions-sdk/dev/api/reference/interfaces/Docker.md
+++ b/desktop/extensions-sdk/dev/api/reference/interfaces/Docker.md
@@ -6,7 +6,9 @@ skip_read_time: true
 
 # Interface: Docker
 
-**`since`** 0.2.0
+**`Since`**
+
+0.2.0
 
 ## Properties
 
@@ -81,9 +83,9 @@ const containers = await ddClient.docker.listContainers();
 
 #### Parameters
 
-| Name       | Type  | Description                                                                                                                                                                                                                                                                                       |
-|:-----------|:------|:--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
-| `options?` | `any` | (Optional). A JSON like `{   "all": true,   "limit": 10,   "size": true,   "filters": JSON.stringify({ status: ["exited"] }), }`  For more information about the different properties see [the Docker API endpoint documentation](../../../../../../engine/api/v1.37.md#operation/ContainerList). |
+| Name | Type | Description |
+| :------ | :------ | :------ |
+| `options?` | `any` | (Optional). A JSON like `{ "all": true, "limit": 10, "size": true, "filters": JSON.stringify({ status: ["exited"] }), }` For more information about the different properties see [the Docker API endpoint documentation](https://docs.docker.com/engine/api/v1.41/#operation/ContainerList). |
 
 #### Returns
 
@@ -103,9 +105,9 @@ const images = await ddClient.docker.listImages();
 
 #### Parameters
 
-| Name       | Type  | Description                                                                                                                                                                                                                                                      |
-|:-----------|:------|:-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
-| `options?` | `any` | (Optional). A JSON like `{ "all": true, "filters": JSON.stringify({ dangling: ["true"] }), "digests": true }`  For more information about the different properties see [the Docker API endpoint documentation](../../../../../../engine/api/v1.37.md#tag/Image). |
+| Name | Type | Description |
+| :------ | :------ | :------ |
+| `options?` | `any` | (Optional). A JSON like `{ "all": true, "filters": JSON.stringify({ dangling: ["true"] }), "digests": true * }` * * For more information about the different properties see [the Docker API endpoint documentation](https://docs.docker.com/engine/api/v1.41/#tag/Image). |
 
 #### Returns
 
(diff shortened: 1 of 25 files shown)


desktop/extensions-sdk/dev/api/reference/interfaces/Extension.md
```diff
@@ -19,3 +21,13 @@ ___
...
+
+___
+
+### image
+
+• `Readonly` **image**: `string`
+
+**`Since`**
+
+0.3.3
```
gtardif, line 33: This is what's really new in 0.3.3. I don't know why the typedoc template slightly changed and impacted all files in various sections

desktop/extensions-sdk/dev/api/reference/interfaces/Docker.md
```diff
@@ -103,9 +105,9 @@ const images = await ddClient.docker.listImages();
...
-|:-----------|:------|:-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
-| `options?` | `any` | (Optional). A JSON like `{ "all": true, "filters": JSON.stringify({ dangling: ["true"] }), "digests": true }`  For more information about the different properties see [the Docker API endpoint documentation](../../../../../../engine/api/v1.37.md#tag/Image). |
+| Name | Type | Description |
+| :------ | :------ | :------ |
+| `options?` | `any` | (Optional). A JSON like `{ "all": true, "filters": JSON.stringify({ dangling: ["true"] }), "digests": true * }` * * For more information about the different properties see [the Docker API endpoint documentation](https://docs.docker.com/engine/api/v1.37/#tag/Image). |
```
thaJeztah, line 110: Is this content generated as well? We should probably (I guess we're stripping `docs.docker.com` before we generate the HTML, so probably still okay). If these are not generated, then we should probably make these relative links `/engine/api/<version>/`


I'm wondering why this is referencing API v1.37 - do you know if there's a reason for that? Is the SDK using this version?

v1.37 corresponds with Docker 18.05 (which is long obsolete); https://docs.docker.com/engine/api/#api-version-matrix

Current API version is v1.41 (soon: v1.42).

gtardif: Thx, this is generated from [typedoc comments in the API code](https://github.com/docker/pinata/pull/20250/commits/a68748cceba3996eb7e0cb189e8d5dfc0bd0d92a). I just updated the version number to 1.41. There's no "latest" link I could use to avoid a dependency on a pinned version? (I tried "latest" in the URL instead of the version, not supported :) ). 
The link in typedoc points to docs.docker.com, the generation replaces it with the relative path

thaJeztah: We do have a `https://docs.docker.com/engine/api/latest/` redirect (which I think is generated when we publish the docs), so that could be an option. Not sure if deep links work with that redirect though 🤔

gtardif: unfortunately no, it does not work for detail pages (Images for example)...

desktop/extensions-sdk/dev/api/reference/interfaces/DockerDesktopClient.md
```diff
@@ -85,13 +87,15 @@ You can use the option `{"all": true}` to list all the running and stopped conta
...
-|:----------|:--------|:-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
-| `options` | `never` | (Optional). A JSON like `{   "all": true,   "limit": 10,   "size": true,   "filters": JSON.stringify({ status: ["exited"] }), }` For more information about the different properties see [the Docker API endpoint documentation](../../../../../../engine/api/v1.37.md#operation/ContainerList). |
+| Name | Type | Description |
+| :------ | :------ | :------ |
+| `options` | `never` | (Optional). A JSON like `{ "all": true, "limit": 10, "size": true, "filters": JSON.stringify({ status: ["exited"] }), }` For more information about the different properties see [the Docker API endpoint documentation](https://docs.docker.com/engine/api/v1.37/#operation/ContainerList). |
```
thaJeztah, line 98: Looks like this one was missed for the API version

desktop/extensions-sdk/dev/api/reference/interfaces/DockerDesktopClient.md
```diff
@@ -115,13 +119,15 @@ Get the list of images
...
-|:----------|:--------|:-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
-| `options` | `never` | (Optional). A JSON like `{ "all": true, "filters": JSON.stringify({ dangling: ["true"] }), "digests": true }`  For more information about the different properties see [the Docker API endpoint documentation](../../../../../../engine/api/v1.37.md#tag/Image). |
+| Name | Type | Description |
+| :------ | :------ | :------ |
+| `options` | `never` | (Optional). A JSON like `{ "all": true, "filters": JSON.stringify({ dangling: ["true"] }), "digests": true }` For more information about the different properties see [the Docker API endpoint documentation](https://docs.docker.com/engine/api/v1.37/#tag/Image). |
```
thaJeztah, line 130: This one as well

thaJeztah (approved):
LGTM, but found two more v1.37 API links; not a blocker I guess, but let me know if you want to fix those before merging

desktop/extensions-sdk/dev/api/reference/interfaces/Docker.md
thaJeztah: Yeah, I doubted that would work. No problem, then we'll just need to keep an eye on updating if we have a new API version (or look if we can improve the redirects to preserve the anchors)

desktop/extensions-sdk/dev/api/reference/interfaces/DockerDesktopClient.md
gtardif: Thx, both fixed

thaJeztah (approved):
LGTM

PR #16525, opened 2023-01-13
Update naming of wordsmith sample app

mikesir87:
### Proposed changes

The wordsmith sample app repo has been renamed, so updating docs to reflect that (old URL will still redirect to the new).

### Related issues (optional)

None


craig-osterhout (approved):
LGTM. Thank you.

PR #16535, opened 2023-01-17
Show k8s & swarm links in similar way

githubERIK:
On `get-started/orchestration.md`, avoid showing k8s link as a button but swarm link as a regular link that is less visible.
Use an unordered list of two links instead.

(Contribution guidelines: https://docs.docker.com/contribute/overview)


craig-osterhout (approved):
Thank you. LGTM

PR #16536, opened 2023-01-17
Update ubuntu.md

arkel-s:
Repository install: adding permissions update on created `/etc/apt/keyrings/docker.gpg` file. Else, `apt-get update` is not working on Ubuntu 22.04


arkel-s:
I found the solution here: https://stackoverflow.com/questions/60137344/docker-how-to-solve-the-public-key-error-in-ubuntu-while-installing-docker

dvdksn:
Thanks for the contribution @arkel-s  - I'm closing this as I'm implementing this change, using a slightly different approach in #17070

PR #16568, opened 2023-01-20
Update index.md

matty-at-rdc:
The word *both* (exactly two) here when referring to a *set* (anything more than one) is less clear than it could be and all we have to do to eliminate the potentially confusing wording is use one less word!

### DIFF
--- a/compose/compose-file/index.md
+++ b/compose/compose-file/index.md
@@ -43,7 +43,7 @@ to support those running modes:
 
 ## The Compose application model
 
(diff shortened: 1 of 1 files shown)


ChrisChinchilla:
Thanks for the contribution @matty-at-rdc but it looks like since you made it, the state of this file has change completely, so the suggestions doesn't fit anymore. Sorry about that, I hope you make some future contributions :)

PR #16545, opened 2023-01-18
Prevent node_modules from override via mount

hahahannes:
### Proposed changes

In the section `Use Compose to develop locally`, a docker-compose file is shown which mounts the app directory into the `notes` container to be able to change the code without rebuilding the image. Unfortunately, the node_modules directory which was created during the build will be overridden as the whole working directory /app gets mounted. Therefore, this directory needs to be secured by adding it to the volumes.

### Related issues (optional)
Closes #15850

PR #16569, opened 2023-01-20
Update develop.md

Grasume:
Node docker-compose.yml no longer works line 142 is the culprit. 
Fix is to ensure the node_modules folder is correct.
It can now be found inside of /usr/src/app 

sample error is:
test-notes-1  | > image@1.0.0 debug
test-notes-1  | > nodemon --inspect=0.0.0.0:9229 server.js test-notes-1  | 
test-notes-1  | sh: 1: nodemon: not found
test-notes-1 exited with code 127

### DIFF
--- a/language/nodejs/develop.md
+++ b/language/nodejs/develop.md
@@ -139,7 +139,7 @@ services:
    - SERVER_PORT=8000
    - CONNECTIONSTRING=mongodb://mongo:27017/notes
   volumes:
-   - ./:/app
+   - ./:/usr/src/app
   command: npm run debug
 
  mongo:


craig-osterhout:
Thanks for sharing the issue.
When following the guide, I'm unable to reproduce this error.
Ensure that you ran `npm install nodemon` in the correct directory and that your compose file is in the correct directory.

In addition, the bind mount from the compose file will overwrite the node_modules directory in the container with your local folder. So, if nodemon is not in your local node_modules folder, you will also encounter errors. This shouldn't cause any issues when following the guide, but it's good to double check.
You can check the changes in this PR, https://github.com/docker/docs/pull/16545, which will prevent the container's node_modules from being overwritten.

If you need more help troubleshooting, please try the community Slack at https://join.slack.com/t/dockercommunity/shared_invite/zt-1jz09gjit-P2NPysg99kiWExFfof0oBA

Grasume:
@wking  How did you attempt to reproduce the error ?

I have just done the Tutorial on the page an can reproduce the error.
The issue is that 
volumes:
 ./:/app 
Fails for me
volumes:
./:/usr/src/app 
Works for me.

This is potentially related to PR #16545 but the issue / resolution is wrong from what I can tell. the path for the container to find the node_modules does not work with /app  and it has issues finding the packages.
If you do ./usr/src/app the deploy can find the node_modules files.

Grasume:
And I just also Further tested the Tutorial after changing the volume mount to the suggested.
I was successfully able to update the server.js file and see the changes live. Interesting enough I did not see logs that nodemon noticed the changes.

Grasume:
@craig-osterhout are you going to respond ever ?

zinfromsynergy:
the volumes suggestions fixed an issue of mine.

PR #16570, opened 2023-01-21
Add release note about minimum version on macos

mat007:
### Proposed changes

Adds a release note for Mac to say that Docker Desktop now requires macOS 11.

We had actually deprecated macOS 10.15 months ago (and updated the installation docs accordingly), but forgot to change the minimum version in the product.
Docker Desktop 4.16 now fails to start on macOS 10.15, even when previous versions seem to be fine.

### DIFF
--- a/desktop/release-notes.md
+++ b/desktop/release-notes.md
@@ -160,6 +160,7 @@ For frequently asked questions about Docker Desktop releases, see [FAQs](faqs/ge
 
 #### For Mac
 
+- Minimum version to install or update Docker Desktop on macOS is now 11 (Big Sur) or newer.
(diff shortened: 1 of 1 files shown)


desktop/release-notes.md
```diff
@@ -160,6 +160,7 @@ For frequently asked questions about Docker Desktop releases, see [FAQs](faqs/ge
 
 #### For Mac
 
+- Minimum version to install or update Docker Desktop on macOS is now 11 (Big Sur) or newer.
```
dvdksn, line 163:
```suggestion
- Minimum OS version to install or update Docker Desktop on macOS is now macOS Big Sur (version 11) or later.
```

mat007: Sure, thanks!

dvdksn (approved):

PR #16578, opened 2023-01-23
update registry example

aevesdocker:
PR fix as per https://docker.slack.com/archives/C02JGHEPZJ5/p1674235850054199?thread_ts=1674141615.446119&cid=C02JGHEPZJ5

### DIFF
--- a/registry/insecure.md
+++ b/registry/insecure.md
@@ -33,7 +33,7 @@ isolated testing or in a tightly controlled, air-gapped environment.
 
     ```json
     {
-      "insecure-registries" : ["myregistrydomain.com:5000"]
+      "insecure-registries" : ["http://myregistrydomain.com:5000"]
     }
     ```
 


dvdksn (approved):

croensch:
@aevesdocker i'm looking for the root cause of this change, could you please copy some details from this presumably internal slack here?

nitinpatil1992:
anyone who is facing the issue **http: server gave HTTP response to HTTPS client** [ref](https://github.com/moby/moby/blob/master/daemon/containerd/resolver.go), kindly look into [registry parsing](https://github.com/moby/moby/blob/master/daemon/daemon.go#L164-L200). There seems no update around it.

croensch:
Why is it at odds with https://github.com/moby/moby/blob/master/registry/config.go#L183-L245 though?

PR #16584, opened 2023-01-24
Fix typo and make other minor language corrections

peterhauke:
### Proposed changes

Remove comma from "Now, that" as it is not necessary in this particular case
Fix typo "pipleline" to "pipeline"
Remove comma after "extension" as it is not necessary here 
Replace "you" with "them" in one place
Other minor corrections


aevesdocker (approved):
Thank you @peterhauke

PR #16587, opened 2023-01-24
fix broken(anchor) links

rahulpandey70:

### DIFF
--- a/engine/index.md
+++ b/engine/index.md
@@ -74,9 +74,9 @@ For more details, see
     <div class="col-xs-12 col-sm-12 col-md-12 col-lg-4 block">
       <div class="component">
         <div class="component-icon">
-          <a href="/config/pruning/"><img src="/assets/images/engine-pruning.svg" alt="A pair of scissors" width="70px" height="70px"></a>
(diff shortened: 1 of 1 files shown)


dvdksn:
This change seems to break the links rather than fix them. Can you please describe what issue you've observed?

rahulpandey70:
When i click on prune link, it says page not found(The 'docker/docs' repository doesn't contain the 'config/pruning' path in 'main'.)

thaJeztah:
> When i click on prune link, it says page not found(The 'docker/docs' repository doesn't contain the 'config/pruning' path in 'main'.)

Were you viewing the page on GitHub or on docs.docker.com? If it's on GitHub, then that's probably expected; some of the pages linked to are downloaded while the docs are built and rendered, and cannot he viewed on GitHub; in some other cases we cannot link to the markdown (in this case, the link is a HTML link, which doesn't get converted by our static site builder)

rahulpandey70:
Okay i got it. Now I'm closing PR

thaJeztah:
Thanks! And sorry for the confusion. Perhaps we can fix some of this to make it work on GitHub as well (but we'd have to look into that)

rahulpandey70:
Thanks for the consideration @thaJeztah, If you want to look into that which broken link i'm talking about. Please go this path https://github.com/docker/docs/blob/main/engine/index.md

PR #16596, opened 2023-01-25
swarm: call out CA rotation as potentially dangerous with MKE

neersighted:
### Proposed changes

Add a call-out for Mirantis Kubernetes Engine/Docker UCP users re: Swarm CA rotation, which is not supported in the general case and usually dangerous/indicative of other cluster health problems. A PR for the CLI docs will follow once this wording is reviewed.

### DIFF
--- a/engine/swarm/how-swarm-mode-works/pki.md
+++ b/engine/swarm/how-swarm-mode-works/pki.md
@@ -60,6 +60,13 @@ reference for details.
 
 ## Rotating the CA certificate
 
+> **Note**
+>
+> Mirantis Kubernetes Engine (MKE), formerly known as Docker UCP, provides an external
+> certificate manager service for the swarm. If you run swarm on MKE, you shouldn't
+> rotate the CA certificates manually. Instead, contact Mirantis support if you need
+> to rotate a certificate.
+
 In the event that a cluster CA key or a manager node is compromised, you can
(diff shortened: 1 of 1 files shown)


engine/swarm/how-swarm-mode-works/pki.md
```diff
@@ -60,6 +60,13 @@ reference for details.
 
 ## Rotating the CA certificate
 
+> **Note**
+>
+> Mirantis Kubernetes Engine (formerly known as Docker UCP) provides an external
+> CA component to service the swarm. As the lifecycles of certificates in a MKE
+> cluster are controlled differently, users of MKE should not rotate the CA manually.
+> Users should contact Mirantis support if they need to rotate a certificate.
+
```
dvdksn, line 69: Is this a fair rewording?

```suggestion
> **Note**
>
> Mirantis Kubernetes Engine (MKE), formerly known as Docker UCP, provides an external
> certificate manager service for the swarm. If you run swarm on MKE, you shouldn't
> rotate the CA certificates manually. Instead, contact Mirantis support if you need
> to rotate a certificate.

```

neersighted: Sounds better to be!

dvdksn (approved):
LGTM

PR #16598, opened 2023-01-26
docs: extend Golang's testing tutorial with Testcontainers for Go

### LINKED ISSUE #16221: [Golang docs feedback] Extend the existing tutorial using more testing libraries
### Problem description
I think this is not a problem but instead a suggestion.

For the sake of covering more available options, we would like to contribute a section that demonstrates how to test with [Testcontainers for Go](https://github.com/testcontainers/testcontainers-go) (the additional benefit is that it is available in multiple languages too, so we can bring even more content). We believe this will be beneficial for the community, and the way we learned about the article is from someone who started with [ory](https://github.com/ory/dockertest)'s version after reading the page but then migrated to Testcontainers as they found it more convenient for their use case, so we want to make sure that the page reflects all available options for devs to choose from.

### Problem location

- I suggest a change at the following URL: https://docs.docker.com/language/golang/run-tests/

- Other: As a Go developer, I'd like to see different options on testing Docker containers.

### Suggestions for a fix

I would be able to contribute a tutorial using other testing libraries, but I do not know how that would fit into the current page structure, so I'd appreciate your guidance in defining it.

mdelapenya:
### Proposed changes

We'd like to consider using two competing approaches to manage containerised tests: [dockertest](https://github.com/ory/dockertest) and [Testcontainers for Go](https://github.com/testcontainers/testcontainers-go), so readers are able to choose which approach is more convenient for their use case.

To achieve it:
1. we are adding an acknowledge block at the top of the page explaining why we have two different alternatives to write the Go tests. 
1. we are not modifying the existing dockertest section, keeping it as the first section (deserved as it landed in the docs first)
1. we are adding an H2 heading that will enclose the existing sections for dockertest. Therefore, those headings are pushed from H2 to H3.
1. we are cloning the dockertest section (texts and codeblocks) but adapting them to Testcontainers for Go. We do not want to create confusion and use the same language and words than the existing texts.
1. we are just updating the sentences that are specific to Testcontainers for Go, such as the usage of wait strategies or the usage of Ryuk as resource cleaner.
1. we are adding Testcontainers and Ryuk to the Vale vocabulary, to avoid errors in the Vale checks.
1. we are updating how the dockertest tests are run adding Go build tags. This is needed because of the changes in https://github.com/olliefr/docker-gs-ping/pull/7/files#diff-79ce3229c13921b79b1175dcb211336aaf84dfd8edcf19c07698934d4fe70e5eR1, that separates the execution of both test types.

One thing that we would like to notice is the current execution time for dockertest: it says around [6 seconds](https://github.com/docker/docs/blame/c68b34e6055c0ae3c2a489341120421f42fba1c0/language/golang/run-tests.md?plain=1#L94), but we have run the very same tests locally and the times are way low. We'd like to update those numbers to https://github.com/olliefr/docker-gs-ping/pull/7/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R28, which seems fair (and more realistic). If this PR is accepted, we'd like to updated them accordingly (here in this PR or in a follow-up one).

### Related issues (optional)

- Depends on https://github.com/olliefr/docker-gs-ping/pull/7 (cc/ @olliefr)
- Closes #16221

### DIFF
--- a/.github/vale/Vocab/Industry/accept.txt
+++ b/.github/vale/Vocab/Industry/accept.txt
@@ -43,6 +43,7 @@ Python
 QEMU
 RHEL
 Raspbian
+Ryuk
 S3
 SBOMs?
 SLES
@@ -55,6 +56,7 @@ Solr
 SonarQube
 Syft
 TLS
+Testcontainers
 Twilio
 Ubuntu
 WSL
--- a/language/golang/run-tests.md
+++ b/language/golang/run-tests.md
@@ -10,7 +10,13 @@ redirect_from:
 
 Testing is an essential part of modern software development. Yet, testing can mean a lot of things to different development teams. In the name of brevity, we'll only take a look at running isolated, high-level, functional tests.
 
-## Test structure
+> **Acknowledgment**
+>
+> In this section we are going to consider using two competing approaches to manage containerised tests: [dockertest](https://github.com/ory/dockertest) and [Testcontainers for Go](https://github.com/testcontainers/testcontainers-go). We let readers choose which approach is more convenient for their use case.
+
+## Using dockertest
+
+### Test structure
 
 Each test is meant to verify a single business requirement for our example application. The following test is an excerpt from `main_test.go` test suite in our example application.
 
@@ -59,7 +65,7 @@ As you can see, this is a high-level test, unconcerned with the implementation d
 
 The second test in `main_test.go` has almost identical structure but it tests _another_ business requirement of our application. You are welcome to have a look at all available tests in [`docker-gs-ping/main_test.go`](https://github.com/olliefr/docker-gs-ping/blob/main/main_test.go).
 
-## Run tests locally
+### Run tests locally
 
 In order to run the tests, we must make sure that our application Docker image is up-to-date.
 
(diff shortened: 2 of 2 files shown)


mdelapenya:
Hi @StefanScherer sorry for pinging you directly but I'm not sure what is the regular frequency to review this kind of PRs. I saw you participated in the review of https://github.com/docker/docs/pull/16244, which is quite related to this one, that's why I summon you. Please forgive me if you are not in charge of this 🙏 

Do you know if there anything else on my side to do here? I'll be more than happy to help

mdelapenya:
Hi @craig-osterhout sorry for pinging you directly, but I noticed you as a very active maintainer here. 🙏 

Do you know what can we do to unblock this PR? Is there anything on our side that we can do for it?

StefanScherer:
Thanks for your contribution. I don't think we need two test frameworks for one language in our docs. I understand your good intention, but in my opinion this exceeds what our quick tutorials should cover.

mdelapenya:
@StefanScherer thanks for your honest feedback. It seems we were not fast enough! :D In any case, we consider that the existing dockertest tutorial deserves to be there and our intention was never to replace it but extend the tutorial.

What would be the best way to provide make these resources discoverable to the Docker community? We have identified a few resources, although we are not sure which are the most up-to-date and the best place to contribute:

- Self-paced tutorials: https://docs.docker.com/get-started/resources/#self-guided-tutorials
- Samples: https://docs.docker.com/samples/go/
- Play with Docker: https://github.com/play-with-docker/play-with-docker.github.io

Thanks in advance for your assistance.

PR #16608, opened 2023-01-27
Update Extensions SDK API ref with `viewContainerTerminal`

felipecruz91:
### Proposed changes

Update Extensions SDK API ref with `viewContainerTerminal`.

### DIFF
--- a/desktop/extensions-sdk/dev/api/reference/interfaces/NavigationIntents.md
+++ b/desktop/extensions-sdk/dev/api/reference/interfaces/NavigationIntents.md
@@ -100,6 +100,30 @@ A promise that fails if the container doesn't exist.
 
 ___
 
+### viewContainerTerminal
+
+▸ **viewContainerTerminal**(`id`): `Promise`<`void`\>
+
+Navigate to the container terminal window in Docker Desktop.
+
+```typescript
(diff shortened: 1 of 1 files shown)


aevesdocker (dismissed):
🎉

aevesdocker:
@felipecruz91 is this to be published with DD 4.17?

felipecruz91:
@aevesdocker Yes, let's coordinate it with @gtardif

lucbpz (commented):
Can we also add a line with this in the `dashboard-routes-navigation` example?

gtardif (commented):
Nit: we should add a `@since 0.3.4` in the code comment for this specific method, to let the users know in which version this has been introduced.

felipecruz91:
> 

@lucbpz Looking at the bottom of the `dashboard-routes-navigation.md` file, the methods listed are deprecated and will be removed in a future version. Those are the ones from v0 that start with `navigateTo` and have the `window` property (e.g. `window.ddClient.navigateToContainerStats(id)`)

The `viewContainerTerminal(id)` is a new method and as such, it hasn't been added to the v0 API but to v1. Does it make sense?

gtardif (approved):

aevesdocker (approved):

PR #16610, opened 2023-01-27
Update gettingstarted.md

siamak-niroomandi:
Changed FLASK_DEBUG: True --> FLASK_DEBUG: "true"

### Proposed changes

Based on the compose file reference I proposed this change.

> Tip when using boolean values
>
> YAML boolean values (`"true"`, `"false"`, `"yes"`, `"no"`, `"on"`,
> `"off"`) must be enclosed in quotes, so that the parser interprets them as
> strings.

Source: https://docs.docker.com/compose/compose-file/compose-file-v3/

### DIFF
--- a/compose/gettingstarted.md
+++ b/compose/gettingstarted.md
@@ -228,7 +228,7 @@ services:
     volumes:
       - .:/code
     environment:
-      FLASK_DEBUG: True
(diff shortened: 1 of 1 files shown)


aevesdocker (approved):
Thank you @siamak-niroomandi!

PR #16618, opened 2023-01-27
docs(bind-mount): added section to help user use 'jq' to better find …

darklight147:
### Proposed changes

Added a small section to show user how to use `jq` to extract information from the json output of docker inspect in the `bind-mount` docs

### DIFF
--- a/storage/bind-mounts.md
+++ b/storage/bind-mounts.md
@@ -269,6 +269,33 @@ correctly. Look for the `Mounts` section:
 ],
 ```
 
+Alternatively, if you have [`jq`](https://stedolan.github.io/jq/download/) installed, you can use the following command to display the mount information in a more readable format, and help you find the Mounts block faster (note that the `jq` command is not part of the Docker CLI, it is a separate tool):
+
+```console
+$ docker inspect devtest | jq '.[0].Mounts'
+```
+
+The output should look similar to the following:
+
+```json
+[
+  {
+    "Type": "bind",
+    "Source": "/tmp/source/target",
+    "Destination": "/app",
+    "Mode": "ro",
+    "RW": false,
+    "Propagation": "rprivate"
+  }
+]
+```
+
+If the previous instructions failed to mount the directory, the output of the previous command will be empty:
+
+```json
+[]
+```
+
 Stop the container:
 
 ```console


storage/bind-mounts.md
```diff
@@ -269,6 +269,33 @@ correctly. Look for the `Mounts` section:
 ],
 ```
 
+Alternatively, if you have [`jq`](https://stedolan.github.io/jq/download/) installed, you can use the following command to display the mount information in a more readable format, and help you find the Mounts block faster (note that the `jq` command is not part of the Docker CLI, it is a separate tool):
```
thaJeztah, line 272: I'm slightly on the fence if we'd be going too much into the weeds here. I guess we could still use `--format` to extract just the mounts (although it defaults to not using pretty-print);

```bash
docker inspect --format '{{json .Mounts}}' devtest
[{"Type":"bind","Source":"/Users/thajeztah/Projects/docker.github.io/target","Destination":"/app","Mode":"ro","RW":false,"Propagation":"rprivate"}]
```

storage/bind-mounts.md
```diff
@@ -269,6 +269,33 @@ correctly. Look for the `Mounts` section:
...
+    "Source": "/tmp/source/target",
+    "Destination": "/app",
+    "Mode": "ro",
+    "RW": false,
+    "Propagation": "rprivate"
+  }
+]
+```
+
+If the previous instructions failed to mount the directory, the output of the previous command will be empty:
```
thaJeztah, line 293: Not sure if this is the case; if the shorthand (`-v <host-path>:<container-path>`) format is used, the host-path would be created. For the `--mount` syntax, the container would fail.

storage/bind-mounts.md
dvdksn: +1 - I don't think we should bring up `jq` the example and instead favor the built-in format flag.

question: is there a way to select nodes, dot-notation style, using the shorthand `--format=json` ?

thaJeztah: No, currently not (`json` is just "everything").

I think this was the reason we originally went for "look for the `Mounts` section in the output", as we wanted to avoid going too deep into those details. But I can see the issue that `inspect` output is a _lot_, and finding the right part requires a bit of scrolling 😅 

reminds me I should write up that proposal to introduce commands for printing the information in a friendlier format (https://github.com/docker/cli/issues/522#issuecomment-333658421)

thaJeztah: Slightly off-topic; I noticed the examples;

- use `nginx:latest` - we could use `nginx:alpine` (bit smaller/lightweight and also runs as a daemon)
- I noticed we use `-d -it` for those containers; `-d -it` can be a "trick" to keep a (e.g.) busybox  running, but as the nginx image already runs a daemon, we could drop the `-it` (should not be needed)

dvdksn:
Closing as we will not add a reference to `jq` in these docs.

PR #16619, opened 2023-01-28
Update Python example to use Python 3 syntax

Briscoooe:
### Proposed changes

Changed the Python SDK code snippet to use Python 3 syntax. `print` statements without parentheses are only valid in Python 2, which [reached EOL 3 years ago](https://www.python.org/doc/sunset-python-2/)


dvdksn (approved):
Thank you!

PR #16650, opened 2023-02-01
Config: config.md contains a misleading command, that will only give a result if called in the node running the container #16649

zombieleet:
### Proposed changes

The command for getting the ID of a container in swarm mode is 50% wrong and 50% correct, the initial content assumes that the container was created in the node (master) that initiated the `docker service create command.

In a one-node swarm cluster or a service created with `--mode=global`, the command `docker ps --filter name=redis -q` is fine, but when the nodes in the cluster are `N + 1`, the possibility of this command `docker ps --filter name=redis -q` producing the correct result is not certain. 

The update made here, instructs the user to get the node the container is running on, proceed to the node and run the rest of the commands.

### DIFF
--- a/engine/swarm/configs.md
+++ b/engine/swarm/configs.md
@@ -172,7 +172,7 @@ real-world example, continue to
     completion to do this automatically.
 
     ```console
-    $ docker service ps --format "{{.Node}}" redis
+    $ docker service ps --format '{{.Node}}' redis
     
     ip-172-31-46-109
     
--- a/engine/swarm/configs.md
+++ b/engine/swarm/configs.md
@@ -166,11 +166,18 @@ real-world example, continue to
 4.  Get the ID of the `redis` service task container using `docker ps`, so that
     you can use `docker container exec` to connect to the container and read the contents
     of the config data file, which defaults to being readable by all and has the
-    same name as the name of the config. The first command below illustrates
-    how to find the container ID, and the second and third commands use shell
+    same name as the name of the config. The first command retrieves the node ID
+    the container is running on, from the node  `ip-172-31-46-109` run the second command below illustrates
+    how to find the container ID, and the third and fourth commands use shell
     completion to do this automatically.
 
     ```console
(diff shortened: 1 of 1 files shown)


engine/swarm/configs.md
```diff
@@ -166,11 +166,18 @@ real-world example, continue to
...
     of the config data file, which defaults to being readable by all and has the
-    same name as the name of the config. The first command below illustrates
-    how to find the container ID, and the second and third commands use shell
+    same name as the name of the config. The first command retrieves the node ID
+    the container is running on, from the node  `ip-172-31-46-109` run the second command below illustrates
+    how to find the container ID, and the third and fourth commands use shell
     completion to do this automatically.
 
     ```console
+    $ docker service ps --format '{{ "{{" }} .Node }}' redis
```
dvdksn, line 175: is this format value correct? trying to parse it but my brain can't handle

cc @thaJeztah

dvdksn: OH, I see now. This needs the special `{% raw %}` and `{% endraw %}` tags before and after the code block, to escape the double curly braces.

thaJeztah: So, the example was written with a "test" environment (single node) in mind. Finding the right node where the task landed up on would be useful for multi-node swarm clusters.

That said, "from now on command should be run on that node" may also not be correct, as `docker service` commands can only be run from a manager node, so if the task landed on a worker node, those commands won't work.

Given that this step was only to "verify that the config was applied", I'm wondering if we could come up with a better way to show that (without requiring direct access to the node). e.g. could be a "web" service (nginx container, e.g.) that shows the config, or a service that logs the contents of the config (which can be viewed using `docker service logs`).

zombieleet: Thanks for the comments 💯 

What I meant with "from now on command should be run on that node" is that the rest of the commands in `Number 4` aside from the first command (the comment came after the first command) should be executed in the worker node.

Issue #16362, opened 2022-12-18
Additional permission fix required on Debian 11


philippm:
[engine/install/debian.md](https://docs.docker.com/engine/install/debian/)

The document offers a section on solving GPG errors with `apt-get update`:

```bash
sudo chmod a+r /etc/apt/keyrings/docker.gpg
sudo apt-get update
```

That didn't quite fix it for me, I had to grant read/list permissions to the keyrings folder:

```bash
sudo chmod a+rx /etc/apt/keyrings
```

Perhaps it would be good to add this to the docs?

PR #16654, opened 2023-02-02
Update debian.md

### LINKED ISSUE #16626: Documentation on adding download.docker.com/linux/debian as debian repo not working
[engine/install/debian.md](https://docs.docker.com/engine/install/debian/)

Hi,

I followed the instructions how to add download.docker.com/linux/debian as debian repo but I cannot get it to work.

The error is when running apt update:
```
Get:3 https://download.docker.com/linux/debian bullseye InRelease [43.3 kB]
Err:3 https://download.docker.com/linux/debian bullseye InRelease
  The following signatures couldn't be verified because the public key is not available: NO_PUBKEY 7EA0A9C3F273FCD8
```

I did setup the repo as below:
```
$ sudo mkdir -p /etc/apt/keyrings
$ curl -fsSL https://download.docker.com/linux/debian/gpg | sudo gpg --dearmor -o /etc/apt/keyrings/docker.gpg
```

I checked that the key was there and it has the correct format. 

Then I did:
```
$ echo \
  "deb [arch=$(dpkg --print-architecture) signed-by=/etc/apt/keyrings/docker.gpg] https://download.docker.com/linux/debian \
  $(lsb_release -cs) stable" | sudo tee /etc/apt/sources.list.d/docker.list > /dev/null
```

This results in the following file:
```
$ cat /etc/apt/sources.list.d/docker.list
deb [arch=amd64 signed-by=/etc/apt/keyrings/docker.gpg] https://download.docker.com/linux/debian   bullseye stable
```

I also tried adding read perms to key but that did not change the error:
`$ sudo chmod a+r /etc/apt/keyrings/docker.gpg`

I also tried to import the particular key id manually but then I got:
```
$ gpg --keyring /etc/apt/trusted.gpg --recv-keys 7EA0A9C3F273FCD8 
gpg: key 8D81803C0EBFCD88: "Docker Release (CE deb) <docker@docker.com>" not changed
gpg: Total number processed: 1
gpg:              unchanged: 1
```

ArnoSen:
Set correct filemask for `/etc/apt/keyrings` . If this is not 0755, the keys in the directory cannot be used.

### DIFF
--- a/engine/install/debian.md
+++ b/engine/install/debian.md
@@ -96,7 +96,7 @@ Docker from the repository.
 2.  Add Docker's official GPG key:
 
     ```console
-    $ sudo mkdir -m 0755 -p /etc/apt/keyrings 
+    $ sudo mkdir -m 0755 -p /etc/apt/keyrings
     $ curl -fsSL {{ download-url-base }}/gpg | sudo gpg --dearmor -o /etc/apt/keyrings/docker.gpg
     ```
 
--- a/engine/install/debian.md
+++ b/engine/install/debian.md
@@ -96,7 +96,7 @@ Docker from the repository.
 2.  Add Docker's official GPG key:
 
     ```console
-    $ sudo mkdir -p /etc/apt/keyrings
+    $ sudo mkdir -m 0755 -p /etc/apt/keyrings 
     $ curl -fsSL {{ download-url-base }}/gpg | sudo gpg --dearmor -o /etc/apt/keyrings/docker.gpg
     ```
 


ArnoSen:
Closes #16626
Closes #16362

dvdksn:
@vvoland could you ptal? thanks

vvoland (approved):
LGTM; permissions applied by mkdir are dependent on the `umask` value, which in some rare cases may be different from 0022.

This is also conformant with Debian suggestions:
> In releases older than Debian 12 and Ubuntu 22.04, /etc/apt/keyrings does not exist by default. It SHOULD be created with permissions 0755 if it is needed and does not already exist.
(source: https://wiki.debian.org/DebianRepository/UseThirdParty)

engine/install/debian.md
```diff
@@ -96,7 +96,7 @@ Docker from the repository.
 2.  Add Docker's official GPG key:
 
     ```console
-    $ sudo mkdir -p /etc/apt/keyrings
+    $ sudo mkdir -m 0755 -p /etc/apt/keyrings
```
thaJeztah, line 99: Looks like there's a trailing space that sneaked its way in

```suggestion
    $ sudo mkdir -m 0755 -p /etc/apt/keyrings
```

thaJeztah (approved):
LGTM, but left a suggestion; we can probably merge that suggestion and squash the commits 👍

thaJeztah (approved):
LGTM; I'll squash and merge 👍

PR #16659, opened 2023-02-03
engine: 23.0.0 errata

neersighted:
- Closes https://github.com/moby/moby/issues/44912

This is a rollup of various corrections to the 23.0.0 release notes I have prepared over the last 12 hours:

- engine: amend 23.0 release notes for typos/minor factual errors
- engine: amend 23.0 release notes with known Kata issue
- engine: amend 23.0 release notes with overlooked change to the example systemd units
- engine: amend 23.0 release notes with tracking issues for known issues 

/cc @corhere to look at the Kata wording, @dvdksn @thaJeztah for review.

### DIFF
--- a/engine/release-notes/23.0.md
+++ b/engine/release-notes/23.0.md
@@ -78,6 +78,7 @@ For a full list of pull requests and changes in this release, refer to the relev
 - API: Add a `CreateMountPoint` parameter to `POST /containers/create`. [moby/moby#43484](https://github.com/moby/moby/pull/43484)
 - API: Add a `shared-size` parameter to `GET /images/json` to enable shared-size computation of images. [moby/moby#42531](https://github.com/moby/moby/pull/42531)
 - API: Add a `type` parameter to `GET /system/df`, to control what object types to are considered when computing disk usage. [moby/moby#42559](https://github.com/moby/moby/pull/42559)
+- systemd: Use a systemd-managed containerd instead of a daemon-managed containerd. [moby/moby#42373](https://github.com/moby/moby/pull/42373)
 - systemd: Start `docker.service` after `time-set.target`. [moby/moby#43107](https://github.com/moby/moby/pull/43107)
 
 ### Removed
--- a/engine/release-notes/23.0.md
+++ b/engine/release-notes/23.0.md
@@ -45,11 +45,11 @@ For a full list of pull requests and changes in this release, refer to the relev
 - Add support for alternate OCI runtimes on Linux, compatible with the containerd runtime v2 API. [moby/moby#43887](https://github.com/moby/moby/pull/43887), [moby/moby#43993](https://github.com/moby/moby/pull/43993)
 - Add support for the containerd `runhcs` shim on Windows (off by default). [moby/moby#42089](https://github.com/moby/moby/pull/42089)
 - Add `dockerd --validate` to check the daemon JSON config and exit. [moby/moby#42393](https://github.com/moby/moby/pull/42393)
-- Add the ability to configure the daemon's HTTP proxy via flags or JSON config [moby/moby#42835](https://github.com/moby/moby/pull/42835)
+- Add the ability to configure the daemon's HTTP proxy via flags or JSON config. [moby/moby#42835](https://github.com/moby/moby/pull/42835)
 - Add support for RFC 3021 point-to-point networks (IPv4 /31s) and single hosts (IPv4 /32s). For networks with two or fewer addresses, IPAM won't reserve a network and broadcast address. [moby/moby#42626](https://github.com/moby/moby/pull/42626)
 - Add support for setting `ipvlan_flag` and using the `l3s` `ipvlan_mode` in the `ipvlan` network driver. [moby/moby#42542](https://github.com/moby/moby/pull/42542)
 - Add support for displaying the value of the `metacopy` option for the `overlay2` storage driver. [moby/moby#43557](https://github.com/moby/moby/pull/43557)
-- Add support for describing Windows devices using the syntax `IDType://ID`.[moby/moby#43368](https://github.com/moby/moby/pull/43368)
+- Add support for describing Windows devices using the syntax `IDType://ID`. [moby/moby#43368](https://github.com/moby/moby/pull/43368)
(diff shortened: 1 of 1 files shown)


paralin:
@neersighted The updated note mentions that `--containerd` now implies an externally managed containerd, which is good. Please also note that `--containerd` was added to the systemd unit under contrib/ and is therefore default-enabled with systemd.

engine/release-notes/23.0.md
```diff
@@ -42,6 +42,9 @@ For a full list of pull requests and changes in this release, refer to the relev
   - There are differences in how BuildKit and the legacy builder handle multi-stage builds. For more information, see
     [Multi-stage builds](../../build/building/multi-stage.md#differences-between-legacy-builder-and-buildkit).
 - Add support for pulling `zstd` compressed layers. [moby/moby#41759](https://github.com/moby/moby/pull/41759), [moby/moby#42862](https://github.com/moby/moby/pull/42862)
+- Change the `--containerd` option to signify an externally managed containerd at a socket path. [moby/moby#42373](https://github.com/moby/moby/pull/42373)
```
thaJeztah, line 45: This PR (https://github.com/moby/moby/pull/42373) only move the systemd unit back to the moby repo; docker 20.10 already uses that configuration

paralin: The main difference is that --containerd was added to the systemd unit

engine/release-notes/23.0.md
```diff
@@ -81,6 +81,7 @@ For a full list of pull requests and changes in this release, refer to the relev
 - API: Add a `CreateMountPoint` parameter to `POST /containers/create`. [moby/moby#43484](https://github.com/moby/moby/pull/43484)
 - API: Add a `shared-size` parameter to `GET /images/json` to enable shared-size computation of images. [moby/moby#42531](https://github.com/moby/moby/pull/42531)
 - API: Add a `type` parameter to `GET /system/df`, to control what object types to are considered when computing disk usage. [moby/moby#42559](https://github.com/moby/moby/pull/42559)
+- systemd: Use a systemd-managed containerd instead of a daemon-managed containerd. [moby/moby#42373](https://github.com/moby/moby/pull/42373)
```
thaJeztah, line 84: I guess, same for this (as we used a separate containerd service since .. (long ago)

engine/release-notes/23.0.md
thaJeztah: Slightly longer explanation; https://github.com/moby/moby/issues/44912#issuecomment-1415374072

> I don't think this changed in the 23.0 release; the packaged versions of `docker-ce` already used this configuration since docker 18.09;
> 
> * [[18.09 backport] systemd: set --containerd socket patch to prevent race-condition docker/docker-ce-packaging#298](https://github.com/docker/docker-ce-packaging/pull/298)
> 
> The systemd unit in this repository were no longer used at that time (but were left in the repository in case someone would find them useful). [009639e](https://github.com/moby/moby/commit/009639e0f46d270309821998836fb664046b3e07) was part of this PR:
> 
> * [Upstream systemd unit files from docker-ce-packaging repo #42373](https://github.com/moby/moby/pull/42373)
> 
> Which brought back the systemd service into this repository, but effectively no change in the packages (other than the packaging scripts now consuming the systemd unit(s) from this repository instead of from the packaging repository).
> 
> Note that _without_ the `--containerd` option, the daemon would already _prefer_ a running containerd service (if it finds a containerd socket in the default location), but this could sometimes cause a race condition if the containerd _service_ started, but the socket not yet created, which could lead to _two_ instances of containerd being started. The `--containerd` flag was added to prevent this race condition.

paralin: OK, I see, the systemd unit in this repo was outdated. We were using the one from contrib/ in Buildroot.

engine/release-notes/23.0.md
neersighted: I think this is correct, as it is now impossible for the daemon to start a managed containerd when running under systemd.

neersighted:
Looks like I should have done some more fact checking, instead of taking things at face value -- my bad there. I think this should be accurate now.

engine/release-notes/23.0.md
```diff
@@ -218,17 +219,27 @@ For a full list of pull requests and changes in this release, refer to the relev
...
+apt-get install apparmor
 ```
+
+#### Kata Containers ([tracking issue](https://github.com/kata-containers/kata-containers/issues/6154))
+
+The 23.0 branch brings support for alternate containerd shims, such as `io.containerd.runsc.v1` (gVisor) and `io.containerd.kata.v2` (Kata Containers).
+
+Due to a long-standing bug in Moby, exiting an `exec` session when using the Kata Containers runtime will terminate the container, hanging the connected CLI if a TTY was opened. There is no mitigation at this time beyond avoiding execing into containers running on the Kata runtime.
+
+This will be resolved in a future release of Moby; users should keep in mind that alternate OCI runtimes are a new feature and that similar issues may be discovered as more users start exercising this functionality.
```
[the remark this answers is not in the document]
dvdksn: do we have a link to reference the Moby bug in question? in case users want to follow up.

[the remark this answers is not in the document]
neersighted: It's linked as 'tracking issue' in the section header -- the bug is canonically tracked on Kata and not Moby, as it was originally thought to be a Kata bug. It was revealed to be a Moby bug, and while we have several PRs on Moby, they reference the Kata bug and not a new Moby bug.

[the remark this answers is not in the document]
neersighted: Pushed up a new version with a rephrase (not quite yours, as the altered wording introduced some errors) -- PTAL!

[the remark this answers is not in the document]
dvdksn: geee, how did I just forget about the tracking link in the title 🙈 
LGTM now!

dvdksn (approved):

neersighted:
Just bumping this; I think getting some corrections (particularly around apparmor) out in a timely fashion is important as 23.0.0 rolls out more widely to users.

thaJeztah (approved):
LGTM, thanks!

PR #16662, opened 2023-02-04
fix(engine): Minimum --kernel-memory is 6m

rebornplusplus:
### Proposed changes

``linuxMinMemory`` was updated from 4m to 6m [0], and it is 
used to compare ``resources.KernelMemory`` [1, 2]. Thus, the 
actual minimum allowed value for --kernel-memory is 6m.

Refs:
- [0] https://github.com/docker/engine/commit/d2e23405bed2f5388b4a6c4f50c94ed2ab580319
- [1] https://github.com/docker/engine/blob/master/daemon/daemon_unix.go#L473
- [2] https://github.com/moby/moby/blob/master/daemon/daemon_unix.go#L452


ChrisChinchilla:
Thanks @rebornplusplus

ChrisChinchilla (approved):

PR #16681, opened 2023-02-07
engine/install/ubuntu: update instructions to match debian variant

thaJeztah:
Incorporate the changes made in the debian variant from;

- 6d8b08171c734ff66a277183c6b8fd4a72fb1fab (https://github.com/docker/docs/pull/16654)
- 37ae6721930ee3d96a6235b2d94920518ce14383 (https://github.com/docker/docs/pull/15902)

### DIFF
--- a/engine/install/ubuntu.md
+++ b/engine/install/ubuntu.md
@@ -95,7 +95,7 @@ Docker from the repository.
 2.  Add Docker's official GPG key:
 
     ```console
-    $ sudo mkdir -p /etc/apt/keyrings
+    $ sudo mkdir -m 0755 -p /etc/apt/keyrings
     $ curl -fsSL {{ download-url-base }}/gpg | sudo gpg --dearmor -o /etc/apt/keyrings/docker.gpg
     ```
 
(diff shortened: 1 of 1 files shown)


engine/install/ubuntu.md
```diff
@@ -232,6 +232,7 @@ download a new file each time you want to upgrade Docker Engine.
    `hello-world` image:
 
    ```console
+   $ sudo service docker start
```
thaJeztah, line 235: Kept it the same as the `debian` variant, but wondering if we should change this to `sudo systemctl start docker` (as most of our docs assume systemd is used, and `systemctl` is likely the canonical command for that).

dvdksn: should be safe to use systemctl, I think. Can't recall receiving any requests to update this, unlike apt/apt-get. When we get around to defining linux/unix/shell conventions, we could consider including a preference to either command there

dvdksn (approved):

engine/install/ubuntu.md
thaJeztah: Yup; let's stick to this for now to keep them the same.

I saw some other things we could update (e.g. the examples for versions listed by `apt-cache madison`, and some other bits and bops).

thaJeztah:
Let me bring this one in 👍

PR #16682, opened 2023-02-07
Remove broken link in vm-vdi

MihaelaStoica:
### Proposed changes

The page referenced here doesn't exist anymore, so remove the link.


aevesdocker (approved):
Thanks @MihaelaStoica

PR #16698, opened 2023-02-09
fix typo in env-file argument

JensErat:
### Proposed changes

This is a rather obvious typo.

```
$ docker compose --help | grep env-file
      --env-file string            Specify an alternate environment file.
```

### Related issues (optional)

none


aevesdocker (approved):
Thanks for catching this @JensErat

PR #16710, opened 2023-02-10
Update run-tests.md

nbarnwell:
The tests literally didn't work for me. I'm not sure what breaking changes appear to have come in since they were first written, but a major problem was that calling `.WithNetwork()` passing in a not-yet-started network would throw an exception. Most of the changes are around coping with that - moving code from ctor to the `InitializeAsync()` method.

### DIFF
--- a/language/dotnet/run-tests.md
+++ b/language/dotnet/run-tests.md
@@ -38,31 +38,44 @@ using System.Net;
 using DotNet.Testcontainers.Builders;
 using DotNet.Testcontainers.Containers;
 using DotNet.Testcontainers.Networks;
+using Xunit;
+using Xunit.Abstractions;
 
 public sealed class UnitTest1 : IAsyncLifetime, IDisposable
 {
     private const ushort HttpPort = 80;
 
-    private readonly CancellationTokenSource _cts = new(TimeSpan.FromMinutes(1));
+    private CancellationTokenSource _cts = new(TimeSpan.FromMinutes(1));
 
-    private readonly IDockerNetwork _network;
+    private IDockerNetwork _network;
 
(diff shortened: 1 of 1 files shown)


craig-osterhout:
Thank you, @nbarnwell. Your changes fix the exception for me.

Hi @HofmeisterAn, PTAL if you can.

HofmeisterAn:
@nbarnwell can you please share the exception. I will double-check it, but I do not see an issue in the previous code right away.

HofmeisterAn:
I believe the appropriate solution is to set the version to 2.3.0 by using the command `dotnet add package Testcontainers --version 2.3.0`. By doing this, we can avoid running the documentation against an incompatible version of the software. Additionally, we are in the process of implementing module support, which will make the documentation and usage much more straightforward in the future. As part of my responsibilities, I will be updating the Docker documentation once the new version is released.

It should only necessary to update this line:

https://github.com/docker/docs/blob/8275c6f4a6885c82f8f32160e42ddc9dd2127550/language/dotnet/run-tests.md?plain=1#L29

Should I create a PR?

nbarnwell:
I mean, fair enough. But you are opening yourself up to having to revisit this page one day because folks will come here looking for how to use the latest version, only to find the docs stuck at v2.3. I guess choose your pain. :)

HofmeisterAn:
With the next release, the API will be stable. Unfortunately, despite taking great care to avoid introducing breaking changes, I was not completely successful.

> But you are opening yourself up to having to revisit this page one day because folks will come here looking for how to use the latest version, only to find the docs stuck at v2.3.

It is in my own interest to ensure that the Docker documentation is up-to-date. With new TC releases, I am happy to update the documentation as well.

PR #16711, opened 2023-02-10
Update https://docs.docker.com/compose/networking/

mlnorthernswe:
In row 55 to 57 it is written following (document https://docs.docker.com/compose/networking/):
Within the `web` container, your connection string to `db` would look like `postgres://db:5432`, and from the host machine, the connection string would look like `postgres://{DOCKER_IP}:8001`.
As the container is running postgres on port 5432 it should be `postgres://{DOCKER_IP}:5432` in the end of that sentence.

### DIFF
--- a/compose/networking.md
+++ b/compose/networking.md
@@ -54,7 +54,7 @@ the service is accessible outside the swarm as well.
 
 Within the `web` container, your connection string to `db` would look like
 `postgres://db:5432`, and from the host machine, the connection string would
-look like `postgres://{DOCKER_IP}:8001`.
+look like `postgres://{DOCKER_IP}:5432`.
 
 ## Update containers on the network
 


aevesdocker (approved):
Thanks for catching this @mlnorthernswe

ftoh:
Connections from host should be `postgres://{DOCKER_IP}:8001` because in `docker-compose.yml` exposes another port
```yml
  db:
    image: postgres
    ports:
      - "8001:5432"
 //  -----^
```

mlnorthernswe:
If {DOCJE_IP} is the container address it's :5432 because that it's the port it's listen on in the container. But if {DOCKER_IP} is the host address it's :8001 because it's forarded 5432 in the container. So it depends what they mean with {DOCKER_IP}. thought it was the container address.

ftoh:
Thank you for explanation. Then it is better to specify both options.
```
Within the web container, your connection string to db would look like `postgres://db:5432`, and from the host machine, the connection string would look like `postgres://{DOCKER_IP}:5432` or `postgres://localhost:8001`
```

mlnorthernswe:
> Thank you for explanation. Then it is better to specify both options.
> 
> > Within the web container, your connection string to db would look like `postgres://db:5432`, and from the host machine, the connection string would look like `postgres://{DOCKER_IP}:5432` or `postgres://localhost:8001`

Yes, I have suggested that to better illustrate the port forwarding thing, but got the following answer:

Hi @mlnorthernswe, Thanks for your PR. I'm going to close it because in the rare case a user is using a remote engine and their stack is running on a remote server, postgres://localhost:8001 won't work, which is why we have postgres://{DOCKER_IP}:5432 in the docs.

mlnorthernswe:
But if it's the rare case I think you could use postgres://{DOCKER_HOST_IP}:8001. That should work for a remote user.

PR #16719, opened 2023-02-13
fix(registry): correct IAM policy for s3 storage driver

bericp1:
### Proposed changes

The s3 storage driver explicitly uses the `CreateMultipartUpload` s3 API call and this is needed in the suggested IAM policy in the documentation, otherwise the registry will fail to upload blobs with the generic error "AccessDenied".

https://github.com/distribution/distribution/blob/main/registry/storage/driver/s3-aws/s3.go#L651

### Related issues (optional)

None.


bericp1:
Apologies, this isn't actually an s3 action, `PutObject` covers this. I misread some documentation and had a typo elsewhere in my policy.

PR #16729, opened 2023-02-16
Illustrate use of HOST_PORT better

mlnorthernswe:
To illustrate the mapping between HOST_PORT and CONTAINER_PORT better I think the last meaning starting in row 55 should be changed as my suggestions. postgres://localhost:8001 maps to postgres://{DOCKER_IP}:5432.


aevesdocker:
Hi @mlnorthernswe, Thanks for your PR. I'm going to close it because in the rare case a user is using a remote engine and their stack is running on a remote server, `postgres://localhost:8001` won't work, which is why we have `postgres://{DOCKER_IP}:5432` in the docs.

PR #16733, opened 2023-02-16
Put docker-compose into the cli-plugins folder

kojoru:
### Proposed changes

Docker compose v2 is a plugin, and it needs to be installed in a `cli-plugins` directory. Only this way will the next step `docker compose version` work and the note about _also_ being able to put it just to `Program Files/Docker` makes sense.

The amount of time it took me to figure it out is maddening, so I'm putting this fix in the docs for the future generation. Mostly the documentation recommends just installing docker desktop which is not supported on Windows Server.

### Related issues (optional)

N/A

### DIFF
--- a/compose/install/other.md
+++ b/compose/install/other.md
@@ -46,7 +46,7 @@ When asked if you want to allow this app to make changes to your device, click *
 3. Run the following command to download the latest release of Compose ({{site.compose_version}}):
 
     ```powershell
-     Start-BitsTransfer -Source "https://github.com/docker/compose/releases/download/{{site.compose_version}}/docker-compose-Windows-x86_64.exe" -Destination $Env:ProgramFiles\Docker\docker-compose.exe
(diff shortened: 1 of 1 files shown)


kojoru:
I've looked at the docs more thoroughly and realized that maybe a bigger reorganization is needed. This document is about "installing docker standalone" but called `other.md`. The other one is about "installing docker compose plugin" and called `linux.md`. It also indeed only talks about linux.

I could do this bigger reorganization for you to properly split between plugin and standalone (Windows and linux), but would need your prior approval that you're interested.

aevesdocker:
Hi @kojoru thank you for the PR. I'm going to close it as the suggestion you made is not necessary for users as the page refers to the Compose standalone, rather than the plugin. As you can also see from our docs, the easiest and recommended way to install Compose V2 is by installing Docker Desktop.

kojoru:
@aevesdocker Docker Desktop is not available on Windows Server. The documentation never anywhere mentioned how to install a compose plugin there. If there's a different place where this info should rather go can you point me to it?

kojoru:
Like I mentioned earlier I'm happy to do a broader reorganization of the documentation so that it makes more sense. I just need your permission and willingless to work with me.

PR #16739, opened 2023-02-17
Fix grammer in `ipam` description

janpieper:
### Proposed changes

Fix grammer in `ipam` description.


usha-mandya (approved):
Thank you @janpieper :)

PR #16745, opened 2023-02-18
Pass StopOptions to ContainerStop function

### LINKED ISSUE #16744: Example "Stop all running containers" has invalid syntax
[engine/api/sdk/examples.md](https://github.com/docker/docs/blob/main/engine/api/sdk/examples.md?plain=1#L352)

In the example **"Stop all running containers"**, the third parameter of function **ContainerStop(ctx, container.ID, nil)** should not be **nil**.
When trying to run the code snippet with `go run main.go` I get this error:
```bash
./main.go:26:50: cannot use nil as container.StopOptions value in argument to cli.ContainerStop
```

andregri:
Fix #16744

**options** argument in **ContainerStop** [function](https://github.com/moby/moby/blob/master/daemon/stop.go#L23) can't be **nil**

### DIFF
--- a/engine/api/sdk/examples.md
+++ b/engine/api/sdk/examples.md
@@ -331,6 +331,7 @@ import (
 	"fmt"
 
 	"github.com/docker/docker/api/types"
+	containertypes "github.com/docker/docker/api/types/container"
 	"github.com/docker/docker/client"
 )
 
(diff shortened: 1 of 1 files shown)


dvdksn (approved):
TY!

PR #16764, opened 2023-02-22
docs: clarify provenance attestation defaults

sozercan:
### Proposed changes

Clarify provenance defaults
`BUILDX_NO_DEFAULT_ATTESTATIONS` is not cherry picked yet. https://github.com/docker/buildx/pull/1612


build/attestations/index.md
```diff
@@ -49,6 +49,16 @@ You can opt in to add either the SBOM or provenance attestation type, or both.
...
 
+> SBOM generation is not enabled by default. It must be opted in using `--sbom=true` option.
+>
+> However, provenance generation is enabled by default with `mode=min,inline-only=true` option.
+>
+> It can be disabled by setting `--provenance=false` or using `BUILDX_NO_DEFAULT_ATTESTATIONS` environment variable.
+>
+> If explicitly enabled using `--provenance=true`, provenance attestation is created with `mode=max` option by default.
+>
+> See [Provenance attestation](./slsa-provenance.md) section for more details.
```
dvdksn, line 60: Thanks for this contribution!
Just some formatting suggestions.

```suggestion
> **Note**
>
> Provenance attestations are enabled by default, with the `mode=min` option.
> You can disable provenance attestations using the `--provenance=false` flag,
> or by setting the `BUILDX_NO_DEFAULT_ATTESTATIONS` environment variable:
>
> ```console
> $ export BUILDX_NO_DEFAULT_ATTESTATION=1
> ```
>
> Using the `--provenance=true` flag attaches provenance attestations with `mode=max`
> by default. See [Provenance attestation](./slsa-provenance.md) for more details.
```

crazy-max: @dvdksn I don't think we have a common place for `BUILDX_` vars atm i nthe docs right? Might be worth to have smth like https://docs.docker.com/engine/reference/builder/#buildkit-built-in-build-args in our docs too?

dvdksn (dismissed):

crazy-max:
Before merging we should wait for next Buildx 0.10.4 release to avoid confusion.

crazy-max:
Hum now that we have a new page for env vars with build, I wonder if we should just add a link to it: https://docs.docker.com/build/building/env-vars/#buildx_no_default_attestations

@dvdksn WDYT?

sozercan:
@crazy-max @dvdksn updated to link to env vars page, lmk if this works

dvdksn (approved):
LGTM thanks @crazy-max !

PR #16765, opened 2023-02-23
build(gha): updates

crazy-max:
### Proposed changes

* Fixes an issue with our named context example with GitHub Actions that doesn't work with a container builder. This provides an additional example for this case.
* Follow-up of https://github.com/docker/setup-buildx-action/pull/189 to add version pinning when configuring a builder.
* Set correct experimental layout for gha cache exporter

Also while browsing the examples I found out the page was quite huge and user could lost track and also not that great for discoverability (SEO, search).

### Related issues (optional)

* https://github.com/docker/build-push-action/issues/807


dvdksn (approved):
LGTM!

PR #16781, opened 2023-02-27
fix: Pin Testcontainers for .NET version

HofmeisterAn:
### Proposed changes

This pull request pins Testcontainers for .NET to version 2.3.0. Although we could also use the latest version (2.4.0) with some small adjustments, I recommend to stick with version 2.3.0 for now. We are currently preparing a major release that will make it easier for Docker customers to test .NET applications and services, and we plan to update the Docker docs accordingly.

This pull request fixes an issue where Testcontainers expected an already created network, which is not necessary at that point.

### Related issues (optional)

Refer to related PRs or issues:

- Closes #16710


craig-osterhout (approved):
Thank you. LGTM.

PR #16783, opened 2023-02-27
Replace broken image

ChrisChinchilla:
### Proposed changes

It doesn't look like this file exists, or has existed for a while, so re-created it.


aevesdocker (approved):
Thanks @ChrisChinchilla!

PR #16824, opened 2023-03-02
Update initial usage example

jakelevirne:
After running the initial usage example, it wasn't clear how to know whether the command ran successfully. This adds an instruction to view the output of the running Wasm module.

### Proposed changes

Added a step to the first usage example so the reader knows how they can verify that the command ran successfully.

### DIFF
--- a/desktop/wasm/index.md
+++ b/desktop/wasm/index.md
@@ -47,7 +47,7 @@ $ docker run -dp 8080:8080 \
   michaelirwin244/wasm-example
 ```
 
-After running this command, you should be able to visit http://localhost:8080/ and see the "Hello world" output from this example module.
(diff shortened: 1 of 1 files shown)


desktop/wasm/index.md
```diff
@@ -47,6 +47,8 @@ $ docker run -dp 8080:8080 \
   michaelirwin244/wasm-example
 ```
 
+After running this command, you should be able to visit http://localhost:8080/ and see the "Hello world" output from this example module.
```
aevesdocker, line 50:
```suggestion
After running this command, you can visit [http://localhost:8080/](http://localhost:8080/) to see the "Hello world" output from this example module.
```

aevesdocker (approved):
Thank you @jakelevirne!

PR #16847, opened 2023-03-08
Update intext.md - one word typo in composer's aliases section

pawelWritesCode:
Fix typo in aliases section

### Proposed changes

While reading documentation I found typo in `aliases` section. According to [networking in compose](https://docs.docker.com/compose/networking/) services within one network can reach each other by their corresponding service name declared in `docker-compose.yaml` file. Additionally, whole example content does not contain word `db`.


aevesdocker:
Thanks for the PR @pawelWritesCode. This content is now pulled in from an upstream repo. Would you please open a PR in https://github.com/compose-spec/compose-spec?

PR #16878, opened 2023-03-13
Review for freshness and to check reported errors

ChrisChinchilla:
### Proposed changes

Reviewed Python build image docs for freshness and as an error was reported.

### DIFF
--- a/language/python/build-images.md
+++ b/language/python/build-images.md
@@ -15,7 +15,7 @@ description: Learn how to build your first Docker image by writing a Dockerfile
 
 ## Overview
 
-Now that we have a good overview of containers and the Docker platform, let’s take a look at building our first image. An image includes everything needed to run an application - the code or binary, runtime, dependencies, and any other file system objects required.
+Now that you have a good overview of containers and the Docker platform, let’s take a look at building your first image. An image includes everything needed to run an application - the code or binary, runtime, dependencies, and any other file system objects required.
 
 To complete this tutorial, you need the following:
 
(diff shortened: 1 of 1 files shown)


ChrisChinchilla:
I will go through other pages in this section next for some consistency, but don't want to create PRs with too many changes, unless people would rather that?

language/python/build-images.md
```diff
@@ -49,68 +51,69 @@ def hello_world():
...
 
-Once we have our `requirements.txt` file inside the image, we can use the `RUN` command to execute the command `pip3 install`. This works exactly the same as if we were running `pip3 install` locally on our machine, but this time the modules are installed into the image.
+With the `requirements.txt` file inside the image, you can use the `RUN` command to run `pip3 install`. This works exactly the same as running `pip3 install` locally on a machine, but this time pip installs the modules into the image.
 
 ```dockerfile
 RUN pip3 install -r requirements.txt
 ```
 
-At this point, we have an image that is based on Python version 3.8 and we have installed our dependencies. The next step is to add our source code into the image. We’ll use the `COPY` command just like we did with our `requirements.txt` file above.
+At this point, you an image based on Python version 3.8 and have installed the dependencies. The next step is to add the source code into the image. Use the `COPY` command as with the `requirements.txt` file. This `COPY` command takes all the files located in the current directory and copies them into the image.
```
craig-osterhout, line 110:
```suggestion
At this point, you have an image based on Python version 3.8 and have installed the dependencies. The next step is to add the source code into the image. Use the `COPY` command as with the `requirements.txt` file. This `COPY` command takes all the files located in the current directory and copies them into the image.
```

craig-osterhout (changes requested):

PR #16883, opened 2023-03-13
ci: fix validate upstream workflow

crazy-max:
### Proposed changes

Encounter this issue while adding the upstream workflow to the buildkit repo: https://github.com/moby/buildkit/actions/runs/4406503730/jobs/7718839413#step:9:3815

That's because we are not checking `inputs.data-files-folder` and therefore it will remove the `./_data` folder.


usha-mandya (approved):
LGTM. Thank you @crazy-max!

PR #16902, opened 2023-03-15
Update https://docs.docker.com/compose/networking/

ftoh:
### Proposed changes

Updated information on how to reach the container from the host. This will help beginners not get confused in the networks.

Another proposal by @mlnorthernswe
> ... look like `postgres://{DOCKER_IP}:5432` or `postgres://{DOCKER_HOST_IP}:8001`

### Related issues (optional)

Discussion: [#16711](https://github.com/docker/docs/pull/16711#issuecomment-1470258105)
Similar prs:
https://github.com/docker/docs/pull/16729
https://github.com/docker/docs/pull/16894

### DIFF
--- a/compose/networking.md
+++ b/compose/networking.md
@@ -54,7 +54,7 @@ the service is accessible outside the swarm as well.
 
 Within the `web` container, your connection string to `db` would look like
 `postgres://db:5432`, and from the host machine, the connection string would
-look like `postgres://{DOCKER_IP}:5432`.
+look like `postgres://{DOCKER_IP}:5432` or `postgres://localhost:8001`.
 
 ## Update containers on the network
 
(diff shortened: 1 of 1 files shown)


compose/networking.md
```diff
@@ -54,7 +54,7 @@ the service is accessible outside the swarm as well.
 
 Within the `web` container, your connection string to `db` would look like
 `postgres://db:5432`, and from the host machine, the connection string would
-look like `postgres://{DOCKER_IP}:5432`.
+look like `postgres://{DOCKER_IP}:5432` or `postgres://localhost:8001`.
```
glours, line 57:
```suggestion
look like `postgres://{DOCKER_IP}:8001` for example `postgres://localhost:8001` if your container is running locally.
```

glours (commented):
So, in this case, the doc was wrong, the port to connect to the container from outside of Compose network is `8001`. 
And I agree the `{DOCKER_IP}` part can be confusing be we must be very clear that using `localhost` works only when your container is running locally

aevesdocker (approved):
Thank you for raising the PR @ftoh and thank you for the review @glours 🎉

PR #16908, opened 2023-03-17
build: change sections order in toc

crazy-max:
### Proposed changes

As discussed internally with @dvdksn and @jedevc, change sections order in toc and overview page.

Before:

image

After:

image

Maybe we could also add an extra card for attestations in the Overview page. WDYT?

Let me know if that looks good.


dvdksn (approved):
LGTM thanks! I think adding a card for attestations makes sense. I'll create an issue for a followup

PR #16911, opened 2023-03-17
build: dockerfile release notes

crazy-max:
closes https://github.com/docker/docs/pull/15110

### Proposed changes

While waiting for https://github.com/docker/docs/pull/15110#discussion_r1115939206, let's add the Dockerfile release notes that is currently missing in the Build section. Also moves frontend syntax to a dedicated Dockerfile section:

image

### DIFF
--- a/_data/toc.yaml
+++ b/_data/toc.yaml
@@ -1685,12 +1685,14 @@ manuals:
           title: SLSA definitions
         - path: /build/attestations/attestation-storage/
           title: Attestation storage
+    - sectiontitle: Dockerfile
+      section:
+        - path: /build/dockerfile/frontend/
+          title: Custom Dockerfile syntax
     - sectiontitle: BuildKit
       section:
         - path: /build/buildkit/
           title: Overview
-        - path: /build/buildkit/dockerfile-frontend/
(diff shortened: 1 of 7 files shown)


dvdksn:
how about changing `/build/dockerfile/frontend.md` to `/build/dockerfile/index.md` ?

crazy-max:
> how about changing `/build/dockerfile/frontend.md` to `/build/dockerfile/index.md` ?

I'm not sure. While moving the frontend page I was thinking if this was legit to make it the entry page of this section.

Maybe instead we could have a new `index.md` page with a short intro about the Dockerfile and cross linking to reference for full ref?

We could also probably rework and move https://docs.docker.com/develop/develop-images/dockerfile_best-practices/ in this section?

WDYT?

dvdksn:
I completely lost this (blaming parental leave)

I think we should merge this and yes, we need to check the best practices page, I think moving some of that to this section will be clearer. Let's do it in a followup

Looks like this needs a rebase.

crazy-max:
@dvdksn Should be good

dvdksn (approved):
:shipit:

PR #16936, opened 2023-03-22
ENGDOCS-1269

aevesdocker:


compose/environment-variables/envvars.md
```diff
@@ -43,8 +43,7 @@ the container's name on startup.
 For example, if your project name is `myapp` and it includes two services `db` and `web`, 
 then Compose starts containers named `myapp-db-1` and `myapp-web-1` respectively.
 
-Compose sets the project name using the following mechanisms, in order of
-precedence:
+Compose can set the project name in different ways. The level of precedence (from highest to lowest) for each method is as follows:
```
usha-mandya, line 46: Thanks Allie. I like the clarification "from highest to lowest". I'm wondering if it helps to convert this to an ordered list to emphasize the order? Also, as Compose environment variables is one of the popular pages, I was wondering if it helps to add an example here? Something on the lines of "If you specify the '-p' flag, it overrides the default setting...".

aevesdocker: Thanks for the suggestions! Yes agree with the ordered list - have changed this. Not sure adding an example helps though. It's a fairly straightforward concept and I think it would interfere with the flow of the rest of the section.

PR #16938, opened 2023-03-22
Improve uninstall instructions

singhprd:
### Proposed changes

Update uninstall instructions.

The icon is now a bug and doesn't have any text, so tell the user that so they can find it

### DIFF
--- a/desktop/uninstall.md
+++ b/desktop/uninstall.md
@@ -34,7 +34,7 @@ To uninstall Docker Desktop from your Windows machine:
 <br>
 To uninstall Docker Desktop from your Mac:
 
-1. From the Docker menu, select **Troubleshoot** and then select **Uninstall**.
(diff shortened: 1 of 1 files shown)


desktop/uninstall.md
```diff
@@ -34,7 +34,7 @@ To uninstall Docker Desktop from your Windows machine:
 <br>
 To uninstall Docker Desktop from your Mac:
 
-1. From the Docker menu, select **Troubleshoot** and then select **Uninstall**.
+1. From the Docker menu, select **Troubleshoot** (the bug icon) and then select **Uninstall**.
```
aevesdocker, line 37:
```suggestion
1. From the Docker menu, select the **Troubleshoot** icon in the top-right corner of Docker Dashboard and then select **Uninstall**.
```

PR #16947, opened 2023-03-23
engine: No lsb-release for setting up the repo

benz0li:
### Proposed changes

Do not use `lsb-release` to set up the repository. `lsb-release` depends on `python3`.

`docker-ce`, `docker-ce-cli`, `containerd.io`, `docker-buildx-plugin` and `docker-compose-plugin` do not depend on `python3`.

### Related issues (optional)

\-

### DIFF
--- a/engine/install/debian.md
+++ b/engine/install/debian.md
@@ -89,8 +89,7 @@ Docker from the repository.
     $ sudo apt-get install \
         ca-certificates \
         curl \
-        gnupg \
-        lsb-release
+        gnupg
     ```
 
 2.  Add Docker's official GPG key:
@@ -103,9 +102,9 @@ Docker from the repository.
 3.  Use the following command to set up the repository:
 
     ```console
-    $ echo \
+    $ . /etc/os-release && echo \
       "deb [arch=$(dpkg --print-architecture) signed-by=/etc/apt/keyrings/docker.gpg] {{ download-url-base }} \
-      $(lsb_release -cs) stable" | sudo tee /etc/apt/sources.list.d/docker.list > /dev/null
+      $VERSION_CODENAME stable" | sudo tee /etc/apt/sources.list.d/docker.list > /dev/null
     ```
 
 #### Install Docker Engine
--- a/engine/install/ubuntu.md
+++ b/engine/install/ubuntu.md
@@ -88,8 +88,7 @@ Docker from the repository.
     $ sudo apt-get install \
         ca-certificates \
         curl \
-        gnupg \
-        lsb-release
+        gnupg
     ```
 
 2.  Add Docker's official GPG key:
@@ -102,9 +101,9 @@ Docker from the repository.
 3.  Use the following command to set up the repository:
 
     ```console
-    $ echo \
+    $ . /etc/os-release && echo \
       "deb [arch=$(dpkg --print-architecture) signed-by=/etc/apt/keyrings/docker.gpg] {{ download-url-base }} \
-      $(lsb_release -cs) stable" | sudo tee /etc/apt/sources.list.d/docker.list > /dev/null
+      $VERSION_CODENAME stable" | sudo tee /etc/apt/sources.list.d/docker.list > /dev/null
     ```
 
 #### Install Docker Engine


benz0li:
According to his GitHub profile, @dvdksn is on *parental leave*.

@usha-mandya @ChrisChinchilla Could someone else review, please?

usha-mandya:
@thaJeztah Could you PTAL? Thanks

thaJeztah:
Thanks for your PR! I was looking earlier, and want to double-check if there was a specific reason for using `lsb-release`. I know there's been quite some back-and-forth on some of these, and some of these _may_ have been to work around specific corner-cases that users could run into.

For referene, I had a look at the convenience install script (at get.docker.com), which also uses `/etc/os-release`; https://github.com/docker/docker-install/blob/66474034547a96caa0a25be56051ff8b726a1b28/install.sh#L163-L172

```bash
get_distribution() {
	lsb_dist=""
	# Every system that we officially support has /etc/os-release
	if [ -r /etc/os-release ]; then
		lsb_dist="$(. /etc/os-release && echo "$ID")"
	fi
	# Returning an empty string here should be alright since the
	# case statements don't act unless you provide an actual value
	echo "$lsb_dist"
}
```

_However_ that script uses it for the "happy path", which may not always be the case, and because of that has some additional handling; https://github.com/docker/docker-install/blob/66474034547a96caa0a25be56051ff8b726a1b28/install.sh#L305-L307

```bash
# perform some very rudimentary platform detection
lsb_dist=$( get_distribution )
lsb_dist="$(echo "$lsb_dist" | tr '[:upper:]' '[:lower:]')"
```

And fallbacks; https://github.com/docker/docker-install/blob/66474034547a96caa0a25be56051ff8b726a1b28/install.sh#L356-L362

```bash
		*)
			if command_exists lsb_release; then
				dist_version="$(lsb_release --release | cut -f2)"
			fi
			if [ -z "$dist_version" ] && [ -r /etc/os-release ]; then
				dist_version="$(. /etc/os-release && echo "$VERSION_ID")"
			fi
		;;
```

thaJeztah:
Was there a specific issue you ran into because of Python being installed as dependency?

From the above, I'm a bit on the fence what the best thing to do is here; I can see `/etc/os-release` being a more lightweight approach, but the downsides could be that "it may not work", which could lead to more support questions from users, and with the installation steps already quite involved, I do want to avoid adding more complexity (if we can avoid).

Of course, users implementing these steps for deployment scripts could adjust (and optimise) the script for their situation, and we _could_ add a section about optimisations that users could make (but perhaps that should be left as an exercise for the reader).


@tianon @neersighted you have any opinions / thoughts?

benz0li:
> Was there a specific issue you ran into because of Python being installed as dependency?

No.

> From the above, I'm a bit on the fence what the best thing to do is here; I can see `/etc/os-release` being a more lightweight approach, but the downsides could be that "it may not work", which could lead to more support questions from users, and with the installation steps already quite involved, I do want to avoid adding more complexity (if we can avoid).

Both Debian and Ubuntu have `/etc/os-release`.

> Of course, users implementing these steps for deployment scripts could adjust (and optimise) the script for their situation

That's what I did. Just wanted to share this *optimisation*.

neersighted:
I don't see anything wrong with depending on `/etc/os-release`; that is what Mirantis's install script does, which we use across our entire support matrix, and is technically a [FreeDesktop.org standard](https://www.freedesktop.org/software/systemd/man/os-release.html).

benz0li:
> Was there a specific issue you ran into because of Python being installed as dependency?

P.S.: I am using a slightly more customised version here: https://github.com/b-data/jupyterlab-r-docker-stack/blob/10839ba8f300cf58a9b21f04da8cb3c80cbab727/common/subtags/docker/Dockerfile#L11-L26

```Dockerfile
RUN dpkgArch="$(dpkg --print-architecture)" \
  && . /etc/os-release \
  && mkdir -m 0755 -p /etc/apt/keyrings \
  && curl -fsSL https://download.docker.com/linux/$ID/gpg | \
    gpg --dearmor -o /etc/apt/keyrings/docker.gpg \
  && echo "deb [arch=$dpkgArch signed-by=/etc/apt/keyrings/docker.gpg] https://download.docker.com/linux/$ID $VERSION_CODENAME stable" | \
    tee /etc/apt/sources.list.d/docker.list > /dev/null \
  && apt-get update \
  && apt-get -y install \
    docker-ce-cli \
    docker-buildx-plugin \
    docker-compose-plugin \
  && ln -s /usr/libexec/docker/cli-plugins/docker-compose \
    /usr/local/bin/docker-compose \
  ## Clean up
  && rm -rf /var/lib/apt/lists/*
```

👉 This code can be used with both Debian and Ubuntu images.

The parent image I am using already has a Python source build installed at `/usr/local/bin` and I do not want another Python version in that image.

benz0li:
> As to the implementation in the PR, I was slightly wondering if it would be better to inline the code (`"$(. /etc/os-release && echo "$VERSION_CODENAME")"`); perhaps @tianon has opinions on that

Seems reasonable as only `$VERSION_CODENAME` is required.

thaJeztah:
> Seems reasonable as only `$VERSION_CODENAME` is required.

My thinking there was to prevent users overlooking the `.` at the start, or "trying to be smart" and only taking the `echo ..` (and `$VERSION_CODENAME` not being set as a result of that). There's a wide variety of users, and not all of them actually _read_ the docs properly, and now all of them are profound users of command-lines, so every bit that can help avoiding them make mistakes helps 😄

neersighted:
The advantage to containing it in `$()` is that it wouldn't 'escape' and populate variables in the rest of the script. I think that is a good approach here, especially as some of the names are very generic and may collide with variables someone might reasonably use elsewhere (e.g. `ID`).

neersighted (approved):

tianon (approved):

PR #16954, opened 2023-03-23
remote-development: make sure cross-domain links with ambassador have the same params

nicks:
### Proposed changes

make sure cross-domain links with ambassador have the same params,
so ambassador can see where they came from on the other end


nicks:
updated to match the params @aevesdocker added in this pr - https://github.com/docker/docs/pull/16929

PR #16957, opened 2023-03-24
Fix link to healthcheck section

rizerzero:


aevesdocker (approved):
thanks for catching this @rizerzero!

PR #16972, opened 2023-03-24
logging: non-blocking does not enqueue new messages when buffer is full

PettitWesley:
Fixes https://github.com/moby/moby/issues/45200

### DIFF
--- a/config/containers/logging/configure.md
+++ b/config/containers/logging/configure.md
@@ -135,8 +135,7 @@ Docker provides two modes for delivering messages from the container to the log
 driver:
 
 * (default) direct, blocking delivery from container to driver
-* non-blocking delivery that stores log messages in an intermediate per-container
-  ring buffer for consumption by driver
+* non-blocking delivery that stores log messages in an intermediate per-container buffer for consumption by driver
 
 The `non-blocking` message delivery mode prevents applications from blocking due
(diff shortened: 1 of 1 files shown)


config/containers/logging/configure.md
```diff
@@ -135,24 +135,22 @@ Docker provides two modes for delivering messages from the container to the log
 driver:
 
 * (default) direct, blocking delivery from container to driver
-* non-blocking delivery that stores log messages in an intermediate per-container
-  ring buffer for consumption by driver
+* non-blocking delivery that stores log messages in an intermediate per-container buffer for consumption by driver
```
PettitWesley, line 138: I'm guessing this edit will be contentious and I can remove it if you want. But with the way the code actually works it is not a true ring buffer, its just a buffer: https://github.com/moby/moby/blob/master/daemon/logger/ring.go#L160

cpuguy83 (approved):

config/containers/logging/configure.md
cpuguy83: Nah, this is perfect. It describes reality. I don't recall how these docs got written but they are definitely wrong.

thaJeztah (approved):
LGTM, thank you!

thaJeztah:
And, it's merged! Changes won't be visible yet on docs.docker.com, but will be part of the next deploy (probably on Monday)

PR #16994, opened 2023-03-28
ea-label

aevesdocker:
re add ea label as per Dixie and Mercedes' request https://docker.slack.com/archives/C03DVPF0P40/p1680015726257079?thread_ts=1680012881.783629&cid=C03DVPF0P40

### DIFF
--- a/docker-hub/company-owner.md
+++ b/docker-hub/company-owner.md
@@ -4,6 +4,11 @@ keywords: company, owners
 title: Manage company owners
 ---
 
+> **Note**
+>
+> Company layer is an [early access](../release-lifecycle.md#early-access-ea)
+> product, and requires a Docker Team or Business subscription.
+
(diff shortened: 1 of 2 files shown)


docker-hub/company-owner.md
```diff
@@ -4,6 +4,11 @@ keywords: company, owners
 title: Manage company owners
 ---
 
+> **Note**
+>
+> Company layer is an [early access](../release-lifecycle.md#early-access-ea)
```
usha-mandya, line 9: Are we calling this feature "Company layer"? If so, should we also update https://docs.docker.com/docker-hub/creating-companies?

aevesdocker: Checked with Mercedes and 'company layer' is approved. I don't think we need to update the company overview page though.

PR #16996, opened 2023-03-28
Minor changes for freshness to engine install index

ChrisChinchilla:
### Proposed changes

Minor changes for freshness to engine install index

### DIFF
--- a/engine/install/index.md
+++ b/engine/install/index.md
@@ -38,7 +38,7 @@ redirect_from:
 
 ## Supported platforms
 
-Docker Engine is available on a variety of [Linux platforms](../../desktop/install/linux-install.md),
+Docker Engine is available on a variety of [Linux distros](../../desktop/install/linux-install.md),
 [macOS](../../desktop/install/mac-install.md), and [Windows 10](../../desktop/install/windows-install.md)
 through Docker Desktop, and as a [static binary installation](binaries.md). Find
 your preferred operating system below.
@@ -55,7 +55,7 @@ your preferred operating system below.
 
 ### Server
 
-Docker provides `.deb` and `.rpm` packages from the following Linux distributions
+Docker provides `.deb` and `.rpm` packages from the following Linux distros
 and architectures:
 
 | Platform                | x86_64 / amd64         | arm64 / aarch64        | arm (32-bit)           | s390x                  |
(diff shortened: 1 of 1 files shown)


craig-osterhout (dismissed):

engine/install/index.md
```diff
@@ -73,28 +73,28 @@ and architectures:
...
 > **Note**
 >
-> While the instructions below may work, Docker does not test or verify
-> installation on derivatives.
+> While the instructions below may work, Docker doesn't test or verify
+> installation on distribution derivatives.
 
 - Users of Debian derivatives such as "BunsenLabs Linux", "Kali Linux" or 
   "LMDE" (Debian-based Mint) should follow the installation instructions for
-  [Debian](debian.md), substituting the version of their distro for the
```
thaJeztah, line 81: I must admit I got quite used to `distro` to be used when talking about Linux distributions. For me it helped disambiguate "distribution" (which can have quite some meanings, depending on context), which led to the alternative "Linux distribution" (which is a bit of a mouth full)


I guess I mean to say; do we know how common "distro" is (maybe my view is squashed from within my bubble), and could it still be an alternative to be considered ?

engine/install/index.md
```diff
@@ -30,16 +30,16 @@ redirect_from:
 
 > **Docker Desktop for Linux**
 >
-> Docker Desktop helps you build, share, and run containers easily on Mac and
```
thaJeztah, line 33: :heart: one of my pet-peeves to avoid these

engine/install/index.md
ChrisChinchilla: @thaJeztah For me it's mostly not using "slang" for people who may not be familiar with it… But in the case of Linux it's probably safe to say they are, so it's not a change I am dedicated to keeping…

thaJeztah: Yup! That's a fair point, and my view may be skewed indeed. Perhaps we can find out what users would expect to be used for this (so that we can converge on one option)

ChrisChinchilla:
Undone @thaJeztah

dvdksn (approved):
LGTM

thaJeztah:
Hm, link checker found an issue, but perhaps it was elsewhere;

```
- ./_site/compose/compose-file/build/index.html
  *  linking to internal hash #platform that does not exist (line 542)
     <a href="05-services#platform">platforms</a>

HTML-Proofer found 1 failure!
```

ChrisChinchilla:
@thaJeztah That doesn't look like anything to do with this PR, @aevesdocker could you add to your list?

PR #16998, opened 2023-03-28
add docker init

craig-osterhout:
### Proposed changes

Create topic for the docker init command.

### Related issues (optional)

DA-380

### DIFF
--- a/_data/toc.yaml
+++ b/_data/toc.yaml
@@ -511,6 +511,8 @@ reference:
       title: docker import
     - path: /engine/reference/commandline/info/
       title: docker info
+    - path: /engine/reference/commandline/init/
+      title: docker init (Beta)
     - path: /engine/reference/commandline/inspect/
       title: docker inspect
     - path: /engine/reference/commandline/kill/
--- a/engine/reference/commandline/init.md
+++ b/engine/reference/commandline/init.md
@@ -1,0 +1,147 @@
+---
+title: docker init (Beta)
+---
+
+> **Beta**
+>
+> The Docker Init plugin is currently in [Beta](../../../release-lifecycle.md/#beta). Docker recommends that you do not use this in production environments.
+
+Initialize a project with the files necessary to run the project in a container.
+
+## Usage
+
+```console
+$ docker init
+```
+
+`docker init` uses a CLI-based menu. Refer to the [examples](#examples) for more details.
+
+## Description
+
+Docker Desktop 4.18 and later provides the Docker Init plugin with the `docker init` CLI command. Run `docker init` in your project directory to be walked through the creation of the following files with sensible defaults for your project:
+
+ * .dockerignore
+ * Dockerfile
+ * docker-compose.yaml
+
+If any of the files already exist, a prompt appears and provides a warning as well as giving you the option to overwrite all the files.
+
+> **Warning**
+>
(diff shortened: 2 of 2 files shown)


engine/reference/commandline/init.md
```diff
@@ -0,0 +1,147 @@
...
+```
+
+The following appears after selecting the default **1.20**.
+
+```console
+? What version of Go do you want to use? 1.20
+? What's the relative directory (with a leading .) of your main package? (.)
+```
+
+The following appears after selecting the default **.**.
```
aevesdocker, line 122: the bolded "." isn't totally clear. Wonder if it would stand out more with code formatting?

aevesdocker (approved):

_data/toc.yaml
```diff
@@ -511,6 +511,8 @@ reference:
       title: docker import
     - path: /engine/reference/commandline/info/
       title: docker info
+    - path: /engine/reference/commandline/init/
```
crazy-max, line 514: I don't think this should live in the reference docs. This page is not a reference but a set of examples. I think it should go in Manuals instead, like other plugins (e.g., https://docs.docker.com/scout/). cc @thaJeztah

thaJeztah: Hmm.. yes, I agree.

Wondering if the plugin can generate yamldocs (as we do for the other plugins?) @rafd123  ?

rafd123: Is this what you're referring to, @thaJeztah?

https://github.com/docker/docker-init/tree/main/docs/reference

craig-osterhout: Thanks all.
@rafd123, yes, I think it's those files. Can we add the more comprehensive description and examples to those? If it's just a manually adding it to the yaml rather than some auto-generated thing, let me know and I can create a PR to update the yaml in the init repo.

@crazy-max, I'll remove the md files and use the yaml files to generate the content. I'm not sure what you mean by it's just a set of examples because the other reference docs have exactly the same sections. If you just mean it needs to be generated by the yaml, I'll do that, otherwise can you elaborate?

As for adding it to Manuals, we decided against it for this release as it is limited.

Since this goes live on Monday, I plan to just merge this PR to avoid any delays. I'll create a follow-up PR to update it to the yaml file. If any concerns, let me know.

thaJeztah:
@craig-osterhout @rafd123 @crazy-max I did a quick alternative; I took the generated YAML files from the repository, and combined it with the content from the markdown file from this PR (which is "roughly" what the docs generator would produce when setting up markdown content in the right locations). Also made the `docker init` into a section (as there's two commands (`docker init` and `docker init version`);

opened a PR here; https://github.com/docker/docs/pull/17016

craig-osterhout:
Closing this PR as it's superseded by https://github.com/docker/docs/pull/17016

nunocoracao:
@craig-osterhout why did we close this?

thaJeztah:
@nunocoracao see;

> Closing this PR as it's superseded by #17016

nunocoracao:
Thanks @thaJeztah any changes to the final docs file? Url? Content?

thaJeztah:
See the linked PR; main content will be at the same URL, but slightly different format to match the reference section layout

PR #17005, opened 2023-03-29
Refresh registry topics

craig-osterhout:
### Proposed changes

Refresh `/registry/` and `/registry/deploying/` for docs refresh.

### DIFF
--- a/_data/toc.yaml
+++ b/_data/toc.yaml
@@ -1949,7 +1949,7 @@ manuals:
   - sectiontitle: Docker Registry
     section:
     - path: /registry/
-      title: Registry overview
+      title: Overview
     - path: /registry/introduction/
       title: Understand the Registry
     - path: /registry/deploying/
--- a/_includes/registry.md
+++ b/_includes/registry.md
@@ -2,7 +2,7 @@
 
 >
 > This page contains information about hosting your own registry using the
-> [open source Docker Registry](https://github.com/distribution/distribution){: target="_blank" rel="noopener" class="_"}. For information about Docker Hub, which offers a
-> hosted registry with additional features such as teams, organizations, web
-> hooks, automated builds, etc, see [Docker Hub](/docker-hub/).
+> [open source Docker Registry](https://github.com/distribution/distribution){: target="_blank" rel="noopener" class="_"}. See [Docker Hub](/docker-hub/) for more information about Docker's hosted registry, which offers
+> features such as teams, organizations, web
+> hooks, automated builds, and more.
 {: .important }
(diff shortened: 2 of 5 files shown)


ChrisChinchilla (changes requested):
Small typo @craig-osterhout

thaJeztah:
Wah. We need to have a bit of a look at this (ISTR there was a ticket for this somewhere, but maybe @usha-mandya knows).

Let me try to give a quick summary;

- the open source registry was a project that was previously owned by Docker
- at some point, we moved its documentation into our docs repository (as it was one of our open-source projects, and we documented all our projects on our documentation website)

However, the open source registry has now been donated to the CNCF, and is not a Docker _product_. As part of donating to the CNCF, we also moved back the documentation to the project (see https://github.com/distribution/distribution/pull/3397)

Given that this registry is not part of our product catalogue (and it's just "one" way to run your own registry; there's many alternatives, such as Artifactory, or other solutions)

The intent is to remove (most of) the documentation for it, and where needed refer to the upstream project for further details; providing guidelines (and support) for that is out of scope for our documentation.


So basically;

- the open source registry _can_ be an easy way to quickly set up a local registry (in a container) for testing purposes, which can be useful for developer scenarios
- ^^ for that purpose we can show an example somewhere "here's how you could
- but clearly document that the example is for testing purposes, because running a registry _securely_ (TLS support, authentication) gets complex quickly, and we don't want to document "here's how to run a public registry in production"
- Sections like the storage drivers supported by the registry should be all out of scope for our docs (https://docs.docker.com/registry/storage-drivers/)

There's one other caveat to look into;

- Besides docs for running a registry, the open source project also contains some specifications. Most of that is obsoleted by the OCI distribution specification (which is implemented by Docker Hub)
- But there's also a specification for authentication (which is currently not part of the OCI distribution spec), and that part is still implemented by Docker Hub. (See https://docs.docker.com/registry/spec/auth/ )
- ^^ We need to consider how to document that specification _as part of our Docker Hub API reference_ (which can be either "copy the docs and move it to the docker hub section", or update the Docker Hub API docs themselves to document our implementation for Auth).

thaJeztah:
> ^^ for that purpose we can show an example somewhere "here's how you could run a registry for testing"

Maybe such a topic would be a better fit for our "Guides" section 🤔 (currently the registry topic lives under "manuals")

usha-mandya:
> The intent is to remove (most of) the documentation for it, and where needed refer to the upstream project for further details...

@dvdksn  had a PR to remove the registry docs - https://github.com/docker/docs/pull/15888.

We need to discuss this when David is back from parental leave.

thaJeztah:
DOH! Forgot about that one; I _knew_ I wrote something along these lines _somewhere_ but forgot "where" 😂

craig-osterhout:
Thanks all. I'll close this.

PR #17016, opened 2023-03-31
add docker init (carry)

thaJeztah:
- alternative to / replaces https://github.com/docker/docs/pull/16998
- closes https://github.com/docker/docs/pull/16998

Manually combining the generated YAML docs with the Markdown that was written by Craig.

Screenshot 2023-03-31 at 22 05 22
Screenshot 2023-03-31 at 22 06 09

### DIFF
--- a/_data/init-cli/docker_init.yaml
+++ b/_data/init-cli/docker_init.yaml
@@ -1,0 +1,150 @@
+command: docker init
+short: Creates Docker-related starter files for your project
+long: |-
+  Initialize a project with the files necessary to run the project in a container.
+
+  Docker Desktop 4.18 and later provides the Docker Init plugin with the `docker init` CLI command. Run `docker init` in your project directory to be walked through the creation of the following files with sensible defaults for your project:
+    
+    * .dockerignore
+    * Dockerfile
+    * docker-compose.yaml
+    
+    If any of the files already exist, a prompt appears and provides a warning as well as giving you the option to overwrite all the files.
+    
+    > **Warning**
+    >
+    > You can't recover overwritten files.
+    > To back up an existing file before selecting to overwrite it, rename the file or copy it to another directory.
+    {: .warning}
+  
+  After running `docker init`, you can choose one of the following templates:
+    
+    * Go: Suitable for a Go server application.
+    * Other: General purpose starting point for containerizing your application.
+  
+  After `docker init` has completed, you must modify the created files and tailor them to your project. Visit the following topics to learn more about the files:
+    
+    * [.dockerignore](../../../engine/reference/builder.md#dockerignore-file)
+    * [Dockerfile](../../../engine/reference/builder.md)
+    * [docker-compose.yaml](../../../compose/compose-file/03-compose-file.md)
+
+usage: docker init [OPTIONS]
+pname: docker
+plink: docker.yaml
+options:
+    - option: version
+      value_type: bool
+      default_value: "false"
+      description: Display version of the init plugin
+      deprecated: false
+      hidden: false
+      experimental: false
+      experimentalcli: false
+      kubernetes: false
+      swarm: false
+deprecated: false
+experimental: false
+experimentalcli: false
+kubernetes: false
+swarm: false
+examples: |-
+  ### Example of selecting Other
+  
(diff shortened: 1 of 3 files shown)


craig-osterhout:
Thanks @thaJeztah. I get a 404 in the preview. Do we need to add the markdown template files in engine/reference/commandline/?

thaJeztah:
DOH! Forgot to stage those I think 🙈 let me push

thaJeztah:
I guess a `git add` does wonders for that; should be fixed now 🤞

craig-osterhout (dismissed):
Thanks again for the quick help, @thaJeztah.
LGTM.

On a side note, I think the template doesn't handle a single child command well, or there's some other key that I'm missing. 
The "Related commands" table in the version topic is empty. Adding cname & clink in the parent topic fixes it, but then the version command just shows only itself in that table.

crazy-max (dismissed):

thaJeztah:
> On a side note, I think the template doesn't handle a single child command well, or there's some other key that I'm missing.
> The "Related commands" table in the version topic is empty. Adding cname & clink in the parent topic fixes it, but then the version command just shows only itself in that table.

Yes, we need to review some of that layout. I think we should event consider removing the "parent" and "child" command sections overall; it really overlaps with the left-hand navigation, so I wonder if it adds much value.

dvdksn (dismissed):

rafd123:
Heads up that there's no longer a dedicated `version` command. The corresponding docs should have been removed with [this PR](https://github.com/docker/docker-init/pull/16).

I'll update the `docker init` repo accordingly, but just wanted to give a heads up.

rafd123 (changes requested):
We should remove references to the `version` command since `docker init` is not shipping with it.

[This PR in the `docker init` repo](https://github.com/docker/docker-init/pull/16) should have removed it, but it slipped through the cracks. Created[ this PR](https://github.com/docker/docker-init/pull/22) to rectify that.

thaJeztah:
Pushed a quick commit to remove it (and changed the "section" back to a single link)

If it looks good, I can squash the two commits.

_data/init-cli/docker_init.yaml
```diff
@@ -0,0 +1,151 @@
...
+    * [.dockerignore](../../../engine/reference/builder.md#dockerignore-file)
+    * [Dockerfile](../../../engine/reference/builder.md)
+    * [docker-compose.yaml](../../../compose/compose-file/03-compose-file.md)
+
+usage: docker init [OPTIONS]
+pname: docker
+plink: docker.yaml
+options:
+    - option: version
+      shorthand: v
```
thaJeztah, line 36: @rafd123 do you know if the shorthand `-v` has been discussed? We try to avoid shorthands if there's chances they become ambiguous. In this case `-v` could collide with `--verbose` on some other commands (which could potentially be an option we might add to `docker init` in future, and which would be more commonly used by users than `--version`).

thaJeztah: ^^ If we're unable to update the binary, we could decide to just remove it here from the docs for now)

rafd123: It hasn't come up. FWIW, `docker init` was modeled after `docker dev`, which includes a shorthand for `--version`: https://github.com/docker/docker-dev/blob/main/docs/reference/dev.md

Whether we update the binary to exclude it is a function of whether it's worth cutting a new 4.18 release, I *think*.

rafd123 (dismissed):

_data/init-cli/docker_init.yaml
rafd123: With the understanding that it has no bearing on the 4.18 release, any objection to me just removing the `-v` shorthand from `docker init` now so we don't forget?

thaJeztah:
Pushed a commit to remove it from the docs (can squash all commits later if we're good)

rafd123 (dismissed):

_data/init-cli/docker_init.yaml
thaJeztah: Seems fine to me!

(I can also give some pointers on how to make it generate the page as it is now for future builds)

craig-osterhout (approved):

PR #17018, opened 2023-03-31
docs: Move Post install link to infobox

KatrinaAS:
Moved Post-Install link to inbobox to make it more visible.

### Proposed changes

This was done as when people are trying to install Docker to use software I assist in maintain. I have countless times had to link them to the post-install instructions, to the point of one of the people maintain it has suggested linking to it in our own README.md. As they are so unclear to find unless you read what looks to be boilerplate "You have completed install" text

### DIFF
--- a/engine/install/ubuntu.md
+++ b/engine/install/ubuntu.md
@@ -180,11 +180,10 @@ Docker from the repository.
    This command downloads a test image and runs it in a container. When the
    container runs, it prints a confirmation message and exits.
 
-You have now successfully installed and started Docker Engine. The `docker` user
-group exists but contains no users, which is why you're required to use `sudo`
-to run Docker commands. Continue to [Linux post-install](linux-postinstall.md)
-to allow non-privileged users to run Docker commands and for other optional
-configuration steps.
+You have now successfully installed and started Docker Engine. 
+> The `docker` user group exists but contains no users, which is why you're required 
+> to use `sudo` to run Docker commands. Continue to [Linux post-install](linux-postinstall.md)
+> to allow non-privileged users to run Docker commands and for other optional configuration steps.
 
 #### Upgrade Docker Engine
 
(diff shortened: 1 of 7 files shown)


KatrinaAS:
Added a title to the infobox to make it more clear.

engine/install/ubuntu.md
```diff
@@ -180,11 +180,12 @@ Docker from the repository.
    This command downloads a test image and runs it in a container. When the
    container runs, it prints a confirmation message and exits.
 
-You have now successfully installed and started Docker Engine. The `docker` user
-group exists but contains no users, which is why you're required to use `sudo`
-to run Docker commands. Continue to [Linux post-install](linux-postinstall.md)
-to allow non-privileged users to run Docker commands and for other optional
-configuration steps.
+You have now successfully installed and started Docker Engine. 
+> Receiving errors when trying to run without root?
```
thaJeztah, line 184: Could you keep an empty line between the line and the infobox?
```suggestion
You have now successfully installed and started Docker Engine.

> Receiving errors when trying to run without root?
```

thaJeztah (commented):
If we change this page, we should probably change all the other Linux distros (such as the [`debian.md` in the same directory](https://github.com/docker/docs/blob/main/engine/install/debian.md) with the same changes

KatrinaAS:
> If we change this page, we should probably change all the other Linux distros (such as the [`debian.md` in the same directory](https://github.com/docker/docs/blob/main/engine/install/debian.md) with the same changes

Definatly makes sense, will do that

KatrinaAS:
I made sure to follow the existing documentation, so only mentioned the docker group existing if the existing documentation did.

ChrisChinchilla:
Thanks @KatrinaAS I refactored the change into using an include so we can change the text in one place. For some reason my change isn't showing up here… But GitHub had some issues today, maybe it's taking it's time… 🤔

aevesdocker (approved):

ChrisChinchilla:
Thanks @KatrinaAS !

PR #17020, opened 2023-04-01
Docker Community Slack joining URL

shubham-kshetre:


shubham-kshetre:
The Docker Community Slack URL for joining the community only accepts email addresses with the domains docker.com and jackmorton.com. However, I found a link that works perfectly fine for email addresses with other domains as well. I'm eager to discuss other solution if this is not better about this issue.

PR #17021, opened 2023-04-01
Update uninstall.md

cimplylimited:
I discovered a daemon file that was persisting from my old mac machine.  It had been restored on the new computer from a Time Machine backup and inherited many of the system configurations from the Intel macbook.  Here is how I fixed that uninstall.  You may want to investigate as the daemon file was not purged in any of the uninstall methods from the existing documentation.


cimplylimited:
Understood. Just note that it is likely to happen any time someone is upgrading their machine. It's persisted in a number of application setups and tends to occur even with fresh uninstalls/reinstalls. Anyways, appreciate the response.

PR #17037, opened 2023-04-05
Add example of passing build args

ibrahima:
### Proposed changes

Added an example for passing build args in a GitHub action.

I did this because the syntax was not obvious, I had to find it from a GitHub issue: https://github.com/docker/build-push-action/issues/557

### Related issues (optional)

https://github.com/docker/build-push-action/issues/557


ibrahima:
Closing because I realized this is documented on https://github.com/docker/build-push-action#customizing and I just missed it.

PR #17039, opened 2023-04-05
Added link to the Apache 2.0 License #17038

Satyxm:
Now the Apache 2.0 License redirects to the Apache 2.0 license. It would be convenient for the user to get to the license with a single click.

### Proposed changes

I added a link to the Apache 2.0 license, which now is a link that redirects to Apache 2.0 license


craig-osterhout:
Fixes #17038

PR #17044, opened 2023-04-05
Update 07_multi_container.md

Ativerc:
### Proposed changes

In line 198: Made it much more clearer about the CWD from where the `docker run` command needs to be run.

In this article, the commands for setting up the `to-do` docker network, running the `mysql` container and finding its IP can be run from any CWD. 

However, the `docker run` command here needs to be specifically run from `getting-started/app` directory. I know this is implied but I just came upon this article while googling and it's better to be explicit about these things than being implicit.


craig-osterhout (approved):
Thank you @Ativerc, LGTM.

PR #17063, opened 2023-04-07
scout: add consistent banners for "early access"

thaJeztah:
Docker Scout is in early access, but some parts of the documentation (reference docs) did not contain a banner for this.

This patch:

- Changes the "note" to "early access" (just a minor improvment)
- Adds banners to all the CLI reference pages
- Adds (Early Access) in the CLI reference navigation (TOC)

Banners now use "early access" as heading, instead of the blanket "note"

Before:

Screenshot 2023-04-07 at 15 25 19

After:

Screenshot 2023-04-07 at 15 25 07

The CLI reference docs now has "(Early Access)" in the TOC, and pages have a "Early Access" banner;

Screenshot 2023-04-07 at 15 27 23
Screenshot 2023-04-07 at 15 27 57

### DIFF
--- a/engine/reference/commandline/scout.md
+++ b/engine/reference/commandline/scout.md
@@ -10,4 +10,7 @@ in the source repository on GitHub:
 
 https://github.com/docker/scout-cli-plugin
 -->
+
+{% include scout-early-access.md %}
+
 {% include cli.md datafolder=page.datafolder datafile=page.datafile %}
--- a/engine/reference/commandline/scout_compare.md
+++ b/engine/reference/commandline/scout_compare.md
@@ -10,4 +10,7 @@ in the source repository on GitHub:
 
 https://github.com/docker/scout-cli-plugin
 -->
+
+{% include scout-early-access.md %}
+
 {% include cli.md datafolder=page.datafolder datafile=page.datafile %}
--- a/_data/toc.yaml
+++ b/_data/toc.yaml
@@ -619,7 +619,7 @@ reference:
       title: docker run
     - path: /engine/reference/commandline/save/
       title: docker save
-    - sectiontitle: docker scout
+    - sectiontitle: docker scout (Early Access)
       section:
       - path: /engine/reference/commandline/scout/
         title: docker scout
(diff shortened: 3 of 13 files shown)


amyb12345:
These changes LGTM

engine/reference/commandline/scout.md
```diff
@@ -10,4 +10,14 @@ in the source repository on GitHub:
...
 -->
+
+> **Early Access**
+>
+> Docker Scout is an [early access](../../../release-lifecycle.md#early-access-ea)
+> product, and requires a Docker Pro, Team, or Business subscription.
+>
+> If you're interested in Docker Scout for your organization and want to
+> learn more, get in touch by filling out the contact form on the
+> [Docker Scout product page](https://docker.com/products/docker-scout){: target="_blank" rel="noopener" }.
```
crazy-max, line 21: Can we have an include for this to avoid dedup?

thaJeztah: Yeah, I was considering for a bit; it was already duplicated, and there was still some discussion if the "cli" and "scout" (as a whole) have the same stability (early access / beta), so I decided to (for now) just copy/paste.

Let me know if you feel strongly about that; I can update.

thaJeztah: Let me have a quick look at making a include

crazy-max (approved):

engine/reference/commandline/scout_compare.md
```diff
@@ -10,4 +10,7 @@ in the source repository on GitHub:
 
 https://github.com/docker/scout-cli-plugin
 -->
+
+{% include scout-early-access.md %}
+
```
mikeparker, line 15: @thaJeztah did you see the comment above this saying this file is auto generated? Just checking this isn't going to get removed automatically at some point in the future.

thaJeztah: @mikeparker ah, yes, I guess the comment is slightly ambiguous on these; so the markdown "stub" files themselves are not automatically generated, but the content is generated from YAML files (https://github.com/docker/docs/tree/c0402932ca01a11eb663080191ba5f4b7e9194dd/_data/scout-cli) combined with a template.

So unless scout also generates the stubs, I think we're good 👍

PR #17064, opened 2023-04-07
redirect: add /go link for Compose v1 EOL

milas:
### Proposed changes
Add a `/go` link for Compose v1 EOL info.

Redirecting to the "Evolution of Compose" page for the moment. We will probably adjust this page once Compose V1 is removed from Docker Desktop in a couple months.

### Related issues (optional)
* #16641 (Compose v1 EOL banner in docs)

### DIFF
--- a/_redirects.yml
+++ b/_redirects.yml
@@ -105,6 +105,8 @@
   - /go/deprecated/
 "/engine/reference/commandline/cli/#experimental-features":
   - /go/experimental/
+"/compose/compose-v2/":
+  - /go/compose-v1-eol/
 "/config/formatting/":
   # Instructions on using Go templates to format CLI output with --format
   - /go/formatting/


_redirects.yml
```diff
@@ -105,6 +105,8 @@
   - /go/deprecated/
 "/engine/reference/commandline/cli/#experimental-features":
   - /go/experimental/
+"/compose/compose-v2":
```
thaJeztah, line 108: oh! almost missed that (and we should have validation for this); could you add a trailing `/`? (I _think_ otherwise there will be a double redirect. `/go/compose-v1-eol/ -> /compose/compose-v2 -> /go/compose-v1-eol/`

thaJeztah (approved):
LGTM!

PR #17068, opened 2023-04-08
Fix terminology about container creation

nicomwong:
### Proposed changes

Change terminology from "start" to either "run" or "create" as it relates to the `docker run` command which basically performs `docker create` then `docker start`. The usage of "start" in this part of the getting-started tutorial particularly creates confusion since starting a docker container would imply it already exists and was at some point either not started or it had stopped. I am confident that the authors intend to say that upon container creation, file state is not retained, rather than start.

Further, the author(s) may have become confused themselves, seeing by the usage of "restarts" in the following sentence from the documentation:
>If we mount that same directory across container restarts, we'd see the same files.


nicomwong:
Looks like there are further terminology fixes needed in the steps of the "Persisting our Todo Data" section.

craig-osterhout:
Thank you @nicomwong. It looks like there are some conflicts. If you can resolve those, I can get this merged.

Issue #14464, opened 2022-03-29
doc improvement suggestion


nankingrider:
[network/network-tutorial-overlay.md](https://docs.docker.com/network/network-tutorial-overlay/)

HI, the content below this title "Communicate between a container and a swarm service[🔗](https://docs.docker.com/network/network-tutorial-overlay/#communicate-between-a-container-and-a-swarm-service)"

has nothing to do with it ... maybe a mislocation, please check
thanks . !

craig-osterhout:
Thanks for sharing the issue. Closing as duplicate of https://github.com/docker/docker.github.io/issues/8387

Issue #9074, opened 2019-07-17
"Communicate between a container and a swarm service" needs to be updated.  It is a duplicate of the bridge.


zkSNARK:
[network/network-tutorial-overlay.md](https://docs.docker.com/network/network-tutorial-overlay/)

Mikuu:
+1

pluckhuang:
the same #9046

DavidSonoda:
Please fix this as soon as possible. 

I looked for this tutorial for a long time and when I finally found it here it's just not related to the 'Communication between a container and a swarm service' at all. Does anyone have the correct tutorial?

craig-osterhout:
Thanks for sharing the issue. Closing as duplicate of https://github.com/docker/docker.github.io/issues/8387

Issue #8387, opened 2019-03-01
The heading says communicate between swarm service and standalone container but explanation is for communication between two standalone containers


wlmpoc:
[network/network-tutorial-overlay.md](https://docs.docker.com/network/network-tutorial-overlay/)

wlmpoc:
https://docs.docker.com/network/network-tutorial-overlay/#communicate-between-a-container-and-a-swarm-service

This mainly explains about running two alpine Linux containers attached to default bridge network with the ping <hostname> and ping <ip address> outputs.

Please update this with an example for standalone container and swarm service

Thanks
Sathya

eeechoo:
same problem.
It's sad when read the official doc to learn docker network and find there are problems in the doc.

Issue #7534, opened 2018-10-18
Remove section of Overlay Networking Tutorial


tduval:
[network/network-tutorial-overlay.md](https://docs.docker.com/network/network-tutorial-overlay/)

The last section of the page (_Communicate between a container and a swarm service_) is unrelated to the Overlay Networking and is a duplicata of the Bridge Network Tutorial.

craig-osterhout:
Thanks for sharing the issue. Closing as duplicate of https://github.com/docker/docker.github.io/issues/8387

PR #17071, opened 2023-04-08
fix: Fix incorrect information in Communicate between a container and…

### LINKED ISSUE #15384: Wrong content under Communicate between a container and a swarm service
[network/network-tutorial-overlay.md](https://docs.docker.com/network/network-tutorial-overlay/)

Under section, Communicate between a container and a swarm service

The details show only communication among containers in a bridge network and does not show communication of container to Service (overlay network), please check. thanks !

vashian:
… a swarm service section #8387

[network/network-tutorial-overlay.md](https://docs.docker.com/network/network-tutorial-overlay/)

Under the section, Communicate between a container and a swarm service

The details show only communication among containers in a bridge network and do not show communication of container to Service (overlay network)

  Refer to related PRs or issues:

  Closes #15384, #14464, #7534, #9074

### DIFF
--- a/network/network-tutorial-overlay.md
+++ b/network/network-tutorial-overlay.md
@@ -433,203 +433,76 @@ example also uses Linux hosts, but the same commands work on Windows.
 
 ## Communicate between a container and a swarm service
 
-In this example, you start two different `alpine` containers on the same Docker
-host and do some tests to understand how they communicate with each other. You
-need to have Docker installed and running.
+In order to communicate between a container and a swarm service in Docker, you can use Docker's built-in networking functionality. Docker provides a default overlay network called **ingress** that allows containers to communicate with each other and with swarm services.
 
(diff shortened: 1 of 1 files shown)


thaJeztah:
Hm... looking at that example, I feel like that example (but need to check) on itself may be good, but somehow ended up in the wrong section.

I see that example was added (or moved from elsewhere) in https://github.com/docker/docs/commit/a8e84a80e1b9968b55d3487bbfef0a98bcd5c5a4, which was part of some large rewrite of the networking  section;

- https://github.com/docker/docs/pull/5936


Perhaps we need to look if the existing example could be (should be) used elsewhere, and then indeed have an example that illustrates how to run a "regular" (non-swarm) container, and attach it to a overlay network (that has the `--attachable` option set) 🤔 


/cc @dvdksn @akerouanton

PR #17074, opened 2023-04-09
Add a missing dash

dhrions:
### Proposed changes

Replace `docker compose` by `docker-compose` to correct a syntax error

### DIFF
--- a/language/python/develop.md
+++ b/language/python/develop.md
@@ -228,7 +228,7 @@ If you have any containers running from the previous sections, [stop](./run-cont
 Now, to start our application and to confirm that it is running properly, run the following command:
 
 ```console
-$ docker compose -f docker-compose.dev.yml up --build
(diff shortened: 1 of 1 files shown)


thaJeztah (changes requested):
I don't think this is the correct change. What version of docker do you have installed ? Compose is now integrated with the `docker` The default command for compose is now `docker compose` (but it's possible to set up aliases for the old (`docker-compose`) command); see https://docs.docker.com/compose/compose-v2/#differences-between-compose-v1-and-compose-v2

dhrions:
> I don't think this is the correct change. What version of docker do you have installed ? Compose is now integrated with the `docker` CLI, and the default command for compose is now `docker compose` (but it's possible to set up aliases for the old (`docker-compose`) command); see https://docs.docker.com/compose/compose-v2/#differences-between-compose-v1-and-compose-v2

PR #17076, opened 2023-04-09
docs(docker-ps): correct wrong example command

ansidev:
### Proposed changes

- Correct the wrong example command.

### DIFF
--- a/_data/engine-cli/docker_ps.yaml
+++ b/_data/engine-cli/docker_ps.yaml
@@ -521,7 +521,7 @@ examples: |-
     To list all running containers in JSON format, use the `json` directive:
 
     ```console
-    $ docker ps --format json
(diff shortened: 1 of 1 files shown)


_data/engine-cli/docker_ps.yaml
```diff
@@ -521,7 +521,7 @@ examples: |-
     To list all running containers in JSON format, use the `json` directive:
 
     ```console
-    $ docker ps --format json
+    $ docker ps --format '{{json .}}'
```
thaJeztah, line 524: The example is correct, but requires docker 23.0 or up. If you're using docker desktop, you may still be on the previous (docker 20.10) version, which doesn't support this option

thaJeztah (changes requested):

PR #17085, opened 2023-04-11
Add acronym terms to unpack to Vale

ChrisChinchilla:


aevesdocker (approved):

.github/vale/Docker/Acronyms.yml
```diff
@@ -63,3 +63,5 @@ exceptions:
   - XSS
   - YAML
   - ZIP
+  - SBOM
+  - CVE
```
thaJeztah, line 67: OCD me: perhaps we should keep this list sorted (otherwise it may be easy to miss if there's duplicates, or harder to find if things are already in the list) 😅

dvdksn: > perhaps we should keep this list sorted

Agree, I always try to do a `:sort` when I add a new item to these (mostly because I'm too lazy to try and find where I should put the new ones! 😅 )

ChrisChinchilla: Didn't even notice the comment from @dvdksn there sorry. I am not sure if there is any kind of automatic sorted for these files, but they are just text files, so I guess. Will add it to the list…

PR #17108, opened 2023-04-13
Remove misleading "This is the newest version." from "Compose file version 3 reference"

dcki-ts:
### Proposed changes

Remove misleading "This is the newest version." from "Compose file version 3 reference" article.

This change is consistent with https://github.com/docker/docs/blob/main/compose/compose-file/index.md

### Related issues (optional)

None.


aevesdocker (approved):
Thank you @dcki-ts

PR #17131, opened 2023-04-18
ENGDOCS-1303

aevesdocker:
Makes it clearer that this is legacy content and addresses some HotJar feedback

https://insights.hotjar.com/sites/3169877/feedback/responses?date=last-30-days&url_contains=compose&has_message=yes&frid=92881187

Also moves some content around

### DIFF
--- a/compose/compose-file/compose-versioning.md
+++ b/compose/compose-file/compose-versioning.md
@@ -4,6 +4,17 @@ keywords: fig, composition, compose, versions, upgrading, docker
 title: Compose file versions and upgrading
 ---
 
+>**Warning**
+>
+>This page contains information on the legacy versions of Compose, also collectively referred to as Compose V1.
+>From the end of June 2023 Compose V1 won’t be supported anymore.
+>
+>The latest and recommended version of Compose is the [Compose Specification](index.md). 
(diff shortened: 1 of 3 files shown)


compose/compose-file/compose-versioning.md
```diff
@@ -4,14 +4,24 @@ keywords: fig, composition, compose, versions, upgrading, docker
 title: Compose file versions and upgrading
 ---
 
+>**Warning**
+>
+>This page contains information on the legacy versions of Compose, also collectively referred to as Compose V1.
+>From the end of June 2023 Compose V1 won’t be supported anymore.
```
ChrisChinchilla, line 10: @aevesdocker Is there a better way to say this? It feels a bit passive… But I am not sure of the internal processes behind this…

ChrisChinchilla (approved):
Some thoughts, but approving to unblock

compose/compose-file/compose-versioning.md
aevesdocker: This matches what we have elsewhere in the docs since Jan. But I hear your point. I'll maybe address this in a follow-up PR

PR #17134, opened 2023-04-18
Update index.md

Bernardstanislas:
### Proposed changes

Update potential grammar mistake if we consider "Dev Environments" is a plural.


aevesdocker (approved):
Thanks for catching this @Bernardstanislas

PR #17144, opened 2023-04-19
Clarify GPU support

MihaelaStoica:
### Proposed changes
Add a note to clarify that the GPU support is only available in DD4W with the WSL backend


aevesdocker (approved):
Thank you @MihaelaStoica

PR #17148, opened 2023-04-19
Add modal check

craig-osterhout:
### Proposed changes

When visiting some pages, the following error occurs:
```
modal.js:26 Uncaught TypeError: Cannot read properties of null (reading 'addEventListener')
    at modal.js:26:7
```
For example, can reproduce on 
https://docs.docker.com/reference/ and the 404 page https://docs.docker.com/adsfasdf

Added check to see if modal exists before doing the operations.
I didn't dig down to find the cause, so there may be a better way to fix this.

### Related issues (optional)

ENGDOCS-1308

### DIFF
--- a/assets/js/modal.js
+++ b/assets/js/modal.js
@@ -1,31 +1,33 @@
 // Get the modal for image 1 
 var modal = document.getElementById("img-modal1");
 
-// Get the image and insert it inside the modal - use its "alt" text as a caption
-const images = document.querySelectorAll("main img");
-var modalImg = document.getElementById("img-modal-img1");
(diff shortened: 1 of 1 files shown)


aevesdocker:
I still can't reproduce this error for some reason. @dvdksn any ideas?

dvdksn:
I'm able to reproduce the issue too.

@aevesdocker if you open the dev console on any of the pages that craig linked in the PR description, you should see an uncaught typeerror:

image

dvdksn (approved):
This fix LGTM 👍🏻

PR #17159, opened 2023-04-21
standalone docker-compose is still supported, while discouraged

ndeloof:
### Proposed changes
don't tell users we no longer support docker-compose used as standalone executable, just discourage this usage

### Related issues (optional)
closes https://github.com/docker/compose/issues/10485

### DIFF
--- a/compose/install/index.md
+++ b/compose/install/index.md
@@ -41,5 +41,5 @@ You can [install the Compose standalone](other.md) on Linux or on Windows Server
 
 >Note
 >
->This install scenario is not recommended and only supported for backward compatibility purpose.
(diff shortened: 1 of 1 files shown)


compose/install/index.md
```diff
@@ -41,5 +41,5 @@ You can [install the Compose standalone](other.md) on Linux or on Windows Server
 
 >Note
 >
->This install scenario is no longer supported.
+>This install scenario is not recommended and only supported for backward compatibility purpose.
```
aevesdocker, line 44:
```suggestion
>This install scenario is not recommended and is only supported for backward compatibility purposes.
```

PR #17172, opened 2023-04-24
ENGDOCS-1316

aevesdocker:
Removes outdated compose information and refreshes some old links

### DIFF
--- a/get-started/08_using_compose.md
+++ b/get-started/08_using_compose.md
@@ -20,6 +20,8 @@ So, how do we get started?
 If you installed Docker Desktop for Windows, Mac, or Linux you already have Docker Compose!
 Play-with-Docker instances already have Docker Compose installed as well. 
 
+Standalone installations of Docker Engine requires Docker Compose to be installed as a separate package, see [Install the Compose plugin](../compose/install/linux.md).
+
 After installation, you should be able to run the following and see version information.
 
 ```console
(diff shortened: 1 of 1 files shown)


get-started/08_using_compose.md
```diff
@@ -17,9 +17,8 @@ So, how do we get started?
 
 ## Install Docker Compose
 
-If you installed Docker Desktop/Toolbox for either Windows or Mac, you already have Docker Compose!
-Play-with-Docker instances already have Docker Compose installed as well. If you are on
-a Linux machine, you will need to [install Docker Compose](../compose/install/index.md).
+If you installed Docker Desktop for Windows, Mac, or Linux you already have Docker Compose!
+Play-with-Docker instances already have Docker Compose installed as well.
```
dvdksn, line 21: I think we should keep the reference for how to install the compose plugin manually, and mention that it is required for users who install Docker Engine standalone.

```suggestion
If you installed Docker Desktop for Windows, Mac, or Linux you already have Docker Compose.
Play-with-Docker instances already have Docker Compose installed as well.
Standalone installations of Docker Engine requires Docker Compose to be installed as a separate package, see [Install the Compose plugin](../compose/install/linux.md).
```

aevesdocker: Ah yes. That's a Monday-morning-vibes mistake from me

PR #17175, opened 2023-04-24
FIX - switch docker repo to centos

### LINKED ISSUE #13463: Provide x86_64 RPMs for RHEL 8
### Problem description

In the manual installation steps for Docker CE for RHEL 8 (https://docs.docker.com/engine/install/rhel/#install-from-a-package), it says to browse to https://download.docker.com/linux/rhel/8/ to download an offline RPM. But no x86_64 binaries are listed.

### Problem location

- I saw a problem on the following URL: https://download.docker.com/linux/rhel/8/

- I couldn't find the information I wanted. I expected to find it near the following URL: https://download.docker.com/linux/rhel/8/x86_64

### Suggestions for a fix

Upload x86_64 RPMs to https://download.docker.com/linux/rhel/8/x86_64

There are x86_64 RPMs for CentOS 8, strangely enough: https://download.docker.com/linux/centos/8/x86_64/stable/Packages/

But it's not clear if those are officially supported (or why they aren't also under `/rhel`. Also CentOS 8 is end of life soon, so it's not clear how that factors into this.

mschreib28:
### Proposed changes
Update config-manager repo to point to centos instead of RHEL as referenced in https://access.redhat.com/discussions/6249651 and graciously solved by https://stackoverflow.com/questions/70358656/rhel8-fedora-yum-dns-causes-cannot-download-repodata-repomd-xml-for-docker-ce

This was done because RHEL7 is nearing EOL (https://endoflife.date/rhel), and while RHEL8 is nearing EOL too, 8 is at least more relevant.

### Related issues (optional)

#13463
Closes #13463

### DIFF
--- a/engine/install/rhel.md
+++ b/engine/install/rhel.md
@@ -82,7 +82,7 @@ from the repository.
 
 #### Set up the repository
 
-{% assign download-url-base = "https://download.docker.com/linux/rhel" %}
+{% assign download-url-base = "https://download.docker.com/linux/centos" %}
 
 Install the `yum-utils` package (which provides the `yum-config-manager`
 utility) and set up the repository.
@@ -90,7 +90,7 @@ utility) and set up the repository.
 ```console
 $ sudo yum install -y yum-utils
 
-$ sudo yum-config-manager \
+$ sudo dnf config-manager \
     --add-repo \
     {{ download-url-base }}/docker-ce.repo
 ```


dvdksn:
ping @thaJeztah

thaJeztah (changes requested):
This is not the correct change, as it would switch the package repo for users installing the packages on s390x, for which there's no CentOS packages.

This is expected; we currently only provide packages for RHEL running on s390x (because CentOS does not support s390x). For other architectures, users running RHEL can try installing the CentOS packages, as outlined in the prerequisites; https://docs.docker.com/engine/install/rhel/#prerequisites

Screenshot 2023-04-25 at 12 06 12

We may provide packages for RHEL in future, but no decision was made on that yet.

thaJeztah:
I'll close this PR because of the above

PR #17180, opened 2023-04-25
build(ci): distribute build across multiple runners

crazy-max:
### Proposed changes

Adds an example using a matrix strategy to distribute the build for each platform across multiple runners and merge the resulting manifests into a single image using the `buildx imagetools create` command.

### Related issues (optional)

fixes https://github.com/docker/docs/issues/16593

cc @kaovilai

### DIFF
--- a/build/ci/github-actions/multi-platform.md
+++ b/build/ci/github-actions/multi-platform.md
@@ -51,3 +51,125 @@ jobs:
           tags: user/app:latest
 ```
 {% endraw %}
+
+## Distribute build across multiple runners
+
+In the previous example, each platform is built on the same runner which can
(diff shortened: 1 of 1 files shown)


dvdksn (approved):
Using a local registry for this is so cool

crazy-max:
I go ahead and merge this one. In follow-up I think we could provide a composite action for a better integration so user would not need to add extra steps to load, push the local image but also handling of imagetools command.

jedevc:
SGTM :tada:

Realized I hadn't shared my thoughts from a DM with @crazy-max:

> alternatively. we can use the push-by-digest key i think (might wanna check my sources on that), to only push the digest - then we grab the digests from the buildx output digest, and combine those

PR #17184, opened 2023-04-25
Update DVP Data API spec

mcapell:
### Proposed changes

Updated the DVP Data API spec.

For instance, see: https://deploy-preview-17151--docsdocker.netlify.app/docker-hub/api/dvp/#tag/discovery/operation/getUserNamespaces

### Related issues (optional)

This PR is the same as https://github.com/docker/docs/pull/17151, but I'm creating a new PR as @felipecruz91 is on parental leave and I don't have permissions to update his fork.

### DIFF
--- a/docker-hub/api/dvp.yaml
+++ b/docker-hub/api/dvp.yaml
@@ -18,7 +18,7 @@ info:
      - Tag- or digest-level, a summary of every namespace, repository, and reference
        (tag or digest)
 
-      The summary data formats contain the following data points:
+     The summary data formats contain the following data points:
 
      - Unique IP address count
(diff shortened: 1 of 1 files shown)


ChrisChinchilla (approved):
There's some grammar i'd generally like to fix, but it's not part of the PR, so all good ;)

mcapell:
Cool, thank you for the review! We sync the openapi spec from [here](https://github.com/docker/hub.publishers-dvp-data/blob/main/api/data/v1/data.oapi), so probably we should fix the grammar there.

Feel free to merge, I don't have permissions to do so 😃

PR #17209, opened 2023-04-28
Compose v2 guide [ENGDOCS-1319]

milas:
See #17198 -- these are a bunch of changes I made on top.


_includes/compose-eol.md
```diff
@@ -1,6 +1,6 @@
 > **Important**
 >
-> From the end of June 2023 Compose V1 won't be supported anymore and will be removed from all Docker Desktop versions.
+> Compose V1 is no longer supported and will not be available in new releases of Docker Desktop after June 2023.
```
aevesdocker, line 3: Don't we want to merge this now? This implies we have to wait till June to merge

compose/upgrade.md
```diff
@@ -0,0 +1,133 @@
...
+redirect_from:
+  - /compose/compose-v2/
+---
+
+{% include compose-eol.md %}
+
+## What are the functional differences between Compose V1 and Compose V2?
+
+Compose V2 introduces a new internal architecture integrated with the rest of the Docker platform.
+It offers an improved CLI experience, improved build performance with BuildKit, and continued new feature development.
```
aevesdocker, line 14:
```suggestion
It offers an improved CLI experience, improved build performance with BuildKit, and continued new-feature development.
```

compose/upgrade.md
```diff
@@ -0,0 +1,133 @@
...
+Ask yourself...
+- Is my project using multiple levels of environment variable overrides (e.g. `.env` file and `--env` CLI flags)?
+- Do any `.env` file values have escape sequences or nested quotes?
+- Do any `.env` file values contain literal `$` signs in them? (This is common with PHP projects!)
+- Do any variable values use advanced expansion syntax? (e.g. `${VAR:?error}`)
+
+Run `docker compose config` on the project to preview the configuration after Compose V2 has performed interpolation to
+verify that values appear as expected.
+
+Maintaining backwards compatibility with Compose V1 is **typically** achievable by ensuring that literal values (no
```
aevesdocker, line 78:
```suggestion
Maintaining backwards compatibility with Compose V1 is typically achievable by ensuring that literal values (no
```

_includes/compose-eol.md
aevesdocker: This is the winner:
>Compose V1 no longer receives updates and will not be available in new releases of Docker Desktop after June 2023.

_includes/compose-eol.md
```diff
@@ -1,6 +1,6 @@
...
 >
-> Make sure you switch to [Compose V2](/compose/compose-file/) with the `docker compose` CLI plugin or by activating the **Use Docker Compose V2** setting in Docker Desktop. For more information, see the [Evolution of Compose](/compose/compose-v2/).
-{: .important}
\ No newline at end of file
+> Compose V2 is included with all currently supported versions of Docker Desktop.
+> For more information, see [Migrate to Compose V2](/compose/upgrade).
```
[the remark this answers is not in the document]
aevesdocker: Just bumping this again, can't see the change come through in the latest commits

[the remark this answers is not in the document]
milas: Ahhh I accidentally had it as part of the quote (with a `>`), which doesn't work apparently. Fixed - nice catch :)

milas:
# ship it!
squirrel giving a thumbs up

compose/upgrade.md
```diff
@@ -1,184 +1,133 @@
...
+Compose generates container names based on the project name, service name, and scale/replica count.
 
-```console
-$ docker compose cp my-service:~/path/to/myfile ~/local/path/to/copied/file
-```
+In Compose V1, an underscore (`_`) was used as the word separator.
+In Compose V2, a hyphen (`-`) is used as the word separator.
 
-We can also copy from the local filesystem to all the running containers of a service:
+Underscores are not valid characters in DNS hostnames.
```
jayaddison: @milas @aevesdocker this probably seems micro-nitpicky of me, but: underscores _are_ valid in DNS hostnames (unless there have been recent RFC updates I'm unaware of that alter that?).

Note: I'm not challenging Docker's choice to migrate to use hyphens instead -- that seems fine.  But I think it'd be worth ensuring that the documentation is accurate.

milas: _I no longer work at Docker, so I'm commenting purely in a personal capacity as an OSS developer_

I don't think this warrants a change on its own: quick search supports that underscore is **not** widely regarded as acceptable in DNS names. The separator change was made for better compatibility with this real world expectation.

jayaddison: To provide some missing background: while preparing to migrate a V1 `docker-compose.yml` file using the migration guide, I found some service names containing mid-name underscores (`foo_bar.invalid`) and began wondering about any network-level implications of renaming those.  The statement that underscores aren't valid in DNS names temporarily made me skip considering those implications (thinking: if they're not valid, they can't possibly be used in any connection-strings), until I remembered and checked later to find that they are valid (perhaps not recommendable, sure, but nonetheless valid).

We can probably chalk it up to my misunderstanding and hastiness.  In theory I'd like to reduce the possibility for other people to follow that same reasoning detour, and would happily offer a pull request, but I'm also OK if an edit for this is not considered worthwhile.

PR #17220, opened 2023-05-01
Add Fedora 38 to supported Fedora versions

AeroBliss:
### Proposed changes

Added Fedora 38 to supported versions since there is a repository for it.


dvdksn:
Thanks for contributing. This will be updated with #17233, so I'm closing this one

Issue #17192, opened 2023-04-26
Error docs empty environment variable will produce an empty string


Kibubu:
https://github.com/docker/docs/blob/45e5838f13346e4e2d5ee4969ffb95352ea51363/compose/environment-variables/set-environment-variables.md?plain=1#L112

While the sentence is correct the conclusion is wrong. The final value is `postgres:` without the period. Since a colon is provided but not tag is given docker will refuse to run this image.

This should be either fixed in the docs or inside compose.

aevesdocker:
Closing as this issue has now been resolved in the linked PR. Thanks again @Kibubu

PR #17224, opened 2023-05-02
Add info invalid references

Kibubu:
The example shows a configuration that would raise a invalid image reference error if the environment variable for the image is set to an empty string or is not set at all.

Also the code format ticks included the period.

#17192

A warning was added to highlight the incorrect configuration and the ticks now exclude the period.

### DIFF
--- a/compose/environment-variables/set-environment-variables.md
+++ b/compose/environment-variables/set-environment-variables.md
@@ -109,7 +109,11 @@ db:
 
 When you run `docker compose up` with this configuration, Compose looks for the `POSTGRES_VERSION` environment variable in the shell and substitutes its value in. For this example, Compose resolves the image to `postgres:9.3` before running the configuration.
 
-If an environment variable is not set, Compose substitutes with an empty string. In the example above, if `POSTGRES_VERSION` is not set, the value for the image option is `postgres:.`
+If an environment variable is not set, Compose substitutes with an empty string. In the example above, if `POSTGRES_VERSION` is not set, the value for the image option is `postgres:`.
+
+> **Important**
+>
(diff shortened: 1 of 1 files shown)


compose/environment-variables/set-environment-variables.md
```diff
@@ -109,7 +109,11 @@ db:
 
 When you run `docker compose up` with this configuration, Compose looks for the `POSTGRES_VERSION` environment variable in the shell and substitutes its value in. For this example, Compose resolves the image to `postgres:9.3` before running the configuration.
 
-If an environment variable is not set, Compose substitutes with an empty string. In the example above, if `POSTGRES_VERSION` is not set, the value for the image option is `postgres:.`
+If an environment variable is not set, Compose substitutes with an empty string. In the example above, if `POSTGRES_VERSION` is not set, the value for the image option is `postgres:`.
+
+> **Important**
```
aevesdocker, line 114:
```suggestion
> **Note**
```

compose/environment-variables/set-environment-variables.md
```diff
@@ -109,7 +109,11 @@ db:
...
+>
+> `postgres:` is **not** valid image reference. Docker expects either a reference without a tag like `postgres` which defaults to latest or with a tag `postgres:15`.
```
aevesdocker, line 116:
```suggestion
> `postgres:` is not a valid image reference. Docker expects either a reference without a tag, like `postgres` which defaults to the latest image, or with a tag such as `postgres:15`.
```

aevesdocker (approved):
Thank you @Kibubu

PR #17232, opened 2023-05-03
ENGDOCS-1331

aevesdocker:
As requested by support. 

(This whole page could do with a further clean up ***adds to to-do list***)

### DIFF
--- a/desktop/troubleshoot/topics.md
+++ b/desktop/troubleshoot/topics.md
@@ -288,7 +288,7 @@ You can also refer to the [Microsoft TechNet article](https://social.technet.mic
 
 #### Enable nested virtualization
 
-If you are using Hyper-V and you get the following error message:
+If you are using Hyper-V and you get the following error message when running Docker Desktop in a VDI environment:
 
 ```console
 The Virtual Machine Management Service failed to start the virtual machine 'DockerDesktopVM' because one of the Hyper-V components is not running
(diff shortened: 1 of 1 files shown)


desktop/troubleshoot/topics.md
```diff
@@ -261,16 +261,17 @@ To install Hyper-V manually, see [Install Hyper-V on Windows 10](https://msdn.mi
 From the start menu, type **Turn Windows features on or off** and press enter.
 In the subsequent screen, verify that Hyper-V is enabled.
 
-#### Virtualization must be enabled
+#### Nested virtualization must be enabled
```
dvdksn, line 264: I **think** the term nested virtualization only applies to the VDI context, right? Virtualization must be enabled is still correct, however if you're running DD in a VDI environment, you must also ensure that it supports nested virtualization, and that nested virtualization is enabled ([ex][1])

(I could be wrong here -- just thinking out loud, would be good with a 2nd opinion)

[1]: https://learn.microsoft.com/en-us/virtualization/hyper-v-on-windows/user-guide/nested-virtualization

aevesdocker: You are indeed correct. I went back to support to confirm as well. Have amended it accordingly - thank you for catching this. VM stuff makes my brain warm to the touch 🤯

dvdksn: very much the same for me!

dvdksn (approved):

PR #17233, opened 2023-05-03
engine: update versions in examples to use 23.0 (current release)

thaJeztah:

### DIFF
--- a/engine/api/index.md
+++ b/engine/api/index.md
@@ -32,7 +32,7 @@ Engine API), as well as SDKs for Go and Python. The SDKs allow you to build and
 scale Docker apps and solutions quickly and easily. If Go or Python don't work
 for you, you can use the Docker Engine API directly.
 
-For information about Docker Engine SDKs, see [Develop with Docker Engine SDKs](/engine/api/sdk/).
+For information about Docker Engine SDKs, see [Develop with Docker Engine SDKs](sdk/index.md).
 
 The Docker Engine API is a RESTful API accessed by an HTTP client such as `wget` or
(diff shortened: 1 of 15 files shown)


thaJeztah:
I tried to split the changes into logical commits (so please don't use "squash and merge"); I have a couple of follow-ups stashed as well, and will open a follow-up PR with those 😅

engine/api/index.md
```diff
@@ -119,10 +117,9 @@ You can specify the API version to use, in one of the following ways:
 - The Docker Go SDK allows you to enable API version negotiation, automatically
   selects an API version that is supported by both the client, and the Docker Engine
   that is used.
-
 - For the SDKs, you can also specify the API version programmatically, as a
   parameter to the `client` object. See the
-  [Go constructor](https://github.com/moby/moby/blob/v19.03.6/client/client.go#L119){: target="_blank" rel="noopener" class="_"}
+  [Go constructor](https://pkg.go.dev/github.com/docker/docker@v23.0.5+incompatible/client#NewClientWithOpts){: target="_blank" rel="noopener" class="_"}
```
dvdksn, line 122: Maybe we shouldn't pin the version here, it's gonna go out of date quickly

Instead we could point to:

https://pkg.go.dev/github.com/docker/docker/client#NewClientWithOpts

dvdksn (approved):

engine/api/index.md
thaJeztah: Ah, perhaps we can now, yes. Previously Go modules would pick `v1.13.1` as "latest", which may have been why we used a fixed version.

dvdksn (approved):
TY!

PR #17270, opened 2023-05-05
Proofread Compose GPU

bagratte:
Minor fixes in Compose GPU manual. (https://docs.docker.com/compose/gpu-support/)


ChrisChinchilla (approved):
Thanks @bagratte

PR #17271, opened 2023-05-05
Proofread Compose Extends Manual

bagratte:
Minor fixes in Compose Extends Manual. (http://docs.docker.com/compose/extends/)


ChrisChinchilla (approved):
Thanks @bagratte

PR #17283, opened 2023-05-08
Proofread Swarm

bagratte:
### Proposed changes

This is a proofreading of Swarm documentation.

### DIFF
--- a/engine/swarm/how-swarm-mode-works/services.md
+++ b/engine/swarm/how-swarm-mode-works/services.md
@@ -52,7 +52,7 @@ fails its health check or crashes, the orchestrator creates a new replica task
 that spawns a new container.
 
 A task is a one-directional mechanism. It progresses monotonically through a
-series of states: assigned, prepared, running, etc. If the task fails the
+series of states: assigned, prepared, running, etc. If the task fails, the
 orchestrator removes the task and its container and then creates a new task to
 replace it according to the desired state specified by the service.
 
@@ -60,7 +60,7 @@ The underlying logic of Docker swarm mode is a general purpose scheduler and
 orchestrator. The service and task abstractions themselves are unaware of the
 containers they implement. Hypothetically, you could implement other types of
 tasks such as virtual machine tasks or non-containerized process tasks. The
-scheduler and orchestrator are agnostic about the type of task. However, the
+scheduler and orchestrator are agnostic about the type of the task. However, the
 current version of Docker only supports container tasks.
 
 The diagram below shows how swarm mode accepts service create requests and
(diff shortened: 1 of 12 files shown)


engine/swarm/how-swarm-mode-works/services.md
```diff
@@ -109,9 +109,8 @@ serving the same content.
 A global service is a service that runs one task on every node. There is no
 pre-specified number of tasks. Each time you add a node to the swarm, the
 orchestrator creates a task and the scheduler assigns the task to the new node.
-Good candidates for global services are monitoring agents, an anti-virus
-scanners or other types of containers that you want to run on every node in the
-swarm.
+Good candidates for global services are monitoring agents, anti-virus  scanners
```
dvdksn, line 112: Double space
```suggestion
Good candidates for global services are monitoring agents, anti-virus scanners
```

engine/swarm/secrets.md
```diff
@@ -926,10 +926,8 @@ use it, then remove the old secret.
     ```
 
 
-7.  If you want to try the running all of these examples again or just want to
-    clean up after running through them, use these commands to remove the
-    WordPress service, the MySQL container, the `mydata` and `wpdata` volumes,
-    and the Docker secrets.
+7.  Run these commands to remove the WordPress service, the MySQL container,
+    the `mydata` and `wpdata` volumes, and the Docker secrets.
```
dvdksn, line 930:
```suggestion
7.  Run the following commands to remove the WordPress service, the MySQL container,
    the `mydata` and `wpdata` volumes, and the Docker secrets:
```

engine/swarm/services.md
```diff
@@ -893,7 +893,7 @@ $ docker service create --name=my_redis \
 ### Give a service access to volumes or bind mounts
 
 For best performance and portability, you should avoid writing important data
-directly into a container's writable layer, instead using data volumes or bind
+directly into a container's writable layer, using instead data volumes or bind
```
dvdksn, line 896:
```suggestion
directly into a container's writable layer, and instead use data volumes or bind
```

engine/swarm/services.md
```diff
@@ -920,7 +920,7 @@ $ docker service create \
   <IMAGE>
 ```
 
-If a volume with the same `<VOLUME-NAME>` does not exist when a task is
+If a volume with the name `<VOLUME-NAME>` does not exist when a task is
```
dvdksn, line 923:
```suggestion
If a volume with the name `<VOLUME-NAME>` doesn't exist when a task is
```

dvdksn (changes requested):
Thank you so much for doing this! ❤️ I found a couple nits while reading through the changes.

engine/swarm/services.md
bagratte: I like "using instead" better :)
Alternatively, we could split this into two sentences. "...you should avoid writing...layer. Instead, you should use..."
I leave this one to you to decide.

dvdksn: Splitting into two sentences SGTM!

dvdksn (approved):
LGTM

PR #17302, opened 2023-05-11
Add multiple device deployment to registry.json

craig-osterhout:
### Proposed changes

Users have asked how they can use Docker to deploy registry.json to multiple devices. Added a section to inform users that they must use a third-party solution.

Added toc_max to frontmatter because the tab sections don't work well with the toc.

### Related issues (optional)

ENGDOCS-1339


aevesdocker (approved):
Thanks @craig-osterhout !

PR #17306, opened 2023-05-12
Incorrect docker image used registry hasn't htpasswd

pavljiks:
Incorrect docker image used. (unable to start container process: exec: "htpasswd": executable file not found in $PATH: unknown.)

### Proposed changes

replace **registry** docker image with **httpd**


dvdksn (approved):
Thx!

PR #17313, opened 2023-05-14
Update "Ways to set environment variables in Compose"

RafaelWO:
### Proposed changes
* Add information that if an environment variable is declared by a single key (no value to equals sign), no warning will be printed if the variable is not set in the shell environment.
* Add an example of setting variables explicitly to get a warning from Compose if variables are not set. 

### Related issues (optional)

Relates to https://github.com/docker/compose/issues/10551


aevesdocker (approved):
Thank you for your contribution @RafaelWO!

PR #17317, opened 2023-05-15
Use Debian instructions for 64-bit Raspbian

### LINKED ISSUE #17311: Raspbian Docker Installation Documentation Does Not Work
When using the latest Raspberry Pi OS (64-bit) *Raspbian*, following the dedicated webpage for installing docker on raspbian does not work. I ran the following command as stated **exactly** by the documentation:

Here are the commands in chronological order:

```sh
 sudo apt-get update
 sudo apt-get install ca-certificates curl gnupg
```

```sh
 curl -fsSL https://download.docker.com/linux/raspbian/gpg | sudo gpg --dearmor -o /etc/apt/keyrings/docker.gpg
 sudo chmod a+r /etc/apt/keyrings/docker.gpg
```

```sh
echo \
  "deb [arch="$(dpkg --print-architecture)" signed-by=/etc/apt/keyrings/docker.gpg] https://download.docker.com/linux/raspbian \
  "$(. /etc/os-release && echo "$VERSION_CODENAME")" stable" | \
```

```sh
sudo apt-get update
sudo apt-get install docker-ce docker-ce-cli containerd.io docker-buildx-plugin docker-compose-plugin
```

When ran on an Raspberry Pi 4 Model B with latest operating system (literally flashed on May 11, 2023), I get the following error as exactly stated:

```sh
Package docker-ce is not available, but is referred to by another package.
This may mean that the package is missing, has been obsoleted, or
is only available from another source

Package docker-ce-cli is not available, but is referred to by another package.
This may mean that the package is missing, has been obsoleted, or
is only available from another source

E: Package 'docker-ce' has no installation candidate
E: Package 'docker-ce-cli' has no installation candidate
E: Unable to locate package containerd.io
E: Couldn't find any package by glob 'containerd.io'
E: Couldn't find any package by regex 'containerd.io'
E: Unable to locate package docker-buildx-plugin
E: Unable to locate package docker-compose-plugin
```

### Problem location

- I saw a problem at the following URL: [https://docs.docker.com/engine/install/raspbian/](https://docs.docker.com/engine/install/raspbian/)

StefanScherer:
### Proposed changes

With the updated [installation instructions for Raspbian](https://docs.docker.com/engine/install/raspbian/) there is a problem for users of the 64-bit version Raspbian. We don't ship special 64-bit Raspbian packages, only the 32-bit because of special differences for armhf. The 64-bit version are compatible with the normal Debian packages.

I propose this small change to guide the readers over to the instruction page for Debian when they run a 64-bit version of Raspbian.

### Related issues (optional)

Closes #17311

### DIFF
--- a/engine/install/raspbian.md
+++ b/engine/install/raspbian.md
@@ -16,7 +16,7 @@ To get started with Docker Engine on Raspbian, make sure you
 
 ### OS requirements
 
-To install Docker Engine, you need the 64-bit version of one of these Raspbian
+To install Docker Engine, you need the 64-bit version or 32-bit version of one of these Raspbian
 versions:
 
 - Raspbian Bookworm 12 (testing)
@@ -24,6 +24,8 @@ versions:
 - Raspbian Buster 10 (oldstable)
 
 Docker Engine for Raspbian is compatible with the armhf architecture.
+
+For the 64-bit version of Raspbian follow the instructions for [Debian](debian.md).
 
 ### Uninstall old versions
 


dvdksn (approved):
Thank you!

PR #17319, opened 2023-05-15
Add mention of arch package but why steps are recommended instead

ChrisChinchilla:
### Proposed changes

A popular arch package exists but we don't mention it, as noted in https://github.com/docker/docs/issues/16913

This PR explains the differences between it and the steps we recommend.

### DIFF
--- a/desktop/install/archlinux.md
+++ b/desktop/install/archlinux.md
@@ -6,7 +6,7 @@ redirect_from:
   - /desktop/linux/install/archlinux/
 ---
 
-This topic discusses installation of Docker Desktop from an [Arch package](https://desktop-stage.docker.com/linux/main/amd64/docker-desktop-4.19.0-x86_64.pkg.tar.zst?utm_source=docker&utm_medium=webreferral&utm_campaign=docs-driven-download-linux-amd64) that Docker provides in addition to the supported platforms. Docker has not tested or verified the installation.
+This topic discusses installation of Docker Desktop from an [Arch package](https://desktop-stage.docker.com/linux/main/amd64/docker-desktop-4.19.0-x86_64.pkg.tar.zst?utm_source=docker&utm_medium=webreferral&utm_campaign=docs-driven-download-linux-amd64) that Docker provides in addition to the supported platforms. Docker hasn't tested or verified the installation.
 
 [Arch package (experimental)](https://desktop.docker.com/linux/main/amd64/docker-desktop-4.19.0-x86_64.pkg.tar.zst?utm_source=docker&utm_medium=webreferral&utm_campaign=docs-driven-download-linux-amd64){: .button .primary-btn }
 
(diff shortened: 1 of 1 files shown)


aevesdocker (approved):

desktop/install/archlinux.md
```diff
@@ -6,7 +6,7 @@ redirect_from:
   - /desktop/linux/install/archlinux/
 ---
 
-This topic discusses installation of Docker Desktop from an [Arch package](https://desktop-stage.docker.com/linux/main/amd64/docker-desktop-4.19.0-x86_64.pkg.tar.zst?utm_source=docker&utm_medium=webreferral&utm_campaign=docs-driven-download-linux-amd64) that Docker provides in addition to the supported platforms. Docker has not tested or verified the installation.
+This topic discusses installation of Docker Desktop from an [Arch package](https://desktop-stage.docker.com/linux/main/amd64/docker-desktop-4.19.0-x86_64.pkg.tar.zst?utm_source=docker&utm_medium=webreferral&utm_campaign=docs-driven-download-linux-amd64) that Docker provides in addition to the supported platforms. Docker hasn't tested or verified the installation.
```
thaJeztah, line 9: Do we know why this is referring to the "desktop-stage" domain? Looks like that probably shouldn't be documented in our docs 🤔

mat007: Yeah, it’s the stage bucket, URLs have a 30 days decay time there…
We should use https://desktop.docker.com/linux/main/amd64/106363/docker-desktop-4.19.0-x86_64.pkg.tar.zst instead.

ChrisChinchilla:
@dvdksn There's a small amount of context in the issue, but maybe it's more that using something like packman on Arch is **the** way to install something, whereas Homebrew is **a** way…? I am open to feedback though, but I get the impression that lots of people use it anyway, so might be worth calling out what it will do if you do. I also wonder if in the long run, maybe we should mention some of the more popular community methods. For example, I use nothing but the Homebrew formula 🤷

dvdksn (commented):
I agree that it would be nice to have a page listing unofficial options for installing Docker, I suggest that we do that instead. Linking off to AUR here seems to me like too much of an endorsement. Also, presumably that AUR package is for the daemon + cli, and deps, not Docker Desktop.

ChrisChinchilla:
Alright, then let's do that @dvdksn . Will close this, tell the original issue creator, and add to backlog.

desktop/install/archlinux.md
usha-mandya: @aevesdocker Could you update the URL?

qumberrizvi:
For future reference, the package seems to have moved from `Community` to `Extra` repository.
The older link to the package in the original issue now returns 404.
The package is now located here - https://archlinux.org/packages/extra/x86_64/docker/

PR #17326, opened 2023-05-16
Update container-networking.md for consistency

erfantkerfan:
ports inside the container are not quoted so it looks like an inconsistency


erfantkerfan:
for clarity, it is just an inconsistency in the docs; some ports are inside a quote and some are not.
I thought it would matter to some people like myself so I fixed it.

erfantkerfan:
@craig-osterhout

PR #17341, opened 2023-05-17
fix helm oci registry push docs

ishankhare07:
### Proposed changes

Fix the command for helm oci upload to match with the username in the example


craig-osterhout:
Thanks for creating the PR. The command appears to be correct.

While the user could push to their own namespace, the rest of the tutorial assumes that they pushed to the docker org namespace. The sentence right below the command explains which namespace it was pushed to. If it's unclear, it's probably better to update that sentence rather than changing the command, which affects the following steps.

PR #17350, opened 2023-05-17
Update section to create an org

craig-osterhout:
### Proposed changes

Large enterprises may contact sales directly and pay for an organization before it has been created. These users need to create an org without paying again.

The self-serve UI can be unintuitive in this scenario, and it causes users to think they will be billed again. Specifically the "Buy Now" button and the payment screen. 
Talks to update the UI are already in motion. Updating the docs as a quick fix.

- Added notes to the steps for the Buy Now button and payment screen.
- Cleaned up a few styles in the section.

### Related issues (optional)
ENGDOCS-1363


stephaurelio (approved):
This is much clearer! LGTM 🚀

PR #17356, opened 2023-05-18

wushu037:
After repeated verification, I think we should add this note

### DIFF
--- a/network/network-tutorial-standalone.md
+++ b/network/network-tutorial-standalone.md
@@ -436,6 +436,10 @@ connected to both networks.
     address. This capability is called **automatic service discovery**. Let's
     connect to `alpine1` and test this out. `alpine1` should be able to resolve
     `alpine2` and `alpine4` (and `alpine1`, itself) to IP addresses.
+    
+    > **Note**
+    > 
+    > Only valid for custom container names, the default generated container names cannot be resolved
 
     ```console
     $ docker container attach alpine1
--- a/network/network-tutorial-standalone.md
+++ b/network/network-tutorial-standalone.md
@@ -439,7 +439,7 @@ connected to both networks.
     
     > **Note**
     > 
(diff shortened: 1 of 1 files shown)


network/network-tutorial-standalone.md
```diff
@@ -436,6 +436,10 @@ connected to both networks.
     address. This capability is called **automatic service discovery**. Let's
     connect to `alpine1` and test this out. `alpine1` should be able to resolve
     `alpine2` and `alpine4` (and `alpine1`, itself) to IP addresses.
+    
+    > **Note**
+    > 
+    > Only valid for custom container names, the default generated container names cannot be resolved
```
ChrisChinchilla, line 442: Thanks @wushu037 but I think the paragraph above already covers this?

> On user-defined networks like alpine-net, containers can not only communicate by IP address, but can also resolve a container name to an IP address. This capability is called automatic service discovery. Let's connect to alpine1 and test this out. alpine1 should be able to resolve alpine2 and alpine4 (and alpine1, itself) to IP addresses.

But let me know if you think that "user-define" part needs clarifying and maybe we could move the note into that line instead?

wushu037: Thank you for your reply.

I don't think the paragraph you mentioned covers this, it only highlights "user-defined networks" and I added "user-defined container names".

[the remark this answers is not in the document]
wushu037: @ChrisChinchilla , Thank you for asking me what I think.

It would be nice if the second half of the sentence emphasized that the name of the container is "automatically generated" or "generated by default". It would be more straightforward and would not require the reader to spend a few more seconds thinking "Generated? Does that mean automatically generated?"

That's what I thought.

ChrisChinchilla (approved):

ChrisChinchilla:
Thanks for the contribution @wushu037 !

PR #17367, opened 2023-05-19
Clarify email domains and subdomains on SSO config page

stephaurelio:
### Proposed changes

This PR addresses a question that came from a customer that needed clarity and an example of what the SSO docs mean when they tell users to include email domains and subdomains. In this PR, I added an example domain and subdomain. I also opted to style this as inline code to be consistent with domain examples in [this docs page](https://docs.docker.com/config/daemon/systemd/).

Page updated:
- `single-sign-on/configure/index.md`

### DIFF
--- a/single-sign-on/configure/index.md
+++ b/single-sign-on/configure/index.md
@@ -18,7 +18,7 @@ Follow the steps on this page to configure SSO for your organization or company.
 
     >**Note**
     >
-    > Format your domains without protocol or www information, for example, `yourcompany.com`. This should include all email domains and subdomains users will use to access Docker, for example `user@yourcompany.com` and `user@us.yourcompany.com`. Public domains such as `gmail.com`, `outlook.com`, etc. aren’t permitted. Also, the email domain should be set as the primary email.
+    > Format your domains without protocol or www information, for example, `yourcompany.com`. This should include all email domains and subdomains users will use to access Docker, for example `yourcompany.com` and `us.yourcompany.com`. Public domains such as `gmail.com`, `outlook.com`, etc. aren’t permitted. Also, the email domain should be set as the primary email.
 
 4. Once you have waited 72 hours for the TXT Record verification, you can then select **Verify** next to the domain you've added, and follow the on-screen instructions. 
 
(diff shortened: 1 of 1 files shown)


single-sign-on/configure/index.md
```diff
@@ -18,7 +18,7 @@ Follow the steps on this page to configure SSO for your organization or company.
 
     >**Note**
     >
-    > Format your domains without protocol or www information, for example, yourcompany.com. This should include all email domains and subdomains users will use to access Docker. Public domains such as gmail.com, outlook.com, etc. aren’t permitted. Also, the email domain should be set as the primary email.
+    > Format your domains without protocol or www information, for example, `yourcompany.com`. This should include all email domains and subdomains users will use to access Docker, for example `yourcompany.com` and `us.yourcompany.com`. Public domains such as `gmail.com`, `outlook.com`, etc. aren’t permitted. Also, the email domain should be set as the primary email.
```
thaJeztah, line 21: It's best to use designated example domains for this (see https://www.rfc-editor.org/rfc/rfc2606.html#section-2), as domains like `yourcompany.com` can be actual domains owned by someone.

For example;

- `yourcompany.example.com`
- `company.example.com`
- `yourcompany.example`

thaJeztah (changes requested):

single-sign-on/configure/index.md
stephaurelio: Sure, I can make that update.

thaJeztah: Thanks!

I keep; forgetting to look at adding that to the docs guidelines 😅. In _most_ cases it's not problematic, but we've had some cases where we used an existing domain in the examples, and users (being users) copy/pasted the commands as-is (which isn't, ahem, "ideal" 😅

thaJeztah (approved):
domain changes LGTM, thanks!

single-sign-on/configure/index.md
stephaurelio: That makes sense!

ChrisChinchilla: I can add this to our guide and style checkers at some point

aevesdocker: If you have the time @ChrisChinchilla, that'd be great

ChrisChinchilla: Yeah was just talking with @thaJeztah about it :)

PR #17388, opened 2023-05-22
reference: prevent options/flags column from wrapping

thaJeztah:
Make sure that options (flags) are not wrapped in the table.

Before:

Screenshot 2023-05-22 at 23 18 46

After:

Screenshot 2023-05-22 at 23 18 05


thaJeztah:
ISTR we had a CSS rule for this (prevent any inline `code` from being wrapped), but perhaps that was removed at some point. In either case; let's at least prevent these from being wrapped in the tables.

thaJeztah:
@dvdksn @crazy-max PTAL

thaJeztah:
We should also make these vertically align "top" 🤔

dvdksn (approved):
LGTM

PR #17390, opened 2023-05-23
Engdocs 1374

aevesdocker:
Fixes https://github.com/docker/docs/issues/12021 and https://github.com/docker/docs/issues/12214 - there was never a version 3.9 

Also removes the 'version' attribute from any and all examples as this is no longer required


engine/swarm/stack-deploy.md
```diff
@@ -116,9 +116,7 @@ counter whenever you visit it.
 5.  Create a file called `docker-compose.yml` and paste this in:
 
     ```none
-    version: "{{ site.compose_file_v3 }}"
-
-    services:
+        services:
```
dvdksn, line 119: Indentation seems off for this key - I think it should hug left. But I noticed this page isn't even in the TOC... I'll check why...

```suggestion
    services:
```

dvdksn (approved):
LGTM!

PR #17401, opened 2023-05-24
Make Scout compare command consistent with release cycle and other changes

ChrisChinchilla:
### Proposed changes

Make Scout compare command consistent with release cycle and other changes

### DIFF
--- a/_data/scout-cli/docker_scout_compare.yaml
+++ b/_data/scout-cli/docker_scout_compare.yaml
@@ -1,9 +1,9 @@
 command: docker scout compare
-short: '[early preview] Compare two images and display differences'
+short: '[experimental] Compare two images and display differences'
 long: |-
     The `docker scout compare` command analyzes two images and displays a comparison of both.
 
-    > This command is in **early preview** and its behaviour might change in the future
+    > This command is **experimental** and its behaviour might change in the future
 
     The main usage is to compare two versions of the same image.
     For instance when a new image is built and compared to the version running in production.


ChrisChinchilla:
@eunomie I notice bringing this in updates other cli options, are they released?

eunomie:
They are released and available yes.
But some changes might not be part of Desktop, but not sure how we can easily make it visible (if we want it)

_data/scout-cli/docker_scout_compare.yaml
```diff
@@ -1,13 +1,16 @@
 command: docker scout compare
-short: '[early preview] Compare two images and display differences'
```
usha-mandya, line 2: @ChrisChinchilla Another reference to early preview

ChrisChinchilla:
@usha-mandya Yup, that's what we did

ChrisChinchilla:
@eunomie I am not sure we have any way to differentiate that unless there's some way we highlight it with flags / a support matrix currently? Maybe @thaJeztah or @dvdksn might now

thaJeztah:
With "Docker Desktop" being our main distribution "vehicle", we sometimes decided to hold back documentation for updated components until they shipped with Desktop. (e.g. if Desktop ships with v1.0.0, then document that version, even if there is a `v1.1.0` available).

That was mostly with the expectation that these situations would "resolve" themselves within a reasonable timeframe.

Another solution _could_ be to add notes about what version specific flags were added in, but we tried to avoid that in most cases, as that may clutter the documentation considerably (and making it hard to navigate the docs" oh, this feature requires X, and this requires Y").

We may still think a bit about that option, but perhaps in a different format/presentation ("badges" e.g., similar to what we did with some flags on the engine that depend on certain versions of the Engine API).

thaJeztah:
e.g.; 

Screenshot 2023-05-24 at 15 39 14


But that presentation needs some work as well (most users won't know how to correlate API version with engine version)

dvdksn:
I like the idea of badges. I have in some cases [used notes](https://docs.docker.com/build/attestations/) for explicitly stating the release version, but we often do not do it at all and lean on release notes.

We should try to make it more obvious, without being too noisy. I quite like [how GitLab does it](https://docs.gitlab.com/runner/executors/docker.html#use-windows-containers), with a little note underneath the heading.

ChrisChinchilla:
And identifying flag and option support I think is something we can address later.

dvdksn:
Yes, we should bisect the immediate and future changes imo

ChrisChinchilla:
@dvdksn Done, I feel like the other small changes here are fine though and kind are part of the same change I made.

thaJeztah:
> I like the idea of badges. I have in some cases [used notes](https://docs.docker.com/build/attestations/) for explicitly stating the release version, but we often do not do it at all and lean on release notes.
> 
> We should try to make it more obvious, without being too noisy. I quite like [how GitLab does it](https://docs.gitlab.com/runner/executors/docker.html#use-windows-containers), with a little note underneath the heading.

Our main concern with including versions was that it becomes irrelevant at some point, after which it is _really_ only noise. e.g., consider a Microsoft Windows manual where every command has a note "this command requires at least 640KB of memory and a Pentium 2 processor". We've had similar cases where docs were still mentioning "Starting with Docker 1.10.. " "In versions of Docker prior to 17.03" etc; those versions reached EOL long time ago. When including it as part of the actual docs / text, this adds additional overhead to maintain. Annotations / metadata that can be used to _generate_ information about can go a longer way for that, because we can decide what to do with the information when generating the documentation.

For example (yes, it' hacky), I added a "threshold" configuration for the API version badges that I mentioned above; https://github.com/docker/docs/blob/7ecfd29e5769a4dab7534c82f816c64e9d237a15/_config.yml#L60-L77


That configuration allows us to hide API version requirement badges for obsolete engine versions (if your engine doesn't meet this minimum API requirement, you're running an EOL version, so you likely have a much bigger problem); information is still there, and (API) changelog for those that need it. We also have similar annotations for features/flags that are only supported on (say) Windows, or require BuildKit (etc) (last one I need to check, because we may have removed the actual options in 23.0)

Long time ago, when CSS was still cool, and all docs were still maintained in the "docker/docker" repository I suggested we could add more of such annotations (also in the docs themselves, not just "generated"), and use the information more dynamically as well; 

- optionally show all API version badges when viewing the page (if you need the full history)
- configure what the "latest" version is, and if `<some option> / <some feature>` was added in that version, then render a "NEW!" badge
- highlight / summarise an overview of all pages that contain content with `<NEW!>` content
- and ... once it's no longer "new!" (version introduced is < latest release), we can only render it as "added in version X", and hide it once it's no longer relevant to mention.

If we want to do that, we should;

- for generated docs: define annotations (we defined some in https://github.com/docker/cli-docs-tool/blob/main/annotation/annotation.go, but need to thing what's the best approach), and a format to use.
- similar for "in-text" annotations
- brainstorm a bit how we can best consume the information.

dvdksn:
Thanks for the great summary @thaJeztah. Let me schedule a call to discuss this further, I think it'd be a nice enhancement to our docs.

ChrisChinchilla:
@dvdksn It's become too hard to follow the "correct" process for generating the docs in the docs repo from the CLI docs in this case due to the inconsistent releasing, so for now I am just adding what we actually want changed manually and we can address the functionality changes separately.

dvdksn (approved):
LGTM - please squash!

dvdksn (approved):

PR #17410, opened 2023-05-25
Tweak accordion CSS

ChrisChinchilla:
### Proposed changes

This PR makes an attempt to tweak the CSS for the Accordion component to add a bit more padding.

### DIFF
--- a/_scss/_overrides.scss
+++ b/_scss/_overrides.scss
@@ -105,6 +105,14 @@
     background-color: #fafdfe;
 }
 
+.collapse {
+    padding: 10px 15px;
+}
+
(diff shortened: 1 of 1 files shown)


aevesdocker:
I'd be tempted not to merge this as it does look quite buggy 🤔 given Hugo is right around the corner, is it worth spending the time trying to fix this?

ChrisChinchilla:
@aevesdocker I was thinking the same tbh, but was also worried that when I ask people to look at the PR that caused me to create this one, all they'll focus on is the lack of padding… We'll see I guess and I can mention there's a fix in progress.

PR #17416, opened 2023-05-26
engine: add release notes for v24.0.2

thaJeztah:

### DIFF
--- a/engine/release-notes/24.0.md
+++ b/engine/release-notes/24.0.md
@@ -19,6 +19,30 @@ For more information about:
 - Deprecated and removed features, see [Deprecated Engine Features](../deprecated.md).
 - Changes to the Engine API, see [Engine API version history](../api/version-history.md).
 
+## 24.0.2
+
+{% include release-date.html date="2023-05-26" %}
+
+For a full list of pull requests and changes in this release, refer to the relevant GitHub milestones:
+
+- [docker/cli, 24.0.2 milestone](https://github.com/docker/cli/issues?q=is%3Aclosed+milestone%3A24.0.2)
+- [moby/moby, 24.0.2 milestone](https://github.com/moby/moby/issues?q=is%3Aclosed+milestone%3A24.0.2)
+
+### Bug fixes and enhancements
+
+* Fix a panic during build when referencing locally tagged images. [moby/buildkit#3899](https://github.com/moby/buildkit/pull/3899), [moby/moby#45582](https://github.com/moby/moby/pull/45582)
+* Fix builds potentially failing with `exit code: 4294967295` when performing many concurrent build stages. [moby/moby#45620](https://github.com/moby/moby/pull/45620)
+* Fix DNS resolution on Windows ignoring `etc/hosts` (`%WINDIR%\System32\Drivers\etc\hosts`), including resolution of `localhost`. [moby/moby#45562](https://github.com/moby/moby/pull/45562)
+* Apply a workaround for a containerd bug that causes concurrent `docker exec` commands to take significantly longer than expected. [moby/moby#45625](https://github.com/moby/moby/pull/45625)
+* containerd image store: Fix an issue where the image `Created` field would contain an incorrect value. [moby/moby#45623](https://github.com/moby/moby/pull/45623)
+* containerd image store: Adjust the output of image pull progress so that the output has the same format regardless of whether the containerd image store is enabled. [moby/moby#45602](https://github.com/moby/moby/pull/45602)
+* containerd image store: Switching between the default and containerd image store now requires a daemon restart. [moby/moby#45616](https://github.com/moby/moby/pull/45616)
+
+### Packaging updates
+
+* Upgrade Buildx to `v0.10.5`. [docker/docker-ce-packaging#900](hhttps://github.com/docker/docker-ce-packaging/pull/900)
+
+
 ## 24.0.1
 
 {% include release-date.html date="2023-05-19" %}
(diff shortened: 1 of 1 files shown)


thaJeztah:
Looks like the linter has an off-by-one for the annotation, or are they supposed to be "one line above the issue?" 😂 

Screenshot 2023-05-26 at 09 54 35

engine/release-notes/24.0.md
```diff
@@ -19,6 +19,34 @@ For more information about:
...
+
+For a full list of pull requests and changes in this release, refer to the relevant GitHub milestones:
+
+- [docker/cli, 24.0.2 milestone](https://github.com/docker/cli/issues?q=is%3Aclosed+milestone%3A24.0.2)
+- [moby/moby, 24.0.2 milestone](https://github.com/moby/moby/issues?q=is%3Aclosed+milestone%3A24.0.2)
+
+### Bug fixes and enhancements
+
+* Fix a panic when building Dockerfiles that used locally built images. [moby/moby#45582](https://github.com/moby/moby/pull/45582)
+* Fix `docker build` hanging when running many concurrent builds. [moby/moby#45620](https://github.com/moby/moby/pull/45620)
```
neersighted, line 34:
```suggestion
* Fix builds potentially failing with `exit code: 4294967295` when performing **many** concurrent build stages. [moby/moby#45620](https://github.com/moby/moby/pull/45620)
```

engine/release-notes/24.0.md
```diff
@@ -19,6 +19,34 @@ For more information about:
...
+* Fix DNS resolution on Windows ignoring `etc/hosts` (`%WINDIR%\System32\Drivers\etc\hosts`),
+  including resolution of `localhost`. This was caused by a bug in Go's `netgo`
+  implementation, which was enabled in Go 1.19, but does not yet implement file-based
+  resolution. We disabled this feature in the Windows builds, now using the Windows
+  host resolver instead. [moby/moby#45562](https://github.com/moby/moby/pull/45562)
```
neersighted, line 39:
```suggestion
* Fix DNS resolution on Windows ignoring `etc/hosts` (`%WINDIR%\System32\Drivers\etc\hosts`),
  including resolution of `localhost`. [moby/moby#45562](https://github.com/moby/moby/pull/45562)
```

I don't think we need to go much more long-winded than this.

engine/release-notes/24.0.md
```diff
@@ -19,6 +19,34 @@ For more information about:
...
+* Apply a workaround for a containerd bug that causes concurrent `docker exec` commands to take significantly longer than expected. [moby/moby#45625](https://github.com/moby/moby/pull/45625)
+* containerd-storage: Fix an issue where the image `Created` field would contain an incorrect value. [moby/moby#45623](https://github.com/moby/moby/pull/45623)
+* containerd-storage: Adjust the output of image pull progress to match the output without the containerd-storage enabled. [moby/moby#45602](https://github.com/moby/moby/pull/45602)
+* containerd-storage: Prevent switching between containerd-storage and non-containerd-storage without restarting the daemon. [moby/moby#45616](https://github.com/moby/moby/pull/45616)
```
neersighted, line 43: It looks like we actually called the feature-flag `containerd-snapshotter` -- I think we should update all the release notes for uniformity/to match.

neersighted (changes requested):

engine/release-notes/24.0.md
dvdksn: Nit: we shouldn't use bold here.

engine/release-notes/24.0.md
thaJeztah: I changed it, but then changed it back; I guess it's a bit of a combo. Maybe "snapshotter" is too much implementation specific, and the preferences mention "containerd image storage", so probably fine to leave it (for now)
Screenshot 2023-05-26 at 10 45 00

engine/release-notes/24.0.md
thaJeztah: ah, removed bold 👍

engine/release-notes/24.0.md
dvdksn: IMO this should say `containerd image store:` instead of referring to the configuration flag or `containerd-storage`

thaJeztah: works for me; let me update

thaJeztah: Updated, and slightly rephrased the last bullet (the one for switching the feature)

rumpl (dismissed):

neersighted (dismissed):

engine/release-notes/24.0.md
```diff
@@ -19,6 +19,30 @@ For more information about:
...
+* Fix builds potentially failing with `exit code: 4294967295` when performing many concurrent build stages. [moby/moby#45620](https://github.com/moby/moby/pull/45620)
+* Fix DNS resolution on Windows ignoring `etc/hosts` (`%WINDIR%\System32\Drivers\etc\hosts`), including resolution of `localhost`. [moby/moby#45562](https://github.com/moby/moby/pull/45562)
+* Apply a workaround for a containerd bug that causes concurrent `docker exec` commands to take significantly longer than expected. [moby/moby#45625](https://github.com/moby/moby/pull/45625)
+* containerd image store: Fix an issue where the image `Created` field would contain an incorrect value. [moby/moby#45623](https://github.com/moby/moby/pull/45623)
+* containerd image store: Adjust the output of image pull progress to match the output without the containerd image store enabled. [moby/moby#45602](https://github.com/moby/moby/pull/45602)
+* containerd image store: Prevent switching the containerd image store feature without restarting the daemon. [moby/moby#45616](https://github.com/moby/moby/pull/45616)
```
dvdksn, line 39:
```suggestion
* containerd image store: Fix an issue where the image `Created` field would contain an incorrect value. [moby/moby#45623](https://github.com/moby/moby/pull/45623)
* containerd image store: Adjust the output of image pull progress so that the output has the same format regardless of whether the containerd image store is enabled. [moby/moby#45602](https://github.com/moby/moby/pull/45602)
* containerd image store: Make the daemon restart when switching between the containerd image store and the default image store. [moby/moby#45616](https://github.com/moby/moby/pull/45616)
```

thaJeztah: Looking at the last bullet: "make the daemon restart" is probably not correct. The fix is that changing the option does not take effect until restarting the daemon. Not sure what the best wording is for that though; perhaps you have suggestions.

dvdksn: aha, I misunderstood then. How about; "Switching between the default and containerd image store now requires a daemon restart"?

thaJeztah: Ah, yes, that works

thaJeztah: Updated 👍

dvdksn (approved):

PR #17421, opened 2023-05-26
Change DSO link to Scout Web App

ChrisChinchilla:
### Proposed changes

Now the Scout Web App is live, we can change the link from DSO to it.


stephaurelio (approved):
LGTM 🚢

PR #17432, opened 2023-05-30
Add version note

ChrisChinchilla:
### Proposed changes

Adds note to Scout Web App page about version calculation on base image.


aevesdocker (approved):

scout/dashboard.md
```diff
@@ -29,6 +29,11 @@ Each entry in the list shows the following details:
 - The most recent version of the image and the vulnerabilities for that version. Clicking the link for the base image opens [the image layer view](#image-layer-view).
 - The operating system and architecture of the image.
 - The base image and version used by the repository and the vulnerabilities for that version. Clicking the link for the base image opens [the image layer view](#image-layer-view).
+
+  > **Note**
+  >
+  > Docker Scout detects the base image used by a repository by using the digest hash. However, if the hash is linked to different tags, Docker Scout can't show the precise version.
```
dvdksn, line 35: ?

```suggestion
  > Docker Scout detects the base image used by a repository by using the digest hash. However, if the hash is associated with more than one tag, Docker Scout can't show the precise version.
```

eunomie (approved):
Sounds good, just quick comment 
Also, wondering if we shouldn't also add a note about provenance that might help to fix that (when we have provenance attestation then we know exactly which image/tag)
And while writing, just thinking it's more than tag. Sometimes two images have the same digest (same image) in two different repository so we don't know which one it is.

scout/dashboard.md
ChrisChinchilla: @dvdksn Thanks, I was stuck on that particular word for a while, perfect.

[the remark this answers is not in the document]
ChrisChinchilla: @eunomie Fair point, I think at the time I was trying to differentiate two different applications of the term.

scout/dashboard.md
```diff
@@ -29,6 +29,13 @@ Each entry in the list shows the following details:
 - The most recent version of the image and the vulnerabilities for that version. Clicking the link for the base image opens [the image layer view](#image-layer-view).
 - The operating system and architecture of the image.
 - The base image and version used by the repository and the vulnerabilities for that version. Clicking the link for the base image opens [the image layer view](#image-layer-view).
+
+  > **Note**
+  >
+  > Docker Scout detects the base image of an image used in a repository by matching layer `diff_ids`.
+  > However, it's possible for different images and tags to use these layers. In these cases, Docker Scout's
```
mark-dr, line 36:
```suggestion
  > However, it's possible for multiple tags to point at the same layers. In these cases, Docker Scout's
```

scout/dashboard.md
```diff
@@ -29,6 +29,13 @@ Each entry in the list shows the following details:
 - The most recent version of the image and the vulnerabilities for that version. Clicking the link for the base image opens [the image layer view](#image-layer-view).
 - The operating system and architecture of the image.
 - The base image and version used by the repository and the vulnerabilities for that version. Clicking the link for the base image opens [the image layer view](#image-layer-view).
+
+  > **Note**
+  >
+  > Docker Scout detects the base image of an image used in a repository by matching layer `diff_ids`.
```
mark-dr, line 35: I feel like diff IDs are a bit of an implementation detail - we should obscure that complexity if possible. Maybe just this?

(also suggested removing "in a repository", since this is true of exclusively local images as well)

```suggestion
  > Docker Scout detects the base image of an image by matching layer IDs.
```

scout/dashboard.md
```diff
@@ -29,6 +29,13 @@ Each entry in the list shows the following details:
...
+  > However, it's possible for different images and tags to use these layers. In these cases, Docker Scout's
+  > base image detection might not be precise and potentially return a different tag to the one used in the Dockerfile.
```
mark-dr, line 37: Maybe it's just me, but my eyes stumbled a bit over parsing the scope of the _not_ when I first read this. The first half of the conjunction is essentially a subset of the second half, so we could just skip it?

```suggestion
  > base image detection could return a different tag to the one used in the Dockerfile.
```

mark-dr (changes requested):

scout/dashboard.md
ChrisChinchilla: I took bits of this into a change

scout/dashboard.md
ChrisChinchilla: I like this suggestion. @cdupuis Is this enough clarity/accuracy compromise?

dvdksn (commented):
Can you make sure you squash these commits? thanks

scout/dashboard.md
cdupuis: Don't think this is technically really correct but it shouldn't matter much. As a reader I wonder what a `layer ID` is. What about just:

```suggestion
  > Docker Scout detects the base image of an image by matching layer content hashes.
```

cdupuis (approved):

dvdksn (approved):

PR #17435, opened 2023-05-30
Reorganise trusted content pages pages

ChrisChinchilla:
### Proposed changes

An idea to see if reorganising some of the trusted content pages makes the content clearer and easier to find.


_data/toc.yaml
```diff
@@ -1842,16 +1842,12 @@ manuals:
         title: Manage
   - path: /docker-hub/official_images/
     title: Docker Official images
-  - sectiontitle: Docker Verified Publisher
-    section:
-      - path: /docker-hub/publish/
-        title: Overview
-      - path: /docker-hub/publish/insights-analytics/
-        title: Insights and analytics
-      - path: /docker-hub/publish/repository-logos/
```
gtardif, line 1851: Is this an old page that should get deleted as well?

gtardif (commented):
Overall LGTM, all makes sense

_data/toc.yaml
ChrisChinchilla: @docker/docs-team any thoughts on if there's a better way to do this? Or just remove the page? As I don't think it has received much traffic anyway.

aevesdocker: I would just remove the page. It has a lot of inaccurate info on it and now that it has been pulled out of the sub-folder it'll be more visible which is not what we want

docker-hub/dvp-program.md
```diff
@@ -17,6 +17,7 @@ redirect_from:
   - /docker-hub/publish/trustchain/
   - /docker-hub/publish/byol/
   - /docker-hub/publish/publisher-center-migration/
+  - /docker-hub/publish/
 ---
```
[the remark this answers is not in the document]
ChrisChinchilla: @aevesdocker This was because the content has actually been merged into two different files, but I will add it to here so at least it redirects somewhere and DVP is preferable.

aevesdocker (approved):
🚀

PR #17437, opened 2023-05-30
Clarify image sizing

ChrisChinchilla:
### Proposed changes

I was trying to set an image explicitly by height and realised it wasn't possible, so this PR makes that clear in the contribution docs.


aevesdocker (approved):
Thank you!

PR #17462, opened 2023-06-01
Fix "scope" default value for GHA cache options

james-warren0:
Update the `cache-to` and `cache-from` `scope` config value to the correct default of `buildkit`.

### Proposed changes

The documentation for the GitHub Action cache is incorrect in listing that the branch name is the default scope value. The correct value is `buildkit`, which is also listed as the default value in the buildkit docs [here](https://github.com/moby/buildkit#github-actions-cache-experimental).

### Related issues (optional)

N/A


dvdksn:
@crazy-max could you ptal?

crazy-max (dismissed):

crazy-max:
Arg needs rebase :sweat_smile:

crazy-max (changes requested):
Needs rebase :pray:

dvdksn:
Fixed in 17eb61acf31656c09751fa5bc4b4cbd60030ed90

PR #17477, opened 2023-06-05
Add scout URLs to allow list doc and tidy

ChrisChinchilla:
### Proposed changes

This PR adds the Scout service URLs to the allowlist doc and whilst I was there I tidied it a little.


aevesdocker (approved):
Thanks @ChrisChinchilla

PR #17488, opened 2023-06-07
Add "-a" flag to "Prune everything" section so it includes all unused images, not just dangling ones

coritanic:
### Proposed changes

"Prune everything" should prune _all_ unnecessary components, including unused images. If someone doesn't want to delete unused images, they can omit the "-a" flag. This way, people who want to clean unused components can attain their purpose.

### DIFF
--- a/config/pruning.md
+++ b/config/pruning.md
@@ -158,16 +158,16 @@ WARNING! This will remove:
 Are you sure you want to continue? [y/N] y
 ```
 
-To also prune volumes, add the `--volumes` flag:
+To also prune volumes and unused images, add the `--volumes -a` flag:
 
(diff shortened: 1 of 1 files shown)


config/pruning.md
```diff
@@ -158,16 +158,16 @@ WARNING! This will remove:
 Are you sure you want to continue? [y/N] y
 ```
 
-To also prune volumes, add the `--volumes` flag:
+To also prune volumes and unused images, add the `--volumes -a` flag:
```
dvdksn, line 161:
```suggestion
To also prune volumes and unused images, add the `--volumes` and `-a` flags:
```

PR #17508, opened 2023-06-08
Update journald.md

kurthaegeman:
Convert Python 2.7 example to Python 3.6+ example


kurthaegeman:
Not a class instance but a dict.

PR #17511, opened 2023-06-09
WIP: reference: show aliases for commands

thaJeztah:
- relates to https://github.com/docker/docs/pull/10471
- relates to https://github.com/docker/docs/pull/10471#issuecomment-702344459

### DIFF
--- a/_includes/cli.md
+++ b/_includes/cli.md
@@ -71,6 +71,16 @@ your client and daemon API versions.
 {% endcapture %}{{ command-orchestrator }}
 
 
+{% if controller_data.aliases %}
+{% assign aliases = controller_data.aliases | split: ', ' %}
+### Aliases
+
+The following commands are equivalent and redirect here:
+{% for alias in aliases %}
+{%- assign fname = alias | remove_first: "docker " | replace: " ", "_" -%}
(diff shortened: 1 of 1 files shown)


_includes/cli.md
```diff
@@ -71,6 +71,16 @@ your client and daemon API versions.
...
 
 
+{% if controller_data.aliases %}
+{% assign aliases = controller_data.aliases | split: ', ' %}
+### Aliases
+
+The following commands are equivalent and redirect here:
+{% for alias in aliases %}
+{%- assign fname = alias | remove_first: "docker " | replace: " ", "_" -%}
+- [{{ alias }}](/engine/reference/commandline/{{ fname }}/)
```
thaJeztah, line 81: I constructed the URLs for each alias here, but this won't work because we don't have redirects for all aliases.

Some options;

- (for now) just show the aliases only (without a link)
- add redirects for all aliases
- ^^^ but if we do, we need to update those redirects to the canonical URL once we moved the content (see https://github.com/docker/docs/pull/10471 and https://github.com/docker/cli/pull/2389)
- ^^^ if we add redirects now, those would already have a "moved _permanently_" redirect, so I'm not sure if it's a good idea to do that if we'll be updating the location in future (SEO-wise)

PR #17519, opened 2023-06-09
Update 02_our_app.md

meetp2022:
### Proposed changes

Added a hyphen 
in between to and do (to-do)  as it is much clearer


craig-osterhout:
Thanks for sharing the PR.
I don't think we have a standard defined for this, but it is consistently used without the hyphen in several more places. 
If you'd like to update it in this one topic, it should be updated in every other topic. There are several throughout the rest of the getting started guide and even in other guides like the orchestration section (https://docs.docker.com/get-started/kube-deploy/).

PR #17524, opened 2023-06-11
Update raspbian.md

fvillafa:
Although https://download.docker.com/linux/raspbian/ exists, when adding the repos with that URL, apt won't get a list of packages from there.

This can be checked by using the convenience script. The convenience script sets the repo as:

https://download.docker.com/linux/debian/

And it works, while the manual steps don't.

### DIFF
--- a/engine/install/raspbian.md
+++ b/engine/install/raspbian.md
@@ -92,7 +92,7 @@ Docker from the repository.
 
 #### Set up the repository
 
-{% assign download-url-base = "https://download.docker.com/linux/raspbian" %}
+{% assign download-url-base = "https://download.docker.com/linux/debian" %}
 
 1.  Update the `apt` package index and install packages to allow `apt` to use a
     repository over HTTPS:


thaJeztah:
Do you have more details about the failure? What version of raspbian are you running? There for sure are packages in that repository, so I'm curious what failed; https://download.docker.com/linux/raspbian/dists/bookworm/pool/stable/armhf/

fvillafa:
Hello Sebastiaan,

I found the error on my side. The reason why with Debian works is because there is a small text:

Docker Engine for Raspbian is compatible with the armhf architecture.

For the 64-bit version of Raspbian follow the instructions for [Debian](https://docs.docker.com/engine/install/debian/).

Just under the section "OS requirements".

I'm very sorry for my confusion!

fvillafa:
Closing as it was my mistake.

thaJeztah:
Ah, thanks, yes, that would explain. Perhaps we should highlight that line better

PR #17526, opened 2023-06-11
Windows appdata references

jmartens:
### Proposed changes

Updated patch for #16606 since apparently I cannot reopen it.

### Related issues (optional)
#16606


aevesdocker (approved):
Thank you @jmartens

PR #17534, opened 2023-06-13
Update build-images.md

handestd:


language/golang/build-images.md
```diff
@@ -188,7 +188,7 @@ and can figure out what the user meant (most of the time), Docker's `COPY` comma
 is quite sensitive in its interpretation of the trailing slash.
 
 ```dockerfile
-COPY go.mod go.sum ./
+COPY go.mod go.sum .1121212/
```
ChrisChinchilla, line 191: @handestd I don't think this is correct unless I am misunderstanding something here? This should be the current directory.

PR #17551, opened 2023-06-15
Updated installation doc for Ubuntu

MokhFn:
added command to start docker engine before running the first test


dvdksn:
The docker service starts automatically when installed using the package manager
https://github.com/docker/docs/issues/16849

PR #17552, opened 2023-06-15
debian.md: fix invalid grammar

perlun:


engine/install/debian.md
```diff
@@ -41,9 +41,9 @@ architectures.
 Before you can install Docker Engine, you must first make sure that any
 conflicting packages are uninstalled.
 
-Distro maintainers provide an unofficial distributions of Docker packages in
-APT. You must uninstall these packages before you can install the official
-version of Docker Engine.
+Distro maintainers provide an unofficial distribution of Docker packages in
+their repositories. You must uninstall these packages before you can install the 
+official version of Docker Engine.
```
perlun, line 46: The typo is what triggered this PR. :+1: It's an obvious, uncontroversial change.

----

However, I'm also tempted to rephrase the whole "unofficial" part to at least put it in quotes. The way this is phrased now has an extremely Docker Inc-centric world view. The packages provided in Debian are in no way "unofficial", they are in fact "the official Debian packages" that forms the official Debian distribution!

So, I think that trying to phrase this a bit more neutrally would make sense. Perhaps using a phrase like "vendor-provided" would work? At least to me, the current wording feels like "marketing speech" in a negative way.

PR #17555, opened 2023-06-16
Fix capitalization error in sentence by adding an uppercase letter at…

### LINKED ISSUE #17520: Missing upper letter on first word of sentence
### Is this a docs issue?

### Type of issue

There's a problem with the website

### Description

This sentence:
> **the** following command starts the `web` service and runs `bash` as its command:

Does not start with an uppercase `T`.

### Location

https://docs.docker.com/engine/reference/commandline/compose_run/#description

### Suggestion

This would be the fix:
```diff
- the following command starts the `web` service and runs `bash` as its command:
+ The following command starts the `web` service and runs `bash` as its command:
```

aroramrinaal:
### Proposed changes
I made a documentation change to fix a missing uppercase letter at the beginning of a sentence. The specific fix is as follows:

Change in the original documentation:
From: "the following command starts the web service and runs bash as its command:"
To: "The following command starts the web service and runs bash as its command:"
The purpose of this change is to ensure proper capitalization and improve the readability and consistency of the documentation.

### Related issues (optional)
  Closes #17520

### DIFF
--- a/_data/compose-cli/docker_compose_run.yaml
+++ b/_data/compose-cli/docker_compose_run.yaml
@@ -3,7 +3,7 @@ short: Run a one-off command on a service.
 long: |-
     Runs a one-time command against a service.
 
-    the following command starts the `web` service and runs `bash` as its command:
+    The following command starts the `web` service and runs `bash` as its command:
 
     ```console
     $ docker compose run web bash


dvdksn:
Hi, and thank you for contributing.

These yaml files are generated from a combination of code and Markdown files. For that reason, to make the change persistent, we need to change it at the source, which for this file is here: https://github.com/docker/compose/blob/v2/docs/reference/compose_run.md#description

Would you mind raising a PR in docker/compose too, with the same change? That way we can ensure this doesn't get overwritten the next time we re-generate the docs for this command.

aroramrinaal:
> Hi, and thank you for contributing.
> 
> These yaml files are generated from a combination of code and Markdown files. For that reason, to make the change persistent, we need to change it at the source, which for this file is here: https://github.com/docker/compose/blob/v2/docs/reference/compose_run.md#description
> 
> Would you mind raising a PR in docker/compose too, with the same change? That way we can ensure this doesn't get overwritten the next time we re-generate the docs for this command.

Yes I will raise a PR in docker/compose too, with the same change.

dvdksn (approved):

PR #17561, opened 2023-06-18
Update debian versions

casperklein:
### Proposed changes

Debian 12 is the new stable release. This PR updates the debian versions list.


dvdksn (approved):
Thanks!

PR #17565, opened 2023-06-19
Fix typo on "Create an advanced frontend extension" page

scotthallock:
### Proposed changes

Fix typo on the [Create an advanced frontend extension](https://docs.docker.com/desktop/extensions-sdk/build/frontend-extension-tutorial/) page of the docs. ("event" should be "even").

**Current:**
> The `docker extension init` generates a React based extension. But you can still use it as a starting point for your own extension and use any other frontend framework, like Vue, Angular, Svelte, etc. or **event** stay with vanilla Javascript.

**Fix:**
> The `docker extension init` generates a React based extension. But you can still use it as a starting point for your own extension and use any other frontend framework, like Vue, Angular, Svelte, etc. or **even** stay with vanilla Javascript.


ChrisChinchilla (approved):
Thanks @scotthallock !

PR #17573, opened 2023-06-20
Update release-notes.md

manojmandya07:
Changed "New" to "New feature" to make it consistent throughout the doc.


aevesdocker:
Thanks for the PR @manojmandya07. We changed our release note template for the whole of docs to use 'New' instead of 'New features' (and only retroactively set this for the last two years), so this change would then be inconsistent with the rest of docs so I'm going to close this PR

PR #17588, opened 2023-06-21
engine: remove Ubuntu 18.04 Bionic Beaver LTS, as it reached end of standard support

thaJeztah:
Ubuntu 18.04 LTS reached end of standard support. Expanded Security Maintenance
(ESM) is available, but requires a subscription, and we don't provide packages
for those.

- https://wiki.ubuntu.com/Releases
- https://ubuntu.com//blog/18-04-end-of-standard-support

Also removing Ubuntu 18.04 from examples in other places.


dvdksn (approved):
Thanks!

PR #17600, opened 2023-06-22
Docs on using Docker Scout in CI

ChrisChinchilla:
### Proposed changes

Adds docs for using Docker Scout in CI


scout/ci.md
```diff
@@ -0,0 +1,145 @@
...
+  steps:
+    - name: Checkout repository
+    uses: actions/checkout@v3
+    with:
+        ref: $\{\{ env.SHA \}\}
+
+    - name: Setup Docker buildx
+    uses: docker/setup-buildx-action@v2.5.0
+    with:
+        driver-opts: |
```
cdupuis, line 70: The `driver-opts` should go. We don't need that any longer.

scout/ci.md
```diff
@@ -0,0 +1,142 @@
...
+    uses: docker/metadata-action@v4.4.0
+    with:
+        images: $\{\{ env.REGISTRY \}\}/$\{\{ env.IMAGE_NAME \}\}
+        labels: |
+        org.opencontainers.image.revision=$\{\{ env.SHA \}\}
+        tags: |
+        type=edge,branch=$repo.default_branch
+        type=semver,pattern=v\{\{version\}\}
+        type=sha,prefix=,suffix=,format=short
+```
```
dvdksn, line 92: Looks like there are some indentation issues with this example.

dvdksn:
We have a section covering all other Docker GitHub actions, here: https://docs.docker.com/build/ci/github-actions/

ChrisChinchilla:
@dvdksn Hmm, though that entire section is in the build section, we also have a CI section in the extensions docs, although it's not related to any GHA. Let me ask around and get some opinions.

crazy-max:
> We have a section covering all other Docker GitHub actions, here: https://docs.docker.com/build/ci/github-actions/
> 
> I think we should move the Scout CI page to that section too, so we have all GHA docs in the same place.

I don't think it should be added to "Docker Build" section as this is specific to Scout which is not under the build tooling umbrella tbh.

ChrisChinchilla:
@crazy-max and @dvdksn Take a look at last changes, I added a brief mention both ways, but left the majority of the content where it is.

build/ci/github-actions/index.md
```diff
@@ -28,6 +28,8 @@ The following GitHub Actions are available:
   builds.
 - [Docker Buildx Bake](https://github.com/marketplace/actions/docker-buildx-bake){:target="blank" rel="noopener" class=""}:
   enables using high-level builds with [Bake](../../bake/index.md).
+- [Docker Scout](https://github.com/docker/scout-action){:target="blank" rel="noopener" class=""}:
+  analyze Docker images for security vulnerabilities.
```
crazy-max, line 32: Would be better to have this published to the GitHub Marketplace like other actions instead of repo link for better discoverability.

cdupuis: It be there in the time but it isn't there now.

PR #17619, opened 2023-06-26
Update 02_our_app.md

carlosbarlv:
For some reason if we don't pass the path to the docker file it causes the error that requires 1 argument.

Without .
image

When add .
image

### DIFF
--- a/get-started/02_our_app.md
+++ b/get-started/02_our_app.md
@@ -103,7 +103,7 @@ To build the [container image](../get-started/overview.md/#docker-objects){:targ
 
    Build the container image.
    ```console
-   $ docker build -t getting-started .
+   $ docker build . -t getting-started .
    ```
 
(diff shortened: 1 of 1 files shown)


craig-osterhout:
Thanks for sharing the issue.
The path (`.`) is already at the end of the current command. As your first screenshot is cut off, I cannot tell if you included it. Based on the second screenshot, it doesn't look like it. Closing this as it's already included in the command.

`docker build -t getting-started .`

carlosbarlv:
> Thanks for sharing the issue. The path (`.`) is already at the end of the current command. As your first screenshot is cut off, I cannot tell if you included it. Based on the second screenshot, it doesn't look like it. Closing this as it's already included in the command.
> 
> `docker build -t getting-started .`

Oh, thank you, sorry

PR #17621, opened 2023-06-26
Add raw filter to Scout CI doc

ChrisChinchilla:
Sorry @dvdksn and @crazy-max I attempted to fix up commits in the publish PR and then that caused wider Git conflicts, so this adds the correct liquid filters as discussed and then I will manually recreate the publish PR.

### DIFF
--- a/scout/ci.md
+++ b/scout/ci.md
@@ -24,6 +24,7 @@ with Docker in general, see [the GitHub Actions documentation](../build/ci/githu
 
 Add the following to a GitHub action YAML file:
 
+{% raw %}
 ```yaml
 name: Docker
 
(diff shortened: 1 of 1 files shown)


scout/ci.md
```diff
@@ -89,13 +92,15 @@ Add the following to the YAML file:
           type=semver,pattern=v\{\{version\}\}
```
crazy-max, line 92: You can remove escape sequences now like this one:

```suggestion
          type=semver,pattern=v{{version}}
```

scout/ci.md
```diff
@@ -130,6 +137,7 @@ Add the following to the YAML file:
     only-severities: critical,high
     token: $\{\{ secrets.DOCKER_PAT \}\}
```
crazy-max, line 138: Same here and everywhere else

PR #17630, opened 2023-06-27
redirects: add redirect for runtime feedback form

laurazard:
### Proposed changes

Added a redirect for the runtime feedback form in order to remove the hardcoded URLs in case we need to redirect things in the future.


laurazard:
cc @thaJeztah @dvdksn

_redirects.yml
```diff
@@ -557,6 +557,8 @@
 # feedback
 "https://docs.google.com/forms/d/e/1FAIpQLSenjK8KiBiOrwCpq06u_iVEZGv4tfTEpfmY_cQDmsqJrHL9Gw/viewform":
   - /feedback/extension/
+"https://docs.google.com/forms/u/1/d/e/1FAIpQLScgzKwqQyr7BmnphdrcGTO3wkye_McHJVsWzOIgEESfqmditQ/formResponse":
+  - /feedback/runtime
```
thaJeztah, line 561: Will we use this for the containerd-integration only, or also beyond that? (thinking if this would be redirected elsewhere once that goes "GA", or if we intent to use this for other feedback after that)

Can you add a trailing slash to the URL?

```suggestion
  - /feedback/runtime/
```

laurazard: Original thought was that we'd probably reuse this (even if we end up switching to a different form/survey platform).
Added the trailing `/`!

laurazard:
@dvdksn PTAL

_redirects.yml
```diff
@@ -563,6 +563,8 @@
 # feedback
 "https://docs.google.com/forms/d/e/1FAIpQLSenjK8KiBiOrwCpq06u_iVEZGv4tfTEpfmY_cQDmsqJrHL9Gw/viewform":
   - /feedback/extension/
+"https://docs.google.com/forms/u/1/d/e/1FAIpQLScgzKwqQyr7BmnphdrcGTO3wkye_McHJVsWzOIgEESfqmditQ/formResponse":
```
thaJeztah, line 566: Interesting; what's the difference between `viewForm` and `formResponse` ?

laurazard: `formResponse` is the path where form response payloads get `POST`ed to, which is what we're interested in for this usecase (feedback plugin submitting the response programatically). `GET`ing it redirects to `viewForm` though, so if we intend to use this URL in other scenarios (such as linking from DD/docs), it'll work there as well.

thaJeztah: Oh! I thought this was the link that we used in Docker Desktop (which ideally wouldn't be hard-coded to a form either). Wondering if we can find some way to use the same redirect, but guess that's not an option 🤔

dvdksn (approved):
LGTM

_redirects.yml
laurazard: That's what I meant – we can use the same one, but the one we define as the redirect should be `formResponse` since we can `GET` that and it will redirect to the correct page (`viewForm`), but for `POST`ing from the plugin it needs to be `formResponse`.

PR #17632, opened 2023-06-27
Fix small typo on "Containerize an application" in "Start an app container" section

cklanac:
The `-d` flag is (short for `--detach`). Not `--detached`.


craig-osterhout (approved):
LGTM. Thank you.

PR #17635, opened 2023-06-28
desktop: remove BuildKit feature from daemon config

crazy-max:
### Proposed changes

This is a follow-up of a future change in Docker Desktop:

> 
> Since Docker 23, `docker build` command is forwarded to `docker buildx build` which always enables BuildKit. If user tries to set `buildkit` to `false` in the daemon config it will have no effect. Only setting the env var `DOCKER_BUILDKIT=0` client side makes it use the legacy builder when using the `docker build` command.
> 
> So better to remove the BuildKit feature from the daemon config to avoid any confusion.

### Related issues (optional)

* https://github.com/docker/for-win/issues/13548

### DIFF
--- a/desktop/settings/linux.md
+++ b/desktop/settings/linux.md
@@ -136,10 +136,7 @@ You configure the daemon using a JSON configuration file. Here's what the file m
       "enabled": true
     }
   },
-  "experimental": false,
-  "features": {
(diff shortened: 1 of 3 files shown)


dvdksn:
👍🏻 removed the same from `/config/daemon/`

PR #17643, opened 2023-06-29
Update develop.md

ilivestrong:
### Proposed changes
While going through the Go specific guide to run database container, I found a typo, where it was mentioned that: `database container was started with --name db` but instead it should be `database container was started with --hostname db`


dvdksn:
Hi. Can you rebase this PR please? thanks

craig-osterhout (approved):

thaJeztah:
I wonder why the example uses --hostname; by default, containers can be resolved using their name (roach in this example), so wondering why the example opted to pick a hostname different from that; removing that part can simplify the examples.


Any idea @craig-osterhout ?

PR #17664, opened 2023-06-30
Document restricted callout

ChrisChinchilla:
### Proposed changes

Documents restricted call out option.


ChrisChinchilla:
@aevesdocker Turns out we already had one, it just wasn't documented.

aevesdocker (approved):
Ooooo that looks great! Niiice

PR #17669, opened 2023-07-02
Fix default address pool size

cu:
The default pools are correct in the first example, but a typo made its way into the second.
  https://github.com/moby/libnetwork/blob/master/ipamutils/utils.go


dvdksn (approved):
Thanks @cu !

PR #17671, opened 2023-07-02
Define container portability more precisely

jwhb:
### Proposed changes

The Docker docs overview page states that "[a container] Is portable (can be run on any OS)".
The detail "run on any OS" is not true due to the fact that containers run binary code that is constrained to execution on a specific operating system (mostly Linux, and Windows containers only run on Windows).

Furthermore, the portability of containers is constrained to CPU architectures.

I propose to correct the portability claim to detail the actual constraints of portability.


dvdksn:
The containers themselves aren't OS-agnostic, but container runtimes let you run containers on any operating system, and architecture by means of emulation. It's more nuanced than "runs on any OS", but this is not the right place to go into this.

PR #17702, opened 2023-07-06
Update rust guide

craig-osterhout:
### Proposed changes

Issue: The migrations directory is not found when building the application.
Fix: Add step to add bind mount for migrations directory in the build stage. 

### Related issues (optional)
ENGDOCS-1496

### DIFF
--- a/language/rust/develop.md
+++ b/language/rust/develop.md
@@ -177,7 +177,7 @@ For the sample application, you'll use a variation of the backend from the react
    $ docker build -t rust-backend-image .
    ```
 
-5. Run `docker run` to run the image as a container on the same network as the database.
+5. Run `docker run` with the following options to run the image as a container on the same network as the database.
 
    ```console
    $ docker run \
--- a/language/rust/develop.md
+++ b/language/rust/develop.md
@@ -96,13 +96,88 @@ For the sample application, you'll use a variation of the backend from the react
    ? What port does your server listen on? 8000
    ```
 
-3. In the cloned repository's directory, run `docker build` to build the image.
+3. In the cloned repository's directory, open the `Dockerfile` in an IDE or text editor to update it.
+
+   `docker init` handled creating most of the instructions in the Dockerfile, but you'll need to update it for your unique application. In addition to a `src` directory, this application includes a `migrations` directory to initialize the database. Add a bind mount for the `migrations` directory to the build stage in the Dockerfile. The following is the updated Dockerfile.
+
+   ```dockerfile
+   # syntax=docker/dockerfile:1
+
+   # Comments are provided throughout this file to help you get started.
+   # If you need more help, visit the Dockerfile reference guide at
+   # https://docs.docker.com/engine/reference/builder/
+   
+   ################################################################################
+   # Create a stage for building the application.
+   
+   ARG RUST_VERSION=1.70.0
+   ARG APP_NAME=react-rust-postgres
+   FROM rust:${RUST_VERSION}-slim-bullseye AS build
+   ARG APP_NAME
+   WORKDIR /app
+   
+   # Build the application.
+   # Leverage a cache mount to /usr/local/cargo/registry/
+   # for downloaded dependencies and a cache mount to /app/target/ for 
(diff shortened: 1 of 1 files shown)


language/rust/develop.md
```diff
@@ -96,13 +96,88 @@ For the sample application, you'll use a variation of the backend from the react
...
+   # for downloaded dependencies and a cache mount to /app/target/ for 
+   # compiled dependencies which will speed up subsequent builds.
+   # Leverage a bind mount to the src directory to avoid having to copy the
+   # source code into the container. Once built, copy the executable to an
+   # output directory before the cache mounted /app/target is unmounted.
+   RUN --mount=type=bind,source=src,target=src \
+       --mount=type=bind,source=Cargo.toml,target=Cargo.toml \
+       --mount=type=bind,source=Cargo.lock,target=Cargo.lock \
+       --mount=type=cache,target=/app/target/ \
+       --mount=type=cache,target=/usr/local/cargo/registry/ \
```
dvdksn, line 130: I think that the cache dir for rust build cache is `target/release/` by default, so maybe we should use that instead of the registry cache?

https://doc.rust-lang.org/cargo/guide/build-cache.html

dvdksn: cc @jedevc

craig-osterhout: Also cc @cmrigney if related to docker init template

cmrigney: We actually cache both of these:
1) `--mount=type=cache,target=/app/target/` includes `target/release` in addition to some other build cache (like deps).
2) `--mount=type=cache,target=/usr/local/cargo/registry/` is to avoid a lengthy re-fetch of the cargo packages every time the image is built.

cmrigney: (Also see comment above the command)

language/rust/develop.md
```diff
@@ -96,13 +96,88 @@ For the sample application, you'll use a variation of the backend from the react
    ? What port does your server listen on? 8000
    ```
 
-3. In the cloned repository's directory, run `docker build` to build the image.
+3. In the cloned repository's directory, open the `Dockerfile` in an IDE or text editor to update it.
+
+   `docker init` handled creating most of the instructions in the Dockerfile, but you'll need to update it for your unique application. In addition to a `src` directory, this application includes a `migrations` directory to initialize the database. Add a bind mount for the `migrations` directory to the build stage in the Dockerfile. The following is the updated Dockerfile.
```
stephaurelio, line 101: **Thought:** Is there a way our styling for code snippets can include line numbers? That way it's easier to identify what line needs to be changed. It was easy enough to compare the original with the updated one here, but just thought that would make it extra clear.

language/rust/develop.md
```diff
@@ -96,13 +96,88 @@ For the sample application, you'll use a variation of the backend from the react
...
+   ```
+
+4. In the cloned repository's directory, run `docker build` to build the image.
 
    ```console
    $ docker build -t rust-backend-image .
    ```
 
-4. Run `docker run` to run the image as a container on the same network as the database.
+5. Run `docker run` to run the image as a container on the same network as the database.
```
stephaurelio, line 180: I wonder if beginners may just copy paste `docker run` and miss that they need all the options? So it may help them to specify that: 
```suggestion
5. Run `docker run` with the following options to run the image as a container on the same network as the database.
```

stephaurelio (approved):
LGTM ✨ I started from the beginning of this section and tested it out and didn't run into any issues, everything worked as expected. Just left a couple of thoughts as comments.

language/rust/develop.md
craig-osterhout: No way to include line numbers currently.
One alternative we have used is specifying the line number and using a diff codeblock. It doesn't work good with copy-paste though.
For example, https://docs.docker.com/get-started/03_updating_app/

stephaurelio: Gotcha, this works as-is then I would say!

stephaurelio (approved):
LGTM 🚀

PR #17710, opened 2023-07-10
Add Scout go links

ChrisChinchilla:
### Proposed changes

Adds convenience go links for Scout

### DIFF
--- a/_redirects.yml
+++ b/_redirects.yml
@@ -81,6 +81,12 @@
 # provide a short, permanent link to refer to a topic in the documentation.
 # For example, the docker CLI can output https://docs.docker.com/go/some-topic
 # in its help output, which can be redirected to elsewhere in the documentation.
+"/scout/":
+  - /go/scout/
+"/scout/quickstart/":
+  - /go/scout-quickstart/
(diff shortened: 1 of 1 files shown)


thaJeztah (approved):
"LGTM"

but I agree with your earlier mention of these not being much more memorable than the links they point to, so slightly wondering if it would be fine to use the regular URLs (unless we want to redirect to anchors in future)

thaJeztah:
Perhaps minor nit; should we sort the list of `/go/` redirects in the file, to make it easier to find them?

ChrisChinchilla:
@thaJeztah I felt having them by section might be easier? I think for now I will create them and let the team decide if they want to use them or not and we can always remove them after a trial period to see if they are used at all?

thaJeztah:
> @thaJeztah I felt having them by section might be easier? I think for now I will create them and let the team decide if they want to use them or not and we can always remove them after a trial period to see if they are used at all?

Ah, yes that makes sense as well. I guess as long as we stick to one approach, it works. "hunting down" redirects in a list that becomes longer can be a pain 😅

PR #17719, opened 2023-07-11
ENGDOCS-1507

aevesdocker:
PR makes it clear that the documentation refers to Compose V2 and reduces the frequency on the banner

### DIFF
--- a/_includes/compose-eol.md
+++ b/_includes/compose-eol.md
@@ -5,5 +5,5 @@
 > Compose V2 is included with all currently supported versions of Docker Desktop. 
 > For more information, see [Migrate to Compose V2](/compose/migrate).
 >
-> All documentation describes Compose V2 functionality. 
+> Docker's documentation refers to and describes Compose V2 functionality. 
 {: .important}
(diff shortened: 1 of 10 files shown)


aevesdocker:
@usha-mandya As per our sync, would something like this work?

usha-mandya:
Thanks @aevesdocker. I was wondering if it makes sense to remove it from the other topics  such as Service profiles, GPU support, Extend services, and Networking?

aevesdocker:
GPU support, yes for sure, I'll remove it from there. But I also have other feedback from another Captain saying that the visibility of the migration was not clear, so I'm going to leave it on the other pages. Networking is very popular page, and Extend services may also get an increase in viewings with the latest release of Compose. Profiles is also on an upwards trajectory as I increased the visibility of it in the Compose spec!

usha-mandya:
Sure.  Please go ahead and remove the banner from the GPU topic. We can go back and update the pages if we start getting more feedback on any of these.

usha-mandya (approved):

PR #17722, opened 2023-07-11
Fixed broken link to ELK docker compose example

dutts:
### Proposed changes

Fixed broken link to ELK docker compose example


aevesdocker (approved):
Thank you for catching this @dutts!

PR #17724, opened 2023-07-11
Style and contributor guide updates

ChrisChinchilla:
### Proposed changes

Adds more context to some of our contributor guide and supporting Vale configuration.

### DIFF
--- a/.github/vale/Docker/Substitute.yml
+++ b/.github/vale/Docker/Substitute.yml
@@ -8,7 +8,7 @@ action:
 swap:
   '\b(?:eg|e\.g\.)[\s,]': for example
   '\b(?:ie|i\.e\.)[\s,]': that is
-  (?:sign on|log on|log in): sign in
+  (?:sign on|log on|log in|logon|login): sign in
   above: previous
   adaptor: adapter
   account name: username
@@ -27,6 +27,7 @@ swap:
   assembler: assembly language
   below: following
   check box: checkbox
+  check boxes: checkboxes
   click: select
   deselect: clear
   disable: turn off|toggle off
--- a/.github/vale/Docker/UIElement.yml
+++ b/.github/vale/Docker/UIElement.yml
@@ -1,0 +1,10 @@
+# Checks for use of bold that doesn't start with a capital letter
+# which might indicate a reference to something other than a UI label
+extends: existence
+message: "Check for correct usage of tab versus view"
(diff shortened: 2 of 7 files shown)


.github/vale/Docker/UIElement.yml
aevesdocker, line 1: yessss, love this

aevesdocker (commented):
Nice one @ChrisChinchilla. 
The only other thing that's been on my to-do here is to add more clarity around when to use the two new call outs. 
I think the purple one with the rocket ship should be reserved for when we highlight beta and EA products, and the test-tube one should be for calling out experimental products. WDYT?
Screenshot 2023-07-12 at 09 13 08

dvdksn:
Hi. No comments regaring the vale or style updates, but I would prefer if we do not change the docs about the site generator/components now. I've started to update those already in the Hugo branch and it seems unnecessary to make these updates here if we're migrating soon.

.github/vale/Docker/Substitute.yml
aevesdocker, line 1: can we add 'drop-down' here (as opposed to dropdown) ? We're currently inconsistent.

contribute/components/links.md
```diff
@@ -28,12 +27,10 @@ An example of a link to an auto-generated reference page that we pull in during
 ## HTML
 
 ```html
-
-It is best practice if [a link opens in a new window](https://docker.com/){: target="_blank" rel="noopener" class="_" }
+It is best practice if [a link to an internal page opens in the same window](/) and [a link to an external page opens in a new window](https://docker.com/){: target="_blank" rel="noopener" class="_" }
 
 You can also have [a markdown link to a custom target ID](#formatting-examples)
```
stephaurelio, line 32: Would it be possible to include internal link best practices here? I've seen both internal links to the path (`/subscription/add-seats/`) and to the markdown file (`../docker-id/index.md`). What is the preferred format? Do we want to callout the best practice to add `/` to the end of the link?

ChrisChinchilla: @stephaurelio Updated somewhat. While on links I also noticed with

`{: target="_blank" rel="noopener" class="_" }` apparently `rel="noopener"` is no longer needed and as far as I can tell looking at HTML, `class="_"` doesn't do anything.

aevesdocker (approved):

PR #17739, opened 2023-07-15
Fix grammar in official_images.md

Frezzle:
### Proposed changes

Just a grammar correction.

Alternatively it could be "we recommend" but the rest of the doc doesn't use "we".


thaJeztah (approved):
LGTM

- failure is unrelated and will be fixed by https://github.com/docker/docs/pull/17748

PR #17743, opened 2023-07-16
Update linux-install.md

koleajeolayinka:
### Proposed changes

- [ ] Commenting out the lines following the `lsmod | grep kvm` command prevents unintended copying of the commented lines along with the command when users attempt to copy only the `lsmod | grep kvm` command.

### DIFF
--- a/desktop/install/linux-install.md
+++ b/desktop/install/linux-install.md
@@ -85,10 +85,10 @@ To check if the KVM modules are enabled, run:
 
 ```console
 $ lsmod | grep kvm
-kvm_amd               167936  0
-ccp                   126976  1 kvm_amd
-kvm                  1089536  1 kvm_amd
-irqbypass              16384  1 kvm
+# kvm_amd               167936  0
+# ccp                   126976  1 kvm_amd
+# kvm                  1089536  1 kvm_amd
+# irqbypass              16384  1 kvm
 ```
 
 #### Set up KVM device user permissions


thaJeztah:
@craig-osterhout I wonder if we can make the copy button better for this. Fun thing is that we disable selecting the output, so if you "mouse-drag" to select the example, it excludes those lines;

Screenshot 2023-07-17 at 19 12 16


But with this PR applied, those lines are actually selectable (of course, they would be pasted as "comment", so won't do much harm).

Screenshot 2023-07-17 at 19 12 07



But the copy button isn't that smart, and copies the whole block irregardless (both before and after this PR);

copy/paste button before;

```console
lsmod | grep kvm
kvm_amd               167936  0
ccp                   126976  1 kvm_amd
kvm                  1089536  1 kvm_amd
irqbypass              16384  1 kvm
```

copy/paste button after;

```console
lsmod | grep kvm
# kvm_amd               167936  0
# ccp                   126976  1 kvm_amd
# kvm                  1089536  1 kvm_amd
# irqbypass              16384  1 kvm
```

craig-osterhout:
We could do a quick, messy hack to piggy back off the existing logic. For example, if `user-select:none`, then remove it when copying. I don't think the existing implementation of the copy code button will migrate to hugo though, so probably not worth fixing it until after the migration.

A workaround is to put the output in a separate codeblock. Readability is meh though.
image

PR #17750, opened 2023-07-17
Add link to JP community docs

usha-mandya:
This PR adds a link to the community-translated docs in JP


ChrisChinchilla (approved):
Looks good to me. I wonder if we want to use this opportunity to say what you could do if you want to make your own translation?

thaJeztah (approved):
LGTM as well

let me know if you want more things addressed before merging

thaJeztah:
Yes, I don't think we should _encourage_ this too much, as (even well-intended) such translations often get out of sync, and I frequently run into issues where users use such things and run into issues (outdated instructions etc).

If we _do_ document guidelines, it should probably go into a separate section, and more along the lines of _restrictions_ when using / translating docs, or hosting a mirror of the docs (for offline / internal use). The documentation itself has a permissive license, but that does not include _trademarks_, so we should write guidelines on how to use the content without violating our trademark.

thaJeztah (approved):
LGTM

PR #17759, opened 2023-07-19
fix grammatical typo

calebcartwright:
### Proposed changes

I suspect the diff is self-explanatory but will leave a note inline in case this proposed typo fix is unclear

### DIFF
--- a/engine/install/ubuntu.md
+++ b/engine/install/ubuntu.md
@@ -89,7 +89,7 @@ You can install Docker Engine in different ways, depending on your needs:
 
 - [Install it manually](#install-from-a-package) and manage upgrades manually.
 
-- Use a [convenience scripts](#install-using-the-convenience-script). Only
+- Use a [convenience script](#install-using-the-convenience-script). Only
   recommended for testing and development environments.
 
 ### Install using the apt repository {#install-using-the-repository}


engine/install/ubuntu.md
```diff
@@ -89,7 +89,7 @@ You can install Docker Engine in different ways, depending on your needs:
 
 - [Install it manually](#install-from-a-package) and manage upgrades manually.
 
-- Use a [convenience scripts](#install-using-the-convenience-script). Only
+- Use a [convenience script](#install-using-the-convenience-script). Only
```
calebcartwright, line 92: Based on the link text (and associated content) this appears to be singular, and if so, should reflect that.

If there are multiple scripts or a desire to reflect that, then an alternative would be:

```suggestion
- Use [convenience scripts](#install-using-the-convenience-script). Only
```

ChrisChinchilla: @calebcartwright Personally I think your first suggestion is correct here as it leads to a link that explains using a singular script.

```suggestion
- Use a [convenience script](#install-using-the-convenience-script). Only
```

dvdksn (approved):
Thanks!

PR #17765, opened 2023-07-19
community issue solution

aevesdocker:
This PR solves user confusion around the ID range. The text is clear but because of the way the numbers are displayed it can be hard to read and cause confusion. 

Fixes https://github.com/docker/docs/issues/17752


dvdksn (approved):
👍🏻

PR #17780, opened 2023-07-20
Add instruction to enter "host port" and clarify instructions

peterhauke:
Add an instruction to add a host port in the Optional settings dialog (for me on Windows, this was necessary).

Add "To" and a colon to make it clear that the points are headings and the sub-points that follow are the step-by-step instructions.


craig-osterhout (approved):
LGTM. Thank you.

PR #17816, opened 2023-07-24
Modify "stop" to "stop and remove" to avoid container name conflict error

Ekwinder:
### Proposed changes

Added instruction to remove the container along with stopping it to avoid error - 
```
docker: Error response from daemon: Conflict. The container name "/rest-server" is already in use by container "abc". 
You have to remove (or rename) that container to be able to reuse that name.
```


craig-osterhout (approved):
Thank you. LGTM.

PR #17819, opened 2023-07-24
Fix typos and make minor language edits

peterhauke:
### Proposed changes

Fix typos and make minor language edits

"make it so that" in "the Dockerfile instruction make it so that the builder" sounds a bit clumsy. I tried to improve it without rewriting the whole sentence.

I believe "downloading and installing dependencies" is plural so the verb "occur" should also be plural.


dvdksn (approved):
Thank you!

PR #17828, opened 2023-07-26
Update zfs-driver.md (remove section "Limit a container's writable storage quota")

Harliff:
Instruction how to limit container's size has been removed because this option doesn't work and prevent the docker service from starting up.

See https://github.com/moby/moby/issues/33847 for bug description

### Proposed changes

I've removed the section of docker documentation since it contains wrong instruction.

### Related issues (optional)
(https://github.com/moby/moby/issues/33847)

### DIFF
--- a/storage/storagedriver/zfs-driver.md
+++ b/storage/storagedriver/zfs-driver.md
@@ -118,26 +118,6 @@ the Docker host, and then add it to the `zpool` using the `zpool add` command:
 $ sudo zpool add zpool-docker /dev/xvdh
 ```
 
-### Limit a container's writable storage quota
-
-If you want to implement a quota on a per-image/dataset basis, you can set the
-`size` storage option to limit the amount of space a single container can use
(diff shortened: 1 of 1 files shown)


thaJeztah:
DOH! Looking at `parseOptions`, not `parseStorageOpt`, which still has the `size` option; https://github.com/moby/moby/blob/7a44e4cde0b85c81642aac69f7a5d5fe76ea86f4/daemon/graphdriver/zfs/zfs.go#L324-L336


```go
func parseStorageOpt(storageOpt map[string]string) (string, error) {
	// Read size to change the disk quota per container
	for k, v := range storageOpt {
		key := strings.ToLower(k)
		switch key {
		case "size":
			return v, nil
		default:
			return "0", fmt.Errorf("Unknown option %s", key)
		}
	}
	return "0", nil
}
```

dvdksn:
@thaJeztah should this be closed then?

Issue #17785, opened 2023-07-20
"Use multi-stage builds" uses outdated Go code


miparnisari:
### Is this a docs issue?

- [X] My issue is about the documentation content or website

### Type of issue

Information is incorrect

### Description

The Dockerfile mentioned in the example references `RUN go get github.com/golang/dep/cmd/dep`

But `github.com/golang/dep` was deprecated in 2020 

### Location

https://docs.docker.com/develop/develop-images/dockerfile_best-practices/

### Suggestion

Update to use `go.mod` files

PR #17829, opened 2023-07-26
develop: Update multi-stage example to use go mod

chris-crone:
### Proposed changes

Updated [an example of a multi-stage Go build](https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#use-multi-stage-builds) to use `go mod` as [`dep`](https://github.com/golang/dep) is no longer maintained.


craig-osterhout:
Fixes #17785

craig-osterhout (approved):
LGTM

PR #17831, opened 2023-07-26
desktop: initialize pass after install

craig-osterhout:
### Proposed changes

Initializing pass is a requirement to sign in.
Adding this to Linux system requirements was suggested as it's a system setting and should be part of the install/setup flow. But, sign in is not a requirement and therefore this setting is not necessarily a requirement.

"Where to go next" seems like a good alternative place to surface this information in the install/setup flow.

### Related issues (optional)
DA-646

### DIFF
--- a/desktop/install/linux-install.md
+++ b/desktop/install/linux-install.md
@@ -141,9 +141,10 @@ Make sure you meet the system requirements outlined earlier and follow the distr
 
 ## Where to go next
 
+- [Initialize `pass`](../get-started.md#credentials-management-for-linux-users) to manage credentials and sign in to Docker Desktop.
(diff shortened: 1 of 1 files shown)


aevesdocker (dismissed):
Makes sense! Thanks @craig-osterhout

aevesdocker:
Should it also be added to each linux distro page? 🤔

craig-osterhout:
> Should it also be added to each linux distro page? 🤔

@aevesdocker, thank you. Yea that makes sense. I've also been thinking about how to make a requirement optional without it being an oxymoron. What do you think about using "recommended"? For example, `Recommended: Initialize pass for credentials management`.
The rationalization being that there can be minimum system requirements and recommended system requirements.
This would solve the problem above because all the distro installation topics already point to the system requirements.

aevesdocker (approved):
:shipit:

PR #17870, opened 2023-08-04
build: add redirect to early access program

nicks:
### Proposed changes

Add redirect to build early access program

### Related issues (optional)
n/a

### DIFF
--- a/_redirects.yml
+++ b/_redirects.yml
@@ -594,3 +594,5 @@
   - /go/builders/
 "/build/builders/#selected-builder":
   - /go/builders/selected/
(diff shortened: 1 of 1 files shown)


_redirects.yml
```diff
@@ -594,3 +594,5 @@
   - /go/builders/
 "/build/builders/#selected-builder":
   - /go/builders/selected/
+"https://www.docker.com/build-early-access-program/?utm_campaign=onboard-30-customer-zero&utm_medium=in-product-ad&utm_source=desktop_v4":
+  - /go/build-eap
```
dvdksn, line 598: Could you add a trailing slash, just so it's consistent with the other redirects? Thx!

```suggestion
  - /go/build-eap/
```

nicks: updated!

PR #17877, opened 2023-08-07
Add role management limitation to SCIM

stephaurelio:
### Proposed changes

This tiny PR includes a limitation to the SCIM docs to callout that SCIM doesn't support role management.


craig-osterhout (approved):
LGTM

PR #17890, opened 2023-08-09
one off freshness

aevesdocker:

### DIFF
--- a/support/index.md
+++ b/support/index.md
@@ -33,7 +33,7 @@ If you have a paid Docker subscription, you can get support for the following ty
       * Installation crashes
       * Failure to launch Docker Desktop on first run
    * Desktop update issues
-   * Sign in issues in both the command line interface and Docker Hub user interface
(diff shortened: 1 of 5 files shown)


support/index.md
```diff
@@ -39,15 +33,15 @@ If you have a paid Docker subscription, you can get support for the following ty
       * Installation crashes
       * Failure to launch Docker Desktop on first run
    * Desktop update issues
-   * Login issues in both the command line interface and Docker Hub user interface
+   * Sign in issues in both the command line interface and Docker Hub user interface
```
stephaurelio, line 36: Would this work better since this is the adjective form:
```suggestion
   * Sign-in issues in both the command line interface and Docker Hub user interface
```
I always have to look this up!

stephaurelio (approved):
Tiny comment but otherwise LGTM! Yay freshness ✨

support/index.md
aevesdocker: oo yup! 🔍

dvdksn (approved):

PR #17917, opened 2023-08-11
ci: add Compose Watch configuration

milas:
### Proposed changes
This is a proof-of-concept that adds live update support while developing on the docs by adding the experimental `watch` section.

The existing Compose configuration has been separated out into a `compose.preview.yaml`, which builds the full site and serves it from nginx to mirror production.

The new Compose configuration does a `jekyll serve` and has a couple hacks for performance, namely that it does not run any of the local plugins from `_plugins`. They're slow and run on every change but aren't strictly necessary.

https://docker.atlassian.net/browse/ENV-206


milas:
To run, check out this PR branch and then:
```
docker compose up --build --wait && docker compose alpha watch
```

You might also want to open `docker compose logs -f` in another tab to see the Jekyll rebuild progress when making a change, as it takes 1-2 secs.

NOTE: It's necessary to manually refresh the page after making changes - there's no JS reloader in Jekyll.

PR #17919, opened 2023-08-13
Update debian.md

uberhacker:
Fixes `docker: Cannot connect to the Docker daemon at unix:///var/run/docker.sock. Is the docker daemon running?` This issue may or may not be isolated to Debian 12.

### DIFF
--- a/engine/install/debian.md
+++ b/engine/install/debian.md
@@ -151,7 +151,7 @@ Docker from the repository.
    To install the latest version, run:
 
    ```console
-    $ sudo apt-get install docker-ce docker-ce-cli containerd.io docker-buildx-plugin docker-compose-plugin
+    $ sudo apt-get install docker-ce docker-ce-cli containerd.io docker-buildx-plugin docker-compose-plugin cgroupfs-mount
    ```
 
    </div>


dvdksn:
this does not seem like it would be needed, unless perhaps you've tampered with your OS or init system in some way. cc @thaJeztah

thaJeztah:
Hm, yes, I don't think this would be needed, unless on a non-standard install without systemd perhaps? What kind of setup are you running, @uberhacker ?

uberhacker:
Yes, this is without systemd which is how MX Linux is set up out of the box.  I don't think we should be making the assumption that systemd is installed; especially for Debian which tends to generally have a minimal install by default.  I want to emphasize, I DID NOT tamper.  The install instructions should include steps for initd or systemd.  Again, we should not be making any assumptions about how any services are installed.

thaJeztah:
> Yes, this is without systemd which is how MX Linux is set up out of the box

Right, but MX Linux is a Debian derivative, and not a standard Debian install? While things _may_ work on such distros, [nothing is tested on those setups, so YMMV](https://docs.docker.com/engine/install/#other-linux-distros)

dvdksn:
Unfortunately, we can't maintain installation instructions for every possible distro variation and derivative, so we have to prioritize. MX Linux and other systemd-free distros are not among our top priorities at the moment, so I'll close this suggestion.

There's an old saying: no is temporary, yes is forever 🙂 We're always open to revisiting this in the future.

PR #17922, opened 2023-08-14
Add info on specific build that is targeted

kahenya-anita:
### Proposed changes

Added an extra explanation that breaks down which build stage is targeted by the multi-stage build command;

`docker build --target build -t hello .`

### Related issues

[Issue 17921](https://github.com/docker/docs/issues/17921)

### DIFF
--- a/build/building/multi-stage.md
+++ b/build/building/multi-stage.md
@@ -98,6 +98,14 @@ the stage named `builder`:
 $ docker build --target build -t hello .
 ```
 
+The builder targeted here is actually the first stage build responsible for building the Go application
+
+```dockerfile
+FROM golang:1.21 as build
+WORKDIR /src
+COPY <<EOF /src/main.go
+```
+
 A few scenarios where this might be useful are:
 
 - Debugging a specific build stage


build/building/multi-stage.md
```diff
@@ -98,6 +98,14 @@ the stage named `builder`:
...
 ```
 
+The builder targeted here is actually the first stage build responsible for building the Go application
+
+```dockerfile
+FROM golang:1.21 as build
+WORKDIR /src
+COPY <<EOF /src/main.go
+```
+
```
dvdksn, line 108: Thank you for your PR @kahenya-anita. Suggesting a small change to the wording here:

````suggestion
This build targets the stage responsible for building the Go application:

```dockerfile
FROM golang:1.21 as build
WORKDIR /src
COPY <<EOF /src/main.go
```

````


While reviewing the related issue, I noticed we have a typo on line 95 as well. It says the stage name is `builder`, but it should in fact be `build`. Could I ask you to please fix this typo as well? 

```
the stage named `builder`:
```

🙏🏻 thank you!

PR #17934, opened 2023-08-16
Improve volumes.md 

MHM98:
I was reading the docker volume [documentation](https://docs.docker.com/storage/volumes/) and I noticed something is wrong in it. here [here](https://docs.docker.com/storage/volumes/#block-storage-devices) discussed about how mount a block device to a container.
the `fallocate`  does not have -f option at all.
so the correct way of using this command is like this:
`fallocate -l 1G disk.raw`


dvdksn (approved):
Thank you for fixing this!

PR #17935, opened 2023-08-16
Update style guide word recommendations

stephaurelio:
### Proposed changes

The UI copy in Hub has changed from `register` to `sign up`. This PR updates the docs style guide to align with this change, and removes the vale recommendation to use register instead of sign up. Since register is used throughout the docs in other contexts besides account creation, I removed instead of swapping (i.e. recommend sign up for register).

### Related issues (optional)

Closes ENGDOCS-1582

### DIFF
--- a/.github/vale/Docker/Substitute.yml
+++ b/.github/vale/Docker/Substitute.yml
@@ -46,7 +46,6 @@ swap:
   repo: repository
   scroll: navigate
   (?:sign on|log on|log in|logon|login): sign in
-  sign up: register
   tb: TB
   vs: versus
   wish: want
--- a/contribute/style/recommended-words.md
+++ b/contribute/style/recommended-words.md
@@ -120,7 +120,7 @@ Don't use `please` in the normal course of explaining how to use a product, even
 
 #### register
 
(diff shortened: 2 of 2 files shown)


.github/vale/Docker/Substitute.yml
```diff
@@ -46,7 +46,6 @@ swap:
   repo: repository
   scroll: navigate
   (?:sign on|log on|log in|logon|login): sign in
-  sign up: register
```
aevesdocker, line 49: should we add the opposite rule in?
`register: sign up`?

stephaurelio: Was on the fence as it seems like register in the docs is also used for things other than create account (e.g. `Register a domain` or `Register a service`) so I wasn't sure if the recommendation would be additional noise or not 🤔 Happy to update this though!

aevesdocker: Ah, good point, ignore me!

aevesdocker (approved):

dvdksn (approved):

PR #17947, opened 2023-08-18
ci(validate-upstream): allow cleanup of stub files

crazy-max:
### Proposed changes

follow-up https://github.com/docker/buildx/pull/2006#issuecomment-1683505890

Adds `data-files-placeholder-prefix` input to handle removal of stub files. This can be useful in case commands on downstream repo are removed.

Tested here: https://github.com/crazy-max/buildx/actions/runs/5909610217/job/16030404526

cc @ktock

### DIFF
--- a/.github/workflows/validate-upstream.yml
+++ b/.github/workflows/validate-upstream.yml
@@ -19,6 +19,9 @@ on:
       data-files-placeholder-folder:
         required: false
         type: string
+      data-files-placeholder-prefix:
+        required: false
+        type: string
 
 jobs:
   run:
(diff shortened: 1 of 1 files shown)


ktock (approved):

ktock:
Thanks!

crazy-max:
Switched to draft, I will make something more reliable and generic than this.

dvdksn:
@crazy-max this workflow will break after we migrate the site to Hugo. Since we're planning to do that on Tuesday, I don't think there's a point in merging this. Wdyt?

dvdksn:
We'll need to reimplement a validation workflow that works with the go module system that Hugo supports

PR #17976, opened 2023-08-22
fix shortcodes and update callouts

stephaurelio:
### Proposed changes

This PR fixes areas where shortcodes are used for Docker Admin and Administration and Security pages. It also fixes the styling of a callout box on the SSO docs and in the contribution guide.


craig-osterhout (dismissed):
LGTM.
Noticed another at /single-sign-on/manage/

PR #17993, opened 2023-08-23
Leading space typo in a few console commands

ThomasBurgess2000:
### Proposed changes

I noticed an instance where one too many leading spaces was before the "$" in a console command, causing the "$" to be included when you click the copy button for the command. So I wrote a quick script to find any other places this occurred and fix them.


aevesdocker (approved):
Thank you @ThomasBurgess2000!

PR #18012, opened 2023-08-23
Fixed explanation error in second stage base image reference

Hopertz:
- Update explanation

### Proposed changes
Corrected explanation about second stage base image instead of  alpine:latest to scratch


dvdksn (approved):
🙏

PR #18030, opened 2023-08-24
icon consistency

aevesdocker:

### DIFF
--- a/content/admin/_index.md
+++ b/content/admin/_index.md
@@ -6,11 +6,11 @@ title: Docker Admin overview
 grid:
 - title: Company administration
   description: Explore how to manage a company in Docker Admin.
-  icon: lan
+  icon: apartment
   link: /admin/company/
 - title: Organization administration
   description: Learn about organization administration in Docker Admin.
-  icon: contact_page
+  icon: store
   link: /admin/organization/
 ---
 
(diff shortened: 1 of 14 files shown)


dvdksn:
❓ does it make sense to add somewhere in our contribution docs an "inventory" of the icons that we've used in our grids? Maybe just a grid like:

```yaml
grid:
  - title: contact_page
    icon: contact_page
```

aevesdocker:
I think it makes sense for a Notion page somewhere, but not sure if we need to add it to our public-facing contribute section - when do users change card content?

dvdksn:
Contributors may not add/change cards, but it would be easier for us (or at least me) to have this reference in the repo directly. Then I don't have to open notion 🐌 and find the right page - it's faster to find the reference among the files that I am already working with.

stephaurelio:
These look really good! I like the idea of a database or inventory of icons used. I'm the opposite of David and I think I'd prefer Notion just because I think it's easier to quickly reference but I'd be fine with whatever option means having a guide so we are consistent with what icons are used where.

stephaurelio (approved):
Thanks for adding the inventory page! This LGTM. I noticed a small styling error on that page that I'll drop a screenshot of in Slack.

PR #18039, opened 2023-08-26
Identify Linux/Windows container support requirements with System req…

softwarepronto:
…uirements

The page is not written clearly for those who want to use Windows containers. Systems requirements make it appear that Docker, carte blanche supports Windows/Linux containers with the system requirements. To see the Windows Pro or Windows Enterprise edition is require a users has to scroll down two screens and read a note. Windows Home/Education users who only look at "System requirements" have no idea that their machine will not run Windows containers.


aevesdocker:
Addressed as part of https://github.com/docker/docs/pull/18322

aevesdocker:
Thanks for contributing @softwarepronto, closed PR as this has now been fixed

PR #18042, opened 2023-08-26
Inverting dpkg install order

flavienbwk:
### Proposed changes

Installing the `docker-ce.deb` package before the `docker-ce-cli.deb` package causes a dpkg error.

This PR just inverts the dpkg arguments for these packages.

```bash
$ dpkg -i ./docker-ce_24.0.5-1~ubuntu.22.04~jammy_amd64.deb
Selecting previously unselected package docker-ce.
(Reading database ... 64311 files and directories currently installed.)
Preparing to unpack .../docker-ce_24.0.5-1~ubuntu.22.04~jammy_amd64.deb ...
Unpacking docker-ce (5:24.0.5-1~ubuntu.22.04~jammy) ...
dpkg: dependency problems prevent configuration of docker-ce:
 docker-ce depends on docker-ce-cli; however:
  Package docker-ce-cli is not installed.

dpkg: error processing package docker-ce (--install):
 dependency problems - leaving unconfigured
Errors were encountered while processing:
 docker-ce
```

### DIFF
--- a/content/engine/install/ubuntu.md
+++ b/content/engine/install/ubuntu.md
@@ -215,8 +215,8 @@ download a new file each time you want to upgrade Docker Engine.
    and Docker Compose packages:
 
    - `containerd.io_<version>_<arch>.deb`
-   - `docker-ce_<version>_<arch>.deb`
    - `docker-ce-cli_<version>_<arch>.deb`
+   - `docker-ce_<version>_<arch>.deb`
    - `docker-buildx-plugin_<version>_<arch>.deb`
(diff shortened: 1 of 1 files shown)


neersighted (changes requested):
This is not an issue if the command as provided is used. When `dpkg -i` is invoked with multiple packages, it will solve with the complete set of packages to be installed in that invocation.

In other words: the issue here is that you're installing packages one-at-a-time. You must install them all together, as demonstrated with the suggested invocation.

flavienbwk:
You are absolutely right. Sorry for the misunderstanding.

PR #18043, opened 2023-08-26
remove useless class declaration

yatoogamii:
### Proposed changes

Removing class declaration on an image of _index.md of BuildKit, it might be a typo or something from old markdown formatter used where we could put class to an image like that. But it didnt work in the actual docs website and is not necessary anyway I think.

```
![Directed acyclic graph (DAG)](../images/buildkit-dag.svg){:class="invertible" style="width:60%"}
```


dvdksn (approved):
my script didn't catch this one! thank you for fixing it.

PR #18048, opened 2023-08-28
Fix layout/style issue in "Important" callout

yatoogamii:
### Proposed changes

Fix layout/style issue in **Important** callout according to https://docs.docker.com/contribute/components/call-outs/


dvdksn (approved):
Thanks!

PR #18059, opened 2023-08-30
chore: remove images

aevesdocker:
removes unused images


dvdksn (approved):
🤩 This is a niiiice one

I did a quick and dirty verification with the following, which showed me no dangling references 👏🏻

```console
$ git --no-pager show --name-only HEAD^ | tail -n +7 > paths
$ git --no-pager show --name-only HEAD | tail -n +7 >> paths
$ awk -F "/" '{ print $NF }' paths > filenames
$ while read img; do rg "$img"; done < filenames
```

PR #18090, opened 2023-09-01
ENGDOCS-1600-scoutAndGetStarted

aevesdocker:


craig-osterhout (dismissed):
LGTM

Nits: Should "Pull Image" be sentence case.
I liked the old transparent backgrounds (at least in light mode) because the image felt more like part of the page. The white background of the image with the off-white background of the page feels more disjointed. I'm not a fan of dark mode, so I can't speak to that.
Not sure if adding a thin frame around the image would help or something else. 
For example, the old multi-container image vs. this new one. The old one has a frame with no white padding on the outside. The new one has a frame, but also adds white padding outside of the frame.

dvdksn:
In a follow-up at some point, we should probably add guidelines for images to the contribution section. The issue with transparent backgrounds is usually dual modality - it's near impossible to create a graphic that's transparent and that can work on both dark and light backgrounds. We could add support for inverting graphic colors based on theme selection, but we'd still need to have some rules for which colors we could use in that case, since simply inverting colors can still lead to contrast issues.

Also worth adding to an image guideline would be formats. I like to use `webp` over `png` or `jpeg` since they're much smaller and all browsers support them. It means taking an extra step to convert the image using https://squoosh.app/, but imo worth it.

aevesdocker:
Thanks for the feedback both!

I've changed 'Pull Image' to sentence case, and have addressed the frame issue for the multi-container image. 
Re: the transparency of the old images. I agree, this was better, but the issue was the some parts of some of the images weren't always visible when you clicked on them to make them bigger :/ 

I totally agree David, we need image guidelines. I'll add this to our wish list page, along with converting images to the new format (oh lordy that's gonna be long).

craig-osterhout (approved):

PR #18096, opened 2023-09-02
Update compatibility note for Ubuntu derivative distros

dapoadedire:
Added elementary OS

### Proposed changes

- Update compatibility note for Ubuntu derivative distros


dvdksn:
Thanks. This note mentions Linux Mint as an example only. I don't want to start listing alternatives here.

dapoadedire:
Thanks for the review, well appreciated.

PR #18119, opened 2023-09-06
engine: fix ipv6 example for Compose

milas:
### Proposed changes
Fixed the example for creating an IPv6 network in Compose YAML.

See https://github.com/compose-spec/compose-spec/blob/796804a634e4a1fd48a70f0b0b0e02f1a3a97890/06-networks.md#L168-L196

(Note: there's a typo there too, which is being fixed as well. The indentation is wrong, but the description of the fields/hierarchy is correct.)
 
### Related issues (optional)
* compose-spec/compose-spec#417
* docker/compose#10978

### DIFF
--- a/content/config/daemon/ipv6.md
+++ b/content/config/daemon/ipv6.md
@@ -54,10 +54,12 @@ The following steps show you how to create a Docker network that uses IPv6.
    - Using a Docker Compose file:
 
      ```yaml
-     networks:
-       ip6net:
-         enable_ipv6: true
-         subnet: 2001:0DB8::/112
+      networks:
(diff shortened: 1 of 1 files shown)


milas:
I think we might want to cross-link to this guide from the Compose docs, and perhaps have a link back to https://docs.docker.com/compose/compose-file/06-networks/ here? 😇 

(cc @aevesdocker as Compose 🧙🏻‍♀️)

aevesdocker (approved):

aevesdocker:
Yes agreed! Will create follow up PRs for this. Muchas gracias @milas

PR #18129, opened 2023-09-07
Update kube-deploy.md

aylusltd:
Fix formatting


craig-osterhout:
Thank you. There's definitely something wrong with the current formatting.
In your update, there seems to be an extra 4 spaces before each line in the yaml when copy-pasting it.
To fix it, you probably need to also delete the beginning 4 spaces from each line in that codeblock.

PR #18137, opened 2023-09-08
Fix malformed links

samy:
Some links were using bad syntax

### Proposed changes

Fix of broken links


craig-osterhout (approved):
LGTM. Thank you.

PR #18174, opened 2023-09-13
Update ubuntu.md

EvanCarroll:
You do not have to list every package individually.


EvanCarroll:
Ohh I see what ya'll are trying to one sec, let me fix this do not merge.

EvanCarroll:
This is not possible. =( You're doing it the right way. Your concern is the packages may not exist and you don't want the command to fail. Makes sense.

PR #18195, opened 2023-09-14
docs: Update develop.md for password auth failed for nodejs & postgresql

obedNuertey1:
Added a detailed work around step to password authentication error when running nodejs and postgresql with docker compose

### DIFF
--- a/content/language/nodejs/develop.md
+++ b/content/language/nodejs/develop.md
@@ -115,6 +115,40 @@ Run the following command to start your application.
 $ docker compose up --build
 ```
 
+> **Note**
+> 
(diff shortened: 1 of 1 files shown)


craig-osterhout:
Thanks for sharing the issue and update.

Did you run the compose app initially without creating the db/password.txt file?
I've done that a few times and it causes the db to be initialized without the password.  
If this occurs then you need to set the password as you have (even after creating a db/password.txt file) or delete the volume and run the compose app with the db/password.txt file already created.

obedNuertey1:
Yeah you are right. It was giving me problems earlier. But upon trying it again with what you said in the comment I found out that perhaps I might have made some mistake along the way that was why I had errors earlier.

Thanks a lot!

PR #18202, opened 2023-09-15
Update 02_our_app.md

RishiPradeep:
Fixed an issue where open_in_new would show after each link.

### Proposed changes

I was going through the Docker documentation and i noticed that the three lines under Prerequisites in the containerise an application page had open_in_new text at the end. I have removed that text so that it looks normal now.

### DIFF
--- a/content/get-started/02_our_app.md
+++ b/content/get-started/02_our_app.md
@@ -15,9 +15,10 @@ don't worry. This guide doesn't require any prior experience with JavaScript.
 
 ## Prerequisites
 
-- You have installed the latest version of [Docker Desktop](../get-docker.md).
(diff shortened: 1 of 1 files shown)


craig-osterhout:
Thanks for sharing the issue. 
I may be misunderstanding the issue and the fix.
If possible can you share a screenshot of what you see for the issue and what browser you're using? 

Here's what I see for the prereqs. I don't see `open_in_new` text after the links.
image

RishiPradeep:
I am extremely sorry for this as i realised too late that my browser was causing the issue due to a poor internet connection and i thought it was an error in the application. I however cannot seem to be able to recreate this issue. I was on Chrome and i was on an Office network. If this issue arises again i will be sure to let you know. Once again my deepest apologies.

PR #18208, opened 2023-09-15
Remove note suggesting running NTP in containers is "best practice" from develop/dev-best-practices/

### LINKED ISSUE #14834: NTP should be avoided always in containers... unless paravirtualization is broken
[develop/dev-best-practices.md](https://docs.docker.com/develop/dev-best-practices/)

> Always run an NTP client on the Docker host and within each container process and sync them all to the same NTP server. If you use swarm services, also ensure that each Docker node syncs its clocks to the same time source as the containers.

No. 

Userspace programs on Linux get their date-time via the system call `gettimeofday(2)`, which is usually implemented in the glibc as a VDSO call, which is like a kernel call without the context switch.  The values from that call are in turn obtained from the kernel's clock-discipline driver. In  para-virtualized kernels, this should simply delegate the information from the host -- including adjtime() run on the host. On bare-metal, either CPU registers or a timer-chip are being read for these values. 

If Docker isn't providing Linux kernels with such a paravirtualized clock discipline, get your act together and do that! Running NTP as a service within each container is incredibly wasteful and will end up producing lots of false alarms should delays result from many implementations. It also creates a networking configuration nightmare. And possibly overloading centralized NTP servers.  See notes provided by the original ntpd service: http://www.ntp.org/ntpfaq/NTP-s-algo-kernel.htm In the ideal case, NTP is simply doing network synchronization and finding that the container's clock is always correct, since those values are ultimately coming from the host.

martinky24:
### Proposed changes

The note about standing up an in-container NTP server felt off to me when I read this section. I did some googling, and came across https://github.com/docker/docs/issues/14834, in which @thaJeztah and @otheus came to agreement that the section was misinformed.

No one ever seems to have actually gotten around to opening a PR though, so here we are.

### Related issues

  Closes #14834


thaJeztah (approved):
oh, thanks! I recall that one yes, but looks like the stale-bot closed it.

this LGTM

(fun fact: with time-namespaces now being a thing in recent kernel versions, MAYBE some day this becomes relevant again 😂)

PR #18219, opened 2023-09-17
Update gettingstarted.md, corrected docker-compose command

ashishkel:
Update gettingstarted.md to correct the docker-compose command


dvdksn:
Closing as this change is incorrect. Use Compose V2 instead.

PR #18220, opened 2023-09-17
Update ubuntu.md

tenyuuk:
### Proposed changes

Added a service start command to the 'Install using the Apt repository' section to prevent confusion, mirroring the approach used in the 'Install from a package' section."

### DIFF
--- a/content/engine/install/ubuntu.md
+++ b/content/engine/install/ubuntu.md
@@ -165,6 +165,7 @@ Docker from the repository.
    `hello-world` image.
 
    ```console
+   $ sudo service docker start
    $ sudo docker run hello-world
    ```
 


content/engine/install/ubuntu.md
```diff
@@ -165,6 +165,7 @@ Docker from the repository.
    `hello-world` image.
 
    ```console
+   $ sudo service docker start
```
thaJeztah, line 168: The service should be running automatically after you installed the packages; what method did you use to install the packages?

tenyuuk: I use the command just above here in the `Latest` tab. It returns some symbolic error, but I didn't pay much attention to it. When I run the hello world command, it gives the following error: `docker: Cannot connect to the Docker daemon at unix:///var/run/docker.sock. Is the Docker daemon running?` It worked after I started the docker daemon service manually. I'm trying to install it on WSL2 Ubuntu 22.04."

thaJeztah: Wondering if part of the issue is WSL2; ISTR WSL2 doesn't use `systemd`, making it a non-standard setup for ubuntu (I'm not on a Windows machine though, so can't check 😅)

PR #18232, opened 2023-09-19
Update profiles.md

avpps:
Fixed profiles.md snippet that looks like copy paste issue.

### DIFF
--- a/content/compose/profiles.md
+++ b/content/compose/profiles.md
@@ -83,7 +83,7 @@ $ docker compose --profile frontend --profile debug up
 ```
 
 ```console
-$ COMPOSE_PROFILES=frontend,debug docker compose up
+$ COMPOSE_PROFILES=frontend,debug
 ```
 
 ## Auto-starting profiles and dependency resolution


content/compose/profiles.md
```diff
@@ -83,7 +83,7 @@ $ docker compose --profile frontend --profile debug up
 ```
 
 ```console
-$ COMPOSE_PROFILES=frontend,debug docker compose up
+$ COMPOSE_PROFILES=frontend,debug
```
thaJeztah, line 86: Looking at this example, I think this was intentional; this line does 2 things;

- set the `COMPOSE_PROFILES` env-var to `frontend,debug` (for the duration of the command)
- execute `docker compose up` (which now enables the `frontend` and `debug` profiles)

thaJeztah (changes requested):
I don't think this change is correct (see my comment)

content/compose/profiles.md
thaJeztah: Perhaps the docs could describe that though, and explain why the env-var is used like this (commonly this is something when you want a setting to only be used for a single command, without making the change permanent)

PR #18278, opened 2023-09-26
Add a missing line to start docker service before running test command in Ubuntu

theoctober19th:
The docker service needs to be started at first before running the command `sudo docker run hello-world` which is missing in the docs (for installation instructions in Ubuntu).

### Proposed changes

When installing docker via apt-get in Ubuntu for the first time, the docker service needs to be started before running the test command `sudo docker run hello-world`. Else, `docker` will complain docker daemon is not running.


theoctober19th:
Happens to be duplicate of #18220.

PR #18286, opened 2023-09-26
fix watch

aevesdocker:
Compose Watch is now GA and now means you only have to run one command. This PR fixes the contribute section to reflect the changes and updates the compose.yaml file as well. 

Screenshot 2023-09-26 at 14 40 29

### DIFF
--- a/compose.yaml
+++ b/compose.yaml
@@ -6,7 +6,7 @@ services:
     ports:
       - "1313:1313"
     entrypoint: ["hugo", "server", "--bind", "0.0.0.0"]
-    x-develop:
+    develop:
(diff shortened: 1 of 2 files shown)


dvdksn (approved):

stephaurelio:
Did `docker compose watch` work for you @aevesdocker to build the docs locally? When I try it, I just get a list of the options commands.

aevesdocker:
Ah! This is because I'm on the un-released version of DD, so this was a slightly premature merge of the PR... I'll send you the link to the new builds, or you can continue to use the old way until Thursday

PR #18296, opened 2023-09-27
Removed output from console text

IgorSamurovic:
When someone is copying everything in the console section in this instance, they are also copying the output. This is unnecessary and potentially confusing.


aevesdocker (approved):
Thank you @IgorSamurovic, a useful fix

PR #18334, opened 2023-09-30
fix: remove excessive quoting in engine/install

ftambara:
The removed quotes cause ShellCheck SC2027 and SC2046 warnings.


dvdksn (commented):
lgtm cc @thaJeztah

dvdksn (approved):

PR #18343, opened 2023-10-02
Remove text on how to set WSL autoMemoryReclaim.

ctalledo:
### Proposed changes

Since the WSL autoMemoryReclaim feature is experimental, the way it's configured is not yet set in stone (and has changed recently). Therefore remove the example text on how to configure it and instead refer users to the appropriate WSL webpage.

### Related issues (optional)

https://github.com/docker/docs/issues/18329


aevesdocker (approved):
Thank you @ctalledo

PR #18376, opened 2023-10-06
Update profiles.md

Coteh:
### Proposed changes

Was reading the docs and I noticed this passage. I believe that it should say start `dev` profile explicitly rather than `db` profile, as `db` is the service container in this case and `dev` is the profile that service `db` requires.


aevesdocker (approved):
Thanks for catching this @Coteh

PR #18394, opened 2023-10-11
Make it clearer the Raspian instructions are just for 32-bit installs

dimo414:
Although mentioned in the OS requirements section the existing note is easy to gloss over because readers are likely to assume that the "Raspberry Pi" page covers both 32-bit and 64-bit versions of the operating system.


dvdksn:
Agree that it's too subtle now, you're not the first person to point it out. I think I might just change the title of the page to make it absolutely clear...

dvdksn:
Fixed in #18395

PR #18403, opened 2023-10-11
Update _index.md

### LINKED ISSUE #18402: Switch the wording on Port mapping details in the documentation
### Is this a docs issue?

### Type of issue

Other

### Description

image

Why in code is the mapping `<host port>:<container port>` but then in the documentation next to it, it reads _"Map TCP port 80 in the container to port 8080 on the Docker host."_? I think it would be much easier to understand if we switched the wording of the documentation, and instead said _"Maps port 8080 on the host machine to TCP port 80 in the container"_, since this is the order it's specified in the code.

It seems like a nit-pick but for someone learning Docker, these little things can confuse people.

### Location

https://docs.docker.com/network/

### Suggestion

Reword the sentences in the docs!

BrettC5159:
Make the port mapping table easier to read.

### Proposed changes

Reworded the table text to make it read in the same order as the code to the left of it. This makes it easier for people scanning the documentation to understand it.

### Related issues (optional)

 #18402 
 Closes #18402

### DIFF
--- a/content/network/_index.md
+++ b/content/network/_index.md
@@ -38,12 +38,12 @@ This creates a firewall rule in the host,
 mapping a container port to a port on the Docker host to the outside world.
 Here are some examples:
 
-| Flag value                      | Description                                                                                                                                           |
-| ------------------------------- | ----------------------------------------------------------------------------------------------------------------------------------------------------- |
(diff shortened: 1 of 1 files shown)


dvdksn (approved):
Thanks!

PR #18412, opened 2023-10-12
Remove docs for already-removed workflow input

TBBle:
### Proposed changes

I removed the documentation of the `repo` parameter from the validate-upstream workflow.

### Related issues (optional)

#17427 removed the `repo` parameter, but left the docs behind.


TBBle:
Fixed in #18433

PR #18427, opened 2023-10-13
Add example Dockerfile

matthiasgubler:
### Proposed changes
The workaround describes the usage of mounted caches in the Dockerfile i.e. `--mount=type=cache`, but the example only shows the action and not how the Dockerfile should look like


dvdksn (approved):
LGTM

PR #18441, opened 2023-10-16
update titles for language guides

craig-osterhout:
### Proposed changes

The language-specific guides use the same titles for multiple topics and it's not great for search.
Some have already been updated or will be updated in pending PRs.

Update titles for the remaining guides: Rust, Java, and Go.

### Related issues (optional)

ENGDOCS-1731


stephaurelio (approved):
LGTM!

PR #18450, opened 2023-10-17
get-started: update bind mount topic

craig-osterhout:
### Proposed changes

Issue: Containers can only access bind mounts in shared locations based on the file sharing setting. If users create a bind mount in a non-shared location, an error message appears but it may not be obvious to users what the issue is.

Fix: In the get-started bind mount topic, add something to make users aware of where they can create a bind mount.

### Related issues (optional)

ENGDOCS-1730

### DIFF
--- a/content/get-started/06_bind_mounts.md
+++ b/content/get-started/06_bind_mounts.md
@@ -18,6 +18,11 @@ changes you make to the code immediately, as soon as you save a file. This means
 that you can run processes in the container that watch for filesystem changes
 and respond to them.
 
+To view which parts of your filesystem you can share with containers, see
+the file sharing setting for [Mac](../desktop/settings/mac.md/#file-sharing),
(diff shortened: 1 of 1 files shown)


content/get-started/06_bind_mounts.md
```diff
@@ -45,10 +45,17 @@ work.
 >
 > If you use Windows and want to use Git Bash to run Docker commands, see [Working with Git Bash](../desktop/troubleshoot/topics.md#working-with-git-bash) for syntax differences.
 
-1. Open a terminal and change directory to the `getting-started-app`
+1. In the Docker Desktop Dashboard, view the file sharing setting. This setting
```
dvdksn, line 48: should we add something like "make sure that your `getting-started-app` project is located under a file share" (really not sure how to phrase it but you know what I mean 😆)

dvdksn (approved):
LGTM!

PR #18455, opened 2023-10-18
redirects: add redirect for link used in docker init templates

lizzthabet:
### Proposed changes

This PR adds a redirect for a link used in templates generated by Docker Init based on a recommendation from @craig-osterhout. 🌟


craig-osterhout (approved):
Thank you. LGTM

PR #18456, opened 2023-10-18
Update release-notes.md

fpisani:
Line 31 had 4.24.1 as the version, but it is supposed to be 4.24.2.

### Proposed changes

Current links to GNU/Linux packages are wrong and lead to a page with a 403 Forbidden error. The number of the version is supposed to be 4.24.2, but the links say:
- docker-desktop-4.24.1-amd64.deb
- docker-desktop-4.24.1-x86_64.rpm
- docker-desktop-4.24.1-x86_64.pkg.tar.zst


aevesdocker (approved):
Thank you for catching this @fpisani!

PR #18464, opened 2023-10-18
Engine install freshness

stephaurelio:
### Proposed changes

Freshness updates for Docker Engine installation docs


content/includes/engine-install-methods-apt.md
```diff
@@ -0,0 +1,13 @@
+You can install Docker Engine in different ways, depending on your needs:
+
+- Docker Engine comes bundled with
+  [Docker Desktop for Linux](../../desktop/install/linux-install.md). This is
+  the easiest and quickest way to get started.
+
+- Set up and install Docker Engine from
+  [Docker's Apt repository](#install-using-the-repository).
```
stephaurelio, line 8: Note for reviewer: We inconsistently say `apt` and Apt in our docs, and I've seen it multiple ways in other docs. Thoughts on how we want to standardize this here?

content/engine/install/centos.md
```diff
@@ -33,12 +33,11 @@ CentOS versions:
 - CentOS 9 (stream)
 
 The `centos-extras` repository must be enabled. This repository is enabled by
-default, but if you have disabled it, you need to
-[re-enable it](https://wiki.centos.org/AdditionalResources/Repositories).
```
stephaurelio, line 37: Note for reviewer: this link is a 404 and didn't include much helpful info about this topic. I opted to not explain how to do this, but wanted feedback on if we should include the commands for how to enable repositories using `yum` or `dnf`.

stephaurelio: Related issue: https://github.com/docker/docs/issues/18364

dvdksn: I'm OK with removing the link!

content/includes/engine-install-methods-apt.md
dvdksn: I've noticed this as well and deferred making my mind up 🙈 This is an acronym for Advanced Packaging Tool, so technically we should use APT. But you only really care about this tool as a CLI invocation, much like `curl`. So IMO, we're better off using `apt` for this (and similar utilities) - let's do that

dvdksn: Apt is definitely wrong and my bad - not sure what I was thinking

dvdksn (changes requested):
I'd prefer if we don't move the install methods to an include. While reuse is good, these docs get a lot of contributions, and the include complicates this.

content/includes/engine-install-methods-apt.md
stephaurelio: K will update to use `apt` 👍

content/desktop/install/debian.md
```diff
@@ -50,8 +50,8 @@ To install Docker Desktop successfully, you must:
 
 Recommended approach to install Docker Desktop on Debian:
 
-1. Set up Docker's Apt repository.
-   See [Install using the Apt repository, step 1](../../engine/install/debian.md#install-using-the-repository).
+1. Set up Docker's 'apt' repository.
+   See [Install using the 'apt' repository, step 1](../../engine/install/debian.md#install-using-the-repository).
```
dvdksn, line 54:
```suggestion
1. Set up Docker's `apt` repository.
   See [Install using the `apt` repository, step 1](../../engine/install/debian.md#install-using-the-repository).
```

dvdksn (approved):
LGTM thanks!

PR #18465, opened 2023-10-18
fix access token links

craig-osterhout:
### Proposed changes

https://github.com/docker/docs/pull/18404 and https://github.com/docker/docs/pull/18400 were tested and then some files were moved around before they were merged to main. This caused the links to pass and merge into main, but fail when merging to published.


stephaurelio (approved):
LGTM

PR #18487, opened 2023-10-22
Fix typo in _index.md

Amnish04:
I think there is a typo in the docs where they talk about use the `--mount` option for the cache.

Changed
`RUN --mount type=cache`
to
`RUN --mount=type=cache`


dvdksn:
It's not a typo

PR #18499, opened 2023-10-24
Set correct default value for fluentd-buffer-limit

mstinsky:
Docker is setting a higher (1048576) default buffer limit than the fluent logger library (8192). 

Here is the relevant code in the latest docker release:
https://github.com/moby/moby/blob/v24.0.6/daemon/logger/fluentd/fluentd.go#L39


thaJeztah:
nice catch; looks indeed that I added a `docs/revisit` label on the PR, but that never happened 🙈 https://github.com/moby/moby/pull/18041#event-527662229

PR #18509, opened 2023-10-24
chore: add --rm in docker run hello-world command

unitythemaker:
### Proposed changes

I did add `--rm` in the `sudo docker run hello-world` command. I do that every time and wanted to edit it this time. I do not think people will need the hello-world container later on so I thought we can just make it destroy itself after it is done running.

### Related issues (optional)

None


dvdksn:
This doesn't seem necessary to me

PR #18519, opened 2023-10-25
Add filtering section on docker system prune

### LINKED ISSUE #18518: Missing filter statement on system prune
### Is this a docs issue?

### Type of issue

Other

### Description

The pruning unused objects page does not make mention of the filter CLI command, which is supported per https://docs.docker.com/engine/reference/commandline/system_prune/ but only for API 1.28+. 

### Location

https://docs.docker.com/config/pruning/

### Suggestion

Reference docs should note this and link to the page similarly to how it's done for the other 'prune' commands.

thedoc31:
Docker system prune is allowed for API 1.28+, as shown in the doc https://docs.docker.com/engine/reference/commandline/system_prune/, but this is not shown in the Prune unused objects page.

### Proposed changes

Copied and pasted the "filtering" section from an above prune command and modified to make it accurate for system prune. Also noted that API 1.28+ is required.

### Related issues (optional)

Closes #18518

### DIFF
--- a/content/config/pruning.md
+++ b/content/config/pruning.md
@@ -175,3 +175,15 @@ Are you sure you want to continue? [y/N] y
 
 By default, you are prompted to continue. To bypass the prompt, use the `-f` or
 `--force` flag.
+
+By default, all unused containers, networks, images (both dangling and unreferenced)
+are removed. If you are using docker {{< badge color=blue text="API 1.28+" >}}, you can limit the scope using the 
+`--filter` flag. For instance, the following command removes items older than 24 hours:
+
+```console
+$ docker system prune --filter "until=24h"
+```
+
+Other filtering expressions are available. See the
+[`docker system prune` reference](../engine/reference/commandline/system_prune.md)
+for more examples.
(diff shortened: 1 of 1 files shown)


content/config/pruning.md
```diff
@@ -174,4 +174,16 @@ Are you sure you want to continue? [y/N] y
 ```
 
 By default, you are prompted to continue. To bypass the prompt, use the `-f` or
-`--force` flag.
\ No newline at end of file
+`--force` flag.
+
+By default, all unused containers, networks, images (both dangling and unreferenced)
+are removed. If you are using docker {{< badge color=blue text="API 1.28+" >}}, you can limit the scope using the 
+`--filter` flag. For instance, the following command removes items older than 24 hours:
```
dvdksn, line 181: Thanks! I think we can remove the version disclaimer, API v1.28 is very old.

```suggestion
By default, all unused containers, networks, images (both dangling and unreferenced)
are removed. You can limit the scope using the 
`--filter` flag. For instance, the following command removes items older than 24 hours:
```

dvdksn (approved):
thank you

PR #18537, opened 2023-10-27
build: set missing frontmatter for remote pages

crazy-max:
### Proposed changes

follow-up https://github.com/docker/docs/pull/18473

Set missing frontmatter for remote pages. Values transposed from lab branch: https://github.com/docker/docs/blob/lab/build


dvdksn (approved):
♥️

PR #18557, opened 2023-10-31
Remove wrong statement from fluentd logging driver

mstinsky:
Since docker 20.10 dual logging is supported which enables logging drivers to use the `docker logs` command.
This is already stated in the following section: https://docs.docker.com/config/containers/logging/configure/#supported-logging-drivers

This PR just removes the 'wrong' line from the fluentd logging driver documentation.


dvdksn (approved):
LGTM, ptal @vvoland

vvoland (approved):

PR #18567, opened 2023-11-01
refresh diagrams

aevesdocker:
Closes https://github.com/docker/docs/pull/18314 fixes merging issues and improves color contrast and converts to webp


aevesdocker:
@dvdksn I'm ready to close this one out now. If you get a spare second this week, I'd appreciate a review. 
I think this covers most of the images, perhaps they're a lil big on certain pages but I'd prefer to do that in a follow-up

dvdksn:
Not a full review but some quick finds. 🤔 

Some images appear kind of blurry, or with aliased lines. Like they've been compressed too aggressively.

image image image

Many were kind of blurry before too, so it's probably not worse 😅 but still, good to keep in mind for follow-ups (if we're updating these anyway we might as well make them sharp 🗡️)

This one seems to be missing some lines (and the lines resembling the directory tree look strangely hand-drawn) https://deploy-preview-18567--docsdocker.netlify.app/storage/storagedriver/btrfs-driver/#image-and-container-layers-on-disk

And then it looks like the old png/jpeg images in storage/storagedrivers weren't deleted. I only see the new webp ones added

aevesdocker:
I've deleted the storage drivers png/jpegs, and fixed the obviously blurry ones. 
I think the slight blur/lack of crispness is not from the images themselves but from the zoomable image component, if you look at this page https://docs.docker.com/desktop/use-desktop/, there is the same issue, but when you expand the image, it's fine. Also here https://docs.docker.com/desktop/dev-environments/, https://docs.docker.com/desktop/extensions/, https://docs.docker.com/scout/dashboard/

aevesdocker:
Maybe the solution is to have the zoomable image be an optional component, so you can choose to have it if the image/diagram is super detailed. But a lot of the images don't need it

dvdksn:
I don't think it has anything to do with the modal or the downscaling. View the images directly and they still appear quite blurry some of them. Compare the font rendering in these:

- https://docs.docker.com/desktop/images/dev-env.PNG
- https://docs.docker.com/scout/images/dd-image-view.png

dvdksn (approved):

PR #18615, opened 2023-11-08
ENGDOCS-1803

aevesdocker:
slight tweak to compose banner to shine a light on v2 rather than starting with the negatives of v1


stephaurelio (approved):
LGTM

PR #18626, opened 2023-11-10
light and dark mode images

aevesdocker:
PR adds the functionality for docs to support light and dark mode image switching


dvdksn (commented):
a couple out-loud thoughts on how we could improve this:

- Keep it simple and use CSS to hide images for the unselected theme (still loads all variants, but doesn't cause a repaint until the user changes theme). With this we wouldn't need any script changes, just a shortcode with the appropriate markup. And we could add lazy loading to make it less expensive on initial load.
- Don't load an image at all (use a blank placeholder, e.g. base64 data string) and update the source only after the dom content is loaded. This would still trigger a repaint I think but it would mean that only the appropriate format is loaded.

No solution is going to be perfect from a performance standpoint because the site is fully static and we can't predetermine which image to load beforehand. So we should probably try to check which alternative is better/worse. It's not a massive hit but it all adds up.

curious to hear if @craig-osterhout has any ideas

PR #18630, opened 2023-11-10
added -y for apt install for ubuntu version

anvarknian:


dvdksn:
This is not a script so I don't view this as necessary

PR #18634, opened 2023-11-10
build: fix gha cache scope

crazy-max:
### Proposed changes

* fixes https://github.com/docker/build-push-action/issues/706

`scope` is an attribute defaulting to `buildkit` (https://github.com/moby/buildkit#github-actions-cache-experimental) while there is also the scope of the JWT token. So you have the token's scope and on top of it the cache scope.


dvdksn (approved):
:shipit:

PR #18646, opened 2023-11-12
Update ubuntu.md

pawelpt:
changes applied to trusted.gpg.d instead of keyrings


dvdksn:
Hello. Adding a key to `/etc/apt/trusted.gpg.d` is insecure. See accepted answer in https://stackoverflow.com/questions/68992799/warning-apt-key-is-deprecated-manage-keyring-files-in-trusted-gpg-d-instead

PR #18648, opened 2023-11-12
Update _index.md

fussybeaver:
Add bollard as Rust SDK.

Bollard is currently the most actively maintained docker SDK in the Rust programming language, so it should be mentioned on this page.

### DIFF
--- a/content/engine/api/sdk/_index.md
+++ b/content/engine/api/sdk/_index.md
@@ -205,6 +205,7 @@ file them with the library maintainers.
 | Perl                  | [Eixo::Docker](https://github.com/alambike/eixo-docker)                     |
 | PHP                   | [Docker-PHP](https://github.com/docker-php/docker-php)                      |
 | Ruby                  | [docker-api](https://github.com/swipely/docker-api)                         |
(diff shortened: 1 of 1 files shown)


dvdksn (approved):
LGTM

We should probably remove the other rust alternatives from this list. cc @thaJeztah

thaJeztah:
> We should probably remove the other rust alternatives from this list. cc @thaJeztah

These overviews have always been a bit tricky; they were really added to help discovery, and to list some possible options. These lists tend to be hard to keep up-to-date (projects come and go, or get unmaintained). Overall, it's ok to list multiple alternatives, as we don't extensively curate the list, and there may be multiple alternatives, all "valid" options to choose from.

We should be very clear on intent though, because we don't do a full review on these (other than looking if they're _somewhat_ widely used, and are actively maintained), so it's hard to officially "endorse" any of them. I see there's some wording we added for that, which hopefully covers that, but suggestions to improve that are of course welcome;

> There are a number of community supported libraries available for other languages. They have not been tested by Docker, so if you run into any issues, file them with the library maintainers.

PR #18657, opened 2023-11-14
Update configure-ci-cd.md

wmstantoniii:
Create multi-platform build which can be downloaded and run on an M1 Mac.


dvdksn (commented):
I don't think this is in scope for this tutorial, at least not in its current form cc @craig-osterhout

craig-osterhout:
Thanks for sharing this.
I'm trying to determine if this is necessary. In the next section, test your deployment, did you receive an error when running the amd64 image?

PR #18670, opened 2023-11-14
WIP: Engdocs 1814

aevesdocker:

### DIFF
--- a/content/compose/environment-variables/best-practices.md
+++ b/content/compose/environment-variables/best-practices.md
@@ -1,0 +1,22 @@
+---
+title: Best practices for working with environment variables in Docker Compose
+description: Explainer on the ways to set, use and manage environment variables in
+  Compose
+keywords: compose, orchestration, environment, env file
+---
+
+1. Handle Sensitive Information Securely:
+   Be cautious about including sensitive data in environment variables. Consider using [Secrets](../use-secrets.md) for managing sensitive information.
+
+2. Understand Environment variable precedence:
+    Be aware of how Docker Compose handles the precedence of environment variables from different sources (`.env` files, shell variables, Dockerfiles).
+
+3. Use specific environment files:
+   Consider how your application adapts to different environments. for example development, testing, production, and use different `.env` files as needed.
+
+4. Know interpolation:
+   Understand how interplation works within Docker Compose files for dynamic configurations.
+
+5. Command line overrides:
+    Be aware that you can override environment variables from the command line when starting containers, useful for testing or temporary changes.
+
--- a/content/compose/environment-variables/envvars-precedence.md
+++ b/content/compose/environment-variables/envvars-precedence.md
@@ -7,14 +7,14 @@ aliases:
 - /compose/envvars-precedence/
 ---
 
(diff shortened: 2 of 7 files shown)


ndeloof (approved):

glours (approved):
LGTM

ndeloof (approved):

content/compose/environment-variables/best-practices.md
```diff
@@ -0,0 +1,27 @@
+---
+title: Best practices for working with environment variables in Docker Compose
+description: Explainer on the best ways to set, use, and manage environment variables in
+  Compose
+keywords: compose, orchestration, environment, env file, environment variables
+---
+
+#### Handle sensitive information securely
```
dvdksn, line 8: The heading levels on this page are off, these should be h2s

aevesdocker: Yes this was done on purpose. The page itself doesn't have a lot of content so having h2s would look out of place. We do a similar thing in the FAQs, so figured a little bit more rule breaking wouldn't be a problem

content/compose/environment-variables/set-environment-variables.md
```diff
@@ -1,13 +1,15 @@
 ---
-title: Ways to set environment variables in Compose
-description: How to set, use and manage environment variables in Compose
+title: Ways to set environment variables with Compose
+description: How to set, use ,and manage environment variables with Compose
```
dvdksn, line 3:
```suggestion
description: How to set, use, and manage environment variables with Compose
```

content/compose/environment-variables/envvars-precedence.md
```diff
@@ -65,7 +65,7 @@ Each column represents a context from where you can set a value, or substitute i
 
 The columns `Host OS environment` and `.env file` is listed only as an illustration lookup. In reality, they don't result in a variable in the container by itself.
 
-Each row represents a combination of contexts where `TAG` is set, substituted, or both.
+Each row represents a combination of contexts where `TAG` is set, substituted, or both. The **Result** column indicates the final value for TAG in each scenario.
```
dvdksn, line 68:
```suggestion
Each row represents a combination of contexts where `TAG` is set, substituted, or both. The **Result** column indicates the final value for `TAG` in each scenario.
```

content/compose/environment-variables/set-environment-variables.md
```diff
@@ -44,12 +48,31 @@ services:
...
+- You can use multiple `.env` files in your `compose.yml` with the [`env_file` attribute](../compose-file/05-services.md#env_file), and Docker Compose reads them in the order specified. If the same variable is defined in multiple files, the last definition takes precedence:
+  ```yaml
+  services:
+    webapp:
+      image: my-webapp-image
+      env_file:
+        - .env
+        - .env.override
+  ```
+- You can place your `.env` file in a location other than the root of your project's directory, and then use one of the following methods so Compose can navigate to it:
```
dvdksn, line 71: I think there might be some nuance missing here. The .env file can be referenced from an arbitrary location with the env_file attribute or --env-file flag (as described on line 73) but with the --file flag, I assume that the .env file needs to be placed adjacent to the compose file referenced by --file. You couldn't point to a .env file with -f

dvdksn (approved):

PR #18680, opened 2023-11-15
Update multi-stage.md

julianharty:
Added the wildcard to the filter so that all the matching images are shown. The command without the wildcard didn't match either of the newly named images.

### Proposed changes

Added the wildcard and quotes so that the newly named images would be listed in the results. I didn't see either of the matching images without the wildcard.


dvdksn (approved):
LGTM, thank you @julianharty

PR #18689, opened 2023-11-16
Update containerize.md - correction of path to ../activate

MiaLisen:
The path to .//activate was incorrect. ( .venv/bin/activate). I There is no bin folder in this repo. I've corrected it to source .venv/Scripts/activate that is the existing path.


dvdksn:
cc @craig-osterhout - looks like this path is different for win/unix. Maybe add a tab here for both versions?

PR #18703, opened 2023-11-17
Update proxy.md

WestFarmer:
there is no `default` key under `proxies` in daemon.json at all!, this sentence is misleading, especially for non-english speaker.

I guess the author means 
> The `default` key under `proxies` in `~/.docker/config.json`


dvdksn (approved):
You're totally right, thanks for catching and fixing this 🙈

PR #18739, opened 2023-11-21
fix(healthcheck): extend start_period to 2m to clarify initialization

yepayepayepa:
### Proposed changes

Revised the start_period in the Docker healthcheck example to 2 minutes to make sure the property's role is clear. This change highlights the intended use of start_period as a grace period for container initialization. Setting start_period to a duration longer than the interval (1 minute and 30 seconds) makes it more evident that health check failures during this initial period will not count towards the retries limit.


dvdksn (approved):
LGTM, thanks @yepayepayepa

PR #18743, opened 2023-11-21
redirects: add urls used by docker init files

lizzthabet:
### Proposed changes

This PR adds a few more short url redirects for docs pages that are referenced in files generated by Docker Init. If the docs page structure changes in the future, we can change the redirects to point to the current page, while the url that's included in the Init-generated files remains the same.


lizzthabet:
@craig-osterhout, would you mind taking a look and merging if the changes look good to you? thank you!

dvdksn (approved):
LGTM

PR #18763, opened 2023-11-22
update docs to reflect it is docker-compose not docker-compose-plugin

thams:
### Proposed changes

Updated docs to reflect that the plugin appears to be named docker-compose, not docker-compose-plugin


thams:
It appears the existing documentation is correct; I had been running an old version of docker compose.

PR #18776, opened 2023-11-24
Fixed referencing to the wrong brach /s/master/main

tippexs:
In the README.md there is still a reference to the `master` branch. Renamed this into main.


aevesdocker (approved):
Thank you for catching this @tippexs

Issue #18666, opened 2023-11-14
Clarify that compose.override.yml is also applied


kwint:
### Is this a docs issue?

- [X] My issue is about the documentation content or website

### Type of issue

Information is incorrect

### Description

> The configuration in the docker-compose.override.yml file is applied over and in addition to the values in the docker-compose.yml file.

this is also true for `compose.yml` and `compose.override.yml` files.

### Location

https://docs.docker.com/compose/reference/

### Suggestion

Might be good to use `compose.yml` everywhere?

PR #18780, opened 2023-11-24
Engdocs 1871

aevesdocker:
Closes https://github.com/docker/docs/issues/18750
Closes https://github.com/docker/docs/issues/18666

### DIFF
--- a/content/compose/reference/_index.md
+++ b/content/compose/reference/_index.md
@@ -8,20 +8,7 @@ aliases:
 
 {{< include "compose-eol.md" >}}
 
-This page provides the usage information for the `docker compose` Command.
-
-> Important
->
-> The new Compose V2, which supports the `compose` command as part of the Docker
-> CLI, is now available.
->
-> Compose V2 integrates compose functions into the Docker platform, continuing
-> to support most of the previous `docker-compose` features and flags. You can
-> run Compose V2 by replacing the hyphen (`-`) with a space, using `docker compose`,
-> instead of `docker-compose`.
-{ .important }
-
-For more information about Docker Compose V2 GA, see the blog post [Announcing Compose V2 General Availability](https://www.docker.com/blog/announcing-compose-v2-general-availability/).
+This page provides usage information for the `docker compose` command.
 
 ## Command options overview and help
 
@@ -36,7 +23,7 @@ Define and run multi-container applications with Docker.
 Options:
       --ansi string                Control when to print ANSI control characters ("never"|"always"|"auto") (default "auto")
       --compatibility              Run compose in backward compatibility mode
-      --env-file stringArray       Specify an alternate environment file.
+      --env-file stringArray       Specify an alternate environment file
   -f, --file stringArray           Compose configuration files
       --parallel int               Control max parallelism, -1 for unlimited (default -1)
       --profile stringArray        Specify a profile to enable
@@ -91,10 +78,10 @@ add to their predecessors.
 For example, consider this command line:
 
 ```console
-$ docker compose -f docker-compose.yml -f docker-compose.admin.yml run backup_db
+$ docker compose -f compose.yml -f compose.admin.yml run backup_db
 ```
 
-The `docker-compose.yml` file might specify a `webapp` service.
+The `compose.yml` file might specify a `webapp` service.
 
 ```yaml
 webapp:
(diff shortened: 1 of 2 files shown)


content/compose/reference/_index.md
```diff
@@ -36,7 +23,7 @@ Define and run multi-container applications with Docker.
 Options:
       --ansi string                Control when to print ANSI control characters ("never"|"always"|"auto") (default "auto")
       --compatibility              Run compose in backward compatibility mode
```
dvdksn, line 25: The `--dry-run` flag is missing

content/compose/reference/_index.md
```diff
@@ -36,7 +23,7 @@ Define and run multi-container applications with Docker.
...
-      --env-file stringArray       Specify an alternate environment file.
+      --env-file stringArray       Specify an alternate environment file
   -f, --file stringArray           Compose configuration files
       --parallel int               Control max parallelism, -1 for unlimited (default -1)
       --profile stringArray        Specify a profile to enable
```
dvdksn, line 29: The `--progress` flag is missing

content/compose/reference/_index.md
```diff
@@ -36,7 +23,7 @@ Define and run multi-container applications with Docker.
 Options:
       --ansi string                Control when to print ANSI control characters ("never"|"always"|"auto") (default "auto")
       --compatibility              Run compose in backward compatibility mode
-      --env-file stringArray       Specify an alternate environment file.
+      --env-file stringArray       Specify an alternate environment file
```
dvdksn, line 26: This period is there in the cobra help string so we probably want to update it in the code as well

content/compose/reference/_index.md
```diff
@@ -126,12 +113,12 @@ relative to the current working directory.
```
dvdksn, line 113: unrelated to the changes here, but we could probably use an example here for reading the compose file from stdin with a heredoc

content/compose/reference/_index.md
```diff
@@ -105,7 +92,7 @@ webapp:
     - "/data"
 ```
 
-If the `docker-compose.admin.yml` also specifies this same service, any matching
+If the `compose.admin.yml` also specifies this same service, any matching
 fields override the previous file. New values, add to the `webapp` service
 configuration.
```
dvdksn, line 97: Not related to this change, but this description is inaccurate. Some fields are merged, not overridden. Labels, for example:

```yaml
# compose.yaml
services:
  webapp:
    build:
      context: .
      labels: 
        com.example.foo: "foo"
```

```yaml
# overrides.yaml
services:
  webapp:
    build:
      context: .
      labels: 
        com.example.name: "Alice"
```

```console
$ docker compose -f compose.yaml -f overrides.yaml build
$ docker image inspect compose-webapp --format json | jq '.[].Config.Labels'
{
  "com.docker.compose.project": "compose",
  "com.docker.compose.service": "webapp",
  "com.docker.compose.version": "2.23.0",
  "com.example.foo": "foo",
  "com.example.name": "Alice"
}
```

`buildx bake` has similar behavior and I've described how duplicate fields either merge with or override previous definitions in the bake file reference, see here: https://docs.docker.com/build/bake/reference/#file-format

content/compose/reference/_index.md
```diff
@@ -126,12 +113,12 @@ relative to the current working directory.
 
 The `-f` flag is optional. If you don't provide this flag on the command line,
 Compose traverses the working directory and its parent directories looking for a
-`docker-compose.yml` and a `docker-compose.override.yml` file. You must supply
-at least the `docker-compose.yml` file. If both files are present on the same
+`compose.yml` and a `compose.override.yml` file. You must supply
+at least the `compose.yml` file. If both files are present on the same
```
dvdksn, line 117: Both `docker-compose.yml` and `compose.yml` and corresponding override files are supported. Compose has some logic for deciding the lookup order of these files, which is what I think #18666 is asking for. How do these variants interact?

Again the bake file has a similar behavior so you can take a look at how I did it there: https://docs.docker.com/build/bake/reference/#file-format

dvdksn (changes requested):
I have one specific comment related to the change, which is about file lookup order. The current change doesn't address the fact that you can have multiple compose filenames and they take presedence in different ways.

Then I had several comments on things unrelated to these changes but that I noticed in passing.

aevesdocker:
Thanks @dvdksn. It's on my radar that this page needs some TLC. I've addressed your main blocking comment. The rest I'll do in a follow-up

dvdksn (approved):

PR #18794, opened 2023-11-27
add php language-specific guide

craig-osterhout:
### Proposed changes

Add PHP language-specific guide based on init template in 4.26.

### Related issues (optional)

ENGDOCS-1634

https://deploy-preview-18794--docsdocker.netlify.app/language/php/


content/language/php/run-tests.md
```diff
@@ -0,0 +1,117 @@
...
+COPY ./tests /var/www/html/tests
+RUN mv "$PHP_INI_DIR/php.ini-development" "$PHP_INI_DIR/php.ini"
+COPY --from=dev-deps app/vendor/ /var/www/html/vendor
+
+FROM base as final
+RUN mv "$PHP_INI_DIR/php.ini-production" "$PHP_INI_DIR/php.ini"
+COPY --from=prod-deps app/vendor/ /var/www/html/vendor
+USER www-data
+
+FROM development as test
```
mikesir87, line 79: I wouldn't put the tests as the last stage here, but let the `final` stage be the last. Otherwise, you're going to be required to specify `--target=final` every time you do a build. Otherwise, a `docker build -t php-app .` will build an image ready to do tests.

content/language/php/run-tests.md
```diff
@@ -0,0 +1,117 @@
...
+
+FROM composer:lts as dev-deps
+WORKDIR /app
+RUN --mount=type=bind,source=./composer.json,target=composer.json \
+    --mount=type=bind,source=./composer.lock,target=composer.lock \
+    --mount=type=cache,target=/tmp/cache \
+    composer install --no-interaction
+
+FROM php:8.2-apache as base
+RUN apt-get update && apt-get install -y libpq-dev && docker-php-ext-install pdo pdo_pgsql
```
mikesir87, line 66: You might want to go ahead and clear the apt cache (`rm -rf /var/lib/apt/lists/*`) to ensure we're aligning with [our own best practices](https://docs.docker.com/develop/develop-images/instructions/#apt-get).

content/language/php/develop.md
```diff
@@ -0,0 +1,354 @@
+---
```
mikesir87, line 1: Just one thought (so feel free to throw out/ignore)... a _lot_ of PHP developers will probably be more familiar with using MySQL/MariaDB, as it's the only supported database for Wordpress and the recommended db for Drupal. If you used MySQL, you can easily add another service to the Compose file for phpMyAdmin, which is yet another service PHP devs use quite often, many hosting services provide as part of hosting, etc. It could help show the ability to bolt on additional containers to the stack to make it easier to do development/debug data. Just a thought...

mikesir87 (commented):
Fantastic writeup! Added a few comments, but feel free to adopt or ignore them 😄

content/language/php/develop.md
lizzthabet: ah, that is a great point! we currently use the postgres image as a commented out example in the generated compose file for init, but it wouldn't be hard to swap it out for a more relevant image. (i also don't see any issues with using a different image in the docs than whatever's currently used by init and will leave that up to craig.)

lizzthabet (approved):
🎉 this looks great. i noted some minor thoughts + questions.

content/language/php/deploy.md
```diff
@@ -0,0 +1,140 @@
...
+1. In a terminal, navigate to the `docker-php-sample` directory
+   and deploy your application to Kubernetes.
+
+   ```console
+   $ kubectl apply -f docker-php-kubernetes.yaml
+   ```
+
+   You should see output that looks like the following, indicating your Kubernetes objects were created successfully.
+
+   ```shell
```
dvdksn, line 89: I think `shell` == `bash` in hugo's syntax highlighting (chroma). I think we should use `text` for these instead
https://gohugo.io/content-management/syntax-highlighting/#list-of-chroma-highlighting-languages

content/language/php/run-tests.md
```diff
@@ -0,0 +1,117 @@
...
+COPY --from=dev-deps app/vendor/ /var/www/html/vendor
+
+FROM development as test
+WORKDIR /var/www/html
+RUN ./vendor/bin/phpunit tests/HelloWorldTest.php
+
+FROM base as final
+RUN mv "$PHP_INI_DIR/php.ini-production" "$PHP_INI_DIR/php.ini"
+COPY --from=prod-deps app/vendor/ /var/www/html/vendor
+USER www-data
```
dvdksn, line 81: I think you received a comment about moving the "test" stage to before the final stage to not override the default. As an alternative solution, we could add the test stage to the end of the Dockerfile, but also add a final "FROM final" line to force `final` to be the default stage.

```dockerfile
FROM development as test
WORKDIR /var/www/html
RUN ./vendor/bin/phpunit tests/HelloWorldTest.php

FROM final
```

I think this could help make this instruction more clear. ("Add the following to your Dockerfile")

content/language/php/run-tests.md
```diff
@@ -0,0 +1,117 @@
...
+```
+
+Run the following command to build an image using the test stage as the target and view the test results. Include `--progress=plain` to view the build output, `--no-cache` to ensure the tests always run, and `--target test` to target the test stage.
+
+```console
+$ docker build -t php-docker-image-test --progress=plain --no-cache --target test .
```
dvdksn, line 87: I think we should avoid mixing the space and `=` separator for flags, at least in a single command

```suggestion
Run the following command to build an image using the test stage as the target and view the test results. Include `--progress plain` to view the build output, `--no-cache` to ensure the tests always run, and `--target test` to target the test stage.

```console
$ docker build -t php-docker-image-test --progress plain --no-cache --target test .
```

content/language/php/deploy.md
```diff
@@ -0,0 +1,140 @@
...
+
+2. Make sure everything worked by listing your deployments.
+
+   ```console
+   $ kubectl get deployments
+   ```
+
+   Your deployment should be listed as follows:
+
+   ```shell
```
dvdksn, line 102:
```suggestion
   ```text
```

dvdksn (commented):
great work @craig-osterhout - I found some small things, mostly formatting stuff

content/language/php/run-tests.md
craig-osterhout: I think I'll add the highlighted lines here and leave the test stage above final. Otherwise, I need to briefly explain the last line, which might take focus away from the main objective.

dvdksn (approved):

PR #18801, opened 2023-11-28
Update buildx setup instructions to use `install: true`

nicks:
### Proposed changes

Currently, `install` defaults to false.

Setting `install` to true makes `docker build` use the new builder. I think this is probably the behavior most people want.

### DIFF
--- a/content/build/hydrobuild.md
+++ b/content/build/hydrobuild.md
@@ -236,6 +236,7 @@ jobs:
           version: "lab:latest"
           driver: cloud
           endpoint: "<ORG>/default"
+          install: true
       - name: Build and push
         uses: docker/build-push-action@v5
         with:


nicks:
@crazy-max i also wonder if it makes sense to change the defaults on the upstream action? but i guess that's a separate discussion.

crazy-max:
I don't think this is necessary, since Docker 23 `docker build` defaults to `docker buildx build`, see https://docs.docker.com/build/architecture/.

> @crazy-max i also wonder if it makes sense to change the defaults on the upstream action? but i guess that's a separate discussion.

In contrary I think we should remove the `install` input in the `setup-buildx-action` as it doesn't make a lot of sense anymore.

nicks:
Hmmm...in my tests, it's still necessary. I'll send you what some logs off-list

crazy-max (approved):

dvdksn (approved):

PR #18808, opened 2023-11-29
get-started: add landing page w/ walkthroughs

craig-osterhout:
### Proposed changes

Add initial get started landing page w/ walkthroughs and rearrange toc.
Update prereq in walkthroughs intro to emphasize that the latest version is required.

### Related issues (optional)

ENGDOCS-1761
ENGDOCS-1881

https://deploy-preview-18808--docsdocker.netlify.app/guides/get-started/


dvdksn (commented):
This is great. I wonder if we could make some of the images a little bit larger? Both in terms of size on page and resolution. Some of them are illegible unless maximized, and/or appear somewhat cramped because they're cropped very hard. WDYT?

image
image

craig-osterhout:
Re-captured some of the shots and adjusted the widths. Some still look a bit awkward. I'll create a follow up ticket to come back to it.

PR #18812, opened 2023-11-30
apt-get can remove more than one package in one run

ratopi:
### Proposed changes

Keep things simple.

apt-get can remove more than one package in one run.


dvdksn:
This for loop is intentional, see: https://github.com/docker/docs/issues/17405

PR #18868, opened 2023-12-06
Updating install location and note to match Microsoft docs

cbittencourt:
The $env:ProgramFiles/docker folder does not exist when installing docker on Windows Server using the Microsoft documentation and install script: https://learn.microsoft.com/en-us/virtualization/windowscontainers/quick-start/set-up-environment?tabs=dockerce#windows-server-1

### DIFF
--- a/content/compose/install/standalone.md
+++ b/content/compose/install/standalone.md
@@ -50,12 +50,11 @@ on Microsoft Windows Server and want to install Docker Compose.
 3. Run the following command to download the latest release of Compose ({{% param "compose_version" %}}):
 
     ```powershell
-     Start-BitsTransfer -Source "https://github.com/docker/compose/releases/download/{{% param "compose_version" %}}/docker-compose-windows-x86_64.exe" -Destination $Env:ProgramFiles\Docker\docker-compose.exe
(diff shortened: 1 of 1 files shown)


content/compose/install/standalone.md
```diff
@@ -50,12 +50,11 @@ on Microsoft Windows Server and want to install Docker Compose.
 3. Run the following command to download the latest release of Compose ({{% param "compose_version" %}}):
 
     ```powershell
-     Start-BitsTransfer -Source "https://github.com/docker/compose/releases/download/{{% param "compose_version" %}}/docker-compose-windows-x86_64.exe" -Destination $Env:ProgramFiles\Docker\docker-compose.exe
+     Start-BitsTransfer -Source "https://github.com/docker/compose/releases/download/{{% param "compose_version" %}}/docker-compose-windows-x86_64.exe" -Destination $env:windir\System32\docker-compose.exe
```
thaJeztah, line 53: Not sure if this is correct; this would change the compose CLI to be installed in the System32 directory (e.g. `C:\Windows|System32`), which doesn't look like the correct location for this.

thaJeztah (changes requested):
I don't think this change is correct (see comments)

PR #18875, opened 2023-12-07
Update 02_our_app.md

torressam333:
### Proposed changes

When going through the get started example I ran into an issue with running the container with the -d flag and the container would never actually run. Running the container in interactive mode helped for me to be able to run the container successfully. I added supplemental documentation to help other users by providing them a 2nd way to run a docker container if they run into an issue with the first method.

My machine specs: `Darwin LAP09162TN 23.1.0 Darwin Kernel Version 23.1.0: Mon Oct  9 21:27:27 PDT 2023; root:xnu-10002.41.9~6/RELEASE_X86_64 x86_64`

### Related issues (optional)

N/A

### DIFF
--- a/content/get-started/02_our_app.md
+++ b/content/get-started/02_our_app.md
@@ -134,13 +134,25 @@ Now that you have an image, you can run the application in a container using the
    The command publishes the container's port 3000 to `127.0.0.1:3000` (`localhost:3000`) on the host.
    Without the port mapping, you wouldn't be able to access the application from the host.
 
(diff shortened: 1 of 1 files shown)


dvdksn:
What you encountered was actually the expected behavior - running a container in detached mode prints the containerid. The container still runs as you would see if you ran the docker ps command.

I do see how this can be confusing for someone getting started. But changing this behavior is probably not gonna happen, as I'm sure there are lots of scripts that rely on this. I guess we could rewrite it to be clearer that this is expected. //cc @craig-osterhout

PR #18877, opened 2023-12-07
get-started: remove run sample app without docker

craig-osterhout:
### Proposed changes

Remove the optional sections that ask the user to run sample apps without Docker.

Some users may install the tools without realizing they have another version installed, or not realize that it may cause issues to have multiple versions installed. We could put a big warning, but the section is already optional and I think it doesn't really add much value. 

### Related issues (optional)

ENGDOCS-1897
Closes #18689

### DIFF
--- a/content/language/java/build-images.md
+++ b/content/language/java/build-images.md
@@ -31,44 +31,6 @@ $ git clone https://github.com/spring-projects/spring-petclinic.git
 $ cd spring-petclinic
 ```
 
-## Test the application without Docker (optional)
-
-In this step, you'll test the application locally without Docker, before you
-continue with building and running the application with Docker. This section
-requires you to have Java OpenJDK version 15 or later installed on your machine.
-[Download and install Java](https://jdk.java.net/)
-
-If you prefer to not install Java on your machine, you can skip this step, and
-continue straight to the next section, in which you'll build and run the
-application in Docker. Building or running the application in Docker doesn't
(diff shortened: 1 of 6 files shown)


content/language/java/build-images.md
```diff
@@ -31,44 +31,6 @@ $ git clone https://github.com/spring-projects/spring-petclinic.git
 $ cd spring-petclinic
 ```
 
-## Test the application without Docker (optional)
```
thaJeztah, line 34: We need to check the Java example; ISTR that there was a step that generated a file on the host, which is later copied into the Dockerfile. Don't know if this was updated later  on, but it may mean that some tools are still needed on the host (if that's the case, we should fix it though!)

craig-osterhout: I ran through the Java guide and it appears to work without running it outside of Docker first. I also double checked at least the first docker run in the Go guide.

The Python and Node guides used to require the installs to build the apps from scratch, but have been updated.

craig-osterhout: And just noticed the Go guide shows examples to run tests outside of Docker. I'll remove those.

thaJeztah: Thank you SO MUCH for checking! I noticed the PR, and recalled looking at these parts in the past, so would have blamed myself for "not commenting" (better safe than sorry 🥳)

thaJeztah (approved):
LGTM!!

PR #18883, opened 2023-12-08
update get docker cards

craig-osterhout:
### Proposed changes

Revert the size of the images on the cards back to icon size.

Received some internal feedback that the images look too large. It's subjective, but I do agree.
I think they were never specifically updated, but indirectly changed from wider styling changes.

### Related issues (optional)
ENGDOCS-1901

https://deploy-preview-18883--docsdocker.netlify.app/get-docker/


craig-osterhout:
Hi @aevesdocker, I know you have bigger plans for this page that you may have already started working on, so if this PR causes any conflicts, please close it.

PR #18884, opened 2023-12-08
doc: comment about async behavior of cli.ImagePull

zainq11:
### Proposed changes

Updated the example for `cli.ImagePull`. This is to educate about the behavior described in this [issue](https://github.com/moby/moby/issues/28646).

### Related issues (optional)
https://github.com/moby/moby/issues/28646


dvdksn (approved):
👍

PR #18888, opened 2023-12-10
Added a note to alert users installing on Ubuntu to not use the Debian installation

gkpln3:
Added note to not use debian installation on ubuntu

### Proposed changes

Added a note to alert users installing on Ubuntu to not use the Debian installation


dvdksn:
Sorry, I think this is self-explanatory.

PR #18906, opened 2023-12-12
`docker/build-push-action`: Update attestations.md

dsotirakis:
### Proposed changes

Small typo fix - and also removing a section as duplicated.


dvdksn (approved):
huh! thanks, that's odd that we had duplicate sections!

PR #18909, opened 2023-12-13
docker compose => docker-compose

gossie:
### Proposed changes

In the examples the cli command was written `docker compose`. The hyphen was missing.


gossie:
Ich just notices, that you do not need the hyphen  anymore.

dvdksn:
Yeah, the hyphen is for the legacy compose v1

PR #18917, opened 2023-12-14
Fix permission for Scout GHA

mikesir87:
The permission needed to allow Scout to write back to the PR was incorrect (docs here - https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs)

### Proposed changes

Fixed the permission name


dvdksn (approved):
Thanks

PR #18960, opened 2023-12-19
fix(docs-cache-mount): invalid go build with cache command

worapolburaphan:
### Proposed changes

Wrong command of example cache mount in Dockerfile

screenshot:
image


crazy-max (approved):
Thanks!

dvdksn (approved):
Thanks, LGTM

Issue #18904, opened 2023-12-12
Incorrect mention that hostname is the container name


thernstig:
### Is this a docs issue?

- [X] My issue is about the documentation content or website

### Type of issue

Information is incorrect

### Description

> Each container for a service joins the default network and is both reachable by other containers on that network, and discoverable by them at a hostname identical to the container name.

This might (or might not be true). But it is a garden path sentence as it leads you to believe that it is the container name you should code in your applications code, when in fact it change due to many reasons, whereas the service name would be the proper mention here as that does not change. The rest of the same page for networking explain this as well.

For example this:

```yaml
name: foo
services:
  postgres:
```

The service would be reachable at the service name `postgres` but the container name would be `foo-postgres-1`.

### Location

https://docs.docker.com/compose/networking/

### Suggestion

Change `hostname` to `service name` in the quoted text.

PR #19009, opened 2024-01-02
ENGDOCS-1926

aevesdocker:
Addresses confusion in https://github.com/docker/docs/issues/18904 and closes https://github.com/docker/docs/issues/18924#issuecomment-1855981646

### DIFF
--- a/content/compose/networking.md
+++ b/content/compose/networking.md
@@ -68,7 +68,7 @@ If any containers have connections open to the old container, they are closed. I
 
 ## Link containers
 
-Links allow you to define extra aliases by which a service is reachable from another service. They are not required to enable services to communicate. By default, any service can reach any other service at that service's name. In the following example, `db` is reachable from `web` at the names `db` and `database`:
+Links allow you to define extra aliases by which a service is reachable from another service. They are not required to enable services to communicate. By default, any service can reach any other service at that service's name. In the following example, `db` is reachable from `web` at the hostnames `db` and `database`:
 
 ```yaml
 services:


content/compose/networking.md
```diff
@@ -9,8 +9,7 @@ title: Networking in Compose
 By default Compose sets up a single
 [network](../engine/reference/commandline/network_create.md) for your app. Each
 container for a service joins the default network and is both reachable by
-other containers on that network, and discoverable by them at a hostname
-identical to the container name.
+other containers on that network, and discoverable by the container name.
```
dvdksn, line 12: #18904 is suggesting that the hostname is the service name, not the container name. E.g. the container name in the following example would be `svcs-web-1`, but the service would be discoverable on the bridge network via the `web` hostname.

```yaml
services:
  web:
    image: nginx:alpine
    ports:
      - 80:80
  sh:
    depends_on:
      - web
    build:
      context: .
      dockerfile_inline: |
        FROM alpine
        RUN apk add curl
        CMD curl http://web:80
```

content/compose/networking.md
```diff
@@ -41,7 +40,7 @@ When you run `docker compose up`, the following happens:
 3.  A container is created using `db`'s configuration. It joins the network
     `myapp_default` under the name `db`.
 
-Each container can now look up the hostname `web` or `db` and
+Each container can now look up the name `web` or `db` and
```
dvdksn, line 43: hostname is better here I think

content/compose/networking.md
```diff
@@ -69,7 +68,7 @@ If any containers have connections open to the old container, they are closed. I
 
 ## Link containers
 
-Links allow you to define extra aliases by which a service is reachable from another service. They are not required to enable services to communicate. By default, any service can reach any other service at that service's name. In the following example, `db` is reachable from `web` at the hostnames `db` and `database`:
+Links allow you to define extra aliases by which a service is reachable from another service. They are not required to enable services to communicate. By default, any service can reach any other service at that service's name. In the following example, `db` is reachable from `web` at the names `db` and `database`:
```
dvdksn, line 71: same as previous comment

dvdksn (approved):

PR #19036, opened 2024-01-04
Update fedora.md

kocic11:
The repo URL is not working. I updated it based on the solution from Stackowerflow: https://stackoverflow.com/questions/70358656/rhel8-fedora-yum-dns-causes-cannot-download-repodata-repomd-xml-for-docker-ce

### DIFF
--- a/content/engine/install/fedora.md
+++ b/content/engine/install/fedora.md
@@ -10,7 +10,7 @@ aliases:
 - /engine/installation/linux/fedora/
 - /engine/installation/linux/docker-ce/fedora/
 - /install/linux/docker-ce/fedora/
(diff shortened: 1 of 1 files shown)


dvdksn:
What version of Fedora are you using? The Fedora installation instruction supports fedora 38 and 39.

kocic11:
Good question. It may not be Fedora OS after all, although the release file says so:

dvdksn:
so the reason your installation fails is because you're using an unsupported distro. Please switch to a different distribution or try installing Docker using a different method (that may work, but that we will not cover in the documentation, since it's not a supported distribution)

thank you

PR #19056, opened 2024-01-08
Use consistent array style in Profiles example

allanlewis:
### Proposed changes

The config example for Profiles uses a mixture of different array styles. In order to avoid confusion and reduce the vertical size of the example config, this PR amends it to use minimal quoting and inline arrays.


dvdksn (approved):
Thanks!

PR #19067, opened 2024-01-09
Simplify removal of conflicting packages

farmergreg:
### Proposed changes
A for loop isn't needed because apt-get can handle multiple packages as parameters.


dvdksn:
This for loop is intentional, see: https://github.com/docker/docs/issues/17405

PR #19081, opened 2024-01-11
ENGDOCS-1931

aevesdocker:
- Removes EA banner
- Updates naming
- Remove steps to enable
- Adds links to content elsewhere in docs
- Makes content visible in TOC
- Updated known issues
- Updated content with pending GUI changes

### DIFF
--- a/content/desktop/synchronized-file-sharing.md
+++ b/content/desktop/synchronized-file-sharing.md
@@ -1,6 +1,6 @@
 ---
-title: Synchronized File Shares
-description: Get started with Synchronized File Shares on Docker Desktop.
+title: Synchronized file shares
+description: Get started with Synchronized file shares on Docker Desktop.
 keyword: mutagen, file sharing, docker desktop, bind mounts
 ---
 


content/storage/_index.md
```diff
@@ -106,6 +106,12 @@ You can't use Docker CLI commands to directly manage bind mounts.
 > non-Docker processes on the host system.
 { .important }
 
+> **Tip**
+>
+> Working with large repositories or monorepos, or with virtual file systems that are no longer scaling with your codebase?
+> Checkout [Synchronized File Shares](../desktop/synchronized-file-sharing.md) which provides fast and flexible host-to-VM file sharing by replacing bind mounts with synchronized filesystem caches.
```
dvdksn, line 112: question: We say "replacing bind mounts" here and in a few other places, but I don't think this is the correct way to phrase this. You'd still use a bind mount to get the files into the container, correct? So it doesn't replace bind mounts, but it makes bind mounts faster by adding a file synchronization layer between host and VM, and then bind mounts from VM to container I guess?

```suggestion
> Check out [Synchronized File Shares](../desktop/synchronized-file-sharing.md) which provides fast and flexible host-to-VM file sharing by replacing bind mounts with synchronized filesystem caches.
```

dvdksn:
Not part of these changes but the code span here is incorrect: https://github.com/docker/docs/blob/7798949d91d248930577944de03a8994117029b2/content/desktop/synchronized-file-sharing.md?plain=1#L67

"directories" shouldn't be wrapped in `<code>`

content/storage/_index.md
dvdksn: Also, should this be changed?

"Synchronized File Shares" -> "Synchronized file sharing"

dvdksn: > Nope this is correct, and also how it is phrased on the Mutagen docs site https://mutagen.io/documentation/docker-desktop-extension

How do you get those filesystem caches into the containers? You would use `--mount=type=bind` or `-v <src>:<dest>` -- a bind mount. Bind mounts aren't Docker Desktop implementations, they're a feature of the Docker Engine. It's my understanding that this feature changes how files move between the host and the container, but at the end of the day you're still using a bind mount when you create the container. That's why I don't think "replacing bind mounts" is a good choice of terminology.

cc @thaJeztah

content/desktop/synchronized-file-sharing.md
```diff
@@ -1,60 +1,41 @@
...
 >
-> Synchronized file sharing is an early-access feature. 
->
-> If you would like to be considered for Synchronized file sharing testing, you can
-> [sign up for the early access program](https://www.docker.com/docker-desktop-preview-program/).
-{ .restricted }
+> Synchronized File Shares is available for customers with a Docker subscription. 
 
-Synchronized file sharing is an alternative file sharing mechanism powered by [Mutagen](https://mutagen.io/). It provides fast and flexible host-to-VM file sharing by replacing bind mounts with synchronized filesystem caches. 
+Synchronized File Shares is an alternative file sharing mechanism powered by [Mutagen](https://mutagen.io/). It provides fast and flexible host-to-VM file sharing by replacing bind mounts with synchronized filesystem caches.
```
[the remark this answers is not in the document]
xenoscopic: I think I would opt to remove it; I think we want to avoid any confusion if people start veering into the Mutagen docs.  I'd just say "Synchronized File Shares is an alternative file sharing mechanism that provides fast and flexible host-to-VM ..."

content/desktop/synchronized-file-sharing.md
```diff
@@ -1,73 +1,55 @@
...
 
-Synchronized file sharing is an alternative file sharing mechanism powered by [Mutagen](https://mutagen.io/). It provides fast and flexible host-to-VM file sharing by replacing bind mounts with synchronized filesystem caches. 
+Synchronized File Shares is an alternative file sharing mechanism powered by [Mutagen](https://mutagen.io/). It provides fast and flexible host-to-VM file sharing by replacing bind mounts with synchronized filesystem caches. 
 
 ![Image of Synchronized file shares pane](images/synched-file-sharing.png)
  
 ## Who is it for?
 
-Synchronized file sharing is for developers who: 
+Synchronized File Shares is for developers who:
```
xenoscopic, line 17: Maybe we could say "is ideal for developers who:"?  I don't necessarily want to make people think this won't be helpful to them if they don't have the world's largest monorepo; there could be pathological cases that benefit from this functionality with smaller projects.

content/desktop/synchronized-file-sharing.md
```diff
@@ -1,73 +1,55 @@
...
 
-> **Early Access**
+> **Note**
 >
-> Synchronized file sharing is an early-access feature. 
->
-> If you would like to be considered for Synchronized file sharing testing, you can
-> [sign up for the early access program](https://www.docker.com/docker-desktop-preview-program/).
-{ .restricted }
+> Synchronized File Shares is available for customers with a Docker subscription.
```
xenoscopic, line 9: I know this was already addressed below, but the plurality mismatch between the singular feature and its plural name does make for a tough read.

I noticed that the Docker Extensions documentation takes an approach where they primarily refer to "extensions" (not the "Docker Extensions" feature name) to avoid a mismatch.  Maybe that's an approach we could take on a future revision (referring to "shares" or "file shares").

Though even the D.E. docs aren't always consistent (e.g. "Docker Extensions that are not in the Marketplace haven't gone through Docker's review process." and "Docker Extensions is switched on by default.").

content/desktop/synchronized-file-sharing.md
```diff
@@ -1,60 +1,41 @@
...
+Synchronized File Shares is for developers who: 
 - Have large repositories or monorepos with 100 000 files or more totaling hundreds of megabytes or even gigabytes.
 - Are using virtual filesystems, such as VirtioFS, gRPC FUSE, and osxfs, which are no longer scaling well with their codebases. 
 - Regularly encounter performance limitations.
 - Don't want to worry about file ownership or spend time resolving conflicting file-ownership information when modifying multiple containers.
 
-## How does Synchronized file sharing work?
+## How does Synchronized File Shares work?
 
 A Synchronized file share behaves just like a virtual file share, but takes advantage of Mutagen's high-performance, low-latency code synchronization engine to create a synchronized cache of the host files on an ext4 filesystem within the Docker Desktop VM. If you make filesystem changes on the host or in the VM’s containers, it propagates via bidirectional synchronization.
```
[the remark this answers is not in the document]
xenoscopic: +1 to this suggestion.

content/desktop/synchronized-file-sharing.md
```diff
@@ -1,73 +1,55 @@
...
 
-After creating a file share instance, any container using a bind mount that points to a location on the host filesystem matching the specified synchronized file share location, utilizes the Synchronized file sharing feature. Bind mounts that don't satisfy this condition are passed to the normal virtual filesystem bind-mounting mechanism, for example VirtioFS or gRPC-FUSE.
+After creating a file share instance, any container using a bind mount that points to a location on the host filesystem matching the specified synchronized file share location, utilizes the Synchronized File Shares feature. Bind mounts that don't satisfy this condition are passed to the normal virtual filesystem [bind-mounting mechanism](../storage/bind-mounts.md), for example VirtioFS or gRPC-FUSE.
```
xenoscopic, line 27: I'd clarify "matching the specified synchronized file share location (or a subdirectory thereof)"

content/desktop/synchronized-file-sharing.md
```diff
@@ -1,73 +1,55 @@
...
-
-When you re-open the Docker Dashboard, Synchronized file sharing is enabled and ready for you to create your first file share instance.
-
-## Create a file share instance 
-
-To create a synchronized file share:
-1. In **Settings**, navigate to the **File sharing** tab within the **Resources** section. 
-2. In the **Synchronized file shares** section, select the **plus** icon.
-3. Select a host folder to share. The synchronized file share should initialize and be usable.
+3. In the **Synchronized File Shares** section, select the **plus** icon.
```
xenoscopic, line 39: We can now replace "select the **plus** icon" with "click the 'Create share' button"

content/desktop/synchronized-file-sharing.md
```diff
@@ -83,19 +64,19 @@ You can use a `.syncignore` file at the root of each file share, to exclude loca
...
 - Large dependency directories, for example `node_modules` and `composer` directories (unless you rely on accessing them via a bind mount)
-- `.git directories` (again, unless you need them)
+- `.git` directories (again, unless you need them)
 
 In general, use your `.syncignore` file to exclude items that aren't critical to your workflow, especially those that would be slow to sync or use significant storage.
 
 ## Frequently asked questions
 
-### What is the difference between Synchronized file sharing and the Mutagen extension?
+### What is the difference between Synchronized File Shares and the Mutagen extension?
```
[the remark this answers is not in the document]
xenoscopic: I'd defer to @smnovick for the final decision, but we recently informed all paying users of the extension about the transition by email, so they should be up-to-date.

content/desktop/synchronized-file-sharing.md
```diff
@@ -83,19 +65,19 @@ You can use a `.syncignore` file at the root of each file share, to exclude loca
...
+Since Docker [acquired Mutagen](https://www.docker.com/blog/mutagen-acquisition/), Synchronized File Shares is essentially the direct integration of the Mutagen extension into Docker Desktop. However, Synchronized File Shares provides an improved user experience and the ability to use a [`.syncignore` file](#use-syncignore). You also no longer have to set the default ownership on files as this is now handled automatically.
 
 ## Known issues
 
-- Upon launching Docker Desktop, it can take between 5-10 seconds for Synchronized file sharing to fully initialize. During this window, file share instances display as **Connecting** and any new containers created during this window won't replace bind mounts with Synchronized file sharing.
+- Upon launching Docker Desktop, it can take between 5-10 seconds for Synchronized File Shares to fully initialize. During this window, file share instances display as **Connecting** and any new containers created during this window won't replace bind mounts with Synchronized File Shares.
```
xenoscopic, line 80: Given some recent improvements, I think we can remove this issue.  There's still a delay, but it should now be almost imperceptible.

content/desktop/synchronized-file-sharing.md
```diff
@@ -105,10 +87,10 @@ Since Docker [acquired Mutagen](https://www.docker.com/blog/mutagen-acquisition/
 
 - You cannot remove a file share instance during the initial synchronization. You have to wait for it to complete before **Delete** has any effect.
 
-- Synchronized file sharing proactively reports temporary issues, which can result in occasional **Conflict** and **Problem** indicators appearing in the GUI during synchronization. These can be ignored. However, if they persist, you can report the issue.
+- Synchronized File Shares proactively reports temporary issues, which can result in occasional **Conflict** and **Problem** indicators appearing in the GUI during synchronization. These can be ignored. However, if they persist, you can report the issue.
 
 - If you switch from WSL2 to Hyper-V on Windows, Docker Desktop needs to be fully restarted.
```
xenoscopic, line 93: I would add one more known issue, which is that we don't support POSIX-style Windows paths (which aren't really a standard anyway), so things like `/c/Users/jacob/some/path` won't use synchronized file shares (they'll fall back to normal file sharing). This is tracked by POS-2168, but it won't be fixed before 4.27.  The biggest impact is that users should avoid setting `COMPOSE_CONVERT_WINDOWS_PATHS`, but this isn't on by default and is a legacy setting that most won't use anyway.

content/storage/_index.md
```diff
@@ -106,6 +106,12 @@ You can't use Docker CLI commands to directly manage bind mounts.
 > non-Docker processes on the host system.
 { .important }
 
+> **Tip**
+>
+> Working with large repositories or monorepos, or with virtual file systems that are no longer scaling with your codebase?
+> Check out [Synchronized File Shares](../desktop/synchronized-file-sharing.md) which provides fast and flexible host-to-VM file sharing by replacing bind mounts with synchronized filesystem caches.
```
xenoscopic, line 112: Comma before "which" (including below)?

content/desktop/synchronized-file-sharing.md
```diff
@@ -105,10 +87,10 @@ Since Docker [acquired Mutagen](https://www.docker.com/blog/mutagen-acquisition/
...
 
 ## Feedback
 
-To give feedback, report bugs, or receive support, email `desktop-preview@docker.com`. There is also a dedicated Slack channel. To join, simply send an email to the provided address.
+To give feedback, report bugs, or receive support (TODO)
```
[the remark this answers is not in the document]
smnovick: Feedback and support are through normal channels. There's no specific avenue for this feature. We'll be closing out the public roadmap feature around it.

content/storage/_index.md
xenoscopic: I agree with @dvdksn, and perhaps the verbage on the Mutagen extension documentation isn't ideal.  At the end of the day it's a nuanced technical distinction that we should hide from users (without being incorrect in our description).

How about a description like "Synchronized File Shares is an alternative file sharing mechanism that provides fast and flexible host-to-VM file sharing that enhances bind mount performance using synchronized filesystem caches."

thaJeztah: Yes, I like that approach. Effectively, it's providing the existing functionality, but making it _better_ (same UX, better experience)

dvdksn (approved):
LGTM

PR #19083, opened 2024-01-11
Engdocs 1941

aevesdocker:
Part of https://www.notion.so/dockerinc/Compose-spec-deprecation-427b924a1aa0435e959d68ddd8716dbb

This simple PR moves the content into our repo. More to follow. 

cc @glours @ndeloof


dvdksn:
I don't think the data file for `docker init` is maintained in an upstream repo.

https://github.com/docker/docs/blob/766c54fe380317cefeb33b182112b3cc8a084721/data/init-cli/docker_init.yaml

cc @craig-osterhout

craig-osterhout:
> I don't think the data file for `docker init` is maintained in an upstream repo.
> 
> https://github.com/docker/docs/blob/766c54fe380317cefeb33b182112b3cc8a084721/data/init-cli/docker_init.yaml
> 
> cc @craig-osterhout

Correct, it's currently not maintained upstream.

dvdksn:
Could you rebase so that the vendor commit (fd6fb9c77b31013d9afa5e6ba41ec872bad8c40a) is in a separate commit, and the other commits are squashed?

dvdksn:
A squashed diff will show these files as having been moved from the _vendor dir to the content dir. Also I see that the content is different in these files. Maybe we should instead do this in three commits

- update the compose import first so it's on the latest commit
- then remove the vendoring
- add the files, ideally preserving history?

ndeloof (approved):

usha-mandya (dismissed):
Thanks for moving forward with this update @aevesdocker. Maintaining the upstream compose-spec will involve a wider discussion. Let's go ahead with the docs changes so we are not blocked on this.

content/compose/compose-file/05-services.md
```diff
@@ -1,166 +1,8 @@
...
-  - NET_ADMIN
-  - SYS_ADMIN
-```
-
-## cgroup
+---
+title: Services top-level elements
+description: Explore all the attributes the services top-level element can have.
+keywords: compose, compose specification, services, compose file reference
+---
```
glours, line 5: Why do we lost all the attributes from `attach` to `cap_drop` during the move of the file? Also the title for the `cgroup` argument is missing

glours (changes requested):
I think we have lost parts of the `services.md` file

glours (approved):
LGTM

usha-mandya (dismissed):

usha-mandya:
:shipit:

usha-mandya (approved):

PR #19108, opened 2024-01-16
Explain explicit property in default non-root user policy

felipecruz91:
### Proposed changes

This PR explains the `Explicit` property in the evaluation result of the Default non-root user policy.

### DIFF
--- a/content/scout/policy/_index.md
+++ b/content/scout/policy/_index.md
@@ -229,7 +229,10 @@ default `root` user. To comply with this policy, images must specify a non-root
 user in the image configuration. Images violate this policy if they don't
 specify a non-root default user for the runtime stage.
 
-In instances where an image does not adhere to this policy, the evaluation result will communicate through the `Explicit` property whether the Dockerfile explicitly specifies the use of the `USER root` instruction. This information serves to enhance developers' comprehension regarding the deliberate or inadvertent setting of the `root` user, aiding in the identification of intentional configurations.
+For non-compliant images, evaluation results show whether or not the `root`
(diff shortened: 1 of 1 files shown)


content/scout/policy/_index.md
```diff
@@ -229,6 +229,22 @@ default `root` user. To comply with this policy, images must specify a non-root
 user in the image configuration. Images violate this policy if they don't
 specify a non-root default user for the runtime stage.
 
+In instances where an image does not adhere to this policy, the evaluation result will communicate through the `Explicit` property whether the Dockerfile explicitly specifies the use of the `USER root` instruction. This information serves to enhance developers' comprehension regarding the deliberate or inadvertent setting of the `root` user, aiding in the identification of intentional configurations.
```
dvdksn, line 232:
```suggestion
For non-compliant images, evaluation results show whether or not the `root`
user was set explicitly for the image. This helps you distinguish between
policy violations caused by images where the `root` user is implicit, and
images where `root` is set on purpose.
```

content/scout/policy/_index.md
```diff
@@ -229,6 +229,22 @@ default `root` user. To comply with this policy, images must specify a non-root
...
+
+The following Dockerfile runs as `root` by default despite not being explicitly set:
+```Dockerfile
+FROM alpine
+RUN echo "Hi"
+```
+
+Whereas in the case below the `root` user is explictly set:
```
dvdksn, line 240:
```suggestion
Whereas in the following case, the `root` user is explicitly set:
```

PR #19111, opened 2024-01-16
Update tmpfs.md to correct the example tab about using "--tmpfs" flag

slashdreama:
To specify the real flag "--tmpfs" used for the example tab. So we avoid the confusion with volume. Hope that is helpful


dvdksn (approved):
Thank you 🎉

slashdreama:
@dvdksn It's a pleasure

PR #19129, opened 2024-01-18
Improve "Extend your Compose file"

allanlewis:
### Proposed changes

This PR makes various minor improvements to `extends.md`.

### Related issues (optional)

n/a


allanlewis:
> Hi @allanlewis, as per this page in the docs https://docs.docker.com/compose/compose-yaml-file/
> 
> 'The default path for a Compose file is compose.yaml (preferred) or compose.yml that is placed in the working directory. Compose also supports docker-compose.yaml and docker-compose.yml for backwards compatibility of earlier versions. If both files exist, Compose prefers the canonical compose.yaml.' So `compose.yaml` is correct.
> 
> If you could remove those changes from your PR I can accept the spacing fixes

I had those in a previous PR, #19128, but they're not in this one - I found precisely the doc you linked! I haven't intentionally renamed any of the uses of `compose.yaml` in this PR. (I closed the old PR because it was under a temporary branch name and I wanted to split the commit for readability, so I started again with a better-named branch.)

PR #19135, opened 2024-01-18
add docker debug cli reference

craig-osterhout:
### Proposed changes

Add CLI reference docs for Docker Debug.
https://deploy-preview-19135--docsdocker.netlify.app/engine/reference/commandline/debug/

### Related issues (optional)

DACC-26

### DIFF
--- a/data/debug-cli/docker_debug.yaml
+++ b/data/debug-cli/docker_debug.yaml
@@ -145,7 +145,7 @@ examples: |-
   The `--host` flag may be used with all other examples, including getting a
   shell into an image.
   
-  Example 4: Get a shell into a running container that has no shell
+  ### Example 4: Get a shell into a running container that has no shell
   
   ```console
   $ docker run --detach --rm --volume=/var/run/docker.sock:/var/run/docker.sock -p 8080:8080 --name my-container amir20/dozzle
--- a/data/debug-cli/docker_debug.yaml
+++ b/data/debug-cli/docker_debug.yaml
@@ -1,0 +1,162 @@
+command: docker debug
+short: Get a shell into any container or image.
+long: |-
+  Get an enhanced shell with additional tools into any container or image.  
+  
+  The shell contains the following builtin commands:
+  - `install [tool1] [tool2]`: Add Nix packages from: https://search.nixos.org/packages.
+  - `uninstall [tool1] [tool2]`: Uninstall NixOS package(s).
+  - `entrypoint`: Print/lint/run the entrypoint.
+  - `builtins`: Show builtin commands.
+
+  For images and stopped containers, all changes are discarded when leaving the
+  shell. At no point, do changes affect the actual image or container. When
+  accessing running or paused containers, all filesystem changes are directly
+  visible to the container. The `/nix` directory is never visible to the actual
+  image or container.
+usage: debug [OPTIONS] {CONTAINER|IMAGE}
+pname: docker
+plink: docker.yaml
+options:
+    - option: shell
+      value_type: shell
+      default_value: "auto"
+      description: "Select a shell. Supported: `bash`, `fish`, `zsh`, `auto`."
+      deprecated: false
+      hidden: false
+      experimental: false
+      experimentalcli: false
+      kubernetes: false
+      swarm: false
+    - option: command
+      shorthand: c
+      value_type: string
+      default_value: false
+      description: Evaluate the specified commands instead, passing additional positional arguments through $argv.
+      deprecated: false
+      hidden: false
+      experimental: false
+      experimentalcli: false
+      kubernetes: false
+      swarm: false
+    - option: host
+      value_type: string
+      default_value: false
+      description: "Daemon docker socket to connect to. E.g.: `ssh://root@example.org`, `unix:///some/path/docker.sock`"
+      deprecated: false
+      hidden: false
+      experimental: false
+      experimentalcli: false
+      kubernetes: false
+      swarm: false
+    - option: version
+      value_type: bool
+      default_value: "false"
+      description: Display version of the docker-debug plugin
+      deprecated: false
+      hidden: false
+      experimental: false
+      experimentalcli: false
+      kubernetes: false
+      swarm: false
+deprecated: false
+experimental: false
+experimentalcli: false
+kubernetes: false
+swarm: false
+examples: |-
+  ### Example 1: Get a shell into any image
+  This example pulls an nginx image and gets a shell into it without ever running a container.
+  
+  ```console
+  $ docker pull nginx
+  $ docker debug nginx
+  
+  Builtin commands:
+  - install [tool1] [tool2] ...    Add Nix packages from: https://search.nixos.org/packages
+  - uninstall [tool1] [tool2] ...  Uninstall NixOS package(s).
+  - entrypoint                     Print/lint/run the entrypoint.
+  - builtins                       Show builtin commands.
+  
+  Checks:
+  ✓ distro:            Debian GNU/Linux 12 (bookworm)
+  ✓ entrypoint linter: no errors (run 'entrypoint' for details)
+
+  Note: This is a sandbox shell. All changes will not affect the actual image.
+                                                                                                                     Version: 0.0.18 (BETA)
+  root@a8758716bb6a / [nginx:latest]
+  docker >
+  ```
+  
+  The console now has a shell in the read-only filesystem of the nginx image.
+  You can now run `entrypoint` to inspect the entrypoint or run `builtins` to
+  see all builtin commands.
+  
+  ### Example 2: Dumping a file from an image to stdout by using the -c mode
+  
+  Use the `-c` (`--command`) flag to directly evaluate a command instead of
+  running the shell in interactive mode.
+  
+  The following example dumps the default `index.html` file of nginx to stdout.
+
+  ```console
+  $ docker pull nginx
+  $ docker debug nginx -c "cat /usr/share/nginx/html/index.html"
+  <!DOCTYPE html>
+  <html>
+  <head>
+  <title>Welcome to nginx!</title>
+  <style>
+  html { color-scheme: light dark; }
+  body { width: 35em; margin: 0 auto;
+  font-family: Tahoma, Verdana, Arial, sans-serif; }
+  </style>
+  </head>
+  <body>
+  <h1>Welcome to nginx!</h1>
+  <p>If you see this page, the nginx web server is successfully installed and
+  working. Further configuration is required.</p>
+  
+  <p>For online documentation and support please refer to
+  <a href="http://nginx.org/">nginx.org</a>.<br/>
+  Commercial support is available at
+  <a href="http://nginx.com/">nginx.com</a>.</p>
+  
+  <p><em>Thank you for using nginx.</em></p>
+  </body>
+  </html>
+  ```
+  
+  ### Example 3: Get a shell into a stopped container via SSH
+  
+  Use the `--host` flag to connect to a different Docker Engine other than the
+  default.
+  
+  The following example connects to a remote Docker instance via SSH.
+  
+  ```console
+  $ docker debug --host ssh://root@example.org my-container
+  ```
+  
+  The `--host` flag may be used with all other examples, including getting a
+  shell into an image.
+  
+  Example 4: Get a shell into a running container that has no shell
+  
+  ```console
+  $ docker run --detach --rm --volume=/var/run/docker.sock:/var/run/docker.sock -p 8080:8080 --name my-container amir20/dozzle
+  $ docker debug my-container
+  ```
+  
+  Once the shell is attached to the container, you can add further tools with
+  the built-in `install` command. For example, run `install nmap`. Adding
+  further tools to your debugging shell does not modify the target container. 
+  
+  The built-in `forward` command lets you use remote tools. The following example starts a web file browser.
+  
+  ```console
+  $ forward --port 9000:9012 -- webfsd -F -p 9012
+  ```
+  
+  You can access the file browser on your host's browser at
+  http://localhost:9000.
--- a/data/debug-cli/docker_debug.yaml
+++ b/data/debug-cli/docker_debug.yaml
@@ -1,7 +1,12 @@
 command: docker debug
 short: Get a shell into any container or image.
 long: |-
-  Get an enhanced shell with additional tools into any container or image.  
+  > **Note**
+  >
+  > Docker Debug requires a [Pro, Team, or Business subcription](/subscription/details/).
+  > You must [sign in](/desktop/get-started/) to use this command.
+
+  Get an enhanced shell with additional tools into any container or image.
   
   The shell contains the following builtin commands:
   - `install [tool1] [tool2]`: Add Nix packages from: https://search.nixos.org/packages.
@@ -14,6 +19,7 @@ long: |-
   accessing running or paused containers, all filesystem changes are directly
   visible to the container. The `/nix` directory is never visible to the actual
   image or container.
+  
 usage: debug [OPTIONS] {CONTAINER|IMAGE}
 pname: docker
 plink: docker.yaml
--- a/data/debug-cli/docker_debug.yaml
+++ b/data/debug-cli/docker_debug.yaml
@@ -7,13 +7,19 @@ long: |-
   > You must [sign in](/desktop/get-started/) to use this command.
 
   Get an enhanced shell with additional tools into any container or image.
-  
+
   The shell contains the following builtin commands:
   - `install [tool1] [tool2]`: Add Nix packages from: https://search.nixos.org/packages.
-  - `uninstall [tool1] [tool2]`: Uninstall NixOS package(s).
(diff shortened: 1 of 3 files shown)


data/debug-cli/docker_debug.yaml
```diff
@@ -0,0 +1,162 @@
+command: docker debug
```
smnovick, line 1: We should probably add a section somewhere about the requirement to be logged in with a subscription to Docker Desktop.

smnovick (changes requested):

data/debug-cli/docker_debug.yaml
```diff
@@ -0,0 +1,166 @@
...
+      hidden: false
+      experimental: false
+      experimentalcli: false
+      kubernetes: false
+      swarm: false
+    - option: command
+      shorthand: c
+      value_type: string
+      default_value: false
+      description: Evaluate the specified commands instead, passing additional positional arguments through $argv.
```
dvdksn, line 41: suggestion:

```suggestion
      description: Evaluate the specified commands instead of starting an interactive session.
```

data/debug-cli/docker_debug.yaml
```diff
@@ -0,0 +1,166 @@
...
+  ✓ distro:            Debian GNU/Linux 12 (bookworm)
+  ✓ entrypoint linter: no errors (run 'entrypoint' for details)
+
+  Note: This is a sandbox shell. All changes will not affect the actual image.
+                                                                                                                     Version: 0.0.18 (BETA)
+  root@a8758716bb6a / [nginx:latest]
+  docker >
+  ```
+  
+  The console now has a shell in the read-only filesystem of the nginx image.
```
dvdksn, line 96: Not sure about the word "console" here, I'd associate that with a virtual console in Linux. Maybe "terminal"?

```suggestion
  Your terminal now has a shell in the read-only filesystem of the Nginx image.
```

data/debug-cli/docker_debug.yaml
```diff
@@ -0,0 +1,166 @@
...
+  ### Example 1: Get a shell into any image
+  This example pulls an nginx image and gets a shell into it without ever running a container.
+  
+  ```console
+  $ docker debug nginx
+  
+  Builtin commands:
+  - install [tool1] [tool2] ...    Add Nix packages from: https://search.nixos.org/packages
+  - uninstall [tool1] [tool2] ...  Uninstall NixOS package(s).
+  - entrypoint                     Print/lint/run the entrypoint.
```
dvdksn, line 83: What does print/lint/run refer to? Does the `entrypoint` command when invoked in an interactive debug shell only display the container's entrypoint, or does it do more?

From the output below it seems a lint of the entrypoint happens on starting the debug shell.

```suggestion
  - entrypoint                     Print/lint/run the entrypoint.
```

data/debug-cli/docker_debug.yaml
```diff
@@ -0,0 +1,166 @@
...
+    - option: version
+      value_type: bool
+      default_value: "false"
+      description: Display version of the docker-debug plugin
+      deprecated: false
+      hidden: false
+      experimental: false
+      experimentalcli: false
+      kubernetes: false
+      swarm: false
```
dvdksn, line 67: 😮 

I don't think we should use `--version` flags in the Docker CLI...

data/debug-cli/docker_debug.yaml
```diff
@@ -0,0 +1,166 @@
...
+    - option: host
+      value_type: string
+      default_value: false
+      description: "Daemon docker socket to connect to. E.g.: `ssh://root@example.org`, `unix:///some/path/docker.sock`"
+      deprecated: false
+      hidden: false
+      experimental: false
+      experimentalcli: false
+      kubernetes: false
+      swarm: false
```
dvdksn, line 57: 🤔 we already have a `--host` command on the base command. Why do we have another one here? Is it different, how do they co-exist?

data/debug-cli/docker_debug.yaml
```diff
@@ -0,0 +1,166 @@
...
+  further tools to your debugging shell does not modify the target container. 
+  
+  The built-in `forward` command lets you use remote tools. The following example starts a web file browser.
+  
+  ```console
+  docker > forward --port 9000:9012 -- webfsd -F -p 9012
+  ```
+  
+  You can access the file browser on your host's browser at
+  http://localhost:9000.
```
dvdksn, line 166: I am not sure I understand this example. Where is `webfsd` expected to be installed? On the host?

dvdksn (commented):
In addition to comments, here are a couple suggestions that likely would also affect the Go source. So maybe for a followup.

- Replace NixOS with Nix. I think we refer to the package manager in all cases, not the distro.
- We're mixing `builtin` and `built-in` here currently. Unfortunately it looks like the command is `builtins`. Should we use `built-in` consistently in the prose (and help messages) anyway?

data/debug-cli/docker_debug.yaml
```diff
@@ -0,0 +1,166 @@
...
+long: |-
+  > **Note**
+  >
+  > Docker Debug requires a [Pro, Team, or Business subcription](/subscription/details/).
+  > You must [sign in](/desktop/get-started/) to use this command.
+
+  Get an enhanced shell with additional tools into any container or image.
+  
+  The shell contains the following builtin commands:
+  - `install [tool1] [tool2]`: Add Nix packages from: https://search.nixos.org/packages.
```
joe0BAB, line 12: We should add a note here that installed tools never get added to the actual container/image, just to your debug toolbox. This is very different behaviour people are used to from eg `docker exec`. People ask for clarification on this a lot.

data/debug-cli/docker_debug.yaml
thaJeztah: Ah! Bookmarked this PR Yesterday, and was about to ask the same; if we have options on both levels, this will be confusing;

```bash
docker --host unix:///var/run/docker.sock debug --host <something else> ...
```

(maybe this was intended / used when running the plugin standalone, so not as plugin?)

data/debug-cli/docker_debug.yaml
thaJeztah: Yes, perhaps we should hide it for now. 

I think there's some discussion to be had on this "in general" (e.g.  we also have `docker buildx version`, `docker compose version`, which may also be confusing)

To some extent, I'd rather have a `--version` flag than a whole `version` command, but we should make sure that we align on this, and that we're consistent accros the board.

data/debug-cli/docker_debug.yaml
```diff
@@ -0,0 +1,166 @@
+command: docker debug
+short: Get a shell into any container or image.
+long: |-
+  > **Note**
+  >
+  > Docker Debug requires a [Pro, Team, or Business subcription](/subscription/details/).
+  > You must [sign in](/desktop/get-started/) to use this command.
+
+  Get an enhanced shell with additional tools into any container or image.
```
joe0BAB, line 9: We should explain a bit more. It's important to somehow introduce the concept of the toolbox here: 

The users can get a shell into any container/image. But they can also bring all the tools with them in that shell. Ie, the debug shell comes with a toolbox. However, this toolbox is always kept separated from the actual container/image. At no point the actually image/container will be affected/modified by that or see any of the tools. Eg, the user can run `install nmap` in the debug shell to add `nmap`. Than `nmap` can be used in the debug shell. But the actual container/image stays the same. If the user debugs something else, `nmap` can still be used because it was added to the toolbox. By default, the toolbox is pre-filled with common linux tools, eg `vim`, `nano`, `htop`..

cc @ipedrazas maybe you have something more catchy to explain this

data/debug-cli/docker_debug.yaml
```diff
@@ -0,0 +1,166 @@
...
+  ```
+  
+  The console now has a shell in the read-only filesystem of the nginx image.
+  You can now run `entrypoint` to inspect the entrypoint or run `builtins` to
+  see all builtin commands.
+  
+  ### Example 2: Dumping a file from an image to stdout by using the -c mode
+  
+  Use the `-c` (`--command`) flag to directly evaluate a command instead of
+  running the shell in interactive mode.
```
thaJeztah, line 103: Wondering if this should be working similar as  `--entrypoint` (like we have on `docker run` 🤔 ), which can disable the `/bin/sh` (or other entrypoint that the container / image may have; in `debug`'s case that entrypoint is the shell.

Or perhaps a cleaner approach would be to add a `none` option to the `--shell` flag;

```bash
docker debug --shell=none mycontainer /bin/some-command
```

data/debug-cli/docker_debug.yaml
```diff
@@ -0,0 +1,166 @@
...
+  The following example connects to a remote Docker instance via SSH.
+  
+  ```console
+  $ docker debug --host ssh://root@example.org my-container
+  ```
+  
+  The `--host` flag may be used with all other examples, including getting a
+  shell into an image.
+  
+  ### Example 4: Get a shell into a running container that has no shell
```
joe0BAB, line 148: This should be example number 1 or 2 as it's the most common use case.

data/debug-cli/docker_debug.yaml
```diff
@@ -0,0 +1,166 @@
...
+  
+  ```console
+  $ docker run --detach --rm --volume=/var/run/docker.sock:/var/run/docker.sock -p 8080:8080 --name my-container amir20/dozzle
```
joe0BAB, line 151: docs people, are there any other more popular images we can use here that doesn't include any shell, ie is a slim image? @mikesir87 any chance you know of some better image we could use here?

The problem with the image in the example is the `--volume=..` stuff. It's distracting and some people asked me if that is important for the example. But it's totally not. The intent of the example is to showcase that for slim images/containers `docker debug my-container` works while `docker exec -it my-container sh` does not.

data/debug-cli/docker_debug.yaml
joe0BAB: So maybe we should group the examples into 
* category A: the use cases where `docker debug` can be applied
* category B: extra things that are possible in the debug shell / special tools we added to your toolbox

Currently we have category A and two cases of category B are mixed within (the `entrypoint` and the `forward` example). Probably we should move them out and create a section `Things you can do inside the debug shell that you would not expect` (better title welcome). And then there we'd have
1. installing/removing tools
2. checking the entry point
3. dynamic port forwarding using the `forward` command
4. ...(there will come more in future)

data/debug-cli/docker_debug.yaml
thaJeztah: `hello-world` is very minimal, and an official image 😉

data/debug-cli/docker_debug.yaml
joe0BAB: docker debug doesn't have subcommands. I followed docker init on this which also has a --version flag

I'd agree to hiding it in the official docs. It doesn't really help the user and just makes the docs longer to read.

data/debug-cli/docker_debug.yaml
joe0BAB: yes, this duplication is intended (it's similar to the -n flag in kubectl)

From during the interviews I learned many people don't associate the top level --host flag with commands. They wouldn't know about it or not expect it there. This duplication intends to emphasise docker debug really is designed to work with --host flag. A lot of people asked for the remote case. That's also why it's important to extra have an example with the --host flag on this doc page.

data/debug-cli/docker_debug.yaml
joe0BAB: yeah, while it still was named `docker shell` this was way cleaner and clearer. It's just the equivalent of `bash/fish/zsh -c ARGS`. With `docker shell -c` that would have been clear directly, now we might need to explain..

data/debug-cli/docker_debug.yaml
thaJeztah: Hm.... not sure I'm happy with that; we'd now have `docker context`, `DOCKER_HOST`, `docker --host` _AND_ `docker debug --host`.

Note that this may have consequences with other logic in the CLI and elsewhere as well (where data should be sent, OTEL, etc), so we should be careful with this.

thaJeztah: I can also see this being confusing, e.g. use `docker ps` to list the container you want to debug, then `docker debug --host other-host` ? Wondering what the flow is in that case.

data/debug-cli/docker_debug.yaml
craig-osterhout: Added examples for these.

data/debug-cli/docker_debug.yaml
craig-osterhout: Removed this example for this release as it seems forward may be buggy and it isn't listed as part of the builtins for this reason.

Reorganized the examples by options, builtin commands, and other. I think any quirky behavior or use-cases we can add to the "other" section. To highlight use-cases, we can also consider adding it to the language-specific guides, or create something like a Debug with Docker guide for more visibility.

data/debug-cli/docker_debug.yaml
craig-osterhout: Added some more info below that paragraph and in the examples.

data/debug-cli/docker_debug.yaml
craig-osterhout: Added more examples to show the usage of all the options.

data/debug-cli/docker_debug.yaml
craig-osterhout: Created an example with hello-world. The new example has the user use debug and then `cat /etc/shells` to show that there are no shells in the image. I'm not sure if they'll understand that. Is there a better way to show this?

data/debug-cli/docker_debug.yaml
dvdksn: Is the --host flag actually duplicated or did you just include it here for the purpose of illustrating an example? If it it's only to show the debug _works_ with --host, then I'd remove it from this table and include a section on how to use it with remote hosts, with the syntax `docker --host x debug y` (and with `context` as well if that also works). Then it would be clear that it's using the existing mechanisms. Again, if that's the intention here.

thaJeztah: I'm also wondering how this would work, as "remote" won't be a Docker Desktop install, but a regular engine, which doesn't have the services running for debug;

Here's my list of contexts; I have one for a remote engine;

```bash
docker context ls
NAME              DESCRIPTION                               DOCKER ENDPOINT                                   ERROR
default           Current DOCKER_HOST based configuration   unix:///var/run/docker.sock
desktop-linux *   Docker Desktop                            unix:///Users/thajeztah/.docker/run/docker.sock
remote                                                      ssh://swarm-test-01
```

Doing a quick check; it looks like the remote engine has a `hello-world` image;

```bash
docker --context=remote image ls
REPOSITORY                TAG       IMAGE ID       CREATED        SIZE
busybox                   latest    a416a98b71e2   6 months ago   4.26MB
alpine                    latest    9ed4aefc74f6   9 months ago   7.05MB
hello-world               latest    feb5d9fea6a5   2 years ago    13.3kB
```

Trying to debug the image on the remote machine, it pulls the `docker/desktop-docker-debug-service:0.0.18` image on the remote machine, but then ... also pulls the current version of the  `hello-world` image (probably the `hello-world` image on the remove machine was outdated), and then fails?

```bash
docker debug --host ssh://swarm-test-01 hello-world
...
30a05b81ac91: Pull complete
0b540ad1848e: Pull complete
Digest: sha256:d3dc0f309dc6bc8aa69c59dcc5316c84f14a065d8a06a2bfa2bb6d44be33242e
Status: Downloaded newer image for docker/desktop-docker-debug-service:0.0.18
latest: Pulling from library/hello-world
c1ec31eb5944: Pull complete
Digest: sha256:4bd78111b6914a99dbc560e6a20eab57ff6655aea4a80c50b0c5491968cbc2e6
Status: Downloaded newer image for hello-world:latest
Error: error during connect: Get "http://labs.debug-tools.localhost/lookup": unable to upgrade to tcp, received 404
```

```bash
REPOSITORY                            TAG       IMAGE ID       CREATED        SIZE
docker/desktop-docker-debug-service   0.0.18    5a76776fd058   3 days ago     1.59GB
busybox                               latest    a416a98b71e2   6 months ago   4.26MB
hello-world                           latest    d2c94e258dcb   8 months ago   13.3kB
alpine                                latest    9ed4aefc74f6   9 months ago   7.05MB
hello-world                           <none>    feb5d9fea6a5   2 years ago    13.3kB
```

data/debug-cli/docker_debug.yaml
```diff
@@ -0,0 +1,374 @@
...
+      description: "Daemon docker socket to connect to. E.g.: `ssh://root@example.org`, `unix:///some/path/docker.sock`"
+      deprecated: false
+      hidden: false
+      experimental: false
+      experimentalcli: false
+      kubernetes: false
+      swarm: false
+deprecated: false
+experimental: false
+experimentalcli: false
```
dvdksn, line 66: Should we flag this command as experimental? We usually don't use "Beta" for CLI commands. See e.g. buildx debug

PR #19137, opened 2024-01-18
docker init: add java, remove beta

craig-osterhout:
### Proposed changes

Add Java template to docker init.
Remove Beta label from docker init.
Remove the version (4.18) info. I don't think we list the version on any other command, and versions older than 4.18 are no longer officially supported. 

### Related issues (optional)

DCKR-68


dvdksn (approved):
LGTM

PR #19138, opened 2024-01-19
Fix apt-get remove and ascii GPG key

audunmg:
Running apt-get remove in a loop is probably the slowest possible way to uninstall packages.
apt-get will not fail if any of the packages are already uninstalled, so this is unneccesary.

Changing the file extension for the GPG key to .asc lets apt-get know the GPG key is in ascii armor format, no need to dearmor.

### DIFF
--- a/content/engine/install/debian.md
+++ b/content/engine/install/debian.md
@@ -98,12 +98,11 @@ Docker from the repository.
    sudo apt-get update
    sudo apt-get install ca-certificates curl gnupg
    sudo install -m 0755 -d /etc/apt/keyrings
-   curl -fsSL {{% param "download-url-base" %}}/gpg | sudo gpg --dearmor -o /etc/apt/keyrings/docker.gpg
-   sudo chmod a+r /etc/apt/keyrings/docker.gpg
+   curl -fsSL {{% param "download-url-base" %}}/gpg -O /etc/apt/keyrings/docker.asc
 
    # Add the repository to Apt sources:
    echo \
-     "deb [arch=$(dpkg --print-architecture) signed-by=/etc/apt/keyrings/docker.gpg] {{% param "download-url-base" %}} \
+     "deb [arch=$(dpkg --print-architecture) signed-by=/etc/apt/keyrings/docker.asc] {{% param "download-url-base" %}} \
      $(. /etc/os-release && echo "$VERSION_CODENAME") stable" | \
      sudo tee /etc/apt/sources.list.d/docker.list > /dev/null
    sudo apt-get update
--- a/content/engine/install/raspberry-pi-os.md
+++ b/content/engine/install/raspberry-pi-os.md
@@ -100,12 +100,11 @@ Docker from the repository.
    sudo apt-get update
    sudo apt-get install ca-certificates curl gnupg
    sudo install -m 0755 -d /etc/apt/keyrings
-   curl -fsSL {{% param "download-url-base" %}}/gpg | sudo gpg --dearmor -o /etc/apt/keyrings/docker.gpg
-   sudo chmod a+r /etc/apt/keyrings/docker.gpg
+   curl -fsSL {{% param "download-url-base" %}}/gpg -O /etc/apt/keyrings/docker.asc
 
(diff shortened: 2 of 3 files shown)


tianon (approved):
I think APT has supported `.asc` files since ~1.4 (and the oldest Ubuntu LTS, 20.04/Focal has APT 2.0) :+1:

LGTM

tianon:
(even the oldest LTS-supported release of Debian has APT 1.8 :rocket:)

audunmg:
Thank you for verifying the `.asc` support in APT and for looking at it so quickly

dvdksn (changes requested):
Thanks for the PR @audunmg , could I ask you not to remove the `for` loop for apt? It's by design, see https://github.com/docker/docs/issues/17405

tianon:
Oh right, it'll still error for things it doesn't know at all (just not for things it knows but aren't installed) :facepalm:

We could "fix" that by getting clever with patterns, but that's also somewhat dangerous (as the patterns then might match unrelated things).  The loop is probably best. :sweat_smile:

dvdksn:
I patched the commit to revert the `for` loop change, and replicated the change to raspian/debian as well. ptal @tianon

tianon (approved):
:partying_face:

dvdksn (approved):

thaJeztah:
Thanks for looking, @tianon ! <3

content/engine/install/debian.md
```diff
@@ -98,12 +98,11 @@ Docker from the repository.
    sudo apt-get update
    sudo apt-get install ca-certificates curl gnupg
    sudo install -m 0755 -d /etc/apt/keyrings
-   curl -fsSL {{% param "download-url-base" %}}/gpg | sudo gpg --dearmor -o /etc/apt/keyrings/docker.gpg
-   sudo chmod a+r /etc/apt/keyrings/docker.gpg
+   curl -fsSL {{% param "download-url-base" %}}/gpg -O /etc/apt/keyrings/docker.asc
```
thaJeztah, line 101: Oh, erm, silly question; would need `sudo curl` now? (or `| sudo tee`)? Looks like `/etc/apt/keyrings` may not be accessible by everyone;

```bash
ls -l /etc/apt/keyrings
total 4
-rw-r--r-- 1 root root 2760 Jan 25 21:02 docker.gpg
```


The `sudo chmod a+r /etc/apt/keyrings/docker.gpg` was added for some cloud systems where permissions were not allowing traversing the directories, causing things to fail; https://github.com/docker/docs/pull/17070#discussion_r1161130453

thaJeztah: ^ @dvdksn @tianon

dvdksn: ah, yes. Would `sudo curl` do or do we need tee?

thaJeztah: `sudo curl` is probably the easy one, but thinking if cURL depends on user-directories (therefore `sudo` potentially looking for config in `root`'s home-dir and such. running cURL as root may have a slightly bigger attack surface (vs `tee`)  as well, but perhaps that's just looking for issues.

I'm sure @tianon has opinions as well if `curl | sudo tee` is preferred over `sudo curl`

tianon: Yeah, attack surface is all that came to mind for me (generally, running `curl` as root is probably safe, but not the best idea).  That being said, if we use `tee`, we should probably also send the output to `/dev/null`, so I'd personally think `sudo curl` is probably the simpler answer to keep the docs easier to understand (users that paranoid are hopefully not even using this line and are instead doing things like fetching the key by full fingerprint from elsewhere and/or doing deeper verification after download).

thaJeztah: Yup, `sudo curl` is definitely easier to grasp, so perhaps it's an ok trade-off (readability over "fully correct").

We should look at the directory permissions though (the sudo chmod a+r /etc/apt/keyrings/docker.gpg)

as I recall there were some real-life scenarios outside of the user's control where things broke without https://github.com/docker/docs/pull/17070

PR #19146, opened 2024-01-19
engine: deprecate API versions < 1.24

thaJeztah:
Docker Engine v25.0 deprecates old API versions, and disables them by default.

### DIFF
--- a/content/engine/api/_index.md
+++ b/content/engine/api/_index.md
@@ -151,14 +151,12 @@ You can specify the API version to use in any of the following ways:
 | 1.13.1         | [1.26](/engine/api/v1.26/) | [changes](/engine/api/version-history/#v126-api-changes) |
 | 1.13           | [1.25](/engine/api/v1.26/) | [changes](/engine/api/version-history/#v125-api-changes) |
 | 1.12           | [1.24](/engine/api/v1.24/) | [changes](/engine/api/version-history/#v124-api-changes) |
-| 1.11           | [1.23](/engine/api/v1.23/) | [changes](/engine/api/version-history/#v123-api-changes) |
-| 1.10           | [1.22](/engine/api/v1.22/) | [changes](/engine/api/version-history/#v122-api-changes) |
-| 1.9            | [1.21](/engine/api/v1.21/) | [changes](/engine/api/version-history/#v121-api-changes) |
-| 1.8            | [1.20](/engine/api/v1.20/) | [changes](/engine/api/version-history/#v120-api-changes) |
-| 1.7            | [1.19](/engine/api/v1.19/) | [changes](/engine/api/version-history/#v119-api-changes) |
-| 1.6            | [1.18](/engine/api/v1.18/) | [changes](/engine/api/version-history/#v118-api-changes) |
 
-### Archived API versions
+### Deprecated API versions
 
-You can find archived documentation for older versions of the API
-in the [Docker code repository on GitHub](https://github.com/moby/moby/tree/v1.9.1/docs/reference/api)
+API versions before v1.24 are [deprecated](/engine/deprecated/#deprecate-legacy-api-versions).
+You can find archived documentation for deprecated versions of the API in the
+code repository on GitHub:
+
+- [Documentation for API versions 1.23 and before](https://github.com/moby/moby/tree/v24.0.7/docs/api).
+- [Documentation for API versions 1.6 and before](https://github.com/moby/moby/tree/v1.9.1/docs/reference/api).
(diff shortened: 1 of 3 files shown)


data/redirects.yml
```diff
@@ -85,6 +85,14 @@
   - /go/access-tokens/
 "/desktop/mac/apple-silicon/":
   - /go/apple-silicon/
+"/engine/api/#deprecated-api-versions":
+  - /engine/api/v1.23/
+  - /engine/api/v1.22/
+  - /engine/api/v1.21/
+  - /engine/api/v1.20/
+  - /engine/api/v1.19/
+  - /engine/api/v1.18/
```
thaJeztah, line 94: Hm... these don't work in previews, is that correct? Or is it because there's still a page available at this URL?

https://deploy-preview-19146--docsdocker.netlify.app/engine/api/v1.23/

content/engine/api/_index.md
```diff
@@ -151,14 +151,12 @@ You can specify the API version to use in any of the following ways:
...
-| 1.7            | [1.19](/engine/api/v1.19/) | [changes](/engine/api/version-history/#v119-api-changes) |
-| 1.6            | [1.18](/engine/api/v1.18/) | [changes](/engine/api/version-history/#v118-api-changes) |
 
-### Archived API versions
+### Deprecated API versions
 
-You can find archived documentation for older versions of the API
-in the [Docker code repository on GitHub](https://github.com/moby/moby/tree/v1.9.1/docs/reference/api)
\ No newline at end of file
+API versions before v1.24 are [deprecated](/engine/deprecated/#deprecate-legacy-api-versions).
```
thaJeztah, line 157: This anchor is broken currently (which is expected), but also doesn't appear to be caught  in CI 🤔

data/redirects.yml
dvdksn: yeah preview deployments do not have cloudfront, which is what handles these pageless redirects. Hugo doesn't have a way to generate pages from data files, which we had a plugin for with Jekyll.

content/engine/api/_index.md
dvdksn: Not sure why, I'll look into it. Using relative paths would probably report it as broken though.

dvdksn:
We should also stop vendoring those old files, but I can handle that in a followup

data/redirects.yml
dvdksn: we should change this to point to only the 1.24 markdown file instead of all md files

https://github.com/docker/docs/blob/main/hugo.yaml#L241-L243

thaJeztah: Isn't this redirecting `/engine/api/<old version>` to `/engine/api/#deprecated-api-versions` ? Or did I misunderstand your comment 😅

dvdksn: No this is defining our imports from moby/moby and the target location to where we mount it in the virtual filesystem mount created by Hugo. So "*.md" files in docs/api are mounted to content/engine/api

thaJeztah: Is there anything I should change? (also feel free to push to my branch 😅)

dvdksn: I tried removing those pages.. but we'll be getting broken links then.

```
Start building sites … 
hugo v0.121.2-6d5b44305eaa9d0a157946492a6f319da38de154+extended darwin/arm64 BuildDate=2024-01-05T12:21:15Z VendorInfo=brew

ERROR [en] REF_NOT_FOUND: Ref "v1.24.md" from page "engine/api/version-history.md": page not found
ERROR [en] REF_NOT_FOUND: Ref "v1.23.md" from page "engine/api/version-history.md": page not found
ERROR [en] REF_NOT_FOUND: Ref "v1.22.md" from page "engine/api/version-history.md": page not found
ERROR [en] REF_NOT_FOUND: Ref "v1.21.md" from page "engine/api/version-history.md": page not found
ERROR [en] REF_NOT_FOUND: Ref "v1.20.md" from page "engine/api/version-history.md": page not found
ERROR [en] REF_NOT_FOUND: Ref "v1.19.md" from page "engine/api/version-history.md": page not found
ERROR [en] REF_NOT_FOUND: Ref "v1.18.md" from page "engine/api/version-history.md": page not found
Built in 5417 ms
Error: error building site: logged 7 error(s)
```

dvdksn (approved):
I think it is fine to merge this for now. I am not sure whether the redirects will work until we remove the pags (but first we need to remove the links in upstream if so) 

I am pretty sure the redirects will work though, because these are handled by cloudfront which means before the browser hits the .html page.

data/redirects.yml
thaJeztah: Ah, hm... yeah we need to take a look at that, and possibly remove them in upstream.

Do you know if redirects take priority over "page content" ? Because if that's the case, I guess people would just hit the redirect, and things should be fine.

thaJeztah:
Moved out of draft; feel free to merge when ready 👍

PR #19164, opened 2024-01-20
Update file-watch.md

### LINKED ISSUE #19163: Docs specify to use addUser command, but invalid otpions given, should be the useradd command
### Is this a docs issue?

### Type of issue

Information is incorrect

### Description

This code snippet  on this page seems to have a typo with the adduser command. If the code is added to a dockerfile, it will fail with the output:

Unknown option: ms
```
watch also requires that the container's USER can write to the target path so it can update files. A common pattern is for initial content to be copied into the container using the COPY instruction in a Dockerfile. To ensure such files are owned by the configured user, use the COPY --chown flag:

# Run as a non-privileged user
[RUN](https://docs.docker.com/engine/reference/builder/#run) adduser -ms /bin/bash -u 1001 app
[USER](https://docs.docker.com/engine/reference/builder/#user) app

# Copy source files into application directory
[COPY](https://docs.docker.com/engine/reference/builder/#copy) --chown=app:app . /app
```

### Location

https://docs.docker.com/compose/file-watch/

### Suggestion

Replace adduser with useradd

matthewbal:
### Proposed changes
Fixed typo of adduser -> useradd, which makes the example dockerfile command valid.

### Related issues (optional)
Closes #19163

### DIFF
--- a/content/compose/file-watch.md
+++ b/content/compose/file-watch.md
@@ -59,7 +59,7 @@ by the configured user, use the `COPY --chown` flag:
 
 ```dockerfile
 # Run as a non-privileged user
-RUN adduser -ms /bin/bash -u 1001 app
+RUN useradd -ms /bin/bash -u 1001 app
 USER app
 
 # Copy source files into application directory


content/compose/file-watch.md
```diff
@@ -59,7 +59,7 @@ by the configured user, use the `COPY --chown` flag:
 
 ```dockerfile
 # Run as a non-privileged user
-RUN adduser -ms /bin/bash -u 1001 app
+RUN useradd -ms /bin/bash -u 1001 app
```
thaJeztah, line 62: As this is a generic example, perhaps we should look which one is more portable. For example, `adduser` is available in minimal images, such as `busybox` (and `alpine`, which ships with a busybox userland).

If we want it to be more portable, we should also avoid using `/bin/bash` for the example (which is not available in many images), and `/bin/sh` instead.

(Perhaps change the example to be a complete dockerfile, including a `FROM` lins so that it can actually be run).


@dvdksn @glours wdyt?

dvdksn: isn't `useradd` the "canonical" command?

dvdksn: Agree that it would be nice to have full Dockerfile. `/bin/sh` is fine for me. Arguably `/usr/bin/env sh` is technically more correct :p

PR #19173, opened 2024-01-22
Add a note about moby/moby#46603 in v25.0 release notes

akerouanton:
### Proposed changes

- Add a note about moby/moby#46603 in v25.0 release notes

### Related issues (optional)

- moby/moby#47145


thaJeztah:
@akerouanton @dvdksn do we have a section describing `internal`? If so, perhaps we should add a description what it does (including this)?

akerouanton:
> If so, perhaps we should add a description what it does (including this)?

The way the default routing table is set up is an implementation detail and could be modified in the future, so I'm a bit on the fence. At least, we should mention that if we document it.

PR #19174, opened 2024-01-22
ENGDOCS-1953

aevesdocker:
addresses https://docker.slack.com/archives/C04300R4G5U/p1705695450299939
and fixes another formatting issue on the same page

### DIFF
--- a/content/desktop/install/windows-install.md
+++ b/content/desktop/install/windows-install.md
@@ -174,7 +174,7 @@ The `install` command accepts the following flags:
   - It must be used together with the `--allowed-org=<org name>` flag. 
   - For example:
 
-    ```console
+    ```text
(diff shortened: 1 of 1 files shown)


content/desktop/install/windows-install.md
```diff
@@ -174,8 +174,8 @@ The `install` command accepts the following flags:
   - It must be used together with the `--allowed-org=<org name>` flag. 
   - For example:
 
-    ```text
-    --allowed-org=<org name> --admin-settings="{'configurationFileVersion': 2, 'enhancedContainerIsolation': {'value': true, 'locked': false}}"
+    ```console
```
dvdksn, line 177: This isn't valid for console since it's just a flag. In this case we should either use text or the whole command

dvdksn (approved):
LGTM

PR #19215, opened 2024-01-25
fix(docker-desktop): wrong linux links

maxcleme:
### Proposed changes

Change few links

### Related issues (optional)

Links to download latest Docker Desktop from the dedicated Linux pages were wrong.


dvdksn (commented):
tyvm @maxcleme

PR #19236, opened 2024-01-26
ENGDOCS-1965

aevesdocker:
Brings in the following changes from https://github.com/compose-spec/compose-spec/pull/458

### DIFF
--- a/content/compose/compose-file/13-merge.md
+++ b/content/compose/compose-file/13-merge.md
@@ -185,7 +185,7 @@ services:
 ### Replace value
 
 While `!reset` can be used to remove a declaration from a Compose file using an override file, `!override` allows you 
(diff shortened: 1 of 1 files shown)


content/compose/compose-file/13-merge.md
```diff
@@ -180,4 +180,28 @@ services:
...
     environment: {}
     ports: []
-```
\ No newline at end of file
+```
+
+### Replace value
+
+While `!reset` can be used to remove a declaration from a Compose file using an override file, `!override` allows you 
+to fully replace an attribute, bypassing the standard merge rules. A typical example is to fully replace a resource definition, to rely on a distinct model but using same name.
```
stephaurelio, line 188: Potential typo: 
```suggestion
to fully replace an attribute, bypassing the standard merge rules. A typical example is to fully replace a resource definition, to rely on a distinct model but using the same name.
```

stephaurelio (approved):
LGTM

PR #19239, opened 2024-01-26
add nlp use-case guides

craig-osterhout:
### Proposed changes

Add more use-case guides based on the sample apps at https://github.com/harsh4870/Docker-NLP
https://deploy-preview-19239--docsdocker.netlify.app/guides/use-case/nlp/

### Related issues (optional)

ENGDOCS-1957

### DIFF
--- a/content/guides/use-case/nlp/language-translation.md
+++ b/content/guides/use-case/nlp/language-translation.md
@@ -1,0 +1,296 @@
+---
+title: Build a language translation app
+keywords: nlp, natural language processing, text summarization, python, language translation, googletrans
+description: Learn how to build and run a language translation application using Python, Googletrans, and Docker.
+---
+
+## Overview
+
+This guide walks you through building and running a language translation
+application. You'll build the application using Python with Googletrans, and
+then set up the environment and run the application using Docker.
+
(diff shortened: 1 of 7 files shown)


content/guides/use-case/nlp/language-translation.md
```diff
@@ -0,0 +1,296 @@
...
+## Explore the application code
+
+The source code for the application is in the
+`Docker-NLP/05_language_translation.py` file. Open `05_language_translation.py`
+in a text or code editor to explore its contents in the following steps.
+
+1. Import the required libraries.
+   ```python
+   from googletrans import Translator
+   ```
```
dvdksn, line 55: while not required by commonmark, I wonder if maybe we should consistently add blank lines before and after code blocks. I find that it improves readability and, aside from aesthetics, some parser would have problems with the lack of blank lines.

The same is true for lists. (A list can be tight, but blank lines before and after is usually a good idea, and makes it easier to catch unwanted [lazy continuation lines](https://spec.commonmark.org/0.31.2/#lazy-continuation-line)).

dvdksn (commented):
only a formatting comment/question from me

dvdksn (approved):
LGTM

PR #19244, opened 2024-01-29
Use a maintained version of Python in `compose` getting started guide

Viicos:
3.10 seems reasonable for now

### DIFF
--- a/content/compose/gettingstarted.md
+++ b/content/compose/gettingstarted.md
@@ -89,7 +89,7 @@ In your project directory, create a file named `Dockerfile` and paste the follow
 
 ```dockerfile
 # syntax=docker/dockerfile:1
-FROM python:3.7-alpine
+FROM python:3.10-alpine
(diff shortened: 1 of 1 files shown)


content/compose/gettingstarted.md
```diff
@@ -351,4 +351,4 @@ $ docker compose down --volumes
 - Try the [Sample apps with Compose](https://github.com/docker/awesome-compose)
 - [Explore the full list of Compose commands](reference/index.md)
 - [Explore the Compose file reference](compose-file/index.md)
-- To learn more about volumes and bind mounts, see [Manage data in Docker](../storage/index.md)
\ No newline at end of file
+- To learn more about volumes and bind mounts, see [Manage data in Docker](../storage/index.md)
```
Viicos, line 354: Not sure why this is counted as a diff, I haven't touched it and nothing seems to have changed

PR #19255, opened 2024-01-30
fix: GitHub Action rename token to github-token

mcapell:
### Proposed changes

This change fixes the example about setting up the Scout GitHub Action.


dvdksn (approved):
thx!

PR #19295, opened 2024-02-02
Add runc, buildkit and moby CVEs to security announcements

gabriellavengeo:
### Proposed changes

Add runc, buildkit and moby CVEs to security announcements


stephaurelio (approved):
LGTM

PR #19319, opened 2024-02-06
Engdocs 1969

aevesdocker:
Addresses: https://github.com/docker/docs/issues/18809 and https://github.com/docker/docs/issues/18700

Adds version notes for major Docker Compose functionality additions going back to June 2023. Establishes a pattern of how we can do this going forward with new releases

### DIFF
--- a/content/compose/compose-file/05-services.md
+++ b/content/compose/compose-file/05-services.md
@@ -656,7 +656,7 @@ When not explicitly set, `tcp` protocol is used.
 expose:
   - "3000"
   - "8000"
-  - "8080-8085/tcp
+  - "8080-8085/tcp"
 ```
 
 > **Note**
@@ -1151,7 +1151,7 @@ There is a performance penalty for applications that swap memory to disk often.
 
 - `none`: Turns off all container networking.
 - `host`: Gives the container raw access to the host's network interface.
-- `service:{name}`: Gives the containers access to the specified service only. For more information, see [Container networks](../../network/_index.md#container-networks).
+- `service:{name}`: Gives the containers access to the specified service only.
 
 ```yml
     network_mode: "host"


glours (approved):
LGTM 👍

content/compose/compose-file/05-services.md
```diff
@@ -222,8 +222,7 @@ Services can only access configs when explicitly granted by the `configs` attrib
 Compose reports an error if `config` doesn't exist on the platform or isn't defined in the
 [`configs` top-level element](08-configs.md) in the Compose file.
 
-There are two syntaxes defined for configs. To remain compliant to this specification, an implementation
-must support both syntaxes. Implementations must allow use of both short and long syntaxes within the same document.
+There are two syntaxes defined for configs.
```
dvdksn, line 225: ?

```suggestion
There are two syntaxes defined for configs: a short syntax and a long syntax.
```

content/compose/compose-file/05-services.md
```diff
@@ -1154,7 +1147,7 @@ There is a performance penalty for applications that swap memory to disk often.
 
 ## network_mode
 
-`network_mode` sets a service container's network mode. Available values are platform specific, but Compose defines specific values which must be implemented as described if supported:
+`network_mode` sets a service container's network mode.
```
dvdksn, line 1150: Is `service:{name}` the same as a container network described here? https://docs.docker.com/network/#container-networks
Where the container gets the networking stack of another container.
Would it be useful to expand and/or add a link?

content/compose/compose-file/05-services.md
```diff
@@ -1500,7 +1491,6 @@ section of the Docker run reference page.
 
 `runtime` specifies which runtime to use for the service’s containers.
 
-The value of `runtime` is specific to the implementation.
 For example, `runtime` can be the name of [an implementation of OCI Runtime Spec](https://github.com/opencontainers/runtime-spec/blob/master/implementations.md), such as "runc".
```
dvdksn, line 1494: We could call out that `runc` is the default, and to use a different runtime you need to install the shims and/or configure the daemon as described here: https://docs.docker.com/engine/alternative-runtimes/

content/compose/compose-file/05-services.md
```diff
@@ -1118,13 +1105,53 @@ logging:
...
+
 ## network_mode
 
-`network_mode` sets a service container's network mode. Available values are platform specific, but Compose defines specific values which must be implemented as described if supported:
+`network_mode` sets a service container's network mode. 
 
 - `none`: Turns off all container networking.
 - `host`: Gives the container raw access to the host's network interface.
-- `service:{name}`: Gives the containers access to the specified service only.
+- `service:{name}`: Gives the containers access to the specified service only. For more information, see [Container networks](../../network/_index.md#container-networks).
```
dvdksn, line 1154: does this lgty @glours ?

dvdksn (dismissed):
LGTM, just would like confirmation that the container networks suggestion I proposed is actually the correct interpretation

dvdksn (approved):
LGTM

PR #19334, opened 2024-02-07
Clarify how JIT works with SAML attributes and impacts SCIM

stephaurelio:
### Proposed changes

This PR adds callouts warning that JIT values overwrite SCIM values on user log in, and a workaround to prevent a discrepency is to make sure these values match. Also adds to the callout to emphasize which attributes are required.

### Related issues (optional)

Closes JIRA https://docker.atlassian.net/browse/ENGDOCS-1978

### DIFF
--- a/content/security/for-admins/single-sign-on/_index.md
+++ b/content/security/for-admins/single-sign-on/_index.md
@@ -36,7 +36,12 @@ When a user signs in using SSO, Docker obtains the following attributes from the
 - **Full name** - name of the user
 - **Groups (optional)** - list of groups to which the user belongs
 
-If you use SAML for your SSO connection, Docker obtains these attributes from the SAML assertion message. Your IdP may use different naming for SAML attributes than those in the previous list. The following table lists the possible SAML attributes that can be present in order for your SSO connection to work. 
+If you use SAML for your SSO connection, Docker obtains these attributes from the SAML assertion message. Your IdP may use different naming for SAML attributes than those in the previous list. The following table lists the possible SAML attributes that can be present in order for your SSO connection to work.
+
+> **Important**
+>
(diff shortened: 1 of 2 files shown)


content/security/for-admins/single-sign-on/_index.md
```diff
@@ -36,7 +36,12 @@ When a user signs in using SSO, Docker obtains the following attributes from the
...
+If you use SAML for your SSO connection, Docker obtains these attributes from the SAML assertion message. Your IdP may use different naming for SAML attributes than those in the previous list. The following table lists the possible SAML attributes that can be present in order for your SSO connection to work.
+
+> **Important**
+>
+>SSO uses Just-in-Time (JIT) provisioning by default. If you [enable SCIM](../scim.md#set-up-scim), JIT values still overwrite the attribute values set by SCIM provisioning whenever users log in. To avoid conflicts, make sure your JIT values match your SCIM values. For example, to make sure that the full name of a user displays in your organization, you would create a `name` attribute in your IdP and ensure the value includes their first name and last name. The exact syntax for these values (for example, `user.firstName + " " + user.lastName`) varies depending on your IdP.
```
yarabarla, line 43: > you would create a `name` attribute in your IdP

I would rephrase this to something like: "you would set a `name` attribute in your SAML attributes" to clarify that we expect it as part of the assertion

yarabarla, line 43: > exact syntax for these values

I wouldn't say syntax for these values since in some IDPs you may not need to construct the Full Name value. There could be a direct attribute for that. 

Maybe something like "The exact method for setting this value, (for example constructing it with `user.firstName + " " + user.lastName`) varies depending on your IdP.

craig-osterhout (approved):

PR #19348, opened 2024-02-08
Fix list like environments attribute

lnksz:
### Proposed changes

`envrionments` attribute is a mapping not a list

### Related issues (optional)

See caused confusion https://github.com/docker/compose/issues/1171#issuecomment-705534861

### DIFF
--- a/content/compose/environment-variables/set-environment-variables.md
+++ b/content/compose/environment-variables/set-environment-variables.md
@@ -65,7 +65,7 @@ services:
       webapp:
         image: my-webapp-image
         environment:
-          - DEBUG=${DEBUG}
+          DEBUG=${DEBUG}
   ```
   Docker Compose replaces `${DEBUG}` with the value from the `.env` file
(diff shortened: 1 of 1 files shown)


thaJeztah:
It should support both?; https://docs.docker.com/compose/compose-file/05-services/#environment

Screenshot 2024-02-08 at 13 11 01

content/compose/environment-variables/set-environment-variables.md
```diff
@@ -65,7 +65,7 @@ services:
       webapp:
         image: my-webapp-image
         environment:
-          - DEBUG=${DEBUG}
+          DEBUG=${DEBUG}
```
dvdksn, line 68: This doesn't look valid. There's no colon delimiter for the map key/value, so this would be interpreted as a string.

lnksz:
> It should support both?; https://docs.docker.com/compose/compose-file/05-services/#environment
> Screenshot 2024-02-08 at 13 11 01

My bad, thanks I blindly believed 

> I've this config that works, you have defined your environment variables as array is the issue

from https://github.com/docker/compose/issues/1171#issuecomment-705534861

lnksz:
Added my observation/question rather [here](https://github.com/docker/compose/issues/1171#issuecomment-1934171005)

PR #19353, opened 2024-02-08
minor UI fix for Docker Desktop release notes

chaizeg:
### Proposed changes

Noticed this in the Docker Desktop [release notes](https://docs.docker.com/desktop/release-notes/#:~:text=2024%2D02%2D08-,Download%20Docker,-Desktop) : 
image
so this PR : 
- removes the extra pipe before Linux installers
- adds the missing pipe before Mac with intel chip installer


dvdksn (approved):
LGTM

PR #19368, opened 2024-02-09
Fix the GenAI usecase guide

p1-0tr:
Attempt to make it clearer where users can expect to be able to use GPU acceleration. Currently that is (1) docker-ce on Linux; (2) docker desktop on windows wsl2.


craig-osterhout (approved):
LGTM. Thank you!

PR #19373, opened 2024-02-10
build: fix stage name references

eriksjolund:
### Proposed changes

Today I read this web page
https://docs.docker.com/build/guide/build-args/
and noticed that a Dockerfile contained
```
--from=build 
```
but the Dockerfile had no  stage with that name.
I assume this is a documentation typo.


dvdksn (approved):
thanks @eriksjolund !

PR #19374, opened 2024-02-10
Update tmpfs.md for unlimited warning

horner:
Adding a little warning about using unlimited on temps

### DIFF
--- a/content/storage/tmpfs.md
+++ b/content/storage/tmpfs.md
@@ -114,8 +114,15 @@ as the `--tmpfs` flag does not support them.
 
 | Option       | Description                                                                                           |
 |:-------------|:------------------------------------------------------------------------------------------------------|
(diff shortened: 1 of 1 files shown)


dvdksn:
Thanks, I pushed an update to change from a subheading to a footnote (I think it's more appropriate for this callout.)

dvdksn:
ptal @vvoland

horner:
Thank you!

vvoland (changes requested):
Thanks for adding the warning! However, this part of the docs seems to be wrong (or outdated).
The `tmpfs-size` options does not default to "unlimited". In fact there's not default assigned by Docker, meaning that no particular size is specified to the `mount` syscall, so the exact value here is up to the kernel.

According to `tmpfs` [manpages](https://www.man7.org/linux/man-pages/man5/tmpfs.5.html), the default is `50%` (of the host physical RAM).

PR #19375, opened 2024-02-10
refactor(docs): fix typos

its-dibo:


its-dibo:
closed because `docker ps` is an alias to `docker container ls`

PR #19393, opened 2024-02-14
Update include.md

kevinmarrec:
https://github.com/docker/compose/issues/10932#issuecomment-1696396507

### Proposed changes

Docker Compose `v2.20.3` is required to have proper `include` feature support.

I had a co-worker using `v2.20.2` which was shipped on Linux `apt` repository, and couldn't understand why it was not working, as the documentation didn't mention the patch version that included a fix regarding `include` feature.

### Related issues
https://github.com/docker/compose/issues/10932#issuecomment-1696396507

### DIFF
--- a/content/compose/multiple-compose-files/include.md
+++ b/content/compose/multiple-compose-files/include.md
@@ -6,7 +6,7 @@ title: Include
 
 > **Note**
 >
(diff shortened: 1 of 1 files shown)


EtiennePASSOT:
I was a co-worker, thank's @kevinmarrec 🥹❤️❤️❤️

PR #19395, opened 2024-02-14
Engdocs 1985

aevesdocker:
Docs freshness for DD topics.

### DIFF
--- a/content/desktop/wsl/use-wsl.md
+++ b/content/desktop/wsl/use-wsl.md
@@ -23,5 +23,5 @@ The following section describes how to start developing your applications using 
     This opens a new VS Code window connected remotely to your default Linux distro which you can check in the bottom corner of the screen.
 
 
-Alternatively, you can open your default Linux distribution from the Start menu, navigate to your project directory, and then run code .
+Alternatively, you can open your default Linux distribution from the **Start** menu, navigate to your project directory, and then run `code .`
 


content/desktop/networking.md
```diff
@@ -101,10 +108,10 @@ ping the Windows containers.
 
 ### Per-container IP addressing is not possible
 
-The docker bridge network is not reachable from the host.
+The Docker-bridge network is not reachable from the host.
 However if you are a Windows user, it works with Windows containers.
```
dvdksn, line 112: suggestion: These two sentences read odd together like this.

- The first is about the network not being accessible from the host
- The second says it "works" (what is it that works? is it reachable from the host if you use hyper-v?)

```suggestion
The Docker `bridge` network is not reachable from the host.
However if you are a Windows user, it works with Windows containers.
```

content/desktop/wsl/use-wsl.md
```diff
@@ -21,5 +21,5 @@ The following section describes how to start developing your applications using
 
     This opens a new VS Code window connected remotely to your default Linux distro which you can check in the bottom corner of the screen.
 
-    Alternatively, you can type the name of your default Linux distro in your Start menu, open it, and then run `code` .
+    Alternatively, you can type the name of your default Linux distro in your Start menu, open it, and then run `code .`
```
dvdksn, line 24: suggestion: `code .` was probably intentional here because it would open VS Code in the current working directory. But that could be clarified/worded better - and we don't really mention the current working directory implications here. I assume opening WSL either with `wsl` or from the start menu would put you in the WSL home directory? But that's not a very useful workspace to launch your editor in. You'd probably want to `cd` to the project directory first (inside wsl)

dvdksn (approved):
LGTM

PR #19416, opened 2024-02-15
ENGDOCS-1989

aevesdocker:
Improves `extends` use cases as discussed in https://docker.slack.com/archives/C03J0AFQ3A9/p1708000160002839

Preview here https://deploy-preview-19416--docsdocker.netlify.app/compose/multiple-compose-files/extends/

### DIFF
--- a/content/compose/multiple-compose-files/extends.md
+++ b/content/compose/multiple-compose-files/extends.md
@@ -60,7 +60,7 @@ You get exactly the same result as if you wrote
 `docker-compose.yml` with the same `build`, `ports` and `volumes` configuration
 values defined directly under `web`.
 
-To include the service `webapp` in the final project when extending services from another file, you need to explicitly include both services in your current Compose file. For example:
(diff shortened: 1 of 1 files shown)


glours (approved):
LGTM

content/compose/multiple-compose-files/extends.md
```diff
@@ -52,30 +56,44 @@ services:
...
 ```
-
-In this case, you get exactly the same result as if you wrote
+You get exactly the same result as if you wrote
 `docker-compose.yml` with the same `build`, `ports` and `volumes` configuration
 values defined directly under `web`.
 
-You can go further and define, or re-define, configuration locally in
-`compose.yaml`:
+To include the service `webapp` in the final project when extending services from another file, you need to explicitly include both services in your current Compose file. For example:
```
ndeloof, line 63: Not sure I get it
Do you want the end result to have two services, with the same config (which doesn't make sense to me)?

ndeloof (approved):

craig-osterhout (approved):
LGTM

PR #19420, opened 2024-02-15
docs: update wrong numeral adjectives in multi-platform.md

zevisert:
An example build command used three platforms, then quantified the number of concurrent processes as four.

The image just after the command that created those three builds shows a diagram with two builds. Instead of using another different quantifier for the same thought that should exist in the reader's mind, I changed the sentence after to just say "multiple".


dvdksn (approved):
Thank you!

PR #19444, opened 2024-02-19
Reorder installation names

doringeman:


content/engine/install/rhel.md
```diff
@@ -92,13 +92,8 @@ Docker from the repository.
...
 #### Set up the repository
 
-
-Install the `yum-utils` package (which provides the `yum-config-manager`
-utility) and set up the repository.
-
 ```console
-$ sudo yum install -y yum-utils
-$ sudo yum-config-manager --add-repo {{% param "download-url-base" %}}/docker-ce.repo
+$ sudo dnf config-manager --add-repo {{% param "download-url-base" %}}/docker-ce.repo
```
dvdksn, line 96: This page states that installation is supported on RHEL 7,8, and 9. I think RHEL 7 uses `yum`, and `yum` is aliased to `dnf` in RHEL 8 and 9 for backward compatibility. So would this change break RHEL 7 and is that a concern? @neersighted 

> Although RHEL 8 and RHEL 9 are based on DNF, they are compatible with YUM used in RHEL 7.

https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/9/html-single/managing_software_with_the_dnf_tool/index

> Red Hat Enterprise Linux 7 will reach end of maintenance on June 30, 2024, but it is not reaching End of Life.

https://www.redhat.com/en/technologies/linux-platforms/enterprise-linux/rhel-7-end-of-maintenance

thaJeztah: Yeah, looking forward to RHEL and CentOS 7 reaching EOL, but still a couple of months out. Changing to `dnf` very likely would break users installing.

Also, I don't think we have packages already for RHEL on architectures other than s390x. I know @neersighted was looking into that, but they're not there yet (?)

doringeman: We've been given some packages that we'd like to upload manually to those repos before the release.
See [this thread](https://docker.slack.com/archives/C01HU919A2C/p1708081894991699?thread_ts=1704988913.564649&cid=C01HU919A2C).

dvdksn (changes requested):
for RHEL on Docker Desktop, let's inline the steps to set up the repository for x86/arm in the DD installation guide rather than link out to the s390x guide for docker-ce, to avoid any confusion and/or compatibility concerns

content/desktop/install/rhel.md
```diff
@@ -0,0 +1,136 @@
...
+
+- Meet the [system requirements](linux-install.md#general-system-requirements).
+- Have a 64-bit version of either RHEL 8 or RHEL 9.
+
+If you don't have `pass` installed, or it can't be installed, you must enable
+[CodeReady Linux Builder (CRB) repository](https://access.redhat.com/articles/4348511)
+and
+[Extra Packages for Enterprise Linux (EPEL)](https://docs.fedoraproject.org/en-US/epel/).
+
+   {{< tabs >}}
```
dvdksn, line 33: tip: if you add the sam group to these tab groups will switch synchronously

```suggestion
   {{< tabs group="os_version" >}}
```

doringeman: Nice!

content/desktop/install/rhel.md
```diff
@@ -0,0 +1,136 @@
...
+```console
+$ sudo subscription-manager repos --enable codeready-builder-for-rhel-8-$(arch)-rpms
+$ sudo dnf install https://dl.fedoraproject.org/pub/epel/epel-release-latest-8.noarch.rpm
+$ sudo dnf install pass
+```
+
+   {{< /tab >}}
+   {{< /tabs >}}
+
+Additionally, for a Gnome Desktop environment you must install AppIndicator and KStatusNotifierItem [Gnome extensions](https://extensions.gnome.org/extension/615/appindicator-support/). You need EPEL enabled, as described above.
```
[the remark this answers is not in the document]
doringeman: Also for https://github.com/docker/docs/pull/19444/files#diff-8de446c72955d96bf1de732898008fbac2d2294fb1129a24c9cbeff7cf32b44bR74 and `content/desktop/install/fedora.md`?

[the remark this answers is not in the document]
dvdksn: yeah, Gnome -> GNOME

content/engine/install/rhel.md
neersighted: RHEL 7 is not supported on x86_64 and aarch64, but RHEL 8 and 9 are. These are both distributions that include DNF.

dvdksn (approved):
LGTM but should this be called out as an EAP or is it generally available for DD?

PR #19454, opened 2024-02-20
update docker init reference

craig-osterhout:
`docker init` will overwrite and use `docker-compose.yaml` as the Compose file name if it already exists. It will prompt the user, but the behavior can be confusing as the examples all use `compose.yaml`.

ENGDOCS-1998
Ran into the issue in #19390


dvdksn (approved):
LGTM

PR #19519, opened 2024-02-27
update use-case guides toc

craig-osterhout:
- Reordered the Jupyter use-case guide in the navigation to be closer to the other AI/ML guides.
- Added the "new" label to the use-case guides to try to drum up more awareness. 
  
image

ENGDOCS-2005


dvdksn (approved):
LGTM

PR #19583, opened 2024-03-07
desktop: add debug gui

craig-osterhout:
## Description

Added "Debug Mode" setting to integrated terminal.
Added "Enable Docker Debug by default" option in settings.
Added go redirect for "Learn more" links in Docker Desktop.

## Related issues or tickets

ENGDOCS-2009

## Reviews

- [ ] Technical review
- [ ] Editorial review
- [ ] Product review


content/desktop/use-desktop/container.md
```diff
@@ -29,46 +29,92 @@ When you [inspect a container](#inspect-a-container), the **Stats** tab displays
...
+  you're using debug mode, it defaults to `root`.
+
+#### Open the integrated terminal
 
 To open the integrated terminal, either:
-- Hover over your running container and select the **Show container actions** menu. From the drop-down menu, select **Open in terminal**.
+
+- Hover over your running container and select the **Show container actions**
+  menu. From the drop-down menu, select **Open in terminal**.
 - Select the container and then select the **Exec** tab.
```
sfavello, line 82: I _think_ there's two ways to open the integrated terminal. 

1. Would be clicking the name of the container, which would automatically send user's to the logs tab where they can then click the exec tab (if docker debug is not enabled). 
2. Would be clicking the menu action on a container (the 3 dots) and clicking the `open in terminal` option. When clicking the `open in terminal` option, it should automatically send users to the exec tab. 

I only bring this up because of the step where it says to select the Exec tab. If it's the first way to get to the integrated terminal it'd make sense to keep but if we are going with the second way in going to the integrated terminal, it'd already send users to that tab so they don't need to have that step included.

content/desktop/use-desktop/container.md
```diff
@@ -29,46 +29,92 @@ When you [inspect a container](#inspect-a-container), the **Stats** tab displays
...
+Debug mode is in [Beta](../../release-lifecycle.md#beta) and requires a [Pro, Team, or Business subcription](/subscription/details/). You must be signed in to use debug mode. Debug mode has several advantages, such as:
+
+- A customizable toolbox. The toolbox comes with many standard Linux tools
+  pre-installed, such as `vim`, `nano`, `htop`, and `curl`. For more details, see the [`docker debug` CLI reference](/reference/cli/docker/debug/).
+- The ability to access containers that don't have a shell, for example, slim or
+  distroless containers.
+
+To open the integrated terminal in debug mode, either:
+
+- Hover over your running container and select the **Show container actions**
```
sfavello, line 98: The same thing for Debug mode, if they click on the `Use Docker Debug`, in the menu actions, it means that the user is already logged in and a paid user.
User's shouldn't be able to see that as an option if they are not logged in and a paid user. The only way for them to get to the integrated terminal and use it in debug mode if they are not logged and or not a paid user would be to navigate to the exec tab, where they'd see a paywall banner and either be prompted to login. After logging in, if they are a paid user, they can enable Docker Debug mode. If they are not, they would then see an upgrade option where, if they do upgrade, they'd be able to enable Docker debug. 

I am soo sorry for such a long winded comment btw! I tried cutting it short but I tend to over explain. 🙇🏼 😅  
Basically what I was thinking was that it might confuse users to see the step  
From the drop-down menu, select **Use Docker Debug**. If the **Debug** tab
  isn't visible, select the **Exec** tab and then enable the **Debug mode**
  setting. 
Since if they didn't meet the requirements to use Docker Debug, they won't see the **Use Docker Debug** in the drop-down menu. 
 Again full apologies for such long comments. Please ignore if none of this made sense btw!

content/desktop/use-desktop/container.md
craig-osterhout: I think it is correct, although easy to misinterpret. That unordered list is supposed to be like, do either or. I think it can be misinterpreted as an ordered list without `or` between the choices. I added `Or,` before the second item.

content/desktop/use-desktop/container.md
craig-osterhout: Same as the other comment. I also rearranged it so that the sign in information appears in the steps rather than the intro paragraph.

sfavello (approved):
LGTM! Thanks for making the changes and creating the doc for debug!! This looks fintastic!! 🚀

content/desktop/use-desktop/container.md
```diff
@@ -29,46 +29,102 @@ When you [inspect a container](#inspect-a-container), the **Stats** tab displays
...
-    - Delete unnecessary files when you right-click on a file
-    - Download file and folders from the container straight to the host
+Select **Files** to explore the filesystem of running or stopped containers. You
+can also:
+
+- See which files have been recently added, modified, or deleted
+- Edit a file straight from the built-in editor
+- Drag and drop files and folders between the host and the container
+- Delete unnecessary files when you right-click on a file
+- Download file and folders from the container straight to the host
```
stephaurelio, line 127: Nit to align with the previous list item:
```suggestion
- Download files and folders from the container straight to the host
```

stephaurelio (approved):
LGTM, left a tiny comment

content/desktop/use-desktop/container.md
sfavello: Ah that makes total sense now! I didn't realize! 🙇🏼  thank you for clarifying it for me!!

PR #19590, opened 2024-03-08
Describe a '--internal' network

robmry:
## Description

Add a description of a `--internal` network (written by @neersighted) to the CLI docs - as they currently only mention overlay networks.

(Also, while I was there, update the description for the table listing 'network create' options, to make it consistent with the table listing 'bridge' options.)

## Reviews

We should probably add similar words to [compose docs](https://docs.docker.com/compose/compose-file/06-networks/#internal).

@neersighted 
@dvdksn 

- [ ] Technical review
- [ ] Editorial review
- [ ] Product review

### DIFF
--- a/data/engine-cli/docker_network_create.yaml
+++ b/data/engine-cli/docker_network_create.yaml
@@ -313,7 +313,8 @@ examples: |-
     | `com.docker.network.container_iface_prefix`      | -           | Set a custom prefix for container interfaces          |
 
     The following arguments can be passed to `docker network create` for any
-    network driver, again with their approximate equivalents to `docker daemon`.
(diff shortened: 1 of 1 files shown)


dvdksn:
Thank you @robmry - the verbiage looks good but we'll need to update this in `docker/cli` - we generate this yaml file from there. This is the source of truth: https://github.com/docker/cli/blob/master/docs/reference/commandline/network_create.md

robmry:
> Thank you @robmry - the verbiage looks good but we'll need to update this in `docker/cli` - we generate this yaml file from there. This is the source of truth: https://github.com/docker/cli/blob/master/docs/reference/commandline/network_create.md

Doh! Sorry ... will do.

PR #19599, opened 2024-03-11
Update fedora.md

donaldmurf:
WRONG URL:
sudo dnf config-manager \
    --add-repo \
    https://download.docker.com/linux/fedora/docker-ce.repo
CORRECT URL:
sudo dnf config-manager \
--add-repo=https://download.docker.com/linux/centos/docker-ce.repo

https://stackoverflow.com/questions/70358656/rhel8-fedora-yum-dns-causes-cannot-download-repodata-repomd-xml-for-docker-ce

## Reviews

- [ ] Technical review
- [ ] Editorial review
- [ ] Product review

### DIFF
--- a/content/engine/install/fedora.md
+++ b/content/engine/install/fedora.md
@@ -81,7 +81,8 @@ your DNF repositories) and set up the repository.
 
 ```console
 $ sudo dnf -y install dnf-plugins-core
-$ sudo dnf config-manager --add-repo {{% param "download-url-base" %}}/docker-ce.repo
+$ sudo dnf config-manager \
+--add-repo=https://download.docker.com/linux/centos/docker-ce.repo
 ```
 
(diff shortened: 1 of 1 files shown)


donaldmurf:
The link that is produced in the documentation is incorrect for Rocky 9. I
don’t see the correct link produced in either RHEL or fedora. I believe
Rocky is fedora. Spin up a Rocky 9 vm and follow the docker installation
guide on here. It doesn’t work, which would tell me the documentation isn’t
correct.

dvdksn:
Rocky is RHEL. There's no build for RHEL x86, only s390x. The docs for RHEL cross-reference the centos instruction as a plausible (untested) option for other architectures. https://docs.docker.com/engine/install/rhel/

donaldmurf:
Thank you for the clarification.

PR #19630, opened 2024-03-14
ENGDOCS-2031

aevesdocker:
https://github.com/compose-spec/compose-spec/pull/472/files

## Reviews

- [ ] Editorial review

### DIFF
--- a/content/compose/compose-file/05-services.md
+++ b/content/compose/compose-file/05-services.md
@@ -1418,6 +1418,7 @@ ports:
     host_ip: 127.0.0.1
     published: "8080"
     protocol: tcp
+    app_protocol: http
     mode: host
 
   - name: web-secured
@@ -1425,6 +1426,7 @@ ports:
     host_ip: 127.0.0.1
     published: "8083-9000"
     protocol: tcp
+    app_protocol: https
     mode: host
 ```
 


content/compose/compose-file/05-services.md
```diff
@@ -1407,23 +1407,26 @@ expressed in the short form.
 - `published`: The publicly exposed port. It is defined as a string and can be set as a range using syntax `start-end`. It means the actual port is assigned a remaining available port, within the set range.
 - `host_ip`: The Host IP mapping, unspecified means all network interfaces (`0.0.0.0`).
 - `protocol`: The port protocol (`tcp` or `udp`). Defaults to `tcp`.
+- `app_protocol`: The application procotol (TCP/IP level 4 / OSI level 7) this port is used for. This is optional and can be used as a hint for Compose to offer richer behavior for protocols that it understands.
```
dvdksn, line 1410: suggestion:
```suggestion
- `app_protocol`: Layer 7 (application layer) protocol this port is used for. This is optional and can be used as a hint for Compose to offer richer behavior for protocols that it understands.
```

dvdksn: question: Should we list valid options? Is there an enum of valid options like `http|https|ssh|ftp|...`

dvdksn (approved):

PR #19632, opened 2024-03-14
Update 06_bind_mounts.md with a semantic fix

Ativerc:
`--mount` takes in a few key-value pairs and its`type=bind` that specifies that the volume is bind mount.

## Description

"The `--mount` option tells Docker to create a bind mount, where..." is wrong. 

According to the docs:
`--mount` flag tells docker to create a volume and consists of multiple key-value pairs, separated by commas and each consisting of a `<key>=<value>` tuple, where `type` key's value sets the type of volume to be created: `bind`, `tmpfs`, and `volume`. For bind mounts, it's set to `bind`.

There was an error in the description of the command. I fixed it.

## Reviews

As I'm typing this, I found even my change to be bit lacking, so IMO, the final text should start as:

"The `--mount` option tells Docker to create a volume, where `type` is the type of volume to be created (here, `bind` for bind mounts), and where `src` is...." 

Please change it to that.

Thank you for your time. 

- [ ] Technical review
- [ ] Editorial review
- [ ] Product review

### DIFF
--- a/content/get-started/06_bind_mounts.md
+++ b/content/get-started/06_bind_mounts.md
@@ -79,7 +79,7 @@ setting, see the topic for [Mac](../desktop/settings/mac.md/#file-sharing),
    {{< /tab >}}
    {{< /tabs >}}
    
-   The `--mount` option tells Docker to create a bind mount, where `src` is the
+   The `--mount type=bind` option tells Docker to create a bind mount, where `src` is the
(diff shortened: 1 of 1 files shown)


dvdksn (approved):
LGTM

dvdksn:
Technically bind mounts and tmpfs mounts aren't volumes, so I think your suggestion is good

PR #19645, opened 2024-03-18
Fix verification command

Fuhrmaaj:
## Description

Make the cat command check for the file that the user created.

In the current version, the user makes a file called 'greeting.txt' but then checks for a file called 'greetings.txt'.


dvdksn (approved):
Thx

PR #19685, opened 2024-03-25
Update ubuntu.md

ZvonimirBandic:
added $ sudo service docker start into section 3. Verify that the Docker Engine installaton is successfull

## Description

I have added " sudo service docker start" into section 3, as the Run Hello World would not work without startin docker 

## Reviews

- [ ] Technical review
- [ ] Editorial review
- [ ] Product review

### DIFF
--- a/content/engine/install/ubuntu.md
+++ b/content/engine/install/ubuntu.md
@@ -164,6 +164,7 @@ Docker from the repository.
    `hello-world` image.
 
    ```console
+   $ sudo service docker start
    $ sudo docker run hello-world
    ```
 


dvdksn:
The Docker service starts automatically when you install Docker Engine on Ubuntu using the APT repositories, unless you've somehow tampered with the init system or otherwise deviated from the installation steps we provide. What version of Ubuntu are you running?

ZvonimirBandic:
Running standard Ubuntu 22.04, no changes. I used exactly the steps in 1. and 2.I will try to repoduce again, as I did try to install docker differently before I found this link. Close for now.

PR #19692, opened 2024-03-26
Update Docker Engine API docs to use "go install" for installing the Go SDK

sneakerhax:
## Description

When installing the Docker Go SDK using the instruction at the page [Develop with Docker Engine SDKs -> Go SDK](https://docs.docker.com/engine/api/sdk/#install-the-sdks) I received the following error:
```
go get github.com/docker/docker/client
go: go.mod file not found in current directory or any parent directory.
        'go get' is no longer supported outside a module.
        To build and install a command, use 'go install' with a version,
        like 'go install example.com/cmd@latest'
        For more information, see https://golang.org/doc/go-get-install-deprecation
        or run 'go help get' or 'go help install'.
```

I was able to install using the updated command in the proposed change.

### DIFF
--- a/content/engine/api/sdk/_index.md
+++ b/content/engine/api/sdk/_index.md
@@ -23,7 +23,7 @@ installed and coexist together.
 ### Go SDK
 
 ```console
-$ go get github.com/docker/docker/client
+$ go install github.com/docker/docker/client@latest
 ```
 
 The client requires a recent version of Go. Run `go version` and ensure that you're running a currently supported version of Go.


sneakerhax:
After reviewing it, I see that my pull request doesn't solve the issues.

The module location ```github.com/docker/docker``` no longer exists and points to ```github.com/moby/moby```:

When attempting to install modules from the new location, I get the following error:

```
go install github.com/moby/moby@latest       
go: github.com/moby/moby@latest: module github.com/moby/moby@latest found (v26.0.0+incompatible), but does not contain package github.com/moby/moby
```

The goal here is to be able to develop code with the Docker Go SDK.

dvdksn:
Hi, the error suggests you're not using Go modules for your project. Run `go mod init` to start using modules. See this recording:

[asciicast](https://asciinema.org/a/xMotZVgpTY2ci2G8wFaKm9ynr)

PR #19697, opened 2024-03-27
ds-950

aevesdocker:
Adds another way users can self-diagnose. For DD4.29

## Reviews

- [x] Technical review
- [ ] Editorial review
- [x] Product review


KatTomrushka (approved):
Good to go! Thanks, Allie

PR #19782, opened 2024-04-11
DS-884 Add an admin setting for displayedOnboarding

gavin-timbrell:
## Description

Adds the new admin setting for `displayedOnboarding` that's arriving in the next release to the docs.

## Reviews

- [ ] Product review


gavin-timbrell:
Thanks for the quick review @aevesdocker! I've added it to that list now

PR #19789, opened 2024-04-12
FIX: Update 05_persisting_data.md (error in the cmd for running a container and mounting a volume in Git Bash)

sudarshanmg:
## Description
An extra space space after:
`--mount type=volume,src=todo-db,`

gives the following error:

invalid argument "type=volume,src=todo-db" for "--mount" flag: target is required See 'docker run --help'.
image

---

## Reviews

After removing the extra space, it works as expected:

image


dvdksn (approved):
LGTM thanks!

PR #19867, opened 2024-04-24
Update Dotnet Containerization Docs

jpbulman:
## Description
This condenses two separate commands to connect to a docker container and to log into postgres into one.


craig-osterhout:
Thank you for updating this. For this change, line 203 and line 207 also need to be updated as the user will now only need to run `exit` once to exit both.

jpbulman:
@craig-osterhout Nice catch! Updated

craig-osterhout (approved):
LGTM

PR #19900, opened 2024-04-26
Add note that docker build are not always protected by ECI.

ctalledo:
## Description

**NOTE: Wait until the Docker Desktop 4.30 release to publish this change.**

Adds a note in the Enhanced Container Isolation (ECI) limitations section regarding docker build protections.

## Reviews

- [ ] Technical review
- [X] Editorial review
- [ ] Product review


ctalledo:
> Thanks @ctalledo, looks fine to me. I noticed that `limitations.md` is no longer visible in the table of content for some reason 🤔 . Would you mind adding it to the `toc.yaml` file?

Thanks Allie (@aevesdocker); yes I noticed that too, I think it used to be tab within the ECI FAQs doc, but got lost when the ECI FAQs were moved to the security FAQs section. 

I've added it to the `toc.yaml` file now. 

As an aside, I liked it better when the ECI FAQs where under the ECI section, compared to where they are now (the security FAQ section). That's because the FAQs were closer to the most relevant content and could be more easily spotted from the TOC.

PR #19946, opened 2024-05-01
add fuse.js extended search

craig-osterhout:
## Description

Added extended search option to enable fuzzy search.

It does return more results that may or may not be appreciated. It doesn't feel slower.

### DIFF
--- a/assets/js/src/search.js
+++ b/assets/js/src/search.js
@@ -20,6 +20,7 @@ async function initializeIndex() {
     ],
     minMatchCharLength: 2,
     threshold: 0.2,
+    useExtendedSearch: true,
   };
 
   handler = new Fuse(index, options);


assets/js/src/search.js
```diff
@@ -20,6 +20,7 @@ async function initializeIndex() {
     ],
     minMatchCharLength: 2,
     threshold: 0.2,
+    useExtendedSearch: true,
```
dvdksn, line 23: I think extended search is for enabling operators: https://www.fusejs.io/examples.html#extended-search

The fuzzy search is controlled by threshold, I believe

craig-osterhout: Thanks! I misinterpreted the doc as extended search enabled all the stuff in the table below.
It does seem to work for the specific case of searching for `create org`. I'll play with the threshold to see if it makes a difference.

craig-osterhout: - With UseExtendedSearch, and bumping the threshold up to 0.3, `create org` finds the topic and no other topics.
- With UseExtendedSearch, and leaving the threshold at 0.2, `create org` finds the topic and no other topics.
- With no UseExtendedSearch, and bumping the threshold up to 0.3, `create org` finds the topic plus 12 other topics. 
- With no UseExtendedSearch, and leaving the threshold at 0.2, `create org` finds nothing.

dvdksn: Interesting! I wonder why that is... I'm not opposed to enabling extended search anyway though

dvdksn (approved):

assets/js/src/search.js
craig-osterhout: I can't find a clear answer either.
It doesn't work as good in some cases too.
`create a` returns result with our current config, but returns nothing with extended search.

craig-osterhout: I think I got it. Extended search treats each word as a separate token unless quoted. Then merges the results from each token using AND logic. In the case above, `a` is less then the minimum of 2 characters, and has no results. When the results from searching `a` (nothing) are AND'd with the results from `create`, nothing shows. 

Non-extended search treats the whole string as one token.

dvdksn (changes requested):
On closer review, I find that the fuse.js without extended search seems to provide better matches in most cases. I think that for the cases where we don't get the match we expect, maybe we should try to work the metadata instead. For example, by adding keywords. I noticed for example that we don't have the `org` keyword anywhere, but maybe it would be good to add? (We have the full "organization" but fusejs doesn't do partial word matches by default I guess? so org != organization. We do have "organization" as a keyword, but that is unnecessary since we already have that word in the title.)

craig-osterhout:
Closing this. Will create a new PR to add a keyword to at least fix the "create org" issue without breaking other searches.

PR #19950, opened 2024-05-02
fix: docker desktop macos completions

Benehiko:
## Description

Removed the shell completion section in the FAQ for Docker Desktop Mac completions, and created a new, dedicated page for it.


dvdksn:
@Benehiko I've pushed an update here, PTAL

content/config/completion.md
```diff
@@ -0,0 +1,96 @@
...
+
+After installing `bash-completion`, source the script in your shell
+configuration file (in this example, `.bashrc`):
+
+```console
+$ cat <<EOT >> ~/.bashrc
+if [ -f /etc/bash_completion ]; then
+    . /etc/bash_completion
+fi
+```
```
Benehiko, line 44: This is tricky since installing with brew you need to use the `brew --prefix` command to get the path to where brew installs the binaries. So instead of `/etc/bash_completion` you would run `$(brew --prefix)/etc/bash_completion`.

Like this example here:

```console
$ cat <<EOT >> ~/.bash_profile
[[ -r "$(brew --prefix)/etc/profile.d/bash_completion.sh" ]] && . "$(brew --prefix)/etc/profile.d/bash_completion.sh"
EOT
```

content/config/completion.md
```diff
@@ -0,0 +1,96 @@
...
+
+- [Bash](#bash)
+- [Zsh](#zsh)
+- [fish](#fish)
+- PowerShell
+
+## Bash
+
+Bash does not support completions natively, and requires the `bash-completion` package to be installed.
+
```
Benehiko, line 22: I was wrong to say `bash` does not support completions natively.

https://www.gnu.org/savannah-checkouts/gnu/bash/manual/bash.html#Programmable-Completion

 https://tuzz.tech/blog/how-bash-completion-works#how-completion-functions-work

But it's a cumbersome setup. So we just simplify this to, to get completions working correctly use `bash-completion`.

Benehiko (commented):
Looks good! Just some minor comments

content/config/completion.md
```diff
@@ -0,0 +1,106 @@
...
+You can generate completion scripts for the following shells:
+
+- [Bash](#bash)
+- [Zsh](#zsh)
+- [fish](#fish)
+
+## Bash
+
+To Get Docker CLI completion with Bash, you first need to install the
```
Benehiko, line 20:
```suggestion
To get Docker CLI completion with Bash, you first need to install the
```

content/config/completion.md
```diff
@@ -0,0 +1,106 @@
...
+fi
+EOT
+
+# On macOS / with Homebrew:
+cat <<EOT >> ~/.bashrc
+if [ -f $(brew --prefix)/etc/bash_completion ]; then
+    . $(brew --prefix)/etc/bash_completion
+fi
+EOT
+```
```
Benehiko, line 54: Sorry I should've been more specific here. This is shown after installing `bash-completion` with homebrew.

```suggestion
cat <<EOT >> ~/.bash_profile
[[ -r "$(brew --prefix)/etc/profile.d/bash_completion.sh" ]] && . "$(brew --prefix)/etc/profile.d/bash_completion.sh"
EOT
```

Benehiko: https://formulae.brew.sh/formula/bash-completion@2#default

Benehiko (commented):
LGTM

dvdksn (approved):

PR #19953, opened 2024-05-02
Engdocs 2080

aevesdocker:
## Description

PR creates includes for the Compose GUI feature. No new content, just rejigging.

## Reviews

- [ ] Technical review
- [ ] Editorial review
- [ ] Product review

### DIFF
--- a/content/compose/compose-file/10-fragments.md
+++ b/content/compose/compose-file/10-fragments.md
@@ -4,7 +4,7 @@ description: Understand how to use fragments
 keywords: compose, compose specification, fragments, compose file reference
 ---
 
-With Compose, you can use built-in [YAML](https://www.yaml.org/spec/1.2/spec.html#id2765878) features to make your Compose file neater and more efficient. Anchors and aliases let you create re-usable blocks. This is useful if you start to find common configurations that span multiple services. Having re-usable blocks minimizes potential mistakes.
+{{< include "compose/fragments.md" >}}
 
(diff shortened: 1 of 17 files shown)


dvdksn:
How is Docker Desktop pulling these descriptions? If it's happening runtime, I think we need an intermediate layer, such as a JSON API like we have for the search index, which decouples the source files and the distribution of headless fragments.

dvdksn:
> How is Docker Desktop pulling these descriptions? If it's happening runtime, I think we need an intermediate layer, such as a JSON API like we have for the search index, which decouples the source files and the distribution of headless fragments.

I discussed this with @jhrotko and I think we need to provide a more reliable endpoint that Docker Desktop can use to fetch these strings. For the same reason that we use `/go/` links, to avoid unintentionally breaking an application. Fetching these strings from the markdown files is too brittle.

I can do the implementation. I think we should serve these as a JSON file. We may need to serve other applications in a similar way in the future, so if we want to be future-proof we should make this a generic implementation. The most straightforward and generic approach is to just assign each string a unique identifier:

```json
{
  "composeBuild": "Build is an optional part of the Compose Specification. It tells Compose how to (re)build an application from source and lets you define the build process within a Compose file in a portable way.",
  "composeVolumes": "Volumes are persistent data stores implemented by the container engine. Compose offers a neutral way for services to mount volumes, and configuration parameters to allocate them to infrastructure. The top-level `volumes` declaration lets you configure named volumes that can be reused across multiple services."
}
```

We should probably stick a `/go/` redirect in front of this too.

dvdksn (approved):
Feel free to merge @aevesdocker and I can open a follow-up

PR #19958, opened 2024-05-03
Update rootless.md

adamoutler:
Adding quotes prevents the `$` from being omitted from the beginning of this line.  When a line starts with `$` it is assumed to be an included shell prompt by the formatter and so does get coped.

## Reviews

- [ ] Technical review
- [ ] Editorial review
- [ ] Product review


dvdksn (approved):
Thank you! 🙏🏻

Issue #19933, opened 2024-05-01
Missing instructions in ubuntu install from repo


Kristianwhittick:
### Is this a docs issue?

- [X] My issue is about the documentation content or website

### Type of issue

Information is incorrect

### Description

Missing command in step 3 of section **Install using the apt repository**, the command is to run the hello world example, but the daemon has not been started.

Command
```
sudo docker run hello-world`
```

Output
```
docker: Cannot connect to the Docker daemon at unix:///var/run/docker.sock. Is the docker daemon running?.
See 'docker run --help'.
```

The command is present in step 6 of **Install from a package**. Although it does state at the end of section 5 that the daemon will automatically start.
```
sudo service docker start
sudo docker run hello-world
```

### Location

https://docs.docker.com/engine/install/ubuntu/
https://docs.docker.com/engine/install/debian/

### Suggestion

Please add the missing daemon start command to step 3 of section **Install using the apt repository**,

Also, add the comment to the effect for both section 3 & 6, alternatively the daemon will auto-start on Debian and Ubuntu when the system is rebooted. This comment is present in the "**Linux post-installation steps for Docker Engine**" page. 

> On Debian and Ubuntu, the Docker service starts on boot by default.To automatically start Docker and containerd on boot for other Linux distributions using systemd, run the following commands:

PR #19981, opened 2024-05-07
fix: ubuntu/debian apt install message

EmilyG2413:
## Description
Added the necessary instructions as described in the issue to start the daemon.

## Related issues or tickets
Git issue: #19933


dvdksn:
The engine should start automatically when you install using the apt package. If this isn't what you're seeing, may I ask what version of Ubuntu or Debian you're using?

PR #19989, opened 2024-05-07
Fix URL used for images in twitter/og previews

mikesir87:
## Description

When using docs in Twitter/other locations, the preview cards aren't pulling an image due to an invalid URL in the `meta` tag. Example here:

```html
<meta name=twitter:image:src content="https://docs.docker.com//assets/images/thumbnail.webp">
```

The additional `/` is causing the image to 404. This PR removes the extra slash.

### DIFF
--- a/layouts/_default/hub-api-baseof.html
+++ b/layouts/_default/hub-api-baseof.html
@@ -11,7 +11,7 @@
   <!-- favicon -->
   {{ partialCached "favicon.html" "favicon" }}
   <!-- make the latest API version the canonical page as that's what we want users to be using mostly -->
-  <link rel="canonical" href="{{ site.BaseURL }}/docker-hub/api/latest/" />
+  <link rel="canonical" href="{{ "/docker-hub/api/latest/" | relURL }}" />
   <style>
     body {
       margin: 0;
--- a/layouts/index.robots.txt
+++ b/layouts/index.robots.txt
@@ -2,7 +2,7 @@
 User-agent: *
 
 
-Sitemap: {{ site.BaseURL }}/sitemap.xml
+Sitemap: {{ "sitemap.xml" | absURL }}
 {{- else -}}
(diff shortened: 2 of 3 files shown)


thaJeztah:
I notice that baseURL is used in many other locations, so wondering if the correct fix would be for that to not include the trailing Slash, eg https://github.com/docker/docs/blob/67d6da5cb50f6f95e7bd4adfee3cd92de331cc08/layouts/index.robots.txt#L5

But looking at the config, we configure without the trailing slash, so now I'm wondering where it comes from https://github.com/docker/docs/blob/67d6da5cb50f6f95e7bd4adfee3cd92de331cc08/hugo.yaml#L2

thaJeztah:
@dvdksn any idea?

mikesir87:
Good callout @thaJeztah. I'm guessing Hugo is appending the slash at the end 🤷 

I found [an issue](https://github.com/gohugoio/hugo/issues/3262) where it was mentioned that we should be using `relURL` and `absURL` to generate URLs. So, just pushed another small commit to update the meta tags and other locations we were building using baseURL (such as the sitemap reference you mentioned).

dvdksn (approved):
Changes look good. I think this got broken quite recently, for reasons I don't know. We've had double slashes in a couple of these links for a long time. Just the other week the double slash caused the favicon to disappear, so I had to push a similar fix. And now this.

"Normalizing" double slashes in a request URL is really something I would expect a web server to be able to handle, and I think it used to do that. I don't know if it's on S3 or CloudFront, but something changed somewhere, causing these isses to appear.

thaJeztah:
> Good callout @thaJeztah. I'm guessing Hugo is appending the slash at the end 🤷

Yes, this was indeed the part I was looking at, and made me wonder "are we doing the right thing, or are we tapering over some underlying issue?"

Looks like we found the correct approach for these; thanks both!

PR #19992, opened 2024-05-07
Update rootless.md

adamoutler:
I just accomplished install via both manual and docker-ce-rootless-extras packages on Ubuntu 24.04.  There were some pain points which I hope to address here and make this run more smoothly for others. 

* As a `$` user, you cannot `cat<< EOF > /etc/apparmor.d`.  The existing command should never work because you cannot be the user and access the apparmor folder at the same time. So we create the file and then move the file.  This separates into creation which should be done as a user, and moving which requires sudo access.  If the move command fails, that's a basic issue which can be resolved by an admin. 
* I added missing apt-get commands which may be required on Ubuntu.  I noticed `dbus-user-session` is a default package in Ubuntu and it was mentioned. So i took the liberty to add uidmap which is absolutely required, and systemd-container may be necessary for accessing via non-login terminals.
* The command `sudo systemctl disable --now docker.service docker.socket` was insufficient and requires a reboot to activate.  The `sudo systemctl stop docker.service docker.socket` will immediately shut down the docker service. 
* I added a note about not shutting down the docker service.
* Troubleshooting now includes an installation error where the user was a pure service account without login capabilities. `systemd-container` and `machinectl` is required when not logging in via terminal with user/pass.

## Reviews

- [ ] Technical review
- [ ] Editorial review
- [ ] Product review

### DIFF
--- a/content/engine/security/rootless.md
+++ b/content/engine/security/rootless.md
@@ -437,6 +437,7 @@ For example:
 ## Troubleshooting
 
 ### Unable to install with systemd when systemd is present on the system
+
 ``` console
 $ dockerd-rootless-setuptool.sh install
 [INFO] systemd not detected, dockerd-rootless.sh needs to be started manually:
--- a/content/engine/security/rootless.md
+++ b/content/engine/security/rootless.md
@@ -436,8 +436,12 @@ For example:
 
 ## Troubleshooting
 
-### Errors installing the Docker daemon
-
+### Unable to install with systemd when systemd is present on the system
+``` console
+$ dockerd-rootless-setuptool.sh install
+[INFO] systemd not detected, dockerd-rootless.sh needs to be started manually:
+...
+```
 `rootlesskit` cannot detect systemd properly if you switch to your user via `sudo su`. For users which cannot be logged-in, you must use the `machinectl` command which is part of the `systemd-container` package. After installing `systemd-container` switch to `myuser` with the following command:
 ``` console
 $ sudo machinectl shell myuser@
--- a/content/engine/security/rootless.md
+++ b/content/engine/security/rootless.md
@@ -170,7 +170,7 @@ testuser:231072:65536
 > If the system-wide Docker daemon is already running, consider disabling it:
 >```console
 >$ sudo systemctl disable --now docker.service docker.socket
->$ sudo systemctl stop docker.service docker.socket
+>$ sudo rm /var/run/docker.sock
 >```
 > Should you choose not to shut down the `docker` service and socket, you will need to use the `--force`
 > parameter in the next section. There are no known issues, but until you shutdown and disable you're
--- a/content/engine/security/rootless.md
+++ b/content/engine/security/rootless.md
@@ -172,8 +172,8 @@ testuser:231072:65536
 >$ sudo systemctl stop docker.service docker.socket
 >```
 > Should you choose not to shut down the docker service and socket, you will need to use the `--force`
-> parameter in the next section.  There are no known issues, but until you shutdown and disable you're
-> still running rooted Docker. 
+> parameter in the next section. There are no known issues, but until you shutdown and disable you're
+> still running rootful Docker. 
 
 {{< tabs >}}
 {{< tab name="With packages (RPM/DEB)" >}}
--- a/content/engine/security/rootless.md
+++ b/content/engine/security/rootless.md
@@ -73,7 +73,8 @@ testuser:231072:65536
   profile for `rootlesskit` manually:
 
   1. Create and install the currently logged-in user's AppArmor profile:
-     ``` console
+
+     ```console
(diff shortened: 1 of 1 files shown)


content/engine/security/rootless.md
```diff
@@ -70,21 +72,21 @@ testuser:231072:65536
   script](https://get.docker.com/rootless), however, you must add an AppArmor
   profile for `rootlesskit` manually:
 
-  1. Add the AppArmor profile to `/etc/apparmor.d/usr.local.bin.rootlesskit`:
-
-     ```console
-     $ cat <<EOF > /etc/apparmor.d/$(echo $HOME/bin/rootlesskit | sed -e s@^/@@ -e s@/@.@g)
+  1. Create and install the currently logged-in user's AppArmor profile:
+     ``` console
```
dvdksn, line 76: formatting nit:

```suggestion
  1. Create and install the currently logged-in user's AppArmor profile:

     ```console
```

content/engine/security/rootless.md
```diff
@@ -167,7 +169,11 @@ testuser:231072:65536
 > If the system-wide Docker daemon is already running, consider disabling it:
 >```console
 >$ sudo systemctl disable --now docker.service docker.socket
+>$ sudo systemctl stop docker.service docker.socket
 >```
+> Should you choose not to shut down the docker service and socket, you will need to use the `--force`
```
dvdksn, line 174:
```suggestion
> Should you choose not to shut down the `docker` service and socket, you will need to use the `--force`
```

content/engine/security/rootless.md
```diff
@@ -167,7 +169,11 @@ testuser:231072:65536
...
+> parameter in the next section.  There are no known issues, but until you shutdown and disable you're
+> still running rooted Docker.
```
dvdksn, line 176:
```suggestion
> parameter in the next section. There are no known issues, but until you shutdown and disable you're
> still running rootful Docker. 
```

content/engine/security/rootless.md
```diff
@@ -429,6 +435,14 @@ For example:
 
 ## Troubleshooting
 
+### Errors installing the Docker daemon
+
+The `rootlesskit` cannot detect Systemd properly if you switch to your user via `sudo su`.  For users which cannot be logged-in, you must utilize the `machinectl` command which is a part of the `systemd-container` package. After installing `systemd-container` switch to "myuser" with the following command:
```
dvdksn, line 440:
```suggestion
`rootlesskit` cannot detect systemd properly if you switch to your user via `sudo su`. For users which cannot be logged-in, you must use the `machinectl` command which is part of the `systemd-container` package. After installing `systemd-container` switch to `myuser` with the following command:
```

dvdksn (commented):
cc @AkihiroSuda

content/engine/security/rootless.md
```diff
@@ -70,21 +72,22 @@ testuser:231072:65536
   script](https://get.docker.com/rootless), however, you must add an AppArmor
   profile for `rootlesskit` manually:
 
-  1. Add the AppArmor profile to `/etc/apparmor.d/usr.local.bin.rootlesskit`:
+  1. Create and install the currently logged-in user's AppArmor profile:
 
      ```console
-     $ cat <<EOF > /etc/apparmor.d/$(echo $HOME/bin/rootlesskit | sed -e s@^/@@ -e s@/@.@g)
```
AkihiroSuda, line 76: Maybe just instruct to run the old commands as root

content/engine/security/rootless.md
```diff
@@ -167,7 +170,11 @@ testuser:231072:65536
 > If the system-wide Docker daemon is already running, consider disabling it:
 >```console
 >$ sudo systemctl disable --now docker.service docker.socket
+>$ sudo systemctl stop docker.service docker.socket
```
AkihiroSuda, line 173: Why doesn't `disable --now` work?

content/engine/security/rootless.md
adamoutler: Not all users have `sudo` access and the old command includes  `$HOME` which will always be `/root` when run as root. The old command appears to have never been and is not as versatile as the new.

content/engine/security/rootless.md
```diff
@@ -429,6 +436,14 @@ For example:
 
 ## Troubleshooting
 
+### Errors installing the Docker daemon
```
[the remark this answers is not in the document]
adamoutler: Updated.

content/engine/security/rootless.md
adamoutler: I suppose it's just the docker socket is still present on the system.  `/var/run/docker.socket` needs to be forcefully removed in this step rather than simply `stop` on the service and socket.  This allows the user to run `dockerd-rootless-setuptool.sh install` without the `--force` parameter.

AkihiroSuda (approved):
Thanks

dvdksn (approved):

PR #20049, opened 2024-05-16
Networking overview - DNS fixes

robmry:
## Description

Two DNS-related tweaks to the network-overview page ...

Since 26.0.0, IPv6 nameservers in the host's /etc/resolv.conf file have been treated the same as IPv4 nameservers:
- The internal DNS server will use them as upstream servers, they won't be left in the container's resolv.conf. (So, there's no longer any race between the internal server and IPv6 upstream servers when using musl-libc.)

A '--dns' server will not be replaced by default DNS:
-  Google's DNS servers are currently used by containers on the default bridge network, when none of the host's /etc/resolv.conf nameservers are usable (host loopback addresses that the container can't see). _That should change in 27.0, the internal DNS server will be used for the default network too._
- But, if a server is supplied via the --dns option, it'll just appear in the container's resolv.conf (on the default bridge) or as an upstream server (for the internal resolver).
- So, fix the networking-intro page's description, noting that requests to --dns servers happen in the container's namespace.

## Related issues or tickets

https://github.com/moby/moby/pull/47512 - Add IPv6 nameserver to the internal DNS's upstreams.

## Reviews

@dvdksn, @akerouanton

### DIFF
--- a/content/network/_index.md
+++ b/content/network/_index.md
@@ -174,22 +174,6 @@ configuration.
 | `--dns-opt`    | A key-value pair representing a DNS option and its value. See your operating system's documentation for `resolv.conf` for valid options.                                                                                                                            |
 | `--hostname`   | The hostname a container uses for itself. Defaults to the container's ID if not specified.                                                                                                                                                                          |
 
-### Nameservers with IPv6 addresses
-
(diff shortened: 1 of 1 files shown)


akerouanton (approved):
LGTM

PR #20095, opened 2024-05-23
style guide addition

aevesdocker:
## Description

Received a request to add this to our Style Guide. 
Was originally in our UI copy specific guidelines, but is a better fit here and too detailed for the ui copy SG

## Reviews

- [ ] Technical review
- [ ] Editorial review
- [ ] Product review


craig-osterhout (approved):
LGTM. Have you considered adding a vale rule because this will be a hard habit to break.

PR #20133, opened 2024-06-02
The word "container" should be changed to "image" for factual correct…

ItzDeejey:
…ness.

The term "container" should be replaced with "image" for accuracy, as we are searching for a specific image from the local images to run a new container.


craig-osterhout (approved):
Thank you.

ItzDeejey:
> Thank you.

You are welcome :)

PR #20203, opened 2024-06-11
clarify support for PWD/.env vs project_dir/.env

ndeloof:
## Description

follow up for https://github.com/docker/docs/pull/20188
Clarify distinction between PWD/.env and project directory when COMPOSE_FILE is set

## Related issues or tickets

https://docker.atlassian.net/browse/COMP-582

## Reviews

- [ ] Technical review
- [ ] Editorial review
- [ ] Product review


content/compose/environment-variables/variable-interpolation.md
```diff
@@ -211,3 +212,42 @@ If an environment variable is not set, Compose substitutes with an empty string.
 > **Note**
 >
 > `postgres:` is not a valid image reference. Docker expects either a reference without a tag, like `postgres` which defaults to the latest image, or with a tag such as `postgres:15`.
+
+
+#### local `.env` file _vs_ <project directory> `.env` file
+
+An `.env` file can also be used to declare [`COMPOSE_*` variables](envvars.md) used to control Compose behavior and files to be loaded. When executed without an
+explicit `--env-file` flag, Compose will search for a `.env` file in working directory (PWD) and load values both for self-configuration and interpolation. If
```
jhrotko, line 220: if you mean the command -> `pwd` I would change it to this and maybe add a reference https://man7.org/linux/man-pages/man1/pwd.1.html

content/compose/environment-variables/variable-interpolation.md
```diff
@@ -211,3 +212,42 @@ If an environment variable is not set, Compose substitutes with an empty string.
...
+  db:
+    image: "postgres:9.3"
+```
+
+
+#### diagnose interpolatin issue
+
+Variable can be declared by multiple sources to be used by interpolation. Precedence applies as:
+1. variables from shell environment
+2. if `--env-file` is not set, variables set by an `.env` file in local working directory (PWD)
```
jhrotko, line 250: here I would just re-write it to `pwd`

content/compose/environment-variables/variable-interpolation.md
```diff
@@ -211,3 +212,42 @@ If an environment variable is not set, Compose substitutes with an empty string.
...
+services:
+  db:
+    image: "postgres:9.3"
+```
+
+
+#### diagnose interpolatin issue
+
+Variable can be declared by multiple sources to be used by interpolation. Precedence applies as:
+1. variables from shell environment
```
jhrotko, line 249: nit: instead of 1. 2. 3. you can write 1. 1. 1. and if in the future the order changes you dont need to mind about the numbers ;)

content/compose/environment-variables/variable-interpolation.md
```diff
@@ -211,3 +212,42 @@ If an environment variable is not set, Compose substitutes with an empty string.
...
+2. if `--env-file` is not set, variables set by an `.env` file in local working directory (PWD)
+3. variables from file set by `--env-file` or an `.env` file in project directory
```
jhrotko, line 251: What about `COMPOSE_FILE` and the explicitly envs inside the compose config?

content/compose/environment-variables/variable-interpolation.md
ndeloof: Hopefully this won't change (again), so much regression we had to manage with this already 😓

content/compose/environment-variables/variable-interpolation.md
ndeloof: AFAICT this is the purpose of the "local `.env` file _vs_ <project directory> `.env` file" section
or do you have something in mind ?

content/compose/environment-variables/variable-interpolation.md
glours, line 220: nit:
```suggestion
explicit `--env-file` flag, Compose will search for a `.env` file in working directory (`PWD`) and load values both for self-configuration and interpolation. If
```

content/compose/environment-variables/variable-interpolation.md
```diff
@@ -211,3 +212,42 @@ If an environment variable is not set, Compose substitutes with an empty string.
...
+
+
+#### local `.env` file _vs_ <project directory> `.env` file
+
+An `.env` file can also be used to declare [`COMPOSE_*` variables](envvars.md) used to control Compose behavior and files to be loaded. When executed without an
+explicit `--env-file` flag, Compose will search for a `.env` file in working directory (PWD) and load values both for self-configuration and interpolation. If
+those values define `COMPOSE_FILE` variable which results into project directory being set to another folder, then the latter is also used to load a second
+`.env` file, which will have lower precedence. 
+
+This mechanism makes it possible to invoke an existing compose project with a custom set of variables as overrides, without the need to pass environment
```
glours, line 224: nit:
```suggestion
This mechanism makes it possible to invoke an existing Compose project with a custom set of variables as overrides, without the need to pass environment 
```

content/compose/environment-variables/variable-interpolation.md
glours, line 250: nit:
```suggestion
2. if `--env-file` is not set, variables set by an `.env` file in local working directory (`PWD`)
```

glours (approved):
Sounds good to me

content/compose/environment-variables/variable-interpolation.md
jhrotko: Sorry :(

content/compose/environment-variables/variable-interpolation.md
jhrotko: Just thought we could add all the possibilities in this list

PR #20232, opened 2024-06-13
search: add additional Heap properties to event

craig-osterhout:
## Description

Updated existing search to add attributes for index and search query to each result link.
Added listener to send event to Heap when a search result link is clicked.

Tested on Heap dev.

## Related issues or tickets

ENGDOCS-2124

## Reviews

- [ ] Editorial review

### DIFF
--- a/layouts/partials/search.html
+++ b/layouts/partials/search.html
@@ -1,19 +1,19 @@
 <div x-data="{ open: false }"
   @keyup.escape.window="open = false; if (!open) { document.body.classList.remove('overflow-hidden'); }"
   @keydown.window="(e) => {
-switch(e.key) {
-case 'k':
-  if (e.metaKey || e.ctrlKey) {
-     e.preventDefault()
-     open = !open;
-     if (open) {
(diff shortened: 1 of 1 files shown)


layouts/partials/search.html
```diff
@@ -89,6 +88,26 @@
       }
 
       modalSearchInput.addEventListener("input", modalSearch);
+
+      // Event delegation for tracking link clicks
+      modalSearchResults.addEventListener('click', function(event) {
+        if (event.target.tagName === 'A' && event.target.closest('.link')) {
+          if (window.heap === undefined) {
+            return;
+          }
```
dvdksn, line 97: nit: I guess we could add this condition before even attaching the eventlistener.

```
if !(window.heap === undefined) {
  modalSearchResults.addEventListener(...)
}
```

craig-osterhout: Updated in the commit below. Also updated it from capturing the full target url to just the target path.

dvdksn (approved):
LGTM, thanks Craig!

Issue #17405, opened 2023-05-25
"apt-get remove" command doesn't remove installed packages if a package isn't found


noamraph:
### Is this a docs issue?

- [X] My issue is about the documentation content or website

### Type of issue

Information is incorrect

### Description

The docs instruct to remove old versions by running:

```
sudo apt-get remove docker docker-engine docker.io containerd runc
```

However, it turns out that if a package isn't found, the command fails entirely:

![image](https://github.com/docker/docs/assets/1553464/397522a9-2d42-4e10-b873-aa33df2add93)

I'm on Ubuntu 22.04, WSL2.

### Location

https://docs.docker.com/engine/install/ubuntu/

### Suggestion

Perhaps replace the command with something like this:

```
for pkg in docker docker-engine docker.io containerd runc; do sudo apt-get remove -y $pkg; done
```

This would remove all packages, regardless of whether packages are known or not.

dvdksn:
Thanks @noamraph! I've pushed your suggestion in #17415

PR #20271, opened 2024-06-21
Do not use for loops for package removal on debian-based systems

insanolanbiri:
## Description

In the docker engine installation instructions, it is advised to remove unofficial packages in the following way:

```shell
for pkg in <packages>; do sudo apt-get remove $pkg; done
``` 

This invokes apt-get many times, resulting in having to confirm every uninstallation.

This PR changes such instructions to the following, easier-to-read and easier-to-execute format:

```shell
sudo apt-get remove <packages>
``` 

## Reviews

- [ ] Technical review
- [ ] Editorial review
- [ ] Product review

### DIFF
--- a/content/engine/install/debian.md
+++ b/content/engine/install/debian.md
@@ -59,7 +59,7 @@ conflicts with the versions bundled with Docker Engine.
 Run the following command to uninstall all conflicting packages:
 
 ```console
-$ for pkg in docker.io docker-doc docker-compose podman-docker containerd runc; do sudo apt-get remove $pkg; done
+$ sudo apt-get remove docker.io docker-doc docker-compose podman-docker containerd runc
 ```
 
 `apt-get` might report that you have none of these packages installed.
--- a/content/engine/install/raspberry-pi-os.md
+++ b/content/engine/install/raspberry-pi-os.md
@@ -61,7 +61,7 @@ conflicts with the versions bundled with Docker Engine.
 Run the following command to uninstall all conflicting packages:
 
 ```console
-$ for pkg in docker.io docker-doc docker-compose podman-docker containerd runc; do sudo apt-get remove $pkg; done
+$ sudo apt-get remove docker.io docker-doc docker-compose podman-docker containerd runc
 ```
 
 `apt-get` might report that you have none of these packages installed.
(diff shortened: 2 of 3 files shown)


content/engine/install/debian.md
```diff
@@ -59,7 +59,7 @@ conflicts with the versions bundled with Docker Engine.
 Run the following command to uninstall all conflicting packages:
 
 ```console
+$ sudo apt-get remove docker.io docker-doc docker-compose podman-docker containerd runc
```
thaJeztah, line 62: Using a loop was an intentional change; without a loop, this command may fail, and some packages still installed;  see

- https://github.com/docker/docs/pull/17415
- https://github.com/docker/docs/issues/17405


> resulting in having to confirm every uninstallation.

Yeah, that's rather inconvenient, and II can see that being inconvenient. Not sure what a better solutions would be for that, without this step becoming overly complex. We could detect which packages are installed, but probably that would make it either an extra step, or "complex",

If your situation allows it, and needs to be run non-interactively, it's of course use your own variant 😅

thaJeztah (changes requested):
I left a comment, but perhaps someone has good suggestions to "have both" 😅

content/engine/install/debian.md
insanolanbiri: I've just looked to these issues, and tested a few package removal combinations. Turns out, apt-get only fails if a package does not exist. That is, when there is no such package in the package repository. When a *real* package is not installed, apt continues normally:

```
$ sudo apt-get remove docker.io podman-docker
Reading package lists... Done
Building dependency tree... Done
Package 'podman-docker' is not installed, so not removed
...
The following packages will be REMOVED:
  docker.io
0 upgraded, 0 newly installed, 1 to remove and 0 not upgraded.
After this operation, 151 MB disk space will be freed.
Do you want to continue? [Y/n] 
```

According to docs itself, these guides are for debian 11 or higher, ubuntu 20.04 or higher, and their derivatives. All of the packages in the changed lines are in the repositories already. So, I think my change would not break something, or would it?

In #17405, the problem was related with nonexistent docker-engine package. Looks like that package is already removed from here.

dvdksn: I'm not sure when the `docker-engine` package was removed from here but indeed, `apt-get` will only skip and not error out if the package exists but was not installed. Given that we don't list any nonexistent packages here, I think we should be safe to remove the for-loop now. WDYT @thaJeztah?

Besides, it's been a pretty long time since we shipped packages under those nonexistent (e.g. `docker-engine`) names, no? We're only backwards compatible with oldstable, so this should be fine I think.

PR #20292, opened 2024-06-25
cloud-eap-link-fix

aevesdocker:
## Description

Makes link relative

## Reviews

- [ ] Technical review
- [ ] Editorial review
- [ ] Product review

### DIFF
--- a/data/redirects.yml
+++ b/data/redirects.yml
@@ -701,7 +701,7 @@
 
 # Run Cloud links
 "/":
-  - /go/run-cloud-eap
+  - /go/run-cloud-eap/
 
 # CLI backlinks
 "/config/filter/":


data/redirects.yml
```diff
@@ -700,7 +700,7 @@
   - /go/docker-build-cloud/
 
 # Run Cloud links
-"https://docs.docker.com/":
+"/":
   - /go/run-cloud-eap
```
dvdksn, line 704: All other go redirects end with a trailing slash. I don't think it makes a difference, but for consistency's sake, I guess we could have one here as well.

```suggestion
  - /go/run-cloud-eap/
```

dvdksn (dismissed):
Approved with a quibble.

dvdksn (approved):

PR #20294, opened 2024-06-25
Update what-is-a-container.md

mandar-rane:
Hyperlink in **Access the Frontend** section

## Description

Changed the hyperlink from https to http


offensive-vk (changes requested):
No one's gonna actually click on that link.

PR #20311, opened 2024-06-27
ENGDOCS-2141

aevesdocker:
## Description

Adds some examples as per Learning path discussion and adjusts some header levels to acommodate  the examples section

## Reviews

- [x] Technical review
- [ ] Editorial review
- [x] Product review

### DIFF
--- a/content/compose/compose-file/05-services.md
+++ b/content/compose/compose-file/05-services.md
@@ -42,6 +42,10 @@ services:
 
 ### Advanced example 
 
+In the following example, the `proxy` service uses the Nginx image, mounts a local Nginx configuration file into the container, exposes port `80` and depends on the `backend` service. 
+
+The `backend` service builds an image from the Dockerfile located in the `backend` directory that is set to build at stage `builder`.
+
 ```yaml
 services:
   proxy:
@@ -61,10 +65,6 @@ services:
       context: backend
       target: builder
 ```
-
-The `proxy` service uses the Nginx image, mounts a local Nginx configuration file into the container, exposes port `80` and depends on the `backend` service. 
-
-The `backend` service builds an image from the Dockerfile located in the `backend` directory that is set to build at stage `builder`.
 
 For more example Compose files, explore the [Awesome Compose samples](https://github.com/docker/awesome-compose).
 


content/compose/compose-file/05-services.md
```diff
@@ -20,7 +20,58 @@ available resources. Deploy support is an optional aspect of the Compose Specifi
...
+  web:
+    image: nginx:latest
+    ports:
+      - "8080:80"
+
+  db:
+    image: postgres:13
+    environment:
+      POSTGRES_USER: example
+      POSTGRES_PASSWORD: example
```
glours, line 38: Not sure we should show this kind of setup, users will reproduce it and it's definitely not a good practice, we should use a secret for this or at least setup the password value via an env variable. 
for example
```suggestion
      POSTGRES_PASSWORD: ${POSTGRES_USER_PASSWORD}
```
Or just remove the `POSTGRES_PASSWORD` definition in this example

glours (approved):
Just update the first sample to avoid hard coded password and it's good to go

content/compose/compose-file/05-services.md
```diff
@@ -20,7 +20,57 @@ available resources. Deploy support is an optional aspect of the Compose Specifi
...
+      - "8080:80"
+
+  db:
+    image: postgres:13
+    environment:
+      POSTGRES_USER: example
+      POSTGRES_DB: exampledb
+```
+
+This minimal example demonstrates how to define two simple services, set their images, map ports, and configure basic environment variables using Docker Compose.
```
dvdksn, line 41: I suggest that we lead with the descriptions instead

content/compose/compose-file/05-services.md
```diff
@@ -20,7 +20,57 @@ available resources. Deploy support is an optional aspect of the Compose Specifi
...
+
+  backend:
+    build:
+      context: backend
+      target: builder
+```
+
+The `proxy` service uses the Nginx image, mounts a local Nginx configuration file into the container, exposes port `80` and depends on the `backend` service. 
+
+The `backend` service builds an image from the Dockerfile located in the `backend` directory that is set to build at stage `builder`.
```
dvdksn, line 67: The same here, I would put the description before, to introduce the example.

dvdksn (approved):

PR #20316, opened 2024-06-27
Fix deprecated types.ImagePullOptions usage and replace with image.PullOptions

sarrocks1:
## Description

While going through golang docker sdk docs, I found that as per https://github.com/moby/moby/blob/v26.0.0/api/types/types_deprecated.go#L7-L35 , the types.ImagePullOptions has been deprecated and on usage throws `undefined: types.ImagePullOptions` error as below:

Screenshot 2024-06-28 at 12 34 29 AM

Fixed the issue by using images.PullOptions which is the correct replacement as mentioned by the same docs above.

## Reviews

- [ ] Technical review
- [ ] Editorial review
- [ ] Product review

### DIFF
--- a/content/engine/api/sdk/_index.md
+++ b/content/engine/api/sdk/_index.md
@@ -82,6 +82,7 @@ import (
 	"os"
 
 	"github.com/docker/docker/api/types"
+        "github.com/docker/docker/api/types/image"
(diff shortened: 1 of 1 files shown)


thaJeztah (approved):
LGTM, thanks!

dvdksn (approved):

PR #20356, opened 2024-07-04
fix(typo): AS instead of as

xElysioN:
## Description
Use `AS` instead of `as` in the dockerfile for the `Name your build stages` that cause a warning when building: 
```
WARN: FromAsCasing: 'as' and 'FROM' keywords' casing do not match (line 1)
```

## Reviews

- [ ] Technical review
- [ ] Editorial review
- [ ] Product review


dvdksn (commented):
thank you! 🚀

dvdksn (approved):

PR #20364, opened 2024-07-05
fix: update linux links

lorenrh:
## Description

Updates linux links in the installer page

## Reviews

- [ ] Technical review
- [ ] Editorial review
- [ ] Product review


dvdksn:
this change breaks all the download links for earlier versions

PR #20365, opened 2024-07-05
fix: update linux links pointing to latest

lorenrh:
## Description

As we've changed the linux binary name to not depend on the version in the installer name, we can remove the build number, as it will always point to the latest version.

⚠️ This has to be merged after 4.32.0 hast been pushed to 100% in the canary rollout!

## Reviews

- [ ] Technical review
- [ ] Editorial review
- [ ] Product review


maxcleme (approved):

chaizeg (approved):
😄

mat007 (approved):

PR #20385, opened 2024-07-09
Fix FromAsCasing warning

iamdbychkov:
Hello! :) 
## Description

Change:

FROM python:${PYTHON_VERSION}-slim as base

To:

FROM python:${PYTHON_VERSION}-slim AS base

To get rid of warning which shows up on image build:

=> WARN: FromAsCasing: 'as' and 'FROM' keywords' casing do not match (line 3)
 1 warning found (use --debug to expand):
 - FromAsCasing: 'as' and 'FROM' keywords' casing do not match (line 3)

## Reviews

- [ ] Technical review
- [ ] Editorial review
- [ ] Product review


dvdksn (approved):
Thank you! :-)

PR #20407, opened 2024-07-13
Update rhel.md to fix package names (rpms instead of debs)

phillipross:
Update rhel.md to fix package names (rpms instead of debs)

## Description

A bit of content referred to deb packages instead of rpm packages.  I updated the relevant content to specify the rpm urls and filenames.


dvdksn (approved):
Thanks!!

PR #20418, opened 2024-07-16
ENGDOCS-2153

aevesdocker:
## Description

Removes 'Use..' from the TOC

## Reviews

- [ ] Technical review
- [ ] Editorial review
- [ ] Product review

### DIFF
--- a/data/toc.yaml
+++ b/data/toc.yaml
@@ -2111,16 +2111,14 @@ Manuals:
       title: Interpolation
     - path: /compose/environment-variables/best-practices/
       title: Best practices
-  - sectiontitle: Use...
-    section:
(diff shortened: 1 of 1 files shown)


dvdksn:
Given that we don't yet have A/B testing capabilities_\*_, do we want to merge this as-is?

\* I have a PR for adding the VWO script to staging, for trying to figure out the flickering issues. But even with that solved, there are some issues with A/B testing the sidebar navigation because it's an interactive component that requires javascript... and last time we (briefly) tried it, there were issues with expanding and collapsing sections that were experimented with using VWO...

dvdksn (approved):

PR #20449, opened 2024-07-21
Improving Docker build cache docs

WillRy:
Improving documentation, explaining that installing dependencies within the Dockerfile can be overwritten by volumes in development environments

## Description

In this section we saw the importance of layer caching to deal with dependencies when packaging the application within the container, being an ideal scenario for production environments.

But remember, in development environments most of the time the application code is shared with the container through volumes, and the code is not copied into the image through Dockerfile

In these cases, it is not recommended run your package manager (NPM, Composer, Maven, Pip, etc.) inside the image, leave the installation to CMD or ENTRYPOINT. This is important so that the volume that we will normally share when developing does not overwrite the files installed by the package manager in the Dockerfile.

Some beginners, when looking at this documentation, will believe that they should always install dependencies within the Dockerfile, which is valid for production environments and several other scenarios, but in development, volumes can overwrite the installation of dependencies made within the Dockerfile

In the current doc, has this example:
```dockerfile
FROM node
WORKDIR /app
COPY . .          # Copy over all files in the current directory
RUN npm install   # Install dependencies
RUN npm build     # Run build
```

However, if the user creates a volume sharing the code within the /app directory, node_modules will be overwritten by the volume and will lose its dependencies.

I believe it is worth inserting a new section in the documentation, alerting to this difference between the treatment of dependencies in development and production when there is the presence of volumes

Currently, there are other good suggestions regarding the same problem, such as:

[20450](https://github.com/docker/docs/pull/20450) is the link to the similar issue.


Luan1Schons:
Your suggestion is very useful and in agreement I also opened another pull request with the same suggestions and another point about volumes, I believe we can make the documentation more accessible about both scenarios.

[PR#20450](https://github.com/docker/docs/pull/20450)

dvdksn (commented):
I don't think this is the right section for this type of content.

I also think that you're better off splitting the Dockerfile into two stages instead: one stage for building the runtime image, and another stage for the dev environment.

I think that the compose section would be a more natural place for this information. @aevesdocker is there an existing page where this could fit?

PR #20450, opened 2024-07-21
Cache and volumes documentation improvements

Luan1Schons:
As per the following PR "https://github.com/docker/docs/pull/20449" explaining that installing dependencies within the Dockerfile can be rolled back across volumes in development environments, it is interesting that you are also warned about dependency managers.

## Description

For production environments, the current example is perfect. However, in development environments, we could have a new section explaining that using volumes would be sufficient instead of copying the content again with "COPY . .". Additionally, it is not necessary to run "npm run build" just to cache the layer, as this is not advantageous in development.

It is also recommended not to run package managers like npm, Composer, Maven, pip, etc., inside the image. Leave the installation to the CMD or ENTRYPOINT. This is important to ensure that the shared volume in development does not overwrite the files installed by the package manager.

```
FROM node
WORKDIR /app
COPY . .          # Copy over all files in the current directory
RUN pip install -r requeriments.txt   # Install dependencies
RUN python -m flask # RUN Flask
```

## Reviews

- [ ] Technical review
- [ ] Editorial review
- [ ] Product review

### DIFF
--- a/content/build/cache/_index.md
+++ b/content/build/cache/_index.md
@@ -269,6 +269,71 @@ EOF
 (Note the `set -e` command to exit immediately after any command fails, instead
 of continuing.)
 
+## Dealing with dependencies in development
+
+In previous examples we saw the importance of layer caching to deal with dependencies when packaging the application within the container, being an ideal scenario for production environments.
+
+But remember, in development environments most of the time the application code is shared with the container through volumes, and is not copied into the image through Dockerfile
+
+In these cases:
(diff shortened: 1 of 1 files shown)


WillRy:
This observation is important, when I was studying the documentation, I had difficulties because in development environments the volumes overwrite the node_modules generated through Dockerfile.

I wrote a similar issue:
https://github.com/docker/docs/pull/20449

PR #20471, opened 2024-07-24
Update writing-a-dockerfile.md

olu-folarin:
## Reconstruct Part of the Said Sentence

This is to point out, clearly, that the image is not secure as it is currently configured.
I propose this edit because I almost didn't understand what that part meant until I took a step back then contextualised it.

- [x] Editorial review


olu-folarin:
Thanks David.

PR #20480, opened 2024-07-26
Fix typo in command in databases.md

printchard:
## Description

I was going through the documentation of creating databases and noticed a typo. The command shows `-p 3306:3306`, but the description says `-p 3307:3306`. It could be fixed the other way around too.


dvdksn (approved):
Thank you

PR #20493, opened 2024-07-26
build-cloud: replace `Hyrdobuild` with `Build Cloud`

laurazard:
## Description

Noticed a typo (`Hyrdobuild`) and decided to just update it to `Build Cloud`. Let me know if another term is preferred (like **Docker** Build Cloud).

## Reviews

- [ ] Technical review
- [ ] Editorial review
- [ ] Product review


dvdksn (approved):
Ha, thank you! That typo escaped my grep foo.

PR #20495, opened 2024-07-29
add insights

craig-osterhout:
## Description

Added Insights topic for EA. No toc entry and excluded from search.
Added redirects for links from UI.
https://deploy-preview-20495--docsdocker.netlify.app/admin/organization/insights/

## Related issues or tickets
ENGDOCS-2119

## Reviews

- [ ] Technical review
- [ ] Editorial review
- [ ] Product review


content/admin/organization/insights.md
```diff
@@ -0,0 +1,190 @@
...
+| Users opted out of analytics | The number of users that are a member of your organization that have opted out of sending analytics. <br><br>When users opt out of sending analytics, you won't see any of their data in insights. To ensure that your insights include all users, you can use [Settings Management](/desktop/hardened-desktop/settings-management/) to set `analyticsEnabled` for all your users.                                                                                                                                                                                                                                                                                                                                                                                      |
```
dvdksn, line 75: ?

It's kinda hard to tell when we're referring to the feature vs the metrics the feature collects... Have you considered referring to the data points as "metrics" (or some other word: "measure", "indicator", "signal"), and the feature "Insights"?


```suggestion
| Users opted out of analytics | The number of users that are a member of your organization that have opted out of sending analytics. <br><br>When users opt out of sending analytics, you won't see any of their data in Insights. To ensure that your insights include all users, you can use [Settings Management](/desktop/hardened-desktop/settings-management/) to set `analyticsEnabled` for all your users.                                                                                                                                                                                                                                                                                                                                                                                      |
```

craig-osterhout: I had "metric" in a few places before, but I found I was using it inconsistently, so I just nixed it. I'll look over the whole topic again and see if I can replace insights (the concept) with metric instead to avoid the whole is it the insights concept or Insights feature issue.

content/admin/organization/insights.md
```diff
@@ -0,0 +1,190 @@
...
+
+Monitor development efficiency and the time your team invests in builds with
+this chart. It provides a clear view of the build activity, helping you identify
+patterns, optimize build times, and enhance overall development productivity.
+
+The chart contains the following insights.
+
+| Insight                | Description                                                                                                                                                                                        |
+|:-----------------------|:---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| Average build per user | The average number of builds per active user. A build includes any time a user runs a Docker command that builds an image.                                                                         |
```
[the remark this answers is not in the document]
craig-osterhout: Updated to be more specific. It only counts commands run.
Any time a user runs `docker build`, `compose build`, `compose up --build`, or any command beginning with `docker buildx`.

content/admin/organization/insights.md
```diff
@@ -0,0 +1,188 @@
...
+
+Monitor development efficiency and the time your team invests in builds with
+this chart. It provides a clear view of the build activity, helping you identify
+patterns, optimize build times, and enhance overall development productivity.
+
+The chart contains the following data.
+
+| Data                   | Description                                                                                                                                                                                        |
+|:-----------------------|:---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| Average build per user | The average number of builds per active user. A build includes any time a user runs `docker build`, `compose build`, `compose up --build`, or any command beginning with `docker buildx`.          |
```
dvdksn, line 88: This isn't right, flagging all `buildx` invocations as a build . `buildx` has many other commands for building. We should narrow this data point down to `buildx build` and `buildx bake`, and ignore any other `buildx` commands.

If we can't change the data, then we should at least probably call this discrepancy out in this doc.

dvdksn (dismissed):

content/admin/organization/insights.md
craig-osterhout: Thank you. Got a more exact definition and updated it.

dvdksn (approved):
LGTM

PR #20503, opened 2024-07-31
Define units for max-buffer-size

jimschubert:
## Description

This documents the units for `max-buffer-size`. Previously, the allowed sizes were not documented to make it clear this is the binary unit of measure, where 1 KiB = 1024 Bytes. This also clarifies that the `i` can be omitted from the binary unit specification, which surprised me as a user that this didn't change the size calculation.

The single-character lowercased examples also were not previously clear in the docs which units they referred to, so hopefully this adds clarity.

## Reviews

- [ ] Technical review
- [ ] Editorial review
- [ ] Product review


thaJeztah:
- alternative PR in https://github.com/docker/docs/pull/20562

PR #20510, opened 2024-07-31
Add Underscores (`_`) and slashes (`/`) to valid label content

maxmynter:
## Description

We're building a tool that wraps container creation and found that Docker can also work with underscores as well as slashes in labels.

## Reviews

- [ ] Technical review
- [ ] Editorial review
- [ ] Product review


dvdksn (approved):
Thank you!

PR #20528, opened 2024-08-03
Update Docker CLI container guide

### LINKED ISSUE #20527: Hard to follow gaps while setting up Docker using CLI as first time user
### Is this a docs issue?

### Type of issue

Information is incorrect

### Description

I found minor errors in the documentaiton which can be challenging for someone who is using docker for the very first time. My girlfriend was trying to set things up and she was confused why certain errors were raised even though she had followed the steps as needed. 

After her trying to set things up, I took a read as fresh user and here are some mistakes I found out.

- In the "Try Out" section for the "Using the CLI" tab. There is no instructions to pull the image. The very first command directly says to run the welcome image. Which offcourse needs to be pulled and as far as I can understand, The user coming to this page would be a first time user. So, we cannot expect them to pull the image all by themselves and this needs to be asked for. 
- The link in "Access the frontend" section is `http://localhost:8080` in text. But when someone clicks on that link then it is redirecting to `https://localhost:8080` due to which it is making it confusing since in chrome browser it shows only localhost:8080 (protocol is hidden untill someone clicks on the url box) so finding the root cause of image not running on 8080 is tough and time taking without much attention. So this should be fixed too.

I am happy to raise a PR to fix the issues above mentioned in the documentation page. I believe this would really help new user have a smooth experience using Docker and will make the community even more supportive.

### Location

https://docs.docker.com/guides/docker-concepts/the-basics/what-is-a-container/

### Suggestion

_No response_

Harshil-Jani:
## Description

- Added instructions to pull the 'welcome-to-docker' image using `docker pull` command.
- Clarified steps by separating image pulling and container starting into distinct steps.
- Fixed localhost link from https to http.

## Related issues or tickets

Fixes #20527

## Reviews

- [x] Technical review
- [x] Editorial review
- [ ] Product review

### DIFF
--- a/content/guides/docker-concepts/the-basics/what-is-a-container.md
+++ b/content/guides/docker-concepts/the-basics/what-is-a-container.md
@@ -115,7 +115,13 @@ The `docker/welcome-to-docker` container continues to run until you stop it.
 
 Follow the instructions to run a container using the CLI:
 
-1. Open your CLI terminal and start a container by using the [`docker run`](/reference/cli/docker/container/run/) command:
+1. Open your CLI terminal and pull the welcome-to-docker image from Dockerhub by using the [`docker pull`](/reference/cli/docker/compose/pull/) command:
+
+    ```console
+    $ docker pull docker/welcome-to-docker
+    ```
+
+2. Start a container by using the [`docker run`](/reference/cli/docker/container/run/) command:
 
     ```console
     $ docker run -d -p 8080:80 docker/welcome-to-docker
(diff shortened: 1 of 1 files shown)


content/guides/docker-concepts/the-basics/what-is-a-container.md
```diff
@@ -115,7 +115,13 @@ The `docker/welcome-to-docker` container continues to run until you stop it.
...
 Follow the instructions to run a container using the CLI:
 
-1. Open your CLI terminal and start a container by using the [`docker run`](/reference/cli/docker/container/run/) command:
+1. Open your CLI terminal and pull the welcome-to-docker image from Dockerhub by using the [`docker pull`](/reference/cli/docker/compose/pull/) command:
+
+    ```console
+    $ docker pull docker/welcome-to-docker
+    ```
+
+2. Start a container by using the [`docker run`](/reference/cli/docker/container/run/) command:
```
dvdksn, line 124: `docker run` will automatically pull the image if it doesn't already exist in your local image store. This step is not necessary so I suggest we revert this bit.

```suggestion
1. Open your CLI terminal and start a container by using the [`docker run`](/reference/cli/docker/container/run/) command:
```

Harshil-Jani: Thanks. I was not aware about it. 

How about we add this statement in that point itself just to clarify that docker pulls the image if it is not present ??

dvdksn (approved):
thank you!

content/guides/docker-concepts/the-basics/what-is-a-container.md
dvdksn: Yeah, maybe. I think we can leave it out for now but if it seems to cause more confusion I'll revisit.

PR #20547, opened 2024-08-06
Update multi-stage-builds.md

Lakshan-Ram:
added underscores to spring_boot_docker since the spring inializer produces file paths with underscores


offensive-vk (approved):
wow dude

PR #20558, opened 2024-08-07
Update writing-a-dockerfile.md

vedantnaik19:
## Description

Update the WORKDIR in the Dockerfile to match the directory structure in the zip file.


dvdksn (changes requested):
`WORKDIR` refers to the path inside the container as defined in the Dockerfile, which is `/usr/local/app`

vedantnaik19:
> `WORKDIR` refers to the path inside the container as defined in the Dockerfile, which is `/usr/local/app`

Oh I was referring to the `WORKDIR` defined in the Dockerfile of the zip file.
Also, the subsequent steps has WORKDIR as `/app`

dvdksn (approved):
Oh I see the workdir in the zip file is different from what's in the repository. That's interesting.

PR #20594, opened 2024-08-13
ENGDOCS-2180

aevesdocker:
## Description

Moves Hardened Docker Desktop content into Security section

## Reviews

- [ ] Technical review
- [ ] Editorial review
- [ ] Product review


craig-osterhout (approved):
LGTM.
Noting that with this change, ECI now lives under admin security even though it also applies to dev security.

PR #20621, opened 2024-08-15
Fix the error describes of `COPY` chapter in reference.md

Navyum:
## Description

This chapter is describing the usage of `COPY`. While these places is using `ADD` instead，Which could make users confused。

## Reviews

- [  ] Technical review
- [ x ] Editorial review
- [  ] Product review


dvdksn:
Hi, thanks for catching that! This file is vendored from moby/buildkit. I carried your fix in https://github.com/moby/buildkit/pull/5259

Will be published with the next BuildKit release.

PR #20628, opened 2024-08-16
ENGDOCS-2182

aevesdocker:
## Description

Removes EA labels as these features are moving into GA

## Reviews

- [ ] Technical review
- [ ] Editorial review
- [ ] Product review

### DIFF
--- a/content/security/for-admins/enforce-sign-in/methods.md
+++ b/content/security/for-admins/enforce-sign-in/methods.md
@@ -27,7 +27,7 @@ This page outlines the different ways you can enforce sign-in for Docker Desktop
 
 In some cases, a system reboot may be necessary for enforcement to take effect.
 
-> [!Note]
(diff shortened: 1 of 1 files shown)


content/security/for-admins/enforce-sign-in/methods.md
```diff
@@ -28,7 +27,7 @@ It's available with Docker Desktop version 4.32 and later.
 
 In some cases, a system reboot may be necessary for enforcement to take effect.
 
->**Note**
```
dvdksn, line 31: It looks like the alert type must be in all-caps

image

I see that this was fixed in Hugo v0.132.2: https://github.com/gohugoio/hugo/issues/12767

For now, let's maybe keep the uppercase, since the "spec" is somewhat ambiguous.

dvdksn (approved):

PR #20637, opened 2024-08-19
Add "rootless mode" to exceptions of "Privileged user requirement" in `docker login`

### LINKED ISSUE #20415: Add "rootless mode" to exceptions of "Privileged user requirement" in `docker login` CLI command
### Is this a docs issue?

### Type of issue

Information is incorrect

### Description

The documentation for `docker login` specifies that `sudo` or `root` is required, however this does not occur when Rootless mode is set up on the Docker host. https://github.com/docker/docs/blob/b904b7f074780c0debd5c40265780509bc47bdc1/data/engine-cli/docker_login.yaml#L74-L80

### Location

https://docs.docker.com/reference/cli/docker/login/

### Suggestion

Perhaps it would be better to add one line to the list of exceptions regarding this behavior.

Javiery3889:
Fixes #20415

### DIFF
--- a/data/engine-cli/docker_login.yaml
+++ b/data/engine-cli/docker_login.yaml
@@ -77,6 +77,7 @@ examples: |-
 
     - Connecting to a remote daemon, such as a `docker-machine` provisioned `docker engine`.
     - The user is added to the `docker` group. This will impact the security of your system; the `docker` group is `root` equivalent.  See [Docker Daemon Attack Surface](/engine/security/#docker-daemon-attack-surface) for details.
+    - The Docker daemon is running as a non-root user, i.e. [rootless mode](https://docs.docker.com/engine/security/rootless/).
 
     You can log in to any public or private repository for which you have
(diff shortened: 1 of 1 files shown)


dvdksn (changes requested):
Thank you so much for taking the time to raise this PR. Your suggestion looks good, but sadly this is a generated file and we need to make the change in the source or else it will be overwritten the next time we re-generate the CLI documentation. 

The source for this content is here: https://github.com/docker/cli/blob/master/docs/reference/commandline/login.md

May I ask that you raise a pull request on the docker/cli repository instead? And then we'll vendor those changes to this repo once they're merged.

Javiery3889:
Hi @dvdksn, understood and thank you for pointing out the source file to edit. I have created a PR in the `docker/cli` repo as well.

Javiery3889:
Closing this PR with respect to https://github.com/docker/cli/pull/5359

PR #20638, opened 2024-08-19
Update key-concepts.md

rcarmen-btc:
fixed a typo

## Reviews

- [ ] Technical review
- [ ] Editorial review
- [ ] Product review

### DIFF
--- a/content/engine/swarm/key-concepts.md
+++ b/content/engine/swarm/key-concepts.md
@@ -53,7 +53,7 @@ manager node*. The manager node dispatches units of work called
 [tasks](#services-and-tasks) to worker nodes.
 
 Manager nodes also perform the orchestration and cluster management functions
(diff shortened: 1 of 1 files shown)


content/engine/swarm/key-concepts.md
```diff
@@ -53,7 +53,7 @@ manager node*. The manager node dispatches units of work called
 [tasks](#services-and-tasks) to worker nodes.
 
 Manager nodes also perform the orchestration and cluster management functions
-required to maintain the desired state of the swarm. Manager nodes elect a
```
dvdksn, line 56: I'm guessing the original word might have been... selected... on purpose. I don't really think that manager nodes hold elections though. If they do, they're probably rigged anyway!

In the name of clarity, I agree that we should change this to a simpler word.

dvdksn (approved):

PR #20641, opened 2024-08-19
ENGDOCS-2184

aevesdocker:
## Description

Move `compose-file` folder from `compose` to `reference`

## Reviews

- [ ] Technical review
- [ ] Editorial review
- [ ] Product review


content/reference/compose-file/04-version-and-name.md
dvdksn, line 1: Now that we're anyway moving these files and adding redirects, I think we should also take the opportunity and remove the `04-`, `-05` (etc) prefix from the filenames.

dvdksn (approved):

PR #20708, opened 2024-08-26
iam: remove group mapping with sso using azure id

craig-osterhout:
## Description

Support for group mapping with SSO using the Azure ID auth method was removed. Only SAML auth method supported.

- Updated **Use group mapping with SSO** to note SAML is supported and Azure ID is not.
- Removed the `Search for Directory.Read.All, and select this option.` step from Azure ID config as it was used for group mapping which is no longer supported.
- Added FAQ in case users ask why it's not supported.

## Related issues or tickets

ENGDOCS-2194

## Reviews

- [ ] Technical review
- [ ] Editorial review


dvdksn (approved):

ivan-californias (approved):
LGTM

PR #20716, opened 2024-08-27
engine: 27.2.0 release notes

vvoland:
## Description

- Add Docker Engine v27.2 release notes
- Consolidate v27.x release notes to a single page
- Vendor moby and cli upstreams
- udpate config


content/engine/release-notes/27.1.md
```diff
@@ -4,13 +4,6 @@ description: Learn about the new features, bug fixes, and breaking changes for D
 keywords: docker, docker engine, ce, whats new, release notes
 toc_min: 1
 toc_max: 2
-tags:
```
thaJeztah, line 7: Considering if we should've used a "v27" file so that we could put all minor versions in there 🤔 maybe that ship sailed, or perhaps we could still combine them; @dvdksn WDYT?

content/engine/release-notes/27.2.md
```diff
@@ -0,0 +1,62 @@
...
+tags:
+  - Release notes
+aliases:
+- /engine/release-notes/
+- /engine/release-notes/latest/
+- /release-notes/docker-ce/
+- /release-notes/docker-engine/
+---
+
+This page describes the latest changes, additions, known issues, and fixes for Docker Engine version 27.1.
```
thaJeztah, line 16: s/27.1/27.2/

but also see my other comment

content/engine/release-notes/27.2.md
```diff
@@ -0,0 +1,62 @@
...
+- Fix an issue that prevented network creation with a `--ip-range` ending on a 64-bit boundary. [moby/moby#48326](https://github.com/moby/moby/pull/48326)
+- CLI: IPv6 addresses shown by `docker ps` in port bindings are now bracketed. [docker/cli#5365](https://github.com/docker/cli/pull/5365)
+- containerd image store: Fix early error exit from `docker load` in cases where unpacking the image would fail. [moby/moby#48376](https://github.com/moby/moby/pull/48376)
+- containerd image store: Fix the previous image not being persisted as dangling after `docker pull`. [moby/moby#48380](https://github.com/moby/moby/pull/48380)
+
+### Packaging updates
+
+- Update BuildKit to [v0.15.2](https://github.com/moby/buildkit/releases/tag/v0.15.2). [moby/moby#48341](https://github.com/moby/moby/pull/48341)
+- Update Compose to [v2.29.2](https://github.com/docker/compose/releases/tag/v2.29.2). [docker/docker-ce-packaging#1050](https://github.com/docker/docker-ce-packaging/pull/1050)
+- The canonical source for the dockerd(8) man page has been moved back to the same source tree as dockerd itself. [moby/moby#48378](https://github.com/moby/moby/pull/48378)
```
thaJeztah, line 60: Maybe this one is only interesting for the github release notes, not for the docs (we also didn't update our packaging yet, so it's only the source code changes for now)

content/engine/release-notes/27.1.md
dvdksn: yeah should be fine to consolidate all v27 to the same page. We can alias the paths with minor versions.

dvdksn: I updated this PR with this consolidation

dvdksn:
Let me also vendor the upstreams

thaJeztah:
That vale check is really noisy; wondering if we need some exclusion rules (at least it's complaining about any link pointing to `docker/xxxx` on GitHub). Is this warning expected?

```
/usr/bin/pip install docutils
Defaulting to user installation because normal site-packages is not writeable
Collecting docutils
  Downloading docutils-0.21.2-py3-none-any.whl.metadata (2.8 kB)
Downloading docutils-0.21.2-py3-none-any.whl (587 kB)
   ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 587.4/587.4 kB 12.2 MB/s eta 0:00:00
Installing collected packages: docutils
Successfully installed docutils-0.21.2
/usr/bin/gem install asciidoctor --user-install
WARNING:  You don't have /home/runner/.local/share/gem/ruby/3.2.0/bin in your PATH,
	  gem executables will not run.
```

thaJeztah:
This one looks like it could be easy to address;

```
[vale] reported by reviewdog 🐶
[Docker.Acronyms] 'GHSA' has no definition.
```

For the `docker` vs `Docker` case; wondering if it can be configured to ignore `docker/` ("docker" followed by a `/`, or perhaps followed by "non-whitespace") to reduce false positives.

thaJeztah:
I think I found the config-file; let me try opening a PR to fix those 🤞

dvdksn:
> Is this warning expected?

I haven't seen it before but it looks like it's related to asciidoc, which we don't use, so I wouldn't worry about it. 

> For the docker vs Docker case; wondering if it can be configured to ignore docker/ ("docker" followed by a /, or perhaps followed by "non-whitespace") to reduce false positives.

Yeah I think we should ignore these. It used to be easier with earlier versions of vale to ignore linktext, but I can also take a look again and remind myself what the options are nowadays

dvdksn:
LOL, one noise replaced by another

`[Docker.RecommendedWords] Consider using 'sign in' instead of 'login'`

content/engine/release-notes/27.2.md
```diff
@@ -0,0 +1,88 @@
...
+#### Device code login
+
+This release adds support for using device code login when authenticating to
+Docker Hub.
+
+You can still use the old method of logging in with a username and password or
+access token, but device code login is more secure and doesn't require you to
+enter your password in the CLI.
+
+![device-code-login](https://i.imgur.com/GRGjSjV.gif "Device code login")
```
thaJeztah: Silly question; do we already use `imgur.com` for images? Should this be either integrated in the docs, or do we have some other location to store these?

I know some of those services don't keep images forever, and/or perform additional tracking

content/engine/release-notes/27.2.md
```diff
@@ -0,0 +1,88 @@
...
+
+{{< experimental >}}
+This is experimental and may change at any time without any backward compatibility.
+{{< /experimental >}}
+
+With the containerd image store enabled, the `docker image ls` command (or
+`docker images` shorthand) now supports a `--tree` flag that now shows
+if an image is a multi-platform image.
+
+<script src="https://asciinema.org/a/PyE57rUPdNh7LvOMhULtE0Ek2.js" id="asciicast-PyE57rUPdNh7LvOMhULtE0Ek2" async="true"></script>
```
thaJeztah: I guess somewhat similar question here; looks like it can be self-hosted, but not 100% sure what amount of server-side is needed to show these; https://docs.asciinema.org/manual/server/self-hosting/

content/engine/release-notes/27.2.md
dvdksn: I used it only because it's a gif, and I didn't want to commit it to the repository. This being a release note. As far as I know they don't remove images from imgur. But this being a release note, and the gif illustrative, it wouldn't be the end of the world if it went away at some point. I would like to have a better place to host these without having to commit them to the repository. I'll check with the web team if they've got a better solution.

dvdksn (approved):

PR #20726, opened 2024-08-28
vale: add some acronyms and relax validation

thaJeztah:
### vale: add GHSA to acronyms

GHSA is for GitHub Security Advisories;
https://docs.github.com/en/code-security/security-advisories/working-with-global-security-advisories-from-the-github-advisory-database/about-the-github-advisory-database

### vale: add DCT, DSOS and DVP to acronyms

- DSOS: Docker-Sponsored Open Source (DSOS) program
- DVP: Docker Verified Publisher (DVP)
- DCT: Docker Content-Trust

### vale: add TTY to acronyms

    [vale] reported by reviewdog 🐶
    [Docker.Acronyms] 'TTY' has no definition.

### vale: allow 'docker/' in link-captions

Try to prevent false-positives on links to GitHub issues, e.g.;

    [docker/docker-ce-packaging#1050](https://github.com/docker/docker-ce-packaging/pull/1050)
    Check failure on line 93 in content/engine/release-notes/27.md

    [vale] reported by reviewdog 🐶
    [Vale.Terms] Use 'Docker' instead of 'docker'.

## Related issues or tickets

- https://github.com/docker/docs/pull/20716#issuecomment-2315098383

## Reviews

- [x] Technical review
- [ ] Editorial review
- [ ] Product review


.github/vale/config/vocabularies/Docker/accept.txt
```diff
@@ -35,6 +35,7 @@ Ddosify
 Debootstrap
 Dev Environments?
 Django
+docker/
```
thaJeztah, line 38: Hope this one works; looks like this file accepts some RegEx? Not sure how it matches things (so if it would need something like `[^]]+` after it?

thaJeztah: Do we need some exception for `docker-XXXX` as well? (thinking `docker-compose`)?

dvdksn: I think the only valid use for `docker-compose` would be in a code span, when referring to the legacy command, so in that case it wouldn't be flagged by vale anyway.

dvdksn: I don't think this works, at least it doesn't seem to in my editor

```suggestion
docker/
```

thaJeztah: LOL what's your suggested change above? I don't see what it ... changed 🥹

.github/vale/Docker/Capitalization.yml
```diff
@@ -0,0 +1,10 @@
+extends: existence
+message: "Please capitalize Docker."
+level: error
+ignorecase: false
+action:
+  name: replace
+  params:
+    - Docker
+tokens:
+  - 'docker(?!/cli)'
```
thaJeztah, line 10: I guess we may need `buildx`, `compose`, `<other repos>` as well; should those be added as separate items in this list, or do we want to try capturing them in a single RegEx?

dvdksn: ya you're right, we should. Maybe this?

```suggestion
  - 'docker(?!/[a-z-]+)'
```

thaJeztah: Yup; something like that seems good (but make sure to test indeed) 🙈 

I updated with your suggestion

dvdksn (approved):

PR #20739, opened 2024-08-30
release notes: add known issue for CLI login address normalization

laurazard:
## Description

Added known issue for CLI login address normalization issue.

https://github.com/docker/cli/issues/5382
More info in: https://github.com/docker/cli/pull/5383

## Reviews

- [ ] Technical review
- [ ] Editorial review
- [ ] Product review

### DIFF
--- a/content/desktop/release-notes.md
+++ b/content/desktop/release-notes.md
@@ -71,6 +71,12 @@ Take a look at the [Docker Public Roadmap](https://github.com/docker/roadmap/pro
 - Fixed a bug that caused WSL distro to terminate abruptly. Fixes [for-win/14230](https://github.com/docker/for-win/issues/14230).
 - Fixed an issue that caused WSL to update in each startup. Fixes [for-win/13868](https://github.com/docker/for-win/issues/13868), [for-win/13806](https://github.com/docker/for-win/issues/13806).
 
+### Known Issues
+
+- There is a known issue when authenticating against a registry in the Docker CLI (`docker login [registry address]`) where, if the provided registry address includes a repository/image name (such as `docker login index.docker.io/docker/welcome-to-docker`), the repository part (`docker/welcome-to-docker`) is not normalized and results in credentials being stored incorrectly, which causes subsequent pulls from the registry (`docker pull index.docker.io/docker/welcome-to-docker`) to not be authenticated. To prevent this, don't include any extraneous suffix in the registry address when running `docker login`.
+  > [!NOTE]
+  > Using `docker login` with an address that includes URL path segments is not a documented use case and is considered unsupported. The recommended usage is to specify only a registry hostname, and optionally a port, as the address for `docker login`.
+
 ## 4.33.1
 
 {{< release-date date="2024-07-31" >}}
--- a/content/engine/release-notes/27.md
+++ b/content/engine/release-notes/27.md
@@ -93,6 +93,12 @@ if an image is a multi-platform image.
 - Update Compose to [v2.29.2](https://github.com/docker/compose/releases/tag/v2.29.2). [docker/docker-ce-packaging#1050](https://github.com/docker/docker-ce-packaging/pull/1050)
 - Update containerd to [v1.7.21](https://github.com/containerd/containerd/releases/tag/v1.7.21). [moby/moby#48383](https://github.com/moby/moby/pull/48383), [docker/containerd-packaging#389](https://github.com/docker/containerd-packaging/pull/389)
 
+### Known Issues
+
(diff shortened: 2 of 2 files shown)


content/desktop/release-notes.md
```diff
@@ -71,6 +71,10 @@ Take a look at the [Docker Public Roadmap](https://github.com/docker/roadmap/pro
 - Fixed a bug that caused WSL distro to terminate abruptly. Fixes [for-win/14230](https://github.com/docker/for-win/issues/14230).
 - Fixed an issue that caused WSL to update in each startup. Fixes [for-win/13868](https://github.com/docker/for-win/issues/13868), [for-win/13806](https://github.com/docker/for-win/issues/13806).
 
+### Known Issues
+
+- There is a known issue when authenticating against a registry in the Docker CLI (`docker login [registry address]`) where, if the provided registry address includes a repository/image name (such as `docker login index.docker.io/docker/welcome-to-docker`), the repository part (`docker/welcome-to-docker`) is not normalized and results in credentials being stored incorrectly, which causes subsequent pulls from the registry (`docker pull index.docker.io/docker/welcome-to-docker`) to not be authenticated. To prevent this, don't include any extraneous suffix in the registry address when running `docker login`.
+
```
thaJeztah, line 77: Do we need to call out that this was undocumented behavior, and the expected argument is the name of the registry( hostname (with optional port) `hostname[:port]`), so that users are recommended to use that format, 
or would that be too much info?

cc @dvdksn (in case you have too suggestions for wording)



We should probably also have a note in the engine release notes

laurazard: I considered that, then decided against it since it didn't feel super relevant to this section (listing known issues). I'm happy to change it though, I don't feel super strongly about it.

Adding it to the engine release notes makes sense – I'd also considered a little (temporary) callout in the CLI docs for docker login, or maybe a not-temporary one explaining that the expected input there is not to include anything but the registry hostname/port. Actually, that makes even more since now since we know some users are doing this (Gitlab).

laurazard: Let me add a blurb so we can discuss wording.

laurazard: Did we add release notes for 27.2?

dvdksn: I just merged them

content/engine/release-notes/27.md
```diff
@@ -93,6 +93,10 @@ if an image is a multi-platform image.
 - Update Compose to [v2.29.2](https://github.com/docker/compose/releases/tag/v2.29.2). [docker/docker-ce-packaging#1050](https://github.com/docker/docker-ce-packaging/pull/1050)
 - Update containerd to [v1.7.21](https://github.com/containerd/containerd/releases/tag/v1.7.21). [moby/moby#48383](https://github.com/moby/moby/pull/48383), [docker/containerd-packaging#389](https://github.com/docker/containerd-packaging/pull/389)
 
+### Known Issues
+
+- There is a known issue when authenticating against a registry in the Docker CLI (`docker login [registry address]`) where, if the provided registry address includes a repository/image name (such as `docker login index.docker.io/docker/welcome-to-docker`), the repository part (`docker/welcome-to-docker`) is not normalized and results in credentials being stored incorrectly, which causes subsequent pulls from the registry (`docker pull index.docker.io/docker/welcome-to-docker`) to not be authenticated. To prevent this, don't include any extraneous suffix in the registry address when running `docker login`.
```
laurazard, line 98: Copied the text from the Desktop release notes here, but we should adjust it. WDYT about mentioning the fact that this was undocumented behavior? @thaJeztah @dvdksn

dvdksn, line 98: yeah, what about this:

```suggestion
- There is a known issue when authenticating against a registry in the Docker CLI (`docker login [registry address]`) where, if the provided registry address includes a repository/image name (such as `docker login index.docker.io/docker/welcome-to-docker`), the repository part (`docker/welcome-to-docker`) is not normalized and results in credentials being stored incorrectly, which causes subsequent pulls from the registry (`docker pull index.docker.io/docker/welcome-to-docker`) to not be authenticated. To prevent this, don't include any extraneous suffix in the registry address when running `docker login`.

  > [!NOTE]
  > Using `docker login` with an address that includes URL path segments is not a documented use case and is considered unsupported. The recommended usage is to specify only the registry hostname, and optionally a port, as the address for `docker login`.
```

laurazard: Loosk good to me! @thaJeztah WDYT?

thaJeztah: SGTM 👍

laurazard:
Included the suggested changes, PTAL @dvdksn @thaJeztah

laurazard:
(let's try to get this merged soon so people running into this can find it).

dvdksn (approved):

PR #20748, opened 2024-08-30
using gha cache with docker-container

sozercan:
## Description

documents how to use gha cache with `docker-container` driver

## Reviews

- [ ] Technical review
- [ ] Editorial review
- [ ] Product review

### DIFF
--- a/content/build/cache/backends/gha.md
+++ b/content/build/cache/backends/gha.md
@@ -57,6 +57,14 @@ exposed. Consider using the
 [`crazy-max/ghaction-github-runtime`](https://github.com/crazy-max/ghaction-github-runtime),
 GitHub Action as a helper for exposing the variables.
 
+If you are using the `docker-container` driver, `ACTIONS_RUNTIME_TOKEN` and `ACTIONS_RUNTIME_URL` environment variables must be passed into the container. 
+
+```console
+docker buildx create --driver=docker-container --use --bootstrap \
+  --driver-opt=env.ACTIONS_RUNTIME_TOKEN=$ACTIONS_RUNTIME_TOKEN \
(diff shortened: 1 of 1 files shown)


cpuguy83 (approved):

content/build/cache/backends/gha.md
```diff
@@ -57,6 +57,14 @@ exposed. Consider using the
 [`crazy-max/ghaction-github-runtime`](https://github.com/crazy-max/ghaction-github-runtime),
 GitHub Action as a helper for exposing the variables.
 
+If you are using the `docker-container` driver, `ACTIONS_RUNTIME_TOKEN` and `ACTIONS_RUNTIME_URL` environment variables must be passed into the container.
```
crazy-max, line 60: This is not needed, Buildx already sets `token` and `url` attributes for `gha` exporter during build: https://github.com/docker/buildx/blob/96eb69aea4e03baf516f29ec36aa93a9c284f085/util/buildflags/cache.go#L69-L84

cpuguy83: Oh cool. I wonder why we were having an issue getting this to work. Maybe an older buildx not using the controller api?

crazy-max: Do you run `docker buildx` in a `runs` step? If that's the case then `ACTIONS_RUNTIME_TOKEN` and `ACTIONS_RUNTIME_URL` envs are not exposed: https://docs.docker.com/build/cache/backends/gha/#authentication

You would need smth like https://github.com/crazy-max/ghaction-github-runtime?tab=readme-ov-file#usage

cpuguy83: Yeah we're definitely using that. Will have to take a closer look.

PR #20824, opened 2024-09-10
ENGDOCS-2209

aevesdocker:
## Description

Arch install improvements. Addresses issues raised in https://docker.slack.com/archives/C039ZM6GH4Z/p1724677490900139 and https://github.com/docker/docs/issues/20816

Also makes the info that this is experimental and the install instructions aren't verified clearer

## Reviews

- [ ] Technical review
- [ ] Editorial review
- [ ] Product review

### DIFF
--- a/content/manuals/desktop/install/archlinux.md
+++ b/content/manuals/desktop/install/archlinux.md
@@ -16,34 +16,27 @@ aliases:
 > employees OR more than $10 million USD in annual revenue) requires a [paid
 > subscription](https://www.docker.com/pricing/).
 
-This page contains information on how to install, launch and upgrade Docker Desktop on an Arch-based distribution. Docker has not tested or verified the installation.
+This page contains information on how to install, launch and upgrade Docker Desktop on an Arch-based distribution. 
+
+> [!IMPORTANT]
+>
+> This is an experimental installation package. Docker has not tested or verified the installation.
 
 ## Prerequisites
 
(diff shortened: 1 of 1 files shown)


content/manuals/desktop/install/archlinux.md
```diff
@@ -16,35 +16,28 @@ aliases:
...
-$ sudo pacman -S gnome-terminal
-```
-
 ## Install Docker Desktop
 
-1. Install client binaries. Docker does not have an Arch package repository. Binaries not included in the package must be installed manually before installing Docker Desktop.
-
-2. [Install Docker client binary on Linux](/manuals/engine/install/binaries.md#install-daemon-and-client-binaries-on-linux). On Arch-based distributions, users must install the Docker client binary.
-   Static binaries for the Docker client are available for Linux (as `docker`).
+1. [Install the Docker client binary on Linux](/manuals/engine/install/binaries.md#install-daemon-and-client-binaries-on-linux). Static binaries for the Docker client are available for Linux as `docker`.
```
dvdksn, line 31: I wonder if we should inline this.

```console
$ wget https://download.docker.com/linux/static/stable/x86_64/docker-27.2.1.tgz -qO- | tar xvfz - docker/docker --strip-components=1
$ mv ./docker ~/.local/bin
```

the version could be dynamic using the site param in config: `{{% param "docker_ce_version" %}}`

cc @cybercyst - do you have any good suggestions for this prodecure?

PR #20834, opened 2024-09-10
Update ubuntu.md

MaVeron30:
Added a solution to the error

### DIFF
--- a/content/manuals/engine/install/ubuntu.md
+++ b/content/manuals/engine/install/ubuntu.md
@@ -268,6 +268,38 @@ To upgrade Docker Engine, download the newer package files and repeat the
 
 You have to delete any edited configuration files manually.
 
+## Troubleshooting GPG error "NO_PUBKEY 7EA0G9CBF1373FED8"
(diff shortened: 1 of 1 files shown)


dvdksn (commented):
I'm hesitant to add this, you shouldn't have to do this if you're on a supported distro version and you haven't configured your system in some exotic way that breaks things (e.g., default umask). Can you reproduce this in a supported version of Ubuntu? (Focal / 20.04 or later)

MaVeron30:
> I'm hesitant to add this, you shouldn't have to do this if you're on a supported distro version and you haven't configured your system in some exotic way that breaks things (e.g., default umask). Can you reproduce this in a supported version of Ubuntu? (Focal / 20.04 or later)

I had to use these commands on my Ubuntu 22.04.5 LTS (Jammy Jellyfish), which I installed on wsl 2 on Windows 10.
I am not a professional programmer, but just an amateur, but this is the only thing that helped me, although I set everything according to the documentation.

PR #20860, opened 2024-09-12
ENGDOCS-2217

aevesdocker:
## Description
Adds known RAM caveats, as per https://docker.slack.com/archives/C04300R4G5U/p1726126126404169

## Reviews

- [ ] Technical review
- [ ] Editorial review
- [ ] Product review


fazlan-nazeem:
@aevesdocker I think the following would be a bit less confusing. Let me know your thoughts. 

> Images which are pulled by Docker Desktop when enabling Docker Debug or Kubernetes features are not restricted by default, even if Docker Hub is blocked by RAM.

PR #20899, opened 2024-09-15
add removing from apt as part of uninstalling

QWERTIOX:
## Description

Add commands to clean up apt config when uninstalling

## Reviews

Check whitespace

- [ ] Technical review
- [x] Editorial review
- [ ] Product review


dvdksn (approved):
LGTM

thaJeztah:
@dvdksn we should probably apply the same changes to the ubuntu version as well

QWERTIOX:
> @dvdksn we should probably apply the same changes to the ubuntu version as well

I'm on it

PR #20906, opened 2024-09-16
document workaround for missing docker-compose command after update

ndeloof:
## Description

document workaround for missing docker-compose command after update

## Related issues or tickets

closes https://github.com/docker/for-mac/issues/7345
https://docker.atlassian.net/browse/COMP-747

## Reviews

- [ ] Technical review
- [ ] Editorial review
- [ ] Product review


content/manuals/compose/install/_index.md
```diff
@@ -30,6 +30,17 @@ Docker Desktop is available on:
...
 
+> [!NOTE]
+>
+> After removal of End-of-Lived Docker Compose V1 in [4.23.0](/desktop/release-notes/#4230), `docker-compose` command points directly to Docker Compose V2 binary, ran in standalone mode. 
+> If you rely on Docker Desktop auto-update, the symlink might be broken and command unavailable, as update doesn't ask for administrator password. 
+> To fix it, just recreate the symlink:
+> ```console
+> $ sudo rm /usr/local/bin/docker-compose
+> $ sudo ln -s /Applications/Docker.app/Contents/Resources/cli-plugins/docker-compose /usr/local/bin/docker-compose
+> ```
```
glours, line 41: For users who don't have admin right on their machine we should mention the alias creation as an another workaround, no?
Maybe we should also specify this only affect Mac users

ndeloof: IMHO better not give them a non-admin workaround, as installation would then diverge from the canonical one, and let their super-admin manage the required update in its company-wide orchestration automation

glours: Right, make sense 👌

[!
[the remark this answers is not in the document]
ndeloof: @aevesdocker "migrate" is about switching from v1 to v2, while this issue affects users who already are using compose v2, just suddenly miss the binary. We consider `docker-compose` (dash) command to be 100% valid when used with Compose v2, as "standalone" mode

PR #21037, opened 2024-09-30
dnf5 have no config-manager

codecreatede:
in the fedora 40 and above dnf5 is implemented and hence the dnf command listed in the docker will not work. I have updated the dnf command to dnf-3 as then you will be able to use the same.

## Reviews

- [ ] Technical review
- [ ] Editorial review
- [ ] Product review

### DIFF
--- a/content/manuals/engine/install/fedora.md
+++ b/content/manuals/engine/install/fedora.md
@@ -83,7 +83,7 @@ your DNF repositories) and set up the repository.
 
 ```console
 $ sudo dnf -y install dnf-plugins-core
-$ sudo dnf config-manager --add-repo {{% param "download-url-base" %}}/docker-ce.repo
+$ sudo dnf-3 config-manager --add-repo {{% param "download-url-base" %}}/docker-ce.repo
 ```
 
 #### Install Docker Engine


thaJeztah:
Thanks! We published packages for Fedora 40, but it's not yet release, so we did not yet update the docs. Perhaps instead we should differentiate for Fedora 40 and up (dnf5) until Fedora 39 and older are EOL. For that we should check though if the bugs in dnf5 were fixed, as there was an issue in the current version of the Fedora 40 (beta) image that caused some things to fail; https://github.com/docker/docker-install/blob/6d51e2cd8c04b38e1c2237820245f4fc262aca6c/install.sh#L560-L569

(That bug was fixed, but was not yet included in the current fedora:40 images);

- https://github.com/rpm-software-management/dnf5/issues/1603

codecreatede:
Thank you and actually it is not about the Fedora 40. It is also about the variants of the Fedora such as SilverBlue and Ultramine and Nobara, so people might find that docker is not working so i added it. If you merge then it would be ok or else we can close this comment. Adding a tag to that will actually help as some people ares still using the old version of the RedHat. 

Evening to all,
Gaurav

dvdksn (approved):
I think explicitly calling `dnf-3` seems like a good option for now, to retain compatibility with current and adjacent distros like silverblue, while also ensuring that the instruction remains as simple as possible. Fedora 39 [reaches EOL on November 12th](https://fedorapeople.org/groups/schedule/f-39/f-39-key-tasks.html), I propose that we update this instruction to use dnf5 then, with a caution to users that do not yet have dnf5 to upgrade.

codecreatede:
@dvdksn Thank you David and thank you approving the changes. This will actually help even for those containers which are using the older version of the fedora. I have just pushed a new fedora devops here https://github.com/codecreatede/fedora-devops and in the second release intgerating the dnf3 so that it will be easier on the old systems. 

Basically i will write a uname-a and then strings.Split(exec.Command("uname", "-a") and if the [0] is greater than time.Date().String then execute the dnf5 else dnf3. 

Thank you approving the changes to the docker docs. 

Tack,
Gaurav

codecreatede:
Thank you David for the merge into the main branch. @dvdksn

thaJeztah:
> Fedora 39 [reaches EOL on November 12th](https://fedorapeople.org/groups/schedule/f-39/f-39-key-tasks.html), I propose that we update this instruction to use dnf5 then

I realized I had an "off by one"; it's Fedora 41 that uses dnf5 as default, which means we need to wait for Fedora 40 to reach EOL

codecreatede:
@thaJeztah than you and it happens. I am happy that you have merged the pull request. this is for the other users also and i thank you for the interaction and i look forward to interact with you all more. Have a good evening all, Gaurav

PR #21045, opened 2024-10-02
Update best-practices.md

mrexodia:
## Description

Adding `--no-install-recommends` is a common best practice and can massively reduce image sizes. Since people copy from this page it is good to add.


dvdksn (approved):
Why not!

PR #21061, opened 2024-10-04
lifecycle hooks

ndeloof:
## Description

documentation for https://github.com/docker/compose/pull/12166

## Related issues or tickets

[](https://docker.atlassian.net/browse/COMP-760)

## Reviews

- [ ] Technical review
- [ ] Editorial review
- [ ] Product review


content/manuals/compose/how-tos/lifecycle.md
```diff
@@ -0,0 +1,59 @@
...
+aliases:
+- /compose/hooks/
+---
+
+## Services lifecycle hooks
+
+As Docker Compose run a service container, the actual command being ran is managed
+by the combination of `entrypoint` and `command` as documented [here](https://docs.docker.com/engine/containers/run/#default-entrypoint). Those ley you manage
+container setup and termination, catching SITERM signal to process a graceful shutdown. In some circumstances anyway it is easier to consider those a separate
+command to be introduced as lifecycle hooks, to be ran just after the container
```
jhrotko, line 16: I am not sure what you mean here - 
```
In some circumstances anyway it is easier to consider those a separate
command to be introduced as lifecycle hooks, to be ran just after the container
is started, or just before Compose stop it.
```

Could you rephrase it?

content/manuals/compose/how-tos/lifecycle.md
```diff
@@ -0,0 +1,59 @@
...
+is started, or just before Compose stop it.
+
+Another benefit for lifecycle hooks, is that hook command can run with additional
```
jhrotko, line 19: Suggestion: instead of 
`Another benefit for lifecycle hooks, is that hook command can run with additional privileges`
-> 
`Another advantage of lifecycle hooks is the ability to run hook commands with elevated privileges...`

content/manuals/compose/how-tos/lifecycle.md
```diff
@@ -0,0 +1,59 @@
...
+privileges, typically run as `root` or even `privileged` while the container is
+configured to have lower privileges to enhance security.
+
+### Post-Start hooks
+
+Post-Start lifecycle hooks are ran after container has been started. There's no
```
jhrotko, line 25:
```
There's no guarantee on the timing hook executes while the container entrypoint is ran.
```

to 
```
The timing of the hook execution is not guaranteed while the container's entrypoint is running.
```

content/manuals/compose/how-tos/lifecycle.md
```diff
@@ -0,0 +1,59 @@
...
+
+volumes:
+  data: {} # a docker volume is created with root ownership
+```
+
+### Pre-Stop hooks
+
+Pre-Stop lifecycle hooks are ran before an explicit request to stop container by a 
+compose command, either `docker compose down` or interrupting `docker compose up`
+with Ctrl+C. Hook won't execute if container stops on its own or is killed.
```
jhrotko, line 51:
```suggestion
with Ctrl+C. This hook won't execute if container finishes its execution or if it is killed.
```

jhrotko (changes requested):

content/manuals/compose/how-tos/lifecycle.md
```diff
@@ -0,0 +1,59 @@
...
+container setup and termination, catching SITERM signal to process a graceful shutdown. In some circumstances anyway it is easier to consider those a separate
+command to be introduced as lifecycle hooks, to be ran just after the container
+is started, or just before Compose stop it.
+
+Another benefit for lifecycle hooks, is that hook command can run with additional
+privileges, typically run as `root` or even `privileged` while the container is
+configured to have lower privileges to enhance security.
+
+### Post-Start hooks
+
```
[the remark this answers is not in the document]
ndeloof: > These hooks don't wait for the container's main task (entrypoint) to finish.

entrypoint never finish, as this would mean container stopped. Here entrypoint and hook run concurently, with no guarantee on when things happen

ndeloof:
better get this published with the compose release

PR #21109, opened 2024-10-12
Mentioning nuances around the term

azarboon:
Marc Brooker has written [a blog post](https://brooker.co.za/blog/2023/06/19/container.html) explaining that "container" means different thing to different people, and that it should be checked whether audience and narrator use the same definition. Below are his four suggested definition. Given the authority of Docker, I suggest this nuance to be mentioned. So readers are more informed and take cautionary acts to ensure they are in sync with their audience. It can also prevent confusion / misunderstanding for non-experts. I've added it. 

> - An approach to packaging an application along with its dependencies (sometimes a whole operating system user space), that can then run on a minimal runtime environment with a clear contract.
> - A set of development, deployment, architectural, and operational approaches built around applications packaged this way.
> - A set of operational, security, and performance isolation tools that allow multiple applications to share an operating system without interfering with each other. On Linux, this tools include chroot, cgroups, namespaces, [seccomp](https://man7.org/linux/man-pages/man2/seccomp.2.html), and others.
> - A set of implementations of these practices (the proper nouns, Docker, Kubernetes, ECS, etc).

## Reviews

- [ ] Technical review
- [ ] Editorial review
- [ ] Product review

### DIFF
--- a/content/get-started/docker-concepts/the-basics/what-is-a-container.md
+++ b/content/get-started/docker-concepts/the-basics/what-is-a-container.md
@@ -32,6 +32,8 @@ Here's what makes them awesome. Containers are:
 - Independent. Each container is independently managed. Deleting one container won't affect any others.
 - Portable. Containers can run anywhere! The container that runs on your development machine will work the same way in a data center or anywhere in the cloud!
 
+Note that "container" is an overloaded term. Whenever you use the word, check whether your audience uses the same definition.
+
 ### Containers versus virtual machines (VMs)
 
 Without getting too deep, a VM is an entire operating system with its own kernel, hardware drivers, programs, and applications. Spinning up a VM only to isolate a single application is a lot of overhead.


content/get-started/docker-concepts/the-basics/what-is-a-container.md
```diff
@@ -32,6 +32,8 @@ Here's what makes them awesome. Containers are:
 - Independent. Each container is independently managed. Deleting one container won't affect any others.
 - Portable. Containers can run anywhere! The container that runs on your development machine will work the same way in a data center or anywhere in the cloud!
 
+Note that "container" is an overloaded term. Whenever you use the word, check whether your audience uses the same definition.
+
```
dvdksn, line 36: I don't think this is helpful. I think the only thing this does is add ambiguity; these docs are, as you say, authoritative.

dvdksn (changes requested):

PR #21124, opened 2024-10-15
Update deprecated field in configure-builder.md 

Hapshanko:
## Description

Got this warning when using config-inline and decided to update the docs accordingly so anyone implementing will use the new field:

Warning: Input 'config-inline' has been deprecated with message: Use buildkitd-config-inline instead

## Reviews

- [ ] Technical review
- [ ] Editorial review
- [ ] Product review


crazy-max (approved):
Thanks!

PR #21217, opened 2024-10-23
Updated pricing information

sbenhoff007:
## Description

As our Docker subscription plans are being updated, we need to remove subscription information from our guides. Instead of hardcoding pricing information, we need to point to the official pricing pages instead.

- [ ] Technical review
- [ ] Editorial review
- [ ] Product review


dvdksn (approved):
LGTM

PR #21230, opened 2024-10-24
Spelling fixes

ArieHein:
## Description

Christmas is early :) and i had some free time to go over all the documents and fix all spelling mistakes i have found.
Sorry in advance for giving you hard work checking so many files.

I recommend you add something like spelling and grammar step in the PR process on markdown files (i use cspell but there are others).

There also about 15 grammar fixes not included that are specific to US English (ex. Initialize) vs British English (ex. Initialise)
and since according to the docs you are using US grammar - https://docs.docker.com/contribute/style/grammar/
I can create a PR for that, but it might be a bit like a grammar police, so i didn't include it in this one. 
Shout if you want that as well in a separate PR.

- [ ] Technical review
- [x] Editorial review
- [ ] Product review


dvdksn (approved):
Wow, thank you!!!

We do have spell checking, in [vale](https://vale.sh), but it seems like we could improve this part of our validation workflow.

Cheers, appreciate it!

PR #21314, opened 2024-11-01
Remove an duplicate list of parser directives

t-dub:
## Description

The [Dockerfile reference](https://docs.docker.com/reference/dockerfile/) has an extra -- incomplete and unlinked -- set of parser directives included in it.  I'm assuming it was just accidentally left over from a previous revision

image

*Further up the page is the "real" section:*

image


dvdksn (changes requested):
Thank you! This file lives in the [moby/buildkit repository](https://github.com/moby/buildkit/blob/master/frontend/dockerfile/docs/reference.md). We'll need to update it there, would you mind raising a PR on that repository? The fix will be vendored here when it's merged upstream. Thank you!

PR #21326, opened 2024-11-04
Engdocs 2282

aevesdocker:
## Description

Add lifecycle attributes to the compose spec

## Reviews

- [ ] Technical review
- [x] Editorial review
- [ ] Product review

### DIFF
--- a/content/reference/compose-file/services.md
+++ b/content/reference/compose-file/services.md
@@ -1532,6 +1532,8 @@ For more information, see [Use lifecycle hooks](/manuals/compose/how-tos/lifecyc
 
 ## pre_stop
 
+{{< introduced compose 2.30.0 "../../manuals/compose/releases/release-notes.md#2300" >}}
+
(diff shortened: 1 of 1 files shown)


content/reference/compose-file/services.md
```diff
@@ -1505,6 +1505,35 @@ ports:
     mode: host
 ```
 
+## post_start
+
+`post_start` defines a sequence of lifecycle hooks to run after a container has started. The exact timing of when the command is run is not guaranteed.
+
+- `command`: The command to run after the container has started. This attribute is required.
```
dvdksn, line 1512: Probably worth mentioning that this works the same way as `CMD` and the default `command` (E.g., you can use both the shell form and exec form. In the case of exec form, no shell is invoked.)

content/reference/compose-file/services.md
```diff
@@ -1505,6 +1505,35 @@ ports:
     mode: host
 ```
 
+## post_start
+
```
dvdksn, line 1509: Add version it was introduced in

dvdksn (approved):

PR #21345, opened 2024-11-05
release-notes for Compose v2.30.2 version

glours:
## Description
Add release notes of Compose `v2.30.2` version

## Reviews

- [ ] Technical review
- [x] Editorial review
- [ ] Product review


dvdksn (commented):
Build error is due to malformatted file in buildkit, which I patched in a hotfix. It's also fixed upstream but I haven't re-vendored that commit yet, which is causing this vendoring to fail. So we need this first: #21350

dvdksn (approved):

PR #21382, opened 2024-11-08
engine/install: enable engine as system service on rpm installs

thaJeztah:
- related / alternative to https://github.com/docker/docs/pull/21380
- ℹ️ more follow-ups coming; getting this one in first 😄 

Unlike deb-based systems, installing an rpm doesn't automatically start the Docker service, and starting the service doesn't make it automatically start on (re)boot.

Update the instructions so that Docker is automatically started on boot.

## Reviews

- [ ] Technical review
- [ ] Editorial review
- [ ] Product review


dvdksn (commented):
I would really like to avoid starting containerd as a service manually. It doesn't seem necessary. Let's go with my PR instead.

dvdksn:
Superseded by #21380

PR #21384, opened 2024-11-08
engine/install: sync between distros, use `dnf` for RHEL and CentOS, and some touch-ups

thaJeztah:
- follow-up to https://github.com/docker/docs/pull/21380

----

- engine/install: enable engine as system service on rpm installs
    Unlike deb-based systems, installing an rpm doesn't automatically start the Docker service, and starting the service doesn't make it automatically start on (re)boot. Update the instructions so that Docker is automatically started on boot. follow-up to 1514b0dc3662970121091028459abb3ab496cf57, which left some of the old instructions.
- engine/install: Docker -> Docker Engine
    Make it more clear we're starting the engine (not "docker", the CLI).
- engine/install: remove -y from yum installation steps
    Use interactive installation for packages from our repository, so that the user can confirm what will be installed. For the distro-packages (yum-utils), we keep the non-interactive, as these should be generally ok to install. follow-up to fde4750c859ab0cf386b30cc19eb9ea7b9cfe24c
- engine/install: use dnf for CentOS and RHEL
    The versions of CentOS and RHEL we support now have dnf as a default instead of yum, so rewrite these to use dnf.
- engine/install: fix leftover old version in ubuntu
- engine/install: fix indentation in debian, raspberry-pi-os
- engine/install: fix some stray whitespace
- engine/install: ubuntu/rpi-os: sync uninstall with debian steps
- engine/install: sync some wording between distros

## Reviews

- [ ] Technical review
- [ ] Editorial review
- [ ] Product review

### DIFF
--- a/content/manuals/engine/install/centos.md
+++ b/content/manuals/engine/install/centos.md
@@ -36,9 +36,11 @@ default. If you have disabled it, you need to re-enable it.
 
 ### Uninstall old versions
 
-Older versions of Docker went by `docker` or `docker-engine`.
-Uninstall any such older versions before attempting to install a new version,
-along with associated dependencies.
+Before you can install Docker Engine, you need to uninstall any conflicting packages.
+
+Your Linux distribution may provide unofficial Docker packages, which may conflict
+with the official packages provided by Docker. You must uninstall these packages
+before you install the official version of Docker Engine.
 
 ```console
 $ sudo dnf remove docker \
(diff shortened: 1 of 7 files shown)


dvdksn (changes requested):
🤦🏻 ignore me, I was looking at the wrong file.

dvdksn (dismissed):

dvdksn (approved):

PR #21418, opened 2024-11-13
Fix: .docker hidden folder instead of docker

MovsisyanM:
Following the steps you end up with a hidden folder `.docker`, not a normal `docker` folder.

## Description

To avoid errors and debugging steps when installing docker compose

## Reviews

- [ ] Technical review
- [ ] Editorial review
- [ ] Product review

### DIFF
--- a/content/manuals/compose/install/linux.md
+++ b/content/manuals/compose/install/linux.md
@@ -111,7 +111,7 @@ To update the Compose plugin, run the following commands:
     or, if you chose to install Compose for all users:
 
     ```console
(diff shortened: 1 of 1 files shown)


dvdksn (commented):
I'm not sure we'd want to make the globally installed plugin a hidden directory. It's probably better to keep as-is.

Line 101 also refers to this path: https://github.com/docker/docs/pull/21418/files#diff-8512592555c208380ae974bf5442e2c7bf2b21672df616544c1fae9cb9bfca0dR101

PR #21446, opened 2024-11-17
Update gettingstarted.md to use f-string

AJDempsey:
Python's f-strings are the preferred method of string formatting now. Updated the python code examples to use them instead of .fromat()

## Description

Removed the format function in favour of using f-strings instead


dvdksn (approved):
👍🏻

PR #21515, opened 2024-11-27
Add section about conflict with snap version of docker

TheGuit:
## Reviews

- [ ] Technical review
- [ ] Editorial review
- [ ] Product review


dvdksn:
I think this information is better placed in this section: https://docs.docker.com/engine/install/ubuntu/#uninstall-old-versions

TheGuit:
> I think this information is better placed in this section: https://docs.docker.com/engine/install/ubuntu/#uninstall-old-versions

Ok, I've modified.

PR #21562, opened 2024-12-04
Add multi-stage builds to the C++ guide

aerabi:
## Description

Having a multi-stage build is an essential thing to learn for containerizing a compiled language.

## Related issues or tickets

N/A

## Reviews

- [ ] Technical review
- [ ] Editorial review
- [ ] Product review

### DIFF
--- a/content/guides/cpp/multistage.md
+++ b/content/guides/cpp/multistage.md
@@ -1,0 +1,109 @@
+---
+title: Create a multi-stage build for your C++ application
+linkTitle: Containerize your app
+weight: 60
+keywords: C++, containerize, multi-stage
+description: Learn how to create a multi-stage build for a C++ application.
+aliases:
+- /language/cpp/containerize/
+- /guides/language/cpp/containerize/
+---
+
+## Prerequisites
+
+- You have a [Git client](https://git-scm.com/downloads). The examples in this section use a command-line based Git client, but you can use any client.
+
+## Overview
+
+This section walks you through creating a multi-stage Docker build for a C++ application.
+A multi-stage build is a Docker feature that allows you to use different base images for different stages of the build process,
+so you can optimize the size of your final image and separate build dependencies from runtime dependencies.
+
+## Get the sample application
+
+We're using the same sample repository that you used in the previous sections of this guide. If you haven't already cloned the repository, clone it now:
+
+```bash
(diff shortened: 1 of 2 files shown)


content/guides/cpp/multistage.md
```diff
@@ -0,0 +1,109 @@
...
+This section walks you through creating a multi-stage Docker build for a C++ application.
+A multi-stage build is a Docker feature that allows you to use different base images for different stages of the build process,
+so you can optimize the size of your final image and separate build dependencies from runtime dependencies.
+
+## Get the sample application
+
+We're using the same sample repository that you used in the previous sections of this guide. If you haven't already cloned the repository, clone it now:
+
+```bash
+$ git clone https://github.com/dockersamples/c-plus-plus-docker.git
```
aerabi, line 27: This repo should be synced with the upstream repo created by Pradumna.

content/guides/cpp/multistage.md
dvdksn, line 1: I don't think this needs to be a completely separate page. I think it could be a section on the first page: https://docs.docker.com/guides/cpp/containerize/#run-the-application

The sequence could be:

1. Get the repo
2. Run the app
3. Make it multi-stage

aerabi: The issue is that Pradumna's example is a bit complicated to make multi-stage. If I want to do it on the same page, I have to use my own example for the whole thing. Would that be OK?

dvdksn: @aerabi hi, yea that's fine - multi-stage is such an important strategy, especially for compiled programs, that I think it makes sense to present that as the main workflow.

dvdksn: Sounds good to me :-) But let me cc @craig-osterhout

aerabi: Pushed another commit, PTAL. 🙏 
CC @craig-osterhout

content/guides/cpp/multistage.md
aerabi: The repo is already updated. 🎉

aerabi:
Thanks for the reviews, @craig-osterhout! I pushed another commit, PTAL. 🙏 
I'll also update Pradumna's repo separately.

PR #21584, opened 2024-12-06
Update docs gc policy docs to mention that the policies are documented in the buildx prune command

rrauenza:
## Description
I spent a lot of time searching for the documentation of the specific policies (like `all`) and finally realized these are just arguments passed to buildx prune.


dvdksn (approved):
Good start

I need to add more details/examples, but happy to merge this to begin with

rrauenza:
It's also confusing that although docker buildx inspect shows ==, and the toml format expects ==, the json format only wants a single '='...

PR #21589, opened 2024-12-09
vale: add some acronyms and allowed words

thaJeztah:
###  vale: add ZFS as acronym

ZFS; https://en.wikipedia.org/wiki/ZFS

### vale: add userland to accept list

Used for (e.g.) "userland proxy"

## Related issues or tickets

- relates to https://github.com/docker/docs/pull/21546

## Reviews

- [ ] Technical review
- [ ] Editorial review
- [ ] Product review

### DIFF
--- a/_vale/Docker/Acronyms.yml
+++ b/_vale/Docker/Acronyms.yml
@@ -128,4 +128,5 @@ exceptions:
   - XML
   - XSS
   - YAML
+  - ZFS
   - ZIP
--- a/_vale/config/vocabularies/Docker/accept.txt
+++ b/_vale/config/vocabularies/Docker/accept.txt
@@ -169,6 +169,7 @@ tmpfs
 ufw
 umask
 ungated
+userland
 untrusted
 vSphere
 vpnkit


_vale/Docker/Acronyms.yml
```diff
@@ -129,3 +130,4 @@ exceptions:
   - XSS
   - YAML
   - ZIP
+  - ZFS
```
thaJeztah, line 133: Oh sorted wrong; let me fix

dvdksn (commented):
LGTM except `DAD` - I don't think we should add that as an acronym exception. (the `Acronyms.yml` file only lists acronyms that we allow not to be expanded before use, but I think DAD should be expanded. If it is expanded, the vale rule won't fire. The issue in pawel's PR was that the acronym was expanded without capitalization, I left a comment on that PR: https://github.com/docker/docs/pull/21546/files#r1875991549)

thaJeztah:
Ah! I was wondering that indeed; makes sense. I dropped that commit 👍

PR #21595, opened 2024-12-09
Fix typo in local exporter example

mikesir87:
## Description

Fixes a small typo in a sample command


dvdksn (approved):
Not sure if it's a typo but I agree that showing `local` in the example makes more sense, as it's more common.

PR #21596, opened 2024-12-09
Add example that shows the stated "default" settings in json

rrauenza:
(also fix an errant . )

## Description

Add an example that matches the stated defaults.

It isn't obvious from current docs that json requires only a single '=' sign.  Also the "type" settings all need to be in one string, so this gives an example showing that.

### DIFF
--- a/content/manuals/build/cache/garbage-collection.md
+++ b/content/manuals/build/cache/garbage-collection.md
@@ -26,7 +26,7 @@ Depending on the [driver](../builders/drivers/_index.md) used by your builder in
 the garbage collection will use a different configuration file.
 
 If you're using the [`docker` driver](../builders/drivers/docker.md), garbage collection
-can be configured in the [Docker Daemon configuration](/reference/cli/dockerd.md#daemon-configuration-file).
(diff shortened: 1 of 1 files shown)


rrauenza:
.... I see the lint error, but I'm also not convinced this feature actually works.   

I'm gonna open a bug to clarify.

dvdksn (changes requested):
Thank you, but I would like to wait with this change. The buildkitd gc config has changed in v0.18, and so has the defaults. While that version doesn't yet ship with Docker Engine, it soon will and I'll need to make space for both the old and the new config format (incl. buildkitd and daemon.json)

rrauenza:
So that would also address https://github.com/moby/buildkit/issues/5581 ?  With the == / = behavior?

dvdksn:
Hmm, no I don't think the filters are affected by the new config options. KeepBytes was replaced by three, more configurable options in v0.17. GC config via daemon.json does not yet use all of these new options. But I don't believe filtering was changed, it only gave new ways to define the GC threshold (min reserved bytes, maximum used space, min free disk space)

PR #21621, opened 2024-12-12
4.37: propose that (Legacy) Qemu VMM + osxfs will be removed in 4.39

djs55:
FYI @KatTomrushka 

## Description

Highlight 2 deprecations in the release notes. Technically both were highlighted before in the DD UI by adding "(Legacy)" labels, but it's better to over-communicate than under-communicate!

For filesharing, anyone still using osxfs should switch to virtiofs which is faster and more stable.

For running the Linux VM locally, anyone still using Qemu on Apple Silicon should switch to virtualization.framework which is also faster and more stable.

In case someone hits a problem, can't switch and needs help, include links to support and the github issue tracker.

For previous deprecated things in the release notes we've used vague language like "in a future release" but let's be specific and propose 4.39. If we discover a problem we can always postpone.

# Some history:

- In 2020 osxfs was marked as legacy (Docker Desktop 2.4.2.0).

- In 2023 with the release of macOS 14, all supported macOS versions (current release plus previous two) use virtualization.framework by default.

- In 2024 (Docker Desktop 4.36) Qemu as a VMM was marked as legacy in the UI.

## Reviews

- [ ] Technical review
- [ ] Editorial review
- [ ] Product review


content/manuals/desktop/release-notes.md
```diff
@@ -66,6 +66,13 @@ Take a look at the [Docker Public Roadmap](https://github.com/orgs/docker/projec
 
 - Kubernetes cluster may not start if **Registry Access Manager** is enabled. As a workaround, add `registry.k8s.io` and `<geo>-docker.pkg.dev` to **Registry Access Management** policies.
 
+### Deprecation
+
+#### For Mac
+
+- QEMU (Legacy) as a VMM on Apple Silicon will be removed in 4.39. Please switch to virtualization.framework for increased performance and stability. If you encounter an issue, [contact support](https://www.docker.com/support/) or file a [github issue](https://github.com/docker/for-mac/issues).
```
thaJeztah, line 73: - Looks like Vale doesn't like the "Please"; perhaps "Users are recommended to switch to ..."
- It also complains about "VMM" as unknown acronym; I think it works if you put it  like "Virtual Machine Manager (VMM) ..." (is it "manager"?)

KatTomrushka:
I am happy to put a note in the release notes but let's not commit to when we will deprecate without an internal alignment first.
To say that we will completely remove the option from the GUI in DD 4.39 (end of Feb?) is a bit agressive timeline given that the adoption rate is slower specially in enterprises. Suggest identifying which business tier companies (if any) are affected and emailing them about it. Then, the typical process is 180 days for deprecation.

PR #21625, opened 2024-12-13
Update multi-stage-builds.md

shariqahmkhan:
added command to view logs of running container

## Reviews

- [x] Technical review
- [ ] Editorial review
- [ ] Product review

### DIFF
--- a/content/get-started/docker-concepts/building-images/multi-stage-builds.md
+++ b/content/get-started/docker-concepts/building-images/multi-stage-builds.md
@@ -231,7 +231,7 @@ Now that you have the project, you’re ready to create the `Dockerfile`.
 1. Now that you have an image built, it's time to run the container.
 
     ```console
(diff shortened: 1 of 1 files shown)


content/get-started/docker-concepts/building-images/multi-stage-builds.md
```diff
@@ -233,9 +233,14 @@ Now that you have the project, you’re ready to create the `Dockerfile`.
     ```console
     $ docker run -d -p 8080:8080 spring-helloworld
```
dvdksn, line 234: Thanks for catching this error. Actually, I think I would prefer that we rather drop the `-d` flag to have the container logs print directly to the terminal, avoiding a background container and log inspection altogether.

```suggestion
    $ docker run -p 8080:8080 spring-helloworld
```

dvdksn (approved):

PR #21639, opened 2024-12-16
ENGDOCS-2332

aevesdocker:
## Description

Adds new attributes to the spec from the recent 2.32.0 release

## Reviews

- [ ] Technical review
- [ ] Editorial review
- [ ] Product review

### DIFF
--- a/_vale/config/vocabularies/Docker/accept.txt
+++ b/_vale/config/vocabularies/Docker/accept.txt
@@ -153,6 +153,7 @@ kubectl
 kubefwd
 kubelet
 lookup
+label_file
 macOS
 macvlan
 mfsymlinks
--- a/content/reference/compose-file/services.md
+++ b/content/reference/compose-file/services.md
@@ -1156,7 +1156,7 @@ results in a runtime error.
 
 {{< introduced compose 2.23.2 "/manuals/compose/releases/release-notes.md#2232" >}}
 
-The `label_file property` attribute lets you load labels for a service from an external file or a list of files. This provides a convenient way to manage multiple labels without cluttering the Compose file.
+The `label_file` attribute lets you load labels for a service from an external file or a list of files. This provides a convenient way to manage multiple labels without cluttering the Compose file.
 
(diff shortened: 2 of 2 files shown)


content/reference/compose-file/develop.md
```diff
@@ -56,10 +56,36 @@ Compose to monitor source code for changes. For more information, see [Use Compo
...
+
+{{< introduced compose 2.23.2 "/manuals/compose/releases/release-notes.md#2232" >}}
+
+`exec` is only relevant when `action` is set to `sync+exec`. Like [service hooks](services.md#post_start), `exec` is used to define the command to be run inside the container once it has started. 
+
+- `command`: Specifies the command to run once the container starts. This attribute is required, and you can choose to use either the shell form or the exec form.
+- `user`: The user to run the command. If not set, the command is run with the same user as the main service command.
+- `privileged`: Lets the command run with privileged access.
+- `working_dir`: The working directory in which to run the command. If not set, it is run in the same working directory as the main service command.
+- `environment`: Sets the environment variables to run the command. While the command inherits the environment variables defined for the service’s main command, this section lets you add new variables or override existing ones.
```
sarahsanders-docker, line 74: just a consistency question. this bulleted lists uses `command: description` format, while the one before it uses `command, description` format. not sure if this was intentional and if so if there is a reason to use one over the other?

sarahsanders-docker (approved):
overall looks good! I left one nit/consistency question that I was curious about

_vale/config/vocabularies/Docker/accept.txt
```diff
@@ -153,6 +153,7 @@ kubectl
 kubefwd
 kubelet
 lookup
+label_file
```
dvdksn, line 156: Adding this to the vocabulary is not the right way to fix this lint error

If this is to disable the sentence case check, you should disable that rule instead.

dvdksn: OK, but there are a lot more headings that are flagged for the same thing in this and the other compose file docs. You could disable spellcheck, or better, wrap the options in code spans, to bypass this rule. Adding all of these params to the vocabulary doesn't seem like a good solution

PR #21654, opened 2024-12-17
Update ubuntu.md

f1rf1n:
## Description
Instruction did not mention having to start the docker service before testing 'hello world'. Following the instruction (as-was) literally leads to confusing errors.

Updated doc accordingly.

- [ ] Technical review
- [ ] Editorial review
- [ ] Product review

### DIFF
--- a/content/manuals/engine/install/ubuntu.md
+++ b/content/manuals/engine/install/ubuntu.md
@@ -176,6 +176,7 @@ Docker from the repository.
 3. Verify that the installation is successful by running the `hello-world` image:
 
    ```console
+   $ sudo service docker start
    $ sudo docker run hello-world
    ```
 


f1rf1n:
I just ran into this (on pop_os, a ubuntu-based distro) and into https://github.com/docker/docs/issues/16849

Is this is the instruction page for ubuntu, maybe it might be worth changing all references or mentions of `sudo service docker start` to  `sudo systemctl start docker` no?

Or add something ( a tip? a note? a troubleshoot section?) to the page?

dvdksn:
I'd be open to adding a note about it, along the lines of:


> [!NOTE]
> If you see the following error message:  
>  
> `docker: Cannot connect to the Docker daemon at unix:///var/run/docker.sock. Is the docker daemon running? See 'docker run --help'.`  
>  
> This indicates that the Docker daemon is not running. You can start it manually by running the following command:
>  
> ```console
> $ sudo systemctl enable --now docker
> ```  
>  
> This command will start the Docker daemon immediately and ensure it starts automatically on system boot.

(+1 on changing `sudo service` to `sudo systemctl`)

thaJeztah:
Perhaps we could add  to the existing note could work? We may want to avoid sprinkling too many notes in between, specifically because these would be for non-standard installations?

Screenshot 2024-12-18 at 10 03 49


Something like 

```markdown
The `docker` service does not start automatically after installation on some derivative and you may need to run `sudo systemctl start docker` after installation."
````

dvdksn:
@thaJeztah yeah I agree with the noise factor but that note is in the wrong place for this info. Plus a lot of users already miss that note.

dvdksn:
@thaJeztah you know what would fix all of this right? `curl https://get.docker.com/ | bash` 😄

thaJeztah:
I'm not sure the `get.docker.com` works with all derivatives (we had some code in there at some point, but all derivatives had their own issues, and we have no way of testing all flavours out there ... each with their own issues), which was one of the reasons we removed most mentions of derivatives as it was near impossible to maintain.

PR #21697, opened 2024-12-29
Update 05_persisting_data.md

shariqahmkhan:
Added info for those running containers in Windows and are curious to find where is /var or what it would be like if they were running Windows containers

## Reviews

- [ ] Technical review
- [ ] Editorial review
- [ ] Product review

### DIFF
--- a/content/get-started/workshop/05_persisting_data.md
+++ b/content/get-started/workshop/05_persisting_data.md
@@ -21,7 +21,7 @@ changes won't be seen in another container, even if they're using the same image
 
 ### See this in practice
 
-To see this in action, you're going to start two container's. In one container,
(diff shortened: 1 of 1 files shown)


shariqahmkhan (commented):
changed containers to container's based on context

shariqahmkhan (commented):
made it concise as much possible without loosing info

dvdksn (commented):
This section is already mostly an "FYI" / if you're curious. It's really out of scope for a getting started workshop, so I'm hesitant to add this.

usha-mandya:
@shariqahmkhan Thanks for the PR. I'll have to close this PR without accepting changes as the suggested updates are out of scope for the get started docs.

PR #21699, opened 2024-12-29
Update order of images in docker image tag command

nebojsatomic:
There was a mistake in the command for tagging an image that you have present in your list of images, order should be like this:  `docker image tag ALREADY_PRESENT_IMAGE  IMAGE_THAT_YOU_WANT_TO_PUSH` not the other way around

### DIFF
--- a/content/get-started/docker-concepts/building-images/build-tag-and-publish-an-image.md
+++ b/content/get-started/docker-concepts/building-images/build-tag-and-publish-an-image.md
@@ -94,7 +94,7 @@ docker build -t my-username/my-image .
 If you've already built an image, you can add another tag to the image by using the [`docker image tag`](https://docs.docker.com/engine/reference/commandline/image_tag/) command:
 
 ```console
-docker image tag my-username/my-image another-username/another-image:v1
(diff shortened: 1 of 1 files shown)


thaJeztah:
Not sure if there's a direct mistake on the page though; the example looks to continue from the example shown a couple of lines above that shows that an image was built and tagged as `my-username/my-image`;

```console
docker build -t my-username/my-image .
```

nebojsatomic:
> Not sure if there's a direct mistake on the page though; the example looks to continue from the example shown a couple of lines above that shows that an image was built and tagged as `my-username/my-image`;
> 
> ```
> docker build -t my-username/my-image .
> ```

Hm, yes, 
but it is a bit confusing if you already have a built image with another name ( not my-username/my-image ) and you want to push my-username/my-image. 
The text above `docker image tag` says "If you've already built an image" which could mean that you might not have to do the `docker build -t`, and you could have named your image previously web:latest for example.

Anyway, whatever you decide, is good for me :)

usha-mandya:
@nebojsatomic Thanks for the PR. I'm closing this without accepting the changes based on the discussion above. Thank you.

PR #21703, opened 2024-12-31
new genai rag guide

Expeto:
## Description
Added a new guide about RAG with genAI which also explains how RAG actually work

## Reviews

- [ ] Technical review
- [x] Editorial review
- [ ] Product review

### DIFF
--- a/content/guides/genai-rag.md
+++ b/content/guides/genai-rag.md
@@ -1,0 +1,206 @@
+---
+title: Leveraging RAG in GenAI to teach new information
+linkTitle: Leveraging RAG in GenAI
+description:  This guide walks through the process of setting up and utilizing a GenAI stack with Retrieval-Augmented Generation (RAG) systems and graph databases. Learn how to integrate graph databases like Neo4j with AI models for more accurate, contextually-aware responses.
+keywords: Docker, GenAI, Retrieval-Augmented Generation, RAG, Graph Databases, Neo4j, AI, LLM
+summary: |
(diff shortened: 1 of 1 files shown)


dvdksn (commented):
Thank you for your contribution! I appreciate your effort; RAG is an important
topic, and expanding our documentation on it can add significant value for our
readers. However, I think this guide could benefit from more focus and
refinement to align with the needs of our audience and complement our existing
materials. Here are my thoughts:

- The guide's purpose is unclear. It's important to define what this guide
  offers that the existing RAG guide does not. It seems like you want to focus
  on RAG with graph databases and Neo4j in particular. If the goal of the guide
  is to explore and demonstrate the benefits of this technology in particular,
  consider a way of doing that by example. E.g., by comparing this approach to
  using a vector database.
- Th structure and flow of the article feels fragmented and doesn't follow a
  logical progression. For instance, introducing graph databases before RAG
  seems out of place, especially for readers unfamiliar with the primary topic.
  Start with RAG and its benefits, then explain how graph databases fit into
  the picture.
- The "Case Study" section is disconnected from the rest of the content. I
  would suggest removing it. Or if you want to keep it, integrate it into a
  narrative that supports the guide.
- Several sections lack detail and direction. The technical instructions for
  setting up the stack are too sparse to be genuinely useful. I'm missing
  detailed, hands-on workflows and concrete examples that readers can follow
  and apply.

Expeto:
@dvdksn I did the changes, can you take another look. Thanks.

PR #21708, opened 2025-01-01
add another iptables rule to allow dns queries from container

fliespl:
## Description
With only mentioned rule, DNS queries from containers won't work.

## Reviews
- [X] Technical review
- [X] Editorial review
- [ ] Product review

### DIFF
--- a/content/manuals/engine/network/packet-filtering-firewalls.md
+++ b/content/manuals/engine/network/packet-filtering-firewalls.md
@@ -126,6 +126,17 @@ the source and destination. For instance, if the Docker host has addresses
 `2001:db8:1111::2` and `2001:db8:2222::2`, you can make rules specific to
 `2001:db8:1111::2` and leave `2001:db8:2222::2` open.
 
+You may need to allow responses from servers outside the permitted external address
+ranges. For example, containers may send DNS or HTTP requests to hosts that are
+not allowed to access the container's services. The following rule accepts any
+incoming or outgoing packet belonging to a flow that has already been accepted
+by other rules. It must be placed before `DROP` rules that restrict access from
+external address ranges.
+
+```console
+$ iptables -I DOCKER-USER -m state --state RELATED,ESTABLISHED -j ACCEPT
+```
+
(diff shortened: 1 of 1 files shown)


content/manuals/engine/network/packet-filtering-firewalls.md
```diff
@@ -119,6 +119,11 @@ the source and destination. For instance, if the Docker host has addresses
 `2001:db8:1111::2` and `2001:db8:2222::2`, you can make rules specific to
 `2001:db8:1111::2` and leave `2001:db8:2222::2` open.
 
+If your containers are also querying DNS, you should add this rule as well to allow them to work:
```
robmry, line 122: The suggested rule isn't specific to DNS, it accepts any incoming or outgoing packet that's part of a flow that's already been allowed by some other rule.

So, how about ...

```suggestion
You may wish to allow responses from servers outside the permitted external address ranges. For example, containers may send DNS or HTTP requests to hosts that are not allowed to access the container's services. The following rule accepts any incoming or outgoing packet belonging to a flow that has already been accepted by other rules. It must be placed before `DROP` rules that restrict access from external address ranges.
```

robmry (changes requested):
Thank you @fliespl ... a hint like this to go along with the existing examples looks useful, without taking us too far down the path of providing general `iptables` documentation.

content/manuals/engine/network/packet-filtering-firewalls.md
fliespl: Yeap, this makes much more sense :)

content/manuals/engine/network/packet-filtering-firewalls.md
```diff
@@ -119,6 +119,11 @@ the source and destination. For instance, if the Docker host has addresses
...
+```
+$ iptables -I DOCKER-USER -m state --state RELATED,ESTABLISHED -j ACCEPT
+```
```
thaJeztah, line 125: If you're updating, can you also;

- add a  newline before  the code-block
- add a `console` code-hint to  make  sure it's properly highlighted?


````suggestion
```console
$ iptables -I DOCKER-USER -m state --state RELATED,ESTABLISHED -j ACCEPT
```
````

thaJeztah:
@robmry @aevesdocker I rebased this one and applied your suggestions; PTAL

PR #21709, opened 2025-01-01
chore: cleanup instructions for unsupported version

syaghoubi00:
## Description

Debian 10 is no longer a supported version
https://github.com/docker/docs/blob/73d74add2bb1825d562770362da8247e925e8801/content/manuals/engine/install/debian.md?plain=1#L40-L49


thaJeztah:
Thanks! I had to double-check until when the LTS project  maintained it, but looks indeed that  that ended as well (https://wiki.debian.org/DebianReleases), so it's only ELTS (commercial) that's now remaining (https://wiki.debian.org/LTS/Extended), which we don't have to care about in this context

thaJeztah (approved):
LGTM

PR #21717, opened 2025-01-04
Corrected the shell name from dash to bash at line 467

SudheerKumarP0357:
## Description
Simple Correction for Shell name - `dash` to `bash`

I have modified the shell name at line 467 from `dash` to `bash`
## Reviews

- [ ] Technical review
- [X] Editorial review
- [ ] Product review


dvdksn (changes requested):
I am pretty sure dash here is correct, referring to Debian Almquist shell. (`-o pipefail` is supported in bash)

SudheerKumarP0357:
Sorry, rookie mistake, I would have researched before requesting the pr.

PR #21749, opened 2025-01-09
directly support ubuntu based distros in add repository snippet

nisrael:
default to `VERSION_CODENAME` but use `UBUNTU_CODENAME` if present and remove comment to do so manually

## Description

The note, that was added in the documentation to use `UBUNTU_CODENAME` instead of `VERSION_CODENAME` could easily be implemented in the snippet directly, by using shell parameter expansion with a default value. In case `UBUNTU_CODENAME` is present in `/etc/os-release` it is used and otherwise the regular `VERSION_CODENAME`.


dvdksn (commented):
👀 LGTM cc @thaJeztah

thaJeztah (commented):
not sure; still on the fence on this. We don't test on any of these derivatives, which is why we added a section about this; https://docs.docker.com/engine/install/#other-linux-distributions

trying to make it "just work" without the user acknowledging / realizing it's an unsupported setup may make it seem as if it's supported / tested.


@vvoland @tianon @neersighted any thoughts?

content/manuals/engine/install/ubuntu.md
```diff
@@ -127,16 +127,11 @@ Docker from the repository.
...
+     $(. /etc/os-release && echo "${UBUNTU_CODENAME:-$VERSION_CODENAME}") stable" | \
      sudo tee /etc/apt/sources.list.d/docker.list > /dev/null
    sudo apt-get update
    ```
 
-   > [!NOTE]
-   >
-   > If you use an Ubuntu derivative distribution, such as Linux Mint,
-   > you may need to use `UBUNTU_CODENAME` instead of `VERSION_CODENAME`.
-
```
vvoland, line 139: Personally, I'm okay with making it use `UBUNTU_CODENAME` if available, as long as we don't remove the notice that derivatives are not supported officially (but MAY work).

nisrael: I rephrased the notice and moved it to the section, where the supported distributions are listed.

tianon:
Agreed, should be mostly harmless. :+1:

vvoland (changes requested):
LGTM after fixing linter errors

nisrael:
fixed the linter errors

dvdksn (approved):

PR #21784, opened 2025-01-10
hub: clarify service accounts

craig-osterhout:
## Description

Clarify that enhanced service account add-ons are no longer available. Service accounts themselves can still be used.

## Reviews

- [ ] Editorial review


sarahsanders-docker (approved):
LGTM!

PR #21848, opened 2025-01-17
ENGDOCS-2382

aevesdocker:
## Description

Some Compose spec updates:

Adds https://github.com/compose-spec/compose-spec/pull/557
Adds https://github.com/compose-spec/compose-spec/pull/555

## Reviews

- [ ] Technical review
- [ ] Editorial review
- [ ] Product review

### DIFF
--- a/content/reference/compose-file/services.md
+++ b/content/reference/compose-file/services.md
@@ -773,11 +773,9 @@ extends:
 
 #### Restrictions 
 
-When a service uses `extends`, it can also specify dependencies on other resources, an explicit `volumes` declaration for instance. However, it's important to note that `extends` does not automatically incorporate the target volume definition into the extending Compose file. Instead, you are responsible for ensuring that an equivalent resource exists for the service being extended to maintain consistency. Docker Compose verifies that a resource with the referenced ID is present within the Compose model.
+When a service is referenced using `extends`, it can declare dependencies on other resources. These dependencies may be explicitly defined through attributes like `volumes`, `networks`, `configs`, `secrets`, `links`, `volumes_from`, or `depends_on`. Alternatively, dependencies can reference another service using the `service:{name}` syntax in namespace declarations such as `ipc`, `pid`, or `network_mode`.
 
-Dependencies on other resources in an `extends` target can be:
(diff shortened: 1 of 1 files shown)


content/reference/compose-file/services.md
```diff
@@ -257,16 +257,24 @@ cgroup_parent: m-executor-abcd
...
+> 
+> ```yaml
+> command: /bin/sh -c 'echo "hello $$HOSTNAME"'
+> ```
+> 
+> If the `entrypoint` (or the image's `ENTRYPOINT`) is configured to invoke a shell, use the exec form syntax for `command` to ensure proper processing. For example:
+>
+> ```yaml
+> command: [ "bundle", "exec", "thin", "-p", "3000" ]
+> ```
```
polarathene, line 276: This would be incorrect.

If `ENTRYPOINT ["/bin/sh", "-c"]`, then `command` should be a string value for `-c`, thus you would need:

```yaml
# Compose will also treat entrypoint config like below as if it were `ENTRYPOINT ["/bin/sh", "-c"]`:
entrypoint: /bin/sh -c
command: ["bundle exec thin -p 3000"]
# or
command: "'bundle exec thin -p 3000'"
# or
command: |
  'bundle exec thin -p 3000'
```

When the image has no entrypoint to run, the equivalent command only form of this is:

```yaml
command: /bin/sh -c 'bundle exec thin -p 3000'
# or
command: ["/bin/sh", "-c", "bundle exec thin -p 3000"]
```

The exec-form / YAML list style is less ambiguous with how Compose will segment the value into exec-form, it technically doesn't have shell-form (_only really exists for `Dockerfile`_).

polarathene (commented):
I am not familiar with the difference of this document file vs the related PR but you may want to ensure they're in sync? The content is diverging here (_and also incorrect in advice_).

polarathene:
> The difference can be found here https://docs.docker.com/reference/compose-file/

No, I was referring to these source files: https://github.com/compose-spec/compose-spec/pull/557/files

At a glance there is overlap, so I wasn't quite familiar with how changes from there make it into the actual Docker docs here.

PR #21859, opened 2025-01-21
fix: invalid command, unrelated output chunks, too large minimal exam…

SalathielGenese:
FIXes in the Docker Bake documentation.

## Description

[x] Commands not matching outputs: _(e.g.: the presence of a group on the CLI affect the group section in the output)_
[x] Some file not being named bring about confusion to the reader _(e.g.: [Using a Target as Build Context](https://docs.docker.com/build/bake/contexts/#using-a-target-as-a-build-context) is not clear about needed two distinct Dockerfiles, or [Using a secondary source directory](https://docs.docker.com/build/bake/contexts/#using-a-secondary-source-directory) about the fact that the first stage in the Dockerfile does not have to be defined at all)_
[x] Multiple confusing relationships between input and output _(e.g.: the [Variables in Bake](https://docs.docker.com/build/bake/variables/) part shows multiple `webapp` target in the outputs whilst the HCL input only defines a `default` target)_

## Reviews

- [ ] Technical review
- [x] Editorial review
- [ ] Product review

### DIFF
--- a/content/manuals/build/bake/variables.md
+++ b/content/manuals/build/bake/variables.md
@@ -15,7 +15,7 @@ environment variables.
 
 Use the `variable` block to define a variable.
 
-```hcl
+```hcl {title=docker-bake.hcl}
 variable "TAG" {
   default = "docker.io/username/webapp:latest"
 }
@@ -23,8 +23,8 @@ variable "TAG" {
 
 The following example shows how to use the `TAG` variable in a target.
 
-```hcl
-target "default" {
+```hcl {title=docker-bake.hcl}
+target "webapp" {
   context = "."
   dockerfile = "Dockerfile"
   tags = [ TAG ]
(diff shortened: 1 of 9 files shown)


content/manuals/build/bake/variables.md
```diff
@@ -233,7 +241,9 @@ $ docker buildx bake -f vars.hcl -f docker-bake.hcl --print app
 ```json
 {
   "target": {
-    "default": {
+  printf("Hello, %s!\n\n", name_0);
+  printf("Hello, %s!\n\n", name_0);
```
dvdksn, line 245:
```suggestion
```

dvdksn (approved):
Thank you

SalathielGenese:
> Thank you

Thanks to you all for making Docker Great Again ^^

PR #21887, opened 2025-01-23
chore: summary bar cleanup

sarahsanders-docker:
## Description
- Updated "Contribute" guide to reference summary bars, added instructions for using them
- Removed `introduced` partial, this is now replaced by the summary bar `requires` value
- Updated all files using `introduced` partial to use summary bar, updated YAML file w/ feature name and info

## Related issues or tickets
[ENGDOCS-2392](https://docker.atlassian.net/browse/ENGDOCS-2392)

## Reviews
- [ ] Technical review
- [ ] Editorial review

[ENGDOCS-2392]: https://docker.atlassian.net/browse/ENGDOCS-2392?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


dvdksn:
I'd much rather have just one way of showing this data. I thought we ended up with a shortcode for the summary bar because it allows us the flexibility to put it on a section-level? If we only ever want to render this component at the top of a page, then it probably shouldn't even be a shortcode. Pretty sure we discussed this when reviewing the original PR.

sarahsanders-docker:
> I'd much rather have just one way of showing this data. I thought we ended up with a shortcode for the summary bar because it allows us the flexibility to put it on a section-level? If we only ever want to render this component at the top of a page, then it probably shouldn't even be a shortcode. Pretty sure we discussed this when reviewing the original PR.

I would also prefer to have one way of showing this data. I think if we introduce too many types of callouts or visual elements that represent a bunch of different things, we get into a territory that becomes difficult to maintain and be consistent with. Also as a user, the more consistent we are, the better I recognize and understand these elements over time.

For these visual elements, I would prefer keeping things limited to summary bars and callouts, and implementing them at the section level as needed. Also, these are implemented at the section level in other docs from the original PR. 

@aevesdocker I can adjust the text shown in the `Requires:` section of the summary bar to include release notes. that is an easy fix and I can update the contribute guide to have that be standard when there are corresponding release notes

sarahsanders-docker:
@aevesdocker @dvdksn I am able to add the release notes to the summary bar `Requires:` line by slightly adjusting the shortcode file. If this is a good compromise, I can push this change! 
image

dvdksn (commented):
You can also remove these, they were used by the introduced.html shortcode

- https://github.com/docker/docs/blob/ee59422b2d859fdb014d8c28ced0fefea593a84d/i18n/en.yaml#L39-L54
- https://github.com/docker/docs/blob/ee59422b2d859fdb014d8c28ced0fefea593a84d/hugo.yaml#L130-L139

dvdksn (approved):

PR #21916, opened 2025-01-29
Update documentation regarding `target` value in compose file sample

Biplab-Dutta:
## Description
The `WORKDIR` in the Dockerfile was set to `/app`, but the `target` in the `compose.yaml` file was incorrectly set to `/src/web` instead of `/app/web`.

## Reviews
- [x] Technical review
- [ ] Editorial review
- [ ] Product review


Biplab-Dutta:
Closing the PR as I think the doc is fine

PR #21917, opened 2025-01-29
ENGDOCS-2405a

aevesdocker:
## Description

Add new DD CLI command for 4.38

## Reviews

- [ ] Technical review
- [ ] Editorial review
- [ ] Product review


data/summary.yaml
```diff
@@ -45,6 +45,8 @@ Docker Desktop Archlinux:
 Docker Desktop CLI:
   availability: Beta
   requires: Docker Desktop 4.37 and later
+Docker Desktop CLI update:
+  requires: Docker Desktop 4.38 and later
```
sarahsanders-docker, line 49: if you want to add release notes here (not sure if you want to or not!) you can use markdown around the version number. `Docker Desktop [4.38](release-notes-url) and later`

sarahsanders-docker (dismissed):

data/summary.yaml
dvdksn: I wonder if "Docker Desktop CLI update" will be easy to remember what it means in the future? Is there a better/more semantic feature id that we could use? (This is probably a good idea generally - I have not reviewed the feature IDs in this file so far, but if they are too generic, it'll be difficult to keep track of them down the line. How do we manage this inventory of feaures?)

dvdksn: yeah something like that

sarahsanders-docker (approved):

PR #21929, opened 2025-01-30
Improve security documentation with warning around windows containers

twelsh-aw:
## Description

Some background dialog between security, desktop, moby teams and some security researchers. At the present time, this is an accepted risk in Docker Desktop installations for Windows and should be clarified in better detail.

## Related issues or tickets

PSEC-1839

## Reviews

- [] Technical review
- [ ] Editorial review
- [ ] Product review


content/manuals/desktop/setup/install/windows-install.md
```diff
@@ -210,7 +210,7 @@ By default, Docker Desktop is installed at `C:\Program Files\Docker\Docker`.
 The `install` command accepts the following flags:
 - `--quiet`: Suppresses information output when running the installer 
 - `--accept-license`: Accepts the [Docker Subscription Service Agreement](https://www.docker.com/legal/docker-subscription-service-agreement) now, rather than requiring it to be accepted when the application is first run
-- `--no-windows-containers`: Disables the Windows containers integration
+- `--no-windows-containers`: Disables the Windows containers integration. This can have security implication. For more information, see [Windows containers](/manuals/desktop/setup/install/windows-permission-requirements.md#windows-containers).
```
gabriellavengeo, line 213: This phrasing makes it a bit unclear that *not* having it is the less secure thing. Maybe we should rephrase but not sure what's the best wording

twelsh-aw: Agreed. That seems clearer. Will go with  'This can improve security.'

Longer term, we do have plans to tweak these options and make the secure option the default behaviour. So this is all interim text anyways.

gabriellavengeo (approved):

PR #21953, opened 2025-02-03
Fix build example diff in multi-platform.md

mpesari:
## Description

The build example "Diff" tab at https://docs.docker.com/build/building/multi-platform/#cross-compiling-a-go-application did not correctly highlight the updated build command.


crazy-max (approved):
LGTM thanks!

PR #21966, opened 2025-02-04
Fix argument name to docker network create

betelgeuse:
## Description

```
$ docker network create --help | grep opt
      --ipam-opt map         Set IPAM driver specific options (default map[])
  -o, --opt map              Set driver specific options (default map[])
```

So the documentation should be changed from using --option to --opt.


thaJeztah (approved):
LGTM, thanks!

PR #21988, opened 2025-02-07
typo in local cache storage backend

AzureLeMoon:
## Description

there is a typo when describing the usage of `digest` on importing caches specifying the flag as `--cache-to` which is wrong.

this is a simple change modifying the mentioned flag to `--cache-from`


crazy-max (approved):
LGTM thanks!

PR #22016, opened 2025-02-12
quick fix: add tags to build code sample

sarahsanders-docker:
## Description
- User reported the following error `Error: buildx failed with: ERROR: tag is needed when pushing to registry` when using our code sample in [https://docs.docker.com/build/ci/github-actions/multi-platform/#distribute-build-across-multiple-runners](https://docs.docker.com/build/ci/github-actions/multi-platform/#distribute-build-across-multiple-runners)
- Added `:latest` tags to code sample, this is the most general approach and not version specific, which will work for most users who are just trying things out using our code samples

## Related issues or tickets
- https://github.com/docker/docs/issues/22014

## Reviews
- [ ] Technical review

### DIFF
--- a/content/manuals/build/ci/github-actions/multi-platform.md
+++ b/content/manuals/build/ci/github-actions/multi-platform.md
@@ -175,7 +175,7 @@ jobs:
         with:
           platforms: ${{ matrix.platform }}
           labels: ${{ steps.meta.outputs.labels }}
-          outputs: type=image,"name=${{ env.DOCKERHUB_REPO }},${{ env.GHCR_REPO }}",push-by-digest=true,name-canonical=true,push=true
+          outputs: type=image,"name=${{ env.DOCKERHUB_REPO }}:latest,${{ env.GHCR_REPO }}:latest",push-by-digest=true,name-canonical=true,push=true
 
       - name: Export digest
         run: |


content/manuals/build/ci/github-actions/multi-platform.md
```diff
@@ -175,7 +175,7 @@ jobs:
         with:
           platforms: ${{ matrix.platform }}
           labels: ${{ steps.meta.outputs.labels }}
-          outputs: type=image,"name=${{ env.DOCKERHUB_REPO }},${{ env.GHCR_REPO }}",push-by-digest=true,name-canonical=true,push=true
+          outputs: type=image,"name=${{ env.DOCKERHUB_REPO }}:latest,${{ env.GHCR_REPO }}:latest",push-by-digest=true,name-canonical=true,push=true
```
crazy-max, line 178: No we should not set a tag here, only repo name as we are pushing by digest with `push-by-digest=true` and merging these to create tags in `merge` job. If that fails that means a registry doesn't support push by digest which looks strange to me. I will ask to user habing this issue.

KonradHoeffner: I indeed tried adding tags at this step but this leads to strange results as then randomly one of the platform images will have e.g. the latest tag or a semver tag like 0.3.9, which then makes it impossible to use those tags on the other architectures.

KonradHoeffner: Added the link to the workflow and the full workflow file to the issue.

PR #22036, opened 2025-02-14
update doc for apt

epicseven-cup:
## Description
updated apt to apt-get in ubuntu install

## Related issues or tickets

22035

## Reviews

- [ ] Technical review
- [x] Editorial review
- [ ] Product review

### DIFF
--- a/content/manuals/engine/install/ubuntu.md
+++ b/content/manuals/engine/install/ubuntu.md
@@ -88,10 +88,10 @@ conflicts with the versions bundled with Docker Engine.
 Run the following command to uninstall all conflicting packages:
 
 ```console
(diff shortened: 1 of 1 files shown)


thaJeztah:
Thank you for contributing!

We use `apt-get` instead of `apt` in these instructions so that  the example commands can be used by users who want to script these steps (which in some cases can be as part of a Dockerfile). While `apt` is intended for "interactive" use, `apt-get` is still the recommended tool for scripting (as it has a stable API, whereas `apt` does not).

Some earlier discussions on this can be found in https://github.com/docker/docker.github.io/pull/8135, https://github.com/docker/docker.github.io/pull/10547, and probably some other ones.


I'm closing this PR because of the above, but thanks again for contributing, and feel free to continue the conversation.

PR #22086, opened 2025-02-23
Updates for moby 28.0.1

robmry:
## Description

- Describe new iptables chain `DOCKER-FORWARD`, which splits Docker's rules out of the main `FORWARD` chain where they were being antisocial - related to https://github.com/moby/moby/pull/49518
- Update notes about IP Forwarding and the default DROP policy
- Only engine < 28.0.0 allows remote access to ports published to the localhost address

## Reviews

- [ ] Technical review
- [ ] Editorial review
- [ ] Product review


thaJeztah:
Just a quick check, because I tend to zone out on iptables rules 😂 - with the rewrite, the docs are still usable for those who didn't update to v28.0.0 / v28.0.1, or is there a differentiation we need to make ("if you're running < v28.0, then ....")?

I know we try to keep our docs reflect "current version" in most cases, but just in case it's relevant for people who are not yet on latest.

robmry:
> Just a quick check, because I tend to zone out on iptables rules 😂 - with the rewrite, the docs are still usable for those who didn't update to v28.0.0 / v28.0.1, or is there a differentiation we need to make ("if you're running < v28.0, then ....")?
> 
> I know we try to keep our docs reflect "current version" in most cases, but just in case it's relevant for people who are not yet on latest.

I think it's ok ... in the third commit I left in the warning about remote access to ports published to 127.0.0.1 ("In releases older than 28.0.0 ..."). And, in the second commit I noted the change in setting the FORWARD policy to DROP (with another "In releases older than 28.0.0, ..." warning).

PR #22130, opened 2025-02-28
feat: support running vale locally

mdelapenya:
- **chore: add a make goal to run the validations**
- **feat: add a make goal to run vale locally**

## Description
This PR adds two Make goals:

- make vale, which runs vale as a container using the official Vale Docker image.
- make validate, which runs `docker buildx bake validate`

The first one allows contributors to detect grammar errors while working on their changes, and the second is an alias for the build command.

The intention here is to enhance the local development experience when contributing to the docs, not having to wait for the CI to understand the vale errors, but instead making it possible to run it locally.

## Reviews

- [x] Technical review
- [ ] Editorial review
- [ ] Product review


Makefile
```diff
@@ -1,3 +1,22 @@
...
+.PHONY: vale
+vale: ## run vale
+	docker run --rm -v $(PWD):/docs \
+		-w /docs \
+		-e PIP_BREAK_SYSTEM_PACKAGES=1 \
+		$(VALE_IMAGE) $(CONTENT_DIR)
+
+.PHONY: validate
+validate: ## run validations
+	docker buildx bake validate
```
crazy-max, line 18: Not sure what the prupose of this change is but don't think that makes sense to have a make goal that just runs bake tbh. This would also not work on Windows host.

crazy-max (commented):
Can we move to another PR the change related to new `validate` make goal? Doesn't relate to "feat: support running vale locally" and we should not enumerate unrelated change requests in the same PR tbh.

Makefile
mdelapenya: You're right. I'll split the changes, and probably discard the validate one in a follow-up commit.

Regarding the vale goal, anything to highlight?

Makefile
```diff
@@ -1,3 +1,18 @@
...
+
+# Docker image to use for vale.
+VALE_IMAGE := jdkato/vale:latest
+
+.PHONY: vale
+vale: ## run vale
+	docker run --rm -v $(PWD):/docs \
+		-w /docs \
+		-e PIP_BREAK_SYSTEM_PACKAGES=1 \
+		$(VALE_IMAGE) $(CONTENT_DIR)
```
crazy-max, line 14: I was wondering if we could have it running within our Dockerfile so it could be used with bake like `lint`:
* https://github.com/docker/docs/blob/9623874e7eeadf0d892a1f64f43a170ebe1b9d22/docker-bake.hcl#L48-L52
* https://github.com/docker/docs/blob/9623874e7eeadf0d892a1f64f43a170ebe1b9d22/Dockerfile#L53-L60

Then we could just add a `vale` bake target to `validate` group? https://github.com/docker/docs/blob/9623874e7eeadf0d892a1f64f43a170ebe1b9d22/docker-bake.hcl#L38-L40

And would probably be able to remove this ci job https://github.com/docker/docs/blob/9623874e7eeadf0d892a1f64f43a170ebe1b9d22/.github/workflows/build.yml#L77-L86

And just add `vale` to this matrix https://github.com/docker/docs/blob/9623874e7eeadf0d892a1f64f43a170ebe1b9d22/.github/workflows/build.yml#L93-L99 so we have a good sandbox that work both on ci and locally?

mdelapenya: I need to crate a foundational knowledge on bake yet, but it seems reasonable to include the target as part of the current build system. I'd go with what you suggest

ArthurFlag:
Looks like this PR was tackled by #23101.

PR #22220, opened 2025-03-10
Update image source for kafbat

mikesir87:
## Description

Kafbat publishes its images to Docker Hub, so changing the registry source to Hub instead of GHCR. Also, instead of using the latest tag, they publish merged changes onto a `main` tag.

This will also help our AI agent to use the correct image source, as it is currently using the GHCR-based image.


sarahsanders-docker (approved):
lgtm!!

PR #22285, opened 2025-03-20
Update start.md

windmillcode0:
Let the reader know that the executable may not be on the PATH and needs to be manually added

## Reviews

- [ ] Technical review
- [ ] Editorial review
- [ ] Product review


thaJeztah:
Closing for now per my earlier comment

windmillcode0:
Greetings we never saw a response to the comment "These are not standard locations to install the daemon in; I'm not sure if we need to mention such cases.

Do you have more context in what kind of setup dockerd would be installed in such a location?" its not abaout the daemon its about the docker executables, and there are isntalled here on windows as per our installation

PR #22368, opened 2025-04-04
Update services.md

chubi-x:
Explicitly state the dangers if a port mapping binds to all interfaces

## Description

We recently discovered that docker was bypassing our firewall rules when forwarding ports from a container using the standard `<host port>:<container port>` syntax. What this meant was that the container was effectively visible to the entire internet. It was only after some digging did we discover that it is possible and even recommended to explicitly bind the host port to localhost so it doesn't accept connections from everywhere. This PR updates the docs to explicitly state the potential dangers.

## Reviews

- [ ] Technical review
- [ ] Editorial review
- [ ] Product review

### DIFF
--- a/content/reference/compose-file/services.md
+++ b/content/reference/compose-file/services.md
@@ -1572,6 +1572,11 @@ in the form:
 - `CONTAINER` is `port | range`.
 - `PROTOCOL` restricts ports to a specified protocol either `tcp` or `udp`(optional). Default is `tcp`.
 
+>[!CAUTION]
+>
+>If you do not specify an IP address (such as 127.0.0.1) and it binds to all interfaces then any machine on the same network could
+>potentially have access to the container. This could be especially dangerous if the container is exposed to the internet.
+
 Ports can be either a single value or a range. `HOST` and `CONTAINER` must use equivalent ranges. 
 
 You can either specify both ports (`HOST:CONTAINER`), or just the container port. In the latter case,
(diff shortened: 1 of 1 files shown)


content/reference/compose-file/services.md
```diff
@@ -1629,6 +1629,11 @@ in the form:
 - `CONTAINER` is `port | range`.
 - `PROTOCOL` restricts ports to a specified protocol either `tcp` or `udp`(optional). Default is `tcp`.
 
+>[!CAUTION]
```
aevesdocker:
```suggestion
> [!WARNING]
```

content/reference/compose-file/services.md
```diff
@@ -1629,6 +1629,11 @@ in the form:
...
+>
+>If you do not specify an IP address (such as 127.0.0.1) and it binds to all interfaces then any machine on the same network could
```
aevesdocker:
```suggestion
> If you do not specify a host IP (such as `127.0.0.1`), Docker binds to all interfaces (`0.0.0.0`), bypassing host firewall rules. This can expose the container directly to the internet if the host has a public IP address. For more information, see [Port publishing and mapping](/manuals/engine/network/port-publishing.md).
```

content/reference/compose-file/services.md
```diff
@@ -1629,6 +1629,11 @@ in the form:
...
+>potentially have access to the container. This could be especially dangerous if the container is exposed to the internet.
```
aevesdocker:
```suggestion

```

aevesdocker:
@chubi-x could you rebase and then I can get this merged

dvdksn:
@aevesdocker I rebased the PR - we should be good to get this in 👍

PR #22373, opened 2025-04-05
Remove `--kernel-memory` option, description and details from resource_constraints.md

zzoe2346:
## Description

According to the https://docs.docker.com/engine/deprecated/#kernel-memory-limit, the `--kernel-memory` option is no longer supported(Removed Status). 

SCR-20250405-ngdv

However, I noticed that resource_constraints.md still includes this option. I believe removing this outdated information will improve the accuracy and completeness of the documentation, so I'm submitting this PR to address it.

Thank you.

## Related issues or tickets

No related issues or ticket.

## Reviews

- [ ] Technical review
- [ ] Editorial review
- [ ] Product review


thaJeztah:
Thanks! I need to double check, because it looks like we still provide the option on the CLI (but this could've been to allow it to work when connecting to an older version of the engine); possibly should be hidden by default, or removed altogether.

PR #22396, opened 2025-04-09
fix go version in go.mod

crazy-max:
## Description

We should keep minimal version in go.mod and avoid bumping patch.

## Reviews

- [ ] Technical review
- [ ] Editorial review
- [ ] Product review

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/docker/docs
 
-go 1.23.6
+go 1.23.0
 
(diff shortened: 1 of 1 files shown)


crazy-max:
Argh actually we can't because of Compose go.mod strategy https://github.com/docker/compose/blob/482b62228233075fe79c355d41061132e6fd983b/go.mod#L3 :disappointed:

PR #22454, opened 2025-04-18
ISSUE-22453: Update Debian, RPI, Ubuntu apt config

jreslock:
## Description

The installation instructions for Debian, Ubuntu and Raspberry PI all use a complicated `$(. /etc/os-release && echo "VERSION_CODENAME")` command to set the VERSION_CODENAME for the apt source configuration. It would be much simpler and straightforward to use `awk` to get this from the target file and pass to the subsequent `tee` command.

This commit updates the installation instructions for each of the deb-based operating systems to use `awk -F '/VERSION_CODENAME' { print $2 } /etc/os-release` instead of the current, somewhat confusing and complicated sourcing of the file and subsequent `echo` to produce the same result using a single command invocation and built-in utility (awk).

## Related issues or tickets

I created an issue for this before creating this PR.

https://github.com/docker/docs/issues/22453

## Reviews

- [X] Technical review
- [X] Editorial review
- [ ] Product review

### DIFF
--- a/content/manuals/engine/install/debian.md
+++ b/content/manuals/engine/install/debian.md
@@ -116,7 +116,7 @@ Docker from the repository.
    # Add the repository to Apt sources:
    echo \
      "deb [arch=$(dpkg --print-architecture) signed-by=/etc/apt/keyrings/docker.asc] {{% param "download-url-base" %}} \
-     $(. /etc/os-release && echo "$VERSION_CODENAME") stable" | \
+     $(awk -F= '/VERSION_CODENAME/ { print $2 }' /etc/os-release) stable" | \
      sudo tee /etc/apt/sources.list.d/docker.list > /dev/null
    sudo apt-get update
    ```
--- a/content/manuals/engine/install/raspberry-pi-os.md
+++ b/content/manuals/engine/install/raspberry-pi-os.md
@@ -117,7 +117,7 @@ Docker from the repository.
    # Add the repository to Apt sources:
    echo \
      "deb [arch=$(dpkg --print-architecture) signed-by=/etc/apt/keyrings/docker.asc] {{% param "download-url-base" %}} \
-     $(. /etc/os-release && echo "$VERSION_CODENAME") stable" | \
+     $(awk -F= '/VERSION_CODENAME/ { print $2 }' /etc/os-release) stable" | \
      sudo tee /etc/apt/sources.list.d/docker.list > /dev/null
    sudo apt-get update
    ```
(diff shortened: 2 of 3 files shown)


content/manuals/engine/install/debian.md
```diff
@@ -116,7 +116,7 @@ Docker from the repository.
    # Add the repository to Apt sources:
    echo \
      "deb [arch=$(dpkg --print-architecture) signed-by=/etc/apt/keyrings/docker.asc] {{% param "download-url-base" %}} \
-     $(. /etc/os-release && echo "$VERSION_CODENAME") stable" | \
+     $(awk -F= '/VERSION_CODENAME/ { print $2 }' /etc/os-release) stable" | \
```
thaJeztah, line 119: Hm.. not sure I understand the change; how is this simpler than a `source` and `echo` ? The update also requires `awk` to be present, which would be in "most" setups, but is not guaranteed.

thaJeztah (changes requested):
not sure if we should change this, but happy to hear if there's additional motivations here

content/manuals/engine/install/debian.md
jreslock: Hi @thaJeztah! Thank you for reviewing and commenting.

I submitted this PR and related issue because I was having some trouble using these commands in a Dockerfile to install the docker-compose-plugin inside of a devcontainer. My "fix" to use `awk` seemed to work and I thought it would be helpful for others. I have re-tested locally using the original `$(. /etc/os-release && echo "$VERSION_CODENAME")` which is now working fine for me. I very likely had some minor syntax issue causing my initial problem. 

If this change doesn't make sense or is undesired I will not be offended/upset.

thaJeztah: Yeah, looking a bit further, this change is definitely not right as it would break if these would have a quote, or otherwise;

```bash
awk -F= '/PRETTY_NAME/ { print $2 }' /etc/os-release
"Debian GNU/Linux trixie/sid"

. /etc/os-release && echo $PRETTY_NAME
Debian GNU/Linux trixie/sid
```

[os-release(5)](https://man7.org/linux/man-pages/man5/os-release.5.html) man page also describes sourcing to be the correct way; parsing in other languages is possible, but requires much more handling to make sure the values are properly consumed.

jreslock:
Based on @thaJeztah's comments, this change is not viable or desirable so I am closing the PR and will close the related issue.

thaJeztah:
Thanks!

PR #22474, opened 2025-04-22
docs: Added section explaining how RAM configurations restricting Docker hub interacts with mirror registries

chaomonica:
## Description

If a user has configured a mirror registry but has restricted Docker Hub in their RAM settings, they will get an error `Access to docker.io has been restricted by your administrators.` even if that image has already been cached in their registry mirror.

## Related issues or tickets

SEG-1124 (Slack thread linked in ticket for context)

## Reviews

Please do not merge until after technical review! 

- [X] Technical review @karman-docker I have tested the commands in the following section.
- [X] Editorial review
- [ ] Product review

### DIFF
--- a/content/manuals/docker-hub/image-library/mirror.md
+++ b/content/manuals/docker-hub/image-library/mirror.md
@@ -45,6 +45,23 @@ Hub can be mirrored.
 The Registry can be configured as a pull through cache. In this mode a Registry
 responds to all normal docker pull requests but stores all content locally.
 
+### Using Registry Access Management (RAM) with a registry mirror
+
+If Docker Hub access is restricted via your Registry Access Management (RAM) configuration, you will not be able to pull images originating from Docker Hub even if the images are available in your registry mirror.
+
+You may encounter an error like:
+```bash
+Error response from daemon: Access to docker.io has been restricted by your administrators.
+```
+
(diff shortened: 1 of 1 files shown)


content/manuals/docker-hub/image-library/mirror.md
```diff
@@ -45,6 +45,23 @@ Hub can be mirrored.
...
+### Using Registry Access Management (RAM) with a registry mirror
+
+If Docker Hub access is restricted via your Registry Access Management (RAM) configuration, you will not be able to pull images originating from Docker Hub even if the images are available in your registry mirror.
+
+You will encounter the following error:
+```console
+Error response from daemon: Access to docker.io has been restricted by your administrators.
+```
+
+This happens because RAM restrictions are enforced at the API proxy layer, meaning that the request is blocked before Docker Desktop even attempts to pull from a registry mirror or fall back to Docker Hub.
```
karman-docker, line 57: @chaomonica we can remove this part (line 57), as it is an implementation details, end user need not know about it.

karman-docker: Or we can add "The registry mirror needs access to Docker hub to verify if the cached image is the latest as compared to the one on Docker hub."

karman-docker (approved):

PR #22493, opened 2025-04-25
docs: update uninstall.md to fix Arch uninstall commands

j-tag:
## Description

Arch uninstall commands are invalid. there is no `pacman remove`, maybe they wanted to write `pacman --remove`. but its better to use `pacman -Rsn` as it will also purge files. and down there there is an `apt purge` which is also not available in Arch linux.
 

## Reviews

- [ ] Technical review
- [ ] Editorial review
- [ ] Product review


thaJeztah:
Thanks for contributing; looks like a pull request was merged that also addressed this; https://github.com/docker/docs/pull/22458

PR #22602, opened 2025-05-07
Remote access to container ports

robmry:
## Description

(These updates are for current moby releases.)

### Update warning about ports published to 127.0.0.1

Align a warning on the [engine/network](https://docs.docker.com/engine/network/) page with the one on [engine/network/packet-filtering-firewalls](https://docs.docker.com/engine/network/packet-filtering-firewalls/) - which has already been updated, in https://github.com/docker/docs/commit/2090e5bcbbdf38fa1004caa0488e555497ae8a39

### Update text about direct routing

Update the text about direct routing, to note that without direct routing it's not (any longer) possible to access ports on container addresses directly.

## Reviews

- [ ] Technical review
- [ ] Editorial review
- [ ] Product review

### DIFF
--- a/content/manuals/engine/network/_index.md
+++ b/content/manuals/engine/network/_index.md
@@ -160,8 +160,8 @@ Here are some examples:
 >
 > > [!WARNING]
 > >
-> > Hosts within the same L2 segment (for example, hosts connected to the same
-> > network switch) can reach ports published to localhost.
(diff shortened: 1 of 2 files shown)


thaJeztah (approved):
LGTM

austinvazquez (approved):

PR #22616, opened 2025-05-12
fix: use correct variable metrics-addr

ranomier:
rename metrics-address to metrics-addr

## Description

just a little correction on variable naming

## Reviews

- [ ] Technical review
- [ ] Editorial review
- [ ] Product review


thaJeztah (changes requested):
I don't think this change is correct, but perhaps you have more context?

PR #22620, opened 2025-05-12
Update: 4GB -> 8GB minimum specs

LilaHexe0:
## Description

4GB isn't really enough to do anything on windows, let alone run docker on top of that, so I've updated it to 8GB.

## Reviews

- [ ] Technical review
- [ ] Editorial review
- [ ] Product review


aevesdocker (approved):
Agreed! Thanks!

PR #22626, opened 2025-05-13
Update outdated url to ubuntu source code

enoot:
## Description

I set a new URL for the source code because the old ones refer to the commit of the file, which is not used now, because Ubuntu for Docker is built from livecd scripts.

## Reviews

- [ ] Technical review
- [x] Editorial review
- [ ] Product review

### DIFF
--- a/content/manuals/build/building/best-practices.md
+++ b/content/manuals/build/building/best-practices.md
@@ -442,7 +442,7 @@ reduces the image size, since the apt cache isn't stored in a layer. Since the
 `RUN` statement starts with `apt-get update`, the package cache is always
 refreshed prior to `apt-get install`.
 
-Official Debian and Ubuntu images [automatically run `apt-get clean`](https://github.com/moby/moby/blob/03e2923e42446dbb830c654d0eec323a0b4ef02a/contrib/mkimage/debootstrap#L82-L105), so explicit invocation is not required.
+Official Debian and Ubuntu images [automatically run `apt-get clean`](https://git.launchpad.net/livecd-rootfs/tree/live-build/functions#n1003), so explicit invocation is not required.
 
 #### Using pipes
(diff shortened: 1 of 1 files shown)


crazy-max (approved):
LGTM thanks

@tianon @neersighted Can you PTAL? :pray:

content/manuals/build/building/best-practices.md
```diff
@@ -442,7 +442,7 @@ reduces the image size, since the apt cache isn't stored in a layer. Since the
 `RUN` statement starts with `apt-get update`, the package cache is always
 refreshed prior to `apt-get install`.
 
-Official Debian and Ubuntu images [automatically run `apt-get clean`](https://github.com/moby/moby/blob/03e2923e42446dbb830c654d0eec323a0b4ef02a/contrib/mkimage/debootstrap#L82-L105), so explicit invocation is not required.
+Official Debian and Ubuntu images [automatically run `apt-get clean`](https://git.launchpad.net/livecd-rootfs/tree/live-build/functions#n1003), so explicit invocation is not required.
```
tianon, line 445: https://github.com/debuerreotype/debuerreotype/blob/c9542ab785e72696eb2908a6dbc9220abbabef39/scripts/debuerreotype-minimizing-config#L87-L109 would be a better link (especially as it has more explanation; I tried to convince the Ubuntu folks to copy/use it, but I wasn't convincing enough).

```suggestion
Official Debian and Ubuntu images [automatically run `apt-get clean`](https://github.com/debuerreotype/debuerreotype/blob/c9542ab785e72696eb2908a6dbc9220abbabef39/scripts/debuerreotype-minimizing-config#L87-L109), so explicit invocation is not required.
```

crazy-max (approved):

PR #22629, opened 2025-05-13
update go to 1.24

crazy-max:
## Description

Update go to 1.24. Also removes toolchain directive that should not be used as we are doing sandboxed builds.

## Reviews

- [x] Technical review
- [ ] Editorial review
- [ ] Product review


thaJeztah (approved):
LGTM

thaJeztah:
One thing I recently learned (or perhaps knew, but forgot about) was that `go-setup-action` does NOT set "go toolchain" to "local", so regardless what version you ask it to install, it may still download a different version when running 🫠 🫠 🫠

crazy-max:
> One thing I recently learned (or perhaps knew, but forgot about) was that `go-setup-action` does NOT set "go toolchain" to "local", so regardless what version you ask it to install, it may still download a different version when running 🫠 🫠 🫠

Ah good to know!

PR #22632, opened 2025-05-14
feat: add information about fluentd tls connection

enoot:
## Description

I added information about TLS support for Fluentd log driver connection
https://github.com/moby/moby/blob/294f0c36e468dc097d56dcf88d4418b7dc3d2856/vendor/github.com/fluent/fluent-logger-golang/fluent/fluent.go#L443

## Related issues or tickets

- relates to https://github.com/moby/moby/pull/42979
- relates to https://github.com/fluent/fluent-logger-golang/pull/107 
- relates to https://github.com/fluent/fluent-logger-golang/issues/112

## Reviews

- [x] Technical review
- [x] Editorial review
- [ ] Product review

### DIFF
--- a/content/manuals/engine/logging/drivers/fluentd.md
+++ b/content/manuals/engine/logging/drivers/fluentd.md
@@ -84,11 +84,12 @@ Users can use the `--log-opt NAME=VALUE` flag to specify additional Fluentd logg
 ### fluentd-address
 
 By default, the logging driver connects to `localhost:24224`. Supply the
-`fluentd-address` option to connect to a different address. `tcp`(default) and `unix` sockets are supported.
+`fluentd-address` option to connect to a different address. `tcp`(default), `tls` and `unix` sockets are supported.
 
 ```console
 $ docker run --log-driver=fluentd --log-opt fluentd-address=fluentdhost:24224
 $ docker run --log-driver=fluentd --log-opt fluentd-address=tcp://fluentdhost:24224
+$ docker run --log-driver=fluentd --log-opt fluentd-address=tls://fluentdhost:24224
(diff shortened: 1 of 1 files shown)


thaJeztah:
Thanks! Looks indeed that the updated dependency from https://github.com/moby/moby/pull/42979 add this as option (as part of docker engine v23.0 and up).

Looking a bit further, I see that a colleague (@akerouanton) left a comment in the PR that implemented it in the dependency; mentioning that it was "incomplete"; https://github.com/fluent/fluent-logger-golang/pull/107#issuecomment-984184421

> @fujimotos @zonito Although this PR allows using self-signed TLS certs, it doesn't offer proper way of securely doing TLS because there's no way to specify which CA cert is accepted. With this change, if an attacker can mitm/redirect fluentd's network stream, they could still generate their own self-signed cert and bypass TLS promises. IMHO it could give a false sense of security. It would be great to be able to specify which CA certs are accepted (on top of what's in system truststore).

- ☝️ tracking ticket for that is in https://github.com/fluent/fluent-logger-golang/issues/112


So, wondering if it was _intentionally_ left undocumented (because of that), or if it's something we should (somewhere?) call out in the docs, in case users try to use the option and run into issues because of that limitation.

cc @akerouanton @vvoland for thoughts

thaJeztah:
Thanks! Yes, so sounds like it may be a good idea to have some warning / note / callout to outline the limitations / caveats. That way the user can make a more well-informed decision whether that's acceptable for their use.

At least, I generally prefer not trying to shove such things under the rug (even if they're ugly)

Suggestions for that definitely welcome! Also /cc @ArthurFlag  (in case you have any! ❤️ )

PR #22635, opened 2025-05-14
engine: add new supported versions

ArthurFlag:
## Description

Add latest supported versions.


thaJeztah (approved):
LGTM, thanks!

thaJeztah:
I just recalled we also started building for CentOS 10, but it's possible our repository is missing some redirects for "minor" versions; https://download.docker.com/linux/centos/10/

It's probably OK to add CentOS 10 (stream) to the CentOS page though.

Wondering if we should also have corresponding RHEL 10 packages (which I think was also out) /cc @neersighted

PR #22670, opened 2025-05-19
get started: add troubleshooting callout

sarahsanders-docker:
## Description
- Adds a callout in the workshop guide for login issues

## Related issues or tickets
https://github.com/docker/docs/issues/22615

## Reviews
- [ ] Editorial review

### DIFF
--- a/content/get-started/workshop/04_sharing_app.md
+++ b/content/get-started/workshop/04_sharing_app.md
@@ -52,6 +52,10 @@ In the following image, you can see an example Docker command from Docker Hub. T
 
     To fix this, you need to tag your existing image you've built to give it another name.
 
+   > [!NOTE]
+   >
+   > If you encounter the following error: `denied: requested access to the resource is denied`, you can retry with `$ docker login -u "myusername" -p "mypassword" docker.io`.
+
 2. Sign in to Docker Hub using the command `docker login -u YOUR-USER-NAME`.
 
 3. Use the `docker tag` command to give the `getting-started` image a new name. Replace `YOUR-USER-NAME` with your Docker ID.


content/get-started/workshop/04_sharing_app.md
```diff
@@ -52,6 +52,10 @@ In the following image, you can see an example Docker command from Docker Hub. T
 
     To fix this, you need to tag your existing image you've built to give it another name.
 
+   > [!NOTE]
+   >
+   > If you encounter the following error: `denied: requested access to the resource is denied`, you can retry with `$ docker login -u "myusername" -p "mypassword" docker.io`.
```
thaJeztah, line 57: We try to avoid using `-p` because that would include the user's passphrase in shell history; the `docker.io` should not be needed (as it's the default).

But looking at the steps, it looks like step 2 below is to log in; should we make step 2 the first step? I think that's the problem here; we try to push before making sure the user is logged in.

content/get-started/workshop/04_sharing_app.md
```diff
@@ -52,6 +52,10 @@ In the following image, you can see an example Docker command from Docker Hub. T
...
+
 2. Sign in to Docker Hub using the command `docker login -u YOUR-USER-NAME`.
```
thaJeztah, line 59: Perhaps we should remove the `-u` as well, as that skips the OAuth workflow. Logging in without the `-u` option will ask the user for their username, so we may need to call that out (type `docker login` and enter your username .. )

content/get-started/workshop/04_sharing_app.md
```diff
@@ -38,7 +38,8 @@ In the following image, you can see an example Docker command from Docker Hub. T
 
 ## Push the image
 
-1. In the command line, run the `docker push` command that you see on Docker
+1. Sign in to Docker Hub using the command: `docker login YOUR-USER-NAME YOUR-PASSWORD`.
```
thaJeztah, line 41:
```suggestion
1. Sign in to Docker Hub using the command: `docker login`.
```

Slightly tricky; depending on the situation, this will ask for a username (and password), or (with current versions of the CLI) will start a web-based OAuth flow;

```bash
docker login

USING WEB-BASED LOGIN

i Info → To sign in with credentials on the command line, use 'docker login -u <username>'


Your one-time device confirmation code is: VQXV-KGKD
Press ENTER to open your browser or submit your device code here: https://login.docker.com/activate

Waiting for authentication in the browser…
```

So, ... perhaps just `docker login` would be enough?

content/get-started/workshop/04_sharing_app.md
```diff
@@ -52,8 +53,6 @@ In the following image, you can see an example Docker command from Docker Hub. T
 
     To fix this, you need to tag your existing image you've built to give it another name.
 
-2. Sign in to Docker Hub using the command `docker login -u YOUR-USER-NAME`.
-
 3. Use the `docker tag` command to give the `getting-started` image a new name. Replace `YOUR-USER-NAME` with your Docker ID.
```
thaJeztah, line 56: Oh, but wait! Reading the (what was) first step; that one was _intentionally_ failing? Looks like this is trying to describe that you cannot push to `docker/getting-started` (either because the image was not tagged yet, or if it was, because the reader won't have permissions to push to their repository (which is expected)).


So was the problem we tried to solve ... not a problem (just a somewhat confusing first step)?

sarahsanders-docker: that explains a lot :'') I misunderstood the ask. my next commit fixes step one to make it clear that this is supposed to fail.

ArthurFlag (approved):

thaJeztah (approved):
LGTM, thanks!

sorry for the back-and-forth; it took me a bit as well to realise the issue!

sarahsanders-docker:
No worries @thaJeztah! Thanks for the thorough review :)

PR #22675, opened 2025-05-20
engine: install: remove Ubuntu 20.04, Fedora 40 (EOL)

thaJeztah:
Ubuntu 20.04 and Fedora 40 reached EOL, so remove mentions of it in the installation docs.

## Reviews

- [ ] Technical review
- [ ] Editorial review
- [ ] Product review

### DIFF
--- a/content/manuals/engine/install/fedora.md
+++ b/content/manuals/engine/install/fedora.md
@@ -28,7 +28,6 @@ Fedora versions:
 
 - Fedora 42
 - Fedora 41
-- Fedora 40
 
 ### Uninstall old versions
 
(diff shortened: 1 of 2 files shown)


ctalledo (approved):
LGTM.

However I still see another mention of Fedora 40 here: 
https://github.com/docker/docs/blob/main/content/manuals/desktop/setup/install/linux/fedora.md#L28

thaJeztah:
oh, good catch! Let me fix that.


that wording really is a bit redundant, but yeah,  people don't read the requirements first 😂

thaJeztah:
Oh, that's Docker Desktop! Hm.. good one.. I guess we should also update, but not sure what it's tested on. Let me assume that's also "currently non-EOL versions"

thaJeztah:
updated 👍

ctalledo:
LGTM ...

PR #22720, opened 2025-05-29
Fix multi-platform build command

k-kbk:
## Description
Replaced `docker build` with `docker buildx build` for clarity and consistency.


crazy-max:
We prefer to keep `docker build` that is an alias to `docker buildx build` already. More info: https://docs.docker.com/build/builders/#difference-between-docker-build-and-docker-buildx-build

PR #22726, opened 2025-05-30
engine: 28.2.2

vvoland:
## Reviews

- [ ] Technical review
- [ ] Editorial review
- [ ] Product review

### DIFF
--- a/content/manuals/engine/release-notes/28.md
+++ b/content/manuals/engine/release-notes/28.md
@@ -23,6 +23,23 @@ For more information about:
 - Deprecated and removed features, see [Deprecated Engine Features](../deprecated.md).
 - Changes to the Engine API, see [Engine API version history](/reference/api/engine/version-history.md).
 
+## 28.2.2
+
+{{< release-date date="2025-05-30" >}}
+
+For a full list of pull requests and changes in this release, refer to the relevant GitHub milestones:
+
+- [docker/cli, 28.2.2 milestone](https://github.com/docker/cli/issues?q=is%3Aclosed+milestone%3A28.2.2)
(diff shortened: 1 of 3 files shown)


content/manuals/engine/release-notes/28.md
```diff
@@ -23,6 +23,23 @@ For more information about:
...
+- [docker/cli, 28.2.2 milestone](https://github.com/docker/cli/issues?q=is%3Aclosed+milestone%3A28.2.2)
+- [moby/moby, 28.2.2 milestone](https://github.com/moby/moby/issues?q=is%3Aclosed+milestone%3A28.2.2)
+
+### Bug fixes and enhancements
+
+- containerd image store: Fix a regression causing `docker build --push` to fail. This reverts [the fix](https://github.com/moby/moby/pull/49702) for `docker build` not persisting overridden images as dangling. [moby/moby#50105](https://github.com/moby/moby/pull/50105)
+
+### Networking
+
+- When creating the iptables `DOCKER-USER` chain, do not add an explicit `RETURN` rule, allowing users to append as well as insert their own rules. (The rule is not removed on upgrade, but it will not be replaced after a reboot.). [moby/moby#50098](https://github.com/moby/moby/pull/50098)
```
thaJeztah, line 41: Punctuation looks a bit awkward (`(The rule is not removed on upgrade, but it will not be replaced after a reboot.).`); perhaps we can consider the braces to be redundant overall for this;

```suggestion
- When creating the iptables `DOCKER-USER` chain, do not add an explicit `RETURN` rule, allowing users to append as well as insert their own rules. Existing rules are not removed on upgrade, but it won't be replaced after a reboot. [moby/moby#50098](https://github.com/moby/moby/pull/50098)
```

vvoland:
This one can be merged

ArthurFlag (approved):

PR #22731, opened 2025-05-30
vale: add "BGP" (Border Gateway Protocol) to acronyms

thaJeztah:
## Related issues or tickets

- relates to https://github.com/docker/docs/pull/22601

## Reviews

- [ ] Technical review
- [ ] Editorial review
- [ ] Product review

### DIFF
--- a/_vale/Docker/Acronyms.yml
+++ b/_vale/Docker/Acronyms.yml
@@ -17,6 +17,7 @@ exceptions:
   - ASP
   - AUFS
   - AWS
+  - BGP     # Border Gateway Protocol
   - BIOS
   - BPF
   - BSD


ArthurFlag (approved):

_vale/Docker/Acronyms.yml
```diff
@@ -17,6 +17,7 @@ exceptions:
   - ASP
   - AUFS
   - AWS
+  - BGP     # Border Gateway Protocol
```
thaJeztah, line 20: Actually, wondering now if vale failed to detect how it was used, because the acronym was not on the same line 🤔 

```markdown
Docker host. This can be achieved using static routes, Border Gateway Protocol
(BGP), or any other means appropriate for your network. For example, within
```

I wonder if it would be OK with it used if it was on the same line;

```markdown
Docker host. This can be achieved using static routes, Border Gateway Protocol (BGP),
or any other means appropriate for your network. For example, within
```


https://github.com/docker/docs/blob/dc01252b40c86ddf86c06793a6aec310f1240a25/_vale/Docker/Acronyms.yml#L3-L10

thaJeztah: Maybe that regex needs tweaking to take newlines into account (in addition to spaces 🤔)

robmry: Yes - putting it on the same line sorted it.

PR #22753, opened 2025-06-03
add `platform-split` parameter

ArthurFlag:
## Description

Parameter was missing.

## Related issues or tickets

https://github.com/docker/buildx/issues/3215#issuecomment-2931258854

## Reviews

- [ ] Technical review
- [ ] Editorial review
- [ ] Product review


content/manuals/build/exporters/local-tar.md
```diff
@@ -25,9 +25,10 @@ $ docker buildx build --output type=tar[,parameters] .
 
 The following table describes the available parameters:
 
-| Parameter | Type   | Default | Description           |
-| --------- | ------ | ------- | --------------------- |
-| `dest`    | String |         | Path to copy files to |
+| Parameter        | Type    | Default | Description                                                                                                                                                                                 |
+|------------------|---------|---------|---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| `dest`           | String  |         | Path to copy files to                                                                                                                                                                       |
+| `platform-split` | Boolean | `false` | With a multi-platform build, a subfolder matching each target platform is created in the destination directory. Set it to `true` to merge files from all platforms into the same directory. |
```
crazy-max: It defaults to `true`, not `false`. See example in https://github.com/moby/buildkit?tab=readme-ov-file#local-directory

Also with upcoming https://github.com/moby/buildkit/pull/6007 and opting-in with `true` it would create the platform dir even if a single platform is being built. Which is not the case atm. So default should be _nil_ in such case (not defined). cc @tonistiigi

crazy-max (changes requested):
Can we have some examples similar to ones in BuildKit README? https://github.com/moby/buildkit?tab=readme-ov-file#local-directory

Like https://docs.docker.com/build/exporters/image-registry/#annotations for image export annotations.

content/manuals/build/exporters/local-tar.md
crazy-max: Also this attribute is only supported by the `local` exporter.

ArthurFlag: updated! Let's update the docs with this PR  to document the current state of things, I'll update the docs again once the linked bk PR is merged.

PR #22825, opened 2025-06-11
docs: Added section for WSL installation and verification

chaomonica:
## Reviews

- [ ] Technical review
- [ ] Editorial review
- [ ] Product review

### DIFF
--- a/content/manuals/desktop/setup/install/windows-install.md
+++ b/content/manuals/desktop/setup/install/windows-install.md
@@ -158,6 +158,34 @@ again when you switch back.
 
 {{< /accordion >}}
 
+## WSL: Verification and Setup
+
+If you've chosen to use WSL, first verify that your installed version meets system requirements by running the following command in your terminal:
+
+```console
+wsl --version
+```
+
+If version details do not appear, you’re likely using the inbox version of WSL. This version does not support modern capabilities and must be updated.
+
+You can update or install WSL using one of the following methods:
+
+### Option 1: Install or update WSL via the terminal
+1. Open PowerShell or Windows Command Prompt in administrator mode.
+2. Run either the install or update command. For more information, refer to [Install WSL](https://learn.microsoft.com/en-us/windows/wsl/install).
+```console
+wsl --install
+
+wsl --update
+```
+
+### Option 2: Install WSL via the MSI package
+If Microsoft Store access is blocked due to security policies:
+1. Go to the official WSL GitHub Releases page: https://github.com/microsoft/WSL/releases
+2. Download the `.msi` installer from the latest stable release (under the Assets dropdown).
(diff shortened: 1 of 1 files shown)


karman-docker:
@chelnak could you also have a look please whenever you can ? Motivation here is, many customers get stuck in WSL install step or complain about non-admin users getting UAC prompts. Hence we thought a this brief section before the install section.

content/manuals/desktop/setup/install/windows-install.md
```diff
@@ -158,6 +158,41 @@ again when you switch back.
...
+```
+
+#### Option 2: Install WSL via the MSI package
+
+If Microsoft Store access is blocked due to security policies:
+1. Go to the official WSL GitHub Releases page: https://github.com/microsoft/WSL/releases
+2. Download the `.msi` installer from the latest stable release (under the Assets drop-down).
+3. Run the downloaded installer and follow the setup instructions.
+
+### Adjust installation based on user's administrator access
```
chelnak: I'm not quite sure about this one.

When in WSL mode, you only activate the Windows Service if you select "Add the *.docker.internal names to the host's /etc/hosts file (Requires password)" from settings.

chelnak (changes requested):

content/manuals/desktop/setup/install/windows-install.md
chaomonica: Thanks Craig, I'll double check this on a VM.

chaomonica: I see that UAC prompt appears in the scenario you mentioned and oops, I missed the part on "com.docker.service is required by Windows containers and Hyper-V backend". 

I'll run through installation again tomorrow and fix this. Thanks again @chelnak!

karman-docker: @chaomonica / @chelnak we can word it something like this:

Title can be "Administrator Privileges and Installation Requirements", since this is generic, it can be outside the "WSL: Verification and Setup" section.

Installing Docker Desktop requires administrator privileges. However, once installed, it can be used without administrative access. Some actions, though, still need elevated permissions. For details, see [Understand permission requirements for Windows].

If your users do not have administrator rights and plan to perform operations that require elevated privileges, be sure to install Docker Desktop using the --always-run-service installer flag. This ensures those actions can still be executed without prompting for UAC elevation.

chaomonica: @chelnak I've updated the "Administrator Privileges and Installation Requirements" with @karman-docker 's section above and additionally moved this section ahead of the WSL sections to avoid confusion.

karman-docker: LGTM, thank you.

PR #22864, opened 2025-06-18
Fix `What is a registry?` formatting

JackPGreen:
Fixes missing space on https://docs.docker.com/get-started/docker-concepts/the-basics/what-is-a-registry/

image


sarahsanders-docker (approved):
Thanks for catching this @JackPGreen!

PR #22870, opened 2025-06-18
hub: add base url to api paths

craig-osterhout:
## Description

Display the base url specifically for Hub API and Registry API docs. 

Is this enough, or do we also need to write something about it in the description at the top?

image

## Related issues or tickets

ENGDOCS-2748

## Reviews

- [ ] Editorial review

### DIFF
--- a/layouts/_default/api-baseof.html
+++ b/layouts/_default/api-baseof.html
@@ -50,7 +50,7 @@
 
 <body>
   {{ $specURL := urls.Parse (printf "/%s%s.yaml" .File.Dir .File.ContentBaseName) }}
-  {{ if strings.HasPrefix .RelPermalink "/reference/api/hub/" }}
+  {{ if or (strings.HasPrefix .RelPermalink "/reference/api/hub/") (strings.HasPrefix .RelPermalink "/reference/api/registry/") }}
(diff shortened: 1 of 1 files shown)


thaJeztah:
Thanks!


Did a quick check if they used the correct URL (especially the registry, as it uses `registry-1.docker.io`, not `hub.docker.com`), but looks like it does;

Screenshot 2025-06-18 at 22 52 44

I have not checked all of them though, and perhaps @jcarter3  could help double-check.


> Is this enough, or do we also need to write something about it in the description at the top?

I'm somewhat tempted to say that it would still be good to have a page _somewhere_ that outlines the domains; specifically because we've historically not been good at documenting them, and in some case have introduced alternative domains that should not be used / not be considered the canonical domain. Having something at the top for each of these API docs won't hurt either.

I know many users have been confused about the domain to use, as many have assumed that `hub.docker.com` is the domain to use _both_ for the WebUI and APIs related to Docker Hub (which also has been a cause of issues when they want to block access to docker hub _registry_ 😅)

I don't have the full list of the correct ones though, but perhaps @jcarter3 (or someone else closer to Docker Hub and related services) can assist with that as well.

jcarter3:
Yup, looks good to me!

aevesdocker (approved):

PR #22923, opened 2025-06-25
dhi: add to homepage

craig-osterhout:
## Description

Add DHI card to homepage.
Refine wording as it's not just "base" images.

## Reviews

- [ ] Editorial review


craig-osterhout:
> LGTM. Just a thought - should we remove the Scout tile from the homepage since it's sort of becoming part of DHI?

Scout is still a separate product. Originally this grid had every product. With that said, for me, no concerns removing Scout as it's not high-traffic nor in the spotlight.

PR #22942, opened 2025-06-30
services: add notice of broken feature

sillydan1:
## Description

Waiting for healthy dependencies is not supported in version 3 of docker compose, as explained in: 
https://docs.docker.com/compose/releases/release-notes/#1100

The `version` key is also ignored now, so this whole feature is actually broken. I think at the very least, this should be mentioned in the documentation.

This took me 2 days of debugging to figure out, and a notice here would've saved me a lot of time.

## Related issues or tickets

- https://docs.docker.com/compose/releases/release-notes/#1100

## Reviews

- [ ] Technical review
- [ ] Editorial review
- [ ] Product review

### DIFF
--- a/content/reference/compose-file/services.md
+++ b/content/reference/compose-file/services.md
@@ -451,7 +451,8 @@ expressed in the short form.
   - `service_started`: An equivalent of the short syntax described previously
   - `service_healthy`: Specifies that a dependency is expected to be "healthy"
     (as indicated by [`healthcheck`](#healthcheck)) before starting a dependent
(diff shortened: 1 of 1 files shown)


aevesdocker:
Hi @sillydan1, the docs site shows the current version of the compose file reference which is the Compose Specification, which is the latest and recommended version. Your fix is not applicable in this case. For more info on the Compose Spec and how it relates to compose file v3, see this page https://docs.docker.com/compose/intro/history/

PR #22968, opened 2025-07-01
chore: reduce font size

ArthurFlag:


layouts/partials/md-dropdown.html
sarahsanders-docker, line 1: image

the padding is a little squished now, especially on the left-hand side. not as noticeable on the right with the arrow

ArthurFlag: fixed! I also made the text thinner, I feel it balances the top of the page a bit more?

sarahsanders-docker: Looks great!!

PR #23011, opened 2025-07-07
Gordon DHI migration

ArthurFlag:


content/includes/gordondhi.md
```diff
@@ -0,0 +1,24 @@
+1. Ensure Gordon is [enabled](/manuals/ai/gordon.md#enable-ask-gordon).
+1. In Gordon's Toolkit, ensure Gordon's [Developer MCP toolkit is enabled](/manuals/ai/gordon/mcp/built-in-tools.md#configuration).
```
jeanlaurent: Gordon's Toolkit is confusing, and should not be used.
I think you can refer to Gordon tools selection or Gordon tools selection ui, Maybe you could also give a visual hint ( the hammer in the chat interface.

content/includes/gordondhi.md
```diff
@@ -0,0 +1,24 @@
...
+1. Start a conversation with Gordon:
+   ```bash
+   docker ai
+   ```
+1. Type:
+   ```console
+   "Migrate my dockerfile to DHI"
+   ```
+1. Follow the conversation with Gordon. When it requests access to the filesystem and more,
+   type `yes` to enable it to update your Dockerfile.
```
jeanlaurent: I think adding something like (if you want to know more about Gordon's data retention and what it can access you can there https://docs.docker.com/ai/gordon/#what-data-does-gordon-access. DHI migration is about updating a Dockerfile, so we specifically request here access in write access to your local dockerfile.

content/includes/gordondhi.md
ArthurFlag: That's all behind the link on line 2

jeanlaurent: Ok the key part is *not* using the wording: Gordon's Toolkit as it clashed with the overall docker toolkit.

ArthurFlag: indeed, that's called the toolbox.

PR #23078, opened 2025-07-12
poc drilldown multigroup

craig-osterhout:
## Description

Groups: https://deploy-preview-23078--docsdocker.netlify.app/manuals/
Example drilldown: https://deploy-preview-23078--docsdocker.netlify.app/docker-hub/

A quick POC to explore how drilldown and multi-group manuals feel.

Multi-group assignment lets us not have to try to find non-overlappig groups.
Drilldown reduces the overwhelming left nav and enables multi-groups.

Used javascript since it was quicker to poc, but we already have a similar static implementation for tags/samples we can reuse if this direction looks good.

Didn't make any effort on styling. But imagine flashy cards instead of simple list. Also, dark mode doesn't work in this poc.

Inspired by the product category component at https://docs.aws.amazon.com/

Added a fake "Desktop" group to demonstrate cross-group membership.

### DIFF
--- a/content/manuals/_index.md
+++ b/content/manuals/_index.md
@@ -13,6 +13,7 @@ params:
       - AI
       - Products
       - Platform
+      - Desktop
   notoc: true
   open-source:
   - title: Docker Build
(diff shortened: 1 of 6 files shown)


craig-osterhout:
Single page manuals don't drill down correctly. No left nav is generated.
image
Although, if using a folder with just a single index page, the menu generates but has a useless dropdown because there are no child topics. This is actually a bug in our current implementation too.

image

PR #23084, opened 2025-07-14
ci: deploy from main

ArthurFlag:
PR that changes the deploy workflow of the docs. 

## What has changed?

This PR simplifies the deployment flow by deleting our staging build. We publish what is on the main branch instead of relying on the `publish` and `publish-update` branches.


.github/workflows/deploy.yml
```diff
@@ -30,38 +28,16 @@ jobs:
...
-          elif [ "${{ github.ref }}" = "refs/heads/published" ]; then
-            HUGO_ENV=production
-            DOCS_URL="https://docs.docker.com"
-            DOCS_AWS_IAM_ROLE="arn:aws:iam::710015040892:role/prod-docs-docs.docker.com-20220818202218674300000001"
-            DOCS_S3_BUCKET="prod-docs-docs.docker.com"
-            DOCS_S3_CONFIG="s3-config.json"
-            DOCS_CLOUDFRONT_ID="E228TTN20HNU8F"
-            DOCS_LAMBDA_FUNCTION_REDIRECTS="DockerDocsRedirectFunction-prod"
-            DOCS_SLACK_MSG="Successfully deployed docs from published branch. $DOCS_URL"
-          elif [ "${{ github.ref }}" = "refs/heads/lab" ]; then
```
StefanScherer, line 53: Maybe keep the `lab` branch mechanism to have a place/branch to try something new like a new tool or refactoring?

ArthurFlag: Good point. Might be useful in potential migration scenarios, too. I brought back the big about this branch.

.github/workflows/deploy.yml
```diff
@@ -8,9 +8,7 @@ on:
   workflow_dispatch:
   push:
     branches:
-      - lab
```
StefanScherer, line 11: Keep this branch to be able to run the workflow when your team does some testing without wondering why nothing happens. 😅

ArthurFlag: good catch hahah

StefanScherer (approved):

PR #23125, opened 2025-07-21
freshness: engine

ArthurFlag:


content/manuals/engine/_index.md
```diff
@@ -51,16 +51,11 @@ aliases:
...
 
-- A server with a long-running daemon process
-  [`dockerd`](/reference/cli/dockerd).
-- APIs which specify interfaces that programs can use to talk to and instruct
-  the Docker daemon.
-- A command line interface (CLI) client
-  [`docker`](/reference/cli/docker/).
+- A server with a long-running daemon process [`dockerd`](/reference/cli/dockerd)
+- APIs that specify interfaces that programs can use to talk to and instruct the Docker daemon
+- A command line interface (CLI) client [`docker`](/reference/cli/docker/)
```
thaJeztah: I think this block only un-wraps lines; could you keep those the same to prevent code-churn?

content/manuals/engine/install/_index.md
```diff
@@ -54,70 +54,51 @@ see: [Overview of Docker Desktop](/manuals/desktop/_index.md).
...
+Do not file a public issue. Instead, send your report privately to security@docker.com.
 
-Security reports are greatly appreciated, and Docker will publicly thank you for it.
+Docker appreciates security reports and will publicly thank you for your contribution.
 
 ## Get started
 
-After setting up Docker, you can learn the basics with
-[Getting started with Docker](/get-started/introduction/_index.md).
+After setting up Docker, you can learn the basics with [Getting started with Docker](/get-started/introduction/_index.md).
```
thaJeztah: Looks like this, and other changes are only un-wrapping lines; we try to wrap at ~80-chars; can you revert those changes?

content/manuals/engine/install/debian.md
```diff
@@ -22,25 +22,16 @@ To get started with Docker Engine on Debian, make sure you
 
 ### Firewall limitations
 
-> [!WARNING]
->
-> Before you install Docker, make sure you consider the following
-> security implications and firewall incompatibilities.
```
thaJeztah: Need to double-check with @robmry - but possibly this may need to remain a big fat, ugly warning, due to the security impact.

content/manuals/engine/install/debian.md
```diff
@@ -51,13 +42,8 @@ and ppc64le (ppc64el) architectures.
 
 ### Uninstall old versions
 
-Before you can install Docker Engine, you need to uninstall any conflicting packages.
-
-Your Linux distribution may provide unofficial Docker packages, which may conflict
-with the official packages provided by Docker. You must uninstall these packages
-before you install the official version of Docker Engine.
```
thaJeztah, line 58: Wondering if we should keep some context about "why" this is done; we've had questions about that, and providing context could help on why we need to do this.

content/manuals/engine/install/sles.md
```diff
@@ -58,7 +51,7 @@ $ sudo zypper addrepo $opensuse_repo
 
 ### Uninstall old versions
 
-Before you can install Docker Engine, you need to uninstall any conflicting packages.
+Uninstall any conflicting packages before you install Docker Engine.
```
thaJeztah: In case it helps; we kept most of the deb and rpm distros aligned, so when updating (say) CentOS, you can do a side-by-side compare of the file with RHEL; there should be only minor changes in between (related to distro name), otherwise identical.

If possible, it'd be good to keep them aligned, so that when updates are made, we can sync them across.

thaJeztah (commented):
just some quick blurbs after having a first glance.

content/manuals/engine/install/debian.md
robmry: Yes, it's probably worth keeping it fat and ugly for now, because of the `ufw` issue ... @akerouanton's got a cunning plan for dealing with that - coming soon! We should be able to drop the warning when that's done.

(We'll need to tweak the nftables text a bit for moby 29.0, but it's probably best to deal with all the nftables updates together.)

ArthurFlag:
@thaJeztah Any chance you could take another look?

ArthurFlag:
@thaJeztah, any chance you could take another look over the next few days? :)

PR #23134, opened 2025-07-22
Update configure-github-actions.md

alanmehio:
The user needs to create a DockerHub repository which has the same name as the GitHub repository  for example, if my GitHub repository is my-proj  so the DockerHub needs to be the same name which is my-proj
The tag yaml elements needs to give an example which show the mirror on the DockerHub

## Description

The github actions does not work if the repository name are not the same 

## Reviews

- [ ] Technical review
- [ ] Editorial review
- [ ] Product review


content/guides/python/configure-github-actions.md
```diff
@@ -21,8 +21,9 @@ If you didn't create a [GitHub repository](https://github.com/new) for your proj
 2. Under the **Variables** tab, create a new **Repository variable** named `DOCKER_USERNAME` and your Docker ID as a value.
 
 3. Create a new [Personal Access Token (PAT)](/manuals/security/access-tokens.md#create-an-access-token) for Docker Hub. You can name this token `docker-tutorial`. Make sure access permissions include Read and Write.
+4. Create DockerHub repository use the **same name** as your GitHub repository
```
crazy-max, line 24: This is not necessary to create the Docker Hub repo manually. It will be created when the image is being pushed.

alan-techdev: Thanks

PR #23145, opened 2025-07-23
fix: open in llm dropdown

sarahsanders-docker:
## Description
- The original implementation was trying to pass the entire content of the documentation page through the URL parameters
- Web servers/browsers have limits on URLs
- Instead of passing the whole page as a URL parameter, the fix prompts Claude with a reference to the markdown file, putting the fetching and parsing onto the LLM, not the browser (which can't handle long files/pages)
- Removing ChatGPT option, as OpenAI appears to no longer support reading RAW markdown files (I tested this using our site and several other docs sites)

## Related issues or tickets
Fixes https://github.com/docker/docs/issues/23144

## Reviews
- [ ] Editorial review

## Testing
Claude
image


ArthurFlag:
The trade-off is that we cannot use these features locally or from preview deployments, so maybe we should remove these 2 buttons from non-production builds. Wdyt?

sarahsanders-docker:
> The trade-off is that we cannot use these features locally or from preview deployments, so maybe we should remove these 2 buttons from non-production builds. Wdyt?

added the production environment variable in the latest commit! good idea :)

PR #23169, opened 2025-07-28
cli: add generated MCP cli refs

ArthurFlag:
Add generated MCP docs.


_vendor/github.com/docker/mcp-gateway/docs/generator/reference/mcp_catalog.md
thaJeztah, line 1: For the docs repository, we probably only need the YAML files (as the docs uses its own templates to convert those to markdown). Not sure how we did that for other repositories that committed the generated files in source control though (do we put the markdown and yaml in separate directories for those?). @crazy-max do you know?

content/reference/cli/docker/mcp/catalog_reset.md
thaJeztah, line 1: Starting to wonder if we need to start considering adding another level in our hierarchy (`reference/cli/docker/mcp/catalog_reset/` -> `reference/cli/docker/mcp/catalog/reset/`). I know we had some of that with `docker buildx imagetools inspect`, but I think we kept the deeply nested at the same level at the time.

thaJeztah: Oh! Actually, we didn't that's also using an extra hierarchy; https://docs.docker.com/reference/cli/docker/buildx/imagetools/inspect/#format

thaJeztah: Right, yeah, looks like we have some changes to make for the navigation; looking at the preview;

Screenshot 2025-07-28 at 12 26 29
Screenshot 2025-07-28 at 12 26 39

content/reference/cli/docker/mcp/catalog.md
crazy-max, line 1: You need to create a dedicated folder for each subcommands if there are nested ones. This `mcp/catalog.md` should be `mcp/catalog/_index.md`.

_vendor/github.com/docker/mcp-gateway/docs/generator/reference/mcp_catalog.md
crazy-max: Only yaml files are consumed https://github.com/docker/docs/pull/23169/files#diff-5928f4025679c9eee397ca3795367f71f667ce1f25018b7f7ae7c574da1f2f08R359 but markdown files seem to also be vendored like compose for example: https://github.com/docker/docs/tree/main/_vendor/github.com/docker/compose/v2/docs/reference

That might be how hugo works :thinking:

ArthurFlag:
@crazy-max  Would you able to take another look at your convenience?

crazy-max (commented):
LGTM

Added @dgageot as reviewer for final approval

ArthurFlag:
@crazy-max David's off until Sept 1st, I'd rather we merged now and let him take a look later if we need tweaks. Wdyt?

PR #23172, opened 2025-07-28
build: freshness

ArthurFlag:
Apply various linguistic efficiency improvements.


content/manuals/build/concepts/context.md
```diff
@@ -437,14 +435,14 @@ improving build speed, especially when using a remote builder.
 ### Filename and location
 
 When you run a build command, the build client looks for a file named
-`.dockerignore` in the root directory of the context. If this file exists, the
-files and directories that match patterns in the files are removed from the
-build context before it's sent to the builder.
+`.dockerignore` in the root directory of the context. If this file exists, files
+and directories that match patterns in the file are removed from the build context
+before Docker sends them to the builder.
```
crazy-max: It's not Docker but the build command so "it" is correct here.

```suggestion
before it sends them to the builder.
```

content/manuals/build/concepts/context.md
```diff
@@ -437,14 +435,14 @@ improving build speed, especially when using a remote builder.
...
 
-If you use multiple Dockerfiles, you can use different ignore-files for each
-Dockerfile. You do so using a special naming convention for the ignore-files.
-Place your ignore-file in the same directory as the Dockerfile, and prefix the
-ignore-file with the name of the Dockerfile, as shown in the following example.
+If you use multiple Dockerfiles, you can use different ignore files for each
```
crazy-max: I would say "ignore-files" is better to qualify dockerignore feature imo.

```suggestion
If you use multiple Dockerfiles, you can use different ignore-files for each
```

content/manuals/build/concepts/context.md
```diff
@@ -717,7 +715,7 @@ where you have two Dockerfiles:
 - `base.Dockerfile`: for building a base image
 - `app.Dockerfile`: for building an application image
 
-The `app.Dockerfile` uses the image produced by `base.Dockerfile` as it's base
+The `app.Dockerfile` uses the image produced by its `base.Dockerfile` as its base
```
crazy-max: Don't think "its base.Dockerfile" is correct. That would mean app.Dockerfile depends on base.Dockerfile. Here we are talking about the build result of the Dockerfile. So either remove "its" or replace it with "the":

```suggestion
The `app.Dockerfile` uses the image produced by `base.Dockerfile` as its base
```

content/manuals/build/concepts/dockerfile.md
```diff
@@ -1,66 +1,58 @@
 ---
 title: Dockerfile overview
 weight: 20
-description: Learn about Dockerfiles and how to use them with Docker Images to build and package your software
-keywords: build, buildx, buildkit, getting started, dockerfile
+description: Learn how to use Dockerfiles to build and package your software into Docker images.
+keywords: dockerfile, docker build, buildx, buildkit, container image, getting started, image layers, dockerfile instructions
```
crazy-max:
```suggestion
keywords: dockerfile, docker build, buildx, buildkit, container image, getting started, image layers, dockerfile, instructions
```

content/manuals/build/concepts/dockerfile.md
```diff
@@ -1,66 +1,58 @@
...
 ### Filename
 
-The default filename to use for a Dockerfile is `Dockerfile`, without a file
-extension. Using the default name allows you to run the `docker build` command
-without having to specify additional command flags.
+The default filename for a Dockerfile is `Dockerfile`, without a file extension. Using
+the default name lets you run the `docker build` command without extra flags.
 
-Some projects may need distinct Dockerfiles for specific purposes. A common
+Some projects need different Dockerfiles for specific purposes. A common
```
crazy-max: "may" is important here and I think the current sentence sounds better.

```suggestion
Some projects may need distinct Dockerfiles for specific purposes. A common
```

content/manuals/build/concepts/dockerfile.md
```diff
@@ -1,66 +1,58 @@
...
+> We recommend using the default (`Dockerfile`) for your project's main Dockerfile.
 
 ## Docker images
 
-Docker images consist of layers. Each layer is the result of a build
-instruction in the Dockerfile. Layers are stacked sequentially, and each one is
-a delta representing the changes applied to the previous layer.
+Docker images consist of layers. Each layer results from a build instruction in the
+Dockerfile. Layers are stacked in order, and each one represents changes applied to
+the previous layer.
```
crazy-max: Current one is more technically correct than this imo, I would not change it.

```suggestion
Docker images consist of layers. Each layer is the result of a build
instruction in the Dockerfile. Layers are stacked sequentially, and each one is
a delta representing the changes applied to the previous layer.
```

content/manuals/build/concepts/dockerfile.md
```diff
@@ -1,66 +1,58 @@
...
+|-----------------------------------------------------------|----------------------------------------------------------------------------------------------------------------------------------------|
+| [`FROM <image>`](/reference/dockerfile.md#from)           | Defines a base for your image.                                                                                                         |
+| [`RUN <command>`](/reference/dockerfile.md#run)           | Executes commands in a new layer on top of the current image and commits the result. `RUN` also has a shell form for running commands. |
+| [`WORKDIR <directory>`](/reference/dockerfile.md#workdir) | Sets the working directory for any `RUN`, `CMD`, `ENTRYPOINT`, `COPY`, and `ADD` instructions that follow it in the Dockerfile.        |
+| [`COPY <src> <dest>`](/reference/dockerfile.md#copy)      | Copies new files or directories from `<src>` and adds them to the container at the path `<dest>`.                                      |
+| [`CMD <command>`](/reference/dockerfile.md#cmd)           | Defines the default program that runs when you start the container. Only the last `CMD` in the Dockerfile is used if multiple exist.   |
```
crazy-max: This is important to keep "based on this image" so users know this runs with the built image and not during build.

```suggestion
| [`CMD <command>`](/reference/dockerfile.md#cmd)           | Defines the default program that runs when you start the container based on this image. Only the last `CMD` in the Dockerfile is used if multiple exist.   |
```

content/manuals/build/concepts/dockerfile.md
```diff
@@ -1,66 +1,58 @@
...
 
-Dockerfiles are crucial inputs for image builds and can facilitate automated,
-multi-layer image builds based on your unique configurations. Dockerfiles can
-start simple and grow with your needs to support more complex scenarios.
+Dockerfiles are essential for image builds and can automate multi-layer image builds
```
crazy-max: Why essential? That breaks the meaning, Dockerfile is crucial not essential. I would keep it as before.

.github/prompts/freshness-tier1.prompt.md
```diff
@@ -12,5 +12,6 @@ how fresh and up to date it is. Apply the following:
 5. Ensure all the components formatted correctly.
 6. Improve the SEO keywords.
 7. If you find numbered lists, make sure their numbering only uses 1's.
+8. Keep line length to a maximum of 90 characters, and do not add a space character at the end of lines.
```
crazy-max: It should be 80 to be conform with what we currently have across all markdown. This is also part of our conventions: https://github.com/docker/docs/blob/331240357cedab35b722d321b46d96aa241e1c09/content/contribute/file-conventions.md?plain=1#L66

content/manuals/build/concepts/dockerfile.md
```diff
@@ -112,46 +103,40 @@ Here's a breakdown of what this Dockerfile does:
...
 ```
 
 The [`FROM` instruction](/reference/dockerfile.md#from) sets your base
-image to the 22.04 release of Ubuntu. All instructions that follow are executed
-in this base image: an Ubuntu environment. The notation `ubuntu:22.04`, follows
-the `name:tag` standard for naming Docker images. When you build images, you
-use this notation to name your images. There are many public images you can
-leverage in your projects, by importing them into your build steps using the
-Dockerfile `FROM` instruction.
+image to Ubuntu 22.04. All following instructions run in this Ubuntu environment. The
```
crazy-max: This assume people know what means 22.04 for Ubuntu which might not be the case for introducing to concepts. I think we should keep the current copy:

```suggestion
image to the 22.04 release of Ubuntu. All following instructions run in this Ubuntu environment. The
```

content/manuals/build/concepts/dockerfile.md
```diff
@@ -164,23 +149,17 @@ the container.
 
 ### Comments
 
-Note the `# install app dependencies` line. This is a comment. Comments in
-Dockerfiles begin with the `#` symbol. As your Dockerfile evolves, comments can
-be instrumental to document how your Dockerfile works for any future readers
-and editors of the file, including your future self!
+The `# install app dependencies` line is a comment. Comments in Dockerfiles start
+with `#`. Comments help document your Dockerfile for future readers and editors.
```
crazy-max: It makes assumption and I preferred the previous one that explains the purpose and meaning of comments. I think we should revert this.

content/manuals/build/concepts/dockerfile.md
```diff
@@ -164,23 +149,17 @@ the container.
...
-> treated as a comment.
+> Comments use the same symbol as the [syntax directive](#dockerfile-syntax) on the
+> first line. The symbol is only a directive if it matches a directive pattern and is
+> at the start of the Dockerfile. Otherwise, it is a comment.
 
 ### Installing dependencies
 
-The second `RUN` instruction installs the `flask` dependency required by the
-Python application.
+The second `RUN` instruction installs the `flask` dependency for the Python app.
```
crazy-max, line 159: Why is "app" better than "application"?

.github/prompts/freshness-tier1.prompt.md
ArthurFlag: Unfortunately, this is not applied in many places of the docs and this is also worded very loosely, as a simple preference. I'll bring this up to the team.

content/manuals/build/concepts/dockerfile.md
ArthurFlag: Most of our guides refer to applications as app. We can harmonize either way.

ArthurFlag:
It's indeed mostly AI-touched up to see how it runs our styleguide on the content. Maybe that's a bit of  waste of time (and review time) for the outcome. I'm not super convinced with this initiative.

.github/prompts/freshness-tier1.prompt.md
crazy-max: Ah yes indeed some pages are not aligned. Wonder if we could have a way to validate this with a linter.

crazy-max:
> It's indeed mostly AI-touched up to see how it runs our styleguide on the content. Maybe that's a bit of waste of time (and review time) for the outcome. I'm not super convinced with this initiative.

Yes me too but that's good to test it out and maybe improve the model and prompts if we can.

ArthurFlag:
@crazy-max, any chance you could take another look this week? 🙇

ArthurFlag:
Ping @crazy-max :)

crazy-max:
> > It's indeed mostly AI-touched up to see how it runs our styleguide on the content. Maybe that's a bit of waste of time (and review time) for the outcome. I'm not super convinced with this initiative.
> 
> Yes me too but that's good to test it out and maybe improve the model and prompts if we can.

Sorry maybe there was a misunderstanding but I don't think we should merge this related to my comment:

> Didn't finish reviewing yet, but I noticed a few places where the text seems to take shortcuts that subtly change the meaning of core concepts or lose important nuance. For pages like concepts, which are relied on for precision (especially by folks deep in the ecosystem) it's really important to preserve the intent and technical accuracy of the original terms.
> 
> Also, looking at the freshness prompt files, I'm guessing this might have been generated with the help of an AI assistant? If so, it'd be helpful to split AI-generated changes and your own edits into separate commits, so it's easier to see where the AI may have gone off track.

If you could split AI-generated changes and your own edits into separate commits, so it's easier to see where the AI may have gone off track because currently it's pretty hard to review.

PR #23193, opened 2025-08-01
Update etwlogs.md

kanwal-kaur:
## Description

The command doesn't work as intended. Explanation for the fix:
1. Windows requires the GUID inside quotes because it's treated as a string (the provider name) in this context. Without quotes, the command parser may interpret the braces or hyphens incorrectly.
2. Passing 0x0 as the only flag after the provider GUID is valid. Supplying two numbers like 0 0 (as in some bad examples) misleads logman, which tries to interpret the second value as outputFormat.
3. The -o trace.etl must be at the end and not mistaken for a different parameter like -f (output format), unless you're also explicitly setting -f to text or xml.

## Related issues or tickets

Worked out in Case number: [00164653](https://docker.lightning.force.com/lightning/r/Case/500U100000VvQDdIAN/view)

## Reviews

- [ ] Technical review
- [ ] Editorial review
- [ ] Product review

### DIFF
--- a/content/manuals/engine/logging/drivers/etwlogs.md
+++ b/content/manuals/engine/logging/drivers/etwlogs.md
@@ -25,7 +25,7 @@ before the provider has been registered with the system.
 Here is an example of how to listen to these events using the logman utility program
 included in most installations of Windows:
 
-1. `logman start -ets DockerContainerLogs -p {a3693192-9ed6-46d2-a981-f8226c8363bd} 0 0 -o trace.etl`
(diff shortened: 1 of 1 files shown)


thaJeztah:
Thanks for contributing! @chelnak could you have a peek to verify the change? I'm not on a Windows machine myself, so both "before" and "after" look fine to me 😂 ❤️

chelnak:
Hey! I Tested with PowerShell 7, PowerShell 5 and cmd.exe. The new version seems to work in all three.

Whereas previously the example would have only worked in cmd.exe.

chelnak (approved):

thaJeztah (approved):
LGTM, thank you for testing!!!!

PR #23217, opened 2025-08-07
Split iptables and port publishing into separate pages

robmry:
## Description

With nftables on the way - refer to "firewall" instead of "iptables" in the top-level description in `packet-filtering-firewalls.md`, move out the iptables specifics, and port-publishing (which applies to both iptables and nftables).

_The specifics for nftables will go into a new page - `firewall-nftables.md`, alongside `firewall-iptables.md`. The split probably isn't too worthwhile until that exists. When I've made some more progress, I can add the nftables updates here (or a separate PR, if PRs can be merged to main without publishing to the website?)_

## Reviews

- [ ] Technical review - @akerouanton 
- [ ] Editorial review - @ArthurFlag 
- [ ] Product review

### DIFF
--- a/content/manuals/engine/network/_index.md
+++ b/content/manuals/engine/network/_index.md
@@ -129,54 +129,18 @@ $ docker run --rm -it --network container:redis example/redis-cli -h 127.0.0.1
 
 ## Published ports
 
-By default, when you create or run a container using `docker create` or `docker run`,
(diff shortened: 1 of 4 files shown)


robmry:
I'll close this ... was hoping to be able to raise separate PRs, but the later changes depend on earlier ones too much. So, will put it all into https://github.com/docker/docs/pull/23215

PR #23227, opened 2025-08-11
dmr: split up content

ArthurFlag:
## Description

Split up the content for DMR for better UX and maintainability.


content/manuals/ai/model-runner/api-reference.md
```diff
@@ -0,0 +1,192 @@
+---
+title: DMR REST API
```
sarahsanders-docker, line 2: This page should live in our API reference section and link to it from one of the DMR manuals pages: https://docs.docker.com/reference/

The Extensions API is an example of one that isn't generated w/ a spec + Redocly: https://docs.docker.com/reference/api/extensions-sdk/BackendV0/

This will make it easier to keep track of all references in one place!

sarahsanders-docker (changes requested):
The reorg of content makes sense and reads much better! I left a suggestion to move the DMR API page to the reference section just to keep things in one place

content/manuals/ai/model-runner/api-reference.md
ArthurFlag: If it's ok with you, should we merge this PR so that I can create the APi pages in a follow up PR?

sarahsanders-docker (approved):

PR #23242, opened 2025-08-13
update `wsl` to `wsl.exe` for consistency reasons and to allow copy-p…

lobis:
Add .exe extension to allow command to work on windows as the other commands in the page work.


aevesdocker:
thanks!

PR #23243, opened 2025-08-14
ux: move support and theme buttons to topnav

ArthurFlag:
Move the contact support button and the theme switch button to the top nav bar.


layouts/partials/header.html
```diff
@@ -45,19 +45,52 @@
...
+            {{ partialCached "utils/svg.html" "/icons/sparkle.svg" }}
+          </span>
+        </button>
+
+        <div class="topbar-button" id="search-bar-container">
+           {{ partialCached "search-bar.html" "-" }}
+        </div>
+
+        <button class="topbar-button" href="https://hub.docker.com/support/contact"
+            aria-label="Contact support">
```
sarahsanders-docker: Looks good! Just one thing to note: in the preview, the contact support button doesn't redirect to this URL. Will have to test in production to verify

ArthurFlag: good catch 🙃 Fixed!

Issue #23231, opened 2025-08-12
APT install instructions for Debian 13


Matthias-vdE:
### Is this a docs issue?

- [x] My issue is about the documentation content or website

### Type of issue

Information is incorrect

### Description

The documentation gives instructions on writing to the file

`/etc/apt/sources.list.d/docker.list`

In order to add the APT repository. 

However, according to the Debian 12 to 13 upgrade guide (https://www.debian.org/releases/trixie/release-notes/upgrading.en.html?ref=selfh.st#preparing-apt-sources-files), these types of files are no longer recommended:

> As mentioned in [Start from “pure” Debian](https://www.debian.org/releases/trixie/release-notes/upgrading.en.html?ref=selfh.st#system-status), we recommend that you use the new deb822-style format, so you would have to replace /etc/apt/sources.list and any *.list files in /etc/apt/sources.list.d/ by only one file named debian.sources in /etc/apt/sources.list.d/ (if you haven’t done so already)

### Location

https://docs.docker.com/engine/install/debian/#install-using-the-repository

### Suggestion

While the "old" docker.list still works, the documentation could also make note of Debian 13's new APT format, as stated in https://www.debian.org/releases/trixie/release-notes/upgrading.en.html?ref=selfh.st#start-from-pure-debian;

>APT is moving to a different format for configuring where it downloads packages from. The files /etc/apt/sources.list and *.list files in /etc/apt/sources.list.d/ are replaced by files still in that directory but with names ending in .sources, using the new, more readable (deb822 style) format.

eshattow:
duplicate issue of #15382 ?

oconnordaniel:
A quick test works with the file 

``` sources
Types: deb
URIs: https://download.docker.com/linux/debian 
Suites: trixie
Components: stable
Architectures: amd64
Signed-By: /etc/apt/keyrings/docker.asc
```

On the [documentation](https://docs.docker.com/engine/install/debian/#install-using-the-repository) page, you could swap the `echo < .list file > | sudo tee filename > /dev/null` line with the following command

``` bash
cat << EOF | sudo tee -a /etc/apt/sources.list.d/docker.sources 
Types: deb
URIs: https://download.docker.com/linux/debian 
Suites: $(. /etc/os-release && echo "$VERSION_CODENAME") 
Components: stable
Architectures: $(dpkg --print-architecture)
Signed-By: /etc/apt/keyrings/docker.asc
EOF
```

Haven't done any testing except on a single debain 13 x86 system.

[...]

Issue #15382, opened 2022-08-14
Compliance with Deb822: Deprecated sources.list use .sources instead


eshattow:
[engine/install/debian.md](https://docs.docker.com/engine/install/debian/)

sources.list(5) says of the Deb822 ".sources" format "It is intended to make this format gradually the default format, deprecating the previously described one-line-style format, as it is easier to create, extend and modify for humans and machines alike..."

https://wiki.debian.org/DebianRepository/UseThirdParty

Note for clarity /etc/apt/keyrings as the Docker install guide suggest is in fact the correct location for a keyring installed by the system operator, so the guide gets it half-right already.

eshattow:
deb822 format will be the default for the next Debian release (either trixie or trixie+1) c.f. https://github.com/debuerreotype/debuerreotype/pull/128

See comparable examples of how to form documentation for users:
https://wiki.winehq.org/Debian
https://qgis.org/en/site/forusers/alldownloads.html#debian-ubuntu

eshattow:
Trixie is released and now stable see also #23231

oconnordaniel:
I probably put these comments on the wrong issue... but for the googlers finding this... 

https://github.com/docker/docs/issues/23231#issuecomment-3369233157

dvdksn:
Implemented in #23249

PR #23249, opened 2025-08-15
Add deb822 apt sources to install section in Debian manual

LLdaniel:
## Description

With Debian Trixie released the new apt sources format [RFC822](https://manpages.debian.org/trixie/apt/sources.list.5.en.html) is now the default. 
This PR proposes a way to take this new format into account.

The goal would be to have the follwoing `docker.sources`:
```
$ cat /etc/apt/sources.list.d/docker.sources 
Types: deb
URIs: https://download.docker.com/linux/debian/
Suites: trixie
Components: stable
Signed-By: /etc/apt/keyrings/docker.gpg
```

## Related issues or tickets

#15382 and #23231 

## Reviews

I am happy to change my PR if needed. Currently I was not sure if we should keep both formats or make the new deb822 format as default. 
Maybe we could move the legacy format into an info box?

### DIFF
--- a/content/manuals/engine/install/debian.md
+++ b/content/manuals/engine/install/debian.md
@@ -114,11 +114,21 @@ Docker from the repository.
    sudo curl -fsSL {{% param "download-url-base" %}}/gpg -o /etc/apt/keyrings/docker.asc
    sudo chmod a+r /etc/apt/keyrings/docker.asc
 
-   # Add the repository to Apt sources:
+   # Add the repository to Apt sources (legacy format):
    echo \
      "deb [arch=$(dpkg --print-architecture) signed-by=/etc/apt/keyrings/docker.asc] {{% param "download-url-base" %}} \
      $(. /etc/os-release && echo "$VERSION_CODENAME") stable" | \
      sudo tee /etc/apt/sources.list.d/docker.list > /dev/null
+
+   # Add the repository to Apt sources (new deb822 format, introduced with Trixie):
+   echo -e \
+     "Types: deb\n \
+	 URIs: https://download.docker.com/linux/debian/\n \
+	 Suites: $(. /etc/os-release && echo "$VERSION_CODENAME")\n \
+	 Components: stable\n \
+	 Signed-By: /etc/apt/keyrings/docker.gpg" | \
+	 sudo tee /etc/apt/sources.list.d/docker.sources > /dev/null
+
    sudo apt-get update
    ```
 
--- a/content/manuals/engine/install/debian.md
+++ b/content/manuals/engine/install/debian.md
@@ -122,12 +122,12 @@ Docker from the repository.
 
    # Add the repository to Apt sources (new deb822 format, introduced with Trixie):
    echo -e \
-     "Types: deb\n \
-	 URIs: https://download.docker.com/linux/debian/\n \
-	 Suites: $(. /etc/os-release && echo "$VERSION_CODENAME")\n \
-	 Components: stable\n \
-	 Signed-By: /etc/apt/keyrings/docker.gpg" | \
-	 sudo tee /etc/apt/sources.list.d/docker.sources > /dev/null
+   "Types: deb\n\
+   URIs: https://download.docker.com/linux/debian/\n\
+   Suites: $(. /etc/os-release && echo "$VERSION_CODENAME")\n\
+   Components: stable\n\
+   Signed-By: /etc/apt/keyrings/docker.gpg" | \
+   sudo tee /etc/apt/sources.list.d/docker.sources > /dev/null
 
    sudo apt-get update
    ```


content/manuals/engine/install/debian.md
```diff
@@ -114,11 +114,21 @@ Docker from the repository.
...
    sudo chmod a+r /etc/apt/keyrings/docker.asc
 
-   # Add the repository to Apt sources:
+   # Add the repository to Apt sources (legacy format):
    echo \
      "deb [arch=$(dpkg --print-architecture) signed-by=/etc/apt/keyrings/docker.asc] {{% param "download-url-base" %}} \
      $(. /etc/os-release && echo "$VERSION_CODENAME") stable" | \
      sudo tee /etc/apt/sources.list.d/docker.list > /dev/null
+
+   # Add the repository to Apt sources (new deb822 format, introduced with Trixie):
```
eshattow: deb822 format was introduced in apt version 1.1 ten years before the release of Trixie

LLdaniel: oh :sweat_smile: , thanks for that comment. 
Did not know that, I can adjust the comment together with possible other review comments. 

Maybe we should just say:
```
 # Add the repository to Apt sources (deb822 format)
```

content/manuals/engine/install/debian.md
```diff
@@ -114,11 +114,21 @@ Docker from the repository.
...
+   echo -e \
+   "Types: deb\n\
+   URIs: https://download.docker.com/linux/debian/\n\
+   Suites: $(. /etc/os-release && echo "$VERSION_CODENAME")\n\
+   Components: stable\n\
+   Signed-By: /etc/apt/keyrings/docker.gpg" | \
```
stalb:
```suggestion
   Signed-By: /etc/apt/keyrings/docker.asc" | \
```

Just a typo correction

eshattow: Wrong way around. This should be a machine-readable file not ASCII export.

content/manuals/engine/install/debian.md
eshattow: Drop this comment and the use of non-deb822 format sources

LLdaniel: Addressed with [7e2650e](https://github.com/docker/docs/pull/23249/commits/7e2650eb493e67c2fdddef42546bb71f265eb8ae)

(Hope it is still ok to keep the original comment, which just stated `# Add the repository to Apt sources:`. The paragraph seemed to me a bit lonely without the comment. At least explanations about (non-)deb822 are gone now).

content/manuals/engine/install/debian.md
LLdaniel: All right, then `.gpg` should be fine. Kept that in the newest commit [7e2650e](https://github.com/docker/docs/pull/23249/commits/7e2650eb493e67c2fdddef42546bb71f265eb8ae)

stalb: @LLdaniel @eshattow, the fact is, if you download an _ASCII armored key_, you have to _dearmor_ it before using the binary version of the key.

So either you have to use the `docker.asc` (armored ASCII version), or you have to dearmor it first to get the binary `docker.gpg` version of the key!

LLdaniel: Thanks for clarifying! Now I switched back to the originally used `.asc`. 
But if you prefer ` gpg --dearmor` I can also include that. Just let me know. :+1:

content/manuals/engine/install/debian.md
```diff
@@ -115,10 +115,14 @@ Docker from the repository.
...
-     "deb [arch=$(dpkg --print-architecture) signed-by=/etc/apt/keyrings/docker.asc] {{% param "download-url-base" %}} \
-     $(. /etc/os-release && echo "$VERSION_CODENAME") stable" | \
-     sudo tee /etc/apt/sources.list.d/docker.list > /dev/null
+   echo -e \
+   "Types: deb\n\
+   URIs: https://download.docker.com/linux/debian/\n\
+   Suites: $(. /etc/os-release && echo "$VERSION_CODENAME")\n\
+   Components: stable\n\
+   Signed-By: /etc/apt/keyrings/docker.asc" | \
+   sudo tee /etc/apt/sources.list.d/docker.sources > /dev/null
```
dvdksn: Thank you for raising this PR!

Maybe for readability we can make this a heredoc instead:

```suggestion
   sudo tee /etc/apt/sources.list.d/docker.sources <<EOF
   Types: deb
   URIs: {{% param "download-url-base" %}}
   Suites: $(. /etc/os-release && echo "$VERSION_CODENAME")
   Components: stable
   Signed-By: /etc/apt/keyrings/docker.asc
   EOF
```

Normally we redirect `tee` to /dev/null but in this case since it involves variable interpolation I think printing to stdout makes sense!

Image

LLdaniel: Thanks for your review. Sure, I agree upon the readability argument :+1: 
Fixed with [e1d7d33](https://github.com/docker/docs/pull/23249/commits/e1d7d339b74775e61fff5170a9a3f5a893a11cc6)

eshattow:
> Looks good to me, thanks. Maybe @thaJeztah / @tianon could you help double-check? 🙇

It is wrong though. Look at https://wiki.debian.org/DebianRepository/UseThirdParty for the recommendations how this should be done.

tianon:
It looks fine to me for documentation given to users - @eshattow can you be more specific about what you'd change?

tianon:
(I don't think you meant for them to, but your comments in this thread are coming off pretty aggressive FYI)

content/manuals/engine/install/debian.md
tianon: Agreed, the armored version is much nicer for users to manage (and is what I personally use on my own systems).

tianon: For even more added context from the wiki page: 

> The reason why we avoid ASCII-Armored files is that they can only be used by SecureApt in version 1.4 or later (which appeared in stretch).

(even stretch _LTS_ has been EOL since 2022, so we're extremely fine here)

stalb:
@tianon, I think @eshattow just says it's better to use the de-armored version of the key (docker.gpg), since the ASCII-armored version will not work in versions older than Debian 12 or Ubuntu 22.04, which indeed is a real problem.

Simply add a line to generate the de-armored  version of the key....

eshattow:
> It looks fine to me for documentation given to users - @eshattow can you be more specific about what you'd change?

There should not be ascii-armoured key installed to the user system. The guide recommended use of curl and sequoia are recent progressions to coincide with the stable release of Debian 13 Trixie.

> (I don't think you meant for them to, but your comments in this thread are coming off pretty aggressive FYI)

I appreciate your code review. If the contributor for this pull request does not implement the required suggestions there is a pull request I have filed to get this done. I can follow up with any needed changes.

tianon:
> There should not be ascii-armoured key installed to the user system.

I'm not sure where you're getting such a strong exclusion from in that wiki page (if it's there, it's incorrect and should be fixed :sweat_smile:).

From the relevant documentation in APT itself: :eyes:

https://manpages.debian.org/trixie/apt/apt-secure.8.en.html#USER_CONFIGURATION

> Alternatively, keys may be placed in /etc/apt/keyrings for local keys, or /usr/share/keyrings for keys managed by packages, and then referenced by Signed-By: /etc/apt/keyrings/example-archive-keyring.asc option in a .sources file or using deb [signed-by=/etc/apt/keyrings/example-archive-keyring.asc] ... in the legacy .list format. This may be useful for APT versions prior to 2.4, which do not support embedded keys. ASCII-armored keys must use an extension of .asc, and unarmored keys an extension of .gpg.

---

> the ASCII-armored version will not work in versions older than Debian 12 or Ubuntu 22.04

I'm confused by this - Ubuntu 20.04 is EOL (back in May), so 22.04 is the oldest Ubuntu that this needs to work on, and it does: :sweat_smile:

```console
$ docker run -it --rm --pull=always ubuntu:22.04
22.04: Pulling from library/ubuntu
Digest: sha256:09506232a8004baa32c47d68f1e5c307d648fdd59f5e7eaa42aaf87914100db3
Status: Image is up to date for ubuntu:22.04
root@ffb1c8368383:/# apt-get update &> /dev/null
root@ffb1c8368383:/# apt-get install -y ca-certificates curl &> /dev/null
root@ffb1c8368383:/# install -m 0755 -d /etc/apt/keyrings
root@ffb1c8368383:/# curl -fsSL https://download.docker.com/linux/ubuntu/gpg -o /etc/apt/keyrings/docker.asc
root@ffb1c8368383:/# chmod a+r /etc/apt/keyrings/docker.asc
root@ffb1c8368383:/# tee /etc/apt/sources.list.d/docker.sources <<EOF
> Types: deb
> URIs: https://download.docker.com/linux/ubuntu
> Suites: $(. /etc/os-release && echo "$VERSION_CODENAME")
> Components: stable
> Signed-By: /etc/apt/keyrings/docker.asc
> EOF
Types: deb
URIs: https://download.docker.com/linux/ubuntu
Suites: jammy
Components: stable
Signed-By: /etc/apt/keyrings/docker.asc
root@ffb1c8368383:/# apt-get update
Get:1 https://download.docker.com/linux/ubuntu jammy InRelease [48.5 kB]
Get:2 https://download.docker.com/linux/ubuntu jammy/stable amd64 Packages [70.3 kB]
Hit:3 http://archive.ubuntu.com/ubuntu jammy InRelease                         
Hit:4 http://security.ubuntu.com/ubuntu jammy-security InRelease               
Hit:5 http://archive.ubuntu.com/ubuntu jammy-updates InRelease
Hit:6 http://archive.ubuntu.com/ubuntu jammy-backports InRelease
Fetched 119 kB in 43s (2767 B/s)
Reading package lists... Done
```

Similarly, Debian 11 is the oldest version of Debian that this needs to work on (even LTS for Debian 10 ended in June of 2024), and it does there too:

```console
$ docker run -it --rm --pull=always debian:11
11: Pulling from library/debian
Digest: sha256:05bc4300f29708b0e6f4fa006ed3d16930baae91613416937c39e79dd9a095ac
Status: Image is up to date for debian:11
root@f7e21b97024e:/# apt-get update &> /dev/null
root@f7e21b97024e:/# apt-get install -y ca-certificates curl &> /dev/null
root@f7e21b97024e:/# install -m 0755 -d /etc/apt/keyrings
root@f7e21b97024e:/# curl -fsSL https://download.docker.com/linux/debian/gpg -o /etc/apt/keyrings/docker.asc
root@f7e21b97024e:/# chmod a+r /etc/apt/keyrings/docker.asc
root@f7e21b97024e:/# tee /etc/apt/sources.list.d/docker.sources <<EOF
> Types: deb
> URIs: https://download.docker.com/linux/debian
> Suites: $(. /etc/os-release && echo "$VERSION_CODENAME")
> Components: stable
> Signed-By: /etc/apt/keyrings/docker.asc
> EOF
Types: deb
URIs: https://download.docker.com/linux/debian
Suites: bullseye
Components: stable
Signed-By: /etc/apt/keyrings/docker.asc
root@f7e21b97024e:/# apt-get update
Hit:1 http://deb.debian.org/debian bullseye InRelease
Hit:2 http://deb.debian.org/debian-security bullseye-security InRelease
Hit:3 http://deb.debian.org/debian bullseye-updates InRelease
Get:4 https://download.docker.com/linux/debian bullseye InRelease [43.0 kB]
Get:5 https://download.docker.com/linux/debian bullseye/stable amd64 Packages [63.9 kB]
Fetched 107 kB in 0s (297 kB/s)     
Reading package lists... Done
```

So even with [my "Debian hat"](https://qa.debian.org/developer.php?login=tianon) on, I think this change is great as-is and that recommending and using both deb822 _and_ armored keys is very sane and very reasonable for all actively supported versions of both Debian _and_ Ubuntu. :+1: :bow:

PR #23279, opened 2025-08-20
fix go.mod

crazy-max:
## Description

It should be the lowest compatible version.

## Reviews

- [ ] Technical review
- [ ] Editorial review
- [ ] Product review

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/docker/docs
 
-go 1.24.5
+go 1.24.0
 
 require (
 	github.com/docker/buildx v0.27.0 // indirect


crazy-max:
```
  > [validate-vendor 1/1] RUN --mount=target=/context   --mount=type=bind,from=vendor,target=/out   --mount=target=.,type=tmpfs <<EOT (set -e...):
0.823 ERROR: Vendor result differs. Please vendor your package with "make vendor"
0.910  M go.mod
```

Ok that's because model-cli and mcp-gateway enforces it:
* https://github.com/docker/mcp-gateway/blob/9da2311fa85dc85282b6ebc5d8ec5825a5a9ae73/go.mod#L3
* https://github.com/docker/model-cli/blob/ebb4723662c628aac216b09d2fe73458511f76cc/go.mod#L3-L5

PR #23286, opened 2025-08-21
dmr: update cli

ArthurFlag:


sarahsanders-docker (approved):
Not a blocker, but there is a wonky visual thing happening on `docker model package` page: https://deploy-preview-23286--docsdocker.netlify.app/reference/cli/docker/model/package/

I don't think this PR caused it, just the long description causing a scroll bar to appear

ArthurFlag:
> Not a blocker, but there is a wonky visual thing happening on `docker model package` page: https://deploy-preview-23286--docsdocker.netlify.app/reference/cli/docker/model/package/
> 
> I don't think this PR caused it, just the long description causing a scroll bar to appear

"code" tags do not allow wrapping by default so we get a horizontal bar as a result. I'd argue it's so rare to get these long code examples that we don't need to update the wrapping behaviour but let's keep an eye on this 🙂

PR #23290, opened 2025-08-21
docs(mcp): update landing page

ArthurFlag:
Add illustrations and tweak example.

### DIFF
--- a/content/manuals/ai/mcp-catalog-and-toolkit/toolkit.md
+++ b/content/manuals/ai/mcp-catalog-and-toolkit/toolkit.md
@@ -17,6 +17,10 @@ from tool usage by offering secure defaults, one-click setup, and support for a
 growing ecosystem of LLM-based clients. It is the fastest way from MCP tool
 discovery to local execution.
 
+> [!NOTE]
+> If you need to run your own MCP gateway,
+> see [Docker MCP Gateway](../mcp-gateway/_index.md).
+
 ## Key features
 
 - Cross-LLM compatibility: Instantly works with Claude Desktop, Cursor, Continue.dev, and [Gordon](/manuals/ai/gordon/_index.md).


content/manuals/ai/mcp-catalog-and-toolkit/images/mcp-overview.svg
liuziyue, line 1: The middle rectangle "MCP Toolkit" seems to be "MCP Servers" to me. MCP Toolkit is a collective feature name for using the MCP. i.e. we're using this name in the Docker Desktop tab which is the entry point to all MCP features

content/manuals/ai/mcp-catalog-and-toolkit/images/mcp-servers-overview.svg
liuziyue, line 1: Should we have some visual representation of the MCP gateway here?

content/manuals/ai/mcp-catalog-and-toolkit/images/mcp-overview.svg
ArthurFlag: I'm trying to show what the Docker parts documented in this section do so I don't think we should have an "MCP server" box since Docker wraps this up in the Docker Tookit. That's also why the Toolkit connects to the clients: it manages the servers and connects to the clients. That might also be confusing with the catalog since the catalog is a list of MCP servers.

The Toolkit is also considered as the gateway sometimes or most specifically the gateway it the backend so that's harder to represent here. Not sure we need this in this landing page, to keep it simple. Wdyt?

content/manuals/ai/mcp-catalog-and-toolkit/images/mcp-servers-overview.svg
ArthurFlag: That's tricky because the Toolkit is the frontend of the gateway and we're in the Toolkit docs page. I think we can represent it in the Gateway docs instead.

liuziyue (approved):

sarahsanders-docker (approved):
images look great! :)

LGTM

PR #23328, opened 2025-08-29
security: add RSS feed option

sarahsanders-docker:
Do not merge, we need to configure some behavior rules in CloudFront

## Description
Adds an RSS feed subscription option to the security announcements page. We can reuse this shortcode for other release notes or announcements pages if we want! 

## Related issues or tickets
https://docker.atlassian.net/browse/ENGDOCS-2939

## Reviews
- [ ] Editorial review


usha-mandya (approved):
LGTM

PR #23330, opened 2025-08-29
add redirect

craig-osterhout:
## Description

Added go redirect as a workaround to GA modifying a pdf link causing it to 404 on www.

It's kind of hacky (and not very sure it's going to work), but only necessary for like 3 more months, then can be nuked. Very open to other alternatives.

For example, https://www.docker.com/legacy-pricing/ 
gets turned into 
https://www.docker.com/legacy-pricing/?_gl=1*hre2s0*_gcl_au*NzE2MTg4NDg0LjE3NTUwNDA1MjU.*_ga*MjM4NDM0NjYxLjE3MjcxMDM2OTk.*_ga_XJWPQMJYHQ*czE3NTY1MDE3MzgkbzEwMDMkZzEkdDE3NTY1MDE3OTYkajIkbDAkaDA.

## Reviews

- [ ] Editorial review

### DIFF
--- a/content/manuals/subscription/details.md
+++ b/content/manuals/subscription/details.md
@@ -141,7 +141,7 @@ Legacy Docker Pro includes:
 - [Auto Builds](/docker-hub/builds/) with 5 concurrent builds
 - 300 [Vulnerability Scans](/docker-hub/vulnerability-scanning/)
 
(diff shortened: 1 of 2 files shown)


craig-osterhout:
aaaaaand it didn't work.

PR #23338, opened 2025-09-01
helm push need the username

benoittgt:
I think `docker` is too generic and we need to specify the default with is the username.

### DIFF
--- a/content/manuals/docker-hub/repos/manage/hub-images/oci-artifacts.md
+++ b/content/manuals/docker-hub/repos/manage/hub-images/oci-artifacts.md
@@ -84,17 +84,17 @@ Steps:
 3. Sign in to Docker Hub with Helm, using your Docker credentials.
 
    ```console
-   $ helm registry login registry-1.docker.io -u hubuser
+   $ helm registry login registry-1.docker.io -u <hubuser>
    ```
 
 4. Push the chart to a Docker Hub repository.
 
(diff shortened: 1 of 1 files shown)


craig-osterhout:
Thanks for flagging this. A couple issues related to this update are the screenshots and other examples on this page should be updated for consistency. In addition, there is a security concern with the existing content using an actual hub username. Please feel free to update this PR, or I can throw a fix up in a few days.

benoittgt:
Thanks @craig-osterhout for the review. I've changed strategy and chosen a more obvious placeholder username. So I do not need to update screenshot. I think it makes the whole doc more explicit about which fields is expected and avoid mixing regular url-path and url-parameter.

content/manuals/docker-hub/repos/manage/hub-images/oci-artifacts.md
```diff
@@ -78,23 +78,24 @@ Steps:
 
    ```console
    $ helm package demo
-   Successfully packaged chart and saved it to: /Users/hubuser/demo-0.1.0.tgz
+   Successfully packaged chart and saved it to: demo-0.1.0.tgz
```
benoittgt, line 81: I think there is no need to have the exact helm output

craig-osterhout (approved):
Thank you. LGTM

PR #23359, opened 2025-09-04
better demonstrate benefits of using heredocs in shell-form

### LINKED ISSUE #23350: Better example for demonstrating the benefit of using heredocs in shell-form
### Is this a docs issue?

### Type of issue

Other

### Description

Within the dockerfile reference section on shell-form - <https://docs.docker.com/reference/dockerfile/#shell-form>, there is a section that explains how you can use heredocs for shell-form structuring:

>You can also use heredocs with the shell form to break up supported commands.
>```
>[RUN](https://docs.docker.com/reference/dockerfile/#run) <<EOF
>source $HOME/.bashrc && \
>echo $HOME
>EOF
>```

I think the benefit of using the heredocs format would be clearer here if the `&& \` was omitted, like so:

```
[RUN](https://docs.docker.com/reference/dockerfile/#run) <<EOF
source $HOME/.bashrc
echo $HOME
EOF
```

### Location

https://docs.docker.com/reference/dockerfile/#shell-form

damonholden:
## Related issues or tickets

Resolves #23350.

### DIFF
--- a/_vendor/github.com/moby/buildkit/frontend/dockerfile/docs/reference.md
+++ b/_vendor/github.com/moby/buildkit/frontend/dockerfile/docs/reference.md
@@ -568,8 +568,8 @@ You can also use heredocs with the shell form to break up supported commands.
 
 ```dockerfile
 RUN <<EOF
-source $HOME/.bashrc && \
-echo $HOME
+  source $HOME/.bashrc
+  echo $HOME
 EOF
 ```
 


crazy-max:
Thanks but this change must be done upstream: https://github.com/moby/buildkit/blob/master/frontend/dockerfile/docs/reference.md

So it can be vendored here if you can open a pull request there :pray:

damonholden:
Understood. Created identical PR on the moby/builkit repo - https://github.com/moby/buildkit/pull/6223.

PR #23360, opened 2025-09-04
bake: add pattern matching for targets input

aevesdocker:
## Description

Adds https://github.com/docker/buildx/pull/3280

## Reviews

- [ ] Technical review
- [ ] Editorial review
- [ ] Product review

### DIFF
--- a/content/manuals/build/bake/targets.md
+++ b/content/manuals/build/bake/targets.md
@@ -107,14 +107,19 @@ Supported patterns:
 - `?` matches any single character
 - `[abc]` matches any character in brackets
 
+> [!NOTE]
+>
+> Always wrap wildcard patterns in quotes. Without quotes, your shell will expand the
+> wildcard to match files in the current directory, which usually causes errors.
+
 Examples: 
 
 ```console
 # Match all targets starting with 'foo-'
-$ docker buildx bake foo-*
+$ docker buildx bake "foo-*"
 
 # Match all targets
-$ docker buildx bake *
+$ docker buildx bake "*"
 
 # Matches: foo-baz, foo-caz, foo-daz, etc.
 $ docker buildx bake "foo-?az"
@@ -126,10 +131,10 @@ $ docker buildx bake "[fb]oo-bar"
 $ docker buildx bake "mtx-a-b-*"
 ``` 
 
-You can also combine multiple patters:
+You can also combine multiple patterns:
 
 ```console
-$ docker buildx bake "*" "tests"
+$ docker buildx bake "foo*" "tests"
 ```
 
 ## Additional resources


content/manuals/build/bake/targets.md
```diff
@@ -96,6 +96,42 @@ command.
...
+- `[abc]` matches any character in brackets
+
+Examples: 
+
+```console
+# Match all targets starting with 'foo-'
+$ docker buildx bake foo-*
+
+# Match all targets
+$ docker buildx bake *
```
thaJeztah: We should probably call out that the value must be quoted to prevent the user's shell from expanding.  Without quotes, the shell will expand to files in the current directory, which will fail in most cases, e.g.;

```bash
docker buildx bake *
[+] Building 0.0s (1/1) FINISHED                                                                                docker:desktop-linux
 => [internal] load local bake definitions                                                                                      0.0s
 => => reading docker-bake.hcl 4.02kB / 4.02kB                                                                                  0.0s
ERROR: failed to find target api
```

crazy-max: Yes should be quoted, a note would be good to explain this

content/manuals/build/bake/targets.md
```diff
@@ -96,6 +96,42 @@ command.
...
+$ docker buildx bake "foo-?az"
+
+# Matches: foo-bar, boo-bar
+$ docker buildx bake "[fb]oo-bar"
+
+# Matches: mtx-a-b-d, mtx-a-b-e, mtx-a-b-f
+$ docker buildx bake "mtx-a-b-*"
+``` 
+
+You can also combine multiple patters:
```
crazy-max:
```suggestion
You can also combine multiple patterns:
```

content/manuals/build/bake/targets.md
```diff
@@ -96,6 +96,42 @@ command.
...
+
+```console
+$ docker buildx bake "*" "tests"
```
crazy-max: `tests` would not be necessary here as we glob everything already. Maybe instead:

```suggestion
$ docker buildx bake "foo*" "tests"
```

content/manuals/build/bake/targets.md
```diff
@@ -96,6 +96,42 @@ command.
...
+
+- `*` matches any sequence of characters
+- `?` matches any single character
+- `[abc]` matches any character in brackets
+
+Examples: 
+
+```console
+# Match all targets starting with 'foo-'
+$ docker buildx bake foo-*
```
crazy-max:
```suggestion
$ docker buildx bake "foo-*"
```

content/manuals/build/bake/targets.md
crazy-max:
```suggestion
$ docker buildx bake "*"
```

PR #23386, opened 2025-09-10
feat: Docker Debug no longer requires a subscription

joe0BAB:
## Description

We have removed the paywall on docker debug.

## Reviews

- [x] Technical review
- [ ] Editorial review
- [ ] Product review

### DIFF
--- a/data/summary.yaml
+++ b/data/summary.yaml
@@ -153,8 +153,7 @@ Docker Offload:
 docker compose alpha:
   availability: Experimental
 Docker Debug:
-  subscription: [Pro, Team, Business]
-  requires: Docker Desktop [4.33.0](/manuals/desktop/release-notes.md#4330) and later
+  requires: Docker Desktop 4.46 and later
 Docker Desktop Archlinux:
   availability: Experimental
 Docker Desktop CLI:


joe0BAB:
As of DD 4.46 release, this feature no longer requires any subscription. I assumed my change would remove that hint that we currently show (this thing shown in the screenshot is no longer needed as of 4.46)
image

aevesdocker:
yeh your change works in the preview. My question was more about is this change backward compatible. So if a user is on 4.45 or 4.44 etc, will they still need a subscription to use Docker Debug? if so, we need to adapt the info further. 

I realise my initial phrasing was confusing. I meant versions below 4.46

joe0BAB:
Ah right, yes for 4.33 - 4.45 a subscription is needed. Do we have any standard way to express this? Or should I just adjust the sentence in `requires` to say this?

aevesdocker:
Yes, I think the simplest is just to add in the `requires` field "Docker Desktop 4.46 and later. For Docker Desktop versions 4.45.0 and earlier, you must have a Pro, Team, or Business subscription"

aevesdocker:
Closing PR as this change was rolled back last min.

PR #23405, opened 2025-09-12
engine/install: add note for service start on deb flavors

austinvazquez:
## Description

This change is to address some feedback from a Ubuntu user that installed Docker Engine using the apt repository but had to start the service before they were able to run containers.

While autostart is the convention on Debian systems, some users might have this behavior disabled for more fine grained control. This change adds a note to the installation step across all the debian flavors to note the service should start by default with explicit instructions how to enable the service if not.

## Reviews

- [ ] Technical review - @vvoland
- [ ] Editorial review - @usha-mandya 
- [ ] Product review

### DIFF
--- a/content/manuals/engine/install/debian.md
+++ b/content/manuals/engine/install/debian.md
@@ -173,6 +173,21 @@ Docker from the repository.
    {{< /tab >}}
    {{< /tabs >}}
 
+    > [!NOTE]
+    >
+    > The Docker service starts automatically after installation. To verify that
+    > Docker is running, use:
+    > 
+    > ```console
+    > $ sudo systemctl status docker
+    > ```
+    >
+    > Some systems may have this behavior disabled and will require a manual start:
+    >
+    > ```console
+    > $ sudo systemctl start docker
+    > ```
+
 3. Verify that the installation is successful by running the `hello-world` image:
 
    ```console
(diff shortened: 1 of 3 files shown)


content/manuals/engine/install/debian.md
```diff
@@ -173,6 +173,27 @@ Docker from the repository.
...
+    > Some systems may have this behavior disabled and will require a manual start:
+    >
+    > ```console
+    > $ sudo systemctl start docker
+    > ```
+    >
+    > To have Docker start automatically when your system boots, use:
+    >
+    > ```console
+    > $ sudo systemctl enable --now docker
```
vvoland: Technically only `enable` is needed to enable automatic start on boot.
```suggestion
    > $ sudo systemctl enable docker
```

The `--now` is just a shorthand for `systemctl enable` + `systemctl start`, so if the user already did `start` in the previous step then `--now` isn't needed

austinvazquez: Good point, let me simplify here. Addressed in https://github.com/docker/docs/pull/23405/commits/9805dcdadacc3c5504127a8f56266fc93e8d615f

content/manuals/engine/install/debian.md
```diff
@@ -173,6 +173,27 @@ Docker from the repository.
...
+    > ```console
+    > $ sudo systemctl start docker
+    > ```
+    >
+    > To have Docker start automatically on system boot, use:
+    >
+    > ```console
+    > $ sudo systemctl enable docker
+    > ```
+
```
vvoland: I noticed we already have a similar section in the post-install page:

https://github.com/docker/docs/blob/a0142e993867ff6d340d3d6e9ebdfbe787c75da8/content/manuals/engine/install/linux-postinstall.md?plain=1#L98-L121

Perhaps we should just leave a small link to that section (like "Docker doesn't start automatically? See <link>") instead of repeating this for each deb distro.

aevesdocker: we can add a lil' "include" shortcode perhaps, like we do for DD https://github.com/docker/docs/blob/main/content/manuals/desktop/setup/install/linux/debian.md?plain=1#L72

austinvazquez: @vvoland , @aevesdocker , I pushed https://github.com/docker/docs/pull/23405/commits/d5a2ba6817a901df60315b6d88229ab9bf6a00f0 which drops the on-boot informational piece. My thinking was an include would blow up this note quite a bit in two places on the page.

Also the page lists post install in next steps so users should naturally discover that content through the documentation flow, but lmk your thoughts here.

PR #23475, opened 2025-09-30
Updated fedora add repo instructions to the new dnf version

StarmanAkremis:
## Description

the instructions for adding the repo on fedora were using an old version of dnf.


StarmanAkremis:
the documentation has already been updated

PR #23484, opened 2025-10-02
Remove warning issue references to docker/for-mac#7527

bsousaa:
## Description

Removing the warning references to the malware detection issue from last January docker/for-mac#7527. The affected versions are no longer supported and the volume of users reporting the issue is minimal.

## Reviews

- [ ] Technical review
- [ ] Editorial review
- [ ] Product review


bsousaa:
> I was wondering about this the other day, so thank you @bsousaa
> 
> Should we keep this page still https://docs.docker.com/desktop/cert-revoke-solution/?

You're welcome @aevesdocker. I think we should keep that page, since it may be linked from public GitHub issues. Probably we can remove it from the navigation bar?

aevesdocker:
> we can remove it from the navigation bar

Will do that in a follow-up. Consider it done 🫡

PR #23485, opened 2025-10-02
scout: update cli to 1.18.4

craig-osterhout:
## Description

Updated scout cli to 1.18.4
- Got vendor'd content from upstream
- Created new placeholders in `content/reference/cli/docker/scout` for new commands
- Updated release notes

https://deploy-preview-23485--docsdocker.netlify.app/reference/cli/docker/scout/
https://deploy-preview-23485--docsdocker.netlify.app/scout/release-notes/cli/

## Related issues or tickets

DHI-449

## Reviews

- [ ] Editorial review

PR #23506, opened 2025-10-06
scout: fix watch docs

craig-osterhout:
## Description

Fixes the generated scout watch docs to use examples from the upstream markdown, instead of the yaml. When previously updated, only the markdown was updated upstream and not the CLI source which generates the yaml.

- Added new layout for the scout cli, based on the generic cli layout. The new layout uses the more extensive markdown examples if present, otherwise use the examples from the yaml.
- Updated the scout watch topic to use the new layout 

Once fixed upstream, will have a follow up to most likely revert this.

https://deploy-preview-23506--docsdocker.netlify.app/reference/cli/docker/scout/watch/#examples

## Related issues or tickets

Original PR to update the version: #23485. The markdown was updated but not the yaml.
Upstream fix: https://github.com/docker/scout-cli-plugin/pull/713

## Reviews

- [ ] Editorial review

### DIFF
--- a/content/reference/cli/docker/scout/watch.md
+++ b/content/reference/cli/docker/scout/watch.md
@@ -1,10 +1,11 @@
 ---
 datafolder: scout-cli
 datafile: docker_scout_watch
+markdownfile: scout_watch.md
 aliases:
 - /engine/reference/commandline/scout_watch
 title: docker scout watch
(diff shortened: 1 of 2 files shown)


sarahsanders-docker (approved):
LGTM!

PR #23513, opened 2025-10-07
Replace braced expressions list with a table

ngdangtu-vn:
## Description

The table can offer better comparision experience. While the keywords for purpose ("Direct substitution", "Default value", "Required value") is useful but I don't think it is neccessary as the purpose can improvise with the nesting case, so let devs go creative is ok.

(if the keywords are required, I can add another colum on the left for it though)

## Related issues or tickets

none

## Reviews

- [ ] Technical review
- [ ] Editorial review
- [ ] Product review

### DIFF
--- a/content/reference/compose-file/interpolation.md
+++ b/content/reference/compose-file/interpolation.md
@@ -10,17 +10,16 @@ weight: 90
 {{% include "compose/interpolation.md" %}}
 
 For braced expressions, the following formats are supported:
-- Direct substitution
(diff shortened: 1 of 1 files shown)


aevesdocker:
Thanks for the PR @ngdangtu-vn. I agree that this could be made clearer but I'm not sure the table quite works, so I'm closing this PR. I've made a note to revisit these docs shortly. Thanks

PR #23515, opened 2025-10-07
Update extension documentation: the SDK to execute commands on the host is not restricted to binaries listed in the extension metadata.

gtardif:
## Description

Update extension doc for more accurate info about SDK API restrictions

## Reviews

- [ ] Technical review
- [ ] Editorial review
- [ ] Product review


LucaGuerra:
Thank you for this, looks good! I think the same clarification should be added to `content/manuals/extensions/extensions-sdk/architecture/security.md` as well.

PR #23546, opened 2025-10-13
docs(github-actions): add missing "checkout" step before docker build

neilime:
## Description

Fixing documentation for GitHub Actions section: adding missing [`actions/checkout`](https://github.com/actions/checkout) step, before building docker images.

To build images we need the Dockerfile to be checked-out. If we don't call the checkout step first, build is failling as it does not find any file: `ERROR: failed to build: resolve : lstat tests: no such file or directory`

## Related issues or tickets

N/A

## Reviews

- [ ] Technical review
- [ ] Editorial review
- [ ] Product review

### DIFF
--- a/content/manuals/build/ci/github-actions/multi-platform.md
+++ b/content/manuals/build/ci/github-actions/multi-platform.md
@@ -37,6 +37,8 @@ jobs:
       - name: Set up Docker Buildx
         uses: docker/setup-buildx-action@v3
 
+      - uses: actions/checkout@v5
+
       - name: Build and push
         uses: docker/build-push-action@v6
         with:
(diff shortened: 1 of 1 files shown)


crazy-max:
This is not needed as build-push-action builds from Git context by default, see https://github.com/docker/build-push-action/?tab=readme-ov-file#git-context

crazy-max:
> To build images we need the Dockerfile to be checked-out. If we don't call the checkout step first, build is failling as it does not find any file: `ERROR: failed to build: resolve : lstat tests: no such file or directory`

If you have a repro of this, please open an issue on action's repo, thanks.

neilime:
> > To build images we need the Dockerfile to be checked-out. If we don't call the checkout step first, build is failling as it does not find any file: `ERROR: failed to build: resolve : lstat tests: no such file or directory`
> 
> If you have a repro of this, please open an issue on action's repo, thanks.

Ok done: https://github.com/docker/build-push-action/issues/1419

PR #23560, opened 2025-10-15
Note that nftables support was added in moby29

robmry:
## Description

Added a note that nftables support was added in moby 29.0.0.

Moby 29.0 release candidates are now available, looking for feedback, so we should publish these changes.

## Related issues or tickets

- https://github.com/moby/moby/issues/49644

## Reviews

- [ ] Technical review
- [ ] Editorial review
- [ ] Product review

### DIFF
--- a/content/manuals/engine/network/firewall-nftables.md
+++ b/content/manuals/engine/network/firewall-nftables.md
@@ -7,8 +7,8 @@ keywords: network, nftables, firewall
 
 > [!WARNING]
 >
-> Support for nftables is experimental, configuration options, behavior and
-> implementation may all change in future releases.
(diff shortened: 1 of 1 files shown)


content/manuals/engine/network/firewall-nftables.md
```diff
@@ -7,8 +7,8 @@ keywords: network, nftables, firewall
 
 > [!WARNING]
 >
-> Support for nftables is experimental, configuration options, behavior and
-> implementation may all change in future releases.
+> Support for nftables introduced in moby 29.0.0 is experimental, configuration
```
vvoland, line 10: Shouldn't we refer to Docker v29 instead of moby 29.0.0?

PR #23564, opened 2025-10-15
feat: Docker Debug no longer requires a subscription

joe0BAB:
## Description

We have removed the paywall on docker debug. It's shipped in DD 4.48.

## Reviews

- [x] Technical review
- [ ] Editorial review
- [ ] Product review

### DIFF
--- a/data/summary.yaml
+++ b/data/summary.yaml
@@ -153,8 +153,7 @@ Docker Offload:
 docker compose alpha:
   availability: Experimental
 Docker Debug:
-  subscription: [Pro, Team, Business]
-  requires: Docker Desktop [4.33.0](/manuals/desktop/release-notes.md#4330) and later
+  requires: Docker Desktop 4.48 and later. For Docker Desktop versions 4.47.0 and earlier, you must have a Pro, Team, or Business subscription
 Docker Desktop Archlinux:
   availability: Experimental
 Docker Desktop CLI:


joe0BAB:
Follow up from #23386

aevesdocker:
Hey @joe0BAB, I checked with Colin before the 4.48 release that this was to be announced with 4.49 with release notes and everything.

joe0BAB:
yes, that was the initial plan, but it already got shipped with 4.48 (we also missed that in the 4.48 release notes, maybe we want to change them as well)

kiview:
@aevesdocker Do you see any issue with just updating the docs, to correctly represent the technical state of things? It's ok to have a distinct or decoupled marketing push that we sync with a 4.49. release.

Also, this repo is public, so it's all out in the open already after all 😃 

Maybe it helps if @colinhemmings confirms that it's fine to update the docs?

aevesdocker:
More than happy to publish this now @kiview if that's what you want. 

But we can make a bigger 'splash' of it if we postpone to 4.49 as we can also have the release highlights within Desktop advertise this. 

In the past, when features have been released before widely advertised, we've just postponed publishing any docs updates to keep everything in sync and cheekily adjusted the version needed. We can do the same here. 

Let me know

aevesdocker:
@kiview @joe0BAB @colinhemmings What's the plan? Merge now or postpone to 4.49?

joe0BAB:
Let's merge!

PR #23577, opened 2025-10-17
Update garbage-collection.md

open-documents:
## Reviews

- [ ] Technical review
- [x] Editorial review
- [ ] Product review


open-documents:
With “-7GB of which is ‘stale’ cache -4GB is other, more valuable cache”,the first policy is "Find 'stale' cache records that haven't been used in the past 48 hours, and delete records until there's maximum 5GB of 'stale' cache left." and the original refernce said "A GC sweep would delete 5GB of stale cache as part of the 1st policy, with a remainder of 6GB", so why the size of collected 'stale' cache is not 7GB-5GB=2GB, but the 5GB.

PR #23604, opened 2025-10-24
Add sudo -E flag to curl command for proxy support

Wenzel:
## Description

The curl command needs the `-E` flag with sudo to preserve environment variables, particularly `HTTP_PROXY` and `HTTPS_PROXY` settings. Without this flag, users behind corporate proxies cannot download Docker's GPG key.

### DIFF
--- a/content/manuals/engine/install/ubuntu.md
+++ b/content/manuals/engine/install/ubuntu.md
@@ -129,7 +129,7 @@ Docker from the repository.
    sudo apt-get update
    sudo apt-get install ca-certificates curl
    sudo install -m 0755 -d /etc/apt/keyrings
-   sudo curl -fsSL {{% param "download-url-base" %}}/gpg -o /etc/apt/keyrings/docker.asc
+   sudo -E curl -fsSL {{% param "download-url-base" %}}/gpg -o /etc/apt/keyrings/docker.asc
    sudo chmod a+r /etc/apt/keyrings/docker.asc
 
    # Add the repository to Apt sources:


content/manuals/engine/install/ubuntu.md
```diff
@@ -129,7 +129,7 @@ Docker from the repository.
    sudo apt-get update
    sudo apt-get install ca-certificates curl
    sudo install -m 0755 -d /etc/apt/keyrings
-   sudo curl -fsSL {{% param "download-url-base" %}}/gpg -o /etc/apt/keyrings/docker.asc
+   sudo -E curl -fsSL {{% param "download-url-base" %}}/gpg -o /etc/apt/keyrings/docker.asc
```
thaJeztah, line 132: Wondering if we should change this so that `curl` isn't executed as root;


```bash
curl -fsSL "{{% param "download-url-base" %}}/gpg" | sudo tee /etc/apt/keyrings/docker.asc > /dev/null
```

or with an intermediate file, but that requires a cleanup after;

```bash
curl -fsSL "{{% param "download-url-base" %}}/gpg" -o docker.asc
sudo install -o root -g root -m 0644 docker.asc /etc/apt/keyrings/docker.asc
rm -f docker.asc
```


@dvdksn @vvoland any thoughts? If we change, we should also update the `debian.md` accordingly

dvdksn: Hmm, we previously agreed that `sudo curl` was an acceptable tradeoff - see https://github.com/docker/docs/pull/19138/files#r1467055019

I think `sudo -E curl` still seems OK to me but no strong opinions. If not `sudo curl` then I'd go for `sudo tee > /dev/null`

vvoland: Not a big deal I think, but if we want something shorter than tee we could use: `sudo cat - >/etc/apt/keyrings/docker.asc`

or  `sudo cat ->/etc/apt/keyrings/docker.asc`  if we want to be funny and use the _cat goes to_ operator (which is a close relative to the  ["downto" operator in C](https://stackoverflow.com/questions/1642028/what-is-the-operator-in-c-c))

hongquan:
Recent Ubuntu switched to `sudo-rs` which doesn't support `-E` option.
I think Debian will also adopt `sudo-rs` because the team said that Rust is now a hard dependency of Debian project.

PR #23610, opened 2025-10-25
docs: update Debian installation instructions to include APT sources for Debian 12 and newer

oconnordaniel:


oconnordaniel:
From the comments here. 
https://github.com/docker/docs/issues/23231#issuecomment-3369233157

Issue #23524, opened 2025-10-08
Instructions lead to empty repository for arm64 architecture


dordepavlovic:
### Is this a docs issue?

- [x] My issue is about the documentation content or website

### Type of issue

Information is incorrect

### Description

Followed instructions to set up and install Docker Engine from [Docker's apt repository](https://docs.docker.com/engine/install/debian/#install-using-the-repository), but the packages for the architecture `arm64` are missing. It seems that the repo only holds `armhf` packages.

The page clearly suggests that `arm64` architecture is supported.

The repo for the bullseye distribution is [here](https://download.docker.com/linux/raspbian/dists/bullseye/).

### Location

https://docs.docker.com/engine/install/debian/

eshattow:
Debian is not Raspbian. The [installation guide for Raspbian](https://docs.docker.com/engine/install/raspberry-pi-os/) clearly calls out armhf support. What do you want?

PR #23612, opened 2025-10-27
Debian install doc updates (block indent, apt UI, Deb822 and cleanup)

### LINKED ISSUE #15382: Compliance with Deb822: Deprecated sources.list use .sources instead

### LINKED ISSUE #23231: APT install instructions for Debian 13

eshattow:
Supersedes #23249
Closes #23231
Closes #15382 
Related to #23524 but I'm not sure if that was still any problem today.

### DIFF
--- a/content/manuals/engine/install/debian.md
+++ b/content/manuals/engine/install/debian.md
@@ -58,9 +58,9 @@ before you install the official version of Docker Engine.
 
 The unofficial packages to uninstall are:
 
-- `docker.io`
 - `docker-compose`
 - `docker-doc`
+- `docker.io`
 - `podman-docker`
 
 Moreover, Docker Engine depends on `containerd` and `runc`. Docker Engine
@@ -71,7 +71,12 @@ conflicts with the versions bundled with Docker Engine.
 Run the following command to uninstall all conflicting packages:
 
 ```console
-$ for pkg in docker.io docker-doc docker-compose podman-docker containerd runc
+$ for pkg in containerd \
+docker-compose \
+docker-doc \
+docker.io \
+podman-docker \
+runc
 do sudo apt remove $pkg
 done
 ```
@@ -91,8 +96,8 @@ You can install Docker Engine in different ways, depending on your needs:
   [Docker Desktop for Linux](/manuals/desktop/setup/install/linux/_index.md). This is
   the easiest and quickest way to get started.
 
-- Set up and install Docker Engine from
-  [Docker's `apt` repository](#install-using-the-repository).
+- Set up and install Docker Engine from the
+  [Docker `apt` repository](#install-using-the-repository).
 
 - [Install it manually](#install-from-a-package) and manage upgrades manually.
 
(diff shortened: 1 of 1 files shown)


content/manuals/engine/install/debian.md
```diff
@@ -71,10 +71,15 @@ conflicts with the versions bundled with Docker Engine.
 Run the following command to uninstall all conflicting packages:
 
 ```console
+sudo apt remove containerd \
+docker-compose \
+docker-doc \
+docker.io \
+podman-docker \
+runc
```
dvdksn, line 79: The `for` loop was intentional to handle the case where one of the packages did not exist in the repository, whereby the command would fail. #17405

However I am not sure that's really necessary anymore. cc @thaJeztah

content/manuals/engine/install/debian.md
```diff
@@ -99,102 +104,102 @@ You can install Docker Engine in different ways, depending on your needs:
...
+
+2.  Install the Docker packages.
+
+    {{< tabs >}}
+    {{< tab name="Latest" >}}
+
+    To install the latest version, run:
+
+    ```console
+    sudo apt install containerd.io \
```
dvdksn: The `console` code blocks should include the prompt character (whereas the previous one, `bash`, should not)

Here and all the examples below.

Also could you not indent these steps 4 spaces? We use three-space indentation for steps throughout. Thank you!

content/manuals/engine/install/debian.md
eshattow: I've reverted the removal of `for` loop. The proper way to do this would be to instrument dpkg output and then only run the `apt remove` once with the filtered selection, but that's bikeshed stuff I don't want to do right now.

content/manuals/engine/install/debian.md
eshattow: ACK.  I don't understand why it's bash here and console there, so I've replaced everything with console. If you have something more to direct me I would appreciate to know. Also the requested indent is applied. I need to be compatible for the heredoc so I do intentionally continue with my idea for command output indent and double-newline rather than command input indent on multi-line. It is possible to have clearspace tab character for heredoc but would be fragile if someone doesn't know about the depths of heredoc syntax in future, that would break easily for anyone substituting out tab characters.

dvdksn:
Implemented in #23249

PR #23620, opened 2025-10-28
Fix grammar in environment variables section

### LINKED ISSUE #23616: Wrong grammar
### Is this a docs issue?

### Type of issue

Other

### Description

https://docs.docker.com/reference/cli/docker/
In the environmental variables section there is a wrong grammar construction:
"The following list of environment variables are supported by the docker command line:"
while it should be:
"The following list of environment variables is supported by the docker command line:"

### Location

https://docs.docker.com/reference/cli/docker/

### Suggestion

Change are to is

ahopp:
## Summary

Corrects grammar error in the environment variables section of the Docker CLI documentation. Specifically, the subject of the sentence is "list" (singular), not "environment variables" (plural). The phrase "of environment variables" is a prepositional phrase that modifies "list" but doesn't change the fact that the subject is singular. So updating. 

## Changes

Changed "The following list of environment variables **are** supported" to "The following list of environment variables **is** supported" 

## Related Issue

Fixes #23616

### DIFF
--- a/content/get-started/docker-concepts/the-basics/what-is-docker-compose.md
+++ b/content/get-started/docker-concepts/the-basics/what-is-docker-compose.md
@@ -81,7 +81,7 @@ Follow the instructions to run the to-do list app on your system.
 
     If this feels overwhelming, don't worry! You'll get there!
 
-5. With everything now up and running, you can open [http://localhost:3000](http://localhost:3000) in your browser to see the site. Feel free to add items to the list, check them off, and remove them.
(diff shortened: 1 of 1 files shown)


ahopp:
@sarahsanders-docker @dvdksn Can you take a look?

data/engine-cli/docker.yaml
```diff
@@ -29,7 +29,7 @@ long: |-
 
     ### Environment variables
 
-    The following list of environment variables are supported by the `docker` command
+    The following list of environment variables is supported by the `docker` command
```
dvdksn, line 32: good catch but unfortunately this file is generated, from this: https://github.com/docker/cli/blob/master/docs/reference/commandline/docker.md

And reading this just now, maybe we could even clarify what we mean here with:

```
The following environment variables configure the Docker CLI:
```

Would you mind submitting a PR to docker/cli and dropping it from here?

ahopp: Closing now. Thanks.

PR #23627, opened 2025-10-29
fix: re-add lint rule

sarahsanders-docker:
## Description
- Reverts a change we made in E2B launch that was blocking


dvdksn (changes requested):
Oops, approved too fast :-) Looks like some files need fixing!

sarahsanders-docker:
> * build / validate (lint) (pull_request)

yes they do! I will get those updated :)

dvdksn (approved):

PR #23634, opened 2025-10-31
mcp: gateway go link

Daniel-Kolev:
## Description

Add a go link for the mcp gateway (https://docs.docker.com/ai/mcp-catalog-and-toolkit/mcp-gateway/) for use in our frontend products

### DIFF
--- a/data/redirects.yml
+++ b/data/redirects.yml
@@ -295,7 +295,7 @@
 "/ai/mcp-catalog-and-toolkit/catalog/":
   - /go/mcp-catalog-and-toolkit-catalog/
 "/ai/mcp-catalog-and-toolkit/mcp-gateway/":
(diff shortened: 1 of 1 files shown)


data/redirects.yml
```diff
@@ -294,6 +294,8 @@
   - /go/mcp-catalog-and-toolkit/
 "/ai/mcp-catalog-and-toolkit/catalog/":
   - /go/mcp-catalog-and-toolkit-catalog/
+"/ai/mcp-catalog-and-toolkit/mcp-gateway/":
+  - /go/mcp-catalog-and-toolkit-gateway/
```
dvdksn: Why not something short and sweet? :-)
```suggestion
  - /go/mcp-gateway/
```

Daniel-Kolev: Was following the pattern we had on the above line 😢

Issue #23407, opened 2025-09-12
Spring Boot 3.4.0-M3 is not available, 3.4.9 has been selected.


TsitouhRanjafy:
### Is this a docs issue?

- [x] My issue is about the documentation content or website

### Type of issue

Information is incorrect

### Description

At the redirected link to generate ZIP file in [**Multi-stage builds**](https://docs.docker.com/get-started/docker-concepts/building-images/multi-stage-builds/) docs, the `spring boot 3.4.0-M3` is no longer available. 

current screenshot in docs:

<img width="579" height="344" alt="Image" src="https://github.com/user-attachments/assets/78153edc-12f9-470a-b41c-4a0aef22aea9" />

now in redirected link:

<img width="579" height="370" alt="Image" src="https://github.com/user-attachments/assets/6dcccf22-a615-48e2-9836-6b57f5990173" />

### Location

https://docs.docker.com/get-started/docker-concepts/building-images/multi-stage-builds/

craig-osterhout:
Fixed in https://github.com/docker/docs/pull/23999

PR #23654, opened 2025-11-04
fix(docs): update Spring Boot version and add note about version auto…

2003Aditya:
…-selection (#23407)

## Reviews

- [ ] Technical review
- [ ] Editorial review
- [ ] Product review


craig-osterhout:
Fixed in #23999

PR #23668, opened 2025-11-06
Update control startup example

vpavic:
## Description

This PR updates control startup example to prevent Compose from interpolating values intended as shell command arguments.

Current example yields the following warnings:

```bash
WARN[0000] The "POSTGRES_USER" variable is not set. Defaulting to a blank string. 
WARN[0000] The "POSTGRES_DB" variable is not set. Defaulting to a blank string. 
```

## Reviews

- [ ] Technical review
- [ ] Editorial review
- [ ] Product review

### DIFF
--- a/content/manuals/compose/how-tos/startup-order.md
+++ b/content/manuals/compose/how-tos/startup-order.md
@@ -42,7 +42,7 @@ services:
   db:
     image: postgres
     healthcheck:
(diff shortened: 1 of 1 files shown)


aevesdocker:
Would you mind rebasing @vpavic and then I can get it merged. Thanks

vpavic:
That's now taken care of @aevesdocker.

PR #23679, opened 2025-11-08
engine: install: debian.md suggest using apt UI

eshattow:
Suggest using apt for user-interactive commands:

- Replace apt-get invocation
- Replace apt-cache madison invocation
- Filter unofficial package names for removal to installed packages

### DIFF
--- a/content/manuals/engine/install/debian.md
+++ b/content/manuals/engine/install/debian.md
@@ -71,10 +71,10 @@ conflicts with the versions bundled with Docker Engine.
 Run the following command to uninstall all conflicting packages:
 
 ```console
-$ for pkg in docker.io docker-doc docker-compose podman-docker containerd runc; do sudo apt-get remove $pkg; done
+$ sudo apt remove $(dpkg --get-selections docker.io docker-compose docker-doc podman-docker containerd runc | cut -f1)
 ```
 
-`apt-get` might report that you have none of these packages installed.
+`apt` might report that you have none of these packages installed.
 
 Images, containers, volumes, and networks stored in `/var/lib/docker/` aren't
 automatically removed when you uninstall Docker. If you want to start with a
@@ -109,8 +109,8 @@ Docker from the repository.
 
    ```bash
    # Add Docker's official GPG key:
-   sudo apt-get update
-   sudo apt-get install ca-certificates curl
+   sudo apt update
+   sudo apt install ca-certificates curl
    sudo install -m 0755 -d /etc/apt/keyrings
    sudo curl -fsSL {{% param "download-url-base" %}}/gpg -o /etc/apt/keyrings/docker.asc
    sudo chmod a+r /etc/apt/keyrings/docker.asc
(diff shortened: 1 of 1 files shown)


dvdksn (commented):
We've used `apt-get` until now because it has a stable ui and is more suitable for scripting... but then otoh we do encourage entering these commands discretely. And if you want a scripted installation there's a great source of inspiration at `https://get.docker.com`.

So I guess changing to apt should be fine. Any concerns @thaJeztah?

And we'd need to update the ubuntu docs with the same.

thaJeztah:
Yeah, `apt-get` is, to my knowledge, still the preferred one when scripting is involved. While these steps are not necessarily part of a script, part of the expectation is that these will also be used to be used in scripting (or at most copy/paste'd by users). Examples we use, e.g. in Dockerfile examples, are all using `apt-get` for that reason

Looking at the changes, I don't think switching to `apt` _significantly_ makes the steps easier, so not sure we should switch these over.


cc @vvoland @tianon thoughts?

content/manuals/engine/install/debian.md
```diff
@@ -71,10 +71,10 @@ conflicts with the versions bundled with Docker Engine.
 Run the following command to uninstall all conflicting packages:
 
 ```console
+$ sudo apt remove $(dpkg --get-selections docker.io docker-compose docker-doc podman-docker containerd runc | cut -f1)
```
vvoland, line 74:
```suggestion
$ sudo apt remove $(dpkg --get-selections docker.io docker-compose docker-doc podman-docker containerd runc 2>/dev/null | cut -f1)
```

To avoid warnings for non-existing packages.

vvoland (approved):
I like it!

apt `update`, `install` are unlikely to change, and we're not piping/parsing their output anyway so I don't think `apt-get` has any benefits.

content/manuals/engine/install/debian.md
eshattow: No, thanks. The intent is to avoid scripting the sudo invocation of apt as the inner of a for-loop. Errors and information are intentionally preserved compared to the for-loop it replaces.

tianon, line 74: Why even filter? Doesn't `apt` have reasonable behavior when you ask it to remove packages that aren't installed?  (I'm sure we've probably discussed this before and I've just forgotten 😅)

tianon (approved):
Overall I'm +1 on switching these; I do think these docs are written expecting users to copy paste into interactive terminals, and users looking for the scripting version will need to understand and adapt them appropriately (as they will for all other software).

content/manuals/engine/install/debian.md
eshattow: all of the apt, apt-get, ... remove package tool actions bail out if a package does not exist. I would be in favor of just making the documentation more instructive "if apt complains that a package does not exist you should remove it from the command and try again". However, as this pull request goes, I remove the scripting of apt-get as it is replaced with apt, the intent is to keep the same errors and information, and at most the user will benefit slightly if they allow sudo to timeout due to slow network access or whatever situation they will just have the one sudo invocation with the update I propose.

PR #23691, opened 2025-11-10
Ruby language DHI

igor-alexandrov:
## Description

Upgraded Ruby language guide to include DHI-based Dockerfile example.

## Reviews

- [ ] Technical review
- [x] Editorial review
- [ ] Product review

### DIFF
--- a/content/guides/ruby/containerize.md
+++ b/content/guides/ruby/containerize.md
@@ -51,7 +51,7 @@ Although the Dockerfile is generated automatically, understanding its purpose an
 
 # Make sure RUBY_VERSION matches the Ruby version in .ruby-version
 ARG RUBY_VERSION=3.4.7
-FROM <your-namespace>/dhi-ruby:$RUBY_VERSION AS base
+FROM <your-namespace>/dhi-ruby:$RUBY_VERSION-dev AS base
 
 # Rails app lives here
 WORKDIR /rails


content/guides/ruby/containerize.md
```diff
@@ -44,7 +50,104 @@ Although the Dockerfile is generated automatically, understanding its purpose an
 # For a containerized dev environment, see Dev Containers: https://guides.rubyonrails.org/getting_started_with_devcontainer.html
 
 # Make sure RUBY_VERSION matches the Ruby version in .ruby-version
-ARG RUBY_VERSION=3.3.6
+ARG RUBY_VERSION=3.4.7
+FROM <your-namespace>/dhi-ruby:$RUBY_VERSION AS base
```
craig-osterhout: Thanks @igor-alexandrov. I think this needs to use the `dev` variant at a minimum because the runtime variant doesn't have a package manager. It could also use the runtime variant in the final stage to make the image even more secure, but not necessary.
```suggestion
FROM <your-namespace>/dhi-ruby:$RUBY_VERSION-dev AS base
```

igor-alexandrov: Thanks, Craig, you are right. Locally I tested this with the following setup:

``` dockerfile
ARG RUBY_VERSION=3.4.7-dev
FROM aleksandrovigor/dhi-ruby:$RUBY_VERSION AS base
```

And forgot to change the guide.

PR #23744, opened 2025-11-19
components: add deprecated badge

craig-osterhout:
## Description

Added gray color for deprecated badge.
Updated contribution guide for usage

## Reviews

- [ ] Editorial review


layouts/partials/components/badge.html
```diff
@@ -1,18 +1,20 @@
+layouts/partials/components/badge.html
```
dvdksn: Looks like this needs to be removed

```suggestion
```

PR #23749, opened 2025-11-20
Add DHI to the Python guide

mfranzon:
## Description

Upgraded Python language guide to include DHI-based Dockerfile example.

## Reviews

- [ ] Technical review
- [ ] Editorial review
- [ ] Product review


craig-osterhout (approved):
Thanks @mfranzon. LGTM
Noting that the develop your app section also has a separate Dockerfile that wasn't updated with DHI, but it's not necessary to follow the guide. I think this is good enough to gather some feedback and see if it's worth investing more time in DHI content.

PR #23759, opened 2025-11-22
Update Docker cache restoration step in CI config

KhalidAlansary:
Updated Docker cache mount restoration step to use the latest version of the buildkit-cache-dance action and added additional parameters for cache management.

### DIFF
--- a/content/manuals/build/ci/github-actions/cache.md
+++ b/content/manuals/build/ci/github-actions/cache.md
@@ -222,6 +222,9 @@ jobs:
   build:
     runs-on: ubuntu-latest
     steps:
+      - name: Checkout code
+        uses: actions/checkout@v5
+
       - name: Login to Docker Hub
         uses: docker/login-action@{{% param "login_action_version" %}}
         with:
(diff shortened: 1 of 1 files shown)


content/manuals/build/ci/github-actions/cache.md
```diff
@@ -219,6 +219,9 @@ jobs:
   build:
     runs-on: ubuntu-latest
     steps:
+      - name: Checkout code
+        uses: actions/checkout@v5
+
```
dvdksn, line 227: Why is checkout needed here? By default build-push action should build with git context.

KhalidAlansary: It's for the build-cache-dance action. It needs to read the dockerfile to parse it and extract the paths of the cache mounts used in the dockerfile. Without checkout, I get this error:
```
No cache map provided. Trying to parse the Dockerfile to find the cache mount instructions...
Error: Failed to parse cache map. Expected JSON, got:
{}
Error: ENOENT: no such file or directory, open 'build/package/Dockerfile'
```

PR #23764, opened 2025-11-24
hack: Add script to sync Moby release notes from docs

vvoland:
This script helps maintain consistency between documentation and GitHub release notes by making the docs the source of truth.

Previously, any djustments to release notes during review, needed to be manually propagated back to the corresponding Moby GitHub releases, creating a maintenance burden and potential for inconsistency.

The script extracts release notes for a given version from the engine documentation markdown files, compares them with the existing GitHub release notes, and provides a command to update the GitHub release if differences are found.

## Reviews

- [ ] Technical review
- [ ] Editorial review
- [ ] Product review

### DIFF
--- a/hack/moby-sync-release.sh
+++ b/hack/moby-sync-release.sh
@@ -1,0 +1,44 @@
+#!/bin/bash
+# Update the Github release notes to match the release notes from the docs.
+set -euo pipefail
+
+version="${1:-}"
+
+if [ -z "$version" ]; then
+    echo "Usage: $0 <version>"
+    exit 1
+fi
+
+moby="moby/moby"
+version="${version#v}" # remove the v prefix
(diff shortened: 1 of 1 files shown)


hack/moby-sync-release.sh
```diff
@@ -0,0 +1,44 @@
+#!/bin/bash
+# Update the Github release notes to match the release notes from the docs.
+set -euo pipefail
+
+version="$1"
+
+if [ -z "$version" ]; then
+    echo "Usage: $0 <version>"
+    exit 1
+fi
```
dvdksn: if $1 is unset, the script errors on variable assignment because `set -u`, so to show the usage string we could do:

```suggestion
version="${1:-}"

if [ -z "$version" ]; then
    echo "Usage: $0 <version>"
    exit 1
fi
```

PR #23803, opened 2025-12-03
compose sdk: release follow-up

aevesdocker:
## Description

Follow-up to https://github.com/docker/docs/pull/23802

- moves major new release to new release notes page
- adds the Compose SDK page 

## Reviews

- [ ] Technical review
- [ ] Editorial review
- [ ] Product review

### DIFF
--- a/content/manuals/compose/releases/release-notes.md
+++ b/content/manuals/compose/releases/release-notes.md
@@ -21,7 +21,7 @@ For more detailed information, see the [release notes in the Compose repo](https
 
 Major changes in this release:
 
-- Compose v5 introduces a new official Go SDK. This SDK provides a comprehensive API that lets you integrate Compose functionality directly into your applications, allowing you to load, validate, and manage multi-container environments without relying on the Compose CLI. 
+- Compose v5 introduces a new official Go SDK. This SDK provides a comprehensive API that lets you integrate Compose functionality directly into your applications, allowing you to load, validate, and manage multi-container environments without relying on the Compose CLI. For more information, see the [Compose SDK docs](/manuals/compose/compose-sdk.md).
 
 > [!NOTE]
 >


ndeloof (approved):

content/manuals/compose/compose-sdk.md
```diff
@@ -0,0 +1,170 @@
...
+        Create: api.CreateOptions{},
+        Start:  api.StartOptions{},
+    })
+    if err != nil {
+        log.Fatalf("Failed to start services: %v", err)
+    }
+
+    log.Printf("Successfully started project: %s", project.Name)
+}
+```
```
dvdksn, line 100: This does not compile for me

content/manuals/compose/compose-sdk.md
dvdksn, line 1: request: some more usage examples. Maybe take inspiration from the engine sdk examples page: https://deploy-preview-23751--docsdocker.netlify.app/reference/api/engine/sdk/examples/

Showing a few common use cases is probably a good idea to help people get going.

content/manuals/compose/compose-sdk.md
```diff
@@ -0,0 +1,170 @@
...
+- `WithEventProcessor(progress.EventProcessor)`: Receive progress events and operation notifications
+
+These options provide fine-grained control over the SDK's behavior, making it suitable for various integration
+scenarios including CLI tools, web services, automation scripts, and testing environments.
+
+## Tracking operations with `EventProcessor`
+
+The `EventProcessor` interface allows you to monitor Compose operations in real-time by receiving events about changes
+applied to Docker resources such as images, containers, volumes, and networks. This is particularly useful for building
+user interfaces, logging systems, or monitoring tools that need to track the progress of Compose operations.
```
dvdksn, line 144: Repeating myself here but an example here would greatly help. It's quite opaque at the moment, hard to understand how this is meant to be used.

content/manuals/compose/releases/release-notes.md
dvdksn, line 1: I know there's a blog post coming to explain the version jump from v2 -> v5, but probably wouldn't hurt to have something here too.

content/manuals/compose/releases/prior-releases.md
dvdksn, line 1: I would recommend `git mv` to ensure history is kept for this file. The history for the current file goes back to 2017.

content/manuals/compose/compose-sdk.md
ndeloof: indeed, see https://github.com/docker/compose/pull/13416

content/manuals/compose/compose-sdk.md
ndeloof: AFAIT the SDK functions are self-explanatory as there's a 1:1 match with compose commands

content/manuals/compose/compose-sdk.md
dvdksn: I tried that change too locally but it still didn't work for me

```
$ go version
go version go1.25.4 darwin/arm64
$ go build main.go 
# github.com/docker/buildx/util/dockerutil
../../go/pkg/mod/github.com/docker/buildx@v0.30.1/util/dockerutil/api.go:41:34: cannot use dockerclient.WithTraceOptions(otelhttp.WithPropagators(propagation.NewCompositeTextMapPropagator(propagation.TraceContext{}, propagation.Baggage{}))) (value of func type "github.com/docker/docker/client".Opt) as "github.com/moby/moby/client".Opt value in argument to append
../../go/pkg/mod/github.com/docker/buildx@v0.30.1/util/dockerutil/api.go:45:53: cannot use clientOpts (variable of type []"github.com/moby/moby/client".Opt) as []"github.com/docker/docker/client".Opt value in argument to dockerclient.NewClientWithOpts
# github.com/docker/compose/v5/cmd/formatter
../../go/pkg/mod/github.com/docker/compose/v5@v5.0.0/cmd/formatter/container.go:225:36: cannot use ports (variable of type []"github.com/docker/docker/api/types/container".Port) as []"github.com/moby/moby/api/types/container".PortSummary value in argument to formatter.DisplayablePorts
```

ndeloof: You need to fix docker/cli dependency in go.mod to match docker/compose:
```
  github.com/docker/cli v28.5.2+incompatible
```
(docker/cli 29.x relies on new Moby module, which is not yet integrated into compose)

content/manuals/compose/compose-sdk.md
dvdksn: You could say that about the engine SDK too but I still think examples are useful

content/manuals/compose/compose-sdk.md
dvdksn: Yep that worked. That would be useful information to include.

content/manuals/compose/compose-sdk.md
ndeloof: wrong, in many places the docker CLI relies on multiple API calls to achieve features people know well.

I'm not against examples, but I don't want to just write useless stuff like 
```
// example: List Compose applications
applications, err := compose.List(ctx, ListOptions{})
```
I'd be happy to add more example once we get feedback on API usability from actual users

dvdksn: yeah so when I say example, I don't mean showing function signatures. But we should be able to show some use cases beyond hello world. For example, using the `EventProcessor`. Or something with watch events, or custom logging, or even something more advanced like dependency visualization. Something to a) demo interesting capabilities, and b) show how these are meant to be used!

content/manuals/compose/releases/prior-releases.md
aevesdocker: So I actually just moved this content onto this new page and kept the old page (I didn't rename the file or anything), so is `git mv` still needed? isn't the history for `release-notes.md` still kept in this instance?

content/manuals/compose/compose-sdk.md
aevesdocker: Roger, and I agree. Can I request that we get this PR merged now and I add and improve the docs in a separate PR? It's going to take me a bit more time to get that together and it'd be good to get what we have now live. I've addressed your other comments already

content/manuals/compose/releases/prior-releases.md
aevesdocker: Hey @dvdksn thanks for approving, just want to check the thinking on this before I merge

dvdksn: hopefully we can remove the release notes from the website and just have them on github anyway, in which case this doesn't matter 😄

PR #23809, opened 2025-12-04
Fix: missing stage name in php-fpm Dockerfile

Matra-Master:
I simply added the stage name that the compose.prod.yml references for service php-fpm since it was not building when I tried using those fiels.

## Description

I added the stage name to php-fpm Dockerfile as it was referenced in compose.prod.yml.
When I tried to build php-fpm service with that compose file it failed because it had a target referencing a non-existent stage.


Matra-Master:
My pleasure 🫡

PR #23816, opened 2025-12-05
Document `extends` lack of support in Docker Stack

oscherler:
## Description

Mention that the `extends` property of services in compose files is not supported in `docker stack`, as it’s documented nowhere.

## Reviews

I’m open to suggestions for better wording, but I’d rather that this PR is rejected because it’s been made obsolete (by just implementing the missing feature.)

- [ ] Technical review
- [ ] Editorial review
- [ ] Product review

### DIFF
--- a/content/manuals/compose/how-tos/multiple-compose-files/extends.md
+++ b/content/manuals/compose/how-tos/multiple-compose-files/extends.md
@@ -21,6 +21,10 @@ application, with the ability to override some attributes for your own needs.
 
 > [!IMPORTANT]
 >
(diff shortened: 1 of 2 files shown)


oscherler:
Well, ideally, since you work at Docker, you’d have changed the error message in `docker stack config` and `docker stack deploy` so that it means something, instead of just rewording the text of my PR.

You could also have asked me to make changes, instead of taking my work and closing my PR, ensuring I get no recognition for pointing out this flaw in your docs.

PR #23855, opened 2025-12-17
engine: 29.2.0

vvoland:


content/reference/api/engine/version/v1.52.md
```diff
@@ -2,6 +2,5 @@
 linkTitle: v1.52
 title: Docker Engine API v1.52 reference
 aliases:
-  - /engine/api/latest/
-  - /reference/api/engine/latest/
+  - /engine/api/v1.52/
```
dvdksn: We don't actually need to keep this redirect on this page because the route `/engine/api/v1.52/` never existed. We had it for some of the older api versions because we changed paths at some point. But this one was only ever available under `/reference/api/engine/version/v1.52/` so no aliases needed.

```suggestion
```

vvoland: Done

PR #23867, opened 2025-12-18
ci: Add automated CLI documentation sync workflow

vvoland:
This adds a GitHub Actions workflow and supporting script to automatically sync CLI documentation from the docker/cli repository on a daily schedule.

## Reviews

- [ ] Technical review
- [ ] Editorial review
- [ ] Product review

### DIFF
--- a/.github/workflows/sync-cli-docs.yml
+++ b/.github/workflows/sync-cli-docs.yml
@@ -1,0 +1,100 @@
+name: sync-cli-docs
+
+on:
+  schedule:
+    # Run daily at 02:00 UTC
+    - cron: '0 2 * * *'
+  workflow_dispatch:
+    inputs:
+      version:
+        description: "(optional) Docker CLI version - defaults to docker_ce_version in hugo.yaml"
+        required: false
+        default: ""
+  pull_request:
+
+permissions:
+  contents: write
+  pull-requests: write
+
+jobs:
+  sync-cli-docs:
+    runs-on: ubuntu-24.04
+    steps:
+      -
+        name: Checkout docs repo
+        uses: actions/checkout@v5
+        with:
+          fetch-depth: 0
+      -
+        name: Get version from hugo.yaml
+        id: get-version
+        run: |
+          if [ -n "${{ inputs.version }}" ]; then
+            VERSION="${{ inputs.version }}"
+          else
+            VERSION=$(grep "docker_ce_version:" hugo.yaml | awk '{print $2}' | tr -d '"')
+          fi
+          # TODO(vvoland): Remove this after 29.2.0 is released
+          # VERSION=v${VERSION}
+          VERSION=60f06cb2df3df36ddfb531c1dae8c6fa96e5f9e7
+
+          echo "version=$VERSION" >> "$GITHUB_OUTPUT"
+          echo "Docker CLI version: **$VERSION**" | tee -a "$GITHUB_STEP_SUMMARY"
+      -
+        name: Checkout docker/cli repo
+        uses: actions/checkout@v5
+        with:
+          repository: docker/cli
+          path: cli-source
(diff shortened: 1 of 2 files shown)


.github/workflows/sync-cli-docs.yml
```diff
@@ -0,0 +1,68 @@
...
+        id: create-branch
+        run: |
+          BRANCH_NAME="bot/sync-cli-docs-$(date +%Y%m%d-%H%M%S)"
+          git checkout -b "$BRANCH_NAME"
+          echo "branch_name=$BRANCH_NAME" >> "$GITHUB_OUTPUT"
+      -
+        name: Run sync script
+        id: sync
+        run: |
+          if ./hack/sync-cli-docs.sh cli-source origin/master; then
```
dvdksn: We do not vendor the CLI docs from master - we vendor a tag or release branch (there might be exceptions to this rule but in general). Maybe we could leverage the docker_ce_version config, and use the release branch?

vvoland:
The latest release needs to include https://github.com/docker/cli/pull/6716

.github/workflows/sync-cli-docs.yml
```diff
@@ -0,0 +1,89 @@
...
+        name: Get version from hugo.yaml
+        id: get-version
+        run: |
+          if [ -n "${{ inputs.version }}" ]; then
+            VERSION="${{ inputs.version }}"
+          else
+            VERSION=$(grep "docker_ce_version:" hugo.yaml | awk '{print $2}' | tr -d '"')
+          fi
+          # TODO(vvoland): Remove this after 29.2.0 is released
+          VERSION=60f06cb2df3df36ddfb531c1dae8c6fa96e5f9e7
```
vvoland, line 39: Gonna remove this after https://github.com/docker/cli/pull/6716 is included in the latest version

.github/workflows/sync-cli-docs.yml
thaJeztah: Ah, yes, I was wondering that as well; we should probably pick the version we vendor (which _should_ be a tagged version)?

vvoland: Changed to pick the one from hugo.yaml

vvoland:
Should be good now!

.github/workflows/sync-cli-docs.yml
thaJeztah: As a follow-up, we should look at having a Dockerfile target that we can build from; that way we don't have to clone the git repository, but instead can just do something like;

```bash
docker build --target=yamldocs --output=./<destination> https://github.com/docker/cli#<git ref>
```

.github/workflows/sync-cli-docs.yml
```diff
@@ -0,0 +1,100 @@
...
+        if: steps.sync.outputs.changes == 'true' && github.event_name != 'pull_request'
+        env:
+          GH_TOKEN: ${{ github.token }}
+          PR_BODY: |
+            ## Summary
+
+            Automated sync of CLI documentation from docker/cli repository.
+        run: |
+          git push -u origin "${{ steps.create-branch.outputs.branch_name }}"
+          gh pr create \
```
thaJeztah, line 96: Curious; where does it create the branch? Is that in a separate repository or directly here in the upstream?

vvoland: Directly in the upstream:
https://github.com/docker/docs/blob/c3652459f9236d79ab2e64df4c817e6dc3be6bac/.github/workflows/sync-cli-docs.yml#L55

Not ideal but... works

vvoland: And this repo already uses dependabot so 😅

dvdksn (approved):
LGTM! Thanks @vvoland

PR #23884, opened 2025-12-21
Update local-models.md

flopwiki:
Models available to the local Docker catalog are only option for list

## Description

Updated documentation to reflect true state of local docker catalog flags. The `--available` option does not exist and the 'list' or 'ls' command is only supported for local viewing of installed models. Documentation currently suggests a user can [1] use the --available flag, and [2] a user can use 'ls' to query an online catalog. 

## Reviews

- [ ] Technical review
- [x] Editorial review
- [ ] Product review


dvdksn (approved):
Thanks 🙏

PR #23925, opened 2026-01-05
gha/sync-cli-docs: Update existing PR instead of creating new ones

vvoland:
Actually we don't need to create a separate PR per each update and we can just modify the previous one if it wasn't merged yet.

Change the workflow to use a fixed branch name and update existing PRs when newer CLI versions are available, rather than creating multiple PRs.

The workflow now uses a consistent branch name "bot/sync-cli-docs" instead of timestamp-based names.

When an existing PR from this branch is found, it force-pushes the new changes and updates the PR title and body with the latest CLI version information.

## Reviews

- [ ] Technical review
- [ ] Editorial review
- [ ] Product review

### DIFF
--- a/.github/workflows/sync-cli-docs.yml
+++ b/.github/workflows/sync-cli-docs.yml
@@ -19,6 +19,9 @@ permissions:
   contents: write
   pull-requests: write
 
+env:
+  BRANCH_NAME: "bot/sync-cli-docs"
+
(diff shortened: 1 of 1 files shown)


dvdksn (approved):

vvoland:
The branch name is now a const so I just extracted it to a global var

PR #23944, opened 2026-01-08
api: include alternate link for swagger / OpenAPI spec

thaJeztah:
The API docs are generated dynamically, which prevents them from being crawled / indexed. Provide a link to the swagger / OpenAPI spec that's used to generate the page, so that that content can be used as alternative.

## Reviews

- [ ] Technical review
- [ ] Editorial review
- [ ] Product review

### DIFF
--- a/layouts/_default/api.html
+++ b/layouts/_default/api.html
@@ -1,7 +1,7 @@
 <!DOCTYPE html>
 <html lang="en">
-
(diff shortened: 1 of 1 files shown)


thaJeztah:
Thx! Let me make CoPilot happier if it also works inside the head tag 😅

PR #23977, opened 2026-01-14
Fixed doc links regarding options (host_ip, gateway_mode) in bridge driver

Teeed:
## Description

This PR fixes docs links for `com.docker.network.host_ipv4` and `com.docker.network.bridge.gateway_mode_ipv4` options (and IPv6 ones). "See more" points to `../port-publishing.md` doc, which seems like much more appropriate page for that, since it documents values for these options.

## Reviews

- [ ] Technical review
- [ ] Editorial review
- [ ] Product review


dvdksn (approved):
Hey! It seems like this one got lost but the update LGTM so let's get it in! Thanks @Teeed

Teeed:
Better late than never ;)

PR #23978, opened 2026-01-14
homepage: update download component

craig-osterhout:
## Description

This might be a my computer thing (because my computer probably isn't the fastest), but the component often flickers on load.

- ~Removed x-collapse and~ added x-cloak to prevent the drop-down from appearing expanded before the js loads.
- Changed the button text to static, so it's not waiting for the OS detection to load.

## Related issues or tickets

ENGDOCS-3149

## Reviews

- [ ] Editorial review

### DIFF
--- a/layouts/index.html
+++ b/layouts/index.html
@@ -89,7 +89,7 @@
 									</svg>
 								</button>
 							</div>
-							<div x-show="open" x-cloak
+							<div x-show="open" x-collapse x-cloak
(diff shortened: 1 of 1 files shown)


layouts/index.html
```diff
@@ -93,7 +89,7 @@ <h1 class="text-xl">Get Docker</h1>
 									</svg>
 								</button>
 							</div>
-							<div x-show="open" x-collapse
+							<div x-show="open" x-cloak
```
dvdksn: You can keep x-collapse along with x-cloak - collapse makes it a smooth animation.

PR #23987, opened 2026-01-15
Update Develop with containers (and) CONTRIBUTING.MD

akristen:
This is a two part PR for some simple edits. I found them while working through docs site ~foundations, and I have a task to file my first PR anyway. 😸 

Includes: 

* Correction to step 3 in `### Change the greeting` section. Live step omits that you need to re-run `docker compose watch` in addition to saving the file after you've updated the variable. I saw that this was also flagged in our feedback channel, too.  
* I updated the indent in CONTRIBUTING.MD to match the other snippets. Right now that particular snippet also copies over the `$`, which the others do not do. The only difference I saw that might create that buggy behavior is the indent. 🖖

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -100,9 +100,9 @@ Before you push your changes and open a pull request, we recommend that you
 test your site locally first. Local tests check for broken links, incorrectly
 formatted markup, and other things. To run the tests:
 
-```console
-$ docker buildx bake validate
-```
+    ```console
+    $ docker buildx bake validate
+    ```
 
 If this command doesn't result in any errors, you're good to go!
 


content/get-started/introduction/develop-with-containers.md
```diff
@@ -97,7 +97,13 @@ The greeting at the top of the page is populated by an API call at `/api/greetin
     };
     ```
 
-3. If you haven't done so yet, save the file. If you refresh your browser, you should see a new greeting. If you keep refreshing, you should see all of the messages appear.
+3. Save the file if you haven't done so yet, then run the following command again:
```
aevesdocker, line 100: With `docker compose watch` the idea is that you don't have to re-enter the command as it's purpose is to automatically update and preview your running Compose services as you edit and save your code. So this shouldn't be necessary. 

SO! It looks like you've found a bug. I'll take this to the Compose team to figure out what's going on.

CONTRIBUTING.md
```diff
@@ -100,9 +100,9 @@ Before you push your changes and open a pull request, we recommend that you
 test your site locally first. Local tests check for broken links, incorrectly
 formatted markup, and other things. To run the tests:
 
-```console
-$ docker buildx bake validate
-```
+    ```console
+    $ docker buildx bake validate
```
aevesdocker, line 104: Looks like it still copies the '$', but it does that for all the commands on that page 🤔 
Probs a github issue rather than us

PR #24003, opened 2026-01-21
Update Dockerfile with secure base image and non-root user

ajeetraina:
- Use node:24-alpine base image
- Run as non-root user (node) for security
- Create /etc/todos directory with proper permissions

## Reviews

- [ ] Technical review
- [ ] Editorial review
- [ ] Product review

### DIFF
--- a/content/get-started/workshop/02_our_app.md
+++ b/content/get-started/workshop/02_our_app.md
@@ -69,10 +69,17 @@ To build the image, you'll need to use a Dockerfile. A Dockerfile is simply a te
    CMD ["node", "src/index.js"]
    ```
 
-   This Dockerfile starts off with a `node:24-alpine` base image, a
-   light-weight Linux image that comes with Node.js and the npm package
-   manager pre-installed. It copies all of the source code into the image,
-   installs the necessary dependencies, and starts the application.
+   This Dockerfile:
+
+   - Starts with a `node:24-alpine` base image, a lightweight Linux image that comes with Node.js and npm pre-installed
+   - Sets `/app` as the working directory for subsequent instructions
+   - Copies `package.json` and `package-lock.json` first to leverage Docker's layer caching for dependencies
+   - Installs production dependencies only (`--omit=dev`)
+   - Removes npm after installation since it's no longer needed at runtime
+   - Creates the `/etc/todos` directory for the SQLite database with proper ownership
+   - Copies the application source code
+   - Switches to the non-root node user for security
+   - Exposes port `3000` and starts the application
 
 2. Build the image using the following commands:
 


content/get-started/workshop/02_our_app.md
```diff
@@ -57,17 +57,19 @@ To build the image, you'll need to use a Dockerfile. A Dockerfile is simply a te
 
    ```dockerfile
    # syntax=docker/dockerfile:1
-
-   FROM node:lts-alpine
+   FROM node:24-alpine
    WORKDIR /app
+   COPY package.json yarn.lock ./
+   RUN yarn install --production && \
```
craig-osterhout: These dockerfile updates look different than the one mentioned at https://github.com/docker/docs/issues/23988#issuecomment-3779465226

The updates in https://github.com/docker/getting-started-app/pull/98 were based on that  one not using yarn. I can update  the app PR based on this dockerfile, I just want to make sure this is the correct one.

ajeetraina: Yes, I changed it to YARN so as to match the instruction of the workshop. I have tested both the NPM and YARN scenario and it works great.

MikeMcC399: The version of Yarn you are using here is unsupported. It would be better to change to npm.

MikeMcC399: If you are basing the Dockerfile on using https://github.com/docker/getting-started-app/pull/98, then you need to use npm, not Yarn.

content/get-started/workshop/02_our_app.md
```diff
@@ -57,17 +57,19 @@ To build the image, you'll need to use a Dockerfile. A Dockerfile is simply a te
...
-   CMD ["node", "src/index.js"]
+   USER node
    EXPOSE 3000
+   CMD ["node", "src/index.js"]
    ```
 
-   This Dockerfile starts off with a `node:lts-alpine` base image, a
-   light-weight Linux image that comes with Node.js and the Yarn package
+   This Dockerfile starts off with a `node:24-alpine` base image, a
+   light-weight Linux image that comes with Node.js and the YARN package
```
MikeMcC399: If you change to npm (bundled with Node.js) you don't need to mention Yarn. In any case "Yarn" is written mixed case, not upper case. See the documentation for the now unsupported Yarn v1 Classic version https://classic.yarnpkg.com/en/docs.

MikeMcC399:
The Dockerfile is now more complex that the original one. The extra steps:

- adding `node` user
- removing the npm package manager

either need to be explained, or they should be omitted and left for a later stage where optimizations, such as security / reducing image size, etc. can be described.

ajeetraina:
@MikeMcC399  If you notice Parts 5 (Bind Mounts), 7 (Docker Compose), and 8 (Image-building best practices) all still use `yarn install && yarn run dev` commands throughout. We might need to work on revamping the entire workshop to migrate consistently from Yarn to npm across all parts. This will ensure a cohesive learning experience while addressing the Yarn v1 deprecation properly. Will update this PR once the comprehensive changes are ready.

Regarding the USER entry in the Dockerfile - it's required. Otherwise, users will encounter permission issues. I can add 1-2 lines explaining what it is and why it is required.

ajeetraina:
@craig-osterhout @MikeMcC399 I will be raising a separate PR that replaces all occurances of Yarn with NPM in order to maintain consistency.

MikeMcC399:
@ajeetraina 

> I will be raising a separate PR that replaces all occurances of Yarn with NPM in order to maintain consistency.

Why not add the changes into this PR? They belong together.

Also, if you are writing "npm", you should use lower case. See https://github.com/npm/cli#faq-on-branding.

ajeetraina:
Closing this as I raised a new PR  https://github.com/docker/docs/pull/24012 that replaces Yarn with npm at all the places.

PR #24017, opened 2026-01-26
feat: add support for Diffusers inference engine and image generation API

ilopezluna:
## Description

Add documentation for the new Diffusers inference engine introduced in Docker Model Runner `v1.0.13`. This enables users to generate images from text prompts using Stable Diffusion models.

## Reviews

- [x] Technical review
- [ ] Editorial review
- [ ] Product review


ericcurtin:
Apparently diffusers works on macOS:

https://github.com/huggingface/swift-coreml-diffusers
https://huggingface.co/blog/fast-mac-diffusers

But apparently mlx can do this too, so it's worth considering vllm-metal also

PR #24023, opened 2026-01-27
update go to 1.25

crazy-max:
## Description

Update Go to 1.25

Needed to update buildx module to latest stable 0.31.0: https://github.com/docker/buildx/actions/runs/21230817252/job/61088675155#step:7:168

## Reviews

- [ ] Technical review
- [ ] Editorial review
- [ ] Product review


hugo.yaml
```diff
@@ -136,7 +136,7 @@ params:
   buildkit_version: "0.27.0"
 
   # Example runtime/library/os versions
-  example_go_version: "1.24"
+  example_go_version: "1.25"
   example_alpine_version: "3.21"
   example_node_version: "20"
```
thaJeztah, line 141: Unrelated, but we should probably update the example Alpine version as well (I think 3.22 is current stable, and 3.23 is there as well).

Not sure what current Node version is though.

crazy-max: For Node I think 24 is fine (LTS) and Alpine 3.23 as well.

PR #24025, opened 2026-01-27
go.mod: use `tool` directive and no `replace`

thaJeztah:
## Reviews

- [ ] Technical review
- [ ] Editorial review
- [ ] Product review


thaJeztah:
Not sure if it's expected for `go mod tidy` to work; probably these are because the top-level doesn't have a `.go` file :thinking:

```
go: finding module for package github.com/docker/buildx
go: finding module for package github.com/docker/compose/v5
go: finding module for package github.com/docker/scout-cli
go: github.com/docker/mcp-gateway: module github.com/docker/mcp-gateway@latest found (v0.37.0), but does not contain package github.com/docker/mcp-gateway
go: github.com/docker/buildx: module github.com/docker/buildx@latest found (v0.31.0), but does not contain package github.com/docker/buildx
go: github.com/docker/cli: module github.com/docker/cli@latest found (v29.2.0+incompatible), but does not contain package github.com/docker/cli
go: github.com/docker/compose/v5: module github.com/docker/compose/v5@latest found (v5.0.2), but does not contain package github.com/docker/compose/v5
go: github.com/docker/scout-cli: module github.com/docker/scout-cli@latest found (v1.19.0), but does not contain package github.com/docker/scout-cli
go: github.com/docker/docker: module github.com/docker/docker@latest found (v28.5.2+incompatible), but does not contain package github.com/docker/docker
```

FWIW; trying to run `go mod tidy` on main will remove all dependencies (as there's no `tools.go` or otherwise), so I don't think that has to be an issue?

crazy-max:
I think it needs changes in module update because there is no more replace directive: https://github.com/docker/docs/blob/92082c472fa5b1e506c202a6782a0967af584418/Dockerfile#L97

thaJeztah:
Good call; looks like we can drop that altogether; 

This seems to work;

```bash
VENDOR_MODULE=github.com/docker/cli docker bake vendor
git diff go.mod
diff --git a/go.mod b/go.mod
index b2f1821168..ffbb0681a7 100644
--- a/go.mod
+++ b/go.mod
@@ -16,7 +16,7 @@ tool (

 require (
        github.com/docker/buildx v0.31.0
-       github.com/docker/cli v29.1.5+incompatible
+       github.com/docker/cli v29.2.0+incompatible
        github.com/docker/compose/v5 v5.0.1
        github.com/docker/docker v28.5.2+incompatible
```

Also works with a version;

```bash
VENDOR_MODULE=github.com/docker/cli@v29.2.0-rc.1+incompatible docker bake vendor

git diff go.mod
diff --git a/go.mod b/go.mod
index b2f1821168..d182fc05c3 100644
--- a/go.mod
+++ b/go.mod
@@ -16,7 +16,7 @@ tool (

 require (
        github.com/docker/buildx v0.31.0
-       github.com/docker/cli v29.1.5+incompatible
+       github.com/docker/cli v29.2.0-rc.1+incompatible
        github.com/docker/compose/v5 v5.0.1
        github.com/docker/docker v28.5.2+incompatible
```

go.mod
```diff
@@ -2,140 +2,25 @@ module github.com/docker/docs
...
-	github.com/docker/compose/v5 => github.com/docker/compose/v5 v5.0.1
-	github.com/docker/mcp-gateway => github.com/docker/mcp-gateway v0.22.0
-	github.com/docker/model-runner/cmd/cli => github.com/docker/model-runner/cmd/cli v1.0.3
-	github.com/moby/buildkit => github.com/moby/buildkit v0.27.0
-	github.com/moby/moby/api => github.com/moby/moby/api v1.53.0
+require (
+	github.com/docker/buildx v0.31.0
+	github.com/docker/cli v29.1.5+incompatible
+	github.com/docker/compose/v5 v5.0.1
+	github.com/docker/docker v28.5.2+incompatible
```
thaJeztah: 🤔 need to check where we use this one for (should be fine for a follow-up though)

dvdksn: I think it might be due to compose
https://github.com/docker/compose/blob/main/go.mod#L20

thaJeztah: oh, right; but if we don't have a direct import ... maybe we don't need to specify it here.

dvdksn: yeah I think you're right

crazy-max: iirc we needed this one for contrib docs on moby repo but maybe I'm wrong

dvdksn: we only pull the api docs from moby now, and that's a separate module; moby/moby/api

go.mod
```diff
@@ -2,140 +2,27 @@ module github.com/docker/docs
...
-	google.golang.org/grpc v1.77.0 // indirect
-	google.golang.org/protobuf v1.36.11 // indirect
-	gopkg.in/tomb.v1 v1.0.0-20141024135613-dd632973f1e7 // indirect
-	gopkg.in/yaml.v3 v3.0.1 // indirect
-	howett.net/plist v1.0.2-0.20250314012144-ee69052608d9 // indirect
+// This go.mod file is used by hugo to vendor documentation from upstream
+// reposities. The "tool" section lists the upstream repositories from which we
+// vendor documentation. Use the "require" section to specify the version of the
+// upstream repository.
+tool (
```
crazy-max, line 20: nit: `tool` block is usually put after `require` and `replace` ones

thaJeztah: Ah, yeah, I was on the fence what would be more clear; I can swap them

thaJeztah: Swapped them 👍

dvdksn:
Bringing this one in so I can rebase #23782

PR #24058, opened 2026-02-02
Add GitHub Copilot CLI to AI sandboxes list

brunoborges:
It works with Copilot CLI


dvdksn:
Ah thanks @brunoborges for noticing and opening a patch 🎉 I actually had a PR in flight, so I will close this one. #24059

PR #24075, opened 2026-02-04
added manual for creating own DHI image

ajeetraina:
## Description

This PR adds a missing manual of building your own custom DHI image based on Alpine and Debian.

## Reviews

- [ ] Technical review
- [ ] Editorial review
- [ ] Product review

### DIFF
--- a/content/manuals/dhi/how-to/build.md
+++ b/content/manuals/dhi/how-to/build.md
@@ -283,7 +283,7 @@ $ docker buildx build . -f base.yaml \
 > The `tags` field in the spec file defines the image metadata (variant and
 > version labels embedded in the image manifest). The `--tag` flag on the CLI
 > sets the OCI image reference used to push or load the image. These serve
-> different purposes — the spec file tags describe *what the image is*, while
+> different purposes - the spec file tags describe *what the image is*, while
 > the CLI tag determines *where it's stored*.
 
 ## Use a Debian base with third-party repositories


ajeetraina:
Currently, working on a scenario that captures "Using OCI artifacts as package sources".

content/manuals/dhi/how-to/build.md
```diff
@@ -0,0 +1,659 @@
+---
+title: Build a custom Docker Hardened Image
```
craig-osterhout: Small nit, be mindful of using `custom` (throughout the topic) to describe this because we use that verb for an enterprise feature / preferred way to customize.
Maybe instead, something like `Define and build an image` or `Create and build an image`,  `Build from definition`, etc.

ajeetraina: I agree with you. Just pushed the changes. Instead of saying "custom DHI image", I kept the title as "Create and build a Docker Hardened image".

ajeetraina:
Ran vale and fixed few formatting issue.

content/manuals/dhi/how-to/build.md
```diff
@@ -0,0 +1,834 @@
...
+
+```yaml
+# syntax=dhi.io/build:2-alpine3.23
+
+name: My Base Image
+image: my-registry/my-base
+variant: runtime
+tags:
+  - "1.0"
+  - "latest"
```
dvdksn: I guess since DHI is all about supply chain security and conduct, maybe we shouldn't include a `latest` tag in this example.

content/manuals/dhi/how-to/build.md
```diff
@@ -0,0 +1,834 @@
...
+- The `annotations` block adds OCI metadata visible in registries and Docker
+  Scout reports.
+
+Build the image:
+
+```console
+$ docker buildx build . -f base.yaml \
+    --sbom=generator=dhi.io/scout-sbom-indexer:1 \
+    --provenance=1 \
+    --tag my-base:latest \
```
dvdksn, line 278: Silly question because I am not familiar with this stuff: why do we specify image/tag in the spec file if we also specify it on the CLI?

Maybe worth explaining how image/variant/tag in the spec file works

content/manuals/dhi/how-to/build.md
```diff
@@ -0,0 +1,834 @@
...
+      includes:
+        - opt/**
+      uid: 0
+      gid: 0
+```
+
+| Field      | Description                                                                  |
+|------------|------------------------------------------------------------------------------|
+| `name`     | Full OCI reference with digest pin. Always use `@sha256:` for reproducibility. |
+| `includes` | Glob patterns for files to extract from the artifact. Use `opt/**` to include everything under a directory. |
```
dvdksn: I guess this is true:

```suggestion
| `includes` | Glob patterns for files to extract from the artifact. Paths are resolved from the filesystem root; `opt/**` includes everything under the `/opt` path. |
```

content/manuals/dhi/how-to/build.md
ajeetraina: You're right.  Let me fix this.

content/manuals/dhi/how-to/build.md
ajeetraina: Fair question. Added a note.

dvdksn (approved):
LGTM

content/manuals/dhi/how-to/build.md
```diff
@@ -0,0 +1,835 @@
...
+Build the image:
+
+```console
+$ docker buildx build . -f base.yaml \
+    --sbom=generator=dhi.io/scout-sbom-indexer:1 \
+    --provenance=1 \
+    --tag my-base:latest \
+    --load
+```
+> **Note:** The `tags` field in the spec file defines the image metadata (variant and version labels embedded in the image manifest). The `--tag` flag on the CLI sets the OCI image reference used to push or load the image. These serve different purposes — the spec file tags describe *what the image is*, while the CLI tag determines *where it's stored*.
```
usha-mandya: @ajeetraina Could you use the Note callout here?

ajeetraina: Fixed. Thanks.

PR #24122, opened 2026-02-10
fix: correct HEALTHCHECK interval documentation

### LINKED ISSUE #24096: Documentation for the --interval flag of the HEALTHCHECK instruction is incorrect
### Is this a docs issue?

### Type of issue

Information is incorrect

### Description

According to https://docs.docker.com/reference/dockerfile/#healthcheck : 

> The health check _will first run **interval** seconds after the container is started_, and then again **interval** seconds after each previous check completes.

(Italics mine)
This doesn't appear to be accurate - by my testing, the health check runs immediately* upon container startup (and _then_ runs again **interval** seconds after each previous check completes).

*Technically I think the health check may run --start-interval seconds after container startup (defaults to 5 seconds). So the corrected docs perhaps should read:

> The health check will first run **start-interval** seconds after the container is started, and then again **interval** seconds after each previous check completes.

But I haven't tested **--start-interval** extensively. (**--interval** definitely doesn't work exactly as documented, however.)

### Location

https://docs.docker.com/reference/dockerfile/#healthcheck

### Suggestion

_No response_

lacatoire:
## Description

The HEALTHCHECK documentation stated that the health check first runs after **interval** seconds. In practice, the initial checks use **start-interval** (default 5s) during the start period, and then **interval** seconds after each previous check completes once the container is considered started.

Fixes #24096

### DIFF
--- a/_vendor/github.com/moby/buildkit/frontend/dockerfile/docs/reference.md
+++ b/_vendor/github.com/moby/buildkit/frontend/dockerfile/docs/reference.md
@@ -2865,8 +2865,9 @@ The options that can appear before `CMD` are:
 - `--start-interval=DURATION` (default: `5s`)
 - `--retries=N` (default: `3`)
 
-The health check will first run **interval** seconds after the container is
-started, and then again **interval** seconds after each previous check completes.
+The health check will first run **start-interval** seconds after the container is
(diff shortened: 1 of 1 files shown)


lacatoire:

PR #24123, opened 2026-02-10
fix: correct CMD form recommendation with exec ENTRYPOINT

### LINKED ISSUE #23122: Exec form ENTRYPOINT example possible incorrection
### Is this a docs issue?

### Type of issue

Information is incorrect

### Description

Under "[Exec form ENTRYPOINT example](https://docs.docker.com/reference/dockerfile/#exec-form-entrypoint-example)" the docs says:

> You can use the exec form of ENTRYPOINT to set fairly stable default commands and arguments and then use **either form** of CMD to set additional defaults that are more likely to be changed.

but [previous](https://docs.docker.com/reference/dockerfile/#cmd) it stated:

> If CMD is used to provide default arguments for the ENTRYPOINT instruction, both the CMD and ENTRYPOINT instructions should be specified in the [exec form](https://docs.docker.com/reference/dockerfile/#exec-form).

Consequently I'd say that's not correct the sentence "then use either form of CMD to set additional defaults that are more likely to be changed".

### Location

https://docs.docker.com/reference/dockerfile/#exec-form-entrypoint-example

### Suggestion

I think the sentence should be rephrased in: You can use the exec form of ENTRYPOINT to set fairly stable default commands and arguments and then use **exec** form of CMD to set additional defaults that are more likely to be changed.

lacatoire:
## Description

The Dockerfile reference under "Exec form ENTRYPOINT example" stated that you can use "either form of CMD" with an exec form ENTRYPOINT. However, earlier in the same document it states that both CMD and ENTRYPOINT should use the exec form when CMD provides default arguments to ENTRYPOINT.

Changed "either form" to "the exec form" for consistency.

## Related issues or tickets

Fixes #23122

## Reviews

- [ ] Technical review
- [ ] Editorial review
- [ ] Product review

### DIFF
--- a/_vendor/github.com/moby/buildkit/frontend/dockerfile/docs/reference.md
+++ b/_vendor/github.com/moby/buildkit/frontend/dockerfile/docs/reference.md
@@ -2071,7 +2071,7 @@ Only the last `ENTRYPOINT` instruction in the Dockerfile will have an effect.
 ### Exec form ENTRYPOINT example
 
 You can use the exec form of `ENTRYPOINT` to set fairly stable default commands
-and arguments and then use either form of `CMD` to set additional defaults that
+and arguments and then use the exec form of `CMD` to set additional defaults that
 are more likely to be changed.
 
 ```dockerfile


lacatoire:

PR #24124, opened 2026-02-10
fix: clarify STOPSIGNAL applies to docker stop, not Ctrl+C

### LINKED ISSUE #23140: STOPSINGAL
### Is this a docs issue?

### Type of issue

Other

### Description

If I'm not mistaken, it should be better to inform that `STOPSIGNAL` is for `docker stop`.
If I run a container without `-d` flag, to run the container in foreground, by sending `CTRL + C`, the signal would be `SIGINT`.

In both cases, with or without `-d` flag, I'm exiting the container. So I think it could be a misunderstanding at first when reading the document.

### Location

https://docs.docker.com/reference/dockerfile/

### Suggestion

```
The STOPSIGNAL instruction sets the system call signal that will be sent to the container to exit with `docker container stop` or `docker stop`. This signal can be a signal name in the format SIG<NAME>, for instance SIGKILL, or an unsigned number that matches a position in the kernel's syscall table, for instance 9. The default is SIGTERM if not defined.

The image's default stopsignal can be overridden per container, using the --stop-signal flag on docker run and docker create
```

lacatoire:
## Description

Clarified that the `STOPSIGNAL` instruction applies specifically when stopping a container via `docker stop` or `docker container stop`. Added a note explaining that `Ctrl+C` on a foreground container sends `SIGINT`, not the signal specified by `STOPSIGNAL`.

## Related issues or tickets

Fixes #23140

## Reviews

- [ ] Technical review
- [ ] Editorial review
- [ ] Product review

### DIFF
--- a/_vendor/github.com/moby/buildkit/frontend/dockerfile/docs/reference.md
+++ b/_vendor/github.com/moby/buildkit/frontend/dockerfile/docs/reference.md
@@ -2832,10 +2832,15 @@ STOPSIGNAL signal
 ```
 
 The `STOPSIGNAL` instruction sets the system call signal that will be sent to the
-container to exit. This signal can be a signal name in the format `SIG<NAME>`,
-for instance `SIGKILL`, or an unsigned number that matches a position in the
-kernel's syscall table, for instance `9`. The default is `SIGTERM` if not
-defined.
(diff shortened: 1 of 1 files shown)


lacatoire:

PR #24126, opened 2026-02-10
fix: remove named volume from volume prune example

### LINKED ISSUE #24007: Inconsistency between example and command description
Hi,

The command description indicates "By default, it only removes anonymous volumes." but the example shows a named volume.

https://github.com/docker/docs/blob/9103dff58a73faa869ad3ac098b63ef187ef7f94/data/engine-cli/docker_volume_prune.yaml#L63

regards

lacatoire:
## Description

The `docker volume prune` documentation states that by default it only removes anonymous volumes. However, the example output included `my-named-vol`, which is a named volume. Removed it from the example to match the documented behavior.

Fixes #24007

### DIFF
--- a/data/engine-cli/docker_volume_prune.yaml
+++ b/data/engine-cli/docker_volume_prune.yaml
@@ -60,7 +60,6 @@ examples: |-
     Are you sure you want to continue? [y/N] y
     Deleted Volumes:
     07c7bdf3e34ab76d921894c2b834f073721fccfbbcba792aa7648e3a7a664c2e
-    my-named-vol
 
     Total reclaimed space: 36 B
     ```


data/engine-cli/docker_volume_prune.yaml
```diff
@@ -60,7 +60,6 @@ examples: |-
     Are you sure you want to continue? [y/N] y
     Deleted Volumes:
     07c7bdf3e34ab76d921894c2b834f073721fccfbbcba792aa7648e3a7a664c2e
-    my-named-vol
```
thaJeztah, line 63: This file is generated, so will be overwritten the next time it's re-generated; to fix this, we should have a pull request in the docker/cli repository to fix the source; https://github.com/docker/cli/blob/df016a3a9538efb6d7d6e7c3baf291dbbe28eee0/docs/reference/commandline/volume_prune.md?plain=1#L31

thaJeztah: I opened a PR in the CLI repository;

- https://github.com/docker/cli/pull/6787

thaJeztah (approved):
LGTM, thanks!

PR #24137, opened 2026-02-12
settings management doc / verified domain stuff

akristen:
Per Tech support sync, adding clarifying verbiage to settings management doc.


craig-osterhout (approved):
LGTM

PR #24149, opened 2026-02-16
vendor: github.com/docker/model-runner/cmd/cli v1.1.9-0.20260303081710-59280ed7abd5

doringeman:
## Description

Updated Go to 1.26.0 for https://github.com/docker/model-runner/blob/main/go.mod.

Update model-cli docs for Docker Model Runner.
See changes in https://github.com/docker/model-runner/compare/cmd/cli/v1.0.3...59280ed7abd5dc3859c6ed83277af7b868e08d79.

## Reviews

- [x] Technical review
- [ ] Editorial review
- [ ] Product review


Dockerfile
```diff
@@ -1,8 +1,8 @@
 # syntax=docker/dockerfile:1
 # check=skip=InvalidBaseImagePlatform
 
-ARG ALPINE_VERSION=3.21
-ARG GO_VERSION=1.25
+ARG ALPINE_VERSION=3.23
+ARG GO_VERSION=1.25.6
```
dvdksn: Fine to update the directive in `go.mod` but I don't think this is necessary - we only point to minor versions, not patch.

go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/docker/docs
 
-go 1.25.0
+go 1.25.6
```
thaJeztah: FWIW; we should not bump patch versions in `go.mod` in our modules (if we control it); ideally even remove the `.patch` from `go.mod`; see

- https://github.com/docker/compose/pull/13417#discussion_r2586218728
- https://github.com/docker/compose/pull/13418

doringeman:
Thanks @thaJeztah @dvdksn, I removed the patch version.

PR #24238, opened 2026-02-27
Remove call outs per style guide

akristen:
This PR captures work for docs in BRIDGE that have >2 but <5 CTAs per doc.

* Keep most pertinent/urgent/helpful CTAs with a maximum of ~~2~~ 3.
    * Some CTAs include security info that should be privileged / override style guidance
    * I've limited scope for this PR to address docs with 4-5 CTAs per doc.
* Find natural home for other CTAs, careful to ensure that the information is still prominent as the CTAs were likely added because users need that information. 

Follow-up work: 

* Some work was started on SCIM and Configure JSON, but both docs need a deeper look into about their efficacy. 
    * For SCIM especially, the doc has a lot of relevant call outs.
    * I want to confirm with stakeholders about possibly segmenting this doc into smaller and more topic-focused docs within the provisioning section
* Need to find point person on legacy notices in docs
* Need to find point person on tax compliance notices, as they're quite noisy. (Not urgent as this is just a personal preference lol)

### DIFF
--- a/content/manuals/subscription/manage-seats.md
+++ b/content/manuals/subscription/manage-seats.md
@@ -31,7 +31,7 @@ organization.
 1. Select **Billing**.
    - If you choose to pay using a US bank account, you must verify the account. 
    - For more information, see [Verify a bank account](manuals/billing/payment-method.md#verify-a-bank-account).
-2. Select **Add seats** and follow the on-screen instructions to complete
+1. Select **Add seats** and follow the on-screen instructions to complete
 adding seats.
    - You can't use pay by invoice for purchasing additional seats.
    - You must use a card or US bank account.
@@ -44,9 +44,9 @@ You can now add more members to your organization. For more information, see [Ma
 To add seats to your Legacy Docker subscription:
 
 1. Sign in to [Docker Hub](https://hub.docker.com).
-2. Select your organization, then select **Billing**.
-3. On the Billing page, select **Add seats**.
-4. Select the number of seats you want to add, then select **Purchase**.
+1. Select your organization, then select **Billing**.
+1. On the Billing page, select **Add seats**.
+1. Select the number of seats you want to add, then select **Purchase**.
 
 To add seats to Docker Build Cloud:
 


.claude/settings.local.json
dvdksn, line 1: Remove this (feel free to add `.claude` to gitignore or .git/info/exclude if you need these)

aevesdocker, line 1: lol classic

akristen: oh weird, good catch

content/manuals/enterprise/security/enforce-sign-in/methods.md
```diff
@@ -245,9 +245,7 @@ Create the `registry.json` file (UTF-8 without BOM) at the appropriate location:
 1. Restart Docker Desktop.
 1. Verify the `Sign in required!` prompt appears in Docker Desktop.
 
-> [!TIP]
->
-> If users have issues starting Docker Desktop after enforcing sign-in,
+If users have issues starting Docker Desktop after enforcing sign-in,
```
aevesdocker: I actually quite like this as a tip. There is an 'Important' CTA further down that talks about desktop versions which can instead be removed as we nolonger support those mentioned versions 🤝

content/manuals/subscription/manage-seats.md
```diff
@@ -19,42 +19,34 @@ You can add or remove seats from your Docker Team or Business subscription at an
 
 ## Add seats to your subscription
 
+If you have a sales-assisted Docker Business subscription, contact your account manager to add seats to your subscription.
+
 {{< tabs >}}
 {{< tab name="Docker subscription" >}}
 
-> [!IMPORTANT]
```
aevesdocker, line 25: I understand why you've made this change, but I think it actually gets lost moving it out of the callout 🤔 

Maybe it's worth having a combined callout at the top saying "If you have a sales-assisted Docker Business subscription, contact your account manager to add or remove seats to your subscription".
I'd then remove the tax callout (pretty sure you'll get the go-ahead to remove it anyway)

aevesdocker (commented):
Great cleanup overall, just a few suggestions

content/manuals/enterprise/security/enforce-sign-in/methods.md
akristen: Good trade off! 🖖

content/manuals/subscription/manage-seats.md
akristen: ok, sweet. i'll make a follow up task to figure out what to do with all of these tax compliance call outs tbh.

PR #24243, opened 2026-02-28
Document Claude Code integration for Docker Model Runner

tfenster:
## Description

Added instructions on how to use Claude Code based on the official blog post for the topic

## Reviews

- [ ] Technical review
- [ ] Editorial review
- [ ] Product review

### DIFF
--- a/content/manuals/ai/model-runner/ide-integrations.md
+++ b/content/manuals/ai/model-runner/ide-integrations.md
@@ -267,7 +267,7 @@ You can find more details in [this Docker Blog post](https://www.docker.com/blog
 1. Install Claude Code (see [docs](https://code.claude.com/docs/en/quickstart#step-1-install-claude-code))
 2. Use the `ANTHROPIC_BASE_URL` environment variable to point Claude Code at DMR. On Mac or Linux, you can do this, for example if you want to use the `gpt-oss:32k` model:
     ```bash
-    ANTHROPIC_BASE_URL=http://localhost:12434 claude --model gpt-oss:32k
+    ANTHROPIC_BASE_URL=http://localhost:12434 claude --model qwen2.5-coder
     ```
     On Windows (PowerShell) you can do it like this:
     ```powershell


content/manuals/ai/model-runner/ide-integrations.md
```diff
@@ -258,6 +258,24 @@ print(response.text)
 
 You can find more details in [this Docker Blog post](https://www.docker.com/blog/opencode-docker-model-runner-private-ai-coding/)
 
+## Claude Code
+
+[Claude Code](https://claude.com/product/claude-code) is [Anthropic's](https://www.anthropic.com/) command-line tool for agentic coding. It lives in your terminal, understands your codebase, and helps you code faster by executing routine tasks, explaining complex code, and handling Git workflows through natural language commands.
```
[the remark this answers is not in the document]
tfenster: adjusted

content/manuals/ai/model-runner/ide-integrations.md
```diff
@@ -258,6 +258,24 @@ print(response.text)
...
+
+### Configuration
+
+1. Install Claude Code (see [docs](https://code.claude.com/docs/en/quickstart#step-1-install-claude-code))
+2. Use the `ANTHROPIC_BASE_URL` environment variable to point Claude Code at DMR. On Mac or Linux you can do this e.g. like this if you want to use the `gpt-oss:32k` model:
```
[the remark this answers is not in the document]
tfenster: adjusted

tfenster:
@dvdksn Is there anything still left to do for me?

content/manuals/ai/model-runner/ide-integrations.md
```diff
@@ -258,6 +258,24 @@ print(response.text)
...
+## Claude Code
+
+[Claude Code](https://claude.com/product/claude-code) is [Anthropic's](https://www.anthropic.com/) command-line tool for agentic coding. It lives in your terminal, understands your codebase, and executes routine tasks, explains complex code, and handles Git workflows through natural language commands.
+
+### Configuration
+
+1. Install Claude Code (see [docs](https://code.claude.com/docs/en/quickstart#step-1-install-claude-code))
+2. Use the `ANTHROPIC_BASE_URL` environment variable to point Claude Code at DMR. On Mac or Linux, you can do this, for example if you want to use the `gpt-oss:32k` model:
+    ```bash
+    ANTHROPIC_BASE_URL=http://localhost:12434 claude --model gpt-oss:32k
```
brandonh6k: The `gpt-oss:32k` example is great, but readers won't know what that is since it's a locally repackaged model. Could we add a brief note explaining that `ai/gpt-oss` defaults to only 4,096 tokens of context (not great for coding), and show how to repackage it?

  Something like:
  > [!TIP]
  > The default context size for `gpt-oss` is 4,096 tokens, which is limiting for coding tasks.
  > You can repackage it with a larger context window:
  > ```console
  > $ docker model pull gpt-oss
  > $ docker model package --from ai/gpt-oss --context-size 32000 gpt-oss:32k
  > ```
  > Alternatively, models like `ai/glm-4.7-flash`, `ai/qwen3-coder`, and `ai/devstral-small-2`
  > come with 128K context by default and work without repackaging.

content/manuals/ai/model-runner/ide-integrations.md
```diff
@@ -258,6 +258,24 @@ print(response.text)
...
 You can find more details in [this Docker Blog post](https://www.docker.com/blog/opencode-docker-model-runner-private-ai-coding/)
 
+## Claude Code
+
+[Claude Code](https://claude.com/product/claude-code) is [Anthropic's](https://www.anthropic.com/) command-line tool for agentic coding. It lives in your terminal, understands your codebase, and executes routine tasks, explains complex code, and handles Git workflows through natural language commands.
+
+### Configuration
+
+1. Install Claude Code (see [docs](https://code.claude.com/docs/en/quickstart#step-1-install-claude-code))
+2. Use the `ANTHROPIC_BASE_URL` environment variable to point Claude Code at DMR. On Mac or Linux, you can do this, for example if you want to use the `gpt-oss:32k` model:
```
brandonh6k, line 268: It might be worth adding a tip about making the env var persistent so users don't have to set it every time:

  > [!TIP]
  > To avoid setting the variable each time, add it to your shell profile
  > (`~/.bashrc`, `~/.zshrc`, or equivalent):
  > ```bash
  > export ANTHROPIC_BASE_URL=http://localhost:12434
  > ```

content/manuals/ai/model-runner/ide-integrations.md
```diff
@@ -258,6 +258,24 @@ print(response.text)
 
 You can find more details in [this Docker Blog post](https://www.docker.com/blog/opencode-docker-model-runner-private-ai-coding/)
 
+## Claude Code
+
+[Claude Code](https://claude.com/product/claude-code) is [Anthropic's](https://www.anthropic.com/) command-line tool for agentic coding. It lives in your terminal, understands your codebase, and executes routine tasks, explains complex code, and handles Git workflows through natural language commands.
```
brandonh6k, line 263: Since every other tool on this page uses the OpenAI-compatible endpoint, it might be worth a one-liner noting that DMR also exposes an Anthropic-compatible API, which is why this integration works differently. Otherwise readers might wonder why this section doesn't use `/engines/v1` like everything else.

brandonh6k:
Nice addition! The blog post this references actually has quite a bit more detail that I think would be valuable to include here so readers don't have to click through. A few suggestions above.

content/manuals/ai/model-runner/ide-integrations.md
tfenster: Good suggestion, but as this applies to other IDEs, I would rather put it here? https://docs.docker.com/ai/model-runner/ide-integrations/#slow-responses-or-timeouts

brandonh6k: You could put the general instructions in the `"Model not found" errors` section as well. Or maybe up top in the https://docs.docker.com/ai/model-runner/ide-integrations/#prerequisites section.  Probably a good idea regardless.  

The reason I had it inline in this guide is all the other IDE configurations on this page all use actual models, not a repackaged DMR model, which means you can't use the commands in this guide verbatim without doing something else beforehand (e.g., repackaging the model).  The other guides all work exactly as written.

tfenster: @brandonh6k ok, then let me change the model here so that it is consistent and in a second PR add the repackaging stuff to the prereqs. Makes sense?

brandonh6k: Works.  Thanks!

tfenster:
@brandonh6k I think now everything should be good? Thanks a lot for all your suggestions

dvdksn (approved):
LGTM - thank you @tfenster (and @brandonh6k)!

PR #24259, opened 2026-03-03
Update release-notes.md

mat007:
## Description

Acknowledge a bug with the MSI preventing updates.

## Related issues or tickets

https://docker.atlassian.net/browse/DKP-2697?atlOrigin=eyJpIjoiNjQ0MTRiNThiM2VkNGM1NmJiZjEwMGU2ZmVhNmU0NjMiLCJwIjoiamlyYS1zbGFjay1pbnQifQ

## Reviews

- [ ] Technical review
- [ ] Editorial review
- [ ] Product review

### DIFF
--- a/content/manuals/desktop/release-notes.md
+++ b/content/manuals/desktop/release-notes.md
@@ -54,6 +54,10 @@ For more frequently asked questions, see the [FAQs](/manuals/desktop/troubleshoo
 
 - Improved startup time on Windows.
 
+### Known issues
+
+- The Windows MSI installer cannot update an existing Docker Desktop installation when the current version is 4.59 or later. As a workaround, uninstall the existing version before reinstalling. Note that uninstalling removes all associated data. A fix will be provided in a future MSI release.
+
 ## 4.62.0
 
 {{< release-date date="2026-02-23" >}}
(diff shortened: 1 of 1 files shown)


mat007:
@aevesdocker I’d love guidance here not just on the wording but also on where to add this exactly, and if we need to mention it in all versions until we fix it?

aevesdocker:
👋  @mat007

I think having it under each entry like you do is the best way forward. What's the timeline on a new fix?

Lets adjust the wording to: 'The Windows MSI installer cannot update an existing Docker Desktop installation when the current version is 4.59 or later. As a workaround, uninstall the existing version before reinstalling. Note that uninstalling removes all associated data. A fix will be provided in a future MSI release.'

Does it also mean developers don't get release notifications?

mat007:
> What's the timeline on a new fix?

We’re thinking 4.66 (March 23).

> Does it also mean developers don't get release notifications?

The notification behaviour shouldn’t change. But there is no in-app updates with MSI-based installations anyway.

aevesdocker:
> We’re thinking 4.66 (March 23).

Ok cool, repeating it until then will be fine 👍

mat007:
I updated the text. Thanks @aevesdocker !

PR #24326, opened 2026-03-10
docs: add provenance flags to CI/CD workflow example in DHI scan page

ajeetraina:
## Description

Updated the "Build Docker image" step in the GitHub Actions workflow example under "Automate DHI scanning in CI/CD with Docker Scout" to include `--provenance=mode=max`, `--sbom=true`, and `--push` flags.

## Reviews

- [ ] Technical review
- [ ] Editorial review
- [ ] Product review

### DIFF
--- a/content/manuals/dhi/how-to/scan.md
+++ b/content/manuals/dhi/how-to/scan.md
@@ -150,8 +150,7 @@ lifecycle.
 
 #### Example GitHub Actions workflow
 
-The following is a sample GitHub Actions workflow that builds an image and scans
-it using Docker Scout:
+The following is a sample GitHub Actions workflow that builds an image, scans it and pushes to the registry only if the scan passes:
 
 ```yaml {collapse="true"}
 name: DHI Vulnerability Scan


content/manuals/dhi/how-to/scan.md
```diff
@@ -205,6 +209,15 @@ The `exit-code: true` parameter ensures that the workflow fails if any critical
 high-severity vulnerabilities are detected, preventing the deployment of
 insecure images.
 
+> [!NOTE]
+>
+> The `--provenance=mode=max` and `--sbom=true` flags are required so that
+> Docker Scout can trace the DHI base image lineage and correctly apply its
+> VEX statements. The `--push` flag is also required, as attestations can
```
fazlan-nazeem: if the `--push` option is mandatory, I would prefer to have the same reasoning mentioned [here](https://docs.docker.com/build/ci/github-actions/attestations/#max-level-provenance) to be consistent. 

However, if there is a way we could avoid --push, I would prefer that approach as the usual flow in CI would be to push the image to the registry only if the CVE scan passes their requirements. Having to to push the image and attestations before the CVE scan, might be a problem in most workflows.

ajeetraina: @fazlan-nazeem Great point. I investigated into this and tested two approaches on GHA. https://github.com/ajeetraina/dhi-containerd-test/actions/runs/22900957211

The `--push` flag can be avoided by enabling the containerd image store, which allows attestations to be stored locally without pushing to a registry first. This enables the correct CI order: build → scan → push only if clean.

Without the containerd image store, Docker Engine rejects the build entirely with:

```
Attestation is not supported for the docker driver.
Switch to a different driver, or turn on the containerd image store, and try again.
```

I've validated this with a two-job GitHub Actions workflow ([test run here](https://github.com/ajeetraina/dhi-containerd-test/actions/runs/22900957211)):

- test-with-containerd - build → scan → push succeeded
- test-without-containerd -  build failed immediately with the above error

I'll update the PR to use `docker/setup-docker-action` with the `containerd` image store enabled, remove `--push` from the build step, and gate the push behind if: success() after the scan. This gives users the security-correct workflow they'd expect.

fazlan-nazeem:
@craig-osterhout would you be able to review and merge this, please?

PR #24338, opened 2026-03-11
engine: slight tweaks to client sdk examples

thaJeztah:
I noticed that building the examples would produce a client with the default ("Go-http-client/1.1") User-Agent; this is something to look into (probably should have a saner default), but regardless, it's good practice to set your own;

- Update examples to show setting the User-Agent
- Use "log.Fatal" instead of "panic"; both exit the binary, but probably are slightly more readable.
- Use the authconfig.Encode utility, which codifies the canonical format to encode the JSON, so that users don't have to replicate the same format.

## Reviews

- [ ] Technical review
- [ ] Editorial review
- [ ] Product review

### DIFF
--- a/content/reference/api/engine/sdk/examples.md
+++ b/content/reference/api/engine/sdk/examples.md
@@ -39,6 +39,7 @@ package main
 import (
 	"context"
 	"io"
+	"log"
 	"os"
 
 	"github.com/moby/moby/api/pkg/stdcopy"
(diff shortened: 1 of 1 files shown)


thaJeztah:
> Note: The review identified pre-existing bugs in the examples (incorrect use of .Items field on slices returned by ContainerList and ImageList APIs), but these are not introduced by this PR and exist in unchanged code lines. They should be addressed in a separate PR

@docker-agent check the docs; examples are valid; is your information outdated? See https://pkg.go.dev/github.com/moby/moby/client

PR #24393, opened 2026-03-17
Fixes https://github.com/docker/docs/issues/23958

mohithshuka:
The `security_opt` field in the Compose services reference did not document 
the accepted syntax variants, which was confusing for users.

Added a note explaining that options accept `option=value`, `option:value`, 
or bare `option` syntax (for booleans). Added examples showing all three 
equivalent forms for `no-new-privileges`.


aevesdocker (approved):
Thanks

PR #24404, opened 2026-03-18
Remove Swarm API Key from Docs

vallieres:
## Description

Removing secrets from documentation.

## Reviews

- [ ] Technical review
- [ ] Editorial review
- [ ] Product review

### DIFF
--- a/content/guides/orchestration.md
+++ b/content/guides/orchestration.md
@@ -191,7 +191,7 @@ Docker Desktop runs primarily on Docker Engine, which has everything you need to
 
    To add a worker to this swarm, run the following command:
 
-       docker swarm join --token SWMTKN-1-3e0hh0jd5t4yjg209f4g5qpowbsczfahv2dea9a1ay2l8787cf-2h4ly330d0j917ocvzw30j5x9 192.168.65.3:2377
+       docker swarm join --token SWMTKN-1-aaa 192.168.65.3:2377
 
    To add a manager to this swarm, run 'docker swarm join-token manager' and follow the instructions.
    ```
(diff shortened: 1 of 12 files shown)


dvdksn:
@vallieres - see bot comment. This needs to be updated in the docker/cli repo upstream. But not sure why we need to change these?

thaJeztah:
Do you know if the secrets detection on GitHub allows for patterns for example values? Ideally we'd still show a key with the same format and length; I recall there was a thread about that, but not sure what the outcome was (if github  supported it)

vallieres:
> Do you know if the secrets detection on GitHub allows for patterns for example values? Ideally we'd still show a key with the same format and length; I recall there was a thread about that, but not sure what the outcome was (if github supported it)

Sadly it's not possible. The most we can do is provide negative lookahead regexp but it's easily getting out of hand if we start adding a rule for each token...

vallieres:
> @vallieres - see bot comment. This needs to be updated in the docker/cli repo upstream. But not sure why we need to change these?

These keep on triggering secrete scanning tools as they look like valid Swarm tokens. And since this is being replicated in a few other places, we easily 50+ notifications regularly. Trying to fix this at the source so we stop tripping the wire for these tools. I'll get the cli repo updated.

https://github.com/docker/cli/pull/6867

thaJeztah:
> > Do you know if the secrets detection on GitHub allows for patterns for example values? Ideally we'd still show a key with the same format and length; I recall there was a thread about that, but not sure what the outcome was (if github supported it)
> 
> Sadly it's not possible. The most we can do is provide negative lookahead regexp but it's easily getting out of hand if we start adding a rule for each token...

Yeah; I recall we discussed a separate prefix for demo purposes as alternative. I somewhat expected them to have some approach for that as it's not uncommon to have tokens for documentation or tests that should be ignored (not just these tokens)

Perhaps we can do something like `SWMTKN-1-aabbccddeeff112233.....` to closer match the expected format

vallieres:
> > > Do you know if the secrets detection on GitHub allows for patterns for example values? Ideally we'd still show a key with the same format and length; I recall there was a thread about that, but not sure what the outcome was (if github supported it)
> > 
> > 
> > Sadly it's not possible. The most we can do is provide negative lookahead regexp but it's easily getting out of hand if we start adding a rule for each token...
> 
> Yeah; I recall we discussed a separate prefix for demo purposes as alternative. I somewhat expected them to have some approach for that as it's not uncommon to have tokens for documentation or tests that should be ignored (not just these tokens)
> 
> Perhaps we can do something like `SWMTKN-1-aabbccddeeff112233.....` to closer match the expected format

If I use the same format, they will once again trigger the secret detection. The only way to skip over that is to "break" the pattern with something that does not fit the length and regexp.

PR #24447, opened 2026-03-20
Fix issue #24010: warn about containerd separate storage path

### LINKED ISSUE #24010: Migration steps to containerd should include details for those not using the default docker data path
### Is this a docs issue?

### Type of issue

Other

### Description

Accidentally filled my root partition today following this guide because i had my docker data path set to a different folder and i didn’t realise that containerd used a separate storage path.
The documentation didn’t make it initially clear that you needed to configure the storage location for containerd separately from the docker data.

the docs lines
```
If disk space is constrained, consider the following:

    Regularly prune unused images with docker image prune
    Use docker system df to monitor disk usage
    [Configure the data directory](https://docs.docker.com/engine/daemon/#configure-the-data-directory-location) to use a partition with adequate space
```
Implied to me that i wouldn't need to make any changes because I had previously configured the data directory.

IMO it might be worth adding a note here for anyone who previously configured a data directory that they need to set one for containerd.

 

### Location

https://docs.docker.com/engine/storage/containerd/

### Suggestion

_No response_

dvdksn:
Adds warning about containerd using a separate storage path from the Docker data directory.

Changes:
- Added IMPORTANT callout in "Disk space usage" section explaining that containerd uses separate storage
- Warns users with custom data directories that they need to configure containerd separately
- Links to containerd configuration documentation for data directory setup

This prevents users from unexpectedly filling their root partition when they assumed their custom data directory configuration would apply to containerd.

Fixes #24010

### DIFF
--- a/content/manuals/engine/storage/containerd.md
+++ b/content/manuals/engine/storage/containerd.md
@@ -62,6 +62,16 @@ uncompressed layers are still de-duplicated through snapshotters. The
 compressed storage adds overhead proportional to the number of images using
 those layers.
 
+> [!IMPORTANT]
+> containerd uses a separate storage path from the Docker data directory.
+> If you previously configured a custom data directory for Docker (for example,
+> to use a different partition), containerd's storage is not automatically
+> moved to that location. You need to configure containerd's data directory
+> separately to avoid filling your root partition.
+>
+> To configure containerd's data directory, see
+> [Configure the data directory location](../daemon/_index.md#configure-the-data-directory-location).
+
 If disk space is constrained, consider the following:
 
 - Regularly prune unused images with `docker image prune`


dvdksn:
Thank you for the feedback on both the style violation and link stability concern. You're right on both counts.

I've made the following changes:
- Removed the bold emphasis (our style guide reserves bold for UI elements only, not for emphasis)
- Updated the link to point to the stable v2.0.0 release instead of the latest tag, which should provide better long-term stability

The warning now follows our documentation standards while maintaining a reliable reference.

SuperNinja-4965:
I would also suggest linking to: https://docs.docker.com/engine/daemon/#configure-the-data-directory-location
The linked page does not specify the config file location and contains alot of information alot of users will not necessarily need.

dvdksn:
Thanks for the suggestion, @SuperNinja-4965. The bullet list below the callout already links to the internal Docker docs page for configuring Docker's data directory. The callout's external link points specifically to the containerd configuration documentation, which covers configuring containerd's own data directory — a separate step from configuring Docker's data directory. The two links serve different purposes.

Generated by [Claude Code](https://claude.com/claude-code)

dvdksn:
You're right, @SuperNinja-4965 — the Docker daemon docs page already covers configuring containerd's data directory. Updated the link to point to the internal docs instead.

dvdksn:
Sorry about that, my robot underling is not behaving this monday

craig-osterhout (approved):

PR #24468, opened 2026-03-22
hack: stop extraction at footer in moby-sync-release.sh

Himanshu2459:
## Reviews

- [ ] Technical review
- [ ] Editorial review
- [ ] Product review

### DIFF
--- a/hack/moby-sync-release.sh
+++ b/hack/moby-sync-release.sh
@@ -21,7 +21,8 @@ github_notes=$(mktemp -t old)
 grep -A 10000 "## ${version}" "content/manuals/engine/release-notes/${major_version}.md" | \
     grep -m 2 -A 0 -B 10000 '^## ' | \
     sed '$d' | \
-    sed '/{{< release-date /{N;d;}' \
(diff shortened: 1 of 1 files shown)


vvoland:
Curious, what's your use case for this script?

Himanshu2459:
I don't use this script myself — I spotted #24533 and traced the 404 to the autogenerated footer not being stripped during extraction. This just adds a sed to cut everything from --- onwards so only the actual notes get pulled in.

dvdksn:
Claude thinks this is incorrect. The `---` thematic break in the 29.0.0 section isn't causing a problem here. The real issue is that the `grep -m 2 -B 10000 '^## '` extraction logic relies on finding two `## ` headings (the target version and the next one). For the last version in a file there's no second heading, so the pipeline produces 0 lines of output — regardless of the `---`.

Additionally, the proposed `sed '/^---$/,$d'` would be harmful if the extraction were fixed: it would delete from the `---` to the end, dropping the `### New`, `### Bug fixes and enhancements`, and all other subsections after the thematic break in the 29.0.0 section.

PR #24503, opened 2026-03-24
dhi: update customizations

craig-osterhout:
## Description

Follow up to #24500. I incorrectly interpreted and tested only the distros and not the distro versions.

The UI wording is dynamic based on distro *and* version. I removed the UI wording instead of updating as it makes the simple step more complex than necessary by having to explain an insignificant UI quirk. Added the explanation to the paragraph below the main step text.

Debian UI wording is **Packages**
Alpine 3.22 UI wording is **Packages**
Alpine 3.23 UI wording is **Hardened Packages**

## Related issues or tickets

ENGDOCS-3218

## Reviews

- [ ] Editorial review

### DIFF
--- a/content/manuals/dhi/how-to/customize.md
+++ b/content/manuals/dhi/how-to/customize.md
@@ -77,14 +77,15 @@ To customize a Docker Hardened Image using the web interface, follow these steps
 1. Select the image version you want to customize.
 1. Optional. Add packages.
 
-   1. In the **Packages** drop-down, select the packages you want to add to the
+   1. In the packages drop-down, select the packages you want to add to the
       image.
 
       The packages available in the drop-down are OS system packages for the
-      selected image variant. For Alpine-based images, these are hardened
-      packages that have been built from source by Docker with cryptographic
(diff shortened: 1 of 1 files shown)


content/manuals/dhi/how-to/customize.md
```diff
@@ -77,14 +77,15 @@ To customize a Docker Hardened Image using the web interface, follow these steps
...
-   1. In the **Packages** drop-down, select the packages you want to add to the
+   1. In the packages drop-down, select the packages you want to add to the
       image.
 
       The packages available in the drop-down are OS system packages for the
-      selected image variant. For Alpine-based images, these are hardened
-      packages that have been built from source by Docker with cryptographic
-      signatures and full supply chain security. For Debian-based images,
-      these are standard Debian system packages.
+      selected image variant. For version 3.23 Alpine-based images, these are
```
dvdksn, line 84: I guess this won't be evergreen - in about two months from now we'll have 3.24 images. So we'll either need to go back here and update at that point, or we rephrase this to say call out say something like Alpine images, except 3.22, contain hardened packages, ...

PR #24526, opened 2026-03-25
engine: 29.3.1

vvoland:


vvoland:
Right! We should automate that part 😅

PR #24551, opened 2026-03-27
Update lifecycle.md

ilya-grigoriev:
Swapped the parameters for `chown` command


content/manuals/compose/how-tos/lifecycle.md
dvdksn, line 41: This example is also wrong - should be a string (`"1001"`) not a number (`1001`)

PR #24560, opened 2026-03-28
Fix: Corrected /. to ./ in bind_mounts under the git bash tab in bind mounts docs

alatos2:
## Description

I was practicing how to use bind mounts via git bash
I used the command as seen on the docs: docker run -it --mount type=bind,src="/.",target=/src ubuntu bash
It didn't work until I changed src="/." to src="./"

## Reviews

- [ ] Technical review
- [ ] Editorial review
- [ ] Product review


dvdksn:
Thanks!

PR #24587, opened 2026-03-31
Clarify legacy builder note for Windows container builds

### LINKED ISSUE #24453: [docs-scanner] Confusing legacy builder note in Builds documentation
**File:** `content/manuals/desktop/use-desktop/builds.md`

### Issue

The note about Windows container builds contains confusing time-relative language:

> When building Windows container images using the `docker build` command, the legacy builder is used which does not populate the **Builds** view.

The phrase "the legacy builder is used" describes current behavior but frames it as temporary/legacy without explaining why this limitation exists or if it will change. This creates uncertainty for readers about whether this is expected behavior or a bug.

### Suggested fix

Clarify the current state more directly:

"When building Windows container images, the `docker build` command uses a legacy builder that doesn't populate the **Builds** view. To use BuildKit and see builds in this view, either:"

This removes the passive "is used" construction and makes it clear this is current expected behavior, while still providing the workaround.

---
*Found by nightly documentation freshness scanner*

lacatoire:
## Description

The note about Windows container builds uses passive voice ("the legacy builder is used") which makes the behavior unclear. Readers can't tell if this is expected behavior or a temporary limitation.

Rewritten in active voice to make the cause and workaround clear: the `docker build` command uses a legacy builder, and here is how to switch to BuildKit to see builds in the Builds view.

Fixes #24453

### DIFF
--- a/content/manuals/desktop/use-desktop/builds.md
+++ b/content/manuals/desktop/use-desktop/builds.md
@@ -16,9 +16,7 @@ connected to the same cloud builder.
 
 > [!NOTE]
 >
-> When building Windows container images, the `docker build` command uses a legacy builder that does not populate the **Builds** view. To use BuildKit and see builds in this view, you can either:
-> - Set `DOCKER_BUILDKIT=1` in the build command, such as `DOCKER_BUILDKIT=1 docker build .` or
-> - Use the `docker buildx build` command
+> Windows container image builds use the legacy builder and do not appear in the **Builds** view. Only BuildKit-powered builds are shown here.
 
 ## Show build list
 


content/manuals/desktop/use-desktop/builds.md
```diff
@@ -16,7 +16,7 @@ connected to the same cloud builder.
 
 > [!NOTE]
 >
-> When building Windows container images using the `docker build` command, the legacy builder is used which does not populate the **Builds** view. To switch to using BuildKit, you can either:
+> When building Windows container images, the `docker build` command uses a legacy builder that does not populate the **Builds** view. To use BuildKit and see builds in this view, you can either:
 > - Set `DOCKER_BUILDKIT=1` in the build command, such as `DOCKER_BUILDKIT=1 docker build .` or
 > - Use the `docker buildx build` command
```
dvdksn: This note is a bit confusing imo. It leads with "When building Windows containers..." and then concludes with setting `DOCKER_BUILDKIT=1` or using `docker buildx build`. BuildKit does not really support WCOW (experimental)... so the more correct thing to say here would be to say that Windows image builds do not appear in the builds view; only BuildKit-powered builds do.

PR #24602, opened 2026-04-01
dhi: refresh how-to explore

craig-osterhout:
## Description

- Refreshed explore based on latest GUI. Removed steps with a direct link. xref'd to previous sections to remove redundant steps.
- Consolidated explore and compare to minimize left nav sprawl and because they're related use-cases.
- Renamed from "explore" to "search" in order to not conflict with the about section that got renamed explore.
- Added in CLI commands to eventually remove the bolted-on, standalone how-to cli topic.
- Cleaned up some fluff from the compare topic that was used to make it feel more substantial when it was a standalone.

https://deploy-preview-24602--docsdocker.netlify.app/dhi/how-to/explore/

## Related issues or tickets

ENGDOCS-3222

## Reviews

- [ ] Editorial review


craig-osterhout:
> Minor observation (not blocking): Several internal links throughout the codebase still reference /dhi/how-to/compare/ rather than pointing directly to /dhi/how-to/explore/. While the alias ensures these links work correctly, updating them to point directly to the new location would follow best practice for internal link management. This is purely a housekeeping suggestion and doesn't affect functionality.
 
I didn't find any other `/how-to/compare/` via search.

PR #24694, opened 2026-04-11
Add nftables compatibility note for Arch Linux in rootless troubleshoot docs

shivani0234:
Arch Linux recently switched the default iptables backend to nftables.

This change can cause compatibility issues with rootless Docker networking.
This PR adds a troubleshooting note for Arch Linux users explaining the
potential issue and referencing the Arch Linux announcement.

Reference:
https://archlinux.org/news/iptables-now-defaults-to-the-nft-backend/


dvdksn:
Closing this. The root cause was outdated Docker binaries installed via the rootless setup script — with a current version, rootless Docker handles the nftables backend without issues (confirmed by @AkihiroSuda with a clean install). The suggestion about the setup script surfacing version warnings is worth raising in moby/moby, but it's not a documentation change. Thanks for the detailed writeup.

PR #24733, opened 2026-04-15
Volumes: Fix a wrong statement about named volume

shallowcosmicjumper:
## Description

[Neither named nor anonymous](https://docs.docker.com/engine/storage/volumes/#when-to-use-volumes)  volumes use outside sources, those are for bind mounts. 

Additionally, referring to a volume's name as a "source" here can be confusing (even though [correct](https://docs.docker.com/engine/storage/volumes/#options-for---mount)), as the term has a [different](https://docs.docker.com/engine/storage/bind-mounts/#options-for---mount) definition for bind mounts. Calling it "name" is clear and straightforward.

## Reviews

- [ ] Technical review
- [ ] Editorial review
- [ ] Product review

### DIFF
--- a/content/manuals/engine/storage/volumes.md
+++ b/content/manuals/engine/storage/volumes.md
@@ -775,8 +775,8 @@ testing using your preferred tools.
 A Docker data volume persists after you delete a container. There are two types
 of volumes to consider:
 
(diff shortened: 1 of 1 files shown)


dvdksn (approved):
Thanks!

PR #24750, opened 2026-04-16
docs: note BuildKit ships QEMU emulators for multi-platform builds

### LINKED ISSUE #22899: Please update multi-platform to mention that the official buildkit will do non-native emulation via QEMU
### Is this a docs issue?

### Type of issue

Information is incorrect

### Description

On https://docs.docker.com/build/building/multi-platform/#prerequisites there is the following:
> If you're using Docker Engine standalone and you need to build multi-platform images using emulation, you also need to install QEMU, see [Install QEMU manually](https://docs.docker.com/build/building/multi-platform/#install-qemu-manually).

However, it turns out that non-Docker desktop buildkit itself will quietly try and do QEMU emulation (but as the name implies only at build time)! From https://github.com/moby/buildkit/blob/master/docs/multi-platform.md :
> When your build needs to run a binary for architecture that is not supported natively by your host, it gets executed using a QEMU user-mode emulator. You do not need to set up QEMU manually in most cases.

I only know this because I spent some time scratching my head as to how a foreign architecture's build was succeeding on a machine that didn't have separate QEMU emulation configured (and yet actually running the built foreign architecture image was failing with `exec format error` as expected). At the time a Google search for "docker inbuilt qemu build" didn't turn up anything useful so hopefully creating this issue helps. 

### Location

https://docs.docker.com/build/building/multi-platform/

### Suggestion

Please add a note that since docker version XYZ architecture emulation will automatically be tried using a built-in QEMU.

lacatoire:
Fixes #22899.

The Prerequisites paragraph on `docs.docker.com/build/building/multi-platform/` tells Docker Engine standalone users they "also need to install QEMU" to build multi-platform images with emulation, but official BuildKit releases already bundle QEMU user-mode emulators ([upstream doc](https://github.com/moby/buildkit/blob/master/docs/multi-platform.md#building-multi-platform-images) : "You do not need to set up QEMU manually in most cases").

Reword the Prerequisites pointer and the Install QEMU manually intro so the manual install is presented as a fallback (for example, with a third-party BuildKit package that doesn't ship the bundled emulators) rather than a default step.

### DIFF
--- a/content/manuals/build/building/multi-platform.md
+++ b/content/manuals/build/building/multi-platform.md
@@ -89,8 +89,11 @@ $ docker buildx create \
 > drivers](/manuals/build/builders/drivers/_index.md).
 
 If you're using Docker Engine standalone and you need to build multi-platform
-images using emulation, you also need to install QEMU, see [Install QEMU
-manually](#install-qemu-manually).
+images using emulation, official BuildKit releases bundle QEMU user-mode
+emulators, so in most cases you don't need to install QEMU manually. See
+[Install QEMU manually](#install-qemu-manually) if emulation fails, for
+example with a third-party BuildKit package that doesn't ship the bundled
+emulators.
 
 ## Build multi-platform images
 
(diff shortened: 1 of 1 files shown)


dvdksn (approved):
LGTM thanks

tonistiigi:
Our position should be that if you know you need emulators, you should still install them. The embedded versions in buildkit should be considered a fallback that may work in some cases, not as a replacement to the emulators in the kernel. There are multiple known cases where the embedded emulators do not work the same (mixed-arch images, some argv0 advanced cases).

PR #24769, opened 2026-04-17
engine: document DNS fallback to Google public servers

### LINKED ISSUE #22945: Missing documentation of Google DNS fallback 8.8.8.8 and 8.8.4.4
### Is this a docs issue?

### Type of issue

### Summary

Add a note to the documentation indicating that Docker falls back to Google DNS if no non-localhost DNS is defined.

### Details

In one of my projects, I started a Docker container on a device without internet access and with no configured DNS other than localhost.
Inside the container, some DNS requests timed out after a long delay. I expected the DNS requests to fail immediately.

Upon checking the dockerd log, I found the following line during startup:

No non-localhost DNS nameservers are left in resolv.conf. Using default external servers: [nameserver 8.8.8.8 nameserver 8.8.4.4]

I reviewed the [DNS section of the networking chapter](https://docs.docker.com/engine/network/#dns-services), but found no mention of this behavior. I only discovered it referenced in another issue: https://github.com/moby/moby/issues/23910

Am I missing something?

Edit: Tested with Docker version 24.0.5, build ced0996

### Location

https://docs.docker.com/engine/network/#dns-services

### Suggestion

Maybe add a simple sentence to the chapter:

```
In case no non-localhost DNS is defined in the `/etc/resolve.conf`, a fallback to `8.8.8.8` and `8.8.4.4` will be added.
```

Exact formulation and placement is of course open for discussion.

Nedunchezhiyan-M:
## What

Documents Docker's fallback behavior when the host's `/etc/resolv.conf` has no
non-localhost nameservers: the daemon uses Google's public DNS servers
(`8.8.8.8` and `8.8.4.4`) as a fallback for external name resolution.

## Why

This behavior was entirely undocumented, causing confusion for users in
offline or restricted-network environments who expected DNS lookups to fail
immediately. Users discovered it only by reading `dockerd` log output:

```
No non-localhost DNS nameservers are left in resolv.conf. Using default external servers: [nameserver 8.8.8.8 nameserver 8.8.4.4]
```

See also: https://github.com/moby/moby/issues/23910

Fixes #22945

### DIFF
--- a/content/manuals/engine/network/_index.md
+++ b/content/manuals/engine/network/_index.md
@@ -267,6 +267,11 @@ The embedded DNS server address is `127.0.0.11`.
 There is no IPv6 equivalent; the IPv4 address works even in IPv6-only containers.
 If an application requires an explicit DNS server address, use `127.0.0.11`.
 
+If the host's `/etc/resolv.conf` contains no non-localhost nameservers, the
+Docker daemon falls back to using Google's public DNS servers (`8.8.8.8` and
+`8.8.4.4`) for external name resolution. This fallback also applies when
+using the embedded DNS server.
+
 You can configure DNS resolution on a per-container basis, using flags for the
 `docker run` or `docker create` command used to start the container.
 The following table describes the available `docker run` flags related to DNS


dvdksn (commented):
The fallback is worth documenting, but the current wording is broader than current Engine behavior. In Moby `master`, the Google defaults are added by `TransformForLegacyNw`, which is used for legacy networking (including the default `bridge`) after loopback resolvers are removed. `TransformForIntNS`, used with the embedded DNS server, leaves the external-resolver list empty when none are configured; it does not add Google defaults.

Please scope this note to legacy/default-bridge networking and remove “This fallback also applies when using the embedded DNS server.” Also consider mentioning that IPv6-enabled legacy networks add Google’s IPv6 resolvers (`2001:4860:4860::8888` and `2001:4860:4860::8844`).

Generated by Codex

PR #24815, opened 2026-04-17
engine/logging: fix json-file compress default value

### LINKED ISSUE #21898: default value of compress option for json-file driver
### Is this a docs issue?

### Type of issue

Information is incorrect

### Description

Under "Description", it says the default is `disabled`, but then under "Example value" it says `--log-opt compress=true`, which implies it is a boolean. Shouldn't the default be `false` then?

### Location

https://docs.docker.com/engine/logging/drivers/json-file/

### Suggestion

_No response_

lacatoire:
The `compress` option is parsed as a boolean via `strconv.ParseBool` in `moby/moby daemon/logger/jsonfilelog/jsonfilelog.go`. The previous wording claimed the default was `disabled`, which isn't a value the parser accepts (users copy-pasting it would hit an error). Aligns with the in-file convention already used by `max-size` and `max-file`.

Closes #21898


dvdksn (approved):
Thank you

PR #24820, opened 2026-04-18
Update Dockerfile code snippet for "Using the official Docker image" tab

iamdbychkov:
The code snippet for Dockerfile under the "Using the official Docker images" contains a comment which says that this Docker file uses DHI image, which is incorrect.

Updated the code snippet to match the tab contents 

## Reviews

- [ ] Technical review
- [ ] Editorial review
- [ ] Product review

### DIFF
--- a/content/guides/python/containerize.md
+++ b/content/guides/python/containerize.md
@@ -146,8 +146,7 @@ Create a file named `Dockerfile` with the following contents.
 
 # Want to help us make this template better? Share your feedback here: https://forms.gle/ybq9Krt8jtBL3iCk7
 
-# This Dockerfile uses Docker Hardened Images (DHI) for enhanced security.
-# For more information, see https://docs.docker.com/dhi/
+# This Dockerfile uses Python Docker Official Image
 ARG PYTHON_VERSION=3.12
 FROM python:${PYTHON_VERSION}-slim
 


content/guides/python/containerize.md
```diff
@@ -146,8 +146,7 @@ Create a file named `Dockerfile` with the following contents.
 
 # Want to help us make this template better? Share your feedback here: https://forms.gle/ybq9Krt8jtBL3iCk7
 
-# This Dockerfile uses Docker Hardened Images (DHI) for enhanced security.
-# For more information, see https://docs.docker.com/dhi/
+# This Dockerfile uses Python Docker Official Image
 ARG PYTHON_VERSION=3.12
```
thaJeztah, line 150: Did a quick check where this came from, but looks like it landed in the wrong tab when this file was updated; https://github.com/docker/docs/pull/23749/changes#diff-8220f5dd3cc751cccded6014817c5369e9d6bc004ec49478bdab02d0629d3d25R149-R151

thaJeztah (approved):
LGTM, thx!

thaJeztah:
@dvdksn wondering if we should either adjust the length of the collapsed examples 

Currently the "preview" only shows comments, and the `FROM` is not shown, so the two tabs look almost identical;

Screenshot 2026-04-18 at 13 05 35


Or perhaps remove the comment about the Dockerfile being ... annotated with comments, which feels a bit self-describing (I guess these were based on the `docker init` generated ones though) 


Screenshot 2026-04-18 at 13 19 17

dvdksn:
@thaJeztah yes agreed. I mean, taking a step back, I think these guides could be more pedagogical anyway. The purpose of these tutorials (IMHO) is to teach folks how to use Docker, idiomatically with their language/framework. Ie the focus is not getting them through the step-by-step procedure of bootstrapping an application, which is kinda what it feels like now.

PR #24827, opened 2026-04-20
dhi: add context for multiple OCI artifacts in customize page

### LINKED ISSUE #24819: [docs-scanner] Missing context for 'the list' in OCI artifact file ordering
**File:** `content/manuals/dhi/how-to/customize.md`

### Issue

In the Docker Hub tab of the customization instructions, the documentation states:

> When combining images that contain directories and files with the same path, images later in the list will overwrite files from earlier images. To manage this, you must select paths to include and optionally exclude from each OCI artifact image.

The phrase "images later in the list" assumes readers understand what "the list" refers to, but the Docker Hub UI instructions don't establish this ordering concept. The preceding text says "select the repository" and "select the tag" but doesn't explain that multiple OCI artifacts can be added or that they have an ordering.

### Why this matters

A reader following the Docker Hub instructions would be confused by "later in the list" because:
1. The UI workflow hasn't established that you can add multiple OCI artifacts
2. There's no mention of how ordering is determined or controlled in the UI
3. The CLI and Terraform tabs make ordering clearer (arrays, resource definitions), but the UI tab doesn't

This creates a comprehension gap where readers understand the consequence (later overwrites earlier) but not the mechanism (how to control ordering in the UI).

### Suggested fix

Add context before the overwriting explanation:

> You can add multiple OCI artifact images to a customization. When combining images that contain directories and files with the same path, images added later will overwrite files from images added earlier. To manage this, you must select paths to include and optionally exclude from each OCI artifact image.

Or restructure to explain the UI mechanism:

> The OCI artifacts drop-down lets you add multiple artifact images. If you add multiple artifacts, they're applied in the order you add them. Files from later artifacts overwrite files from earlier artifacts when paths conflict. To control which files are included, select specific paths to include and optionally exclude from each artifact.

---

Nedunchezhiyan-M:
## Summary

The Docker Hub tab's OCI artifacts section told readers that "images later in the list will overwrite files from earlier images," but never established what "the list" refers to. The step-by-step instructions only described selecting a single OCI artifact (repository → tag → paths), so a reader had no context that multiple artifacts can be added.

## Changes

- Adds a sentence before the overwrite explanation: "You can add multiple OCI artifact images to a customization."
- Tightens the following sentence from "images later in the list will overwrite files from earlier images" to "images added later overwrite files from earlier images" for clarity.
- Fixes a typo: `a another` → `another`.
- Replaces `allows you to` with `lets you` (Vale style rule).

Closes #24819

## Reviews

- [x] Technical review
- [x] Editorial review
- [ ] Product review

### DIFF
--- a/content/manuals/dhi/how-to/customize.md
+++ b/content/manuals/dhi/how-to/customize.md
@@ -48,7 +48,6 @@ You can create customizations using either the DHI CLI or the Docker Hub web int
 
 1. Select the image version you want to customize.
 1. Optional. Add packages.
-
    1. In the packages drop-down, select the packages you want to add to the
       image.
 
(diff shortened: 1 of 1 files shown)


craig-osterhout:
Thanks for the contribution. This was resolved.

PR #24839, opened 2026-04-20
remove requires line from summary.yaml

akristen:
Updating SSO requires line


akristen:
@aevesdocker Do we need to remove all specific references of Desktop versions, or only the 4.47 references?

aevesdocker:
Only the reference you've removed

aevesdocker (approved):

PR #24845, opened 2026-04-21
docs: clarify subuid/subgid mapping in rootless vs userns-remap

### LINKED ISSUE #23664: Add/clarify how subuids and subgids are mapped in rootless docker compared to userns-remap mode
### Is this a docs issue?

### Type of issue

Other

### Description

The documentation page for rootless docker has a couple of paragraphs that discuss user/group ID mapping.

> Rootless mode executes the Docker daemon and containers inside a user namespace. This is similar to [userns-remap mode](https://docs.docker.com/engine/security/userns-remap/), except that with userns-remap mode, the daemon itself is running with root privileges, whereas in rootless mode, both the daemon and the container are running without root privileges.

The linked userns-remap documentation explains how UIDs within the container get mapped to subuids

> This means that testuser is assigned a subordinate user ID range of 231072 and the next 65536 integers in sequence. UID 231072 is mapped within the namespace (within the container, in this case) as UID 0 (root). UID 231073 is mapped as UID 1, and so forth. If a process attempts to escalate privilege outside of the namespace, the process is running as an unprivileged high-number UID on the host, which does not even map to a real user. This means the process has no privileges on the host system at all.

It explains that **UID 0 in the container gets mapped to the first subordinate UID specified in `/etc/subuid`**, but this is only true for `userns-remap`. It is not true for rootless Docker.

In rootless mode, the mapping looks like:

| uid (container) | subuid (host)              | notes                                                                                   |
|-----------------|----------------------------|-----------------------------------------------------------------------------------------|
| 0 (root)        | `$(id -u)` (e.g. `100:john`) | container's root maps to the actual UID of the user running the rootless Docker service |
| 1               | 165536 (`subuid + 0`)        | `/etc/subuid`: `john:165536:100000` -- so 165536 is the first sub uid                   |
| 2               | 165537 (`subuid + 1`)        |                                                                                         |
| n               | `subuid + n - 1`             |                                                                                         |

  
This difference is important when attempting to set file permissions on the host to grant a non-root user in the container permissions to read or write mounted files or folders

The mapping of `0 (root)` to `$UID` (and not the first subuid) is documented in several issues, but it is not easy to locate. The clearest documentation of this behavior is https://github.com/moby/moby/issues/42134#issuecomment-797335148

tagging @AkihiroSuda and @thaJeztah as this relates to a number of open issues regarding permissions and bind mounts in rootless containers.

### Location

https://docs.docker.com/engine/security/rootless/

### Suggestion

Include the above or similar table in the ID mapping section of the rootless Docker docs.

lacatoire:
Fixes #23664.

The rootless docs compared rootless to userns-remap only in terms of daemon privileges, never mentioning that the container-to-host UID mapping also differs. Adds a short bullet list after the existing comparison to spell it out and note the practical impact on bind-mount file ownership.

### DIFF
--- a/content/manuals/engine/security/rootless/_index.md
+++ b/content/manuals/engine/security/rootless/_index.md
@@ -20,6 +20,9 @@ with `userns-remap` mode, the daemon itself is running with root privileges,
 whereas in rootless mode, both the daemon and the container are running without
 root privileges.
 
+The two modes also differ in how they map container UIDs and GIDs to the
+host: see [UID/GID mapping](uid-gid-mapping/) for details.
+
 Rootless mode does not use binaries with `SETUID` bits or file capabilities,
 except `newuidmap` and `newgidmap`, which are needed to allow multiple
 UIDs/GIDs to be used in the user namespace.
--- a/content/manuals/engine/security/rootless/uid-gid-mapping.md
+++ b/content/manuals/engine/security/rootless/uid-gid-mapping.md
@@ -1,0 +1,22 @@
+---
+description: How container UIDs and GIDs are mapped to the host in rootless mode
+keywords: security, namespaces, rootless, uid, gid, subuid, subgid
+title: UID/GID mapping
+weight: 15
+---
+
+Rootless mode and [`userns-remap` mode](../userns-remap.md) map container UIDs
+and GIDs to the host differently.
+
+- In `userns-remap` mode, container UID `0` is mapped to the first subordinate
+  UID listed in `/etc/subuid` for the remap user, and container UID `n` is
+  mapped to `subuid + n`.
+- In rootless mode, container UID `0` is mapped to the host UID of the user
+  running rootless Docker (the result of `id -u`); container UID `n` (for
+  `n >= 1`) is mapped to `subuid + (n - 1)`.
+
+GIDs follow the same rules using `/etc/subgid`.
+
+This difference matters when setting file permissions on bind-mounted
+directories: in rootless mode, files owned by your host user appear as owned
+by `root` inside the container.


dvdksn:
@AkihiroSuda @thaJeztah could you ptal? 🙏

content/manuals/engine/security/rootless/_index.md
```diff
@@ -20,6 +20,21 @@ with `userns-remap` mode, the daemon itself is running with root privileges,
...
+- In rootless mode, container UID `0` is mapped to the host UID of the user
+  running rootless Docker (the result of `id -u`); container UID `n` (for
+  `n >= 1`) is mapped to `subuid + (n - 1)`.
+
+GIDs follow the same rules using `/etc/subgid`.
+
+This difference matters when setting file permissions on bind-mounted
+directories: in rootless mode, files owned by your host user appear as owned
+by `root` inside the container.
+
```
AkihiroSuda, line 25: The changes should be a subpage to avoid inflating the amount of the contents in a single page

AkihiroSuda (changes requested):
https://github.com/docker/docs/pull/24845/changes#r3116127286

content/manuals/engine/security/rootless/_index.md
```diff
@@ -20,14 +20,16 @@ with `userns-remap` mode, the daemon itself is running with root privileges,
...
 root privileges.
 
+The two modes also differ in how they map container UIDs and GIDs to the
+host: see [UID/GID mapping](uid-gid-mapping/) for details.
+
 Rootless mode does not use binaries with `SETUID` bits or file capabilities,
 except `newuidmap` and `newgidmap`, which are needed to allow multiple
 UIDs/GIDs to be used in the user namespace.
 
-
```
AkihiroSuda, line 27: Irrelevant changes

AkihiroSuda: ping @lacatoire

AkihiroSuda (approved):

thaJeztah (approved):
LGTM

@dvdksn ptal

dvdksn (approved):

dvdksn:
Hmm, blocked by commit signature verification. That should not be required here. Looking.

dvdksn:
I guess I was wrong; can you please sign your commit with a GPG/SSH signing key please @lacatoire , thanks

PR #24864, opened 2026-04-23
docs: remove contradictory auto-start note in install guides

### LINKED ISSUE #24861: [docs-scanner] Contradictory information about Docker service auto-start behavior
**Files:** 
- `content/manuals/engine/install/debian.md`
- `content/manuals/engine/install/ubuntu.md`
- `content/manuals/engine/install/raspberry-pi-os.md`

### Issue

Multiple installation guides contain contradictory statements about whether the Docker service starts automatically after installation. The same note appears in Debian, Ubuntu, and Raspberry Pi OS installation guides:

> The Docker service starts automatically after installation. To verify that Docker is running, use:
> 
> ```console
> $ sudo systemctl status docker
> ```
>
> Some systems may have this behavior disabled and will require a manual start:
>
> ```console
> $ sudo systemctl start docker
> ```

This note states both that "The Docker service starts automatically" AND that "Some systems may have this behavior disabled." This is confusing because it doesn't clarify:
- What the default behavior actually is
- Which systems have it disabled
- How users can tell if their system has it disabled

### Why this matters

A user following these instructions doesn't know whether they should expect Docker to be running after installation or not. They're told it "starts automatically" but then immediately told it might not. This creates uncertainty about whether the installation succeeded and what action to take next.

### Suggested fix

Replace the contradictory note with clear, actionable guidance. For example:

> After installation, verify that Docker is running:
> 
> ```console
> $ sudo systemctl status docker
> ```
>
> If Docker is not running, start it manually:
>
> ```console
> $ sudo systemctl start docker
> ```

This removes the contradiction and gives users a clear verification step and recovery action without making confusing claims about what "should" happen.

---

lacatoire:
The `systemctl` note in `debian.md`, `ubuntu.md`, and `raspberry-pi-os.md` claimed the Docker service "starts automatically after installation" and then, in the same block, that "some systems may have this behavior disabled" without telling the reader how to tell which case applies.

Reword the note around a verification step and a recovery step so readers get an actionable check regardless of what the package manager did.

Fixes #24861

### DIFF
--- a/content/manuals/engine/install/debian.md
+++ b/content/manuals/engine/install/debian.md
@@ -178,14 +178,13 @@ Docker from the repository.
 
     > [!NOTE]
     >
-    > The Docker service starts automatically after installation. To verify that
-    > Docker is running, use:
-    > 
+    > After installation, verify that Docker is running:
+    >
     > ```console
     > $ sudo systemctl status docker
     > ```
     >
-    > Some systems may have this behavior disabled and will require a manual start:
+    > If Docker is not running, start it manually:
     >
     > ```console
     > $ sudo systemctl start docker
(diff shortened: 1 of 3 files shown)


dvdksn (approved):
LGTM thanks!

PR #24867, opened 2026-04-23
docs: mention external networks in engine networking overview

### LINKED ISSUE #22498: Mention external networks
### Is this a docs issue?

### Type of issue

### Description

Shouldn't external networks be mentioned here as well?

### Location

https://docs.docker.com/engine/network/

### Suggestion

_No response_

lacatoire:
The engine networking overview did not mention external networks, so readers coming from Compose (where `external: true` is a common pattern) had no entry point to the concept on the Engine side.

Add a short subsection under user-defined networks that defines the term, shows that a pre-existing network can be shared between containers, Compose projects, and Swarm services, and links to the Compose reference.

Fixes #22498

### DIFF
--- a/content/manuals/engine/network/_index.md
+++ b/content/manuals/engine/network/_index.md
@@ -73,6 +73,22 @@ $ docker network create -d bridge my-net
 $ docker run --network=my-net -it busybox
 ```
 
+### External networks
+
+A Docker network persists until you remove it with `docker network rm`. Any container,
+Compose project, or Swarm service can attach to a pre-existing network without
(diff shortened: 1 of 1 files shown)


dvdksn:
The phrase "external network" does not make sense outside of a compose context. In compose, it refers to a docker network that isn't managed by compose. There it makes sense because compose normally creates a network by default, and containers attach to it; it is a managed network. Whereas external networks are ones not managed as part of the compose stack.

For regular containers created with `docker run` or `docker create`, all networks are "external". The container doesn't manage them. Thus this should be documented on the compose side.

PR #24869, opened 2026-04-23
feat(docker-pass): update docs

joe0BAB:
## Reviews

- [x] Technical review
- [ ] Editorial review
- [ ] Product review


dvdksn (commented):
These files are normally generated from source. It doesn't look like they are for the pass plugin (?)

joe0BAB:
@dvdksn it's the other way around. They were manually created before but are auto-generated now.

dvdksn (approved):

PR #24908, opened 2026-04-27
fix: correct chown argument order in Compose lifecycle hooks example

### LINKED ISSUE #24853: Fix examples for Compose lifecycle hooks
### Is this a docs issue?

### Type of issue

Information is incorrect

### Description

The doc describing the lifecycle provides completely irrelevant examples.

The `post-start` hook cannot be used to fix a volume permissions because there is no guarantee that it will finish before a containerized application starts.

The `pre-stop` hook example offers to use it for flushing caches. Taking into account that "these hooks won't run if the container stops by itself or gets killed suddenly", it doesn't make sense to flush caches in the hook because: 1) if a "normal" container stop is ok, then caches are already flushed by a containerized application; 2) if the container gets killed, then the hook will be skipped anyway.

These inaccuracies make the lifecycle feature look like doubtful...

Thank you!

### Location

https://docs.docker.com/compose/how-tos/lifecycle/

### Suggestion

_No response_

mohithshuka:
## What does this PR do?

Fixes an incorrect `chown` command in the `post_start` lifecycle hooks example.

The arguments were in the wrong order:
- ❌ Before: `chown -R /data 1001:1001`
- ✅ After: `chown -R 1001:1001 /data`

The correct `chown` syntax is: `chown [options] [owner:group] [path]`

## Related issue

Fixes #24853


aevesdocker:
@mohithshuka please fix the PR to include only the Compose content.

aevesdocker:
Closing this PR as it contains edited pages not related to the PR description. Please open separate PRs for each change. Thank you

PR #24913, opened 2026-04-28
docs: make VEX prerequisites method-specific

Rohan5commit:
## Description

- make the VEX prerequisites method-specific instead of implying that every workflow requires the containerd image store and registry write access
- reinforce those attestation-only requirements in the attestation section

## Related issues or tickets

- Addresses #24900

## Guideline alignment

- followed `docs/CONTRIBUTING.md`
- kept the change to one documentation file and one focused issue

## Validation

- not run (`docker buildx bake validate`) because this is a docs-only clarification

## Reviews

- [x] Technical review
- [ ] Editorial review
- [ ] Product review

### DIFF
--- a/content/manuals/scout/how-tos/create-exceptions-vex.md
+++ b/content/manuals/scout/how-tos/create-exceptions-vex.md
@@ -24,8 +24,13 @@ To create exceptions using OpenVEX documents, you need:
 
 - The latest version of Docker Desktop or the Docker Scout CLI plugin
 - The [`vexctl`](https://github.com/openvex/vexctl) command line tool.
-- The [containerd image store](/manuals/desktop/features/containerd.md) must be enabled
+
+Additional requirements depend on how you attach the VEX document:
+
+- The [containerd image store](/manuals/desktop/features/containerd.md)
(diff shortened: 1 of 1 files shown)


craig-osterhout (changes requested):
Thank you. Ran through some examples to test this.

My finding is containerd appears to not be needed for `docker scout attestation add` because it pushes directly to the registry and never touches the local image store.

It is needed for the `--provenance=true --sbom=true` in the build command, although `--push` doesn't store the image locally so it still works without containerd. If you remove `--provenance=true --sbom=true` from the build command, we can remove the containerd requirement there too.

One place containerd is genuinely needed: pulling an image locally that has a VEX attestation. Without containerd, the classic image store can't read the OCI image index and the attestation isn't visible. You can work around this using the `registry://` prefix with `docker scout` commands to read directly from the registry instead.

Suggested changes:

1. Remove the containerd prerequisite, or narrow it to "required if pulling and inspecting attestations locally"
2. Remove `--provenance=true --sbom=true` from the build command (unrelated to VEX), which eliminates the main justification for the containerd requirement

craig-osterhout:
superseded by #25189

PR #24929, opened 2026-04-29
Update Docker Desktop log paths from dockerd.log to init.log

vvoland:
Docker Desktop no longer writes separate dockerd.log and containerd.log files. All VM service logs are multiplexed into a single init.log file.

The init.log file uses JSON-formatted lines with a "component" field, not the old `<timestamp> <component> <message>` format.


dvdksn (commented):
in what version did this change

vvoland:
No idea tbh! I tried to look into the DD logs and I noticed that last messages in my `dockerd.log` were from `2025/09/05 20:23:45` so I started digging and found the path changed...

dvdksn (approved):

PR #24951, opened 2026-05-04
docs: clarify both PayloadUUID values must be replaced with unique UUIDs

### LINKED ISSUE #24944: [docs-scanner] Ambiguous UUID replacement instructions in Mac configuration profile
**File:** `content/manuals/enterprise/security/enforce-sign-in/methods.md`

### Issue

The Mac configuration profile example contains two `PayloadUUID` fields with different values, but the replacement instructions are ambiguous about how to handle them.

The example shows:
```xml
<key>PayloadUUID</key>
<string>eed295b0-a650-40b0-9dda-90efb12be3c7</string>
...
<key>PayloadUUID</key>
<string>0deedb64-7dc9-46e5-b6bf-69d64a9561ce</string>
```

The instructions state:
> Replace `PayloadUUID` with a randomly generated UUID

This is ambiguous because:
- There are two PayloadUUID fields (one in the inner payload, one in the outer configuration)
- The instruction uses singular "UUID" but there are two UUIDs in the example
- It's unclear whether both should be replaced or just one
- It's unclear whether they should be different UUIDs (as shown in the example) or the same

### Why this matters

Administrators following these instructions might:
- Replace only one UUID and leave the other as the example value
- Use the same UUID for both fields when they should be different
- Be unsure whether the configuration is correct

This could lead to configuration profile deployment failures or conflicts if multiple administrators use the same example UUIDs.

### Suggested fix

Clarify the instructions to explicitly address both UUIDs:

> Replace placeholders:
> - Change `com.yourcompany.docker.config` to your company identifier
> - Replace `Your Company Name` with your organization name
> - Replace both `PayloadUUID` values with randomly generated UUIDs (each should be unique)
> - Update the `allowedOrgs` value with your organization names (separated by semicolons)
> - Replace `company.proxy:port` with http/https proxy server host (or IP address) and port

---

lacatoire:
The Mac configuration profile example contains two `PayloadUUID` fields (inner payload and outer configuration) with different values, but the replacement instruction was singular and ambiguous about whether one or both should be replaced. Updating to make explicit that both must be replaced with distinct UUIDs.

Closes #24944

### DIFF
--- a/content/manuals/enterprise/security/enforce-sign-in/methods.md
+++ b/content/manuals/enterprise/security/enforce-sign-in/methods.md
@@ -134,7 +134,7 @@ Overriding at least one of the proxy settings via Configuration profiles will au
 1. Replace placeholders:
    - Change `com.yourcompany.docker.config` to your company identifier
    - Replace `Your Company Name` with your organization name
-   - Replace `PayloadUUID` with a randomly generated UUID
+   - Replace both `PayloadUUID` values with randomly generated UUIDs (each value must be unique)
    - Update the `allowedOrgs` value with your organization names (separated by semicolons)
    - Replace `company.proxy:port` with http/https proxy server host(or IP address) and port
 1. Deploy the profile using your MDM solution.


aevesdocker:
Closing as the docs are clear as they are. The docs scanner issue was too pedantic

PR #24961, opened 2026-05-05
Gro 109

aevesdocker:
## Description

Compose networking refresh

## Reviews

- [ ] Technical review
- [ ] Editorial review
- [ ] Product review

### DIFF
--- a/content/manuals/compose/how-tos/networking.md
+++ b/content/manuals/compose/how-tos/networking.md
@@ -30,7 +30,7 @@ services:
       - "8001:5432"
 ```
 
-Since `networks` is empty or absent from the Compose file, Compose considers an implicit definition for the service to be connected to the default network.
+Since `networks` is empty or absent from the Compose file, Compose automatically connects all services to the default network.
 
 When you run `docker compose up`, the following happens:
 
@@ -46,7 +46,7 @@ Each container can now look up the service name `web` or `db` and get back the a
 
 Your app's network is given a name based on the "project name", which is based on the name of the directory it lives in. You can override the project name with either the [`--project-name` flag](/reference/cli/docker/compose/) or the [`COMPOSE_PROJECT_NAME` environment variable](environment-variables/envvars.md#compose_project_name).
 
-It is important to note the distinction between `HOST_PORT` and `CONTAINER_PORT`. In the example above, for `db`, the `HOST_PORT` is `8001` and the container port is `5432` (the Postgres default). Networked service-to-service communication uses the `CONTAINER_PORT`. The host port is only used when accessing the service from outside the network.
+The `HOST_PORT` and `CONTAINER_PORT` serve different purposes. In the example above, for `db`, the `HOST_PORT` is `8001` and the container port is `5432` (the Postgres default). Networked service-to-service communication uses the `CONTAINER_PORT`. The host port is only used when accessing the service from outside the network.
 
 ### Updating containers on the network
 
@@ -54,58 +54,18 @@ If you make a configuration change to a service and run `docker compose up` to u
 
 If any containers have connections open to the old container, they are closed. It is each container's responsibility to detect this condition, look up the name again, and reconnect.
 
-## Network drivers: bridge vs host
+## Change the network mode
 
-Compose supports multiple network drivers. The two most common are `bridge` and `host`.
+By default, each service joins the project's bridge network. It is the most secure networking mode. If you don't specify [`network_mode`](/reference/compose-file/services.md#network_mode), this is the type of network you are creating.
 
-### Bridge (default)
+You can override the networking mode on a per-service basis. `Network_mode` option accepts the following values:
 
-Bridge is the default driver and the most secure option. If you don't specify [`network_mode`](/reference/compose-file/services.md#network_mode), this is the type of network you are creating. Each container gets its own network namespace, and containers communicate with each other using service names:
+- `host`: The container shares the host's network stack. No port mapping is needed or supported, and service name DNS resolution does not work. Use for system-level tools like network monitors that require direct access to host interfaces. A container using `network_mode: host` can access all host ports and observe all network traffic on the host. Use it only when genuinely required.
+- `none`: Turns off all container networking.
+- `service:{name}`: Gives the container access to the specified container by referring to its service name.
+- `container:{name}`: Gives the container access to the specified container by referring to its container ID.
 
-```yaml
-services:
-  web:
-    image: nginx
-    ports:
-      - "8080:80"
-  db:
-    image: postgres:latest
-```
-
-Port mapping is required to expose a service outside the network. Inside the network, containers use service names (`web`, `db`); outside the network, you use the mapped host port.
-
-For more information, see [Bridge network driver](/manuals/engine/network/drivers/bridge.md).
-
-### Host
-
-With [`network_mode: host`](/reference/compose-file/services.md#network_mode), the container shares the host's network stack entirely. There is no network isolation:
-
-```yaml
-services:
-  monitoring:
-    image: prometheus/node-exporter
-    network_mode: host   # No port mapping needed. Uses host ports directly
-```
-
-The container can access all host network interfaces directly, which means no port mapping is required, but also that there is no container-level network isolation.
-
-> [!WARNING]
-> 
-> `host` mode carries security implications. A container using `network_mode: host` can access all host ports and observe all network traffic on the host. Use it only when genuinely required.
-
-| Feature | Bridge | Host |
-|---|---|---|
-| Port mapping | Required | Not needed |
-| Network isolation | Yes | No |
-| Container DNS | Service names work | Use `localhost` or IPs |
-| Performance | Small overhead | Native speed |
-| Security | Better isolation | Less isolation |
-
-Use bridge for application services, databases, and web servers; anything where you want controlled, isolated communication.
-
-Use host for system-level tools that need direct access to the host network, such as network monitors or performance-critical workloads like game servers.
-
-You can mix both modes in a single project:
+You can mix modes in a single project:
 
 ```yaml
 services:
@@ -123,16 +83,6 @@ services:
 networks:
   isolated:
     driver: bridge
-```
-
-With this example, you can test network isolation by running the following:
-
-```bash
-# Bridge mode: can't access host services directly
-$ docker compose exec app curl localhost:5432  # Fails
-
-# Host mode: full access
-$ docker compose exec monitoring curl localhost:5432  # Works
 ```
 
 ## Specify custom networks
@@ -284,7 +234,7 @@ Services on the same external network can reach each other by service name, just
 
 > [!IMPORTANT]
 >
-> The external network must exist before you run `docker compose up`. If it doesn't, Compose fails with a **network not found** error. Always create it first with `docker network create`.
+> The external network must exist before you run `docker compose up`. If it doesn't, Compose fails with a `Network not found` error. Always create it first with `docker network create`.
 
 ## Hybrid networking
 
--- a/content/manuals/compose/how-tos/networking.md
+++ b/content/manuals/compose/how-tos/networking.md
@@ -8,11 +8,11 @@ aliases:
   - /compose/networking/
 ---
 
-Compose handles networking for you by default, but gives you fine-grained control when you need it. This page explains how the default network works and how containers discover each other by name. It also covers when and how to define custom networks, choose between network drivers, connect services across separate Compose projects, map custom hostnames, and debug connectivity issues.
+Compose handles networking for you by default, but gives you fine-grained control when you need it. This page explains how the default network works and how containers discover each other by name. It also covers when and how to define custom networks, connect services across separate Compose projects, map custom hostnames, and debug connectivity issues.
 
 ## Default network and service discovery
 
-By default, Compose sets up a single [network](/reference/cli/docker/network/create/) for your app. Each container for a service joins the default network and is both reachable by other containers on that network, and discoverable by its service name. This network uses the `bridge` driver. To understand when you'd use a different driver, see [Network drivers: bridge vs host](#network-drivers-bridge-vs-host).
+By default, Compose sets up a single [network](/reference/cli/docker/network/create/) for your app. Each container for a service joins the default network and is both reachable by other containers on that network, and discoverable by its service name. This network uses the `bridge` driver. To understand when you'd use a different driver, see [Network drivers: bridge vs host](#change-the-network-mode).
 
 For most development setups, the default network is sufficient. When you run `docker compose up`, Compose creates a network named `<project-name>_default` and attaches all services to it. Each service registers its name with an internal DNS server, so containers can reach each other using the service name directly. No IP addresses or manual configuration is needed.
 
@@ -30,7 +30,7 @@ services:
       - "8001:5432"
 ```
 
-Since `networks` is empty or absent from the Compose file, Compose automatically connects all services to the default network.
+Compose automatically connects all services to the default network, so you don't need to define `networks` explicitly in the Compose file.
 
 When you run `docker compose up`, the following happens:
 
@@ -42,9 +42,9 @@ Each container can now look up the service name `web` or `db` and get back the a
 
 > [!TIP]
 >
-> Always reference services by name, not IP address. When containers are recreated, for example after a configuration change, they receive a new IP address. The service name stays stable.
+> Docker assigns container IP addresses dynamically from the network's subnet each time a container starts so they are not persisted across restarts or recreations. This means you should always reference services by name, not IP address. When containers are recreated, for example after a configuration change, they receive a new IP address. The service name stays stable.
 
-Your app's network is given a name based on the "project name", which is based on the name of the directory it lives in. You can override the project name with either the [`--project-name` flag](/reference/cli/docker/compose/) or the [`COMPOSE_PROJECT_NAME` environment variable](environment-variables/envvars.md#compose_project_name).
+Your app's network is given a name based on the "project name", which is taken from the name of the directory it lives in. You can override the project name with either the [`--project-name` flag](/reference/cli/docker/compose/) or the [`COMPOSE_PROJECT_NAME` environment variable](environment-variables/envvars.md#compose_project_name).
 
 The `HOST_PORT` and `CONTAINER_PORT` serve different purposes. In the example above, for `db`, the `HOST_PORT` is `8001` and the container port is `5432` (the Postgres default). Networked service-to-service communication uses the `CONTAINER_PORT`. The host port is only used when accessing the service from outside the network.
 
@@ -58,7 +58,7 @@ If any containers have connections open to the old container, they are closed. I
 
 By default, each service joins the project's bridge network. It is the most secure networking mode. If you don't specify [`network_mode`](/reference/compose-file/services.md#network_mode), this is the type of network you are creating.
 
-You can override the networking mode on a per-service basis. `Network_mode` option accepts the following values:
+You can override the networking mode on a per-service basis. The `network_mode` option accepts the following values:
 
 - `host`: The container shares the host's network stack. No port mapping is needed or supported, and service name DNS resolution does not work. Use for system-level tools like network monitors that require direct access to host interfaces. A container using `network_mode: host` can access all host ports and observe all network traffic on the host. Use it only when genuinely required.
 - `none`: Turns off all container networking.
@@ -140,18 +140,18 @@ services:
   cache:
     image: redis
     networks:
-      - internal
+      - isolated
 
   worker:
     image: myworker
     networks:
-      - internal
+      - isolated
       - public
 
 networks:
-  internal:
+  isolated:
     internal: true   # No external connectivity
-  public: # Regular network connected to host
+  public:   # Standard bridge network, created by Compose on docker compose up
 ```
 
 Note that a service connected to both an internal and a non-internal network (like `worker` above) can still reach the internet via the non-internal network `public`.
@@ -194,7 +194,7 @@ Instead of creating `<project-name>_default`, Compose looks for a network called
 External networks are particularly useful when services in separate Compose projects need to communicate. Create a shared network once, then reference it as external in each project:
 
 ```bash
-docker network create shared-network
+docker network create inter-project
 ```
 
 backend-compose.yaml:
@@ -210,7 +210,7 @@ services:
 networks:
   shared:
     external: true
-    name: shared-network
+    name: inter-project
 ```
 
 frontend-compose.yaml:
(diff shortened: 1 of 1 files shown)


content/manuals/compose/how-tos/networking.md
```diff
@@ -29,70 +25,115 @@ services:
...
-If any containers have connections open to the old container, they are closed. It is a container's responsibility to detect this condition, look up the name again and reconnect.
+If any containers have connections open to the old container, they are closed. It is each container's responsibility to detect this condition, look up the name again, and reconnect.
 
-> [!TIP]
->
-> Reference containers by name, not IP, whenever possible. Otherwise you’ll need to constantly update the IP address you use.
+## Network drivers: bridge vs host
 
-## Link containers
+Compose supports multiple network drivers. The two most common are `bridge` and `host`.
```
thaJeztah: I'm not sure if `host` is most common, and not sure if we should encourage users to use it; using `host` networking reduces the container's sandbox, and usually is reserved for expert use-cases where containers should not have a security boundary for networking.

Using `host` networking means that the container also won't have its own IP-address, and resolving containers won't be possible (as they share the host's IP-address.

content/manuals/compose/how-tos/networking.md
```diff
@@ -29,70 +25,115 @@ services:
     ports:
       - "8000:8000"
   db:
-    image: postgres:18
+    image: postgres:latest
```
thaJeztah, line 28: Curious why this was switched to `:latest` (it's less maintenance, but not usually best-practice to use `:latest` - even more so for database containers, as updates to (in this case PostgreSQL) usually involves a migration step.

content/manuals/compose/how-tos/networking.md
```diff
@@ -8,17 +8,13 @@ aliases:
   - /compose/networking/
 ---
 
-By default Compose sets up a single
-[network](/reference/cli/docker/network/create/) for your app. Each
-container for a service joins the default network and is both reachable by
-other containers on that network, and discoverable by the service's name.
+Compose handles networking for you by default, but gives you fine-grained control when you need it. This page explains how the default network works and how containers discover each other by name. It also covers when and how to define custom networks, choose between network drivers, connect services across separate Compose projects, map custom hostnames, and debug connectivity issues.
```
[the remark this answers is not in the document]
aevesdocker: since when has this been an issue. this is completely fine

content/manuals/compose/how-tos/networking.md
```diff
@@ -29,70 +25,115 @@ services:
...
+
+networks:
+  isolated:
+    driver: bridge
+```
+
+With this example, you can test network isolation by running the following:
+
+```bash
+# Bridge mode: can't access host services directly
```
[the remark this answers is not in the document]
aevesdocker: removed section as content changed

content/manuals/compose/how-tos/networking.md
aevesdocker: Yes this was a maintenance-related change. Also a consistency one - we do this quite a lot throughout the rest of docs. I personally think this is Ok for these non-normative examples

content/manuals/compose/how-tos/networking.md
aevesdocker: I was keen to demonstrate `network_mode` here and took `bridge` and `host` as my examples. You're right to pick up on `the most common`. I have removed. I've re-framed the rest of the content too to focus on `network_mode` instead of comparing the drivers.

thaJeztah: Ah, yeah, it's a tricky one; the "host" mode networking is really an expert use-case as it has many differences; unlike picking a different network-driver, host-mode means we completely disable the network namespace / network isolation of the container. It's useful for very specific situations, but for users not familiar with it (or aware of the wider consequences), it will be confusing.

Trying to think if we can find a different example, which could be a different network-driver, or if we want to keep it really basic `none` as networking (i.e., disable networking) to prevent going down the deep-end to explain the full network stack; https://docs.docker.com/engine/network/drivers/

We do have other driver (e.g. macvlan or ipvlan), but with those there's definitely a risk of going too much into depth on networking in general.

Perhaps @robmry has some ideas on some illustrative examples.

aevesdocker: Yeh I think I'm leaning towards keeping it lighter with my new changes. We don't get a huge amount (if any) feedback on the lack of depth of the Compose network docs because we point to the Engine docs which are very comprehensive. I view this page as more of an entry point into networking

thaJeztah: Yes, perhaps we should keep it simple for now, and indeed refer to the other docs for those that really need it. Of course we can revisit the if we come with a basic example to illustrate, but it may save us some headaches if users start to follow the example, then get back with questions.

content/manuals/compose/how-tos/networking.md
thaJeztah: Yeah, it's a tricky balance; we should have a closer look at some of these; for the really "one-off" examples (just a `docker run alpine` / `ubuntu` / `busybox`), it's usually fine.

For other, more complete examples that users could use as a "template" / starting point for what they're doing, we should  be more careful, and better to follow best practices.

I know we once had WordPress examples using `mysql:latest` and we had a _ton_ of users that saw their stack break when MySQL moved `:latest` from v5 to something newer and their databases broke.

From a maintenance perspective for this specific case; postgres does a major version once per Year, and those are supported for 5 Years, so even if it's the previous major version, we would still have a good time window to update.

But we should look at either having some variables; we already have some for various versions (engine, alpine, etc); and we could probably automate some of these updates (or have Gordon do it for us); https://github.com/docker/docs/blob/1414b91b7168a58dcb4274cc2d89be68c4d94c9a/hugo.yaml#L159-L184

content/manuals/compose/how-tos/networking.md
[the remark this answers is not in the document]
akristen: LOL

akristen: this is a great intro tbh

content/manuals/compose/how-tos/networking.md
```diff
@@ -29,70 +25,65 @@ services:
     ports:
       - "8000:8000"
   db:
-    image: postgres:18
+    image: postgres:latest
     ports:
       - "8001:5432"
 ```
 
+Since `networks` is empty or absent from the Compose file, Compose automatically connects all services to the default network.
```
akristen: Can we reword "is empty or absent from the Compose file"? I can see this introducing ambiguity. Maybe something like, 'Compose automatically connects all services to the default network, so you don't need to define `networks` explicitly in the Compose file.'

content/manuals/compose/how-tos/networking.md
```diff
@@ -29,70 +25,65 @@ services:
...
-1.  A network called `myapp_default` is created.
-2.  A container is created using `web`'s configuration. It joins the network
-    `myapp_default` under the name `web`.
-3.  A container is created using `db`'s configuration. It joins the network
-    `myapp_default` under the name `db`.
+1. A network called `myapp_default` is created.
+2. A container is created using `web`'s configuration. It joins `myapp_default` under the name `web`.
+3. A container is created using `db`'s configuration. It joins `myapp_default` under the name `db`.
+
+Each container can now look up the service name `web` or `db` and get back the appropriate container's IP address. The `web` service can connect to the database at `postgres://db:5432`. From the host machine, the same database is accessible at `postgres://localhost:8001` if your container is running locally.
```
akristen, line 41: this is really clear and good

content/manuals/compose/how-tos/networking.md
```diff
@@ -29,70 +25,65 @@ services:
...
 
-Each container can now look up the service name `web` or `db` and
-get back the appropriate container's IP address. For example, `web`'s
-application code could connect to the URL `postgres://db:5432` and start
-using the Postgres database.
+> [!TIP]
+>
+> Always reference services by name, not IP address. When containers are recreated, for example after a configuration change, they receive a new IP address. The service name stays stable.
```
akristen: So I understand this and it's clear why I'd use the service name, but I'm not understanding the _why_, for some reason. Is this an aspect of how docker handles networks? if i'm using network: host, would the IP address stay stable or does this rule still apply? (asking in case i'm missing something, or maybe it needs to be clearer in the content before the tip)

content/manuals/compose/how-tos/networking.md
```diff
@@ -29,70 +25,65 @@ services:
...
 
-It is important to note the distinction between `HOST_PORT` and `CONTAINER_PORT`.
-In the above example, for `db`, the `HOST_PORT` is `8001` and the container port is
-`5432` (postgres default). Networked service-to-service
-communication uses the `CONTAINER_PORT`. When `HOST_PORT` is defined,
-the service is accessible outside the swarm as well.
+Your app's network is given a name based on the "project name", which is based on the name of the directory it lives in. You can override the project name with either the [`--project-name` flag](/reference/cli/docker/compose/) or the [`COMPOSE_PROJECT_NAME` environment variable](environment-variables/envvars.md#compose_project_name).
```
akristen: "Your app's network is given a name based on the "project name", which is based on the name of the directory it lives in."

Can we reword? Using 'based' twice is confusing. Do I read th is as the network is based on the project name, and the project name is based on the directory it lives in?

content/manuals/compose/how-tos/networking.md
```diff
@@ -29,70 +25,65 @@ services:
...
 
-Within the `web` container, your connection string to `db` would look like
-`postgres://db:5432`, and from the host machine, the connection string would
-look like `postgres://{DOCKER_IP}:8001` for example `postgres://localhost:8001` if your container is running locally.
+The `HOST_PORT` and `CONTAINER_PORT` serve different purposes. In the example above, for `db`, the `HOST_PORT` is `8001` and the container port is `5432` (the Postgres default). Networked service-to-service communication uses the `CONTAINER_PORT`. The host port is only used when accessing the service from outside the network.
```
akristen, line 49: ohhhhhhh. Is there a benefit to using two different ports? When I've done this, I usually use the same port for each (which I based off the sample compose file in registry.) this is mostly because i'm curious and maybe could be useful for other users

content/manuals/compose/how-tos/networking.md
```diff
@@ -29,70 +25,65 @@ services:
...
-```yaml
-services:
-  web:
-    build: .
-    links:
-      - "db:database"
-  db:
-    image: postgres:18
-```
+You can override the networking mode on a per-service basis. `Network_mode` option accepts the following values:
```
akristen:
```suggestion
You can override the networking mode on a per-service basis. `network_mode` option accepts the following values:
```

is yaml case sensitive? not sure what style guide suggests for variables leading in sentences

content/manuals/compose/how-tos/networking.md
```diff
@@ -142,7 +131,32 @@ networks:
...
+### Internal networks
+
+Setting `internal: true` on a network creates it without a connection to the host's network interfaces. It has no default gateway for external connectivity. This is useful for services like databases that should be completely unreachable from outside the container network:
+
+```yaml
+services:
+  cache:
+    image: redis
+    networks:
+      - internal
```
akristen: The above paragraph defines ` networks: - internal` without the `true` value, but in the example it's just `internal`. is true implied by virtue of defining networks as internal?

content/manuals/compose/how-tos/networking.md
```diff
@@ -142,7 +131,32 @@ networks:
...
+      - public
+
+networks:
+  internal:
+    internal: true   # No external connectivity
+  public: # Regular network connected to host
+```
+
+Note that a service connected to both an internal and a non-internal network (like `worker` above) can still reach the internet via the non-internal network `public`.
```
akristen, line 157: Do users need to define or configure `public` anywhere, or is it an "out of the box" feature (so to speak) that comes with default network

content/manuals/compose/how-tos/networking.md
```diff
@@ -176,7 +187,195 @@ networks:
...
+```
+
+backend-compose.yaml:
+
+```yaml
+services:
+  api:
+    image: myapi:latest
+    networks:
+      - shared
```
akristen, line 207: should it be `shared-network` instead of `shared`?

content/manuals/compose/how-tos/networking.md
```diff
@@ -176,7 +187,195 @@ networks:
...
+
+```yaml
+services:
+  app:
+    image: myapp
+    extra_hosts:
+      - "host.docker.internal:host-gateway"
+```
+
+On Linux, `host-gateway` resolves to the host's IP on the default bridge network. On Mac and Windows, Docker automatically provides this - `host-gateway` resolves to the same internal IP address as `host.docker.internal`.
```
akristen:
```suggestion
On Linux, `host-gateway` resolves to the host's IP on the default bridge network. On Mac and Windows, Docker automatically resolves `host-gateway` to the same internal IP address as `host.docker.internal`.
```

maybe?

content/manuals/compose/how-tos/networking.md
```diff
@@ -176,7 +187,195 @@ networks:
...
+
+## Multi-host networking
+
+When deploying a Compose application on a Docker Engine with [Swarm mode enabled](/manuals/engine/swarm/_index.md), you can use the built-in `overlay` driver to enable multi-host communication. Overlay networks are always created as `attachable`. You can optionally set the [`attachable`](/reference/compose-file/networks.md#attachable) property to `false`.
+
+To learn more, see the [overlay network driver documentation](/manuals/engine/network/drivers/overlay.md).
+
+## Link containers
+
+Links allow you to define extra aliases by which a service is reachable from another service. They are not required for basic service-to-service communication — by default, any service can reach any other service at that service's name. In the following example, `db` is reachable from `web` at both the hostnames `db` and `database`:
```
akristen:
```suggestion
Links allow you to define extra aliases by which a service is reachable from another service. They are not required for basic service-to-service communication. By default, any service can reach any other service at that service's name. In the following example, `web` can reach `db` at both the hostnames `db` and `database`:
```

by all means keep the emdash lol but i think it can be separated into two distinct sentences

akristen (approved):
Made some small suggestions that are non-blocking and perhaps even obvious answers to our users. :) Overall I feel confident that I could do more advanced networking with what you have here. My only other thought:

Is it worth describing that you can test your network mode with the `docker exec` command? Maybe in a what's next section? I've found this helpful when I want to test that I've configured networks correctly.

content/manuals/compose/how-tos/networking.md
akristen: I see you use `shared` later on. What's the difference between the two, if any. lol maybe i'm missing something...

content/manuals/compose/how-tos/networking.md
aevesdocker: oo yes, subtle but makes a difference. done.

content/manuals/compose/how-tos/networking.md
aevesdocker: Yeh this is a Docker thing, not a Compose-specific thing. You're right though that the tip just jumps straight in without setting that up and if you're new to this topic you'll want this info, so have added the line 'Docker assigns container IP addresses dynamically from the network's subnet each time a container starts; they are not persisted across restarts or recreations.'

content/manuals/compose/how-tos/networking.md
aevesdocker: oui yes, good point. I went with 'Your app's network is given a name based on the "project name", which is taken from the name of the directory it lives in.'

content/manuals/compose/how-tos/networking.md
aevesdocker: Using the same is fine as long as nothing else on your host is using it. The main reason to use different ports is to avoid conflicts, for example if you're running multiple projects locally. Again, this is a general Docker thing so not sure whether it needs to be expanded further

content/manuals/compose/how-tos/networking.md
aevesdocker: good spot!

content/manuals/compose/how-tos/networking.md
aevesdocker: ah so this is coz I've given it confusing names. So am renaming 'internal' to 'isolated' then the key value 'internal: true' will be clearer

content/manuals/compose/how-tos/networking.md
aevesdocker: 🔍  So it tells Compose to create it as a standard bridge network. It's not the same as the default network Compose creates automatically, it's just an explicitly named custom network that just happens to have default settings. Have updated the comment to make this clearer!

content/manuals/compose/how-tos/networking.md
aevesdocker: yeh this is weird compose file parsing rules.

- `shared` is the Compose network alias that is used to reference the network within that Compose file
- `shared-network` is the actual Docker network name which gets created by `docker network create` and what Docker itself knows it as.

Having them be similar in the example isn't helpful though so have changed!

content/manuals/compose/how-tos/networking.md
aevesdocker: agreed!

aevesdocker:
>Is it worth describing that you can test your network mode with the docker exec command? Maybe in a what's next section? I've found this helpful when I want to test that I've configured networks correctly.

Yes! (Although it'll be `docker compose exec` ;)). I do have `docker network inspect`, but it achieves a separate thing, so have added.

content/manuals/compose/how-tos/networking.md
```diff
@@ -8,17 +8,13 @@ aliases:
   - /compose/networking/
 ---
 
-By default Compose sets up a single
-[network](/reference/cli/docker/network/create/) for your app. Each
-container for a service joins the default network and is both reachable by
-other containers on that network, and discoverable by the service's name.
+Compose handles networking for you by default, but gives you fine-grained control when you need it. This page explains how the default network works and how containers discover each other by name. It also covers when and how to define custom networks, connect services across separate Compose projects, map custom hostnames, and debug connectivity issues.
```
docker-agent, line 11: **[MEDIUM] Hedge language: "fine-grained control"**

"Fine-grained" is a hedge word that the Docker style guide explicitly flags as language to avoid (along with "simply", "easily", "seamlessly"). It implies a quality rather than describing what the product does.

**Suggestion:** Replace with a factual description, for example:
> "Compose sets up networking automatically, but lets you configure custom networks, drivers, and DNS when you need it."

content/manuals/compose/how-tos/networking.md
```diff
@@ -29,70 +25,65 @@ services:
...
-If any containers have connections open to the old container, they are closed. It is a container's responsibility to detect this condition, look up the name again and reconnect.
+If any containers have connections open to the old container, they are closed. It is each container's responsibility to detect this condition, look up the name again, and reconnect.
 
-> [!TIP]
->
-> Reference containers by name, not IP, whenever possible. Otherwise you’ll need to constantly update the IP address you use.
+## Change the network mode
 
-## Link containers
+By default, each service joins the project's bridge network. It is the most secure networking mode. If you don't specify [`network_mode`](/reference/compose-file/services.md#network_mode), this is the type of network you are creating.
```
docker-agent, line 59: **[MEDIUM] Unqualified superlative: "the most secure networking mode"**

"It is the most secure networking mode" is a marketing-style claim that is also factually questionable. The `none` mode disables all networking entirely, which is arguably more restrictive than bridge. The Docker style guide says to describe what something does rather than making superlative statements.

**Suggestion:** Remove the superlative and describe behaviour instead:
> "By default, each service joins the project's bridge network, which isolates containers from external networks while allowing inter-service communication."

content/manuals/compose/how-tos/networking.md
```diff
@@ -142,7 +131,32 @@ networks:
...
+      - isolated
+      - public
+
+networks:
+  isolated:
+    internal: true   # No external connectivity
+  public:   # Standard bridge network, created by Compose on docker compose up
+```
+
+Note that a service connected to both an internal and a non-internal network (like `worker` above) can still reach the internet via the non-internal network `public`.
```
docker-agent, line 157: **[MEDIUM] Meta-commentary: "Note that"**

The Docker style guide explicitly lists "Note that..." as forbidden meta-commentary that adds no value and should be replaced by stating the information directly (alongside "It's worth noting that...", "Keep in mind that...").

**Suggestion:** Remove the opener and state the information directly:
> "A service connected to both an internal and a non-internal network, like `worker` above, can still reach the internet via `public`."

content/manuals/compose/how-tos/networking.md
```diff
@@ -176,7 +187,201 @@ networks:
...
+
+```yaml
+services:
+  app:
+    image: myapp
+    extra_hosts:
+      - "host.docker.internal:host-gateway"
+```
+
+On Linux, `host-gateway` resolves to the host's IP on the default bridge network. On Mac and Windows, Docker automatically provides this, `host-gateway` resolves to the same internal IP address as `host.docker.internal`.
```
docker-agent, line 286: **[MEDIUM] "Mac" should be "macOS"; comma splice**

Two issues on this line:

1. **Terminology**: Docker documentation uses "macOS" (not "Mac") when referring to Apple's operating system, consistent with Apple's own branding and Docker's terminology conventions.

2. **Comma splice**: "Docker automatically provides this, `host-gateway` resolves..." joins two independent clauses with only a comma. This should be two sentences.

**Suggestion:**
> "On Linux, `host-gateway` resolves to the host's IP on the default bridge network. On macOS and Windows, Docker automatically provides this mapping and `host-gateway` resolves to the same internal IP address as `host.docker.internal`."

content/manuals/compose/how-tos/networking.md
docker-agent, line 11: **[MEDIUM] Scope inflation: focused how-to expanded into comprehensive reference**

The Docker style guide ("Scope preservation") explicitly cautions against expanding a focused document unnecessarily: *"If the document is brief and direct (90 lines), keep it that way. Don't transform a focused guide into an exhaustive tutorial."* and *"Bad additions change the document's character."*

This PR adds roughly 200 lines of new sections to what was a focused how-to:
- Change the network mode
- Internal networks
- Hybrid networking
- Custom DNS with `extra_hosts`
- Debugging (port mapping, network membership, connectivity)
- Connecting multiple Compose projects

Many of these topics (hybrid networking, `extra_hosts` deep-dives, the full debugging workflow) are self-contained enough to live as separate pages or in reference documentation, which would preserve this page's focused character while still making the content available.

docker-agent (commented):

5 medium-severity documentation style issues found in the added lines. No critical or high-severity problems. The page is a significant expansion of the original focused how-to.

PR #24995, opened 2026-05-07
Fix typo "--excludes" in COPY --exclude documentation

### LINKED ISSUE #24993: Typo "--excludes"
### Is this a docs issue?

### Type of issue

Other

### Description

I think its supposed to be "--exclude" and not "--excludes"

> You can specify the --exclude option multiple times for a COPY instruction. Multiple **--excludes** ...

### Location

https://docs.docker.com/reference/dockerfile#copy---exclude

### Suggestion

> You can specify the --exclude option multiple times for a COPY instruction. Multiple **--exclude** ...

alwaysalearner1234:
Closes #24993

## Summary

Fixes a typo in the Dockerfile reference documentation.

Changed incorrect flag reference:

* `--excludes`

- `--exclude`

to match the actual supported Dockerfile option syntax.

### DIFF
--- a/_vendor/github.com/moby/buildkit/frontend/dockerfile/docs/reference.md
+++ b/_vendor/github.com/moby/buildkit/frontend/dockerfile/docs/reference.md
@@ -2062,7 +2062,7 @@ COPY --exclude=*.txt hom* /mydir/
 ```
 
 You can specify the `--exclude` option multiple times for a `COPY` instruction.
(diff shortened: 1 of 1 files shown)


dvdksn:
Thanks for contributing - but please note that this file is vendored from an upstream repository: moby/buildkit

It is not maintained here. The typo fix needs to be committed upstream.

PR #25037, opened 2026-05-12
issue triage 20143

### LINKED ISSUE #20143: Document special behavior when bind-mounting docker socket?
### Is this a docs issue?

### Type of issue

### Description

(After much experimentation, it seems to me that) when running Docker Desktop for Linux, and not Engine, 
the `-v` flag when passed `/var/run/docker.sock` for the host machine filepath will mount `/var/run/docker.sock` from the VM; 
whereas in all other cases, the host machine filepath is interpreted as a path on... the host machine filesystem.

This is confusing in general, but especially if someone [reads](https://docs.docker.com/reference/cli/docker/container/run/#read-only) [somewhere](https://jpetazzo.github.io/2015/09/03/do-not-use-docker-in-docker-for-ci/#the-socket-solution) [that](https://docs.docker.com/desktop/hardened-desktop/enhanced-container-isolation/config/#docker-socket-mount-permissions) one can "bind mount the docker socket",
observes that there is no `/var/run/docker.sock` on their host,
does `docker context list` and see that they are talking to `/home/xxx/.docker/desktop/docker.sock`,
and attempts to mount that instead.

### Location

https://docs.docker.com

### Suggestion

_No response_

aevesdocker:
## Description

With help from my pal Claude, Closes https://github.com/docker/docs/issues/20143

## Reviews

- [ ] Technical review
- [ ] Editorial review
- [ ] Product review

### DIFF
--- a/content/manuals/desktop/troubleshoot-and-support/faqs/linuxfaqs.md
+++ b/content/manuals/desktop/troubleshoot-and-support/faqs/linuxfaqs.md
@@ -123,6 +123,28 @@ To make this permanent, add the export command to your shell profile (`~/.bashrc
 echo 'export DOCKER_HOST=unix://$HOME/.docker/desktop/docker.sock' >> ~/.bashrc
 ```
 
+> [!NOTE]
+>
+> `~/.docker/desktop/docker.sock` is only for connecting SDKs and tools to the Docker daemon from the host. It cannot be bind-mounted into a container. See [How do I bind-mount the Docker socket into a container?](#how-do-i-bind-mount-the-docker-socket-into-a-container)
+
+### How do I bind-mount the Docker socket into a container?
+
+Use `/var/run/docker.sock` as the host path:
+
+```console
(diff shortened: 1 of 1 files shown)


docker-agent (commented):
### Assessment: 🟢 APPROVE

No significant issues found. The new FAQ section about bind-mounting the Docker socket is technically accurate, well-structured, and matches the existing document's scope and tone.

dvdksn:
I think this PR might be solving a problem that just got fixed upstream. There's a commit in Docker Desktop from two weeks ago (https://github.com/docker/pinata/commit/86f61d143baee802ed2d0e8be1293fe719ea0c66) that adds `~/.docker/desktop/docker.sock` to the rewriter — so mounting that path directly now works the same as mounting `/var/run/docker.sock`. The original issue reporter back in 2024 was correct that it didn't work at the time, but it should now (?)

aevesdocker:
@dvdksn Ah! Well there we go! My pal Claude let me down. This PR is no longer needed. Will let the user know

PR #25038, opened 2026-05-12
docs: fix misleading secrets long syntax target description

### LINKED ISSUE #25035: Misleading example in Secrets Long Syntax
### Is this a docs issue?

### Type of issue

Information is incorrect

### Description

The Secrets Long syntax has the statement:
_Note that support for uid, gid, and mode attributes are not implemented in Docker Compose when the source of the secret is a [file](https://docs.docker.com/reference/compose-file/secrets/). This is because bind-mounts used under the hood don't allow uid remapping._

But the example after that uses secrets file with uid, gid, and mode specified.
_The following example sets the name of the server-certificate secret file to server.cert within the container, sets the mode to 0440 (group-readable), and sets the user and group to 103. The value of server-certificate is set to the contents of the file ./server.cert._

```
services:
  frontend:
    image: example/webapp
    secrets:
      - source: server-certificate
        target: server.cert
        uid: "103"
        gid: "103"
        mode: 0o440
secrets:
  server-certificate:
    file: ./server.cert
```
This seems like an inconsistency in documentation and needs to be updated with a relevant example that honours the uid,gid and mode values

### Location

https://docs.docker.com/reference/compose-file/services/#long-syntax-5

### Suggestion

The example could be something like this
```
services:
  app:
    image: busybox
    command: sh -c "ls -lan /run/secrets/ && cat /run/secrets/my-token"
    secrets:
      - source: my-token
        target: /run/secrets/my-token
        uid: "1000"
        gid: "1000"
        mode: 0o440

secrets:
  my-token:
    environment: "MY_TOKEN"
```

mohithshuka:
## What changed
Removed the incorrect claim that `target` in the secrets long syntax 
accepts an absolute path for an alternate mount location.

## Why
The `target` field only accepts a filename, which gets mounted under 
`/run/secrets/<target>`. Specifying an absolute path is a Docker Swarm 
feature and is **not supported** in Docker Compose standalone. This 
was misleading users into thinking they could mount secrets at arbitrary 
paths in Compose.

## Related issue
Fixes #25035


aevesdocker:
again, closing as this contains docs edits that are not relevant to the PR description.

PR #25055, opened 2026-05-13
poc 404 redesign

craig-osterhout:
## Description

Agent redesigned 404 page to be dynamic. 

- Added suggested pages based on the 404 URL. Hackily grabbing the links from Gordon's sources as a semantic search workaround. Kind of sloooow, but works.
- Removed go to homepage link as we have nav right there on the page.
- Removed the image because it has the homepage reference and it feels out of place compared to site designs. Tried to copy more of Hub's/Scout's/etc style.
- Added specifically to create issue if link is from a Docker product. Not sure if anyone will read it, but it seems like many issues are from third-party links or typos.

Example:
https://deploy-preview-25055--docsdocker.netlify.app/install/dd/

## Reviews

- [ ] Technical review
- [ ] Editorial review
- [ ] Product review


layouts/404.html
```diff
@@ -2,46 +2,185 @@
...
+
+          // Drain the full stream into a buffer
+          const reader = gordonResponse.body.getReader();
+          const decoder = new TextDecoder();
+          let buffer = '';
+
+          while (true) {
+            const { done, value } = await reader.read();
+            if (done) break;
+            for (const line of decoder.decode(value, { stream: true }).split('\n')) {
```
docker-agent: **[MEDIUM] SSE stream line parsing drops partial `data:` lines split across `read()` chunks**

`TextDecoder` with `{ stream: true }` only buffers incomplete multi-byte characters — it does **not** buffer incomplete lines. If the server sends a large SSE `data: {...}` line and it happens to be split across two `reader.read()` calls, the first chunk produces a line without a trailing `\n` (so `split('\n')` yields it as a last partial element that may or may not start with `'data: '`), and the second chunk starts mid-JSON, not with `'data: '`, so it is silently skipped by `if (!line.startsWith('data: ')) continue`.

The missing fragment causes `JSON.parse(data)` to fail (caught silently), and the content delta is lost. This results in truncated AI-generated text in `buffer`, which can mean the `Sources:` section is missing or incomplete — leading to fewer or no link suggestions shown to the user.

**Fix:** Maintain a separate accumulator for partial SSE lines across `read()` calls:
```javascript
let sseBuffer = '';
while (true) {
  const { done, value } = await reader.read();
  if (done) break;
  sseBuffer += decoder.decode(value, { stream: true });
  const lines = sseBuffer.split('\n');
  sseBuffer = lines.pop(); // keep incomplete last line for next iteration
  for (const line of lines) {
    if (!line.startsWith('data: ')) continue;
    // ...existing parsing logic...
  }
}
```

layouts/404.html
```diff
@@ -2,46 +2,185 @@
...
+              headers: { 'Content-Type': 'application/json' },
+              body: JSON.stringify({
+                messages: [{
+                  role: 'user',
+                  content: `The page "${path}" was not found. Suggest alternative pages.`,
+                  copilot_references: [{
+                    data: {
+                      origin: 'docs-website',
+                      email: 'docs@docker.com',
+                      uuid: `docs-404-${Date.now()}`,
```
docker-agent: **[LIKELY/MEDIUM] `uuid` field contains a timestamp string, not a UUID**

The `copilot_references[0].data.uuid` field is set to `` `docs-404-${Date.now()}` `` — a millisecond timestamp prefixed with `docs-404-`, not a UUID v4 (e.g. `550e8400-e29b-41d4-a716-446655440000`). If the Gordon API validates this field's format (the field is named `uuid`, strongly implying a UUID is expected), requests from the 404 page may be silently rejected or logged as malformed, causing Gordon to return a non-OK response and the suggestions widget to show nothing.

**Fix:** Use `crypto.randomUUID()` (available in all modern browsers) to generate a proper UUID v4:
```javascript
uuid: crypto.randomUUID(),
```

docker-agent (commented):

Two issues found in the new JavaScript for the Gordon 404 suggestions widget.

dvdksn:
@hamza-jeddad hey could you take a look at this one?

I think the idea is pretty neat - leverage Gordon to provide more useful 404 pages, recommending actually existing pages when someone ends up at a nonexistent route.

But rather than going through the whole retrieval and generation pipeline, this is another one of those cases where it'd be cool to have a lighter, keyword search endpoint to use, eg bm25. Similar to what I mentioned for the MCP server.

Reckon we could poc something here?

PR #25060, opened 2026-05-14
Compose issue triage

### LINKED ISSUE #23678: Setting secrets from environment variables doesn't work in a swarm deploy
### Is this a docs issue?

### Type of issue

Information is incorrect

### Description

It would be nice to have a note in the documentation to the effect that setting secrets from environment variables. eg:

```
secrets:
  borg_backup_ssh_key:
    environment: "BORG_BACKUP_SSH_KEY"
```

work when the compose file is used for `docker compose`, but NOT when it is used for `docker stack deploy`.  It appears that a file source is needed then.

### Location

https://docs.docker.com/reference/compose-file/secrets/

### Suggestion

It would be wonderful if the code could be fixed to allow this, but in the meantime, a note to the effect that the 'environment' clause only works in certain... ahem... environments would be very helpful.

### LINKED ISSUE #24853: Fix examples for Compose lifecycle hooks

### LINKED ISSUE #25035: Misleading example in Secrets Long Syntax

aevesdocker:
## Description

Closes https://github.com/docker/docs/issues/25035
Replaces and closes https://github.com/docker/docs/pull/24650
Closes https://github.com/docker/docs/issues/24853
Replaces and closes https://github.com/docker/docs/pull/24866
Closes https://github.com/docker/docs/issues/23678
Closes https://github.com/docker/docs/pull/24812
Replaces and closes https://github.com/docker/docs/pull/23816

Verified against the Compose code base
## Reviews

- [ ] Technical review
- [ ] Editorial review
- [ ] Product review

### DIFF
--- a/content/manuals/compose/how-tos/lifecycle.md
+++ b/content/manuals/compose/how-tos/lifecycle.md
@@ -53,9 +53,7 @@ command (like `docker compose down` or stopping it manually with `Ctrl+C`).
 These hooks won't run if the container stops by itself or gets killed suddenly.
 
 Because the pre-stop hook runs before the stop signal is sent to the container, it is
-suited for actions that must complete while the application is still fully running. For
-example, deregistering from a load balancer so that no new requests are routed to the
-container before it begins shutting down.
+suited for actions that must complete while the application is still fully running.
 
 In the following example, the hook backs up a data file before the container receives the stop signal.
 
--- a/content/manuals/compose/how-tos/lifecycle.md
+++ b/content/manuals/compose/how-tos/lifecycle.md
@@ -28,25 +28,22 @@ Post-start hooks are commands that run after the container has started, but ther
 set time for when exactly they will execute. The hook execution timing is not assured during 
 the execution of the container's `entrypoint`.
 
-In the example provided:
+Because there is no ordering guarantee between the hook and the container's entrypoint,
+post-start hooks are best suited for tasks that do not need to complete before the
+application begins running, such as registering the container with an external system.
 
-- The hook is used to change the ownership of a volume to a non-root user (because volumes 
-are created with root ownership by default).
-- After the container starts, the `chown` command changes the ownership of the `/data` directory to user `1001`.
+In the following example, after the container starts, a root-level hook registers the
+service with an internal service registry. The application does not depend on registration
+being complete before it starts serving requests.
 
 ```yaml
 services:
   app:
     image: backend
     user: 1001
-    volumes:
-      - data:/data    
     post_start:
-      - command: chown -R 1001:1001 /data
+      - command: /opt/scripts/register-service.sh
         user: root
-
-volumes:
-  data: {} # a Docker volume is created with root ownership
 ```
 
 ### Pre-stop hooks
@@ -55,15 +52,24 @@ Pre-stop hooks are commands that run before the container is stopped by a specif
 command (like `docker compose down` or stopping it manually with `Ctrl+C`). 
 These hooks won't run if the container stops by itself or gets killed suddenly.
 
-In the following example, before the container stops, the `./data_flush.sh` script is 
-run to perform any necessary cleanup.
+Because the pre-stop hook runs before the stop signal is sent to the container, it is
+suited for actions that must complete while the application is still fully running. For
+example, deregistering from a load balancer so that no new requests are routed to the
+container before it begins shutting down.
+
+In the following example, the hook backs up a data file before the container receives the stop signal.
 
 ```yaml
 services:
   app:
     image: backend
+    volumes:
+      - data:/data
     pre_stop:
-      - command: ./data_flush.sh
+      - command: cp /data/app.db /data/app.db.bak
+
+volumes:
+  data: {} # a Docker volume is created with root ownership
 ```
 
 ## Reference information
--- a/content/reference/compose-file/secrets.md
+++ b/content/reference/compose-file/secrets.md
@@ -15,7 +15,8 @@ The top-level `secrets` declaration defines or references sensitive data that is
 application. The source of the secret is either `file` or `environment`.
 
 - `file`: The secret is created with the contents of the file at the specified path.
-- `environment`: The secret is created with the value of an environment variable on the host.
+- `environment`: The secret is created with the value of an environment variable on the host. This is only supported by Docker Compose. It is not supported when deploying with [`docker stack deploy`](/manuals/engine/swarm/stack-deploy.md).
+ 
 
 ## Example 1
 
@@ -38,6 +39,10 @@ secrets:
   token:
     environment: "OAUTH_TOKEN"
 ```
+
+> [!NOTE]
+> `environment` secrets are not supported when deploying with `docker stack deploy`.
+> Use `file` or `external` as the secret source instead.
 
 ## Additional resources
 
--- a/content/reference/compose-file/services.md
+++ b/content/reference/compose-file/services.md
@@ -785,6 +785,8 @@ extends:
 - `service`: Defines the name of the service being referenced as a base, for example `web` or `database`.
 - `file`: The location of a Compose configuration file defining that service.
 
+`extends` is not supported when deploying with `docker stack deploy`.
+
 #### Restrictions
 
 When a service is referenced using `extends`, it can declare dependencies on other resources. These dependencies may be explicitly defined through attributes like `volumes`, `networks`, `configs`, `secrets`, `links`, `volumes_from`, or `depends_on`. Alternatively, dependencies can reference another service using the `service:{name}` syntax in namespace declarations such as `ipc`, `pid`, or `network_mode`.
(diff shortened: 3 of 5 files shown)


content/reference/compose-file/services.md
```diff
@@ -1940,26 +1940,24 @@ the service's containers.
...
 
-Note that support for `uid`, `gid`, and `mode` attributes are not implemented in Docker Compose when the source of the secret is a [`file`](secrets.md). This is because bind-mounts used under the hood don't allow uid remapping.
+Note that support for `uid`, `gid`, and `mode` attributes are only implemented in Docker Compose when the source of the secret is [`environment`](secrets.md). When the source is a [`file`](secrets.md), Compose uses a bind-mount under the hood which doesn't allow `uid` remapping, and these attributes are silently ignored.
 
-The following example sets the name of the `server-certificate` secret file to `server.cert`
-within the container, sets the mode to `0440` (group-readable), and sets the user and group
-to `103`. The value of `server-certificate` is set
-to the contents of the file `./server.cert`.
+The following example sets the name of the `my-token` secret file within the container,
++sets the mode to `0440` (group-readable), and sets the user and group to `103`.
```
docker-agent: **[HIGH] Raw diff `+` markers leaked into prose**

Lines 1946–1947 in the committed file literally begin with a `+` character:

```
+sets the mode to `0440` (group-readable), and sets the user and group to `103`.
+The value of `my-token` is read from the `MY_TOKEN` environment variable.
```

These are bare `+` diff markers that appear to have been accidentally committed into the file. They will render as literal `+` characters in the published documentation, breaking the prose mid-sentence. The sentence starting on line 1945 ("The following example sets the name of the `my-token` secret file within the container,") is left dangling.

Please remove the leading `+` characters so the text reads as intended.

content/reference/compose-file/services.md
```diff
@@ -1940,26 +1940,24 @@ the service's containers.
...
-        target: server.cert
+      - source: my-token
         uid: "103"
         gid: "103"
         mode: 0o440
 secrets:
-  server-certificate:
-    file: ./server.cert
+  my-token:
+    environment: "My_TOKEN"
```
docker-agent: **[HIGH] Case mismatch: prose says `MY_TOKEN`, YAML example uses `My_TOKEN`**

The prose on line 1947 states:
> The value of `my-token` is read from the `MY_TOKEN` environment variable.

But the YAML example on line 1960 uses:
```yaml
my-token:
  environment: "My_TOKEN"
```

On Linux, environment variable names are case-sensitive — `MY_TOKEN` and `My_TOKEN` are two different variables. Users following this example and exporting `MY_TOKEN` in their shell will find the secret is not picked up. Please make both values consistent (conventionally, environment variable names are `ALL_CAPS`, so `MY_TOKEN` is likely the intended form).

content/manuals/compose/how-tos/lifecycle.md
```diff
@@ -55,15 +52,24 @@ Pre-stop hooks are commands that run before the container is stopped by a specif
 command (like `docker compose down` or stopping it manually with `Ctrl+C`). 
 These hooks won't run if the container stops by itself or gets killed suddenly.
 
-In the following example, before the container stops, the `./data_flush.sh` script is 
-run to perform any necessary cleanup.
+Because the pre-stop hook runs before the stop signal is sent to the container, it is
+suited for actions that must complete while the application is still fully running. For
+example, deregistering from a load balancer so that no new requests are routed to the
```
docker-agent: **[MEDIUM] Sentence fragment after "For example"**

The sentence "For example, deregistering from a load balancer so that no new requests are routed to the container before it begins shutting down." is a fragment — it has no main verb. Consider rewriting as a complete sentence, for example:

> For example, you can deregister from a load balancer so that no new requests are routed to the container before it begins shutting down.

content/reference/compose-file/volumes.md
```diff
@@ -71,6 +71,20 @@ volumes:
...
+If you want a named bind mount, use the `local` driver with `driver_opts`. This pattern gives a Compose volume a stable name while mapping it to a specific host path:
+
+```yaml
+volumes:
+  app-data:
+    driver: local
+    driver_opts:
+      type: none
+      o: bind
+      device: /srv/app-data # must be the abosolute host path and already exist
```
docker-agent: **[LOW] Typo in code comment: "abosolute" → "absolute"**

```yaml
device: /srv/app-data # must be the abosolute host path and already exist
```

"abosolute" should be "absolute".

docker-agent (commented):
### Assessment: 🔴 CRITICAL

Two high-severity issues were found in the added content: raw diff markers committed into prose, and a case-sensitive environment variable name mismatch that would cause the documented example to fail.

content/reference/compose-file/secrets.md
```diff
@@ -15,7 +15,8 @@ The top-level `secrets` declaration defines or references sensitive data that is
 application. The source of the secret is either `file` or `environment`.
 
 - `file`: The secret is created with the contents of the file at the specified path.
-- `environment`: The secret is created with the value of an environment variable on the host.
+- `environment`: The secret is created with the value of an environment variable on the host. This is only supported by Docker Compose. It is not supported when deploying with [`docker stack deploy`](/manuals/engine/swarm/stack-deploy.md).
```
docker-agent, line 18: **[LOW] Duplicate information: stack deploy limitation stated twice**

The inline sentence on this line ("This is only supported by Docker Compose. It is not supported when deploying with `docker stack deploy`...") duplicates the `> [!NOTE]` callout a few lines below that conveys the same limitation. Per the style guide's conciseness guidelines, consider removing the inline repetition and keeping only the NOTE callout, or merging both into a single callout.

content/reference/compose-file/services.md
```diff
@@ -785,6 +785,8 @@ extends:
 - `service`: Defines the name of the service being referenced as a base, for example `web` or `database`.
 - `file`: The location of a Compose configuration file defining that service.
 
+`extends` is not supported when deploying with `docker stack deploy`.
```
docker-agent, line 788: **[LOW] Inconsistent formatting: limitation as bare prose vs NOTE callout**

The companion `extends.md` page uses a `> [!NOTE]` callout for the same `docker stack deploy` limitation. For consistency and scannability across related reference pages, consider using the same callout format here.

content/reference/compose-file/volumes.md
```diff
@@ -71,6 +71,20 @@ volumes:
...
+volumes:
+  app-data:
+    driver: local
+    driver_opts:
+      type: none
+      o: bind
+      device: /srv/app-data # must be the absolute host path and already exist
+```
+
+The `type`, `o`, and `device` keys are passed through to the local driver. For a one-off host-path mount on a single service, see [bind mounts](/manuals/engine/storage/bind-mounts.md).
```
docker-agent, line 86: **[LOW] Passive voice: prefer active voice per style guide**

"The `type`, `o`, and `device` keys are passed through to the local driver." uses passive voice. Consider: "The local driver receives the `type`, `o`, and `device` keys directly."

docker-agent (commented):
### Assessment: 🟢 APPROVE

The changes correctly document  and  secrets limitations for , add a named bind mount example to volumes.md, and clean up lifecycle.md examples. No incorrect commands, broken Markdown, missing redirects, or moved pages found. Only minor style observations noted as inline comments.

dvdksn:
Note: "Replaces and closes #xx" would not automatically close the referenced issue. They are only closed if the keyword appears at the beginning of the line or in a list, eg

Closes #xx
Fixes #xx

or:

- Resolves #xx

usha-mandya (approved):

aevesdocker:
Replaces and closes seemed to work! 🎉

dvdksn:
OK good to know :-)

PR #25074, opened 2026-05-16
sbx: clarify sandbox auto-naming in docs

Pradumnasaraf:
## Description

Running `sbx run claude` from `~/my-pro` produces a sandbox named `claude-my-pro`, but the docs tell you to clean up with `sbx rm my-sandbox` and never explain where that name came from. 

This PR documents the `<agent>-<directory>` and updates examples across the sandbox docs to use names that match the rule, so the flow lines up with what users actually see in `sbx ls`. 

Screenshot 2026-05-16 at 2 48 30 PM

## Reviews

- [x] Technical review
- [ ] Editorial review
- [ ] Product review

### DIFF
--- a/content/manuals/ai/sandboxes/customize/kits.md
+++ b/content/manuals/ai/sandboxes/customize/kits.md
@@ -337,7 +337,7 @@ While iterating on a kit, apply changes to a running sandbox with
 `sbx kit add` instead of recreating it:
 
 ```console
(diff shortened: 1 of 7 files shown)


shivaylamba (commented):
Amazing PR - makes a lot of sense and makes it super simple to understand!

dvdksn (approved):
Thank you @Pradumnasaraf

PR #25075, opened 2026-05-16
docs: add environment variables configuration method for Docker daemon

### LINKED ISSUE #25062: [docs-scanner] Daemon configuration overview incomplete - missing environment variables method
**File:** `content/manuals/engine/daemon/_index.md`

### Issue

The daemon configuration overview states there are only two ways to configure the Docker daemon, but omits environment variables as a third configuration method:

> There are two ways to configure the Docker daemon:
> 
> - Use a JSON configuration file. This is the preferred option, since it keeps
>   all configurations in a single place.
> - Use flags when starting `dockerd`.

However, `proxy.md` in the same directory documents environment variables as a distinct configuration method:

> There are two ways you can configure these settings:
> 
> - [Configuring the daemon](#daemon-configuration) through a configuration file or CLI flags
> - Setting [environment variables](#environment-variables) on the system

### Why this matters

A reader following the overview page would not learn about environment variables as a configuration option. When they later encounter proxy configuration documentation that relies on environment variables, they may be confused about whether this is a general daemon configuration method or specific to proxies only.

The overview page should provide a complete picture of all daemon configuration methods so readers understand their options from the start.

### Suggested fix

Update the "Configure the Docker daemon" section in `_index.md` to mention all three configuration methods:

```markdown
There are three ways to configure the Docker daemon:

- Use a JSON configuration file. This is the preferred option, since it keeps
  all configurations in a single place.
- Use flags when starting `dockerd`.
- Set environment variables in the daemon's startup environment (used for specific settings like proxy configuration).
```

Or add a note that some specific configurations (like proxies) can also use environment variables, with a link to the relevant documentation.

---

mohithshuka:
## What changed
Added a new "Configuration using environment variables" section to the
Docker daemon configuration overview page.

## Why
The page stated there are "two ways" to configure the Docker daemon
(JSON file and flags) but was missing the environment variables method
using systemd drop-in overrides.

Fixes #25062

## Testing
- Verified the new section renders correctly in markdown
- No broken links introduced


content/manuals/engine/daemon/_index.md
```diff
@@ -92,6 +93,32 @@ running:
...
+
+````````console
+$ sudo systemctl edit docker
+```````
+
+Add the following:
+
+``````ini
+[Service]
+Environment="DOCKER_OPTS=--debug --tls=true"
```
thaJeztah, line 112: This is not something we support, the `DOCKER_OPTS` env-var was a legacy option before systemd; 
https://github.com/moby/moby/blob/721883f7329e0c555ec00126ecb3bbe5511d50f5/contrib/init/openrc/docker.confd#L31-L32

_some_ options can also be set through environment variables, but the list is limited;
https://docs.docker.com/reference/cli/dockerd/#environment-variables


So, not entirely sure if we want to advertise this as an alternative to the `daemon.json` or CLI flags, because it's not a replacement (only for some options).


cc @dvdksn any thoughts?

content/manuals/engine/daemon/_index.md
```diff
@@ -92,6 +93,32 @@ running:
 $ dockerd --help
 ```
 
+### Configuration using environment variables
+
+You can configure the Docker daemon using environment variables. This is useful
+in systemd service overrides or init scripts.
+
+To set environment variables for the Docker daemon using systemd, create a
+drop-in override file:
```
thaJeztah, line 102: See my other comment; I think this is a very limited use-case, so not 100% sure if we should advertise this here; we could link to the https://docs.docker.com/reference/cli/dockerd/#environment-variables section to make users aware of environment variables that are supported, but possibly leave it an exercise to the user to use those (which is just general systemd knowledge)

thaJeztah (changes requested):

content/manuals/engine/daemon/_index.md
dvdksn: Sorry only saw this now. Yeah I would keep this undocumented personally; I don't see a reason for adding it in now.

PR #25080, opened 2026-05-17
build: clarify registry cache driver support

### LINKED ISSUE #25019: Registry cache backend availability is outdated
### Is this a docs issue?

### Type of issue

Information is incorrect

### Description

[According to the docs](https://docs.docker.com/build/cache/backends/registry/#:~:text=This%20cache%20storage%20backend%20is%20not%20supported%20with%20the%20default%20docker%20driver%2E%20To%20use%20this%20feature%2C%20create%20a%20new%20builder%20using%20a%20different%20driver%2E):
> [The registry] cache storage backend is not supported with the default docker driver. To use this feature, create a new builder using a different driver.

However based on my testing, this is no longer true -- the `registry` cache backend works fine with the default docker driver. 

### Location

https://docs.docker.com/build/cache/backends/registry/

### Suggestion

_No response_

henriquejsza:
## Summary

Clarify that the registry cache backend works with the default `docker` driver when the containerd image store is enabled, and direct users to another driver otherwise.

Closes #25019

### DIFF
--- a/content/manuals/build/cache/backends/registry.md
+++ b/content/manuals/build/cache/backends/registry.md
@@ -18,8 +18,9 @@ everything that the inline cache can do, and more:
 - It works with other exporters for more flexibility, instead of only the
   `image` exporter.
 
-This cache storage backend is not supported with the default `docker` driver.
-To use this feature, create a new builder using a different driver. See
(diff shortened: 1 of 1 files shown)


dvdksn (approved):
LGTM!

PR #25086, opened 2026-05-18
build: GitHub Builder runner mappings

crazy-max:
follow-up https://github.com/docker/github-builder/pull/188

The reusable workflows now accept either a GitHub-hosted Linux runner label or a newline-delimited platform mapping, while the legacy `auto`, `amd64`, and `arm64` values are deprecated.

### DIFF
--- a/content/manuals/build/ci/github-actions/github-builder/architecture.md
+++ b/content/manuals/build/ci/github-actions/github-builder/architecture.md
@@ -27,11 +27,11 @@ conditions, permissions, secrets, target selection, metadata inputs, and the
 choice between image output and local output.
 
 Inside the reusable workflow, the first phase prepares the build. It validates
-the incoming inputs, resolves the appropriate runner, and expands a
-multi-platform request into one job per platform. The execution model is
-easiest to picture as a matrix where `linux/amd64` runs on `ubuntu-24.04` and
-`linux/arm64` runs on `ubuntu-24.04-arm`. Each platform job builds independently,
-then the workflow finalizes the result into one caller-facing output contract.
+the incoming inputs, resolves the runner config, and expands a multi-platform
+request into one job per platform. The execution model is easiest to picture as
+a matrix where `linux/amd64` runs on `ubuntu-24.04` and `linux/arm64` runs on
+`ubuntu-24.04-arm`. Each platform job builds independently, then the workflow
+finalizes the result into one caller-facing output contract.
 
 ```yaml
 requested platforms:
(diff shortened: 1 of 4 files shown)


content/manuals/build/ci/github-actions/github-builder/architecture.md
```diff
@@ -42,6 +42,40 @@ conceptual platform jobs:
...
+```
+
+A mapping must define a `default` runner. Other keys are platform prefixes, and
+the most specific matching prefix wins. For example, `linux/arm` matches
+variants such as `linux/arm/v7`, while `linux/arm64` is a separate prefix:
+
+```yaml
+runner: |
+  default=ubuntu-24.04
+  linux=ubuntu-24.04
```
docker-agent, line 76: **[MEDIUM] Second mapping example introduces `linux=` key alongside `default=` without explaining their interaction**

The second `runner` mapping example adds `linux=ubuntu-24.04` next to `default=ubuntu-24.04`, but the surrounding prose doesn't explain why both are needed or how they differ.

The key distinction — `default` catches non-Linux platforms (e.g. hypothetical Windows/macOS runners) while `linux=` is a more-specific catch-all for any Linux platform not matched by a longer prefix — is left implicit. Since the workflows only support Linux runners in practice, a reader is likely to find `linux=ubuntu-24.04` redundant with `default=ubuntu-24.04` and wonder why it appears at all.

Consider adding a brief explanation before or after the example, such as: "`linux` matches any Linux platform not caught by a more specific key, while `default` covers platforms that don't match any prefix."

docker-agent (commented):

content/manuals/build/ci/github-actions/github-builder/architecture.md
```diff
@@ -42,6 +42,44 @@ conceptual platform jobs:
   linux/arm64 -> ubuntu-24.04-arm
 ```
 
+### Runner selection
+
+The `runner` input accepts either a single GitHub-hosted Linux runner label or a
```
thaJeztah, line 47: Is there a page we could link to in the GitHub docs that shows the available labels?

crazy-max: Yes there is https://docs.github.com/en/actions/reference/runners/github-hosted-runners#supported-runners-and-hardware-resources and https://github.com/actions/partner-runner-images/#available-images but not sure about linking to these pages though :thinking:

content/manuals/build/ci/github-actions/github-builder/architecture.md
```diff
@@ -42,6 +42,44 @@ conceptual platform jobs:
...
+```
+
+The default value is a platform mapping that uses GitHub-hosted Ubuntu runners:
+
+```yaml
+runner: |
+  default=ubuntu-24.04
+  linux/arm=ubuntu-24.04-arm
+  linux/arm64=ubuntu-24.04-arm
+```
```
dvdksn: Suggest we re-order this to show the default first, and what it resolves to after.

````suggestion
newline-delimited platform mapping.

The default value is a platform mapping that uses GitHub-hosted Ubuntu runners:

```yaml
runner: |
  default=ubuntu-24.04
  linux/arm=ubuntu-24.04-arm
  linux/arm64=ubuntu-24.04-arm
```

In the platform job, the runner label resolves to a single value:

```yaml
runner: ubuntu-24.04
```
````

crazy-max: Yes sounds better indeeed

dvdksn (approved):
LGTM ❤️

PR #25092, opened 2026-05-19
Add documentation for `HTTP_AUTH_TOKEN_`/`HTTP_AUTH_HEADER_`.

bglimepoint:
## Description

Add documentation for the special `HTTP_AUTH_TOKEN_<host>` and `HTTP_AUTH_HEADER_<host>` secrets that are supported by Buildkit.

I couldn't find any documentation of this feature anywhere, but stumbled across https://github.com/moby/buildkit/pull/6023 when trying to use secrets in a Dockerfile `COPY`, and found it solved our issue.

Absolutely no worries if this needs a tweak or anything, but I felt like it was worth documenting (as I was very surprised when I couldn't find this feature anywhere) :-)

Thanks :-)

### DIFF
--- a/content/manuals/build/building/secrets.md
+++ b/content/manuals/build/building/secrets.md
@@ -264,3 +264,13 @@ $ docker build \
   --secret id=GIT_AUTH_HEADER.gitlab.com,env=GITLAB_AUTH_HEADER \
   https://github.com/example/todo-app.git
 ```
+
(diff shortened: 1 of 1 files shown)


dvdksn (approved):
Huh good catch!

tonistiigi (approved):
Ah, I used a different separator for `HTTP` and `GIT` secrets? 😬 Maybe we should fix that in a future release.

PR #25108, opened 2026-05-19
Add CloudFront URL to allow list

carloslemus-hub:
## Reviews

- [ ] Technical review
- [ ] Editorial review
- [ ] Product review

### DIFF
--- a/content/manuals/desktop/setup/allow-list.md
+++ b/content/manuals/desktop/setup/allow-list.md
@@ -27,6 +27,7 @@ This page contains the domain URLs that you need to add to a firewall allowlist 
 | https://hub.docker.com                                                               | Docker Hub                                   |
 | https://registry-1.docker.io                                                         | Docker Pull/Push                             |
 | https://production.cloudflare.docker.com                                             | Docker Pull/Push (Paid plans)                |
+| https://production.cloudfront.docker.com                                             | Docker Pull/Push                             |
 | https://docker-images-prod.6aa30f8b08e16409b46e0173d6de2f56.r2.cloudflarestorage.com | Docker Pull/Push (Personal plan / Anonymous) |
 | https://docker-pinata-support.s3.amazonaws.com                                       | Troubleshooting                              |
 | https://api.dso.docker.com                                                           | Docker Scout service                         |


content/manuals/desktop/setup/allow-list.md
```diff
@@ -27,6 +27,7 @@ This page contains the domain URLs that you need to add to a firewall allowlist
 | https://hub.docker.com                                                               | Docker Hub                                   |
 | https://registry-1.docker.io                                                         | Docker Pull/Push                             |
 | https://production.cloudflare.docker.com                                             | Docker Pull/Push (Paid plans)                |
+| https://production.cloudfront.docker.com                                             | Docker Pull/Push                             |
```
docker-agent, line 30: **[MEDIUM] Description 'Docker Pull/Push' lacks a plan/scope qualifier**

The new row describes `https://production.cloudfront.docker.com` as "Docker Pull/Push" with no qualifier, while the two directly adjacent rows carry explicit qualifiers — "Docker Pull/Push (Paid plans)" and "Docker Pull/Push (Personal plan / Anonymous)". This mixed pattern may leave administrators uncertain about which subscription tier or user group this CloudFront CDN URL applies to, potentially resulting in an overly broad or incomplete allowlist configuration.

Note: `https://registry-1.docker.io` (an existing row) also uses the unqualified form, so this is a pre-existing inconsistency in the table. Adding a brief qualifier — such as "(All plans)" if this URL applies universally — would improve clarity for all three unqualified rows and make the table easier to reason about.

docker-agent (commented):

content/manuals/desktop/setup/allow-list.md
carloslemus-hub: This is for BOTH paid and unpaid users.

docker-agent: Thanks for clarifying! If this CloudFront URL is used by both paid and unpaid users, then adding "(All plans)" or "(All users)" as a qualifier would make that clear and help the table remain consistent.

This would also be a good opportunity to clarify https://registry-1.docker.io on line 29, which has the same unqualified description. Making both explicit would help administrators understand at a glance which URLs apply universally versus which are tier-specific.

craig-osterhout (approved):

PR #25120, opened 2026-05-20
Fix restricted shortcode rendering on unassociated machines page

gmondello:
## Description

The restricted shortcode on the [unassociated machines page](https://docs.docker.com/unassociated-machines/) renders with a visible `</span>` tag due to using `{{% %}}` (markdown) delimiters instead of `{{< >}}` (HTML) delimiters.

The `{{% %}}` delimiter causes Hugo to process the shortcode's HTML output through the markdown renderer, which escapes the HTML tags from the icon partial.

## Fix

Switch from `{{% restricted %}}` to `{{< restricted >}}` delimiters.

This matches the pattern used by the similar `experimental` shortcode in other content files (e.g., `layouts/cli.html`).

## Testing

Before (Prod) | After (Preview)
--- | ---
image | image

### DIFF
--- a/content/manuals/unassociated-machines/_index.md
+++ b/content/manuals/unassociated-machines/_index.md
@@ -10,10 +10,10 @@ params:
     group: Enterprise
 ---
 
-{{% restricted title="About unassociated machines" %}}
+{{< restricted title="About unassociated machines" >}}
 Unassociated machines is a private feature that may not be available to all
 accounts.
-{{% /restricted %}}
+{{< /restricted >}}
 
 Docker administrators can identify, view, and manage Docker Desktop machines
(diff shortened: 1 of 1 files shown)


docker-agent (commented):
### Assessment: 🟢 APPROVE

The change correctly fixes the Hugo shortcode delimiter issue. Switching from `{{% restricted %}}` to `{{< restricted >}}` prevents the markdown renderer from escaping the shortcode's HTML output (the `<span>` tags from the icon partial), which was causing a visible `</span>` tag to appear on the page. No content, front matter, links, or other documentation elements were modified. This is a clean, minimal, correct fix.

craig-osterhout (approved):
thank you

PR #25164, opened 2026-05-25
Add case-sensitivity note for Docker ID

0xIkari:
Clarify that the Docker ID is case-sensitive when setting a new organization owner.

## Description

Added explicit instruction the Docker ID is case sensitive.

## Reviews

- [ ] Technical review
- [ ] Editorial review
- [ ] Product review


dvdksn (approved):
thanks!

PR #23598, opened 2025-10-23
iam: idp-initiated doc updates

sarahsanders-docker:
## Description
- Doc updates for IdP-initiated SSO flow, w/ security warnings and recommendations 

## Related issues or tickets
https://docker.atlassian.net/browse/ENGDOCS-3064

## Reviews
@ivan-californias

PR #25287, opened 2026-06-05
stale workflow: update permissions

craig-osterhout:
## Description

Fix stale action not closing PRs due to missing `actions: write` permission

The stale action uses GitHub Actions cache to persist state between runs. Without `actions: write`, it can't delete its `_state` cache entry (fails with 403), so the old state accumulates and PRs in that cached state are permanently skipped with `processed during the previous run`, meaning they never get evaluated for closing.

Confirmed via Actions logs: PR #23598 was marked stale on April 24 but has never been evaluated for closing because it appears in the stuck cache state every run.

## Reviews

- [ ] Technical review
- [ ] Editorial review
- [ ] Product review

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -154,6 +154,7 @@ jobs:
     permissions:
       issues: write
       pull-requests: write
+      actions: write  # required for actions/stale to delete its state cache between runs
 
     steps:
       - uses: actions/stale@b5d41d4e1d5dceea10e7104786b73624c18a190f # v10.2.0


.github/workflows/stale.yml
```diff
@@ -154,6 +154,7 @@ jobs:
     permissions:
       issues: write
       pull-requests: write
+      actions: write  # required for actions/stale to delete its state cache between runs
```
docker-agent, line 157: **[LOW] `actions: write` grants broader workflow permissions beyond cache deletion**

`actions: write` is the minimal permission that satisfies `actions/stale`'s requirement to delete its `_state` cache entry, and the inline comment is accurate. However, this permission also implicitly allows cancelling, re-running, or deleting other workflow runs in the repository. The risk is low given the job only runs the pinned `actions/stale` action, but worth noting for awareness.

No change needed — this is informational only.

docker-agent (commented):
### Assessment: 🟢 APPROVE

This is a correct and well-scoped fix. Adding `actions: write` is the documented requirement for `actions/stale` v9+ to delete its cache-based state between runs, and the inline comment accurately describes the reason. The permission is scoped to the job level, and the job only runs the pinned `actions/stale` action.

dvdksn (approved):
LGTM

PR #25301, opened 2026-06-07
Avoid using the `newgrp` command-line tool in the doc

### LINKED ISSUE #910: The documentation should prompt users to install `util-linux-extra` to use `newgrp docker`
**Tell us about your request**
A clear and concise description of what you want to happen or the change you would like to see

- The documentation should prompt users to install `util-linux-extra` to use `newgrp docker`. See https://docs.docker.com/engine/install/linux-postinstall/ .

- Image

- After all, Ubuntu 26.04 doesn't have `newgrp` installed by default.
```bash
lingh@DESKTOP-T569R06:~$ newgrp docker
Command 'newgrp' not found, but can be installed with:
sudo apt install util-linux-extra
lingh@DESKTOP-T569R06:~$
```

**Which service(s) is this request for?**
Let us know which product(s) you want this for?

- `Docker Engine`

**Tell us about the problem you're trying to solve. What are you trying to do, and why is it hard?**

- Updating documents is not difficult at all.

**Are you currently working around the issue?**
A clear and concise description of any alternative solutions or features you've considered or are using today.

- No.

**Additional context**

- Null.

linghengqian:
## Description

- Fixes https://github.com/docker/roadmap/issues/910 .
- Avoid using the `newgrp` command-line tool in the doc. I don't think requiring users to install an additional apt package is a good option.
- Also see https://github.com/ddev/ddev/issues/8350 .
## Related issues or tickets

- https://github.com/docker/roadmap/issues/910

## Reviews

- [ ] Technical review
- [ ] Editorial review
- [ ] Product review

### DIFF
--- a/content/manuals/engine/install/linux-postinstall.md
+++ b/content/manuals/engine/install/linux-postinstall.md
@@ -63,7 +63,7 @@ To create the `docker` group and add your user:
    You can also run the following command to activate the changes to groups:
 
    ```console
-   $ newgrp docker
+   $ sudo -u $USER bash
    ```
 
 4. Verify that you can run `docker` commands without `sudo`.


content/manuals/engine/install/linux-postinstall.md
```diff
@@ -63,7 +63,7 @@ To create the `docker` group and add your user:
    You can also run the following command to activate the changes to groups:
 
    ```console
-   $ newgrp docker
+   $ sudo -u $USER bash
```
devendrakushwah80, line 66: Thanks for fixing the newgrp dependency problem. Is `sudo -u $USER bash` known to refresh group membership like `newgrp docker` does, uniformly across distributions? Maybe it's worth clarifying this for users following the post-installation steps.

linghengqian: > uniformly across distributions?


Yes.

> Maybe it's worth clarifying this for users following the post-installation steps.

What are you talking about? Isn't this exactly what current PRs are doing?

devendrakushwah80: Thanks for the clarification. My concern was mainly whether readers would understand that `sudo -u $USER bash` is intended to refresh the session with the updated group membership. If that's already clear from the surrounding documentation, then no further clarification is needed from my side.

PR #25305, opened 2026-06-08
fix: add missing useState import and remove duplicate ddClient in backend extension tutorial

### LINKED ISSUE #25192: [docs-scanner] Missing useState import in backend extension React example
**File:** `content/manuals/extensions/extensions-sdk/build/backend-extension-tutorial.md`

### Issue

The React code example in the "Invoke the extension backend from your frontend" section uses the `useState` hook but doesn't import it, which would cause a compilation error.

The import statement (line ~240):
> ```tsx
> import React, { useEffect } from 'react';
> ```

But the code uses `useState` (line ~248):
> ```tsx
> const [hello, setHello] = useState<string>();
> ```

Additionally, the code declares `ddClient` twice - once outside the component and once inside, which is redundant:
> ```tsx
> const ddClient = createDockerDesktopClient(); // Line ~244
> 
> export function App() {
>   const ddClient = createDockerDesktopClient(); // Line ~247 - redundant
> ```

### Why this matters

A developer copying this code would get a compilation error: `'useState' is not defined`. The code won't run until they figure out the missing import. The duplicate `ddClient` declaration, while not causing an error, adds confusion about the correct pattern.

### Suggested fix

Update the import statement to include `useState`:

```tsx
import React, { useEffect, useState } from 'react';
```

And remove the duplicate `ddClient` declaration inside the `App` function, keeping only the one outside (or vice versa, depending on the intended pattern).

---

jasdeepbhalla:
## Summary

- Add `useState` to the React import in the backend extension tutorial code example — it was used but missing from the import, causing a compilation error
- Remove the duplicate `createDockerDesktopClient()` call inside `App()` — `ddClient` is already declared at module scope

## Issue

Fixes #25192

### DIFF
--- a/content/manuals/extensions/extensions-sdk/build/backend-extension-tutorial.md
+++ b/content/manuals/extensions/extensions-sdk/build/backend-extension-tutorial.md
@@ -275,14 +275,13 @@ Replace the `ui/src/App.tsx` file with the following code:
 ```tsx
 
 // ui/src/App.tsx
-import React, { useEffect } from 'react';
+import React, { useEffect, useState } from 'react';
 import { createDockerDesktopClient } from "@docker/extension-api-client";
 
 //obtain docker desktop extension client
 const ddClient = createDockerDesktopClient();
 
 export function App() {
-  const ddClient = createDockerDesktopClient();
   const [hello, setHello] = useState<string>();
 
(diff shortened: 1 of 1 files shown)


aevesdocker (approved):
Thanks!

PR #25307, opened 2026-06-08
docs: restore tasks.<service-name> container discovery documentation

### LINKED ISSUE #24940: DNS Swarm Container discovery documentation about `tasks.<service-name>` appears to have dissapeared
### Is this a docs issue?

### Type of issue

#### History

Back in 2018, there was an issue (#5854) requesting to document a feature of Docker Swarm (as it was at that time) where swarm could respond to internal docker DNS queries matching the pattern `tasks.<service-name>` with a "list of all tasks backing the service."

I've seen anecdotal evidence online that the feature is still present. I haven't had the opportunity to test the presence of the feature in modern swarm, but even as it was back then, the documentation wasn't super clear on:

- The format for the DNS response
- How to best use the feature (perhaps with an example, I'm thinking like an example with a Redis or a Postgres cluster of some sort would help?)

#### Today

I've tried searching for modern documentation of the `tasks.<service-name>` feature and could only find the issue where it was first requested to be documented: #5854. I spent a bit of time trying to track down where the issue went, and documented some notes in [this comment](https://github.com/docker/docs/pull/6420#issuecomment-4362549381).

The documentation on this feature was dropped in #18998 (2 years ago, in 2024). I checked all network, overlay and swarm documentation at that commit (35b77e4cac7e64703c8bc8527a6f38f7ff742347) and found nothing on the topic. This was a very complicated pull request, covering a wide range of topics, and I can't blame anyone for missing something, especially something niche.

### Location

https://docs.docker.com/engine/swarm/networking/

### Suggestion

I don't know if the feature is present in modern swarm.

If it is not present (or is "deprecated"), it would be nice to have clarification on that in the documentation. (probably on this page: https://docs.docker.com/engine/deprecated/)

If it is present (and not "deprecated"), documentation of this feature equal to or better than the last snapshot at 2745b53bc49495c35baf05c01509ba32ea4e1d04 should be restored.

As for where it should go:

- If it is swarm specific, as I think it is, it should go here https://docs.docker.com/engine/swarm/networking/
- elsewise https://docs.docker.com/engine/network/drivers/overlay/

As I mentioned, if restoring the documentation of the feature is the correct route, and improvement of the documentation is worth it, it'd be nice to have an example and be more clear about what is provided from the DNS query and how to use it.

jasdeepbhalla:
## Summary

Restores the "Container discovery" section that was accidentally dropped in #18998 when swarm networking content was deduplicated from `network/drivers/overlay.md` into `engine/swarm/networking.md`.

The removed content documented that `tasks.<service-name>` DNS lookups return individual task IP addresses for a service — a useful swarm-specific feature for service-to-service discovery without going through the VIP.

The new section is placed directly after the VIP/DNSRR service discovery descriptions, within the "Configure service discovery" section, which is the right location for this content.

## Issue

Fixes #24940

### DIFF
--- a/content/manuals/engine/swarm/networking.md
+++ b/content/manuals/engine/swarm/networking.md
@@ -314,6 +314,14 @@ round robin (DNSRR). You can configure this per service.
   `--endpoint-mode dnsrr` when creating a new service or updating an existing
   one.
 
+### Container discovery
+
+For most situations, connect to the service name. Docker load-balances
+requests across all running tasks ("containers") backing the service. To
+resolve the IP addresses of all individual tasks backing a service directly,
+perform a DNS lookup for `tasks.<service-name>`. Docker returns a list of
+all task IP addresses for that service, one per running replica.
+
 ## Customize the ingress network {#customize-ingress}
 
 Most users never need to configure the `ingress` network, but Docker allows you


dvdksn (approved):
LGTM thanks!

PR #25309, opened 2026-06-09
docs: fix LocalStack port, Compose configs key, and a doubled link slash

emmayusufu:
Three small fixes in the guides.

- `guides/localstack`: the prose says `S3_ENDPOINT_URL` is `http://localhost:4556`, but LocalStack's edge port is `4566`, and the guide's own `.env` block later uses `4566`.
- `guides/pgadmin`: the top-level Compose key for the pgpass file is `configs`, not `config`. As written the compose file is invalid.
- `guides/bun/deploy`: the "Turn on Kubernetes" link has a doubled slash (`/manuals//desktop/...`).


craig-osterhout (approved):
thank you. lgtm

PR #25342, opened 2026-06-12
docs: document multi-platform build limitation for Docker driver

### LINKED ISSUE #25044: Page doesn't say it does not support multiplatform
### Is this a docs issue?

### Type of issue

Information is incorrect

### Description

by omitting that multiplatform doesnt work it causes the user to search for an answer that doesn't exist. 

### Location

https://docs.docker.com/build/ci/github-actions/share-image-jobs/

### Suggestion

here is the answer: https://github.com/docker/buildx/discussions/2718#discussioncomment-14182543 - this part of docker doesn't work, but you can use docker's cache to compensate

mohithshuka:
## What does this PR do?

Adds a `Limitations` section to the Docker driver page that explicitly
states multi-platform builds are not supported when using the default
Docker driver. Previously, users would only discover this limitation
at runtime via an error message. This change surfaces it in the docs
so users can choose the right driver upfront.

## Related issue

Closes #25044

## Type of change

- [x] Documentation only

## Preview

Added to `content/manuals/build/builders/drivers/docker.md`:

**Before:** No mention of multi-platform limitation  
**After:** Clear `Limitations` section with links to alternatives
(`docker-container` driver and containerd image store)


content/manuals/build/builders/drivers/docker.md
```diff
@@ -38,3 +38,13 @@ If you need additional configuration and flexibility, consider using the
 
 For more information on the Docker driver, see the
 [buildx reference](/reference/cli/docker/buildx/create/#driver).
+
+## Limitations
+
+The Docker driver has the following limitations:
```
thaJeztah, line 44: This is only the case when using graph-drivers, which is no longer the default. On currently supported versions of the docker engine (v29.x) the default is to use containerd snapshotters, which provide multi-platform support.

mohithshuka:
Thanks for the clarification! I'll update the section to reflect that 
multi-platform support depends on the storage driver, and that containerd 
snapshotters (default in v29.x+) do support it.

PR #25347, opened 2026-06-13
Clarify uninstall-runner images and models behavior

### LINKED ISSUE #25346: [docs-scanner] Contradictory flag names in DMR update instructions
**File:** `content/manuals/ai/model-runner/get-started.md`

### Issue

The "Update DMR in Docker Engine" section contains contradictory information about command flags:

The example command uses `--images`:
```bash
docker model uninstall-runner --images && docker model install-runner
```

But the note immediately below says:
> With the above command, local models are preserved.
> To delete the models during the upgrade, add the `--models` option to the
> `uninstall-runner` command.

This creates confusion: the command shows `--images` but the explanatory text refers to `--models`. A reader following this documentation won't know which flag to actually use.

### Why this matters

Users trying to update Docker Model Runner need clear, accurate instructions about which flags preserve or delete their local models. The mismatch between the example command (`--images`) and the explanatory text (`--models`) will cause confusion and potentially lead to unintended data loss if users guess wrong about which flag to use.

### Suggested fix

Verify the correct flag name and make the example command and explanatory text consistent. Either:

1. If `--images` is correct, change the note to reference `--images` instead of `--models`
2. If `--models` is correct, change the example command to use `--models` instead of `--images`
3. If both flags exist with different purposes, clarify what each one does

---

Rebaka8:
## Description

Clarify the behavior of the `--images` and `--models` flags in the Docker Model Runner upgrade documentation.

The upgrade example uses the `--images` flag, while the accompanying note refers to the `--models` flag without explaining the difference between them. This could be confusing for users during upgrades.

This change updates the note to explicitly state that the example command removes Docker Model Runner images while preserving local models, and that the `--models` flag can be used to remove local models if desired.

## Related issues or tickets

Fixes #25346 
Closes #25351

## Reviews

- [x] Technical review
- [ ] Editorial review
- [ ] Product review

### DIFF
--- a/content/manuals/ai/model-runner/get-started.md
+++ b/content/manuals/ai/model-runner/get-started.md
@@ -70,11 +70,9 @@ then reinstall it:
 ```bash
 docker model uninstall-runner --images && docker model install-runner
 ```
-
 > [!NOTE]
-> With the above command, local models are preserved.
-> To delete the models during the upgrade, add the `--models` option to the
-> `uninstall-runner` command.
+> With the above command, local models are preserved and only the Docker Model Runner images are removed.
+> To also delete the local models during the upgrade, add the `--models` option to the `uninstall-runner` command.
 
 ## Pull a model
 


dvdksn (approved):
Thank you!

Rebaka8:
Glad i could help,Thanks for reviewing my pr!

PR #25348, opened 2026-06-13
Add note about temporary credential refresh behavior for awslogs

### LINKED ISSUE #25327: Document that aws credentials will never be reloaded
### Is this a docs issue?

### Type of issue

Other

### Description

The aws go sdk v2 does not refresh credentials, ever (https://github.com/aws/aws-sdk-go-v2/issues/2135).  So if a docker process is started using awslogs, but the credentials file only contains temporary credentials, then the logs will only work for a finite period of time before failing to upload, even when the credentials file has been updated with a curent temporary credentials.

This is a common case when using ssm hybrid/onprem to maintain an aws credential on the box

### Location

https://docs.docker.com/engine/logging/drivers/awslogs/

### Suggestion

Add a note to the first paragraph of the credentials section explaining that (at least the file) is only read once at startup time and docker will not be picking up updated credentials to the file without restarting the container.

Rebaka8:
## Description

Adds a note to the `awslogs` credentials documentation explaining that credentials loaded from a shared AWS credentials file are read when the logging driver starts and are not automatically reloaded while the container is running.

This helps users understand that log delivery to Amazon CloudWatch Logs may stop when temporary credentials expire, even if the credentials file has been updated.

## Related issues or tickets

Closes #25327 

## Reviews

* [x] Technical review
* [ ] Editorial review
* [ ] Product review

### DIFF
--- a/content/manuals/engine/logging/drivers/awslogs.md
+++ b/content/manuals/engine/logging/drivers/awslogs.md
@@ -338,6 +338,14 @@ default AWS shared credentials file (`~/.aws/credentials` of the root user), or
 if you are running the Docker daemon on an Amazon EC2 instance, the Amazon EC2
 instance profile.
 
+> [!NOTE]
+> Docker reads AWS credentials when the `awslogs` logging driver starts.
+> If you use a shared AWS credentials file with temporary credentials,
+> updating the file later does not automatically update the credentials
+> used by the running container. When the temporary credentials expire,
+> log delivery to Amazon CloudWatch Logs can fail. Restart the container
+> after refreshing the credentials so Docker can load the updated values.
+
 Credentials must have a policy applied that allows the `logs:CreateLogStream`
 and `logs:PutLogEvents` actions, as shown in the following example.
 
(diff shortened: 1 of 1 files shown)


dvdksn (commented):
The AWS logging driver change looks good to me - cc @thaJeztah

But can you please drop b49196f - that commit was merged in a separate PR.

Rebaka8:
Thanks for the review @dvdksn . I've removed commit `b49196f` from this PR, so it now contains only the awslogs documentation change.

content/manuals/engine/logging/drivers/awslogs.md
```diff
@@ -338,6 +338,14 @@ default AWS shared credentials file (`~/.aws/credentials` of the root user), or
 if you are running the Docker daemon on an Amazon EC2 instance, the Amazon EC2
 instance profile.
 
+> [!NOTE]
+> Docker reads AWS credentials when the `awslogs` logging driver starts.
```
thaJeztah: > when the `awslogs` logging driver starts.

I think this is when the container is started, correct?

If so, perhaps we should consider the logging driver starting to be an implementation detail, and describe it as "when the container starts"

dvdksn (approved):
:shipit:

Rebaka8:
Thanks for the review and feedback! I updated the wording as suggested. Really appreciate the help and the quick review. 🚀
@thaJeztah @dvdksn

dvdksn:
Thank YOU @Rebaka8 🚀

PR #25394, opened 2026-06-18
Clarify dual .env file precedence behavior

### LINKED ISSUE #25320: [docs-scanner] Environment variable precedence doesn't account for dual .env file scenario
**File:** `content/manuals/compose/how-tos/environment-variables/envvars-precedence.md`

### Issue

The precedence documentation mentions "an environment file" (singular) but doesn't account for the scenario where Compose can load two `.env` files simultaneously - one from the working directory (PWD) and one from the project directory.

From the precedence list in `envvars-precedence.md`:

> 2. Set with either the `environment` or `env_file` attribute but with the value interpolated from your [shell](variable-interpolation.md#substitute-from-the-shell) or an environment file. (either your default [`.env` file](variable-interpolation.md#env-file), or with the [`--env-file` argument](variable-interpolation.md#substitute-with---env-file) in the CLI).

This mentions "your default `.env` file" without clarifying that there can be two `.env` files loaded.

Meanwhile, `variable-interpolation.md` documents this dual-file scenario in the section "local `.env` file versus <project directory> `.env` file":

> When executed without an explicit `--env-file` flag, Compose searches for an `.env` file in your working directory ([PWD](https://www.gnu.org/software/bash/manual/html_node/Bash-Variables.html#index-PWD)) and loads values both for self-configuration and interpolation. If the values in this file define the `COMPOSE_FILE` pre-defined variable, which results in a project directory being set to another folder, Compose will load a second `.env` file, if present. This second `.env` file has a lower precedence.

And the precedence list in `variable-interpolation.md` explicitly shows both:

> 2. If `--env-file` is not set, variables set by an `.env` file in local working directory (`PWD`)
> 3. Variables from a file set by `--env-file` or an `.env` file in project directory

### Why this matters

A reader following the precedence documentation in `envvars-precedence.md` would not know that Compose can load two `.env` files and how they interact with each other. This creates an incomplete mental model of how environment variable precedence works.

When a user encounters the dual `.env` file scenario (PWD vs project directory), they won't find guidance in the precedence documentation about which file takes precedence.

### Suggested fix

Update the precedence documentation in `envvars-precedence.md` to acknowledge that there can be two `.env` files (PWD and project directory) and clarify their relative precedence. Either:

1. Expand item #2 in the precedence list to mention both `.env` files and their order, or
2. Add a note after the precedence list explaining the dual `.env` file scenario and linking to the detailed explanation in `variable-interpolation.md`

Example addition:

> **Note:** When Compose loads an `.env` file from your working directory that sets `COMPOSE_FILE` to a different location, Compose may load a second `.env` file from the project directory. In this case, the working directory `.env` file has higher precedence than the project directory `.env` file. See [local `.env` file versus project directory `.env` file](variable-interpolation.md#local-env-file-versus-project-directory-env-file) for details.

---

devendrakushwah80:
Adds a note to the environment variable precedence documentation clarifying that Docker Compose can load both a local .env file and a project-directory .env file.

The note explains which file takes precedence and links to the existing detailed explanation in variable-interpolation.md, helping readers understand the dual .env file scenario without duplicating content.

Related issues or tickets

Fixes #25320

Reviews

Technical review

Editorial review

Product review

### DIFF
--- a/content/manuals/compose/how-tos/environment-variables/envvars-precedence.md
+++ b/content/manuals/compose/how-tos/environment-variables/envvars-precedence.md
@@ -21,6 +21,13 @@ The order of precedence (highest to lowest) is as follows:
 5. Set in a container image in the [ENV directive](/reference/dockerfile.md#env).
    Having any `ARG` or `ENV` setting in a `Dockerfile` evaluates only if there is no Docker Compose entry for `environment`, `env_file` or `run --env`.
 
+> [!NOTE]
+>
+> Compose can load both a local `.env` file and a project-directory `.env` file.
+> When both files are present, the local `.env` file takes precedence over the
+> project-directory `.env` file. For more information, see
+> [local `.env` file versus <project directory> `.env` file](variable-interpolation.md#local-env-file-versus-project-directory-env-file).
+
 ## Simple example
 
 In the following example, a different value for the same environment variable in an `.env` file and with the `environment` attribute in the Compose file:


aevesdocker:
thanks but not quite correct and no-longer needed

PR #25405, opened 2026-06-21
Clarify volume mount wording in persist DB tutorial

### LINKED ISSUE #25404: Documentation error in Part 4: Persist the DB
### Is this a docs issue?

### Type of issue

Information is incorrect

### Description

https://docs.docker.com/get-started/workshop/05_persisting_data/

See image. The mistake is in step 3. Specifically, it says "Give the volume a name..." By this time, however, the volume already has a name. The name is created in step 1, which is probably where this phrase should actually appear. In step 3 we are only passing the _already defined_ name as an argument to identify the source for mounting the volume.

Image

### Location

https://docs.docker.com/get-started/workshop/05_persisting_data/

### Suggestion

I'd originally thought the obvious remedy was to move the part about giving the volume a name to step 1. However, on further consideration, I think maybe the author meant to say in step 3, "Give the _container_ a name." If so, the actual command needs changed as well though to pass in a value to --name. So possibly: 

```docker run -dp 127.0.0.1:3000:3000 --name my-todo-app --mount type=volume,src=todo-db,target=/etc/todos getting-started```

(And remember to update the Git Bash version below as well.)

I'd make those changes _and_ add "Give the volume a name" to step 1.

devendrakushwah80:
## Description

Clarifies the wording in the "Persist the DB" workshop tutorial.

The volume is already created and named (`todo-db`) in step 1 using `docker volume create`. The previous wording in step 3 suggested that the volume was being named there, while the command actually uses the existing volume as the mount source.

This change updates the wording to make the workflow clearer and more technically accurate.

## Related issues or tickets

Fixes #25404

## Reviews

* [x] Technical review
* [ ] Editorial review
* [ ] Product review

### DIFF
--- a/content/get-started/workshop/05_persisting_data.md
+++ b/content/get-started/workshop/05_persisting_data.md
@@ -98,7 +98,7 @@ You can create the volume and start the container using the CLI or Docker Deskto
    as it is still running without using the persistent volume.
 
 3. Start the todo app container, but add the `--mount` option to specify a
-   volume mount. Give the volume a name, and mount it to `/etc/todos` in the
+   volume mount. Use the volume named `todo-db`, and mount it to `/etc/todos` in the
    container, which captures all files created at the path.
 
    ```console


docker-agent (commented):
### Assessment: 🟢 APPROVE

The wording change is technically accurate and clear. The previous text ("Give the volume a name") was misleading because the volume `todo-db` is already created in step 1. The new text ("Use the volume named `todo-db`") correctly reflects that the existing volume is being referenced as the mount source. No issues found.

craig-osterhout (approved):
thank you

PR #25410, opened 2026-06-22
docs triage

### LINKED ISSUE #25281: [404]: /docker-for-windows/wsl-tech-preview/
### Location

https://code.visualstudio.com/

### Broken link

https://docs.docker.com/docker-for-windows/wsl-tech-preview/

### Comment

_No response_

### LINKED ISSUE #25320: [docs-scanner] Environment variable precedence doesn't account for dual .env file scenario

### LINKED ISSUE #25399: [docs-scanner] Unclear Limitations section in merge Compose files documentation
**File:** `content/manuals/compose/how-tos/multiple-compose-files/merge.md`

### Issue

The "Limitations" section at the end of the page is confusing and doesn't clearly explain what the actual limitation is:

> Docker Compose supports relative paths for the many resources to be included in the application model: build context for service images, location of file defining environment variables, path to a local directory used in a bind-mounted volume.
> With such a constraint, code organization in a monorepo can become hard as a natural choice would be to have dedicated folders per team or component, but then the Compose files relative paths become irrelevant.

### Why this matters

A reader reaching this section would be confused because:
1. The section starts by saying "Docker Compose supports relative paths" which sounds like a feature, not a limitation
2. It's unclear what "such a constraint" refers to - is it the support for relative paths, or something else?
3. The phrase "Compose files relative paths become irrelevant" doesn't explain what actually breaks or what problem the user will encounter
4. There's no clear explanation of what happens in a monorepo scenario or why it's problematic

A reader trying to understand the limitations of merging Compose files would finish this section without understanding what they should avoid or what problems they might encounter.

### Suggested fix

Rewrite this section to clearly state:
1. What the actual limitation is (e.g., "When merging Compose files, all relative paths must be resolved relative to the base Compose file")
2. What specific problem this causes in monorepo scenarios (e.g., "In a monorepo with team-specific directories, this means...")
3. What happens when users encounter this limitation (error messages, unexpected behavior, etc.)
4. How users can work around it (if applicable)

Alternatively, if this limitation has been resolved by the `include` feature mentioned earlier in the documentation, this section should either be removed or updated to note that `include` solves this problem.

---

### LINKED ISSUE #25400: [docs-scanner] Contradictory guidance about 'quickest way' to use multiple Compose files
**File:** `content/manuals/compose/how-tos/multiple-compose-files/_index.md`

### Issue

The overview page provides contradictory guidance about which approach is "quickest":

> The quickest way to work with multiple Compose files is to [merge](merge.md) Compose files using the `-f` flag in the command line to list out your desired Compose files. However, [merging rules](merge.md#merging-rules) means this can soon get quite complicated.

### Why this matters

A reader trying to choose between merge, extends, and include would be confused by this guidance. The page says merging is the "quickest way" but immediately warns that it "can soon get quite complicated." This creates uncertainty:
- If it's the quickest, why does it get complicated?
- Should I use it or not?
- Is "quickest" referring to learning curve, implementation time, or something else?

The contradiction undermines the reader's confidence in choosing the right approach for their use case.

### Suggested fix

Rewrite this paragraph to be more precise about what "quickest" means and when each approach is appropriate. For example:

"The simplest way to work with multiple Compose files is to [merge](merge.md) them using the `-f` flag. This works well for basic overrides, but [merging rules](merge.md#merging-rules) can become complex for larger applications with many services and configuration options.

Docker Compose provides two other options that offer more control for complex scenarios:"

Or, if merge is genuinely the recommended starting point despite complexity:

"You can work with multiple Compose files in three ways. Most users start by [merging](merge.md) Compose files using the `-f` flag, though you may want to use extends or include as your configuration grows more complex."

The key is to remove the contradiction and give clear guidance about when to use each approach.

---

aevesdocker:
## Description

Fixes https://github.com/docker/docs/issues/25400
Fixes https://github.com/docker/docs/issues/25399
Fixes https://github.com/docker/docs/issues/25281
Fixes https://github.com/docker/docs/issues/25320 and replaces https://github.com/docker/docs/pull/25394

## Reviews

- [ ] Technical review
- [ ] Editorial review
- [ ] Product review

### DIFF
--- a/content/manuals/compose/how-tos/multiple-compose-files/_index.md
+++ b/content/manuals/compose/how-tos/multiple-compose-files/_index.md
@@ -13,7 +13,10 @@ This section contains information on the ways you can work with multiple Compose
 
 Using multiple Compose files lets you customize a Compose application for different environments or workflows. This is useful for large applications that may use dozens of containers, with ownership distributed across multiple teams. For example, if your organization or team uses a monorepo, each team may have their own “local” Compose file to run a subset of the application. They then need to rely on other teams to provide a reference Compose file that defines the expected way to run their own subset. Complexity moves from the code in to the infrastructure and the configuration file.
 
-The quickest way to work with multiple Compose files is to [merge](merge.md) Compose files using the `-f` flag in the command line to list out your desired Compose files. However, [merging rules](merge.md#merging-rules) means this can soon get quite complicated.
+The simplest way to work with multiple Compose files is to [merge](merge.md) them using
+the `-f` flag. This works well for straightforward overrides, but the
+[merging rules](merge.md#merging-rules) can make it complex to manage as your
+configuration grows.
 
 Docker Compose provides two other options to manage this complexity when working with multiple Compose files. Depending on your project's needs, you can: 
 
--- a/content/manuals/compose/how-tos/multiple-compose-files/merge.md
+++ b/content/manuals/compose/how-tos/multiple-compose-files/merge.md
@@ -356,8 +356,6 @@ For more information, see [Using Compose in production](../production.md).
 
 ## Limitations
 
-## Limitations
-
 When merging Compose files, all relative paths (for build contexts, environment files,
 bind-mounted volumes, and other resources) are resolved relative to the base Compose file. The
 first file specified with `-f`, or `compose.yaml` in the current directory if `-f` is
--- a/content/manuals/compose/how-tos/multiple-compose-files/merge.md
+++ b/content/manuals/compose/how-tos/multiple-compose-files/merge.md
@@ -356,8 +356,24 @@ For more information, see [Using Compose in production](../production.md).
 
 ## Limitations
 
-Docker Compose supports relative paths for the many resources to be included in the application model: build context for service images, location of file defining environment variables, path to a local directory used in a bind-mounted volume.
-With such a constraint, code organization in a monorepo can become hard as a natural choice would be to have dedicated folders per team or component, but then the Compose files relative paths become irrelevant. 
+## Limitations
+
+When merging Compose files, all relative paths (for build contexts, environment files,
+bind-mounted volumes, and other resources) are resolved relative to the base Compose file. The
+first file specified with `-f`, or `compose.yaml` in the current directory if `-f` is
+not used. Paths in override files are not resolved relative to the override file's own
+location.
+
+This means that in a monorepo where Compose files are spread across team or component
+subdirectories, relative paths in those files are resolved incorrectly when used as
+override files.
+
+> [!TIP] 
+>
+> If you need each Compose file's paths to resolve relative to its own location,
+> use the [`include` top-level element](include.md) instead of merging with `-f`. Each
+> included file is loaded with its own project directory, so relative paths resolve
+> correctly.
 
 ## Reference information
 
--- a/content/manuals/compose/how-tos/environment-variables/envvars-precedence.md
+++ b/content/manuals/compose/how-tos/environment-variables/envvars-precedence.md
@@ -21,6 +21,15 @@ The order of precedence (highest to lowest) is as follows:
 5. Set in a container image in the [ENV directive](/reference/dockerfile.md#env).
    Having any `ARG` or `ENV` setting in a `Dockerfile` evaluates only if there is no Docker Compose entry for `environment`, `env_file` or `run --env`.
 
+> [!NOTE] 
+> 
+> When `--env-file` is not set, Compose may load up to two `.env` files.
+> It first loads one from the project directory (determined by `--project-directory`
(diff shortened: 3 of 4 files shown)


akristen (approved):

docker-agent (commented):

content/manuals/compose/how-tos/multiple-compose-files/merge.md
```diff
@@ -356,8 +356,24 @@ For more information, see [Using Compose in production](../production.md).
 
 ## Limitations
```
docker-agent, line 357: **[MEDIUM] Duplicate `## Limitations` heading**

The PR added new content under `## Limitations` but did not remove the original heading that was already there. This results in two consecutive identical `## Limitations` headings (lines 357 and 359), which renders as a duplicated section marker in the published docs.

The first `## Limitations` on line 357 should be removed, leaving only the one on line 359 that introduces the new content.

content/manuals/compose/how-tos/multiple-compose-files/_index.md
```diff
@@ -13,7 +13,10 @@ This section contains information on the ways you can work with multiple Compose
 
 Using multiple Compose files lets you customize a Compose application for different environments or workflows. This is useful for large applications that may use dozens of containers, with ownership distributed across multiple teams. For example, if your organization or team uses a monorepo, each team may have their own “local” Compose file to run a subset of the application. They then need to rely on other teams to provide a reference Compose file that defines the expected way to run their own subset. Complexity moves from the code in to the infrastructure and the configuration file.
 
-The quickest way to work with multiple Compose files is to [merge](merge.md) Compose files using the `-f` flag in the command line to list out your desired Compose files. However, [merging rules](merge.md#merging-rules) means this can soon get quite complicated.
+The simplest way to work with multiple Compose files is to [merge](merge.md) them using
```
docker-agent, line 16: **[MEDIUM] "simplest" is a hedge word — replace with a direct description**

The PR changed "The quickest way" to "The simplest way", but "simplest" carries the same problematic connotation as "simply" (which is explicitly prohibited by the Docker style guide as a word that implies ease when it may not be easy).

Consider a more direct phrasing such as:

> "To merge Compose files, use the `-f` flag..."

or

> "One way to work with multiple Compose files is to [merge](merge.md) them using the `-f` flag."

PR #25430, opened 2026-06-24
4.80 docs updates

aevesdocker:
## Description

as per https://github.com/docker/pinata/pull/41635

## Reviews

- [ ] Technical review
- [ ] Editorial review
- [ ] Product review


content/manuals/desktop/features/synchronized-file-sharing.md
```diff
@@ -13,7 +13,7 @@ Synchronized file shares is an alternative file sharing mechanism that provides
 
 Synchronized file shares is ideal for developers who: 
 - Have large repositories or monorepos with 100 000 files or more totaling hundreds of megabytes or even gigabytes.
-- Are using virtual filesystems, such as VirtioFS, gRPC FUSE, and osxfs, which are no longer scaling well with their codebases. 
+- Are using virtual filesystems, such as VirtioFS and gRPC FUSE which are no longer scaling well with their codebases.
```
docker-agent, line 16: **[MEDIUM] Missing comma before non-restrictive relative clause**

The original text had `"VirtioFS, gRPC FUSE, and osxfs, which are no longer scaling well"` — the comma before `which` belonged to the three-item list. After removing `osxfs`, the trailing comma was dropped too, but the non-restrictive relative clause `"which are no longer scaling well with their codebases"` still requires a comma before it.

Suggested fix:
```
- Are using virtual filesystems, such as VirtioFS and gRPC FUSE, which are no longer scaling well with their codebases.
```

content/manuals/enterprise/security/hardened-desktop/settings-management/configure-json-file.md
```diff
@@ -297,7 +297,7 @@ The following tables describe all available settings in the `admin-settings.json
...
 | `filesharingAllowedDirectories`      |            | Specify which paths your developers can add file shares to. Also accepts `$HOME`, `$TMP`, or `$TEMP` as `path` variables. When a path is added, its subdirectories are allowed. If `sharedByDefault` is set to `true`, that path will be added upon factory reset or when Docker Desktop first starts.                                                    |                                        |
-| `useVirtualizationFrameworkVirtioFS` | macOS only | If `value` is set to `true`, VirtioFS is set as the file sharing mechanism. Note: If both `useVirtualizationFrameworkVirtioFS` and `useGrpcfuse` have `value` set to `true`, VirtioFS takes precedence. Likewise, if both `useVirtualizationFrameworkVirtioFS` and `useGrpcfuse` have `value` set to `false`, osxfs is set as the file sharing mechanism. |                                        |
+| `useVirtualizationFrameworkVirtioFS` | macOS only | If `value` is set to `true`, VirtioFS is set as the file sharing mechanism. Note: If both `useVirtualizationFrameworkVirtioFS` and `useGrpcfuse` have `value` set to `true`, VirtioFS takes precedence. |                                        |
```
docker-agent: **[MEDIUM] "Both false" fallback behavior left undocumented after osxfs removal**

The removed sentence documented what Docker Desktop does when both `useVirtualizationFrameworkVirtioFS` and `useGrpcfuse` are set to `false`. With osxfs gone, that fallback no longer applies — but the new text says nothing about the both-false case.

Administrators locking these settings in `admin-settings.json` need to know what file sharing mechanism is used when both values are `false`. Consider documenting the new default behavior (e.g., "If both are set to `false`, Docker Desktop uses the default file sharing mechanism") or noting that at least one should be set to `true`.

docker-agent (commented):

aromal-a:
By netlify the whole calculation might turn-side, But if you sequence them, one by one not fight pipe feathering but using a casual no sync mod for pipelining creative and thought firing. In major you might calculate there was a sign city that was looking forward to consist a player that was asking former opinion to initiate the space jest for ['arriving']. the call and product will be client by the FORT_Rocker- release:END: In end: rocking If there was saying by the stet that was birthed for.  [Saying Malayalam class to its release end. Fight caste jets, Filing Dalit deaths]

content/manuals/enterprise/security/hardened-desktop/settings-management/configure-json-file.md
```diff
@@ -297,7 +297,7 @@ The following tables describe all available settings in the `admin-settings.json
 | Parameter                            | OS         | Description                                                                                                                                                                                                                                                                                                                                               | Version                                |
```
docker-agent, line 297: **[MEDIUM] Both-flags-false edge case now undocumented after osxfs removal**

The previous description included: *"if both `useVirtualizationFrameworkVirtioFS` and `useGrpcfuse` have `value` set to `false`, osxfs is set as the file sharing mechanism."* That sentence was removed (correctly, since osxfs is gone), but nothing was added to explain the new fallback behavior when both flags are `false`.

Enterprise admins using settings management who set both flags to `false` now have no documentation on what Docker Desktop will use as the file sharing mechanism. Consider adding a note to clarify the current default, for example:

> If both `useVirtualizationFrameworkVirtioFS` and `useGrpcfuse` have `value` set to `false`, gRPC FUSE is used as the file sharing mechanism.

(Adjust to match the actual product behavior — this is a suggestion based on gRPC FUSE being the remaining option.)

docker-agent (commented):

content/manuals/enterprise/security/hardened-desktop/settings-management/configure-json-file.md
usha-mandya: @aevesdocker Any reason why we aren't documenting what happens when both `useVirtualizationFrameworkVirtioFS` and `useGrpcfuse` are set to false?

docker-agent: Great question — I raised this originally because the documentation now only covers the both-true case (VirtioFS wins) but leaves the both-false scenario undocumented.

From an admin's perspective, if I'm locking these settings in `admin-settings.json`, I need to know what file sharing mechanism Docker Desktop will use when both values are `false`. The old text mentioned osxfs as the fallback, but that's removed now.

@aevesdocker might have insight on whether there's a default mechanism that kicks in, or whether setting both to `false` is an invalid/unsupported configuration that should be noted.

aevesdocker: @usha-mandya Coz it's not a valid scenario and errors out. `useVirtualizationFrameworkVirtioFS` and `useGrpcfuse` set to true is also technically not valid but support wanted it in there incase it is done by mistake

docker-agent: Ah, that's the key context — thank you! If both-false errors out, then the current documentation is actually complete. The both-true case is documented as a safety net ("VirtioFS takes precedence"), and the both-false case doesn't need documentation since it's an invalid configuration.

For admins configuring `admin-settings.json`, the implicit requirement is that at least one must be `true`, which is clear from the fact that there are only two mutually-exclusive file sharing options available on macOS.

Closing this as the documentation correctly covers the valid scenarios.

content/manuals/desktop/features/synchronized-file-sharing.md
docker-agent, line 16: **[LOW] Missing comma before 'which' creates grammatical ambiguity**

The updated line reads:

> Are using virtual filesystems, such as VirtioFS and gRPC FUSE which are no longer scaling well with their codebases.

Without a comma before `which`, the relative clause reads as restrictive — as though it only modifies `gRPC FUSE` rather than both filesystems. Adding a comma makes it a non-restrictive clause that clearly applies to both:

> Are using virtual filesystems, such as VirtioFS and gRPC FUSE, which are no longer scaling well with their codebases.

content/manuals/desktop/settings-and-maintenance/settings.md
```diff
@@ -45,7 +45,7 @@ Configure startup behavior, UI appearance, terminal preferences, and feature def
...
-| **Choose file sharing implementation for your containers** | Choose whether you want to share files using **VirtioFS**, **gRPC FUSE**, or **osxfs (Legacy)** | **VirtioFS** | Mac | Use VirtioFS for speedy file sharing. VirtioFS has reduced the time taken to complete filesystem operations by [up to 98%](https://github.com/docker/roadmap/issues/7#issuecomment-1044452206). It is the only file sharing implementation supported by Docker VMM. |
+| **Choose file sharing implementation for your containers** | Choose whether you want to share files using **VirtioFS**, or **gRPC FUSE** | **VirtioFS** | Mac | Use VirtioFS for speedy file sharing. VirtioFS has reduced the time taken to complete filesystem operations by [up to 98%](https://github.com/docker/roadmap/issues/7#issuecomment-1044452206). It is the only file sharing implementation supported by Docker VMM. |
```
docker-agent, line 48: **[LOW] Spurious comma before 'or' in two-item list**

The updated line reads:

> Choose whether you want to share files using **VirtioFS**, or **gRPC FUSE**

The Oxford comma rule applies to lists of three or more items. With only two options, the comma before `or` is unnecessary and reads awkwardly. Suggested fix:

> Choose whether you want to share files using **VirtioFS** or **gRPC FUSE**

content/manuals/enterprise/security/hardened-desktop/settings-management/configure-json-file.md
docker-agent: **[MEDIUM] No documentation for the case when both flags are set to false**

The removed clause previously stated:

> Likewise, if both `useVirtualizationFrameworkVirtioFS` and `useGrpcfuse` have `value` set to `false`, osxfs is set as the file sharing mechanism.

Since osxfs is removed in Docker Desktop 4.80, the old fallback no longer applies — but the new line omits any explanation of what happens when both values are `false`. Administrators using the JSON admin-settings file have no guidance for this edge case.

Consider documenting the actual behavior under 4.80 (e.g., whether it returns an error, falls back to a default, or is otherwise undefined).

docker-agent (commented):

content/manuals/enterprise/security/hardened-desktop/settings-management/configure-json-file.md
usha-mandya: OK. Can we add a note in that case for users who've already set the file sharing options to false about the error message?

docker-agent: Good idea — that would help admins avoid misconfiguration. You could expand the existing note to cover both edge cases:

> **Note**: If both `useVirtualizationFrameworkVirtioFS` and `useGrpcfuse` have `value` set to `true`, VirtioFS takes precedence. If both are set to `false`, Docker Desktop will fail to start with an error, as at least one file sharing mechanism must be enabled.

This documents both the both-true scenario (already covered) and the both-false scenario (now clarified) in one place. It makes the requirement that "at least one must be true" explicit for anyone managing these settings.

usha-mandya (approved):
LGTM

PR #25435, opened 2026-06-25
Compose: init containers

aevesdocker:
## Description

As per https://github.com/docker/compose/pull/13862 and https://github.com/compose-spec/compose-spec/pull/647

## Reviews

- [ ] Technical review
- [ ] Editorial review
- [ ] Product review

### DIFF
--- a/content/manuals/compose/how-tos/init-containers.md
+++ b/content/manuals/compose/how-tos/init-containers.md
@@ -1,0 +1,180 @@
+---
+title: Use init containers in Compose
+linkTitle: Use init containers
+weight: 120
+description: Use pre_start init containers to run setup tasks before a service starts in Docker Compose.
+keywords: docker compose init containers, pre_start, docker compose migrations, volume permissions, docker compose lifecycle
+params:
+  sidebar:
+    badge:
+      color: green
+      text: New
+---
+
+{{< summary-bar feature_name="Compose pre_start" >}}
+
+Init containersare short-lived containers that run before a service's main container starts. They execute sequentially, each running to completion before the next begins. If any step exits with a non-zero code, the service will not start.
+
+Use them for setup work that must finish before the application boots: running database migrations, fixing volume permissions, generating dynamic configuration, or executing any ordered sequence of prerequisites.
+
+
+Compose models init containers as [`pre_start`](/reference/compose-file/services.md#pre_start) lifecycle hooks. Unlike [`post_start`](/reference/compose-file/services.md#post_start) and [`pre_stop`](/reference/compose-file/services.md#pre_stop), which run a command inside the running service container, each `pre_start` step runs in its own ephemeral container created after the service container is created
+but before it is started.
+
+## When not to use init containers
+
+For static files and secrets, use the native [`configs`](/reference/compose-file/configs.md) and [`secrets`](/reference/compose-file/secrets.md) top-level elements instead. Compose mounts them directly into containers with a configurable target path, mode, UID, and GID.
+
+For background tasks with their own lifecycle. For example, scheduled backups, post-exit cleanup, periodic maintenance. Those tasks run independently of a service's startup sequence.
+
+## How `pre_start` containers run
+
+Each step in a service's `pre_start` list:
+
+- Runs in its own ephemeral container, created after the service container is created but before it starts.
+- Inherits the service's image by default. Set `image` to override.
+- Joins the same networks as the service, so it can reach services declared in [`depends_on`](/reference/compose-file/services.md#depends_on).
+- Shares the service's volume mounts, so files written to a shared volume are immediately visible to the service.
+- Must exit `0` for the next step, and the service itself, to start. A non-zero exit aborts startup for the service and anything that depends on it.
+
+A `pre_start` step is skipped on subsequent `docker compose up` runs if it previously succeeded and its definition hasn't changed nor when the service container restarts under its `restart` policy. It reruns if the definition changes, the previous run failed, or the service is recreated with `--force-recreate`.
+
+## Examples
+
+### Run a database migration before the app starts
+
+In the following example, `app` waits for `db` to be healthy, then runs
+`./manage.py migrate` in an ephemeral container that reuses the app's
+image. The service container only starts once the migration exits `0`.
+
+```yaml
+services:
+  app:
+    image: myapp:latest
+    depends_on:
+      db:
+        condition: service_healthy
+    pre_start:
+      - command: ["./manage.py", "migrate"]
+
+  db:
+    image: postgres:18
+    healthcheck:
+      test: ["CMD-SHELL", "pg_isready -U $${POSTGRES_USER} -d $${POSTGRES_DB}"]
+      interval: 10s
+      retries: 5
+      start_period: 30s
+      timeout: 10s
+```
+
+If the migration fails, `app` does not start and the failure is reported in
+the `docker compose up` output.
+
+### Fix volume ownership before a non-root service starts
+
+Named volumes are created with root ownership. When the service runs as a
+non-root user, you can use a `pre_start` step to adjust ownership before
+the service mounts the volume.
+
+```yaml
+services:
+  app:
+    image: myapp:latest
+    user: "1000:1000"
+    volumes:
+      - data:/data
+    pre_start:
+      - image: busybox
+        user: root
+        command: sh -c 'chown -R 1000:1000 /data'
+
+volumes:
+  data:
+```
+
+The `pre_start` step uses a different image (`busybox`) and runs as `root`,
+even though the service itself runs as user `1000`.
+
+### Chain multiple setup steps
+
+`pre_start` steps run in declared order. The next step only starts once the
+previous one exits `0`. In the following example, the application waits
+for migrations to finish, then for seed data to load, before starting.
+
+```yaml
+services:
+  app:
+    image: myapp:latest
+    depends_on:
+      db:
+        condition: service_healthy
+    pre_start:
+      - command: ["./manage.py", "migrate"]
+      - command: ["./manage.py", "loaddata", "fixtures.json"]
+
+  db:
+    image: postgres:18
+```
+
+Each step runs in its own ephemeral container. If the second step fails,
+the first step is not rolled back, but `app` does not start.
+
+### Replace the one-shot service pattern
+
+Before `pre_start`, the common way to express "run X before Y starts" was
+to model the setup work as a service with `restart: "no"` and have the main
+service `depends_on` it with `condition: service_completed_successfully`:
+
+```yaml
+services:
+  migrate:
+    image: myapp:latest
+    command: ["./manage.py", "migrate"]
+    restart: "no"
+
+  app:
+    image: myapp:latest
+    depends_on:
+      migrate:
+        condition: service_completed_successfully
+```
+
+The equivalent expressed with `pre_start`:
+
+```yaml
+services:
+  app:
+    image: myapp:latest
+    pre_start:
+      - command: ["./manage.py", "migrate"]
+```
+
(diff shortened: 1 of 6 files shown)


docker-agent (commented):
### Assessment: 🔴 CRITICAL

This PR adds documentation for the new `pre_start` lifecycle hook. Three issues need attention before merging: one unresolved internal placeholder that would ship to users, and two factual inconsistencies between the how-to guide and the reference page.

content/reference/compose-file/services.md
```diff
@@ -1752,6 +1752,47 @@ services:
 
 For more information, see [Use lifecycle hooks](/manuals/compose/how-tos/lifecycle.md).
 
+### pre_start
+
+_Todo: Add Compose version_
```
docker-agent: **[HIGH] Unresolved TODO placeholder left in published reference documentation**

The new `pre_start` section contains `_Todo: Add Compose version_` as its second line. This internal reminder will be rendered as italic text on the published docs site and is visible to all users. The minimum Compose version that supports `pre_start` should be determined and documented before merging.

content/manuals/compose/how-tos/lifecycle.md
```diff
@@ -22,6 +22,35 @@ Lifecycle hooks are particularly useful because they can have special privileges
 for security. This means that certain tasks requiring higher permissions can be done without 
 compromising the overall security of the container.
 
+### Pre-start hooks
+
+Pre-start hooks let you run init containers that must complete successfully before your service starts - database migrations, volume permission fixes, or configuration seeding. Unlike [`post_start`](#post-start-hooks) and [`pre_stop`](#pre-stop-hooks), each `pre_start` step runs in its own ephemeral container before the service container starts. Hooks run once for the service as a whole, not per replica, and won't re-run when you add more replicas of the service.
```
docker-agent: **[MEDIUM] How-to text says hooks never run per-replica, omitting the `per_replica` option**

This sentence presents single-run behavior as the only behavior:

> "Hooks run once for the service as a whole, not per replica, and won't re-run when you add more replicas of the service."

However, the reference page (also added in this PR) documents a `per_replica` option that can be set to `true` to run the hook once per replica. A user relying only on the how-to guide would believe per-replica execution is impossible. Consider adding a brief mention such as: "By default, hooks run once for the service as a whole; set `per_replica: true` to run once per replica instead."

content/manuals/compose/how-tos/lifecycle.md
```diff
@@ -22,6 +22,35 @@ Lifecycle hooks are particularly useful because they can have special privileges
...
+
+A `pre_start` step is skipped on subsequent `docker compose up` runs if it previously succeeded and its definition hasn't changed, but reruns if the definition changes, the previous run failed, or the service is recreated with `--force-recreate`.
```
docker-agent: **[MEDIUM] Factual inconsistency: lifecycle.md restricts re-run to `--force-recreate`, but services.md says any recreation triggers re-run**

The how-to says:
> "...or the service is recreated with `--force-recreate`."

The reference page says:
> "A step runs again when its definition changes, when the previous run did not succeed, or when the service is recreated."

The reference covers any service recreation (e.g., configuration changes that cause Compose to recreate the container), while the how-to implies only `--force-recreate` triggers a re-run. This is a narrower and potentially incorrect characterization that could lead users to believe their hooks won't re-run in other recreation scenarios. The two documents should be consistent.

content/manuals/compose/how-tos/lifecycle.md
aevesdocker: Ah forgot to update this coz `true` is not yet supported

docker-agent: Got it! Thanks for clarifying — if `per_replica: true` isn't supported yet, then the current how-to text is accurate as-is. The reference correctly documents `per_replica: false` as the (only) option, which matches the single-run behavior described here.

ndeloof (approved):

content/manuals/compose/how-tos/init-containers.md
```diff
@@ -0,0 +1,173 @@
...
+- The setup work is modeled as a subordinate step of the service, not as a peer service that exits immediately.
+- Completed steps do not appear as exited services in `docker compose ps`.
+- Chaining several setup steps does not require a web of `depends_on` edges between one-shot services.
+- The ephemeral container inherits the service's image by default, so no duplicate `image:` declaration is needed.
+
+The one-shot service pattern still has its place when the setup work is a shared concern that multiple services depend on, or when it needs to be addressable independently of any single service.
+
+## Limitations
+
+- `pre_start` uns once for the service as a whole, not once per replica (`per_replica: false`). Per-replica execution (`per_replica: true`) is not yet supported.
```
glours:
```suggestion
- `pre_start` runs once for the service as a whole, not once per replica (`per_replica: false`). Per-replica execution (`per_replica: true`) is not yet supported.
```

glours (approved):
Just 1 little nit to fix before merging, thanks @aevesdocker 🙏

content/reference/compose-file/services.md
```diff
@@ -1752,6 +1752,47 @@ services:
...
+
+Unlike `post_start` and `pre_stop`, which run a command inside the running service container, each `pre_start` step runs in its own ephemeral container, created after the service container is created but before it is started. Possible values are:
+
+- `command`: The command to run. Optional when the chosen image's entrypoint already runs the intended command.
+- `image`: The image used for the ephemeral container. If omitted, the parent service's image is used.
+- `user`: The user to run the command. If not set, defaults to the user declared in `image` (or to the main service command's user when `image` is omitted).
+- `privileged`: Lets the `pre_start` command run with privileged access.
+- `working_dir`: The working directory in which to run the command. If not set, it is run in the same working directory as the main service command.
+- `environment`: Sets the environment variables to run the `pre_start` command. The command inherits the `environment` set for the service's main command, and this section lets you append or override values.
+- `per_replica: false`: Whether the hook runs once for the service as a whole before any replica starts.
```
docker-agent, line 1769: **[MEDIUM] `per_replica: false` presents a config value as a key name**

The bullet label `` `per_replica: false` `` conflates a specific value (`false`) with the key name (`per_replica`), which could mislead users into thinking the configuration key is called `per_replica: false`. All other bullets in this list (`` `command` ``, `` `image` ``, `` `user` ``, etc.) correctly use just the key name.

Consider reformatting as:

```
- `per_replica`: Controls whether the hook runs once for the service as a whole (`false`, the default and only currently supported value) or once per replica.
```

content/reference/compose-file/services.md
```diff
@@ -1752,6 +1752,47 @@ services:
...
+        condition: service_healthy
+    pre_start:
+      - command: ["./manage.py", "migrate"]
+      - image: busybox
+        command: sh -c 'chown -R 1000:1000 /data'
+    volumes:
+      - data:/data
+
+  db:
+    image: postgres:16
```
docker-agent, line 1790: **[MEDIUM] Inconsistent Postgres version between examples in the same PR**

This code block uses `postgres:16`, but the how-to guide added by this same PR (`content/manuals/compose/how-tos/init-containers.md`) uses `postgres:18` in its examples. The inconsistency looks like a copy-paste oversight and could confuse readers comparing the two pages — or imply that the feature requires a specific older version of Postgres.

Consider aligning both examples to the same version tag.

docker-agent (commented):

Two medium-severity issues found in the new `pre_start` reference documentation added by this PR. Both are in `content/reference/compose-file/services.md`.

The new how-to page (`content/manuals/compose/how-tos/init-containers.md`) is well-written and clear. No missing redirects, no vendored content concerns, no front matter issues.

content/reference/compose-file/services.md
aevesdocker: this is fun as is

content/reference/compose-file/services.md
aevesdocker: no this is fine. not a deal breaker. it's an ilustrative example

content/reference/compose-file/services.md
docker-agent: I hear you! My concern is more about consistency with the rest of the bullet list. Looking at the other items (command, image, user, privileged, working_dir, environment), they all present just the key name without any values. Having per_replica: false in the label breaks this pattern and might suggest to readers that the config key is literally called "per_replica: false" rather than just "per_replica".

That said, if you feel the current format better emphasizes that false is the only currently supported value, I'm happy to defer to your judgment on what works best for the docs.

content/reference/compose-file/services.md
docker-agent: Understood, thanks for clarifying. You're right that the version difference doesn't affect the functionality being demonstrated.

usha-mandya (approved):
LGTM

content/reference/compose-file/services.md
dvdksn, line 1790: This example is confusing because it mixes two unrelated concerns in one `pre_start` sequence:

1. `./manage.py migrate` — a DB migration (needs the `db` service)
2. `chown -R 1000:1000 /data` — fixing ownership of a volume

The `app` service has `volumes: data:/data` but the `db` service has no volume. A reader trying to understand `pre_start` from this example will wonder why the app is chowning a path that the database doesn't share, and why the database has no persistent volume of its own.

Also, `db: image: postgres:16` has no `POSTGRES_PASSWORD`, `POSTGRES_USER`, or `POSTGRES_DB` — the postgres container won't start.

Suggest splitting into two focused examples: one for migrations (the DB case, already done in `init-containers.md`) and one for volume ownership (the busybox/chown case, already done separately in `init-containers.md`). The combined example here adds confusion rather than showing a realistic use case.

content/manuals/compose/how-tos/init-containers.md
```diff
@@ -0,0 +1,173 @@
...
+  app:
+    image: myapp:latest
+    depends_on:
+      db:
+        condition: service_healthy
+    pre_start:
+      - command: ["./manage.py", "migrate"]
+
+  db:
+    image: postgres:18
```
dvdksn, line 59: Two problems with this `db` service that would prevent it from actually working:

1. **No `POSTGRES_PASSWORD`** — Postgres refuses to start without it (or `POSTGRES_HOST_AUTH_METHOD=trust`), with the error: `Error: Database is uninitialized and superuser password is not specified. You must specify POSTGRES_PASSWORD to a non-empty value for the superuser.`

2. **Version inconsistency** — this example uses `postgres:18` but the `services.md` reference example uses `postgres:16`. Pick one version and use it consistently across both files.

At minimum the `db` service should look like:
```yaml
  db:
    image: postgres:17
    environment:
      POSTGRES_PASSWORD: example
      POSTGRES_USER: myuser
      POSTGRES_DB: mydb
    healthcheck:
      test: ["CMD-SHELL", "pg_isready -U $${POSTGRES_USER} -d $${POSTGRES_DB}"]
      interval: 10s
      retries: 5
      start_period: 30s
      timeout: 10s
```

content/manuals/compose/how-tos/init-containers.md
```diff
@@ -0,0 +1,173 @@
...
+    image: myapp:latest
+    depends_on:
+      db:
+        condition: service_healthy
+    pre_start:
+      - command: ["./manage.py", "migrate"]
+      - command: ["./manage.py", "loaddata", "fixtures.json"]
+
+  db:
+    image: postgres:18
```
dvdksn, line 114: Same issue here: `postgres:18` with no `POSTGRES_PASSWORD` — this db container won't start. Same version inconsistency vs `services.md` (`postgres:16`).

dvdksn (commented):
Three issues with the Postgres examples that need fixing before merge.

PR #25483, opened 2026-07-02
Clarify BuildKit daemon lifecycle and terminal requirements on Windows

ngebremariam-msft:
## Description

Added a warning to keep the buildkitd.exe terminal open during subsequent steps, and a FAQ section addressing common daemon lifecycle questions for first-time users.
                                                                                                          
When looking at the BuildKit on Windows documentation, I encountered several areas where additional clarification would improve the user experience and reduce confusion for first-time users.                
                                                                                                         
#### 1) Importance of BuildKit daemon in Step 7                                                               
The guide instructs users to start `buildkitd.exe` and then open another terminal to run subsequent BuildKit commands. However, it does not clearly explain that the terminal running `buildkitd.exe` must remain open while executing those commands.                                          
Users may close the original terminal, causing the BuildKit daemon to stop unexpectedly. As a result, subsequent BuildKit commands (such as `buildctl` and `docker buildx` commands) may fail because they can  no longer connect to the daemon.                                                                          
                                                                                                          
#### **2) No explanation of daemon lifecycle**               
After completing a build, users may not know:                                                              
   - Whether the BuildKit daemon must continue running.                                                   
   - Whether the image can be run independently of BuildKit.                                              
   - Whether the daemon can safely be stopped after the build completes." Does the commit message and description help justify the reason for request of change? If not, create suggestions and explain.

### DIFF
--- a/content/manuals/build/buildkit/_index.md
+++ b/content/manuals/build/buildkit/_index.md
@@ -152,15 +152,18 @@ see [GitHub issues](https://github.com/moby/buildkit/issues?q=is%3Aissue%20state
    ```
 
    > [!NOTE]
-   > If you are running a _dockerd-managed_ `containerd` process, use that instead, by supplying the address:
+   > If you are running a _dockerd-managed_ `containerd` process, use that instead, by supplying the     address:
    > `buildkitd.exe --containerd-worker-addr "npipe:////./pipe/docker-containerd"`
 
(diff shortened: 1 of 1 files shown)


dvdksn:
Hi @ngebremariam-msft 

I don't think these additions to the current buildkitd on windows docs are necessary. BuildKit on Windows is still experimental and aimed at advanced users. The information you propose adding seems more targeted at inexperienced users. That is not the target audience for buildkitd on windows at this time.

PR #25501, opened 2026-07-06
scout: fix redirect issue and update examples

craig-osterhout:
## Description

Update the new scout GHA examples based on testing.
- Added organization as it's required when comparing against environments
- Added specific version as v1 is currently pinned to v1.20.4, which predates policy-config support (added in v1.23.0).

Fixed redirect issue

## Reviews

- [ ] Editorial review

### DIFF
--- a/content/manuals/scout/policy/local.md
+++ b/content/manuals/scout/policy/local.md
@@ -2,8 +2,6 @@
 title: Evaluate policies
 description: Evaluate Docker Scout policies using the CLI, with built-in and custom Rego policies
 keywords: scout, policy, rego, opa, cli, custom policies, policy bundle
-aliases:
-  - /scout/policy/local/
 ---
 
 {{< summary-bar feature_name="Evaluate policies" >}}
--- a/content/manuals/scout/policy/local.md
+++ b/content/manuals/scout/policy/local.md
@@ -54,6 +54,7 @@ to evaluate policies as part of your workflow:
   with:
     command: policy
     image: ${{ env.IMAGE_NAME }}
+    organization: <ORG>
 ```
 
 For other CI platforms, install the
@@ -68,13 +69,14 @@ dashboard-managed policy settings, replicate those settings locally with
 `--policy-config`:
 
 ```yaml
-- uses: docker/scout-action@v1
+- uses: docker/scout-action@v1.23.0
   with:
     command: compare
     image: ${{ env.IMAGE_NAME }}
     to-env: production
     exit-on: policy
     policy-config: policies.json
+    organization: <ORG>
 ```
 
 See [Configure built-in policies](#configure-built-in-policies) for the


docker-agent (commented):
### Assessment: 🟢 APPROVE

Reviewed `content/manuals/scout/policy/local.md`.

The three changes in this PR are all correct:

1. **Alias removal** — The `aliases: [/scout/policy/local/]` entry was intentionally removed to fix a redirect conflict. No other docs reference that path, so no inbound links break.
2. **`organization: <ORG>` parameter** — Added to both code examples using the correct Docker docs placeholder syntax (`<VAR>`). Accurate and consistent.
3. **Version pin `@v1.23.0`** — The second block specifically demonstrates `policy-config`, which requires v1.23.0+. The first block doesn't use `policy-config` so `@v1` remains correct there. The version difference is intentional and technically accurate.

No style-guide violations, no markdown issues, no vendored files touched.

content/manuals/scout/policy/local.md
```diff
@@ -2,8 +2,6 @@
 title: Evaluate policies
 description: Evaluate Docker Scout policies using the CLI, with built-in and custom Rego policies
 keywords: scout, policy, rego, opa, cli, custom policies, policy bundle
-aliases:
-  - /scout/policy/local/
```
docker-agent, line 6: **[HIGH] Redirect alias removed — `/scout/policy/local/` may now 404**

This PR removes the `aliases` front matter block that was providing a redirect from `/scout/policy/local/` to the canonical page URL. The PR description mentions "Fixed redirect issue," but it's not clear whether this removal is the fix (e.g., resolving a redirect loop) or an accidental side-effect.

**Concern:** If any external links, search engine indexes, or cross-docs references point to `/scout/policy/local/`, those requests will now return a 404 instead of landing on this page.

**Suggested check:** Confirm whether there is another redirect mechanism elsewhere in the repo (e.g., a `_redirects` file, Netlify config, or a separate Hugo alias) that handles `/scout/policy/local/` → the canonical URL. If not, the alias should be restored or the redirect should be documented here.

docker-agent (commented):

content/manuals/scout/policy/local.md
craig-osterhout: it was circular

docker-agent: Got it — thanks for clarifying! A circular redirect makes sense here: the alias `/scout/policy/local/` was pointing to the canonical path `/manuals/scout/policy/local/`, but if there's already a site-wide redirect rule that maps `/scout/*` → `/manuals/scout/*`, then the alias was creating a loop. Removing it was the right call.

Apologies for the noise — I should have checked for broader redirect patterns before flagging this.

usha-mandya (approved):

PR #25502, opened 2026-07-06
check for circular aliases

craig-osterhout:
## Description

Added check for circular aliases.

I've done it a few times. Move something and add an alias. Then move it back and forget to remove the alias.

Currently there is a circular alias, so it should fail successfully.

## Reviews

- [ ] Editorial review


hack/test/validate_aliases
```diff
@@ -0,0 +1,57 @@
...
+    /^---$/ {
+      if (fm == 0) { fm=1; next }
+      else { exit }
+    }
+    fm && /^aliases:/ { in_aliases=1; next }
+    fm && in_aliases && /^  - / { print substr($0, 5) }
+    fm && in_aliases && /^[^ ]/ { in_aliases=0 }
+  ' "$file")
+
+  for alias in $aliases; do
```
docker-agent: **[LOW] Unquoted `$aliases` is subject to glob expansion**

`for alias in $aliases` works correctly for splitting by newlines (one alias per line), but it also triggers shell glob expansion on each value before iteration. If any alias contained `*`, `?`, or `[...]` characters, the shell would expand them to matching filenames instead of treating them as literal path strings. POSIX sh has no implicit protection against this — you'd need `set -f` (disable globbing) before the loop, or read values line-by-line via a `while IFS= read -r alias` pattern instead.

For Docker docs' actual alias values this is unlikely to matter, but it's a latent risk in new code.

docker-agent (commented):
### Assessment: 🟢 APPROVE

The circular alias detection script is well-structured and integrates cleanly with the existing CI pipeline. The logic correctly identifies self-referential aliases across both `manuals/` and standard Hugo paths. One minor latent risk noted inline.

hack/test/validate_aliases
```diff
@@ -0,0 +1,57 @@
...
+
+  # Extract aliases from YAML frontmatter
+  aliases=$(awk '
+    BEGIN { fm=0; in_aliases=0 }
+    /^---$/ {
+      if (fm == 0) { fm=1; next }
+      else { exit }
+    }
+    fm && /^aliases:/ { in_aliases=1; next }
+    fm && in_aliases && /^  - / { print substr($0, 5) }
```
docker-agent: **[MEDIUM] Inline YAML array alias syntax not parsed — circular aliases in `aliases: [...]` format are silently missed**

The awk frontmatter parser only matches block-style alias entries (lines matching `/^  - /`). Hugo also supports inline array syntax, e.g.:

```yaml
aliases: ["/old-path/"]
```

That format is not matched, so any circular alias written in that style would be silently skipped. A quick grep of the repo shows no files currently use inline syntax — all aliases use block style — so there's no immediate false negative. But the check is fragile: a future alias written inline would evade detection without any warning.

Consider adding inline array parsing, or add a comment noting the limitation.

docker-agent (commented):

dvdksn (commented):
Hey, I noticed this PR only now. I think there is a better way to detect these. We already process the aliases in the [home.redirects.json template](https://github.com/docker/docs/blob/main/layouts/home.redirects.json). As part of generating the redirects.json file, we could have hugo compare aliases and permalink, and `errorf` if there's a conflict. Something like:

```diff
 {{- range hugo.Sites -}}
   {{- range .Pages -}}
     {{- if .Params.aliases -}}
+      {{- $page := . -}}
       {{- $target := .RelPermalink -}}
       {{- range .Params.aliases -}}
-        {{ $redirects.SetInMap "paths" . $target }}
+        {{- $alias := . -}}
+        {{- if eq (strings.TrimSuffix "/" $alias) (strings.TrimSuffix "/" $target) -}}
+          {{- errorf "alias %q in %s resolves to the page itself" $alias $page.File.Path -}}
+        {{- end -}}
+        {{- $redirects.SetInMap "paths" $alias $target -}}
       {{- end -}}
     {{- end -}}
   {{- end -}}
 {{- end -}}
```

craig-osterhout:
That took me way longer than it should of to realize that it fails the netlify check and not the build one. 
Looks like it's working.
https://app.netlify.com/projects/docsdocker/deploys/6a62921881577c0008738802

docker-agent (commented):
### Assessment: 🟢 APPROVE

The PR adds a build-time self-referential alias check to the Hugo redirect template. The logic is sound for this repo's alias conventions (all aliases use leading slashes, matching 's format), so the equality check after stripping trailing slashes is effective.

No issues found in the changed code.

dvdksn (approved):

Issue #25491, opened 2026-07-03
Swarm secrets docs: "Defining and using secrets in compose files" section links to an irrelevant legacy-versions page and omits the one example Swarm users need (external: true)


deahtstroke:
### Is this a docs issue?

- [x] My issue is about the documentation content or website

### Type of issue

I can't find what I'm looking for

### Description

Under the "Defining and using secrets in compose files" heading, the docs state that docker-compose and docker stack both support defining secrets in a compose file, and link out to "the Compose file reference for details." That link currently points to: [here](https://docs.docker.com/reference/compose-file/legacy-versions/). That page is about legacy Compose file version numbers (e.g. 2.x vs 3.x schema history) and it has nothing to do with how to declare or reference a secret using Compose + Swarm. For a reader following a heading specifically about secrets in Swarm, this link doesn't answer the question it's attached to.

More importantly, the only Compose secrets example that does appear on this page (further down, in "Use Secrets in Compose") only demonstrates `file:-` sourced secrets. For Swarm specifically, the far more common pattern is referencing a secret that already exists in the swarm via `docker secret create`, using:

```yaml
secrets:
  my_secret:
    external: true
```

This is the property that lets a compose file reference a pre-existing swarm secret rather than having Compose try to create one from a local file. It's arguably the more relevant example for this specific page, since it's Swarm-specific behavior, but it isn't shown anywhere on `engine/swarm/secrets/`, nor is it clearly documented as being swarm-mode-only.

### Location

https://docs.docker.com/engine/swarm/secrets/

### Suggestion

Update the "Compose file reference for details" link to point to the actual secrets reference page (https://docs.docker.com/reference/compose-file/secrets/) instead of the legacy-versions page.

Add a short `external: true` example directly in the "Defining and using secrets in compose files" section (or in "Use Secrets in Compose"), e.g.:

```yaml
services:
  app:
    image: myapp:latest
    secrets:
      - my_secret

secrets:
  my_secret:
    external: true
```

with a note that `docker secret create my_secret ... ` must be run first, and that `external: true` is a Swarm-only feature (unsupported by plain docker compose). Additionally, in the page https://docs.docker.com/reference/compose-file/secrets/, frankly I think that showcasing that `external` is an valid property for settings secrets in compose should be brought back with the big caveat that this is a Swam-specific attribute

[...]

PR #25522, opened 2026-07-09
docs: Add `external` property back to compose and add swarm secrets example using compose

deahtstroke:
## Description

Added back the attribute `external` in the `/reference/compose/secrets` section with a warning that this attribute is only available for Docker Swarm. Additionally, I added a small toy reference example to the section "Defining and using Secrets in compose files" and moved the section outside of the "Examples" section. The reason why I did this is because I noticed the Examples are all runnable steps that showcase concrete usage of secrets, the example I added is just a skeleton/reference to how they should be used in conjunction with `docker secret create...`, not a true able-to-run example like the ones below. I also changed the hyperlink from `/reference/compose-file/legacy-versions.md` to `/reference/compose-file/secrets.md`.

A few caveats that I wasn't sure about when writing this:
- I'm not sure if using the "[!NOTE]" in the attributes list is the approach maintainers and writers might favor and I can change it accordingly
- I'd love to hear feedback on moving "Defining and using secrets in compose files" outside of "Examples" due to my reasoning above

Edit:
I'm also not opposed to making a fully working example using compose, stack and secrets and putting the section back under `Examples`

## Related issues or tickets
 #25491 

## Reviews

- [ ] Technical review
- [ ] Editorial review
- [ ] Product review


aevesdocker:
Thanks @deahtstroke. The `engine/swarm/secrets/` addition makes sense to me and I'm fine with that part.

Could you drop the `reference/compose/secrets` change, though? Plain `docker compose` errors with `unsupported external secret` on that attribute since it requires swarm, so even with the `[!NOTE]` caveat, documenting it on the generic Compose reference risks readers copying it outside a swarm context and hitting an unexplained CLI error. That page's audience is mostly plain-Compose users. Let's keep `external` scoped to the swarm-specific page where it's actually valid, and leave the generic reference as-is.

deahtstroke:
Thanks for the Review! Yeah I hard agree on that rationale, which also makes sense as to why the original issues about the external option being confusing for compose users were opened in the first place. I removed the changes to `/content/reference/compose-file/secrets.md` 😄

content/manuals/engine/swarm/secrets.md
```diff
@@ -125,6 +125,35 @@ Use these links to read about specific commands, or continue to the
 - [`--secret`](/reference/cli/docker/service/create/#secret) flag for `docker service create`
 - [`--secret-add` and `--secret-rm`](/reference/cli/docker/service/update/#secret-add) flags for `docker service update`
 
+## Defining and using secrets in compose files
```
aevesdocker:
```suggestion
## Defining and using secrets in Compose files
```

content/manuals/engine/swarm/secrets.md
```diff
@@ -125,6 +125,35 @@ Use these links to read about specific commands, or continue to the
...
+
+Using secrets in compose files alongside Docker Swarm is as simple as defining a secret
```
aevesdocker:
```suggestion
Using secrets in a Compose file alongside Docker Swarm is as simple as defining a secret
```

content/manuals/engine/swarm/secrets.md
```diff
@@ -125,6 +125,35 @@ Use these links to read about specific commands, or continue to the
...
+            MY_ENV: /run/secrets/some_secret
+        secrets:
+            - some_secret
+
+secrets:
+    some_secret:
+        external: true
+```
+
+Both the `docker-compose` and `docker stack` commands support defining secrets
```
aevesdocker:
```suggestion
Both the `docker compose` and `docker stack` commands support defining secrets
```

content/manuals/engine/swarm/secrets.md
```diff
@@ -125,6 +125,35 @@ Use these links to read about specific commands, or continue to the
...
+in a compose file. See
```
aevesdocker:
```suggestion
in a Compose file. See
```

aevesdocker (approved):
Thank you @deahtstroke!

PR #25528, opened 2026-07-10
docs: fix Swarm secrets Compose reference link

fhshaik:
## Summary

Updates the Swarm secrets page to link to the Compose secrets reference instead of the legacy Compose file versions page.

Addresses the broken-link portion of #25491.

## Validation

- Ran `git diff --check`


dvdksn (approved):
Thanks!

PR #25541, opened 2026-07-11
compose: note that secrets require Linux containers

### LINKED ISSUE #25540: Compose secrets page: file-based secrets don't work on Windows containers (no caveat noted)
### Is this a docs issue?

### Type of issue

Information is incorrect

### Description

The Compose secrets guide presents file-based secrets as supported and states they're delivered via a bind mount at `/run/secrets/<secret_name>`, with no platform caveat. This is incorrect for **Windows containers**: `docker compose up` with a file-based secret fails at container creation with:

```
invalid mount config for type "bind": source path must be a directory
```

(and, with the default target, `invalid mount path: '/run/secrets/<secret_name>'`).

The cause is a platform limitation — Windows containers can't bind-mount a single file, only directories — acknowledged by maintainers in moby/moby#37120 ("Windows does not support mounting files (only directories)… this is a limitation of the platform") and docker/cli#2055 ("Windows itself cannot do this, so not something we can fix in Docker"). Because Compose file-based secrets are single-file bind mounts, they can't work on Windows containers. (Swarm secrets *do* work on Windows Server Core per moby/moby#37945, because they don't use a plain bind mount.)

Tested on a Windows Server 2025 container (`mcr.microsoft.com/windows/server:ltsc2025`), Hyper-V isolation, Docker 29.6.1, Compose v2.

### Location

https://docs.docker.com/compose/how-tos/use-secrets/

### Suggestion

Add a platform-limitations note to the Compose secrets page stating that file-based secrets aren't supported on Windows containers (single-file bind mounts aren't supported by the platform), and point Windows users to Swarm secrets or a directory bind-mount as alternatives.

mickey-mikey:
### Description

Adds a NOTE callout to the Compose secrets guide stating that secrets are supported on Linux containers only, with the reason.

The page states secrets are delivered as a bind mount at `/run/secrets/<secret_name>`, with no platform caveat. On Windows containers this fails at container creation:

```
invalid mount config for type "bind": source path must be a directory
```

Windows containers can't bind-mount a single file, only directories — a platform limitation confirmed by maintainers in moby/moby#37120 and docker/cli#2055. Because Compose secrets are single-file bind mounts, they can't work on Windows containers.

Tested on a Windows Server 2025 container (`mcr.microsoft.com/windows/server:ltsc2025`), Hyper-V isolation, Docker 29.6.1, Compose v2.

Fixes #25540

### Related issues or tickets

- #25540
- moby/moby#37120
- docker/cli#2055

### Reviews

- [ ] Technical review
- [ ] Editorial review

### DIFF
--- a/content/manuals/compose/how-tos/use-secrets.md
+++ b/content/manuals/compose/how-tos/use-secrets.md
@@ -21,7 +21,7 @@ Secrets are mounted as a file in `/run/secrets/<secret_name>` inside the contain
 
 > [!NOTE]
 >
-> Secrets are supported on Linux containers only. Compose delivers each secret by bind-mounting a single file into the container, and Windows containers support bind-mounting directories only. For Windows containers, bind-mount a directory containing the secret file instead.
+> Secrets are supported on Linux containers only. Compose delivers each secret by bind-mounting a single file into the container, and Windows containers support bind-mounting directories only.
 
 Getting a secret into a container is a two-step process. First, define the secret using the [top-level secrets element in your Compose file](/reference/compose-file/secrets.md). Next, update your service definitions to reference the secrets they require with the [secrets attribute](/reference/compose-file/services.md#secrets). Compose grants access to secrets on a per-service basis.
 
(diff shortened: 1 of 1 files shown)


aevesdocker (approved):
Thanks!

PR #25555, opened 2026-07-14
docs: add sbx migration guidance

### LINKED ISSUE #25554: [docs-scanner] Deprecated docker-desktop.md page has unclear migration guidance
**File:** `content/manuals/ai/sandboxes/docker-desktop.md`

### Issue

The deprecation warning at the top of the page states:

> The Docker Desktop-integrated `docker sandbox` commands are deprecated and replaced by the standalone [`sbx` CLI](/manuals/ai/sandboxes/_index.md). This deprecation applies only to the Docker Desktop integration, not to Docker Sandboxes.

However, the rest of the page then provides full documentation for the deprecated `docker sandbox` commands as if they're still the primary way to use Docker Sandboxes. This creates confusion about whether readers should:
1. Use the deprecated commands (since they're fully documented)
2. Migrate to `sbx` (since there's a deprecation warning)
3. Use both (unclear when to use which)

The page also includes a note box saying "Use the standalone `sbx` CLI for sandboxed AI agent workflows" but then continues with extensive `docker sandbox` documentation.

### Why this matters

A reader landing on this page from search or an old bookmark will be uncertain whether they should follow the documented commands or migrate away from them. The deprecation warning says the integration is deprecated but doesn't clearly state whether the commands still work, for how long, or what the migration path looks like.

The page structure suggests these commands are still viable (full command reference, troubleshooting, architecture details), which contradicts the deprecation framing.

### Suggested fix

Either:

1. **If the commands still work**: Add a clear statement about the timeline ("These commands continue to work in Docker Desktop 4.58+ but will be removed in version X.XX") and add a migration section showing the `docker sandbox` → `sbx` command equivalents.

2. **If the commands are already removed or imminent**: Replace the full documentation with a brief "This page documents deprecated commands" notice and a table mapping old commands to new ones, then link to the current `sbx` documentation.

3. **If this page should only exist for historical reference**: Add `sitemap: false` is already present, but also add a more prominent banner at the top stating this is historical documentation only and all users should migrate to `sbx`.

The current middle ground—warning about deprecation while providing full usage documentation—leaves readers unsure what action to take.

---

nielskaspers:
## Description

Clarify how readers should move from the deprecated Docker Desktop-integrated `docker sandbox` commands to the standalone `sbx` CLI.

- Tell readers to use `sbx` for new sandboxes
- Map the common run, list, shell, and remove commands
- Link to current setup and network-policy guidance

Fixes #25554.

## Validation

- `git diff --check`
- `npx --yes markdownlint-cli content/manuals/ai/sandboxes/docker-desktop.md`

The repository lint wrapper completed markdownlint, then stopped because Vale is not installed locally. The prescribed Prettier command is unavailable in this checkout because `prettier-plugin-go-template` is missing.

### DIFF
--- a/content/manuals/ai/sandboxes/docker-desktop.md
+++ b/content/manuals/ai/sandboxes/docker-desktop.md
@@ -20,8 +20,25 @@ running AI coding agents in isolated microVMs. This integration is superseded
 by the standalone `sbx` CLI, which provides the full Docker Sandboxes workflow
 and doesn't require Docker Desktop.
 
-> [!NOTE]
-> Use the standalone `sbx` CLI for sandboxed AI agent workflows.
+## Migrate to the sbx CLI
+
+Use the standalone `sbx` CLI for new sandboxes. Use this page only when you
+need to work with existing Docker Desktop-integrated sandboxes while you
+migrate.
+
+The common commands map as follows:
+
+| Task             | Docker Desktop integration             | Standalone CLI                |
+| ---------------- | -------------------------------------- | ----------------------------- |
+| Run an agent     | `docker sandbox run AGENT [PATH]`      | `sbx run AGENT [PATH]`        |
+| List sandboxes   | `docker sandbox ls`                    | `sbx ls`                      |
+| Open a shell     | `docker sandbox exec -it <name> bash`  | `sbx exec -it <name> bash`    |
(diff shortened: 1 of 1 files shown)


dvdksn:
Hi, thanks for the contribution @nielskaspers but I don't think this addition is really necessary, so I'm closing. The `docker sandbox` plugin was removed entirely from Docker Desktop so we'll also be removing these docs.

PR #25655, opened 2026-07-27
docs: clarify clone mode security boundary

### LINKED ISSUE #25653: [docs-scanner] Clone mode security boundary unclear about untracked file exposure
**File:** `content/manuals/ai/sandboxes/security/isolation.md`

### Issue

The clone mode section creates confusion about what files are protected from the agent. The introduction promises strong isolation:

> When you start a sandbox with [`--clone`](../usage.md#clone-mode), the agent never works directly against your host repository. Even with full root inside the VM, it cannot modify your `.git` directory, your working tree, or any tracked file on your host.

But later in the same section, it reveals that ALL files in the working directory are readable:

> Your repository's Git root is mounted at `/run/sandbox/source` as read-only. The mount covers your entire working directory, including untracked files and files excluded by `.gitignore`. Nothing the agent does inside the VM can write back through that mount, but all files under the Git root are readable inside the sandbox. This includes credential files not tracked by Git, such as `.env`.

The warning about `.env` files comes after the reader has already been told clone mode provides isolation from "your host repository."

### Why this matters

Readers choosing clone mode for security will believe their untracked files (credentials, local config, etc.) are protected from the agent. The opening framing emphasizes what the agent "cannot modify" and "cannot write," leading readers to assume a stronger boundary than actually exists. The revelation that untracked files are readable is buried mid-section, and the advice to store secrets elsewhere comes too late.

A reader following this documentation would:
1. Choose clone mode thinking it isolates the agent from their local files
2. Leave sensitive `.env` files in their working directory
3. Unknowingly expose those credentials to the sandboxed agent

### Suggested fix

Restructure the clone mode section to lead with what IS and ISN'T isolated:

**What clone mode protects:**
- Your `.git` directory and working tree from modification
- Your Git configuration and hooks from being altered

**What clone mode does NOT protect:**
- Untracked files in your working directory are mounted read-only and readable by the agent
- This includes `.env` files, local credentials, and other untracked files

Move the warning about storing secrets outside the working directory to the beginning of the section, before describing the technical implementation. The current structure front-loads the benefits and buries the limitation.

Alternatively, consider whether the read-only mount should exclude untracked files entirely, or provide a flag to control this behavior.

---

devendrakushwah80:
## Description

Moves the clone mode security warning to the beginning of the section to make
the security boundary clearer. This clarifies that clone mode protects the host
repository from modification, but not from inspection, and removes the
duplicated guidance later in the section.

## Related issues or tickets

Fixes #25653

## Reviews

- [x] Technical review
- [ ] Editorial review
- [ ] Product review

### DIFF
--- a/content/manuals/ai/sandboxes/security/isolation.md
+++ b/content/manuals/ai/sandboxes/security/isolation.md
@@ -148,6 +148,13 @@ never works directly against your host repository. Even with full root
 inside the VM, it cannot modify your `.git` directory, your working tree,
 or any tracked file on your host.
 
+> [!IMPORTANT]
+> Clone mode protects your host repository from modification, **not from
+> inspection**. Your repository is still mounted read-only into the sandbox,
+> including untracked files and files excluded by `.gitignore`. Files such as
+> `.env` remain readable by the agent. Store secrets outside your working
+> directory or use [credential isolation](credentials.md) instead.
+
 ```mermaid
 flowchart LR
   subgraph host["Host repository (untouched)"]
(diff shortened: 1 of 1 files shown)


dvdksn (approved):
Thanks @devendrakushwah80

PR #25660, opened 2026-07-27
daemon: Document Unix socket setup on Windows

vvoland:
Show the Engine 29.5 opt-in listener and matching client endpoint. Explain group access and why the socket should use a user-profile path.

## Reviews

- [ ] Technical review
- [ ] Editorial review
- [ ] Product review

### DIFF
--- a/content/manuals/engine/daemon/_index.md
+++ b/content/manuals/engine/daemon/_index.md
@@ -84,6 +84,33 @@ $ dockerd --debug \
   --host tcp://192.168.59.3:2376
 ```
 
+#### Listen on a Unix socket on Windows
+
+Starting with Docker Engine 29.5, you can configure `dockerd` on Windows to
+listen on a Unix socket. This setting is optional. The default Windows endpoint
+remains the `npipe` named pipe.
+
+Run `dockerd` from an elevated PowerShell session:
+
+```powershell
+PS C:\> dockerd -H unix://C:/Users/<USER>/docker.sock --group docker-users
+```
+
+Replace `<USER>` with the name of your Windows user profile directory.
+Without `--group`, only members of the Administrators group and the `SYSTEM`
+account can access the socket. Use a comma-separated list to grant access to
+multiple users or groups.
+
+Use a path in the user profile when granting access with `--group`. A socket
+created directly under `C:\` can inherit a high integrity level that blocks
+access for the configured users and groups.
+
+Configure the Docker client to use the same socket:
+
+```powershell
+PS C:\> docker -H unix://C:/Users/<USER>/docker.sock info
+```
+
 Learn about the available configuration options in the
 [dockerd reference docs](/reference/cli/dockerd.md), or by
 running:


content/manuals/engine/daemon/_index.md
```diff
@@ -84,6 +84,32 @@ $ dockerd --debug \
...
+
+Run `dockerd` from an elevated PowerShell session:
+
+```powershell
+PS C:\> dockerd -H unix://C:/Users/<USER>/docker.sock --group docker-users
+```
+
+Replace `<USER>` with your Windows user directory. Without `--group`, only
+members of the Administrators group and the `SYSTEM` account can access the
+socket. Use a comma-separated list to grant access to multiple users or groups.
```
docker-agent: **[HIGH] `--group` may not accept a comma-separated list of multiple values**

Line 101 states: "Use a comma-separated list to grant access to multiple users or groups."

The Docker daemon's `--group` flag on Linux (and by convention elsewhere) accepts a **single group name**. If the new Windows Unix socket implementation follows the same behavior, this sentence is factually incorrect and will mislead users who try to pass `docker-users,another-group`.

Please verify against the Docker Engine 29.5 changelog or `dockerd --help` output on Windows. If `--group` only accepts one value, remove or correct this sentence.

content/manuals/engine/daemon/_index.md
```diff
@@ -84,6 +84,32 @@ $ dockerd --debug \
...
+listen on a Unix socket. This setting is optional. The default Windows endpoint
+remains the `npipe` named pipe.
+
+Run `dockerd` from an elevated PowerShell session:
+
+```powershell
+PS C:\> dockerd -H unix://C:/Users/<USER>/docker.sock --group docker-users
+```
+
+Replace `<USER>` with your Windows user directory. Without `--group`, only
```
docker-agent: **[MEDIUM] `<USER>` described as "user directory" but should be "username"**

Line 99: "Replace `<USER>` with your Windows user directory."

The placeholder `<USER>` in the path `C:/Users/<USER>/docker.sock` stands for the **username** (e.g., `johndoe`), not the full directory. "User directory" could lead readers to substitute a full path like `C:\Users\johndoe` rather than just `johndoe`, breaking the intended path.

Suggested fix: "Replace `<USER>` with your Windows username."

docker-agent (commented):

content/manuals/engine/daemon/_index.md
docker-agent: **[HIGH] `--group` does not accept a comma-separated list**

The sentence "Use a comma-separated list to grant access to multiple users or groups" is inaccurate. The `dockerd --group` flag accepts a single group name — it does not support comma-separated values. Passing a comma-separated string would be treated as a literal group name (e.g., `"docker-users,another-group"`) and would not grant access to multiple groups.

If Docker Engine 29.5 supports granting access to multiple groups on Windows, the correct syntax should be demonstrated explicitly (for example, by repeating the flag or using a different mechanism). Otherwise, this sentence should be removed or corrected to reflect that only a single group name is accepted.

docker-agent (commented):
### Assessment: 🔴 CRITICAL

One high-severity factual error was found in the new Windows Unix socket section. The claim that `--group` accepts a comma-separated list of groups does not match established Docker Engine behavior, where `--group` is a single-value flag. This will mislead users who try to grant access to multiple groups.

The rest of the new section — the `-H unix://` flag usage, the PowerShell code blocks, the user-profile path guidance — is well-written and follows the document's existing tone and format.

content/manuals/engine/daemon/_index.md
vvoland: It does: 
https://github.com/moby/moby/blob/d16a293b18779e678a707e29655bf3221b9a8cac/daemon/listeners/listeners_windows.go#L13-L20

docker-agent: You're absolutely right, I apologize for the incorrect finding. I missed that the Windows implementation specifically handles comma-separated values — the code at listeners_windows.go clearly splits the socketGroup by commas to support multiple groups and/or users.

The documentation is accurate as written. Thank you for the correction and the reference!

content/manuals/engine/daemon/_index.md
```diff
@@ -84,6 +84,33 @@ $ dockerd --debug \
   --host tcp://192.168.59.3:2376
 ```
 
+#### Listen on a Unix socket on Windows
```
docker-agent, line 87: **[MEDIUM] New section uses `####` heading which is excluded from the table of contents**

The document's `toc_max` defaults to 3 (h3), as no override is set in the front matter. The `#### Listen on a Unix socket on Windows` heading introduced by this PR is at level h4, so it won't appear in the rendered table of contents.

Consider using `###` instead to give this section ToC visibility, or add `toc_max: 4` to the front matter if nesting under `### Configuration using flags` is intentional.

docker-agent (commented):

dvdksn (approved):
LGTM

PR #25690, opened 2026-08-03
docs: clarify NVIDIA GPU access scope

### LINKED ISSUE #25685: [docs-scanner] Unclear 'single engine' restriction for NVIDIA GPU access
**File:** `content/manuals/engine/containers/gpu.md`

### Issue

The page contains a vague note about GPU access restrictions:

> [!NOTE]
>
> NVIDIA GPUs can only be accessed by systems running a single engine.

This statement is unclear and would confuse readers because:

1. "Single engine" is not defined - does it mean:
   - Only one Docker Engine instance per host?
   - Not supported in Docker Swarm (multiple engines)?
   - Only one container can access the GPU at a time?
   - Something else entirely?

2. The note appears without context or explanation
3. There's no link to documentation explaining this limitation or its implications

### Why this matters

Readers trying to use GPUs with Docker will encounter this note and won't know:
- Whether they can run multiple containers with GPU access on the same host
- Whether this is a Docker Swarm limitation
- What happens if they violate this constraint
- How to configure their system to comply with this requirement

The lack of clarity could lead readers to make incorrect assumptions about GPU support in their Docker deployments.

### Suggested fix

Clarify what "single engine" means with specific language, such as:

- If it means Swarm: "NVIDIA GPUs are not supported in Docker Swarm mode. Use standalone Docker Engine for GPU access."
- If it means one engine per host: "Only one Docker Engine instance per host can access NVIDIA GPUs."
- If it means something else: Provide the specific constraint with an explanation

Also consider adding a link to relevant documentation about GPU limitations or multi-engine scenarios.

---

codewithdaniel1:
## Description

Clarify the note about NVIDIA GPU access in the Docker Engine container documentation.

The previous wording said that NVIDIA GPUs can only be accessed by systems running a "single engine", but did not define what "single engine" means. This update reframes the note around the documented scope of the page: standalone Docker Engine containers on a single host.

## Related issues or tickets

Closes #25685

## Reviews

- [ ] Technical review
- [ ] Editorial review
- [ ] Product review

### DIFF
--- a/content/manuals/engine/containers/gpu.md
+++ b/content/manuals/engine/containers/gpu.md
@@ -88,9 +88,11 @@ This exposes the GPUs at index `0` and `2` — the first and third GPUs listed i
 
 > [!NOTE]
 >
-> GPU access with the `--gpus` flag is documented here for standalone Docker
-> Engine containers on a single host. This page doesn't cover GPU scheduling
-> across multiple Docker Engine hosts or orchestrators.
+> The `--gpus` flag provides GPU access to containers created with
+> `docker run`. To make GPUs available to services in a swarm, advertise them
+> as [node generic resources](/reference/cli/dockerd/#node-generic-resources)
+> and request them with
+> [`--generic-resource`](/reference/cli/docker/service/create/#generic-resources).
 
 ### Set NVIDIA capabilities
 
--- a/content/manuals/engine/containers/gpu.md
+++ b/content/manuals/engine/containers/gpu.md
@@ -88,7 +88,9 @@ This exposes the GPUs at index `0` and `2` — the first and third GPUs listed i
 
 > [!NOTE]
 >
-> NVIDIA GPUs can only be accessed by systems running a single engine.
+> GPU access with the `--gpus` flag is documented here for standalone Docker
+> Engine containers on a single host. This page doesn't cover GPU scheduling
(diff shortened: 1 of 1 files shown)


dvdksn (commented):
This changes the meaning of the callout entirely. Maybe that's what you intended but I'm not able to verify whether the new claim is true. The original note is confusing, but I'm not sure the suggestions improves it much.

codewithdaniel1:
Thanks for the review. I looked into the Docker CLI documentation and updated the note to make the distinction more specific and verifiable.

The updated note now explains that:

- `--gpus` is used with `docker run` / `docker container run`
- Swarm nodes advertise NVIDIA GPUs as generic resources
- Swarm services request those resources with `--generic-resource`

The supporting Docker documentation is here:

- https://docs.docker.com/reference/cli/docker/container/run/#access-an-nvidia-gpu
- https://docs.docker.com/reference/cli/dockerd/#node-generic-resources
- https://docs.docker.com/reference/cli/docker/service/create/#generic-resources

I changed the note to:

> The `--gpus` flag provides GPU access to containers created with
> `docker run`. To make GPUs available to services in a swarm, advertise them
> as node generic resources and request them with `--generic-resource`.

I avoided making a claim about what the original wording historically meant and instead described the currently documented behavior directly. Will this version work?

PR #25742, opened 2026-08-10
docs: mention build cache in system prune overview

locker95:
The “Prune everything” prose only listed images, containers, and networks, while the example warning already includes unused build cache. Spelled that out (and in the page intro) so it matches real behavior, including BuildKit cache mounts.

Related to moby/moby#50141

### DIFF
--- a/content/manuals/engine/manage-resources/pruning.md
+++ b/content/manuals/engine/manage-resources/pruning.md
@@ -10,12 +10,13 @@ aliases:
 ---
 
 Docker takes a conservative approach to cleaning up unused objects (often
-referred to as "garbage collection"), such as images, containers, volumes, and
-networks. These objects are generally not removed unless you explicitly ask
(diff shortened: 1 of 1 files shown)


vvoland (approved):

content/manuals/engine/manage-resources/pruning.md
```diff
@@ -166,8 +167,9 @@ for all options, including `--all` to also remove internal and frontend images.
 ## Prune everything
 
 The `docker system prune` command is a shortcut that prunes images, containers,
-and networks. Volumes aren't pruned by default, and you must specify the
-`--volumes` flag for `docker system prune` to prune volumes.
+networks, and unused [build cache](/manuals/engine/manage-resources/pruning.md#prune-build-cache)
```
dvdksn: I don't see why build cache is linked but the others aren't. I'd be consistent and not add a link here

PR #25743, opened 2026-08-10
docs: clarify what replaces the deprecated self-diagnose tool

### LINKED ISSUE #25740: [docs-scanner] Deprecated self-diagnose tool section lacks guidance on alternatives
**File:** `content/manuals/desktop/troubleshoot-and-support/troubleshoot/_index.md`

### Issue

The "Self-diagnose tool" section contains only a deprecation notice with no additional information:

> ### Self-diagnose tool
> 
> [!IMPORTANT]
> 
> This tool has been deprecated.

This stub section doesn't tell users:
- What to use instead of the self-diagnose tool
- When it was deprecated
- Whether it's been removed or just discouraged
- Which of the other diagnostic methods on the page replaces its functionality

### Why this matters

Users who previously relied on the self-diagnose tool or who find references to it elsewhere need clear guidance on what to do instead. A deprecation notice without migration guidance leaves users without a path forward.

### Suggested fix

Either:
1. Remove this section entirely if the tool is fully replaced by the "Diagnose" section above it, or
2. Add a sentence directing users to the appropriate alternative, such as: "Use the diagnostic methods described in the [Diagnose](#diagnose) section above instead."

If the self-diagnose tool was a specific CLI command or UI feature, mention what specifically was deprecated and what replaces it.

---

nielskaspers:
## Summary
- explain that the deprecated Self-diagnose tool is replaced by the current Diagnose workflows
- point readers to the existing `docker desktop diagnose` command instead of leaving the section as a dead-end

## Testing
- git diff --check

Fixes #25740

### DIFF
--- a/content/manuals/desktop/troubleshoot-and-support/troubleshoot/_index.md
+++ b/content/manuals/desktop/troubleshoot-and-support/troubleshoot/_index.md
@@ -194,7 +194,9 @@ If you don't have a paid Docker subscription, create an issue on [GitHub](https:
 
 > [!IMPORTANT]
 >
-> This tool has been deprecated.
+> This tool has been deprecated. Use the diagnostic methods in
+> [Diagnose](#diagnose) above, including the [`docker desktop diagnose`
+> command](/manuals/desktop/features/desktop-cli.md), instead.
 
 ## Check the logs
 


dvdksn (commented):
🤔 I find this entire section a bit odd. Maybe we'd be better off removing this section entirely? It's not obvious to me what "Self-diagnose tool" refers to. wdyt @aevesdocker

PR #25812, opened 2026-08-12
docs: volume prune default is anonymous only

locker95:
The volumes section of the pruning guide said unused volumes are all removed by default and showed a warning that matched that wording.

CLI default is anonymous-only; named unused volumes need `--all`. Aligned the warning text and a short note so it matches the `docker volume prune` reference.


dvdksn (approved):
Thanks!

PR #25852, opened 2026-08-16
docs: say why GHA Scout compare disables SBOM on PRs

### LINKED ISSUE #25851: [docs-scanner] Contradictory guidance on SBOM/provenance in GitHub Actions workflow
**File:** `content/manuals/scout/integrations/ci/gha.md`

### Issue

The page contains contradictory guidance about SBOM and provenance attestations in the GitHub Actions workflow. The workflow code explicitly enables SBOM and provenance for non-PR builds:

```yaml
sbom: ${{ github.event_name != 'pull_request' }}
provenance: ${{ github.event_name != 'pull_request' }}
```

But immediately after, a note states:

> This comparison doesn't work if you push the image to a registry, or if you build an image that can't be loaded to the runner's local image store. For example, multi-platform images or images with SBOM or provenance attestation can't be loaded to the local image store.

### Why this matters

A reader following this workflow will enable SBOM and provenance (as shown in the code), but then read that images with SBOM or provenance attestation "can't be loaded to the local image store." This creates confusion about whether the workflow will work as intended. The workflow shows one thing, the note warns against it.

Additionally, the note says "This comparison doesn't work if you push the image to a registry" but the workflow does push images on non-PR events (`push: ${{ github.event_name != 'pull_request' }}`), which further contradicts the guidance.

### Suggested fix

Clarify the relationship between the workflow steps and the note. Either:

1. Explain that SBOM/provenance are only enabled for push events (not PRs) because the comparison step only runs on PRs where the image is loaded locally, OR
2. If the workflow is incorrect, update it to disable SBOM/provenance when local loading is required, OR
3. If the note is outdated (e.g., if Docker Scout can now handle these cases), update or remove the note

The workflow and the explanatory note should align rather than contradict each other.

---

locker95:
The workflow turns SBOM/provenance on for pushes and off for PRs, then the next note said those attestations break local compare. That's the same split: compare only runs on PRs, where the image is loaded without attestations.

Fixes #25851

### DIFF
--- a/content/manuals/scout/integrations/ci/gha.md
+++ b/content/manuals/scout/integrations/ci/gha.md
@@ -110,7 +110,7 @@ This creates workflow steps to:
 > and leave `sbom` / `provenance` off so the image can land in the runner's
 > local store.
 >
-> On push events the same job turns SBOM and provenance on and `push`es to
+> On push events the same job turns SBOM and provenance on and pushes to
 > the registry. Don't enable those attestations (or a multi-platform build)
 > on the PR path — those images can't be loaded locally, and compare would
 > have nothing to inspect.
--- a/content/manuals/scout/integrations/ci/gha.md
+++ b/content/manuals/scout/integrations/ci/gha.md
@@ -105,14 +105,15 @@ This creates workflow steps to:
 
 > [!NOTE]
 >
-> This CI workflow runs a local analysis and evaluation of your image. To
-> evaluate the image locally, you must ensure that the image is loaded the
-> local image store of your runner.
+> The Scout compare step only runs on pull requests
+> (`if: github.event_name == 'pull_request'`). Those builds set `load: true`
+> and leave `sbom` / `provenance` off so the image can land in the runner's
+> local store.
 >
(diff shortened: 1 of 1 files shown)


content/manuals/scout/integrations/ci/gha.md
```diff
@@ -105,14 +105,15 @@ This creates workflow steps to:
...
+> The Scout compare step only runs on pull requests
+> (`if: github.event_name == 'pull_request'`). Those builds set `load: true`
+> and leave `sbom` / `provenance` off so the image can land in the runner's
+> local store.
 >
-> This comparison doesn't work if you push the image to a registry, or if you
-> build an image that can't be loaded to the runner's local image store. For
-> example, multi-platform images or images with SBOM or provenance attestation
-> can't be loaded to the local image store.
+> On push events the same job turns SBOM and provenance on and `push`es to
```
craig-osterhout:
```suggestion
> On push events the same job turns SBOM and provenance on and pushes to
```

craig-osterhout (approved):

PR #25874, opened 2026-08-17
docs(rootless): setuptool needs a pam_systemd session

### LINKED ISSUE #14491: dockerd-rootless-setuptool.sh install fails to detect systemd unless $XDG_RUNTIME_DIR is set first
### Problem description

When I follow the rootless installation instructions on Linux Mint 20 (based on Ubuntu 20.04) it fails to detect systemd:

```
dockerd@emvc:~$ dockerd-rootless-setuptool.sh install
[INFO] systemd not detected, dockerd-rootless.sh needs to be started manually:

PATH=/usr/bin:/sbin:/usr/sbin:$PATH dockerd-rootless.sh 

[INFO] Creating CLI context "rootless"
Successfully created context "rootless"

[INFO] Make sure the following environment variables are set (or add them to ~/.bashrc):

# WARNING: systemd not found. You have to remove XDG_RUNTIME_DIR manually on every logout.
export XDG_RUNTIME_DIR=/home/dockerd/.docker/run
export PATH=/usr/bin:$PATH
export DOCKER_HOST=unix:///home/dockerd/.docker/run/docker.sock
```

Setting XDG_RUNTIME_DIR as the script suggests does not help. `systemctl --user` still fails with "Failed to connect to bus: No such file or directory".  Following [this StackOverflow answer](https://unix.stackexchange.com/q/587674/107961), what works is setting XDG_RUNTIME_DIR to /run/user/... before running `dockerd-rootless-setuptool.sh install`:

```
export XDG_RUNTIME_DIR=/run/user/$(id -u)
dockerd-rootless-setuptool.sh install
```

(I actually added it to .bash_profile - that would be helpful to add, too. Unlike .bashrc, .bash_profile gets loaded when using `sudo -iu dockerd`)

(I also enabled lingering for the user first, following that answer, but I don't know if that was strictly necessary.)

### Problem location

https://docs.docker.com/engine/security/rootless/#install

### Suggestions for a fix

I'm not sure if this is a problem with the documentation or the dockerd-rootless-setuptool.sh script itself could handle this better, but assuming the script is not changed, it would help if the documentation page suggested setting XDG_RUNTIME_DIR before running the script, or at least in the troubleshooting section if you get

> [INFO] systemd not detected, dockerd-rootless.sh needs to be started manually

locker95:
If you `sudo su` in and `$XDG_RUNTIME_DIR` is empty (or the `~/.docker/run` fallback the script prints), it says systemd isn't there even on a systemd box. `systemctl --user` needs `/run/user/$UID`.

Called that out on the install page and in the existing troubleshoot section.

@netlify /engine/security/rootless/

Fixes #14491

### DIFF
--- a/content/manuals/engine/security/rootless/_index.md
+++ b/content/manuals/engine/security/rootless/_index.md
@@ -91,6 +91,12 @@ export PATH=/usr/bin:$PATH
 export DOCKER_HOST=unix:///run/user/1000/docker.sock
 ```
 
+Run the installer from a real login session (`ssh` as yourself, or
+`machinectl shell $USER@`). `$XDG_RUNTIME_DIR` must be `/run/user/$(id -u)`.
+If it is empty or something like `~/.docker/run`, the script cannot see
+systemd and tells you to start `dockerd-rootless.sh` by hand. See
+[Troubleshooting](./troubleshoot.md#unable-to-install-with-systemd-when-systemd-is-present-on-the-system).
+
 If `dockerd-rootless-setuptool.sh` is not present, you may need to install the `docker-ce-rootless-extras` package manually, e.g.,
 
 ```console
--- a/content/manuals/engine/security/rootless/troubleshoot.md
+++ b/content/manuals/engine/security/rootless/troubleshoot.md
@@ -110,6 +110,19 @@ $ sudo machinectl shell myuser@
 ```
 Where `myuser@` is your desired username and @ signifies this machine.
 
+The same "systemd not detected" message appears when `$XDG_RUNTIME_DIR` is
+missing or is a fallback such as `~/.docker/run`. On a systemd host it must
+be `/run/user/$(id -u)` so `systemctl --user` can reach the session bus.
+Setting `XDG_RUNTIME_DIR` to the path printed by a failed install does not
+fix that. Log in with `pam_systemd` first (`ssh`, graphical console, or
+`machinectl shell`), then check:
+
+```console
+$ echo $XDG_RUNTIME_DIR
+/run/user/1000
+$ systemctl --user status
+```
+
 ### Errors when starting the Docker daemon
 
 **\[rootlesskit:parent\] error: failed to start the child: fork/exec /proc/self/exe: operation not permitted**


content/manuals/engine/security/rootless/_index.md
```diff
@@ -91,6 +91,12 @@ export PATH=/usr/bin:$PATH
 export DOCKER_HOST=unix:///run/user/1000/docker.sock
 ```
 
+Run the installer from a real login session (`ssh` as yourself, or
+`machinectl shell $USER@`). `$XDG_RUNTIME_DIR` must be `/run/user/$(id -u)`.
+If it is empty or something like `~/.docker/run`, the script cannot see
+systemd and tells you to start `dockerd-rootless.sh` by hand. See
+[Troubleshooting](./troubleshoot.md#unable-to-install-with-systemd-when-systemd-is-present-on-the-system).
```
AkihiroSuda, line 98: This is lengthy and decreases readability.

"The installer will fail if `$XDG_RUNTIME_DIR` is not set. See [Troubleshooting](...).` should be enough.

content/manuals/engine/security/rootless/troubleshoot.md
```diff
@@ -110,6 +110,19 @@ $ sudo machinectl shell myuser@
 ```
 Where `myuser@` is your desired username and @ signifies this machine.
 
+The same "systemd not detected" message appears when `$XDG_RUNTIME_DIR` is
+missing or is a fallback such as `~/.docker/run`. On a systemd host it must
```
AkihiroSuda, line 114: "The same" is misleading, as "when `$XDG_RUNTIME_DIR` is
missing" typically happens when `sudo` is used

content/manuals/engine/security/rootless/_index.md
AkihiroSuda, line 1: We already have helpful error messages in https://github.com/moby/moby/blob/e0c1583c7891b34bd199fdf9e02fb7c7b65bb1ca/contrib/dockerd-rootless-setuptool.sh#L97-L109

PR #25875, opened 2026-08-17
docs(sandboxes): correct the on-disk secret store's protection claim

robmry:
## Description

The Sandboxes credential pages describe the keychain-less Linux fallback as "an encrypted file" and quote a CLI notice that no longer exists. They also frame the fallback as weaker than an OS keychain only because a keychain "mediates access per application".

Both understate it. The file is encrypted with a key compiled into `sbx`, not a secret the user holds, so anyone who can read the file can decrypt it. What actually protects it is the `0700` directory mode — which the pages already mention, so this mostly moves the emphasis onto the claim that holds.

`sbx` now prints:

```text
No keychain detected - this secret will be stored on disk, protected by file permissions rather than a password
```

replacing the previous "…stored in an encrypted file on disk". The quoted strings here are byte-matched against the constant in the CLI.

Four passages change:

- `security/credentials.md` — the fallback description, the quoted notice, and the protection claim
- `security/credentials.md` — the **Best practices** list, which repeated "encrypted at rest in the OS keychain (or an encrypted file on Linux hosts without a keychain)"
- `faq.md` — the headless-Linux entry, which carried the same notice and claim

No behaviour or feature is being documented here — only the accuracy of an existing description.

## Related issues or tickets

The CLI-side change is in `docker/sandboxes` and ships in the next release. Worth holding this until that release is out: until then the page would quote a notice users don't see yet.

## Reviews

- [ ] Technical review
- [ ] Editorial review
- [ ] Product review

Two things a reviewer may want to weigh in on:

- **Tone.** The intent is to be accurate without being alarming: it states what protects the file and what the encryption does and does not give you, rather than editorialising about weakness. Happy to soften or sharpen.
- **Adjacent advice, unchanged here.** The FAQ paragraph immediately below the edited text tells headless users to install `gnome-keyring` and start `dbus-run-session`. A *locked* Secret Service collection is known to make `sbx` commands hang rather than fall back, and on a key-only SSH host locked is the default state after the keyring daemon restarts — so that advice can land users somewhere worse than the fallback. It is left alone because the underlying issue is being worked separately, but it may deserve a caveat once that is resolved.

<sub>Investigated, authored and posted by Claude Code on behalf of @robmry.</sub>

### DIFF
--- a/content/manuals/ai/sandboxes/security/credentials.md
+++ b/content/manuals/ai/sandboxes/security/credentials.md
@@ -59,22 +59,25 @@ The store backing `sbx secret set` depends on your operating system:
 The Ubuntu package depends on GNOME Keyring, so a standard desktop install
 needs no extra setup.
 
-On Linux hosts without a running Secret Service — headless servers and some
-WSL setups — `sbx` falls back to an encrypted file under your user config
-directory `$XDG_CONFIG_HOME/com.docker.sandboxes`, which defaults to
+On Linux hosts without a running Secret Service — headless servers and some WSL
+setups — `sbx` falls back to a file under your user config directory
+`$XDG_CONFIG_HOME/com.docker.sandboxes`, which defaults to
 `~/.config/com.docker.sandboxes` when `$XDG_CONFIG_HOME` is unset. The fallback
 is automatic and needs no configuration. When you store a secret this way,
 `sbx` prints a notice:
 
 ```text
-No keychain detected - this secret will be stored in an encrypted file on disk
+No keychain detected - this secret will be stored on disk, protected by file permissions rather than a password
 ```
 
-The file is encrypted at rest and protected by `0700` directory permissions,
-the same posture as `~/.docker/config.json`. This is weaker than an OS
-keychain, which also mediates access per application. If you start a Secret
-Service on the host later, `sbx` stores new secrets in the keychain again. For
-more on running sandboxes without a desktop keyring, see
+`sbx` stores the file in a directory with `0700` permissions, the same
+file-permission model used for `~/.docker/config.json`. Any user or process that
+can read the file can retrieve the stored credentials, so treat the directory as
+sensitive. Where available, prefer a keychain, which mediates access per
+application.
+
+If you start a Secret Service on the host later, `sbx` stores new secrets in the
+keychain again. For more on running sandboxes without a desktop keyring, see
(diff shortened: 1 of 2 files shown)


content/manuals/ai/sandboxes/security/credentials.md
```diff
@@ -59,22 +59,27 @@ The store backing `sbx secret set` depends on your operating system:
...
-No keychain detected - this secret will be stored in an encrypted file on disk
+No keychain detected - this secret will be stored on disk, protected by file permissions rather than a password
 ```
 
-The file is encrypted at rest and protected by `0700` directory permissions,
-the same posture as `~/.docker/config.json`. This is weaker than an OS
-keychain, which also mediates access per application. If you start a Secret
-Service on the host later, `sbx` stores new secrets in the keychain again. For
-more on running sandboxes without a desktop keyring, see
+What protects that file is the `0700` permissions on the directory holding it,
```
dvdksn: The opening doesn’t connect naturally to the preceding text, and “stops the values being read at a glance” is vague. Could we state the protection model directly?

Suggested replacement:

> `sbx` stores the file in a directory with `0700` permissions, the same file-permission model used for `~/.docker/config.json`. Any user or process that can read the file can retrieve the stored credentials, so treat the directory as sensitive. Where available, prefer a keychain, which mediates access per application.

The FAQ repeats the same “What protects…” construction, so please update that passage to match.

Generated by Codex

robmry: Applied as suggested, in both files — thank you, this reads much better. Your version states the consequence for the reader ("any user or process that can read the file can retrieve the stored credentials") where mine described the mechanism and left them to draw the conclusion.

The FAQ passage now uses the identical paragraph, so the two pages no longer diverge in how they explain it.

One consequence worth surfacing rather than leaving implicit: this wording drops any mention that the file is encrypted. That is deliberate on our side and we are happy with it — the encryption is under a key compiled into `sbx` rather than a secret the user holds, so it is not a protection worth documenting, and naming it tends to invite more trust than it earns. Flagging it only so it is a recorded decision rather than something that looks lost in an edit.

Also reflowed one line above the notice that an earlier revision of mine had left wrapped awkwardly.

<sub>Investigated, authored and posted by Claude Code on behalf of @robmry.</sub>

dvdksn (approved):

PR #25881, opened 2026-08-18
docs: say build.network is a mode, not a Compose network

### LINKED ISSUE #21982: docker image build --network refers to netwoking type
### Is this a docs issue?

### Type of issue

Information is incorrect

### Description

The `--network` parameter of `docker build image` actually does not refer to a network, but to the [networking mode for the RUN instructions during build](https://docs.docker.com/reference/cli/docker/buildx/build/#network).
The semantics is not consistent in this case (and it sucks), since in `docker run` , the `--network` connects the container to a network (it should be called `--networking-mode` in `docker build image`).

In the compose docs there is an example showing how to add a custom network during the build phase (which is wrong) and doesn't explain that this parameter is actually for the networking mode (none, host, bridge):

```yaml
build:
  context: .
  network: custom_network_1
```

Some references:
- [Is "docker build --network=host" working as expected? - #3 by nickyayoub](https://forums.docker.com/t/is-docker-build-network-host-working-as-expected/33492/3)
- [How to handle networking between containers at build time](https://forums.docker.com/t/how-to-handle-networking-between-containers-at-build-time/139018)](https://forums.docker.com/t/how-to-handle-networking-between-containers-at-build-time/139018)
- [DOCKER_BUILDKIT=1 prevents custom networks using docker build --network](https://github.com/moby/buildkit/issues/978)

### Location

https://docs.docker.com/reference/compose-file/build/#network

### Suggestion

- Remove `custom_network_1`
- Clarify regarding `service.build.network` being networking mode and different from `service.network`

locker95:
build.network was showing `custom_network_1`, which looks like a user-defined Compose network. it's the RUN networking mode (`default` / `none` / `host`), same as `docker build --network`.

@netlify /reference/compose-file/build/

Fixes #21982

### DIFF
--- a/content/reference/compose-file/build.md
+++ b/content/reference/compose-file/build.md
@@ -348,7 +348,11 @@ build:
 
 ### `network`
 
-Set the network containers connect to for the `RUN` instructions during build.
+Set the networking mode for `RUN` instructions during the build. This is
+the same as [`docker build --network`](/reference/cli/docker/buildx/build/#network).
+It is not a Compose network from the top-level [`networks`](networks.md) key.
+
+The value is a mode (`default`, `none`, or `host`), not a network name:
 
 ```yaml
 build:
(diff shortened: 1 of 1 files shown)


aevesdocker (approved):
Thanks

PR #25911, opened 2026-08-21
docs(sandboxes): recommend the t3code kit for T3 Code prerequisites

mdelapenya:
## Description

The T3 Code integration page still documents the manual `apt-get` toolchain install as the only way to prepare a sandbox, and its persistence note says to "add the package to a custom image or kit instead" — that kit exists now. The [`t3code` kit](https://github.com/docker/sbx-kits-contrib/tree/main/t3code) (published as `docker.io/sbx/t3code-kit:latest`) installs the build toolchain and the `t3` npm package at sandbox creation, so the first connection starts a pre-installed server instead of building `node-pty` from source.

This updates the Prerequisites section to recommend the kit, keeps the manual install as the fallback for existing sandboxes, and adjusts the first-connection note in the Connect section accordingly.

## Related issues or tickets

Follows up docker/sbx-kits-contrib#223, which added the kit.

## Reviews

- [ ] Technical review
- [ ] Editorial review

### DIFF
--- a/content/manuals/ai/sandboxes/integrations/t3-code.md
+++ b/content/manuals/ai/sandboxes/integrations/t3-code.md
@@ -21,7 +21,20 @@ that tunnels back to the app.
 The first connection installs the T3 server in the sandbox, which needs a
 build toolchain. T3 depends on `node-pty`, which ships prebuilt binaries only
 for macOS and Windows. On a Linux sandbox, `node-pty` compiles from source and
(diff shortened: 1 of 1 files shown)


dvdksn (commented):
~I think we can probably remove the manual steps from here and just recommend the mixin that installs the dependencies. It should work for existing sandboxes with sbx kit add, right?~

Hmm never mind, I saw the linked internal issue.

dvdksn (approved):

PR #25942, opened 2026-08-27
docs: fix incorrect description in local cache page

sunm2n:
## Description

The `local` cache page's front matter description reads
"Manage build cache with Amazon S3 buckets", which belongs to the
S3 cache page and is unrelated to the local cache backend.

Updated it to match the wording used by the other storage-location
backends (`azblob`, `registry`, `s3`), which all follow the
"Manage build cache with ..." pattern.

This description appears in search results and page metadata.

## Reviews

- [x] Editorial review


dvdksn (approved):
thanks

PR #25943, opened 2026-08-27
docs: document tag parameter and restructure local cache versioning

sunm2n:
## Description

The `tag` parameter for the `local` cache backend was added in
[moby/buildkit#3111](https://github.com/moby/buildkit/pull/3111) but was
never documented. The page also carried a `FIXME` comment waiting on that
PR, which has long since been released.

Changes:
- Add `tag` to the parameter table
- Document tag-based scoping in the "Cache versioning" section
- Remove the stale `FIXME` comment

Verified against `client/solve.go` and by running builds locally:
- Exports are annotated with `org.opencontainers.image.ref.name`,
  defaulting to `latest`
- Different tags coexist in the same directory as separate manifests
- `--cache-from` with a matching tag restores the cache on a fresh builder
- When `digest` is set, `tag` is ignored

## Reviews

- [x] Technical review
- [x] Editorial review


dvdksn (approved):
Thank you @sunm2n

PR #25978, opened 2026-08-31
Make a python3 example and not a python2 example.

chrisdecker1201:
## Description

Make a python3 example and not a python2 example.

Python 2 is not maintened anymore. I think it is better to just use a Python 3 example.


dvdksn (approved):
thx!

PR #25996, opened 2026-09-01
docs: document Hub API custom role assignment for members and teams

akristen:
## Summary
- Document that member `PUT` `role` accepts a custom role **name**, not only `owner` | `editor` | `member`.
- Document `role` on team `PATCH` (core or custom role name) and on `org_group` so the PATCH 200 can include the assigned role.
- Add `Editor` to `org_member.role` (title case). Live responses already return it; the old enum was Owner / Member / Invitee only.

## Test plan
PM must confirm before merge:

- [ ] Member PUT 200: after assigning a custom role (body is not the previous core role), is `role` still a core value (`Owner` / `Editor` / `Member`), or is it the custom role **name**? If it is the custom name, drop the `org_member.role` enum in a follow-up and describe a string (core or custom name). If it stays a core value, this PR's enum (`Owner` | `Editor` | `Member` | `Invitee`) is enough.
- [ ] List members: same field. Confirm a user with a custom role lists as a core value vs the custom name.
- [ ] Casing: request stays lowercase (`editor`); response stays title case (`Editor`). Confirm that split is real.
- [ ] Team PATCH 200: confirm `org_group.role` is present and is the value just set (core or custom name).
- [ ] Team assignment is PATCH-only, not PUT.
- [ ] Unknown role names return an error; do not document them as valid.
- [ ] Custom role **name** (not label or UUID) is the request identifier, matching Docker Home.

### DIFF
--- a/content/reference/api/hub/changelog.md
+++ b/content/reference/api/hub/changelog.md
@@ -15,6 +15,17 @@ issues for Docker Service APIs.
 
 ---
 
+## 2026-09-01
+
+### Updates
+
+- Document that [Update org member (role)](/reference/api/hub/latest/#tag/orgs)
+  accepts a custom role name in `role`.
+- Document the `role` field on
+  [Update some details for an organization group](/reference/api/hub/latest/#tag/groups).
+
+---
+
 ## 2026-08-27
 
 ### Updates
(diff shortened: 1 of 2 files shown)


gmondello (approved):
This looks good to me, @akristen. Thank you!

gurleensethi-docker (approved):

aevesdocker (approved):
LGTM

PR #26062, opened 2026-09-11
build: explain GitHub Actions cache write restrictions

### LINKED ISSUE #1571: Intermittent "error writing layer blob: failed to reserve cache" during GHA cache export (gha exporter) — not quota, not permissions
### Contributing guidelines

- [x] I've read the [contributing guidelines](https://github.com/docker/build-push-action/blob/master/.github/CONTRIBUTING.md) and wholeheartedly agree

### I've found a bug, and:

- [x] The documentation does not mention anything about my problem
- [x] There are no open or closed issues that are related to my problem

### Description

We're seeing intermittent, recurring failures during the GitHub Actions cache export step across multiple repositories, all using `type=gha`. The build and image push both complete successfully — the failure happens specifically during cache export, on what appears to be the very first blob write of the export.

This has been happening for a while now (previously worked reliably for months before failures started appearing), so it does not appear to be tied to a recent config change on our end.

### Environment

- `docker/build-push-action@v6` (also reproduced on latest `@v7`)
- `docker/setup-buildx-action@v4`
- Buildx `v0.35.0`
- BuildKit `v0.30.0`
- GitHub-hosted `ubuntu-latest` runners
- Docker Engine 28.0.4

### Workflow (relevant excerpt)

\`\`\`yaml
- name: build-push
  uses: docker/build-push-action@v6
  with:
    tags: <ecr-registry>/<repo>:<tag>
    context: .
    cache-from: type=gha
    cache-to: type=gha,mode=max
    push: true

\`\`\`

### Error log

\`\`\`
#17 exporting to image
#17 pushing manifest ... done
#17 DONE 33.5s
#18 exporting to GitHub Actions Cache
#18 preparing build cache for export 21.8s done
#18 sending cache export
#18 sending cache export 0.0s done
#18 writing layer sha256:............................................ 0.0s done
#18 ERROR: error writing layer blob: failed to reserve cache
------
 > exporting to GitHub Actions Cache:
------
ERROR: failed to build: failed to solve: error writing layer blob: failed to reserve cache
Error: buildx failed with: ERROR: failed to build: failed to solve: error writing layer blob: failed to reserve cache
\`\`\`

Notably, only **one blob** ever completes (`writing layer sha256:... done`) before the very next write fails — this pattern has been consistent across multiple separate failed runs, with different blob hashes each time. Image export and push (`#17`) complete without issue every time; only the cache export step (`#18`) fails.

### What we've ruled out

- **Cache storage quota** — repo usage sits at ~3.8GB of the 10GB limit at time of failure (confirmed via `GET /repos/{owner}/{repo}/actions/cache/usage`), so this isn't an eviction/capacity issue.
- **Token permissions** — added `actions: write` to the job's `permissions:` block; no change in behavior.
- **Concurrent builds / scope collision** — this repo never runs concurrent builds (single build per merge, no matrix, no parallel triggers), and only builds one image, so default `scope=buildkit` shouldn't be contended.
- **Retry-specific collision** — reproduced on fresh, non-retried runs as well, not just re-runs of failed attempts.
- **Outdated tooling / legacy v1 cache API** — Buildx/BuildKit versions are current; not seeing the known `"This legacy service is shutting down..."` error string associated with the v1 sunset.
- **Action version** — reproduced on both `docker/build-push-action@v6` and `@v7`.

### Expected behaviour

The build completes and the image pushes successfully. The GitHub Actions cache export either succeeds, or — since caching is a best-effort optimization, not a correctness requirement — a failure during cache export should not fail the entire build when the image itself has already been built and pushed successfully.

### Actual behaviour

The image builds and pushes correctly (#17 exporting to image completes, layers push, manifest pushes). Immediately after, cache export (#18 exporting to GitHub Actions Cache) fails after writing only one layer blob, aborting the whole workflow run with a non-zero exit even though the actual deliverable (the pushed image) is already complete and valid.

### Repository URL

_No response_

### Workflow run URL

_No response_

### YAML workflow

```yaml
- name: Set up Docker Buildx
        uses: docker/setup-buildx-action@v4

      - name: Login into ECR
        id: login-ecr
        uses: aws-actions/amazon-ecr-login@v2
  
      - name: build-push
        uses: docker/build-push-action@v6
        env:
          ECR_REGISTRY: ${{ steps.login-ecr.outputs.registry }}
          IMAGE_TAG: ${{ env.version_name }}
        with:
          tags: ${{env.ECR_REGISTRY}}/${{env.ECR_REPOSITORY}}:${{env.IMAGE_TAG}}
          context: .
           cache-from: type=gha
           cache-to: type=gha,mode=max
          push: true
```

### Workflow logs

#18 ERROR: error writing layer blob: failed to reserve cache
------
 > exporting to GitHub Actions Cache:
------

ERROR: failed to build: failed to solve: error writing layer blob: failed to reserve cache
Reference
Check build summary support
Error: buildx failed with: ERROR: failed to build: failed to solve: error writing layer blob: failed to reserve cache

### BuildKit logs

```text

```

### Additional info

_No response_

crazy-max:
## Description

Document GitHub Actions cache write restrictions that can cause cache export to fail after a successful image push, including workflows triggered when a pull request is merged into the default branch. Explain how to restore cache in read-only workflows and populate it from a workflow with write access.

Clarify how `cache-mode` overrides affect access and why `ignore-error=true` suppresses export failures without fixing denied writes.

## Related issues or tickets

closes https://github.com/docker/build-push-action/issues/1571

### DIFF
--- a/content/manuals/build/cache/backends/gha.md
+++ b/content/manuals/build/cache/backends/gha.md
@@ -80,6 +80,11 @@ GitHub's [cache access restrictions](https://docs.github.com/en/actions/advanced
 still apply. Only the cache for the current branch, the base branch and the
 default branch is accessible by a workflow.
 
+Cache writes also depend on the workflow's cache access. Some events receive
+read-only access in the default-branch context. See
+[Cache write restrictions](../../ci/github-actions/cache.md#cache-write-restrictions)
+for affected triggers and how to configure cache imports and exports.
+
 ## Version
 
 If you don’t set `version` explicitly, the default is v1. However, if the environment variable `$ACTIONS_CACHE_SERVICE_V2` is set to a value interpreted as `true` ( `1`, `true`, `yes`), then v2 is used automatically.
--- a/content/manuals/build/ci/github-actions/cache.md
+++ b/content/manuals/build/ci/github-actions/cache.md
@@ -1,5 +1,6 @@
 ---
 title: Cache management with GitHub Actions
+description: Configure build cache backends for GitHub Actions and handle cache write restrictions.
 linkTitle: Cache management
 keywords: ci, github actions, gha, buildkit, buildx, cache
 ---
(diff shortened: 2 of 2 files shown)


dvdksn (approved):
Nice

PR #26079, opened 2026-09-11
docs(hub): use 'an SSH' for build keypair step

mrchatam:
## Description
Fix article: "a SSH keypair" → "an SSH keypair" in Docker Hub automated builds troubleshooting.

## Related issues
n/a


craig-osterhout (approved):
thank you

PR #26087, opened 2026-09-14
docs: fix install options on the Raspberry Pi OS 32-bit page

### LINKED ISSUE #26082: Confusing for RasPi 32bit
### Is this a docs issue?

### Type of issue

### Description

The page has the note
"This installation instruction refers to the 32-bit (armhf) version of Raspberry Pi OS."
(which is what most people will have if they buy a pre-installed SD card with their Pi)

But then continues down to a section with 4 installation options [ Desktop for linux , apt repo , manual , convenience script ] but most of these links seem to go off to generic pages, particularly the Desktop option (which I'm not even sure is a thing for a 32 bit Pi) goes to a page that says 64bit x86/amd is a requirement.

Also, the top alert box says "32-bit ARM (v7): Install the Debian armhf packages (targets ARMv7 CPUs).". This flummoxed me because it suggested I needed to do a specific thing with no link to that thing. Reading on I can see that "that thing" is in fact the very page I'm looking at - but it took a while for that to be clear.

Sorry if this all makes me seem quite dumb, but I'm a seasoned software professional who has (mis?) spent their career in DOS, so I'm having to learn linux, cope with the notoriously flakey Pi and learn a new technology all at once. I really could use a bit of spoon feeding from the actual "this is how to do stuff" pages.

### Location

https://docs.docker.com/engine/install/raspberry-pi-os/

### Suggestion

Reword as
"32-bit ARM (v7): Install the Debian armhf packages (targets ARMv7 CPUs). as described below on this page"

And remove the Docker Desktop option if it is true that it is not an option for Pi 32 (or clarify 64 bit only) or make a new page for Pi or explain that even though the page is takes you to says 64bit x86, it will still be fine on a Pi.

sunm2n:
## Description

Two problems on the Raspberry Pi OS (32-bit / armhf) install page, both
reported in #26082.

**Docker Desktop isn't a supported option here.** It was listed first under
"Installation methods" as "the easiest and quickest way to get started", but
[Docker Desktop for Linux](https://docs.docker.com/desktop/setup/install/linux/#supported-platforms)
only provides packages for x86_64 / amd64. The supported platforms table has
no ARM column, and the page doesn't mention ARM anywhere. Both 32-bit and
64-bit Raspberry Pi OS are ARM, so neither qualifies. Removed from the list.

**The 32-bit migration path had no link.** Both deprecation notices offer
Debian packages as the migration path, and the 64-bit entry links to the
Debian installation instructions. The 32-bit entry doesn't, so a reader is
told to "install the Debian `armhf` packages" with nowhere to go. That matters
here because this page installs from the raspbian repository
(`download-url-base` in the front matter), not the Debian one, so the notice
can't be read as pointing at the instructions further down this page. Added
the same link to the 32-bit entry in both notices.

`debian.md` lists `armhf (arm/v7)` among its supported architectures and
includes `armhf` in its package selection step, so the link lands somewhere
that covers this case.

## Related issues or tickets

Fixes #26082

## Reviews

- [ ] Technical review
- [ ] Editorial review
- [ ] Product review


sunm2n:
@dvdksn mind taking a look when you get a chance? CI is passing, and GitHub shows no merge conflicts with main.

The main thing worth your eye is removing the Docker Desktop bullet. The
[supported platforms table](https://docs.docker.com/desktop/setup/install/linux/#supported-platforms)
lists only x86_64 / amd64, so Desktop doesn't look like a supported install
option for either 32-bit or 64-bit Raspberry Pi OS. The other change adds the
missing Debian installation links to the two 32-bit migration notices.

Happy to adjust if there's context I'm missing.

PR #26144, opened 2026-09-21
cli: add reference pages for kubernetes status and reset-cluster

mat007:
## Description

`docker desktop kubernetes status` and `docker desktop kubernetes reset-cluster` are implemented in desktop-cli but had no reference page: `data/cli/desktop` was missing their YAML, which is what the reference pages are generated from. Adds both files and lists them in the parent command's `cname`/`clink` (also fixes `clink` pointing at a nonexistent `docker_desktop_engine_images.yaml`).

## Related issues or tickets

https://docker.atlassian.net/browse/DDB-567

## Reviews

- [ ] Technical review
- [ ] Editorial review


aevesdocker (approved):
Thanks @mat007

